PR #102, opened 2026-04-27
fix: enforce CLI pin on live session and all code paths

clubanderson:
## Summary
- `hive pin` now restarts the systemd service immediately so the running session picks up the pinned CLI
- `hive pin` creates the governor `model_lock_<agent>` file; `hive unpin` removes it
- Governor reads `AGENT_CLI_PINNED=true` from env as fallback guard
- Supervisor rate-limit failover skips pinned agents

## Context
Scanner was pinned to copilot but continued running claude because the pin only wrote to the env file without restarting the session or blocking the governor.

## Test plan
- [ ] `hive pin scanner` restarts session and creates lock file
- [ ] `hive unpin scanner` removes lock, governor resumes
- [ ] Pinned agent rate limit sends alert but does not switch


clubanderson:
Closing — the critical pin enforcement paths are now covered by recent PRs (#223, #227). The remaining gap (immediate session restart on pin) isn't causing issues in practice. Can reopen if needed.

PR #894, opened 2026-06-03
fix: increase stale_timeout to exceed cadence intervals

clubanderson:
stale_timeout was shorter than cadence for guide/scanner/quality, causing false stale detection every kick cycle.


clubanderson:
Superseded by #893 which fixed this more comprehensively

PR #1251, opened 2026-06-06
fix: ACMM level persists across pod restarts

clubanderson:
## Summary

- Config file (`hive.yaml`) now takes priority over `HIVE_LEVEL` env var on restarts
- `HIVE_LEVEL` env var is only used as fallback for initial provisioning (when no level is persisted)
- Removed shutdown backup handler that could overwrite PVC backup with stale config

## Root cause

`HIVE_LEVEL=3` env var in the deployment spec was checked **first** on every restart, overriding the user's saved `acmm_level: 6` in config. Other config settings (repos, governor, etc.) were not affected because they have no env var override.

## Test plan

- [ ] Set ACMM to level 6 via dashboard
- [ ] Delete pod (force restart)
- [ ] Verify ACMM remains at level 6 after restart
- [ ] Repeat 5x to confirm persistence

### DIFF
--- a/v2/cmd/hive/main.go
+++ b/v2/cmd/hive/main.go
@@ -736,19 +736,20 @@ func main() {
 			}
 		}
 	} else {
-		// Determine the ACMM level from env var, config, or saved state
+		// Config file is authoritative on restarts; HIVE_LEVEL env var is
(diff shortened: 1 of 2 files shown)


clubanderson:
Wrong branch — superseded by #1252

Issue #1581, opened 2026-06-17
permissions watcher: chmod spam on agent-owned files + goose 1.37 log dir not pre-created


hanthor:
## Summary

The permissions watcher (#1567/#1568) has two issues in production:

### Issue 1: chmod spam on agent-owned files

The watcher runs `fixEntry()` on every file under `/data/home/`, attempting `os.Chmod()` on files owned by OTHER users. Since the watcher runs as `dev` (uid 1001) but agents create files as their own users (e.g., `hive-scanner`, uid 2007), the chmod fails with:

```
permissions watcher: chmod file failed
  path="/data/home/.local/state/goose/logs/cli/2026-06-17/20260617_041606.log"
  error="operation not permitted"
```

This fires every 10 seconds on every agent-owned file, filling logs with noise.

**Fix:** In `fixEntry()`, skip chmod/chown when the file is not owned by root AND is not owned by the current user.

### Issue 2: Goose 1.37.0 log directory not pre-created

Goose 1.37.0 uses `tracing-appender` for rolling file logging. On startup, it tries to create files under `~/.local/state/goose/logs/cli/YYYY-MM-DD/`. If the parent directory tree doesn't exist, goose panics with:

```
initializing rolling file appender failed: InitError { context: "failed to create log file", source: Os { code: 13, kind: PermissionDenied, message: "Permission denied" } }
Error: goose-cli main thread panicked
```

This causes ALL agents using the goose backend to fail on startup. The hub detects "prompt timeout" and "failed to send kick" for every agent.

**Fix:** Add `os.MkdirAll("/data/home/.local/state/goose/logs/cli/", 0o770)` in the permissions watcher's `ensureWatchedDirs()`, or pre-create the directory in the Dockerfile.

## Reproduction

1. Deploy a fresh hive pod with goose 1.37.0
2. Observe agent startups — if any goose-backed agents exist, they panic
3. Observe permissions watcher logs — chmod spam on agent-owned files every 10s

## Severity

Issue 1: Low (cosmetic log noise)
Issue 2: Critical (all goose agents fail to start)

PR #1582, opened 2026-06-17
fix: permissions watcher — skip agent-owned files and pre-create goose logs dir

hanthor:
Two fixes for the permissions watcher (#1581):

### Fix 1: Skip chmod on agent-owned files

The watcher attempted `os.Chmod()` on every file under `/data/home/`, including files owned by agent users (e.g., `hive-scanner`, uid 2007). Since the watcher runs as `dev` (uid 1001), chmod fails with "operation not permitted", producing log spam every 10 seconds:

```
permissions watcher: chmod file failed
  path="/data/home/.local/state/goose/logs/cli/2026-06-17/20260617_041606.log"
  error="operation not permitted"
```

Now skips files not owned by root (pre-chown) or the dev user.

### Fix 2: Pre-create goose 1.37 log directory

Goose 1.37 uses `tracing-appender` for rolling file logging. On first startup, it tries to create files under `~/.local/state/goose/logs/cli/YYYY-MM-DD/`. If the parent directory tree doesn't exist, goose panics:

```
initializing rolling file appender failed: InitError { context: "failed to create log file", source: Os { code: 13, kind: PermissionDenied, message: "Permission denied" } }
Error: goose-cli main thread panicked
```

This causes ALL goose-backed agents to fail on startup (ci-maintainer, sec-check, supervisor, outreach, guide, architect via pi→goose mapping, scanner via goose).

The watcher now pre-creates `/data/home/.local/state/goose/logs/cli/` in `ensureWatchedDirs()`.

## Testing

- [x] Compiles: `go build ./pkg/agent/`
- [x] Unit tests pass
- [x] Verified in production: hive hub with 9 agents, no panics, no chmod spam

### DIFF
--- a/v2/pkg/agent/permissions_watcher.go
+++ b/v2/pkg/agent/permissions_watcher.go
@@ -36,6 +36,11 @@ var WatchedHomeDirs = []string{
 	"/data/home/.local",
 }
 
+// GooseLogsDir is the rolling log directory goose 1.37 creates on startup.
+// Goose panics if this directory doesn't exist, so the watcher ensures it
+// is created at startup with correct permissions.
+const GooseLogsDir = "/data/home/.local/state/goose/logs/cli"
+
 // StartPermissionsWatcher runs a background goroutine that periodically
 // scans WatchedHomeDirs and fixes files/directories that were created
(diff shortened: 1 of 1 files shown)


clubanderson:
Thank you!

PR #1806, opened 2026-07-08
docs: copy-config precedence gotcha in migration guide

clubanderson:
Follow-up to #1805: documents the old-template copy-config init container that prefers the stale PVC backup over the ConfigMap, silently reverting config edits on restart (hit live during the kellyaa migration), with the two-step fix.


clubanderson:
Superseded by #1807 — same gotcha, better placement (inline callout at the ConfigMap step).

PR #2118, opened 2026-07-28
📝 hub landing: move the comparison section up, right after the hero

clubanderson:
Moves **"The only open-source one in the room"** to directly after the hero and before the YouTube video.

## Why

It was buried below the fold — after the video, the loop rail, and the proof cards. It's the sharpest differentiator on the page: everyone now ships an agent that opens PRs, but almost none are free, open-source, and self-hosted. That argument should land while the hero's claim is still on screen.

## New order

```
Hero
How Hive compares     <- moved here
Watch (video)
How it works
Proof / Options
Get started
...
```

Pure block move — no copy changes, the section is byte-identical.

## Verification

- 13/13 `<section>` open/close tags balanced; HTML parses with no unclosed tags
- Rendered DOM order confirmed over CDP: `product` → `compare` → `watch`
- First `<h2>` after the hero is now "The only open-source one in the room"

Applied to all four long-lived branches.

### DIFF
--- a/v2/pkg/hub/static/index.html
+++ b/v2/pkg/hub/static/index.html
@@ -659,73 +659,6 @@
       </div>
     </section>
 
-    <!-- ── Watch: latest presentation ── -->
-    <section class="section-pad video-section" id="watch">
-      <div class="section-heading" style="margin-left:auto;margin-right:auto">
-        <p class="section-label">Watch</p>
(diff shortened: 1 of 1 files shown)


clubanderson:
Closing — the branch went stale once the hero-copy and typo PRs merged into the same file, leaving it CONFLICTING. Reopening the same block move rebased on the current tip.

Issue #2253, opened 2026-07-30
bob shell writing to read-only folder


joewxboy:
bob shell cannot properly run since it cannot write its configuration:

You’re basically seeing Bob-Shell trying to write state somewhere it doesn’t have write access.

Those errors in the Supervisor terminal —

“There was an error saving your latest settings changes.”
“Failed to initialize logger:”
“Do you trust this folder? … Trusting a folder allows Bob-Shell to execute commands it suggests.”

— all point to Bob trying to:

1. Write its own config / trust state (e.g. “trusted folders”), and

2. Initialize a log file,

but the filesystem inside the pod is read‑only or owned by a different user.

In a Hive deployment, that usually comes down to one of:

- The workspace or repo mount is read‑only to the bob user.

- Bob’s config directory (e.g. something like ‎⁠~/.config/bob⁠ or ‎⁠.bob⁠ in the project) lives on a read‑only volume.

- The container is running as a non‑root user that doesn’t match the ownership on the mounted volume.

What to actually change

At a high level: give the bob process a writable directory for its config/logs and ensure the repo workspace it operates in is writable by that same user.

Concretely, in your Kubernetes / Hive deployment:

1. Make sure the workspace volume is writable.
The directory where the repo is checked out (what Bob calls the “folder” it wants to trust) must be mounted with write access for the bob container user. No ‎⁠readOnly: true⁠ on that volume mount.

2. Fix user / group and permissions.

 - Set ‎⁠securityContext.runAsUser⁠ and ‎⁠fsGroup⁠ so the process user can write to the mounted path.

 - Or ‎⁠chown⁠/‎⁠chmod⁠ the workspace path in an initContainer so it’s owned (or group‑writable) for the bob user.

3. Give Bob a writable config dir.
If Hive’s chart or deployment lets you set a ‎⁠BOB_CONFIG_DIR⁠ / similar env var, point it at a path on a writable volume (e.g. ‎⁠/data/bob⁠), and ensure that directory exists and is writable.

4. Restart the Supervisor pod after fixing perms.
Once the container filesystem is writable, the next run should allow:

 - Logger to initialize.

 - “Trust folder” choice to be saved, so that prompt stops reappearing.

As long as the pod’s filesystem remains effectively read‑only to the bob process, Hive’s Supervisor agent will keep re‑prompting for trust and failing to persist settings. The fix is entirely on the container/volume permissions side, not in the Hive UI or in Bob’s CLI commands.

[...]

PR #2256, opened 2026-07-30
🐛 fix(agent): make tmux pane inherit session env so bob can authenticate

clubanderson:
## Summary

Investigated #2253. The reported symptoms are real, but the diagnosis in the issue is **not** what the container shows. The actual defect is a tmux environment-ordering bug that prevents bob from ever authenticating; the "read-only folder" framing does not hold up.

`Refs #2253` rather than `Fixes` — see *What remains* below.

## What the reporter hypothesized vs. what is actually true

| Hypothesis in the issue | Verdict | Evidence |
|---|---|---|
| Bob's config dir is on a read-only volume | **Wrong** | `/data/home/.bob` is `drwxrwx--- dev:node`. The agent user can write it: `su -s /bin/sh -c "test -w /data/home/.bob/settings.json" hive-scanner` succeeds. bob actively rewrites `settings.json` at runtime. |
| The workspace mount is `readOnly` to the bob user | **Wrong** | Agent workspaces are `/data/agents/<name>`, owned by the agent UID. `.bob/.bob-errors/errors-*.log` is actively being written there (708 KB in a day), owned `hive-scanner:node`. |
| Container runs as a user that mismatches volume ownership | **Wrong** | Per-agent UIDs (2001+) all share group `node` (gid 1000); the dirs are group-writable. `StartPermissionsWatcher` already reconciles ownership on a 10s tick. |
| Needs `securityContext` / `fsGroup` / initContainer `chown` | **Not needed** | Nothing is unwritable, so there is nothing to chown. These are also non-starters on OpenShift here: `NET_ADMIN` and `CAP_CHOWN` are both denied. No deployment change is required by this PR. |
| Needs a `BOB_CONFIG_DIR`-style env var | **Not needed** | bob derives its state dir from `os.homedir()`; `$HOME=/data/home` already points at a writable volume. |
| "Do you trust this folder?" banner | **Already fixed** | #2249 added `--trust`; the banner no longer appears. Not re-fixed here. |

## The actual root cause

`ensureTmuxSession` creates the session and only *then* applies the environment:

```
new-session -d   ->   set-environment (xN)   ->   send-keys
```

tmux copies the session environment only into processes it forks **afterwards**. `new-session -d` has already forked the pane's shell, so that shell predates every `set-environment` and never receives any of them.

This silently drops exactly the `Secret: true` pairs — `BOBSHELL_API_KEY`, `CLAUDE_CODE_OAUTH_TOKEN` — because `buildEnvPrefix` deliberately keeps those off the command line so credentials never land in `ps` or pane scrollback. Non-secret vars survive only because they are re-typed on every launch by `buildEnvPrefix`.

Net effect: the session environment genuinely holds the key (`show-environment` lists it) while no bob process does, so bob falls back to the W3ID SSO flow that cannot complete in a pod and parks at the auth prompt — with every precondition green (settings `api-key`, key readable by the agent UID, token delivered). A bob that cannot authenticate also cannot complete the startup writes behind "There was an error saving your latest settings changes."

The codebase already documented this tmux behavior (comment at `manager.go` ~line 377, and the `RelaunchBobAgentsAwaitingKey` comment), but the launch path still depended on the broken ordering.

## Verification (run against real tmux, not reasoned about)

```
new-session -> set-environment -> send-keys        => pane env EMPTY   (reproduces the bug)
new-session -> set-environment -> respawn-pane     => pane env FULL    (the fix)
```

Two obvious alternatives were tested and **rejected**:

- **Env on the `tmux new-session` client process** — works on a cold server, fails once the server is already running, because the *server* forks the pane, not the client. That is precisely the relaunch case (`killSessionForRelaunch` kills the session, not the server), so it would have regressed silently.
- **`set-environment -g` before `new-session`** — cannot run at all on a cold server: `error connecting to <socket>`.

Also verified: pluk's `pipe-pane` still captures correctly, since it attaches after the respawn; and the container's shells are `/bin/bash` (`Dockerfile:52`, `entrypoint.sh:501`), matching the tested case.

## Changes

- **`ensureTmuxSession`** — respawn the pane after the environment is populated so the shell inherits it. Failure is logged and non-fatal (a degraded session beats refusing to launch). Secrets stay off the command line: `respawn-pane` takes no arguments.
- **`verifyBobStateDirsWritable`** — new advisory probe. bob surfaces "error saving your latest settings changes" and "Failed to initialize logger:" only inside its own TUI, so a genuinely unwritable dir is invisible to anyone reading `kubectl logs` — that silence is what made this issue hard to diagnose. Probes both `$HOME/.bob` and `<workdir>/.bob` for the agent UID's group access and logs a remedy. Mirrors `verifyBobKeyReadable`, including its false-positive trap: hive runs as `dev` and *owns* these dirs, so any check on its own behalf passes while the agent UID still gets EACCES. Never blocks a launch. Missing dirs are not faults — bob creates them recursively on first use.
- Named constants only (`bobStateGroupWriteBit`, `config.BobStateDirName`); no magic numbers. No credential values logged — paths, modes and UIDs only.

## Safety

`ensureTmuxSession` takes no locks in the added code — no new `m.mu` acquisition, so the non-reentrant-RWMutex deadlock class from #1980→#1988 is not reintroduced.

## Testing

9 new table-driven cases covering the production-correct `0770`, the `0700` lockout, `0750`/`0760` partial-permission cases, root/unset-UID bypass, missing dirs, and workspace-dir-only failure. `go build ./...`, `go vet`, and the `pkg/agent` + `pkg/config` suites pass. gofmt clean on touched files — the pre-existing drift in `manager.go` and `config.go` was deliberately left alone (verified my hunks add none).

## What remains

Marked `Refs` rather than `Fixes` because I could not re-attach to the reporter's Supervisor pane to confirm their *specific* instance was the same failure. The probe added here will now say so definitively in the hive log if any state dir really is unwritable in their environment. If #2253 was solely the trust banner, that was already resolved by #2249.

## Porting

**mk / dd / v3 still need this.** ks/console's hive runs v3, so a v2-only merge will not reach it.

### DIFF
--- a/v2/pkg/agent/bob_settings.go
+++ b/v2/pkg/agent/bob_settings.go
@@ -21,6 +21,9 @@ const (
 	// bobKeyOtherReadBit is `o+r`. Also sufficient for the agent to read, but
 	// it means the secret is world-readable, which is worth flagging.
 	bobKeyOtherReadBit fs.FileMode = 0o004
+	// bobStateGroupWriteBit is `g+w` — the bit that lets an agent UID create
+	// or replace files inside a dev:node-owned bob state directory.
+	bobStateGroupWriteBit fs.FileMode = 0o020
 )
 
 // verifyBobKeyReadable checks that the resolved bob key file is readable BY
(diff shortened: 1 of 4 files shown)


clubanderson:
Test note, for anyone reproducing locally: `pkg/agent` tests that create tmux sessions (`TestStart_TransientInferencePauseAutoUnpauses`, `TestSendKick_NoTmuxSession`) use fixed session names on the **shared** tmux socket. A session left behind by an earlier interrupted run makes `new-session -s hive-cxa` fail with `exit status 1`, surfacing as `creating tmux session for cxa: exit status 1` — which looks alarming in `ensureTmuxSession`, the function this PR touches, but is unrelated to it.

Verified it is environmental, not a regression:
- `tmux kill-server` first, then the same tests pass.
- The two tests pass in isolation on both this branch and unmodified `origin/v2`.
- CI `build-and-test` is green (clean runner, no leftover sockets).

Separately, `TestConcurrentPauseResume_NoPanic` calls `m.Resume(nil, ...)` with a literal nil context, so it can panic with `cannot create context from nil parent` whenever the race lets `launchInTmux` reach `context.WithCancel`. That is pre-existing on `origin/v2` and untouched here — flagging it as a latent test bug rather than fixing it in a bugfix PR.

PR #2337, opened 2026-07-31
🐛 fix(upgrade): roll floating-tag spokes onto the requested SHA

clubanderson:
## The bug

A spoke whose Deployment tracks a **mutable tag** (`v2-latest` and friends) can never be upgraded to a specific SHA, and the hub retries forever without surfacing a fault. Two hives sat "Upgrading" for 86 and 8 minutes today with fully-settled Deployments — `observedGeneration == generation`, `1/1` ready, zero restarts.

## Root cause (file:line)

`v2/pkg/hub/self_upgrade.go:161` — `UpgradeSelfToSHA` returned `needsRestart = true` for **any** mutable tag, taking a bare `rollout restart`.

A restart re-pulls the tag, so the pod lands on whatever digest CI last published under it — the hub's target only by coincidence. **Verified against the live registry**, three distinct digests:

| tag | digest | SHA |
|---|---|---|
| `v2-latest` | `sha256:709fea5e7d81` | `fb37afe` |
| target | `sha256:1ddcf6bed4a1` | `07ccca1` |
| target | `sha256:c9fe0024086b` | `fc32ae4` |

So the spoke came back reporting a SHA that was neither its old one nor the target. The hub's completion check never matched, and the hive was re-instructed on every heartbeat, forever.

Note this is **not** a stale-cached-digest problem — the restart really does roll and really does pull new code. It just pulls the *wrong* code. `hive-hosted-hosted-available-vllmd-02/03` are live proof: spec says `v2-latest`, running digest is `sha256:1ddcf6…` (`07ccca1`), i.e. already drifted away from the tag they nominally track.

## The fix

Keep the floating tag — a deliberate platform preference that came out of the self-upgrade CrashLoop incident — and make the rollout deterministic anyway.

`upgradeSelfMutableToSHA` writes the target SHA into a **pod-template annotation** (`hive.kubestellar.io/upgrade-target-sha`, alongside the existing `restart-at` so a retry of an identical target is still a distinct template). The template differs even though the image string does not, which is what actually forces Kubernetes to roll, and the Deployment now carries a readable record of what was asked for. **The image is never rewritten**, so the floating tag survives — a test asserts this.

A rejected patch is now returned as an **error** instead of degrading to the silent no-op; the caller already turns that into a recorded, hub-visible failure (#2308) rather than a permanent spinner.

This narrows but does not eliminate the race — if CI republishes between target selection and the pull, the pod lands on a newer build. That is benign and self-correcting: the hub accepts "at target **or** at registry-latest" as success, so an overshoot completes rather than wedges. What it must never do again is silently land on the wrong build with no record of the request.

## What did NOT need changing

- **`imagePullPolicy`** — already defaults to `Always` at provision time (`saas_provision.go:855`) and is `Always` on **all 51** hive Deployments across `vllm-d` (33) and `hive-oke` (18). No exceptions, so no change.
- **Completion detection** — `server.go:1062` already compares the spoke's self-reported `gitHash` against the target, never the image string, so it works unmodified when the image string cannot change. The `|| registryLatestSHA` fallback is what makes the overshoot case above benign.

## Detection gap / interaction with #2327

#2327's sweep clears and re-arms an orphaned upgrade. That is correct for a departed pod, but it is an **infinite loop** for a spoke structurally unable to advance — exactly the floating-tag case: restart, land on the wrong build, report in, get swept, get re-armed, repeat. Nothing ever reaches a human.

This bounds it. After `maxOrphanedUpgradeSweeps` (3, ≈1 hour at the default `staleUpgradeTimeout`, since each sweep costs at least `orphanedUpgradeClearAfter`) the hub stops retrying, drops the armed instruction, and records a terminal `UpgradeFailed` with the cause — surfaced by the existing UI and `AlertTypeStuckUpgrade`, so no new alert type. Below the budget #2327's behaviour is **unchanged**, guarded by a test. The counter resets when an upgrade lands, so the budget is per-upgrade, not per-hive-lifetime.

## Testing

- `go build ./...` clean, `go vet ./pkg/hub/` clean, `go test ./pkg/hub/` fully green.
- New: forced rollout on a mutable tag (asserting the image is *not* rewritten and the annotation lands on the **template**), reported patch failure, unchanged pinned path (#2308 regression guard), escalation at the budget, preserved re-arm below it.
- `saas.go` / `dashboardHTML` untouched.

## Blast radius

Deployments on floating tags, re-verified read-only:

| cluster | floating | pinned | total |
|---|---|---|---|
| `vllm-d` | 30 (29 `v2-latest`, 1 `mk-latest`) | 3 | 33 |
| `hive-oke` | 2 (1 `v2-latest`, 1 `dd-latest`) | 16 | 18 |

Every floating-tag hive would exhibit this the next time its hub targets a SHA.

**No live systems were modified.** The two hives repaired by hand earlier are already on the target. No mass-patch was performed; with this fix the next upgrade cycle rolls them correctly on its own, so a one-off repair looks unnecessary — happy to do one if you'd rather not wait.

### DIFF
--- a/v2/pkg/hub/self_upgrade.go
+++ b/v2/pkg/hub/self_upgrade.go
@@ -148,6 +148,13 @@ func imageTagIsMutable(image string) bool {
 //
 // So: patch the image when pinned, restart when mutable. Returns needsRestart
 // false when the image patch itself rolls the deployment.
+//
+// The mutable case is NOT simply "restart and hope". A restart re-pulls the
+// tag and therefore lands on whatever digest CI last published under it —
+// which is only the requested target by coincidence. Verified on the live
+// fleet: v2-latest resolved to fb37afe while the hub was targeting 07ccca1, so
+// every restart delivered the wrong build, the spoke reported fb37afe, and the
+// hub re-instructed the same upgrade forever. See upgradeSelfMutableToSHA.
 func UpgradeSelfToSHA(logger *slog.Logger, targetSHA string) (needsRestart bool, err error) {
 	current, err := selfDeploymentImage()
 	if err != nil {
@@ -159,7 +166,7 @@ func UpgradeSelfToSHA(logger *slog.Logger, targetSHA string) (needsRestart bool,
 		return true, nil
 	}
 	if imageTagIsMutable(current) {
-		return true, nil
+		return upgradeSelfMutableToSHA(logger, current, targetSHA)
 	}
 	// Pinned. Rewrite the tag in place so the registry/repo (which may be a
 	// mirror or a private registry) is preserved — only the tag changes.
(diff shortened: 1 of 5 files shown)


clubanderson:
### Additional evidence: identical spec, three different running digests

Re-verified read-only on `vllm-d`. Three spokes whose Deployment spec is byte-identical (`ghcr.io/kubestellar/hive:v2-latest`) are each running a **different** image:

| namespace | spec | running digest | resolves to |
|---|---|---|---|
| `…vllmd-04` | `hive:v2-latest` | `sha256:612f79f1e0ef` | *(untagged)* |
| `…vllmd-05` | `hive:v2-latest` | `sha256:1c5e95982401` | `c11643a` |
| `…vllmd-06` | `hive:v2-latest` | `sha256:221b4c853f78` | *(untagged)* |
| `…vllmd-02/03` | `hive:v2-latest` | `sha256:1ddcf6bed4a1` | `07ccca1` |

And `v2-latest` currently resolves to `sha256:709fea5e7d81` (`fb37afe`) — which **none** of them is running.

Two things follow:

1. The image string carries no information about what a pod actually runs, which is why completion detection must key off the spoke's self-reported `gitHash` (it already does) and why the spec-based no-op was invisible.
2. Each of these pods pulled `v2-latest` at a different moment and froze there. A bare restart would move each to a fourth, unrelated build rather than to any requested target — so the pre-fix upgrade path was not merely a no-op, it was actively non-deterministic.

PR #2148, opened 2026-07-28
fix(integrated): stabilize governed Visual Hive lifecycle

DavidDiaz0317:
## Summary

This PR adds Hive-owned Visual Hive lifecycle support while preserving ordinary Hive as the only scheduler, controller, state owner, and GitHub writer.

It consolidates the production-shaped defects found during complete setup → baseline → evidence → governed repair → replay → resolution → uninstall rehearsals:

- recover exact unmerged setup proposals and restore managed-file preimages safely;
- hot-activate a newly installed Visual Hive contract in the existing Hive process;
- keep repositories without an installed contract dormant;
- preserve isolated Codex authentication and role-scoped bead access;
- bind supported backend/model pairs during ACMM transitions;
- prioritize higher-severity verified work before bounded specialist capacity is consumed;
- ingest authoritative Storybook discovery evidence without inventing repository coverage;
- make repair verification, issue intent, advisory publication, and restart replay idempotent;
- recover journal-bound repair cleanup when disposable worktrees or loose refs disappear;
- retire exact obsolete repair proposals without treating retirement as issue resolution; and
- restore the exact pre-install managed-file and baseline inventory during supported uninstall.

Visual Hive remains evidence-only. This PR adds no second dashboard, scheduler, arbitrary-command endpoint, auto-merge path, or independent GitHub publisher.

## Safety model

- Dashboard lifecycle operations remain owner-authenticated and schema-validated.
- Plan/apply operations retain exact request, plan, transaction, repository, repository-ID, state, actor, head, workflow, artifact, and receipt bindings.
- One runtime manager reconciles one controller, ownership lease, cadence owner, and runner.
- Same-contract reconciliation is an idempotent no-op; conflicting repository/state bindings fail closed and require managed uninstall/rebind.
- Pending setup or baseline state may initialize the controller but cannot dispatch production work.
- Partial activation failure releases acquired runtime ownership while ordinary Hive continues.
- Repair retirement requires the exact Hive branch/PR/check receipt and never substitutes for fresh authoritative absence evidence.
- Stale, partial, tampered, unindexed, foreign-repository, or wrong-head evidence remains rejected.

## Current review binding

- PR head: `69898f83e20429e11c60dd91fee75402ee651807`
- Tree: `4f7d8948673b44a4511027fdae4a0ae17f5ee31a`
- Qualified functional parent: `adb1945de1d8cdf51c4c1970dbfe21134c9b2315`
[...]

PR #2360, opened 2026-07-31
✨ feat(hub): switch a hive between forges (github.com / GitHub Enterprise), delivered over the heartbeat

clubanderson:
> *"i need a way to switch people between github.ibm.com or github.com or any forge in the hub for a specific hive's listing. Enrico has github.com and he needs to be on github.ibm.com."*

There was **no API and no UI to change a hive's forge after provisioning**. `github_host` is written once, at provision/assign time (`saas.go:5097`), and no `HandleFunc` mentions it.

## Why a hub-side edit cannot work

`github_host` was edited to `github.ibm.com` in a live hive's `meta.json` and the hub restarted. **It did not stick** — the registry was back to `github.com` within one beat.

`handleHeartbeat` records `RegistryEntry.GitHubHost` from the **spoke's** `HeartbeatPayload.GitHubHost` on every beat, and the spoke derives that from its own config via `GitHubConfig.HostLabel()`. Nothing the hub writes to `meta.json` changes what the spoke *runs*, so the spoke simply re-reports the unchanged truth and the hub adopts it.

This is the adopt-stale-value class fixed for ACMM in #2354, and the fix has the same shape:

- **`RequestedGitHubHost`** records what the operator asked for, so it survives the spoke reporting the old host.
- **`ForgeDelivered`** latches true the moment the spoke reports the requested host back — after which the push stops **for good** and the spoke owns the value again. This is what avoids the #2061 failure mode where a forever-push made the value un-editable on the spoke.
- Both default to their zero values on every existing hive, and the push is gated on a **non-empty** `RequestedGitHubHost`, so the feature is inert for every hive nobody has switched. No migration.

## What actually travels

`HostLabel()` reads `base_url` and **falls back to `api_url`**, so pushing the API URL alone is sufficient to change what the spoke reports. `HeartbeatProjectConfig.GitHubAPIURL` is an existing, already-implemented channel on both ends — the spoke adopts it in `main.go` and persists to the PVC overlay. That is the channel this uses, which is why it works uniformly for the reachable (hive-oke) and firewalled/heartbeat-only (vllm-d) clusters alike.

**Switching back to public github.com needed explicit handling.** Everywhere else an empty `github_api_url` means "leave the spoke alone", and `gheAPIURLForHost("github.com")` is `""` by definition — so the naive implementation can move a hive *to* enterprise but never *back*. The public target is delivered as an explicit `https://api.github.com` instead of as emptiness. It also cannot ride the existing `needGHEAPIPush` gate, which is deliberately conservative about an empty `curAPIURL`; the switch carries its own target and its own read-back.

## App identity

Switching forge swaps the App, resolved **from `clusters.json` only** — `hive-oke` names its github.com App, `vllm-d` names its github.ibm.com App. **No App ID is hardcoded in Go.**

A cluster whose own forge does not match the target contributes **nothing** (app_id 0 = "say nothing"), which leaves the spoke's credentials untouched rather than pushing an App registered on the wrong forge.

**`installation_id` is never carried over.** An ID issued by one forge names nothing on another; pushing it would authenticate as a nonexistent installation and produce a 401/404 that reads as a key problem rather than a forge problem. It is either supplied in the request or left for installation auto-discovery — and when it is absent the response says so explicitly (`app_install_note`), so an operator cannot mistake it for a silent success.

Key material is deliberately **not** sent here: the per-cluster key reconcile (`cluster_app_key.go`) owns that and gates every key push on `HasKey()`, so duplicating it risked blanking a working key with an empty string.

## Authorization

`requireAuth` on the route plus an inner owner-or-admin check returning `only the owner can switch forges` 403 — the same shape as `handleSwitchBranch` and `handleToggleAutoUpgrade`.

## Forge validation, not arbitrary strings

`parseForgeTarget` resolves a validated `ForgeKind` + host, accepting a bare host or a pasted org URL (the likeliest operator input) and normalizing both to one canonical target. Ports, userinfo and paths are stripped; a malformed host is rejected naming what is supported. It deliberately does **not** use `githubHostLabel`, which maps `""` to `github.com` — that would turn a blank field into a silent switch to public GitHub, a destructive default for this endpoint.

GitLab/Gitea will be selected by an explicit kind rather than inferred from a host, since their API paths differ and a host alone cannot distinguish them.

## Tests

Six tests, including the live regression: **beat 1** pushes the enterprise API URL while the spoke reports the old host and refuses to latch on that stale report; **beat 2** completes delivery when the spoke confirms; then delivery is asserted **idempotent** — the push stops permanently and re-adopting is a no-op. Separately covered: the switch-back-to-public direction, "unknown is not a match" (a spoke too old to report the host must never complete a delivery — the silent-failure case), inertness for unswitched hives, and that the App ID comes from cluster config in both directions.

`go build ./...` clean; full `pkg/hub` suite passes with no regressions.

The embedded `dashboardHTML` block is **byte-identical to base** (md5 verified) — no JS was touched, and `<body>` count is unchanged at 3.

## Live systems

**Strictly read-only.** No hive, user record, cluster config or hub state was mutated. Enrico's hive has deliberately **not** been switched — operator steps are in the thread.

### DIFF
--- a/v2/pkg/hub/forge.go
+++ b/v2/pkg/hub/forge.go
@@ -180,13 +180,17 @@ type SwitchForgeRequest struct {
 // identity the hive will now authenticate as, so the operator can see whether
 // the target cluster actually has an App configured before waiting on a beat.
 type switchForgeResponse struct {
-	Status          string `json:"status"`
-	ID              string `json:"id"`
-	FromHost        string `json:"from_host"`
-	ToHost          string `json:"to_host"`
-	ForgeKind       string `json:"forge_kind"`
-	APIURL          string `json:"api_url,omitempty"`
-	AppID           string `json:"app_id,omitempty"`
+	Status    string `json:"status"`
+	ID        string `json:"id"`
+	FromHost  string `json:"from_host"`
+	ToHost    string `json:"to_host"`
+	ForgeKind string `json:"forge_kind"`
+	APIURL    string `json:"api_url,omitempty"`
+	AppID     string `json:"app_id,omitempty"`
+	// AppSlug is the App slug now in effect on the target forge. Reported so an
+	// operator can see the install link will resolve, rather than discovering a
+	// 404 through the user.
+	AppSlug         string `json:"app_slug,omitempty"`
 	InstallationID  string `json:"installation_id,omitempty"`
 	AppInstallNote  string `json:"app_install_note,omitempty"`
 	DeliveryPending bool   `json:"delivery_pending"`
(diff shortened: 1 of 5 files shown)


clubanderson:
## Update — a forge switch is now an atomic swap of the whole identity

The first revision moved `github_host` + `api_url` + `app_id` and left `app_slug` to whatever the cluster happened to have. That is **not sufficient**, and there is now a live example of why.

A hive was moved to `github.ibm.com` by editing `github_host` by hand. The host changed; the identity did not:

```
githubHost:      github.ibm.com               <- changed
app_id:          3568013                      <- STILL the github.com App (GHE needs its own)
app_slug:        ""                           <- falls back to "kubestellar-hive"
installation_id: 0                            <- never resolved for the new forge
api_url:         https://github.ibm.com/api/v3  <- correct
base_url:        ""                           <- should be https://github.ibm.com
```

The user clicked **"Install GitHub App"** and got a **GitHub 404** at `github.ibm.com/github-apps/kubestellar-hive/installations/new`. An empty `app_slug` falls back through `config.ResolvedAppSlug` to `DefaultGitHubAppSlug` — the github.com slug, which names no App on GHE. **A wrong install link is worse than no link: it tells the user the product is broken.**

### What changed

- **`forgeIdentityForTarget` resolves the identity as a SET** — `app_id` **and** `app_slug`, from `clusters.json` only — and returns the names of everything missing. Nothing is hardcoded in Go.
- **The switch refuses rather than half-applies.** A missing `github_app_slug` (or `github_app_id`, or a cluster serving a different forge) returns **409** naming the missing fields and the remedy, and **writes nothing** — not the host, not the URL pins, not the delivery arming, no queued App config. A partial switch is precisely the hand-made failure above.
- `app_slug` is deliberately required **even for public github.com**. The default happens to be correct there, but accepting an unset value would make the atomicity guarantee hold on one forge and not the other, with no signal that the cluster entry is incomplete.
- **`base_url` is now pinned** alongside `api_url` — the other field the manual edit left empty.
- The resolved slug **travels with the app_id** to the spoke, so a switched hive can never be left falling back to the github.com slug on a GHE host.
- The response and audit log report the `app_slug` now in effect, so an operator sees the install link will resolve instead of learning it 404s from the user.

`installation_id` is still never carried over, and key material is still left to the per-cluster reconcile.

### Tests added

- **`TestSwitchForgeRefusesPartialIdentity`** — drives the handler with the live vllm-d shape (app id, GHE URLs, **no** `github_app_slug`) and asserts 409 plus **no partial write of any field**.
- **`TestSwitchForgeSwapsTheWholeIdentity`** — asserts host, `base_url`, `api_url`, queued `app_id`, queued `app_slug`, `installation_id == 0`, no key material, and that a re-run is idempotent.
- **`TestSwitchForgeAuthorization`** — non-owner 403, owner 200, admin 200, invalid host 400.
- **`TestForgeIdentityIsAtomic`** — a refused identity is entirely empty (all-or-nothing) and the refusal names the missing field.

Full `pkg/hub` suite passes. `dashboardHTML` remains **byte-identical to base** (md5 verified).

### Coordination

The separate agent's PR for repairing the live hive and populating the vllm-d cluster slug was **not open at the time of writing** (open PRs checked: #2358, #2359, #2355, #2305, #2293, #2148). This PR deliberately adds **no** cluster-config data and **does not touch** any live hive — it only refuses to switch when the config is incomplete, so it composes cleanly with that repair rather than duplicating it. Once `github_app_slug` is populated for `vllm-d`, this endpoint will accept the switch with no change here.

Note #2293 (`feat/forge-consistency-and-selfheal`) is adjacent — worth sequencing if it lands first.

PR #2431, opened 2026-08-01
✨ feat(hub): replace native browser dialogs with the hub's own modals

clubanderson:
The hub dashboard used `window.confirm()`, `window.prompt()` and `window.alert()` in **eight** places. They can't be styled, block the whole tab, and read as a browser warning rather than part of the product — jarring next to the themed overlay modals the same page already has (assign, access, timeline, OpenRouter funding) and its own `hiveToast` notifier.

## Three helpers, following the existing pattern

```js
hiveConfirm(title, body, {ok, cancel, danger}) -> Promise<bool>
hiveNotify(title, body)                        -> Promise<void>
hivePrompt(title, defaultValue, {ok})          -> Promise<string|null>
```

Promise-based so they drop into existing `await`-style call sites. `hivePrompt` resolves `null` on cancel — the same contract as the native call — so existing null checks keep working. Escape and backdrop click dismiss; the primary button takes focus, and Enter submits in the prompt.

## Converted

| site | was |
|---|---|
| `saveCurrentView` | prompt + confirm (overwrite) |
| `renameActiveView` | prompt |
| `deleteActiveView` | confirm — now styled destructive |
| `resetHiveApp` | confirm + three alerts |

The two enclosing saved-view handlers became `async`.

## Guard

`TestNoNativeBrowserModals` scans the rendered dashboard source and fails on any new native call. A **source scan**, not a behavioural test: the calls sit in unrelated handlers, and what matters is that no *new* one appears — a behavioural test would enumerate today's handlers and pass while a new one did it. It also asserts it read ≥500 lines, because a scan that reads nothing passes for the wrong reason.

| Mutation | Result |
|---|---|
| Reintroduce a native `confirm` | 1 failure ✓ |
| Scan an empty source | 1 failure ✓ |

**Note for anyone editing `dashboardHTML`:** it's a Go raw string, so a single backtick anywhere inside — even in a comment — ends the literal. Two of mine did exactly that while writing this.

`pkg/hub` green.

### DIFF
--- a/v2/pkg/hub/no_native_modals_test.go
+++ b/v2/pkg/hub/no_native_modals_test.go
@@ -1,0 +1,61 @@
+package hub
+
+import (
+	"regexp"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+// TestNoNativeBrowserModals keeps window.confirm/prompt/alert out of the hub UI.
+//
+// Native dialogs cannot be styled, block the whole tab, and read as a browser
+// warning rather than as part of the product — jarring in a themed dashboard
+// that already has its own overlay modals (assign, access, timeline) and a
+// hiveToast notifier.
+//
+// Use hiveConfirm / hiveNotify / hivePrompt instead. They follow the same
+// overlay pattern, are promise-based so they drop into await-style call sites,
+// and resolve false/null on Escape or a backdrop click.
+//
+// The check scans the rendered dashboard source rather than asserting on
+// behaviour: the calls are spread across unrelated handlers, and what matters
+// is that no NEW one appears anywhere — a behavioural test would enumerate the
+// handlers that exist today and pass while a new one did it.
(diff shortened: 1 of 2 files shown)


clubanderson:
**quality review** — `deleteSavedView` async/await mismatch

The diff converts `deleteSavedView` to use `await hiveConfirm(...)` but the function declaration remains `function deleteSavedView()` (not `async`). This is a SyntaxError in strict mode or silently broken cancel behavior in sloppy mode.

`saveCurrentView` was correctly converted to `async function saveCurrentView()` — `deleteSavedView` needs the same treatment.

The guard test (`TestNoNativeBrowserModals`) correctly catches native dialog reintroduction but doesn't catch this class of error. Filed #2533 for a test enhancement.

---
*quality agent (hold-gated mode)*

PR #2651, opened 2026-08-05
🐛 fix flaky TestAlertAcksPersistRoundTrip (shared alertAcksPath global race)

clubanderson:
## Race mechanism

`alertAcksPath` (`v2/pkg/hub/alerts.go`) is a package-level `var`. Multiple
tests in `pkg/hub` — `TestAlertAcksPersistRoundTrip`,
`TestAlertAcksLoadPrunesExpired`, `TestAlertAcksLoadMissingFileIsNotAnError`,
`TestAlertAcksLoadCorruptFileIsNotFatal`, `TestHandleAlertAck`,
`TestHandleAlertAckPersists`, `TestAckFailedUpgradeSucceedsAndSurvivesRestart`,
`TestSaveAlertAcksSurvivesUnwritablePath`, `TestSaveAlertAcksIsAtomic` — each
do:

```go
orig := alertAcksPath
alertAcksPath = filepath.Join(dir, "acks.json")
defer func() { alertAcksPath = orig }()
```

`saveAlertAcks` / `loadAlertAcks` (both methods on `*HubServer`) read that
same **mutable global** directly at call time instead of a value captured on
the server. `TestAlertAcksPersistRoundTrip` builds two servers
(`newTestAlertServer()`) back to back — one to save an ack, a second to
simulate a restart and reload it — and both go through the shared global. Any
call path that reads or writes `alertAcksPath` while it doesn't hold the
value a given server was set up against reproduces exactly the observed
failure:

```
alerts_test.go:1141: an acknowledgement must survive a hub restart
```

(The `connection refused ... localhost:8080` lines seen alongside it are
unrelated background noise from other tests' URL-reachability/namespace
calls.)

This is the same class of bug already fixed once in this package for the hub
registry — see `HubServer.registryPath` / `regPath()` in `server.go`, which
carries a comment describing the identical "TestLoadRegistry -race failure."
`alertAcksPath` never got the same treatment when it was added.

## Fix

Applied the same pattern used for `registryPath`:

- Added a per-instance `HubServer.alertAcksPath` field.
- `NewHubServer` copies the package global into that field once, at
  construction.
- Added `ackPath()` (mirrors `regPath()`): returns the instance field if set,
  else falls back to the package global — so bare `&HubServer{}` test
  literals (`newTestAlertServer` and friends) keep working unmodified.
- `saveAlertAcks` / `loadAlertAcks` now call `s.ackPath()` instead of reading
  the `alertAcksPath` global directly.

No test files were changed — every existing test that redirects
`alertAcksPath` before constructing a bare `&HubServer{}` literal still works
exactly as before, because `ackPath()` falls back to the global for those
servers.

## Why this is now deterministic

`saveAlertAcks`/`loadAlertAcks` never read the mutable global at call time
anymore — a `*HubServer` only ever sees the path it was constructed with.
Within `TestAlertAcksPersistRoundTrip`, both servers it builds resolve to the
*same* path (the global's value at the moment each was constructed, both
within the same test body), and no background goroutine anywhere in the
package calls either method — `fleetAlerts`/`saveAlertAcks`/`loadAlertAcks`
are only invoked synchronously from HTTP handlers, `NewHubServer`'s own
startup path, or tests directly. So there is no path left by which a test
reassigning the global mid-suite can be observed by a server whose
save/reload round-trip is already in progress.

## Impact

This flake was intermittently failing the required `coverage` CI check and
blocking merges on multiple unrelated PRs today.

### DIFF
--- a/v2/pkg/hub/alerts.go
+++ b/v2/pkg/hub/alerts.go
@@ -34,9 +34,17 @@ const alertProvStatusError = "error"
 // one more SOURCE feeding this model — no rule here needs to be rewritten, and
 // nothing here recomputes a drift signal itself.
 
-// alertAcksPath is the on-disk file where admin alert acknowledgements are
-// persisted so a silenced alert stays silenced across a hub restart or upgrade.
-// A var (not a const) so tests can point it at a temp dir.
+// alertAcksPath is the DEFAULT on-disk file where admin alert acknowledgements
+// are persisted so a silenced alert stays silenced across a hub restart or
+// upgrade. A var (not a const) so tests can redirect it at a temp dir before
+// constructing a server; production keeps the default. NewHubServer copies it
+// into the per-instance HubServer.alertAcksPath at construction, and
+// saveAlertAcks/loadAlertAcks read that field via ackPath() — never this
+// global directly — so one test's temporary redirect of this var can never be
+// observed by a *HubServer another test already built (the
+// TestAlertAcksPersistRoundTrip flake: two servers built back-to-back in the
+// same test both resolve to the SAME path this way, while any concurrently
+// running test's redirect is invisible to them).
 var alertAcksPath = "/data/saas/hub-alert-acks.json"
 
 // Alert severities, ordered most- to least-urgent. Kept as named constants so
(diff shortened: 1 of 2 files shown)


clubanderson:
Closing as duplicate — the same TestAlertAcksPersistRoundTrip fix landed in #2652 (merged). Both diagnosed the identical root cause (frozen creation clock vs real-wall-clock prune in loadAlertAcks crossing the 7-day boundary on 2026-08-05).

Issue #2610, opened 2026-08-04
contribute: contributor.env written world-readable, unauthenticated force token reissue, and token_refresh never firing again after a reconnect


castrojo:
Part of #2534 (downstream field notes from `projectbluefin/review`).

Hi again 👋 — three more observations from operating a downstream contributor against
your protocol. As before: evidence and options, no patch and no ask. Every decision here
is yours, and we say plainly where a thing is ours rather than yours.

**What we read and when.** All code below was read at `v2` HEAD
`fc3d7179255d13a613632fd1e982691d2d8bc0ae` on **2026-08-04 22:25 UTC** (`git fetch origin v2`
against `https://github.com/kubestellar/hive`), and every link is pinned to that SHA.
`projectbluefin/review` pins Hive at `4d61ad7ce8b646a4e380865c521d5b12677240c9`; where the
two differ we checked HEAD, not our pin, so nothing here is a stale-pin artifact.
We filed this unlabeled and unassigned — triage is yours.

---

## 1. `contributor.env` is written at umask, unlike the two secrets beside it

**Observation.** `contribute-setup` writes three files into `~/.config/hive`. Two get an
explicit `chmod 600`; the one holding the registration token does not.

- `gh-auth.env` → [`Justfile#L250-L251`](https://github.com/kubestellar/hive/blob/fc3d7179255d13a613632fd1e982691d2d8bc0ae/Justfile#L250-L251)
- `claude-config.json` → [`Justfile#L314-L315`](https://github.com/kubestellar/hive/blob/fc3d7179255d13a613632fd1e982691d2d8bc0ae/Justfile#L314-L315)
- `contributor.env`, containing `HIVE_REGISTRATION_TOKEN` → [`Justfile#L288-L294`](https://github.com/kubestellar/hive/blob/fc3d7179255d13a613632fd1e982691d2d8bc0ae/Justfile#L288-L294) (no `chmod`)

**Reproduction (live, our host, 2026-08-04 22:29 UTC).** `umask` is `0022`; the file was
created by a normal `just contribute-setup` run on 2026-07-28:

```
$ umask
0022
$ stat -c '%n %a' ~/.config/hive/*
/home/.../.config/hive/contributor.env 644
/home/.../.config/hive/gh-auth.env     600
```

**What that token is.** `HIVE_REGISTRATION_TOKEN` is the sole bearer credential for the
contributor WebSocket: `auth_response` looks the connection up by SHA-256 of the token and
[...]

PR #2672, opened 2026-08-05
✨ feat(contribute): hub task lease + operator revoke/requeue with stale-worker guard

### LINKED ISSUE #2568: contributor: task ownership has no hub lease or safe operator revoke/requeue path
> Part of #2534 (contributor onboarding field notes from `projectbluefin/review`).

## Summary

The last 24 hours closed several task-release races, but task ownership still has no hub-enforced lease and operators have no safe task revoke/requeue control. If the relay remains connected but wedges without sending completion, failure, or another `ready`, the issue can remain held indefinitely.

This is not a request for a second queue: Hive remains the assignment authority. It is a request to make Hive’s ownership state bounded and recoverable.

## What shipped

- #2455 added failure cooldown/quarantine and failure-aware selection.
- #2492 releases and cools down a task on WebSocket disconnect.
- #2557 releases and cools down a held task when the client sends another `ready`.
- #2562 added admission controls and contributor trust/revoke/remove operations, but explicitly left task requeue/reprioritize out because no safe endpoint exists.

Those cover observed state transitions. They do not cover a live socket whose worker has stopped making progress.

## Evidence

At `v2` commit [`08518560`](https://github.com/kubestellar/hive/commit/08518560e14cc583215765831cf72487d432d1c1), `wsTaskTimeout = 30 * time.Minute` is declared at [`contribute_ws.go:31`](https://github.com/kubestellar/hive/blob/08518560e14cc583215765831cf72487d432d1c1/v2/pkg/dashboard/contribute_ws.go#L31) and has no enforcement reference elsewhere in the file.

The relay has its own 30-minute watchdog, but that only fires while the relay process and event loop are healthy. A hub cannot distinguish “working slowly” from “connected but wedged,” and an operator viewing the task in Management & Operations cannot safely release it.

## Options

1. A hub-owned renewable lease based on accepted assignment plus progress/heartbeat, with an explicit expiry transition.
2. A safe operator revoke/requeue/quarantine action that carries a reason and is visible to the client.
3. An assignment generation/lease token so a stale worker cannot later complete or push ownership state after reassignment.
4. Keep automatic expiry conservative and offer manual recovery first; the important property is that ownership is bounded and auditable.

Any design should preserve the existing cooldown protections so recovery does not recreate the duplicate-assignment race fixed by #2492.

**Gate:** Concurrency + recovery. At most one current lease may be authoritative, and releasing it must not allow a stale worker to overwrite the new owner’s state.

clubanderson:
## Summary

Closes the remaining half of #2568: contributor task ownership now has a **hub-enforced lease** with a bounded expiry, a **safe operator revoke/requeue** action that carries a reason, and an **assignment-generation guard** so a stale worker cannot overwrite the new owner's state after its task is reassigned.

The issue's **Gate** — *"at most one current lease may be authoritative, and releasing it must not allow a stale worker to overwrite the new owner's state"* — is delivered by three additions, built on the already-shipped manual-requeue slice and the #2492/#2557/#2649 cooldown machinery (nothing rebuilt).

## The lease model

- `wsTaskTimeout` (30m) was declared but never enforced (the evidence cited in the issue). It now backs a real **hub-owned lease**: `lastLeaseRenew` is set on assignment and renewed on every `task_progress`.
- `cleanupLoop` calls `reclaimExpiredLeases`, which releases a task whose lease has gone stale through the **same** `recordTaskFailure` cooldown + generation-bump path a manual requeue uses.
- **Conservative by design** (option 4 in the issue): a task that keeps reporting progress renews its lease and is **never** reclaimed, so "working slowly but alive" is not confused with "connected but wedged". Manual operator recovery is the primary path; auto-expiry is only the backstop.

## Operator revoke / requeue

- `RequeueContributorTask` now takes a **reason**. The endpoint (`POST /api/contributors/{id}/requeue`) reads it from `?reason=` or a JSON body `{"reason": ...}`.
- The reason is recorded in the audit + activity log and **pushed to a still-connected worker** on `task_revoke`, so the client learns *why* its task was released (as the issue requires).
- The UI prompts the operator for a reason.

## Stale-worker guard (the Gate)

- New `WSMessage.TaskGen` (a monotonic **assignment generation / lease token**) is stamped on each `task_assign` and the `task_progress` RESUME adopt, echoed back by the relay, and checked on `task_complete` / `task_progress` / `task_failed` via `generationAccepted()`.
- A message carrying a generation **older** than the connection's current one is **rejected without clearing `currentTask`** — so a revoked worker that wakes and sends `completion` cannot overwrite the new owner's state.
- **Every** release (disconnect, ready-abandon, operator requeue, lease-TTL expiry) bumps the generation, fencing the revoked worker.
- **Backward compatible**: an unversioned relay echoes `0`, which falls back to the pre-existing `TaskID` identity match, so old clients are unaffected.
- The counter is a **lock-free `atomic.Uint64`** so it is safe to mint from paths already holding `h.mu` (the requeue/reclaim loops run under `h.mu.RLock`); a mutex-guarded counter would deadlock, and lock-free keeps the `-race` job happy.

## Reuse of the #2649 configurable cooldown

Every release routes through `recordTaskFailure`, so recovery honors the same operator-configurable cooldown/quarantine and **cannot recreate the duplicate-assignment race #2492 fixed**. No new cooldown logic was introduced.

## Auth boundary

The requeue endpoint stays **owner/read-write only** (`requireContributorWrite`) — never an unauthenticated control (the class of bug #2610 addressed). A `read` viewer gets 403.

## Tests (`contribute_lease_test.go`)

- `generationAccepted` fences a stale generation, accepts the current one and the unversioned `0`.
- Lease expiry auto-releases + cools down + bumps the generation; a **progressing task is NOT reclaimed** (no false reclaim); an idle connection is ignored.
- Requeue records the reason + bumps the generation; a blank reason falls back to the default label; the endpoint enforces the reason plumbing + auth (read viewer 403).
- **The Gate (WS end-to-end):** assign → learn generation → operator revoke → a stale-generation `task_complete` is **rejected** (`TasksCompleted` stays 0, new owner survives); a control leg confirms a current-generation completion IS recorded. Plus at-most-one-authoritative-lease.
- `-race` clean.

Fixes #2568

### DIFF
--- a/v2/pkg/dashboard/api_contribute.go
+++ b/v2/pkg/dashboard/api_contribute.go
@@ -2773,8 +2773,11 @@ onEl('clanker-list','click',function(e){
     // returned to the queue and booked for the same short cooldown as an auto-release,
     // so it is not instantly re-handed to a stale worker. Uses the existing
     // POST /api/contributors/{id}/requeue endpoint (owner/read-write only).
-    adminConfirm('Requeue '+user+'&rsquo;s task','Release the task '+user+' is currently holding back to the ready queue. Use this when a connected clanker is wedged (connected but not progressing). The task is booked for the same short cooldown as an automatic release, so it is not instantly re-assigned to a stale worker. This uses the existing POST /api/contributors/{id}/requeue endpoint.','Requeue',function(){
-      fetch('/api/contributors/'+encodeURIComponent(cid)+'/requeue',{method:'POST'})
+    adminConfirm('Requeue '+user+'&rsquo;s task','Release the task '+user+' is currently holding back to the ready queue. Use this when a connected clanker is wedged (connected but not progressing). The task is booked for the same short cooldown as an automatic release, and the assignment generation is bumped so a stale worker cannot later overwrite the new owner. This uses the existing POST /api/contributors/{id}/requeue endpoint.','Requeue',function(){
+      // #2568: let the operator attach an optional reason. It is recorded in the
+      // audit + activity log and pushed to the still-connected worker on task_revoke.
+      var reason=(window.prompt('Reason for releasing this task (optional):','wedged: no progress')||'').trim();
+      fetch('/api/contributors/'+encodeURIComponent(cid)+'/requeue',{method:'POST',headers:{'Content-Type':'application/json'},body:JSON.stringify({reason:reason})})
         .then(function(r){return r.json().then(function(d){return{ok:r.ok,d:d};});})
         .then(function(x){if(x.ok){toast('Task requeued for '+user,true);opsPoll();}else{toast((x.d&&x.d.error)||'Requeue failed',false);}})
         .catch(function(){toast('Requeue failed',false);});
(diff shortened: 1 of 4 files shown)


clubanderson:
⚠️ **Quality flag (cross-PR integration):** `reclaimExpiredLeases` clears `currentTask`/`currentPrompt`/`currentLabels`/`tokenMintedAt` and bumps `currentTaskGen`, but does not clear `pendingToken`/`credentialDelivered` (introduced by PR #2671). When both merge, a lease expiry on a connection with a pending credential will leave stale token state.

**Fix:** Add `c.pendingToken = ""` and `c.credentialDelivered = false` inside the `if expired` block (alongside the existing field clears), and the same in the updated `RequeueContributorTask`.

Filed as #2675.

PR #2735, opened 2026-08-06
🐛 fix(proxy): reassemble full ClientHello before MITM so Copilot agents can authenticate

clubanderson:
## The bug

Every Copilot agent in the console **v4** hive could not authenticate. The proxy's own log showed, deterministically on fresh pods:

```
WARN transparent proxy TLS handshake failed error="local error: tls: bad record MAC"
```

Agent → `api.github.com:443` → iptables REDIRECT → proxy `:18443` → the proxy's **server-side** TLS handshake **with the agent** fails → the agent gets an empty reply → Copilot reports:

> Authentication token found but could not be validated / error sending request for `https://api.github.com/copilot_internal/user`

A `kubectl rollout restart` did **not** fix it (reproduced on a fresh pod within 90s), so it is a code-level framing bug, not boot state.

## Root cause: single-read peek + partial ClientHello

`handleTransparentTLS` did a **single** `conn.Read` after the 1-byte peek and assumed it captured the whole ClientHello. **TCP guarantees nothing about record boundaries** — when the agent's ClientHello arrives split across TCP segments, that single read returns only a prefix. A truncated ClientHello breaks the MITM deterministically:

1. `extractSNI` parses the partial record, misses the SNI, and `host` falls back to `github.com` instead of the real `api.github.com`;
2. `forgeCert` then mints a cert for the **wrong** host;
3. the agent-side TLS handshake fails — surfacing as `bad certificate` or `local error: tls: bad record MAC` depending on TLS version/timing.

The failure occurs **before** any upstream dial, so the SO_MARK upstream-dial change (#2678) is provably **not** involved and is left untouched.

## The fix (one focused change)

Read the **complete** first TLS record before extracting SNI and replaying it. New `readClientHelloRecord()` reads the record-layer length field and loops until the whole ClientHello record is buffered (bounded by the existing 4096 cap and the caller's read deadline), so SNI extraction and the `prefixConn` replay always see an intact ClientHello regardless of how TCP fragments the bytes. Oversized ClientHellos still complete because the remainder and subsequent records are served to `tls.Server` via `prefixConn`+`conn`.

Scope is limited to the transparent-TLS peek/replay path.

## Tests (new: `v2/pkg/proxy/proxy_transparent_tls_repro_test.go`)

- **`TestTransparentTLSPartialClientHello`** drives the real `readClientHelloRecord → extractSNI → forgeCert → tls.Server` path over a fragmenting connection.
  - **RED** (old single-read peek): `SNI extracted = "github.com", want "api.github.com"` → `remote error: tls: bad certificate`.
  - **GREEN** (this fix): correct SNI, handshake succeeds.
- **`TestReadClientHelloRecordReassembles`** unit-tests the helper against a ClientHello delivered 1 byte at a time.

Focused test run (this fix, on this branch):

```
--- PASS: TestTransparentTLSPartialClientHello (0.00s)
--- PASS: TestReadClientHelloRecordReassembles (0.00s)
ok  github.com/kubestellar/hive/v2/pkg/proxy
```

### DIFF
--- a/v2/pkg/proxy/github_proxy.go
+++ b/v2/pkg/proxy/github_proxy.go
@@ -302,16 +302,20 @@ const (
 // tried to connect to github.com:443 but iptables sent it here instead.
 // We extract the SNI hostname from the TLS ClientHello, then MITM the connection.
 func (p *GitHubProxy) handleTransparentTLS(conn net.Conn, peeked []byte) {
-	// Read enough of the ClientHello to extract SNI.
-	buf := make([]byte, tlsClientHelloMaxSize)
-	copy(buf, peeked)
+	// Read the whole ClientHello TLS record before extracting SNI and replaying
+	// it. TCP guarantees nothing about record boundaries: a single Read can
+	// return only PART of the ClientHello. A partial read makes extractSNI miss
+	// the SNI (so the wrong host is chosen and the wrong cert is forged) and
+	// mis-frames the record stream handed to tls.Server — the agent-side
+	// handshake then fails deterministically (bad certificate / bad record MAC),
+	// which is exactly why every Copilot agent could not authenticate. Loop until
+	// the full record (per the record-layer length) is buffered.
 	conn.SetReadDeadline(time.Now().Add(transparentProxyTimeout))
-	n, err := conn.Read(buf[len(peeked):])
+	fullBuf, err := readClientHelloRecord(conn, peeked)
 	conn.SetReadDeadline(time.Time{})
 	if err != nil {
 		return
 	}
-	fullBuf := buf[:len(peeked)+n]
 
 	host := extractSNI(fullBuf)
 	if host == "" {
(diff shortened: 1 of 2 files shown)


clubanderson (commented):
## Quality Review

**Tests**: The two new tests directly reproduce and guard the root cause — `TestTransparentTLSPartialClientHello` drives the real peek→SNI→forge→handshake path over a fragmenting connection, and `TestReadClientHelloRecordReassembles` unit-tests the helper at 1-byte-per-read granularity. Both are well-targeted.

**Fix correctness**: `readClientHelloRecord` correctly parses the TLS record-layer length and loops until the full ClientHello is buffered, with graceful degradation on short reads and a cap at `tlsClientHelloMaxSize`. The scope is tightly limited to the transparent-TLS peek/replay path.

**Minor observations** (non-blocking):

1. In the second `for n < need` loop (body reassembly), the `if n == 0 { return nil, err }` branch is unreachable — `n` is guaranteed ≥ `tlsRecordHeaderLen` (5) at that point since we passed the header loop. Harmless but dead code.

2. Potential edge case tests for follow-up (not blocking this PR):
   - EOF before the 5-byte header is complete (e.g., connection reset after 3 bytes)
   - `recordLen` declaring a body larger than `tlsClientHelloMaxSize - 5` (the overflow cap path)
   - `peeked` containing >1 byte (e.g., if the peek layer captures 2–3 bytes)

Overall the fix is well-scoped and well-tested. 👍

---
*Quality agent review (hold-gated mode)*

PR #2736, opened 2026-08-06
fix: wrong-forge repair targets the hive's elected forge, not the cluster app

clubanderson:
## The residual bug (proven live 2026-08-05) — two halves, one root cause

The hive's forge record lacked a single authoritative source: the delivery gate compared against the CLUSTER's App instead of the hive's elected forge's, and spoke reports could overwrite the meta record they were themselves downstream of. Together they produced (1) a dead repair gate and (2) a feedback loop that reverted every hand-repair within minutes.

### Half 1: the dead delivery gate

PR #2732's wrong-forge repair gate in `v2/pkg/hub/cluster_app_key.go` `decideAppKeySync` (~line 716) was:

```go
if spokeOnWrongForge && spokeAppID != 0 && spokeAppID != cluster.AppID
```

For a PUBLIC-github.com hive on the GHE cluster whose spoke was mis-flipped onto the CLUSTER's own App (5686), the identity handed to the gate leaked the cluster's App, so the comparison saw the spoke's broken App on both sides (**5686 == 5686**) and never fired — even though `spokeOnWrongForge` was true. **Live evidence:** six public hives (llm-d wva, llm-d-incubation, kellyaa, kalantar, ai-native, alchemy) with meta correctly restored to `github_host=github.com` received **zero re-deliveries** across many heartbeats. Both leak paths reproduce in tests that fail on `origin/v2`:

1. **No cluster names a public App** → resolver produced AppID 0 → `appIdentityForHive` returned nil → repair structurally unreachable.
2. **A leaked cluster App under the public label** — an under-specified cluster record (flat GHE app_id, no urls, no default_forge) defaults to public, so `forgesForCluster` keys the GHE App under `github.com` → the literal 5686 == 5686 dead gate.

**Fix:** the repair compares against — and delivers — the App of the **hive's elected forge**:
- `ResolveHiveIdentityInFleet` refuses to borrow a fleet App whose issuing forge (`config.ForgeOfAppID`) provably differs from the elected forge (unknown Apps still adopted — silence is never evidence).
- `appIdentityForHive` falls back to the build's own App constant for the elected forge (`builtinAppOfForge`: public 3568013 / GHE 5686) when no cluster entry names one — scoped to `FromHiveIntent`.
- `appKeyConfigForHeartbeat` verifies, whenever `spokeOnWrongForge` is true, that the delivered identity genuinely belongs to the elected forge; a leaked wrong-forge App is rebuilt (key looked up by the App it belongs to, urls stated explicitly) or the repair stands down.
- `decideAppKeySync`'s parameter renamed `cluster` → `identity`; the gate now genuinely reads `spokeAppID != identity.AppID` where identity = the elected forge's App (implied by `spokeOnWrongForge` for a coherent identity; kept as a guard).

### Half 2: the hand-repair revert loop

Observed on `hosted-kalantar-msb-soft-reflect-jsro` and threatened on all 9 — **all six hand-restored metas were re-stomped github.com → github.ibm.com within ~30 minutes**:

1. Operator restores meta `github_host` → github.com and the spoke to the public App.
2. `reconcileGitHubHostFromSpoke` re-adopts github.ibm.com from a spoke report made while the spoke was still (or again) on the GHE identity — the stand-down only covered AUTH-FAILING spokes, and a freshly-booted spoke reports before its first github call fails, so it looked healthy.
3. With meta back at GHE and the spoke on public, the EPM-class branch fires (3568013 ≠ 5686) and re-delivers the GHE identity with `ResetInstallation`. Loop complete.

**Fix — the meta host is operator/request truth; a spoke's report is downstream of hub delivery and never independent evidence:**
- `reconcileGitHubHostFromSpoke` **never overwrites a non-empty `github_host`**. Contradictions are resolved by moving the SPOKE onto the recorded forge (the wrong-forge delivery repair), never the record onto the spoke's. The old public→GHE correction of an explicit github.com record (vllmd-06 class) is deliberately no longer automatic — it was indistinguishable from mis-delivery laundering.
- The EMPTY-host backfill requires **positive working evidence**, not absence-of-failure (the fresh-boot gap): the spoke's reported app_id must map to the same forge as its urls, with a non-zero installation_id and no reported auth failure.

### docs_installation_id artifact

Verified: the delivery path never writes `docs_installation_id`. The observed "old public installation parked in docs_installation_id (146551814)" is the provisioned docs-org add-on value (commonly provisioned equal to `installation_id` on public hives) surviving a reset that correctly cleared only `installation_id`. Stale is non-fatal (docs token mint warns and retries) and it becomes valid again on a flip-back; clearing it would permanently disable docs tokens since it has no rediscovery flow. Documented on `nextInstallationID`.

### Both directions + the loop verified

- GHE-host hive + spoke on public App → GHE identity delivered (EPM class) ✔
- public-host hive + spoke on GHE/cluster App → public identity with reset ✔
- elected forge == cluster forge, converged → no delivery ✔
- unknown spoke App → protected as a deliberate pin ✔
- **full loop** (`TestHandRepairDoesNotRevert`): restored meta + fresh-boot GHE report → record holds, same beat delivers the public App back, convergence goes quiet ✔

Tests: `TestWrongForgeRepairTargetsElectedForge` (5 subtests), `TestHandRepairDoesNotRevert`, hardened `TestReconcileGitHubHostFromSpoke` (12 subtests), extended `TestBrokenSpokeForgeReportNotAdopted`. `go build`/`go vet` clean; full `pkg/hub` + `cmd/hive` + `pkg/config` suites pass.

Code only — no live cluster changes.

### DIFF
--- a/v2/cmd/hive/main.go
+++ b/v2/cmd/hive/main.go
@@ -188,6 +188,17 @@ func prospectiveGitHubIdentity(cur config.GitHubConfig, ghCfg *hub.HeartbeatGitH
 //	                     hub does not track; reading that as a clear would blank
 //	                     a working installation fleet-wide and turn a key-only
 //	                     fault into a total auth outage.
+//
+// docs_installation_id is DELIBERATELY untouched by all three cases. It is the
+// optional docs-org add-on installation, has no rediscovery flow (zero simply
+// disables the docs token refresh, permanently), and a stale value is
+// non-fatal: the periodic docs mint warns and retries. After an app-changing
+// delivery it can therefore briefly equal the PREVIOUS app's installation id —
+// which looks like the old installation_id was "parked" there, but is just the
+// provisioned docs value (public hives commonly provision both fields with the
+// same installation) surviving a reset that correctly cleared only
+// installation_id. On a flip-back to the original App it becomes valid again
+// on its own.
 func nextInstallationID(current int64, ghCfg *hub.HeartbeatGitHubAppConfig) (next int64, reset bool) {
 	if ghCfg == nil {
 		return current, false
--- a/v2/pkg/hub/cluster_app_key.go
+++ b/v2/pkg/hub/cluster_app_key.go
@@ -261,6 +261,23 @@ func (s *HubServer) appIdentityForHive(h *SaaSHive, clusterID string) *clusterAp
 		return nil
 	}
 	resolved := ResolveHiveIdentityInFleet(h, &c, s.forgeAppsAcrossFleet())
+	// A hive that ELECTED a forge no cluster entry names an App on still has a
+	// real App when that forge's App is a build constant (public github.com, the
+	// known GHE instance). Without this, a public-elected hive on a GHE-only
+	// fleet resolved AppID 0, this function returned nil, and the wrong-forge
+	// repair could never deliver — the hub held the right answer (github_host
+	// restored to github.com) and answered nothing, beat after beat, which is
+	// how six restored public hives stayed on the GHE App on 2026-08-05.
+	// Scoped to FromHiveIntent: a cluster-default resolution that names no App
+	// still yields nil, so no cluster gets an identity invented for it.
+	if resolved.AppID == 0 && resolved.FromHiveIntent && resolved.Forge != "" {
+		if appID, slug := builtinAppOfForge(resolved.Forge); appID != 0 {
+			resolved.AppID = appID
+			if resolved.AppSlug == "" {
+				resolved.AppSlug = slug
+			}
+		}
+	}
 	if resolved.AppID == 0 {
 		return nil
 	}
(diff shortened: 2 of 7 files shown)


clubanderson (commented):
## Quality Review

**Test coverage: ✅ Strong**

`TestWrongForgeRepairTargetsElectedForge` (5 subtests) covers:
1. Public-elected hive mis-flipped onto cluster App (no public App in fleet) — the exact live failure
2. Leaked cluster App under the public forge label — the 5686 == 5686 dead gate
3. GHE-elected hive on public App → GHE identity (regression guard for #2732's original direction)
4. Converged cluster-forge hive — healthy case untouched
5. Unknown spoke App — deliberate pin protected

All tests reproduce the live symptom on `origin/v2` and pass with the fix — good fail-first methodology.

**Observations:**
- `builtinAppOfForge` is a new exported-level function tested only indirectly through integration paths. A direct unit test would document its contract (especially the "unknown forge → 0" case), but coverage is adequate through subtests 1 and 2.
- The `ResolveHiveIdentityInFleet` coherence check (`config.ForgeOfAppID` guard) is well-explained in comments and exercised by subtest 2.
- The parameter rename `cluster` → `identity` throughout `decideAppKeySync` improves semantic clarity without changing behaviour — all existing callers updated.

**CI:** `build-and-test` ✅ | docker ✅ | all platform builds ✅ | coverage still running

No blocking findings. LGTM from a testing perspective.

---
*quality agent (hold-gated mode)*

PR #2756, opened 2026-08-06
fix(sec): per-hive authorization for hosted terminal (C3, CWE-862)

clubanderson:
### The hole
The hosted `/terminal` **authenticated any hub user but authorized none per-hive**:

1. The main spoke ingress enforces per-hive authz via `nginx.ingress.kubernetes.io/auth-url: .../api/saas/auth-check?hive={{.ID}}...`, but the **`hive-terminal` ingress had no auth-url annotation** — so `/terminal` was never routed through the per-hive check.
2. The Node proxy terminal gate verified only that `hive_hub_user` is a validly-signed cookie for *some* hub user (C2/CWE-345 fixed the signature check), but **never checked that user against this hive** — the same cookie verifies at every hive because the hub secret and the `.hive.kubestellar.io` cookie are shared.
3. The cookie is `Domain=.hive.kubestellar.io`, 7-day, and any GitHub account completes hub OAuth (`ensureSaaSUser` auto-creates).

Net result: **one cookie from any free GitHub account opened a shell in any tenant's credential-holding pod.**

### The fix — two defenses

**1. Ingress annotation (`saas_provision.go`)** — the `hive-terminal` nginx Ingress now carries the same per-hive `auth-url` the main ingress has (`auth-check?hive={{.ID}}`), plus `auth-signin` and `auth-response-headers`. `handleSaaSAuthCheck` already authorizes against `user.Hives[hiveID]` and 403s a user with no access to this hive. This closes the hole on nginx clusters.

**2. Proxy-side per-hive check (`proxy/server.js`)** — defense in depth, and the **only** defense on OpenShift Routes (no nginx auth-proxy in front of the pod). The proxy now:
- reads `HIVE_ID` and the per-hive `HIVE_AUTHORIZED_USERS` allowlist (already injected at provision time by `authorizedUsersForHive`);
- after verifying the cookie signature, requires the verified user to be on **this** hive's allowlist. A validly-signed cookie for a user authorized only on another hive is rejected — **403** on HTTP, **socket closed** on WS;
- normalizes the `Host` header (strips `:port` and a trailing FQDN dot) before the `.hive.kubestellar.io` hosted-host match, closing a bypass where `hive-b.hive.kubestellar.io:443` / `hive-b.hive.kubestellar.io.` would otherwise be treated as non-hosted and skip the gate entirely.

Fail-closed policy: when the allowlist is populated (the normal hosted case — `authorizedUsersForHive` always emits at least the owner), an absent user is rejected. When the allowlist env is unset/blank, the proxy defers to the ingress auth-url (defense #1) rather than locking everyone out of a legitimately allowlist-less hive — so the proxy check is a strict tightening on top of the ingress, never a new outage.

### C2 overlap note
C2 (hub-secret domain separation) is being fixed in a **separate PR** touching the cookie-signing / key-derivation code. This PR **deliberately does not touch cookie minting or key derivation** — the per-hive binding is enforced from the independent per-hive allowlist, not a new cookie claim — so the two PRs should not conflict.

### Testable vs template-only
- **Node tests (`server.test.js`, runnable, temp-dir harness, no real hub):** a validly-signed cookie for a user authorized on hive A → **403 (HTTP) / closed socket (WS)** at hive B's terminal, including **port-suffixed** and **trailing-dot** Host variants; an authorized user still reaches the terminal; a forged-signature cookie is still 401. All green (`node server.test.js`, exit 0).
- **Go tests (`saas_provision_terminal_authz_test.go`, runnable):** the rendered nginx manifest's `hive-terminal` Ingress carries the per-hive `auth-url` (and `auth-signin` / `auth-response-headers`), and the per-hive `auth-url` appears on **both** the main and terminal ingress so they can't drift.

### Deferred / not in scope
- A full short-lived `{user, hive, role, expiry}` terminal-assertion redesign (a hive-scoped cookie minted at terminal-open time). The existing `MintSSOToken`/`ssoClaims` machinery already models this shape; wiring the hosted terminal through an SSO-style handoff would be the more principled long-term design but is a larger change. This PR implements the core the finding asked for: **the proxy rejects a cookie whose per-hive authorization ≠ this hive.**
- C2 (hub-secret domain separation) — separate PR.

Do not merge without review.

### DIFF
--- a/v2/pkg/hub/saas_provision.go
+++ b/v2/pkg/hub/saas_provision.go
@@ -2243,6 +2243,17 @@ spec:
           value: "{{.SessionKey}}"
         - name: HIVE_SSO_KEY
           value: "{{.SSOKey}}"
+{{- if .IsNginxIngress}}
+        # HIVE_INGRESS_AUTHZ tells the Node proxy that an nginx ingress
+        # auth-proxy sits IN FRONT of this pod and per-hive-authorizes every
+        # /terminal request (auth-url=auth-check?hive={{.ID}}) before it ever
+        # reaches the proxy. Only set on the nginx lane. On the OpenShift-Route
+        # lane there is NO auth-proxy — the Route forwards straight to the pod —
+        # so this is unset and the proxy is the ONLY per-hive gate, which makes
+        # it fail CLOSED on an empty allowlist (see server.js). SECURITY (C3).
+        - name: HIVE_INGRESS_AUTHZ
+          value: "true"
+{{- end}}
 {{- if .AuthorizedUsers}}
         - name: HIVE_AUTHORIZED_USERS
           value: "{{.AuthorizedUsers}}"
--- a/v2/pkg/hub/saas_provision.go
+++ b/v2/pkg/hub/saas_provision.go
@@ -2372,6 +2372,16 @@ metadata:
     cert-manager.io/cluster-issuer: {{.CertIssuer}}
     nginx.ingress.kubernetes.io/proxy-read-timeout: "3600"
     nginx.ingress.kubernetes.io/proxy-send-timeout: "3600"
+    # SECURITY (CWE-862, finding C3): the terminal opens a live shell inside a
+    # credential-holding pod, so it MUST enforce the SAME per-hive authorization
+    # the main ingress does — not merely "is the caller a signed-in hub user".
+    # Without this auth-url, any GitHub account that completes hub OAuth (the
+    # shared .hive.kubestellar.io cookie) reaches ANY tenant's /terminal. The
+    # auth-check endpoint verifies the caller against THIS hive's authorized
+    # users (user.Hives[{{.ID}}]) and 403s a user with no access to this hive.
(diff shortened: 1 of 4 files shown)


clubanderson:
Tightening pushed (754eb05a): removed the last fail-open.

Previously an empty `HIVE_AUTHORIZED_USERS` (ownerless hive) made the proxy defer/allow unconditionally — reachable on OpenShift Routes, which have no ingress auth-proxy, leaving the terminal wide open there. Now platform-aware:

- Hub injects `HIVE_INGRESS_AUTHZ=true` **only** on the nginx-ingress lane (where the ingress auth-proxy per-hive-authorizes before the proxy).
- Empty allowlist: **defer** when `HIVE_INGRESS_AUTHZ=true` (ingress is the gate; don't double-deny legit nginx access); **DENY** (403 / closed WS) otherwise — on OpenShift the proxy is the only gate, so it fails closed.

New tests: OpenShift + empty allowlist → cross-hive AND same-hive/any cookie both denied (HTTP+WS); nginx + empty allowlist → deferred. Go: `HIVE_INGRESS_AUTHZ` present on nginx render, absent on OpenShift render. Full proxy suite + Go template tests green.

PR #2759, opened 2026-08-06
✨ feat(hub): reset a live claimed hive back to the unassigned pool (#2748 Phase 2)

### LINKED ISSUE #2748: hub: reset a hive back to the unassigned pool (reuse namespace) instead of delete-and-orphan
## The request (Thread A)

> When I delete a hive, could it simply **RESET** its configuration and storage so I can return it to the unassigned pool?

Today `DELETE /api/saas/hives/{id}` (`handleDeleteHive`, `v2/pkg/hub/saas.go`) tears the hive down completely: `deprovisionHive` deletes the namespace, PV, OCI export/filesystem, disk record, and does user cleanup, then `removeRegistryEntry` drops the registry row. The Kubernetes namespace (`hive-hosted-<id>`) can never be renamed, so a fresh pool placeholder always gets a **new** namespace, and the torn-down one is orphaned.

The ask: instead of tear-down-and-orphan, offer a **"Reset to pool"** action that **wipes the hive's config/data but KEEPS its existing namespace**, re-marking it as an available pool placeholder so the namespace is **reused** rather than stranded.

This is distinct from Delete:
- **Delete** — remove the hive entirely (namespace + storage gone).
- **Reset to pool** — wipe the tenant's config/data/credentials, re-stamp the namespace identity back to a placeholder, and set the `SaaSHive` record back to `available` (org `available-<id>`, empty repos) so it can be re-assigned.

### Security requirement (must-have)
A namespace returned to the pool **must not carry the previous tenant's secrets**. The reset MUST wipe: `gh-user-token`, GitHub App private keys, inference-gateway/OpenRouter keys, Claude/Copilot creds under `/data/home`, beads, contributor profiles, `hive.yaml`, and `hive.yaml.dashboard`. Returning a slot to the pool with a prior tenant's credentials would be a serious credential-leak.

### Correctness pieces the design must cover
1. Storage/config wipe (what in `/data` must be cleared vs. what `deprovisionHive` deletes today).
2. `SaaSHive` record reset to the freshly-seeded placeholder shape.
3. Namespace identity re-stamp (`hive.kubestellar.io/*` labels + persisted `RegistryEntry.Namespace`/project fields) so the reused namespace stops advertising the old tenant.
4. The spoke: the running spoke is configured for the old project and must come back as an unclaimed placeholder.
5. UI surface: a "Reset to pool" row action (owner/admin), distinct from Delete.

A full design write-up (functions/files per step, the security wipe checklist, and the failure/rollback story) will be added as a comment.

## Related bug — partial-delete shown as failure (Thread B)

Deleting a hive named `unknown-1` popped a toast that read like **"delete failed"**, yet the row vanished. Root cause: `handleDeleteHive` always removes the registry entry (deliberately, to avoid stranded ghost rows) and returns one of two **success** statuses — `deleted` or `partially_deleted` (registry row gone, but no cluster config was available so namespace/PV/OCI may survive and need manual cleanup, carrying a `warning`). The dashboard's `deleteHive` rendered `partially_deleted` with an **`error`** toast, so a partial *success* looked like a failed delete.

Fix: render `partially_deleted` as an informational notice ("Removed from the hub; some cloud resources may need manual cleanup"), not an error. This is small and standalone and will be fixed regardless of the Thread A outcome.

clubanderson:
## Phase 2 of #2748 — reset a LIVE claimed hive back to the unassigned pool (credential-safe /data wipe)

Design comment (reviewable spec): https://github.com/kubestellar/hive/issues/2748#issuecomment-5206903131

Phase 1 (the record-only escape hatch for an assigned-but-unclaimed placeholder) already ships and is reused verbatim for the record half. This PR adds the **storage wipe + verify + fail-closed** state machine a delivered claim needs.

> ⚠️ **Destructive-storage feature — left OPEN for operator review.** Please review the wipe/verify logic and the security model before merging.

### The guarantee: wipe-by-default, verify-clean, fail-closed (NOT a delete-list)

The operator's central concern — *"how do we assure there is NO data left — beads, tmux, or anything else?"* — is answered by an **empty seed allowlist**, not an enumerated delete-list (which rots the day a future feature drops `/data/newthing`).

Code-confirmed: on a fresh provision the `/data` PVC starts **completely empty**. The `copy-config` init container seeds the ConfigMap into `/etc/hive` (an emptyDir), never onto the PVC; everything else on `/data` is recreated by `entrypoint.sh` on boot (idempotent `mkdir -p` / `cp -n` / runtime generation). Therefore:

- **Seed allowlist = ∅.** The wipe destroys everything under `/data`; the verify contract is total and trivial: `find /data -mindepth 1` must return **nothing**. Anything left ⇒ **FAIL CLOSED**.
- The wipe and verify Jobs derive their `find(1)` prune expression from the **same** `resetToPoolSeedAllowlist`, so what the wipe keeps and what the verify tolerates can never drift.
- On restart the spoke re-seeds `/data` exactly as on a first-ever boot — byte-for-byte the fresh-provision baseline.

### Sequence (each step fail-closed; reachable clusters only)

1. **Reachability gate** — a heartbeat-only / unreachable cluster is **refused** (a wipe there cannot be verified; use Delete instead).
2. **Scale the spoke to 0** and wait — no agent/tmux/writer can recreate state mid-wipe.
3. **Wipe Job** — a short-lived root Job mounting the same `hive-data` PVC; `find /data -mindepth 1 -maxdepth 1 -exec rm -rf {} +` (root, so it removes files chowned to the agent uids / `.copilot` group-rw / sqlite locks). Bounded by `activeDeadlineSeconds` + `backoffLimit`; OpenShift SCC honoured.
4. **Verify `/data` clean** — a verify Job lists any leftover top-level entry; non-empty ⇒ **FAIL CLOSED**.
5. **Reset the Secret** — delete + recreate `hive-secrets` in the placeholder shape (fresh `dashboard-token` + placeholder `github-token`; **no** `gh-app-key*.pem`, no tenant creds) and **verify its key set** — any extra key ⇒ **FAIL CLOSED**.
6. **Record reset — only after verified clean** — reuse Phase 1's `resetHiveToAvailablePlaceholder`, drop the owner grant (like access-revoke; quota untouched — it's increment-only in the codebase, called out as deliberate), reopen the provision request, re-stamp the namespace identity to placeholder.
7. **Scale the spoke back to 1** — boots fresh from the re-seeded empty `/data` as an unclaimed placeholder.

**Fail-closed state:** `ResetToPoolStatus` = `in-progress` while running, `failed` (with a reason) on any wipe/verify error. The record flips to `available` **only** on a verified-clean result; on failure the hive stays claimed and the namespace is left intact for manual inspection. Idempotent + re-runnable.

### Security model (defense-in-depth for an irreversible destructive action)

- **Admin-only** (`requireAdmin`, which also applies `blockIfImpersonatingWrite` — a "View as user" session cannot trigger it, 403).
- **Typed confirmation** in the body: `confirm` must equal the hive id (mirrors `slackBroadcastConfirm`); else 400. The UI makes the operator TYPE the hive id.
- **Dry-run**: `dry_run:true` returns what would be wiped (the `/data` listing + Secret key names + namespace) and deletes nothing.
- **Same-origin check** over `SameSite=Lax` (rejects a present-but-untrusted `Origin`/`Referer` via the existing `isTrustedOrigin`).
- **State guard**: 409 unless the hive is a resettable LIVE hosted hive (`ClaimDelivered`).
- **Concurrency latch**: 409 while a reset is already in flight.
- **Audit** on every attempt (success AND fail-closed); **secret values never logged** — key names / paths only.

### Left out as unsafe (reported, not shipped half-done)

- **Heartbeat-only / unreachable clusters** (the vllm-d class): the hub cannot run — or verify — a wipe Job there, so reset-to-pool refuses rather than shipping an unverifiable wipe (exactly the leak #2748 forbids). Reachable clusters (the common case) get the full verified wipe.
- **Quota reversal**: `SaaSQuota` is increment-only across the codebase; this drops the owner's `Hives[id]` grant but does not decrement quota, matching existing behavior.

### Tests

Pure/unit-level (no shared-global mutation, `-race`-safe): the empty-seed verify fails closed on a `/data` tree with a stray bead dir / tmux socket / creds file / per-agent CODEX_HOME and passes on the clean sentinel; the seed-allowlist-is-empty invariant is pinned; the Secret-reset placeholder shape and the surviving-App-key fail-closed detection; the live-hive state guard; the record-reset shape; and the same-origin check (trusted passes, untrusted rejected, absent allowed).

### Files

- `v2/pkg/hub/saas_reset_to_pool.go` — handler, wipe/verify/record state machine, security gates, kubectl primitives.
- `v2/pkg/hub/saas_reset_to_pool_manifests.go` — wipe/verify Job + placeholder Secret manifest builders (shared seed allowlist).
- `v2/pkg/hub/saas.go` — route registration + UI row-menu action + `resetToPool()` JS (dry-run preview, typed confirm, fail-closed surfacing).
- `v2/pkg/hub/saas_provision.go` — `ResetToPoolStatus` / `ResetToPoolError` fields on `SaaSHive`.
- `v2/pkg/hub/reset_to_pool_test.go` — tests.

Fixes #2748

### DIFF
--- a/v2/pkg/hub/reset_to_pool_test.go
+++ b/v2/pkg/hub/reset_to_pool_test.go
@@ -1,0 +1,253 @@
+package hub
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"sort"
+	"strings"
+	"testing"
+)
+
+// --- verify / allowlist (the guarantee) --------------------------------------
+
+// TestResetToPoolParseLeftover_FailsClosedOnJunk asserts that a /data tree with
+// arbitrary accreted junk (a bead dir, a tmux socket, a stray creds file, a
+// per-agent CODEX_HOME) is detected as NOT clean — every unexpected entry is
+// surfaced as a leftover, so the reset fails closed. A tree with only the seed
+// (verify prints the clean sentinel) is detected as clean.
+func TestResetToPoolParseLeftover_FailsClosedOnJunk(t *testing.T) {
+	// The verify Job prints one leftover /data path per line (or the clean
+	// sentinel when there is nothing unexpected).
+	dirtyLogs := strings.Join([]string{
+		"/data/beads",
+		"/data/tmux-1001/default",   // a tmux socket dir
+		"/data/gh-user-token",       // a stray credential file
+		"/data/home/.codex-scanner", // a per-agent CODEX_HOME
+		"/data/hive.yaml.dashboard",
+	}, "\n")
+
+	leftover := resetToPoolParseLeftover(dirtyLogs)
+	if len(leftover) != 5 {
+		t.Fatalf("expected 5 leftover entries (fail-closed), got %d: %v", len(leftover), leftover)
+	}
+	// Sanity: the credential file and the per-agent CODEX_HOME must both be caught.
+	found := map[string]bool{}
+	for _, e := range leftover {
+		found[e] = true
+	}
+	for _, must := range []string{"/data/gh-user-token", "/data/home/.codex-scanner"} {
+		if !found[must] {
+			t.Errorf("leftover detection missed %q — would leak", must)
+		}
+	}
+}
+
+// TestResetToPoolParseLeftover_CleanSentinelPasses asserts that the clean
+// sentinel (and only the sentinel) is treated as a clean /data.
+func TestResetToPoolParseLeftover_CleanSentinelPasses(t *testing.T) {
+	if got := resetToPoolParseLeftover(resetToPoolCleanSentinel + "\n"); len(got) != 0 {
+		t.Fatalf("clean sentinel must parse as no leftover, got %v", got)
+	}
+	// Blank/whitespace-only logs are also clean (nothing to wipe).
+	if got := resetToPoolParseLeftover("\n   \n"); len(got) != 0 {
+		t.Fatalf("blank logs must parse as no leftover, got %v", got)
+	}
+}
+
+// TestResetToPoolSeedAllowlistIsEmpty pins the design invariant: the seed
+// allowlist is empty, so the wipe is total and the verify tolerates nothing. If
+// a future change adds a seeded PVC file it must be a deliberate edit here (which
+// this test will flag) so the wipe and verify stay in lockstep.
+func TestResetToPoolSeedAllowlistIsEmpty(t *testing.T) {
+	if len(resetToPoolSeedAllowlist) != 0 {
+		t.Fatalf("seed allowlist must be empty (total wipe); got %v — if intentional, update this test and confirm both wipe and verify honour it", resetToPoolSeedAllowlist)
+	}
+	// With an empty allowlist the find prune expression must be empty, so nothing
+	// is excluded from either the wipe or the verify.
+	if expr := resetToPoolFindPruneExpr(resetToPoolSeedAllowlist); expr != "" {
+		t.Fatalf("empty allowlist must yield an empty prune expr, got %q", expr)
+	}
+}
+
+// TestResetToPoolFindPruneExpr_Allowlisted asserts the prune expression correctly
+// excludes an allowlisted name (used if the allowlist ever becomes non-empty) and
+// shell-quotes it defensively.
+func TestResetToPoolFindPruneExpr_Allowlisted(t *testing.T) {
+	expr := resetToPoolFindPruneExpr([]string{"hive.yaml", "keep me"})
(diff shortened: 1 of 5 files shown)


clubanderson:
Closing per operator decision to hold off on reset-to-pool for now — delete is sufficient. The full design (fail-closed allowlist wipe + verify + security model: admin-only, typed-confirm, dry-run, impersonation-block, same-origin, audit) is preserved on #2748 if we revisit. Not deleting the issue so the design stays discoverable.

PR #2701, opened 2026-08-05
✨ feat(inference): add IBM watsonx.ai as a first-class gateway kind (Granite)

clubanderson:
## What

Adds **IBM watsonx.ai** as a first-class inference gateway kind (`watsonx`), alongside `vllm` / `llm-d` / `openrouter` / `litellm` / `custom`, so hive agents can run **IBM Granite** (and other watsonx-hosted) models.

Fixes #2700

## Research — the three facts, with evidence

**1. Endpoint (OpenAI-compatible surface).** watsonx.ai ships an OpenAI-compatible **model gateway** at `https://<region>.ml.cloud.ibm.com/ml/gateway/v1`. Verified directly against **IBM's own watsonx-ai API reference** (https://cloud.ibm.com/apidocs/watsonx-ai): the `Model Gateway` operations include `POST /ml/gateway/v1/chat/completions`, `GET /ml/gateway/v1/models`, `/ml/gateway/v1/embeddings`, `/ml/gateway/v1/providers`, and the chat/completions responses are OpenAI-shaped (`object: chat.completion`, `choices`, `usage`). All Model-Gateway ops declare `security: [{ Bearer: [] }]`.

Design consequence: the operator configures the endpoint as `.../ml/gateway` and hive's existing `strings.TrimRight(endpoint,"/") + "/v1/models"` / `+ "/v1/chat/completions"` suffixing reaches `.../ml/gateway/v1/models` and `.../ml/gateway/v1/chat/completions` **unchanged** — no special URL handling. The UI preset fills this base for `us-south`.

**2. Auth — IAM token mint (implemented).** The gateway's `Bearer` scheme is defined in the apidocs as *"an access token obtained from the authentication endpoint"* (`bearerFormat: JWT`) — i.e. an **IAM access token**, minted from an IBM Cloud API key via `POST https://iam.cloud.ibm.com/identity/token` with `grant_type=urn:ibm:params:oauth:grant-type:apikey&apikey=<key>` (corroborated by heidloff.net's watsonx REST walkthrough and the ibm-watsonx SDKs; the native `/ml/v1/text/*` API also requires this token). So hive **mints and caches** the IAM token rather than sending the raw key.

`pkg/watsonx` implements the exchange with per-key caching to the token's `expires_in` (minus a 5-min skew margin; all TTLs/timeouts are named consts), re-minting on expiry. The API key and the minted token are **never logged**; IAM/upstream error bodies are redacted of both.

[...]

PR #2777, opened 2026-08-07
feat(agent): register watsonx as an agent backend + validate backends at config time

clubanderson:
## The bug (live, reproduced on a real spoke)

An operator configured agent `ci-maintainer` with `backend: watsonx` (hive `hosted-kubestellar-hive-ojv6`). The agent never launches:

```
HIVE LAUNCH FAILED: backend watsonx did not launch: unknown backend: watsonx.
The CLI for this backend is not installed in this hive image — upgrade the hive image or switch this agent to a different backend.
...
{"level":"WARN","msg":"failed to send kick","agent":"ci-maintainer","error":"agent ci-maintainer cannot be kicked: it failed to start: unknown backend: watsonx"}
```

The pod was bounced onto current `v2-latest` and still fails — **this is not image staleness.** Verified on v2 HEAD: the agent backend registry knew only `claude`, `codex`, `copilot`, `gemini`, `litellm`, `llm-d`, `vllm`. There is no `watsonx`.

### Root cause

watsonx code existed **only on the inference-gateway side**: `v2/pkg/watsonx/`, `v2/pkg/config/config.go`, `v2/pkg/proxy/inference_route.go`, `v2/pkg/dashboard/gateways.go`.

#2701 shipped watsonx as an inference **gateway kind**, and #2709 shipped it as a clanker **onboarding backend option** — but nobody taught the **agent launcher** to launch an agent whose `backend` is `watsonx`. The config layer accepted the value; the launcher rejected it at kick time. Classic silent accept-then-fail: the misconfiguration is invisible at the moment it's made and only surfaces hours later as an agent that quietly never runs.

## Part 1 — register `watsonx` as an agent backend

watsonx.ai is a **model gateway** (OpenAI-compatible endpoint + IAM-minted bearer token), not an agentic CLI, so it belongs with `litellm`/`vllm`/`llm-d` rather than with the CLI backends. It now joins `config.InferenceBackends`, the canonical list every gateway-backend check already reads.

- **Dispatch site extended:** `backendBinary` in `v2/pkg/agent/manager.go` — the exact function that emitted `unknown backend: %s`. Its gateway entries are now **derived from `config.InferenceBackends`** instead of being hardcoded, so a backend accepted by config can never again be unknown to the launcher.
- **Endpoint + auth reuse the existing plumbing, not reimplemented.** The launch path resolves the watsonx gateway by name-then-kind and takes endpoint, IBM Cloud key, project id and region from the configured `gateways:` entry. The IAM mint and `X-IBM-Project-ID` header are factored into a shared `resolveGatewayAuth` used by *both* the named-gateway branch and the new built-in `watsonx` branch, so the two cannot authenticate differently. `watsonx.DefaultMinter` (process-wide token cache) is reused as-is.
- **One endpoint template.** `watsonxEndpointForRegion` + the region/base-URL constants moved from `pkg/dashboard` into `pkg/watsonx` as `EndpointForRegion` / `BaseURLTemplate` / `DefaultRegion`, because the agent launch path needs the same template the UI preset uses. The dashboard helper is now a thin delegate. Two copies of that template is exactly how the gateway half and the agent half were free to drift.
- **No login badge.** `BackendRequiresInteractiveAuth("watsonx") == false` — auth is an API key exchanged for an IAM bearer, so there is no interactive login for an operator to perform and a 🔑 badge is always a false alarm (the rule from #2765).
- Model pinning, token accounting and pause/kick paths all key off `IsInferenceBackend`, so watsonx inherits litellm's behavior automatically.

## Part 2 — stop the silent accept-then-fail

This is the part that prevents the *next* occurrence. Backends are now validated **at set time**, against the same canonical lists the launcher dispatches on:

- **Centralized the list.** Added `config.CLIBackends`, `config.IsCLIBackend`, `config.SupportedBackends()` and `GovernorConfig.ValidateBackend()`. The validator's previously-inline `map[string]bool{"claude": true, ...}` is gone — `validate()` now calls the shared gate.
- **Gated every write path:** `hive.yaml` load (`config.validate`), and all three dashboard handlers that persist a backend (`handleAgentConfigModels`, the `cliPinValue` save, `handleGovernorAddAgent`) now return HTTP 400 with a message naming the offender and enumerating valid backends.
- **Gated `/switch/{backend}`.** `Manager.SetBackendOverride` previously accepted *any* string and restarted the agent into it; it now rejects a backend the launcher cannot dispatch via `validateBackendName`, which honors configured gateway names (resolved live).

Configured gateway *names* remain valid backends everywhere, matched case-insensitively to mirror `ResolveGateway`.

## Tests

- `watsonx` is a recognized backend — launching with `Backend: "watsonx"` never yields `unknown backend`.
- `BackendRequiresInteractiveAuth("watsonx") == false`, and the full auth-state path returns authenticated+known even with no credential files and a login-looking pane.
- Endpoint/region handling asserted against the shared `watsonx.EndpointForRegion` helper (region fallback, no trailing slash, no `/v1` — hive appends it).
- Config validation: `"totally-made-up"` is rejected at set time with a message listing valid backends; `watsonx` is accepted; a configured gateway name is accepted.
- **Anti-drift:** `TestEveryInferenceBackendHasABinary` asserts every backend config accepts is one the launcher can dispatch, and `TestBackendListsAreDisjoint` keeps the CLI and gateway lists from overlapping.
- The existing gateway-backend table tests (in `agent_unit_test.go`, `authprobe_test.go`, `manager_coverage2_test.go`, `config_test.go`, `litellm_test.go`, `agent_auth_peruid_test.go`) now **iterate `config.InferenceBackends`** rather than repeating `{litellm, vllm, llm-d}` inline, so the next backend added is covered everywhere automatically.

## Note on model semantics (follow-up, not this PR)

The reporting spoke had `backend: watsonx` with `model: claude-sonnet-4-6` — almost certainly an operator mis-set, since watsonx serves Granite et al., not Claude. **No model-name validation is added here** (model ids must pass through verbatim; rewriting them is what causes gateway 403s). Worth a follow-up: surface a clear runtime error on the launch path when the gateway rejects an unknown model, so this class of mis-set also fails loudly instead of silently.

### DIFF
--- a/v2/cmd/hive/main.go
+++ b/v2/cmd/hive/main.go
@@ -2460,23 +2460,7 @@ func main() {
 					// key (NOT the raw key), and scopes billing/limits by a
 					// project id sent as X-IBM-Project-ID. Mint (cached) and set
 					// both here; every other kind sends the resolved key verbatim.
-					apiKey := gw.ResolveAPIKey()
-					var extraHeaders map[string]string
-					if strings.EqualFold(gw.Kind, config.GatewayKindWatsonx) {
-						token, terr := watsonx.DefaultMinter.Token(context.Background(), apiKey)
-						if terr != nil {
-							logger.Warn("watsonx IAM token mint failed; agent inference will fail until the key/project are valid",
-								"agent", agentName, "gateway", backend, "error", terr.Error())
-							// Leave apiKey as-is (the raw key). watsonx will reject
-							// it, surfacing a clear upstream 401 rather than a
-							// silent success — better than dropping the route.
-						} else {
-							apiKey = token
-						}
-						if gw.ProjectID != "" {
-							extraHeaders = map[string]string{watsonx.ProjectIDHeader: gw.ProjectID}
-						}
-					}
+					apiKey, extraHeaders := resolveGatewayAuth(gw, agentName, backend, logger)
 					githubProxy.SetInferenceRoute(agentName, &proxy.InferenceRoute{
 						Backend:      backend,
 						Endpoint:     endpoint,
@@ -2539,6 +2523,41 @@ func main() {
 						Model:    model,
 						APIKey:   apiKey,
 						CABundle: caBundle,
+					})
+					return
+				}
+				if backend == config.GatewayKindWatsonx {
+					// Built-in "watsonx" backend: the operator set
+					// `backend: watsonx` without a gateway literally NAMED
+					// watsonx (a named one is handled by the gateway branch
+					// above). Resolve the watsonx gateway by KIND so the
+					// endpoint, IBM Cloud key, project id and region all come
+					// from the existing `gateways:` plumbing rather than being
+					// re-derived here.
+					gw := resolveWatsonxGateway(cfg)
+					if gw == nil {
+						logger.Warn("watsonx backend selected but no watsonx gateway is configured; add one under the Model Gateways tab",
+							"agent", agentName, "model", model)
+						return
+					}
+					// Region-only gateways are legal (the guided form can save a
+					// region without an endpoint), so fall back to the shared
+					// region template — the same helper the dashboard preset uses.
+					endpoint := strings.TrimSpace(gw.Endpoint)
+					if endpoint == "" {
+						endpoint = watsonx.EndpointForRegion(gw.Region)
+					}
+					if model == "" {
+						model = gw.DefaultModel
+					}
+					apiKey, extraHeaders := resolveGatewayAuth(gw, agentName, backend, logger)
+					githubProxy.SetInferenceRoute(agentName, &proxy.InferenceRoute{
+						Backend:      backend,
+						Endpoint:     endpoint,
+						Model:        model,
+						APIKey:       apiKey,
+						CABundle:     gw.CABundle,
+						ExtraHeaders: extraHeaders,
 					})
 					return
 				}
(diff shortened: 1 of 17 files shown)


clubanderson:
## Coverage gate: both failures diagnosed

### 1. \`agent: 88.0% (floor 89%)\` — mine, fixed

Registering watsonx added code to \`pkg/agent\` without matching tests. I did **not** lower the floor; I added tests for the paths this PR touches plus genuinely untested neighbours on those same paths (commit 42e050c7):

- **\`backendBinary\`** — every \`config.InferenceBackends\` entry resolves to the claude CLI (the backend name selects the upstream route, not the binary); agentic CLIs still map to their own binaries; an unknown backend still errors.
- **\`SetBackendOverride\`** — the accept-then-fail door this PR closed: an undispatchable backend is rejected **and** leaves \`BackendOverride\` unmutated; \`watsonx\` is accepted; a missing agent still reports "not found".
- **\`AgentAuthAvailable\`** — an unknown agent returns \`known=false\` ("no opinion" → no badge); a claude-configured agent overridden onto watsonx resolves via the **override**, so a switched agent never shows a spurious login badge.
- **\`agentCopilotConfigPaths\`** — per-UID paths precede the shared path (a stale shared credential must not mask a fresh per-agent login), and the shared path is never duplicated.
- **\`checkBlockedThrash\` / \`recordBlockedAndCheck\`** — non-blocked lines record nothing, every marker is recognised, the breaker trips only at the threshold, stale entries age out of the window, and the cooldown suppresses a repeat trip.

**pkg/agent coverage: 86.4% → 89.4%** (local \`go test ./pkg/agent/ -cover\`), above the 89% floor. Writing these also surfaced a real detail: \`backendBinary\` returns a PATH-resolved absolute path, so the assertions compare basenames.

### 2. \`hub: TESTS FAILING\` → \`TestHandleOAuthCallbackSuccess\` — PRE-EXISTING on v2, not this PR

This PR touches \`pkg/agent\`, \`pkg/config\`, \`pkg/watsonx\`, \`pkg/dashboard\` — no OAuth code. Verified rather than assumed, in a detached worktree at clean \`origin/v2\` (610f5937, which does **not** contain this branch):

\`\`\`
$ go test ./pkg/hub/ -run TestHandleOAuthCallbackSuccess -count=1   # x3 on clean origin/v2
--- FAIL: TestHandleOAuthCallbackSuccess (0.00s)
    oauth_provision_coverage_test.go:110: minted cookie did not verify: ("", false)
FAIL	github.com/kubestellar/hive/v2/pkg/hub
\`\`\`

Failed **3/3 runs on the untouched baseline** — deterministic, not flaky. The failure is therefore pre-existing on \`v2\` and independent of this PR; it is unrelated breakage that happens to block the shared coverage gate. Likely fallout from the recent hub cookie-signing work (#2775 signs \`hive_hub_user\` with a derived session key) whose test was not updated to the dual-verify path.

I have deliberately **not** touched it here — it belongs in its own issue/PR rather than being smuggled into a backend-registration change. It will keep the gate red until fixed separately.

clubanderson:
### Root cause of the pre-existing hub failure (for whoever picks it up)

Confirmed the mechanism, so the follow-up fix is unambiguous. It is fallout from #2775 (\`067f8f19\`), which is the commit immediately preceding this branch on \`v2\`:

- #2775 changed minting so \`hive_hub_user\` is signed with the **derived** session key (\`s.sessionCookieKey\` = \`deriveDomainKey(master, "hive-session-v1")\`), so a v4 spoke's Node proxy can verify it. The two hub-side verify sites (\`handleAuthUser\`, \`getRealAuthUser\`) were updated to \`verifyHubUserDual\`, accepting either the derived key or the raw master.
- \`TestHandleOAuthCallbackSuccess\` was **not** updated. It still verifies the minted cookie with the **raw** master:

\`\`\`go
// pkg/hub/oauth_provision_coverage_test.go:110
if user, ok := verifyHubUserCookieValue(testHubSecret, sessionCookie.Value); !ok || user != "octocat" {
    t.Errorf("minted cookie did not verify: (%q, %v)", user, ok)
}
\`\`\`

Since minting now uses the derived key and this assertion uses the raw one, it fails deterministically — matching the observed \`minted cookie did not verify: ("", false)\`.

**Fix**: point the assertion at \`verifyHubUserDual\` (or derive the key the same way the mint path does), rather than \`verifyHubUserCookieValue(testHubSecret, ...)\`. The production dual-verify behaviour is correct; only the test's expectation is stale.

clubanderson:
### CI confirms the coverage gap is closed

The gate ran on \`42e050c7\`. Comparing gate output before and after:

**Before:**
\`\`\`
COVERAGE GATE FAILED:
hub: TESTS FAILING

agent: 88.0% (floor 89%)
\`\`\`

**Now:**
\`\`\`
COVERAGE GATE FAILED:
hub: TESTS FAILING
\`\`\`

The \`agent:\` line is gone — CI measured \`pkg/agent\` at **89.6%** (and 89.5% on the gate's second pass), clearing the 89% floor:

\`\`\`
ok  github.com/kubestellar/hive/v2/pkg/agent  347.025s  coverage: 89.6% of statements
\`\`\`

Every other check on this commit is green: \`build\` (amd64 + arm64), \`build-hub\`, \`build-contributor\`, \`build-and-test\`, \`docker\`, \`gate\`, \`copilot-automation\`.

**The sole remaining blocker is \`hub: TESTS FAILING\`** — the pre-existing \`TestHandleOAuthCallbackSuccess\` failure from #2775, documented above. It is unrelated to this PR (reproduced 3/3 on clean \`origin/v2\`) and needs its own fix before this PR can go green. Nothing further on this branch will clear it.

PR #2780, opened 2026-08-07
🐛 Harden tmux and process cleanup

### LINKED ISSUE #2743: Security: hive code and tests can terminate unrelated tmux sessions / the whole tmux server (v4)
## Observed behavior

During security reviews of the v4 branch over the last 24 hours we have observed **broad kills of the tmux server a hive test run was executing in** — i.e. termination of sessions hive did not create, including sessions started independently by the user on the same machine. The exact modality has not been pinned down yet.

A code review of the tmux-termination surface turned up several candidate paths, ranked below by likelihood/severity. No literal `tmux kill-server`, `kill-session -a`, or `pkill tmux` exists in the tree; the exposure is indirect.

### 1. Unguarded UID-based process sweep — `v2/pkg/agent/manager.go` (most likely)

`killAgentProcesses(uid, ...)` (manager.go:5268–5317) walks `/proc` and SIGKILLs **every process whose owner UID equals the argument**. Two call sites in `runCopilotDiagnostic` invoke it **without** the `agent.UID > 0` guard present at the other sites:

- manager.go:4001
- manager.go:4052

`agent.UID` is 0 whenever UID isolation is off or the agent is missing from the UID map (`LookupByName` returns 0 on miss, uidmap.go:87–91; `NewManager` treats that as normal, manager.go:637–642). When hive runs as root (its container does), `killAgentProcesses(0)` SIGKILLs every UID-0 process: the tmux server (all sessions on it, including independently started ones), hive itself, and PID 1. The comment at manager.go:5095 claiming this is "a no-op when agents share the dev UID" holds only for an unprivileged user via EPERM — it is false for root. The function also lacks the `os.Getpid()` self-skip its sibling `reapAgentCLI` has (manager.go:5207).

Trigger: a copilot-backend auth failure diagnostic.

### 2. Test suite drives the developer's default tmux server with hardcoded session names

Tests construct `NewManager` with `uidMap == nil`, so `tmuxSocket` is empty and `tmuxBaseArgs` falls back to plain `tmux` (manager.go:752–757) — the invoking user's **shared default tmux server**. Several tests additionally bypass the manager with direct `exec.Command("tmux", "new-session"/"kill-session", ...)`:

- `v2/pkg/agent/tmux_coverage_test.go:88,92` (`newRawTmuxSession`; 13 call sites across branches/dismiss/diag/poll/bob_startup_kick/launch/tmux coverage tests)
- `crash_coverage_test.go:29,32,57,60,87,120`, `launch_coverage_test.go:29,32,...`, `sendkick_coverage_test.go:31,34`

Session names are fixed strings (`hive-cxa`, `hive-covsk1`, ...). A collision with a live same-named session — or any failure mode that makes a kill target resolve to the wrong/default session — terminates sessions the test does not own. `CheckAndRestartCrashedAgents` tests also kill+recreate the session mid-test.

### 3. v2 Manager shares the user's default tmux server when UID isolation is off

All six production `kill-session` sites (manager.go:2660, 4002, 4053, 4965, 5109, 5351) route through `tmuxCmd`, which targets the default server whenever `tmuxSocket == ""`. Targets are always `hive-<agent>`, so kills are deliberate — but a name collision destroys a user session, and there is no validation that a kill target is non-empty and hive-namespaced.

### 4. v1 shell scripts and legacy dashboard kill sessions on the invoking user's default server

`bin/supervisor.sh:137,389,400`; `bin/agent-healthcheck.sh:57`; `bin/contributor-agent.sh:313,431`; `Justfile:455,469`; `dashboard/server.js:1183` (`/api/restart/:agent`); `uninstall.sh:39`. Mostly well-guarded (`:?` required vars, agent allowlist, `sudo -u` isolation), but local/contributor modes intentionally run on the host user's default server with fixed names (`contributor`, `hive`), so collisions kill user sessions.

### 5. Adjacent: `Justfile:410` — `pkill -f "ollama serve"`

Not tmux, but the same pattern class; as root it kills every user's ollama.

## Suggested remediation

1. Harden `killAgentProcesses`: refuse `uid < 1000` (agent UIDs start at 2001), add an `os.Getpid()` self-skip, and guard the two `runCopilotDiagnostic` call sites. Regression test against a fake `procRoot` asserting uid 0 kills nothing.
2. Run all tmux-touching tests against a private tmux server (`tmux -L hive-test-<pid>` socket or per-run `TMUX_TMPDIR`), threaded through `TestMain`; derive session names from `t.Name()`.
3. Give hive a dedicated tmux socket unconditionally (`tmux -L hive` in `tmuxBaseArgs` and `TMUX_BIN` in shell scripts) so hive-managed sessions never share a server with user sessions.
4. Centralize kill-target validation (reject empty/wildcard/non-`hive-` names) in Go, JS, and shell kill paths.

## Notes

- Reviewed on the `v4` branch (HEAD b8c45aef). Identical code exists in several `.slim/worktrees/*` copies.
- Happy to provide the full review notes or help reproduce the observed server kill.

clubanderson:
## Summary
- guard Copilot diagnostic UID sweeps and make killAgentProcesses refuse uid <= 0 while skipping the hive process
- validate tmux kill-session targets so only non-empty hive-* sessions can be killed through tmuxCmd
- isolate agent package tmux tests from the developer default server via a package-private TMUX_TMPDIR
- scope local ollama cleanup to processes owned by the current user

## Validation
- cd v2 && go build ./...
- cd v2 && go test ./pkg/agent/ -count=1
- cd v2 && go vet ./...

Fixes #2743

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -407,7 +407,7 @@ contribute-hive backend="" mode="docker": check-version
       # Start ollama silently if needed for goose
       if [[ "$BACKEND" == "goose" && "${GOOSE_PROVIDER:-}" == "ollama" ]]; then
         # Kill any existing ollama that might spam logs to this terminal
-        pkill -f "ollama serve" 2>/dev/null || true
+        pgrep -u "$(id -u)" -f "ollama serve" 2>/dev/null | while read -r pid; do kill "$pid" 2>/dev/null || true; done
         sleep 1
         echo "Starting ollama (silent)..."
         OLLAMA_FLASH_ATTENTION=1 nohup ollama serve > /dev/null 2>&1 &
--- a/v2/pkg/agent/manager.go
+++ b/v2/pkg/agent/manager.go
@@ -640,7 +640,38 @@ func (m *Manager) tmuxBaseArgs(agent *AgentProcess) []string {
 	return []string{"tmux"}
 }
 
+func validTmuxKillSessionTarget(args []string) (string, bool) {
+	if len(args) == 0 || args[0] != "kill-session" {
+		return "", true
+	}
+	for i := 1; i < len(args); i++ {
+		arg := args[i]
+		switch {
+		case arg == "-t":
+			if i+1 >= len(args) {
+				return "", false
+			}
+			target := args[i+1]
+			return target, strings.HasPrefix(target, "hive-")
+		case strings.HasPrefix(arg, "-t="):
+			target := strings.TrimPrefix(arg, "-t=")
+			return target, strings.HasPrefix(target, "hive-")
+		}
+	}
+	return "", false
+}
+
 func (m *Manager) tmuxCmd(agent *AgentProcess, args ...string) *exec.Cmd {
+	if target, ok := validTmuxKillSessionTarget(args); !ok {
+		agentName := ""
+		if agent != nil {
+			agentName = agent.Name
+		}
+		if m.logger != nil {
+			m.logger.Warn("refusing unsafe tmux kill-session target", "target", target, "agent", agentName)
+		}
+		return exec.Command("false")
+	}
 	base := m.tmuxBaseArgs(agent)
 	tmuxArgs := append(base[1:], args...)
 	if agent.UID > 0 {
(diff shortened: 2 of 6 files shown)


clubanderson:
Closing: this fix must target the v4 branch, not v2 (per operator). It will be re-opened against v4.

PR #2793, opened 2026-08-07
🐛 fix(agent): group-writable beads dirs + Go on agent shell PATH

bketelsen:
## Summary

Two failures observed on a fresh Kubernetes spoke running with per-agent UID isolation (agents as `hive-<name>` users, uid 2001+, sharing only the `node` group):

### 1. Agents cannot persist beads for runtime-created dirs

`beads.NewStore` created its directory with `os.MkdirAll(dir, 0755)`. The hive process runs as `dev` (uid 1001) after privilege drop, so beads dirs it creates are writable only by 1001 — agents fail with:

> `bd create` cannot write `/data/beads/<agent>/beads.json` due to permission denial

Dirs that exist at boot get re-chowned to the agent UID by the entrypoint, which masks the bug — but any beads dir created **at runtime** (an agent added from the dashboard, first bead after a roster change) stays group read-only until the next pod restart.

Fix: create the dir `0775` and explicitly `chmod 2775` (group-writable + setgid so new files inherit the `node` group). The explicit chmod is required because `MkdirAll`'s mode is clipped by the umask — which is exactly how the 755 dirs kept appearing.

### 2. Agents report "no Go toolchain" despite Go being in the image

The image ships Go at `/usr/local/go/bin`, but `/etc/profile` resets `PATH` to the Debian default in agent shells, so agents skip running tests:

> Coverage review is blocked: the environment has no Go toolchain

Fix: the `.bashrc` written by the entrypoint now restores `/usr/local/go/bin` (idempotently), and the entrypoint seeds a `~/.profile` chaining to `.bashrc` so login shells get the same environment (`~/.profile` is only written if absent, so operator customizations survive).

## Testing

- `go test ./pkg/beads/` passes
- `bash -n deploy/entrypoint.sh` clean
- Verified live on a k8s spoke: with the dir at 2775, `touch` as the agent UID succeeds where `bd create` previously failed; with the PATH line, `bash -lc 'go version'` as an agent user resolves go1.25.12.

### DIFF
--- a/v2/deploy/entrypoint.sh
+++ b/v2/deploy/entrypoint.sh
@@ -532,8 +532,18 @@ if [ -f /data/proxy-ca-bundle.pem ]; then
 else
   export SSL_CERT_FILE=/data/proxy-ca.pem
 fi
+# The image ships Go at /usr/local/go/bin but /etc/profile resets PATH to the
+# Debian default, so agent shells lose it — and agents then report "no Go
+# toolchain" when asked to run tests.
+case ":$PATH:" in *:/usr/local/go/bin:*) ;; *) export PATH="$PATH:/usr/local/go/bin" ;; esac
 BASHRC
   chmod 644 /data/home/.bashrc 2>/dev/null || true
+  # Login shells (tmux default-command) read ~/.profile, not ~/.bashrc — chain
+  # them so both shell flavors get the same environment.
+  if [ ! -f /data/home/.profile ]; then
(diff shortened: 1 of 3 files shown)


clubanderson:
Thanks for the fix, @bketelsen. I rebased/retargeted this onto v4 and kept the permissions change aligned with the existing per-UID isolation model: bead dirs are now explicitly chmodded to setgid 2770 (not world-readable 2775), with a regression test asserting 0770 privacy plus setgid on Linux. The Go PATH/profile fix is unchanged.

Validated locally from v2:
- bash -n deploy/entrypoint.sh
- go build ./...
- go test ./pkg/beads -count=1
- go vet ./pkg/beads

/lgtm
/approve

PR #2829, opened 2026-08-08
feat(contribute): add goose to HEADLESS_BACKENDS

### LINKED ISSUE #2828: contributor-relay: HEADLESS_BACKENDS claims goose has no non-interactive entry point — goose run does
## Summary

`bin/contributor-relay.sh`'s `HEADLESS_BACKENDS` table (added in #2660 for #2538) and its surrounding comment state that `goose` has "no known non-interactive entry point" because it "drive[s] an interactive TUI":

> Backends NOT listed here have no known non-interactive entry point (goose / bob / agy / pi drive an interactive TUI), so headless mode refuses them LOUDLY at task time rather than silently stalling.

That claim is accurate for `bob`/`agy`/`pi`, but **it is not accurate for goose**. `contributor-relay.test.js` also pins the claim in a test (`buildHeadlessArgv('x')` asserted to return `null` for `goose`), so it's load-bearing, not just a comment.

I checked whether this had already been corrected on current `v2` HEAD (the pin I read, `98781c25`, is 22 commits behind) — it has not; `bin/contributor-relay.sh` is byte-identical between the pinned commit and `v2` HEAD as of this report.

## Evidence: `goose run` is a documented, working one-shot mode

Verified directly by executing the `goose` binary (`goose-x86_64-unknown-linux-musl`, version `1.45.0-canary+eea5609`, and independently re-verified against a `1.45.0` build from linuxbrew for this report):

```
$ goose run --help
Execute commands from an instruction file or stdin

Usage: goose run [OPTIONS]

Options:
  -i, --instructions <FILE>   Path to instruction file containing commands. Use - for stdin.
  -t, --text <TEXT>           Input text containing commands for goose. Use this in lieu of the instructions argument.
      --no-session            Execute commands without creating or using a session file. Useful for automated runs.
  -q, --quiet                 Quiet mode. Suppress non-response output, printing only the model response to stdout
      --output-format <FORMAT>  Output format (text, json, stream-json) [default: text] [possible values: text, json, stream-json]
      --max-turns <NUMBER>    Set a limit on how many turns (iterations) the agent can take without asking for user input to continue.
  ...
```

And actually running it, twice, to confirm the exit-code contract headless mode relies on (`runHeadlessTask()` completes a task on process exit code, not by scraping output):

```
$ goose run --no-session -t "print exactly the text: HELLO_TEST_OK"
...
goose is ready
HELLO_TEST_OK
$ echo $?
0
```

```
$ goose run --no-session -i /path/to/missing-file.txt >/dev/null 2>&1; echo $?
1
```

`--output-format json` / `stream-json` also work and return structured, parseable output rather than a TUI to scrape (tested with `--output-format json`, confirmed well-formed JSON with a `metadata.status` field).

Corroborating, non-authoritative evidence that this is stable and expected rather than an accident of one version: the downstream `projectbluefin/review` image's `Containerfile` has smoke-tested `goose run --help` at image build time for some time.

## Shape difference from the existing table entries

Worth flagging precisely, since `buildHeadlessArgv()`'s doc comment currently says the prompt is "the single positional argument that follows" the one-shot flag, and mentions a `promptAsArg` notion that (as far as I can find) has no actual implementation in the current code — `buildHeadlessArgv` only ever does `[...permArgs, ...modelArgs, spec.flag, prompt]`, a single flag token.

- `claude -p "<prompt>"`, `copilot -p "<prompt>"` — one flag, prompt follows as a bare positional.
- `codex exec "<prompt>"` — one subcommand token, prompt follows as a bare positional.
- `goose run --no-session -t "<prompt>"` — goose needs its `run` sub-command **and** a flag (`-t`) whose *value* is the prompt, i.e. two tokens ahead of the prompt, not one. The existing `spec.flag` shape (a single string) can't express that without either mis-joining tokens into one argv element or losing the `run` subcommand.

So this is not a pure data-table addition under the current code — `buildHeadlessArgv` needs a small extension to accept more than one leading token.

## Impact

Goose is currently the one interactive-only backend most likely to want headless status: the same downstream that opened #2538 (the headless-relay prerequisite for Kubernetes) runs goose as its primary contributor backend. Today, every goose contributor is forced onto the interactive `checkTmuxIdle()` heuristic — regex-scraping a tmux pane for `/goose is ready|> Enter to send|>\s*$|goose>|G\s*>/` plus a "not working" keyword check — to decide a task is done. Headless mode's exit-code completion in `runHeadlessTask()` is a fact, not a heuristic, and is the only completion signal that makes sense for #2668's Kubernetes contributor workload, where there is no human to attach to a pane at all.

## Options (not prescribing — happy to let a maintainer choose)

1. Extend `HEADLESS_BACKENDS` so an entry's `flag` can be an array of leading tokens (e.g. `goose: { flag: ['run', '--no-session', '-t'] }`), and have `buildHeadlessArgv` spread it ahead of the prompt instead of inserting a single token. I have this working locally (table entry + a ~10-line `buildHeadlessArgv` change + updated/added tests, all passing) and can open a PR if that shape looks right to a maintainer.
2. If there's a reason to keep `buildHeadlessArgv` to a single leading token by design, a maintainer may prefer a different seam (e.g. resolving `goose run` as part of the binary name) — I don't have visibility into constraints that might rule out option 1.
3. Separately, `--output-format stream-json` could let a future PR give the caller structured progress rather than a raw output tail — flagging as a distinct, unimplemented follow-on idea, not part of this report.

## What I did / did not verify

- Verified: `goose run --help` flags as documented above, exit-code behavior on both a success and a failure path, and `--output-format json` producing valid JSON — all via direct local execution of the `goose` binary.
- Verified: pinned-commit vs. `v2` HEAD diff of `bin/contributor-relay.sh` — no change, claim still present.
- Not verified: a live headless Hive task cycle end-to-end (needs a hub endpoint, registration token, and provider credentials I don't have set up) — my confirmation is limited to the CLI invocation and its exit-code contract, and to the (offline) relay unit tests.

I'll open a PR with option 1 shortly, since that's the smallest change I found that fits the existing style — please treat it as a proposal, not a fait accompli, given the shape-mismatch caveat above.

castrojo:
Fixes #2828.

## What

`HEADLESS_BACKENDS`' comment claimed goose has "no known non-interactive entry point" because it "drive[s] an interactive TUI", and `buildHeadlessArgv('x')` was asserted (by a passing test) to return `null` for goose. That claim is stale: `goose run --no-session -t "<prompt>"` is a documented one-shot invocation that exits 0 on success and non-zero on error. Full evidence, the exact flags/version tested, and the impact are in #2828 — this PR is the fix that issue proposes as option 1.

## The shape difference this required

The existing `HEADLESS_BACKENDS` entries (`claude -p`, `codex exec`) each need exactly **one** leading token before the prompt, and `buildHeadlessArgv` hard-coded that: `[...permArgs, ...modelArgs, spec.flag, prompt]`. goose needs **two**: its `run` sub-command *and* a flag (`-t`) whose value is the prompt. So `spec.flag` now accepts either a string (unchanged behavior for the four existing entries) or an array of leading tokens:

```js
goose: { flag: ['run', '--no-session', '-t'] },
```

```js
const oneShotArgs = Array.isArray(spec.flag) ? spec.flag : [spec.flag];
const args = [...permArgs, ...modelArgs, ...oneShotArgs, prompt];
```

I also dropped the doc comment's reference to a `promptAsArg` notion — I couldn't find any implementation of it in the current code (`buildHeadlessArgv` only ever appends the prompt as the trailing element), so it looked like stale documentation rather than an existing seam I should be using. Happy to be corrected if I missed where it lives.

## Tests

Extends `bin/contributor-relay.test.js`:
- New: `goose runs headless via its one-shot 'run' sub-command and reports completion` — full dispatch: `execFile` called with `goose`/`run`/`-t`/prompt, and a default (exit-0) result reports `task_complete`.
- Extended: `buildHeadlessArgv maps each supported backend to its one-shot invocation` — asserts goose's exact trailing argv (`['run', '--no-session', '-t', prompt]`) and that it does NOT get `--model` (goose is already in `NO_MODEL_FLAG_BACKENDS`).
- Changed: the two tests that used goose as the "backend with no one-shot mode" example (`headless refuses an unsupported backend loudly instead of stalling`, and half of `buildHeadlessArgv maps each supported backend...`) now use `bob`, which still has no known one-shot mode, so that coverage isn't lost.

`node bin/contributor-relay.test.js` → **27/27 passed** (26 existing + 1 new).

## What I verified vs. did not

- **Verified directly**, against the `goose` binary (`1.45.0-canary+eea5609`, and re-confirmed against a local `1.45.0` build for this PR): `goose run --help`'s documented flags, `goose run --no-session -t "<prompt>"` printing the response and exiting 0, and `goose run --no-session -i <missing-file>` exiting 1.
- **Verified**: the full relay unit test suite offline (no tmux, no real CLI, no network — `child_process`/`ws` are stubbed, per the existing harness in `contributor-relay.test.js`).
- **Not verified**: an end-to-end live headless Hive task cycle with a real goose backend. That needs a running hub, a registration token, and `GOOSE_PROVIDER`/model credentials I don't have set up here. If a maintainer can run one, the thing worth confirming is that a real task prompt (as opposed to the test harness's canned prompt) round-trips through `goose run --no-session -t` correctly and that `runHeadlessTask()`'s exit-code handling behaves the same against the real binary as it does against the stub.

## Not in scope here

`--output-format stream-json` for structured progress reporting instead of a raw output tail is mentioned as a possible future idea in #2828 — not implemented in this PR, to keep this change to the table addition + the minimal `buildHeadlessArgv` extension it requires.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -238,14 +238,16 @@ function buildLaunchCommand() {
 // status — the property the headless mode needs. Each entry says how to turn
 // (binary, perm-flags, prompt) into an argv:
 //
-//   flag — the sub-command/flag that selects one-shot mode; the prompt is
-//          passed as the single positional argument that follows it
-//          (`claude -p "<prompt>"`, `codex exec "<prompt>"`).
+//   flag — the sub-command/flag(s) that select one-shot mode, as a single
+//          token (`claude -p "<prompt>"`, `codex exec "<prompt>"`) or an
+//          array of tokens when a subcommand and a flag both precede the
+//          prompt (`goose run --no-session -t "<prompt>"`). The prompt is
+//          always appended as the final, distinct argv element.
 //
-// Backends NOT listed here have no known non-interactive entry point (goose /
-// bob / agy / pi drive an interactive TUI), so headless mode refuses them
-// LOUDLY at task time rather than silently stalling. Extending this table is
-// how a future PR adds a backend once its headless invocation is verified.
+// Backends NOT listed here have no known non-interactive entry point (bob /
+// agy / pi drive an interactive TUI), so headless mode refuses them LOUDLY at
+// task time rather than silently stalling. Extending this table is how a
+// future PR adds a backend once its headless invocation is verified.
 const HEADLESS_BACKENDS = {
   // claude -p "<prompt>" — print mode: runs the prompt non-interactively and
   // exits. Same perm flags as the interactive launch (bypass permissions).
@@ -257,6 +259,16 @@ const HEADLESS_BACKENDS = {
   copilot: { flag: '-p' },
   // codex exec "<prompt>" — Codex's non-interactive execution sub-command.
   codex: { flag: 'exec' },
+  // goose run --no-session -t "<prompt>" — `run` is goose's one-shot
+  // sub-command (the bare `goose` binary alone drives the interactive TUI);
+  // --no-session skips creating/resuming a session file, which headless
+  // dispatch never needs; -t takes the prompt as its value rather than as a
+  // trailing positional. Exits 0 on success, non-zero on error. Verified
+  // directly against `goose 1.45.0-canary+eea5609`: `goose run --no-session
+  // -t "<prompt>"` printed the response and exited 0; `goose run --no-session
+  // -i <missing-file>` exited 1. Not yet exercised through a live headless
+  // Hive task cycle — see kubestellar/hive#2828.
+  goose: { flag: ['run', '--no-session', '-t'] },
 };
 
 // headlessSupportsBackend reports whether the configured backend has a known
@@ -267,17 +279,18 @@ function headlessSupportsBackend() {
 
 // buildHeadlessArgv turns a task prompt into the argv for a one-shot,
 // non-interactive backend invocation: [binary, ...permFlags, ...modelFlag,
-// oneShotFlag, prompt] (order adjusted per promptAsArg). Returns null for an
-// unsupported backend. Never shell-interpolates the prompt — it is passed as a
-// distinct argv element to execFile, so apostrophes/quotes in the prompt (the
-// exact #2203 wedge on the interactive path) cannot break anything here.
+// ...oneShotFlags, prompt]. Returns null for an unsupported backend. Never
+// shell-interpolates the prompt — it is passed as a distinct argv element to
+// execFile, so apostrophes/quotes in the prompt (the exact #2203 wedge on the
+// interactive path) cannot break anything here.
 function buildHeadlessArgv(prompt) {
   const spec = HEADLESS_BACKENDS[BACKEND];
   if (!spec) return null;
   const { cmd, perm } = resolveBackend();
   const permArgs = perm ? perm.split(/\s+/).filter(Boolean) : [];
   const modelArgs = MODEL && !NO_MODEL_FLAG_BACKENDS.includes(BACKEND) ? ['--model', MODEL] : [];
-  const args = [...permArgs, ...modelArgs, spec.flag, prompt];
+  const oneShotArgs = Array.isArray(spec.flag) ? spec.flag : [spec.flag];
+  const args = [...permArgs, ...modelArgs, ...oneShotArgs, prompt];
   return { bin: cmd, args };
 }
 
--- a/bin/contributor-relay.test.js
+++ b/bin/contributor-relay.test.js
@@ -544,6 +544,26 @@ test('interactive is the default mode; headless is opt-in via CONTRIBUTOR_MODE',
   } finally { teardown(head); }
 });
 
+test('goose runs headless via its one-shot `run` sub-command and reports completion', () => {
+  const relay = loadRelay({ backend: 'goose', mode: 'headless' });
+  try {
+    assignHeadlessTask(relay);
+
+    assert.strictEqual(relay.__execFileCalls.length, 1, 'expected exactly one one-shot invocation');
+    const call = relay.__execFileCalls[0];
+    assert.strictEqual(call.bin, 'goose', 'goose backend should run the goose binary');
+    assert.ok(call.args.includes('run'), `goose headless must use its 'run' sub-command: ${JSON.stringify(call.args)}`);
+    assert.ok(call.args.includes('-t'), `goose headless must pass the prompt via -t: ${JSON.stringify(call.args)}`);
+    assert.ok(call.args[call.args.length - 1].includes('foo/bar#7'),
+      'the prompt must be the trailing argv element (passed to execFile, never shell-quoted)');
+
+    // A default (exit-0, empty stdout) execFile result reports completion —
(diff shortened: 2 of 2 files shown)


clubanderson:
Thanks for diagnosing this one — the claim checks out. I verified `goose run` against **1.37.0** (the version `v2/Dockerfile` pins via `ARG GOOSE_VERSION=1.37.0`, vs the 1.45.0 you tested), and the flags are present there too: `-t/--text`, `-i/--instructions`, `--no-session`, `--output-format`, `--max-turns`. `goose run --no-session -i /nonexistent.txt` exits 1, which is the exit-code contract `runHeadlessTask()` relies on.

While confirming, I found the capability set is mirrored in a few more places than the issue named — flagging them so this PR can land complete rather than needing a follow-up:

1. `Justfile` — `HEADLESS_BACKENDS="…"` (drives the preflight warning)
2. `v2/pkg/dashboard/api_contribute.go` — the `K8S_HEADLESS_BACKENDS` map **and** the user-facing copy ("Pick Claude Code, LiteLLM, Copilot, Codex or Goose"). Without this one, a goose contributor still sees "goose has no headless mode; it will refuse work in a cluster" in the onboarding UI even though the relay runs it fine. Both files carry comments saying they must match `HEADLESS_BACKENDS`.

One regression to watch: `v2/deploy/test_contribute_k8s_workload.sh` uses `AGENT_BACKEND=goose` as its "unsupported backend warns on stderr" fixture, so adding goose silently breaks that test. Swapping the fixture to `bob` preserves the coverage.

Two smaller notes: goose is the first backend needing two leading tokens before the prompt (`run --no-session -t <prompt>`, where the prompt is `-t`'s value) rather than one like `claude -p` / `codex exec` — so `spec.flag` needs to accept an array. And `promptAsArg` turns out to be a phantom: one doc-comment mention, no implementation.

I have all of the above working in #2831, which I opened before spotting this PR. Happy to close mine in favour of yours — say the word and I'll push these bits to your branch, or you can pull them across, whichever you prefer. The diagnosis and the fix are yours.

clubanderson:
Thanks for the contribution, Jorge — your core fix is right: Goose does have a verified non-interactive path via `goose run --no-session -t <prompt>`, and the relay needs to treat that as headless-capable.

I landed the overlapping maintainer PR #2831 instead because it included the same relay argv change plus the additional surfaces needed on current `v2`: `Justfile` `contribute-k8s`, the /contribute page's `K8S_HEADLESS_BACKENDS` map, the watsonx/headless pinning tests, and the Kubernetes workload smoke test. Without those extra updates, Kubernetes-mode contributors would still see Goose warned/filtered as non-headless even though the relay supports it.

I appreciate you catching the stale assumption in the relay and opening this — it directly informed the merged fix in #2831.

clubanderson:
Closing as superseded by #2831, which has now merged with the same Goose headless relay support plus the Kubernetes contribute UI/Justfile/test updates needed on current v2. Thanks again!

PR #2830, opened 2026-08-08
🐛 fix(gateway): healthcheck no longer probes IPv6 localhost against an IPv4-only nginx

### LINKED ISSUE #2774: hive-gateway healthcheck permanently reports unhealthy: probes localhost (IPv6) but nginx only binds IPv4
## Summary

`hive-gateway` runs correctly but its container healthcheck **never** succeeds, so the container sits in `unhealthy` forever. On a host running since the last reboot this had accumulated a `FailingStreak` of **10390** consecutive failures while the gateway was serving traffic normally the entire time.

This is a healthcheck defect, not a gateway defect — but it makes `podman ps` permanently misreport the service, and it will defeat any orchestration or monitoring that keys off container health.

## Root cause

`v2/quadlets/hive-gateway.container` line 14:

```
HealthCmd=/usr/bin/wget -qO- http://localhost:3001/api/health
```

`v2/deploy/nginx.conf` binds IPv4 only:

```nginx
server {
    listen 3001;
```

Inside the container `/etc/hosts` maps `localhost` to **both** families:

```
127.0.0.1	localhost
::1	localhost
```

BusyBox `wget` resolves `localhost` to `::1` first. Nothing is listening on `::1:3001`, so the probe gets `ECONNREFUSED` every single time.

## Reproduction / evidence

Confirmed on podman 5.x, rootless, `docker.io/library/nginx:alpine`, gateway up 32 hours:

```console
$ podman exec hive-gateway netstat -tlnp
tcp  0  0 0.0.0.0:3001  0.0.0.0:*  LISTEN  1/nginx: master pro
tcp  0  0 0.0.0.0:7681  0.0.0.0:*  LISTEN  1/nginx: master pro

$ podman exec hive-gateway wget -qO- http://127.0.0.1:3001/api/health
{"status":"ok"}

$ podman exec hive-gateway wget -qO- 'http://[::1]:3001/api/health'
wget: can't connect to remote host: Connection refused

$ podman exec hive-gateway wget -qO- http://localhost:3001/api/health
wget: can't connect to remote host: Connection refused
```

The service is genuinely healthy from the host through the published port:

```console
$ curl -s http://localhost:3001/api/health
{"status":"ok"}
```

And the recorded health state:

```console
$ podman inspect hive-gateway --format '{{json .State.Health}}'
{"Status":"unhealthy","FailingStreak":10390,"Log":[{...,"ExitCode":1,
 "Output":"wget: can't connect to remote host: Connection refused\n"}, ...]}
```

Every entry in the health log is the same `Connection refused`, back to container start — it has never once passed.

## Why the upstream nginx image doesn't save you here

`nginx:alpine` ships `/docker-entrypoint.d/10-listen-on-ipv6-by-default.sh`, which normally adds `listen [::]:80;`. The container log confirms it ran:

```
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
```

But it only patches `/etc/nginx/conf.d/default.conf`. The quadlet bind-mounts `deploy/nginx.conf` over `/etc/nginx/nginx.conf`, and that config has no `include /etc/nginx/conf.d/*.conf`, so `default.conf` is never loaded and the IPv6 fixup has no effect on the actual server block. The script's success message is misleading here.

## Suggested fixes

Either one resolves it; (1) is the minimal, most obviously correct change.

**1. Probe IPv4 explicitly** — `v2/quadlets/hive-gateway.container`:

```diff
-HealthCmd=/usr/bin/wget -qO- http://localhost:3001/api/health
+HealthCmd=/usr/bin/wget -qO- http://127.0.0.1:3001/api/health
```

**2. Or make nginx dual-stack** — `v2/deploy/nginx.conf`:

```diff
 server {
     listen 3001;
+    listen [::]:3001;
```

Note the `stream` block for ttyd on 7681 has the same IPv4-only bind, so if anything is later added that probes `localhost:7681` it will hit this same trap. Fixing the listen directives (2) inoculates against that; fixing only the probe (1) does not.

Happy to open a PR with DCO sign-off for whichever you prefer.

---
Reported from a downstream `projectbluefin/review` deployment consuming the v2 quadlets unmodified — our local `hive-gateway.container` is byte-identical to `v2/quadlets/hive-gateway.container`.

clubanderson:
## The report

[#2774](https://github.com/kubestellar/hive/issues/2774), from `castrojo`, running the v2 quadlets unmodified downstream. The gateway served traffic fine for 32 hours while its healthcheck accumulated a `FailingStreak` of **10390** — it had never once passed. His diagnosis:

> `v2/deploy/nginx.conf` binds IPv4 only [...] Inside the container `/etc/hosts` maps `localhost` to **both** families [...] BusyBox `wget` resolves `localhost` to `::1` first. Nothing is listening on `::1:3001`, so the probe gets `ECONNREFUSED` every single time.

And the part that matters most, which is easy to miss:

> `nginx:alpine` ships `/docker-entrypoint.d/10-listen-on-ipv6-by-default.sh` [...] But it only patches `/etc/nginx/conf.d/default.conf`. The quadlet bind-mounts `deploy/nginx.conf` over `/etc/nginx/nginx.conf`, and that config has no `include /etc/nginx/conf.d/*.conf`, so `default.conf` is never loaded and the IPv6 fixup has no effect on the actual server block. The script's success message is misleading here.

## What I confirmed

Verified against the tree rather than taken on faith:

- `v2/deploy/nginx.conf` binds IPv4 only — `listen 3001;` (http) and `listen 7681;` (ttyd stream). No IPv6 listen anywhere in the file.
- The file has **no** `include /etc/nginx/conf.d/*.conf`, so the upstream IPv6 fixup genuinely cannot reach these server blocks. The reporter's subtlest claim holds.
- `v2/docker-compose.yaml` mounts it at `./deploy/nginx.conf:/etc/nginx/nginx.conf:ro`, confirming the override path.

One correction to scope: `v2/quadlets/hive-gateway.container` **does not exist in this repo** — that path is the reporter's downstream deployment. The `HealthCmd=` line he quotes is not ours to change. What is ours is the nginx config, and it is still IPv4-only.

## Why fix (b), not (a)

[#2779](https://github.com/kubestellar/hive/pull/2779) already applied fix (a) — `localhost` → `127.0.0.1` across `docker-compose.yaml`, `docker-compose.architect.yaml`, `blue-green-deploy.sh` and `bin/hive-setup.sh` — and closed #2774. But it never touched `nginx.conf`. So the probes were made unambiguous while the underlying serving gap was left in place: **the gateway still refuses all IPv6 connections.**

That makes (a) alone exactly the papering-over the issue warns about. Fix (a) silences the symptom for probes that we control; it does nothing for a real IPv6 client. This PR adds the missing half:

```nginx
listen 3001;
listen [::]:3001;
```

The reporter also flagged that the ttyd `stream` block on 7681 has the same IPv4-only bind and is the same trap waiting for the next probe added against it, so that block is fixed too. As he put it: fixing the listen directives "inoculates against that; fixing only the probe (1) does not."

Combined with #2779 already in place, the gateway is now dual-stack *and* the probes are unambiguous — both halves, which is what the issue actually asked for.

## Tests

`v2/pkg/config/gateway_nginx_dualstack_test.go`, following the existing `entrypoint_boot_test.go` pattern of asserting against the real deploy file via a relative path (not a paraphrase of it).

- `TestGatewayListensDualStack` — asserts both the IPv4 and the `[::]` listen for 3001 and 7681. Verified it genuinely pins the behavior: with the `listen [::]` lines stripped it fails on both ports; with them it passes.
- `TestGatewayConfDoesNotRelyOnConfD` — guards the *reason* the upstream fixup doesn't save us. If someone later adds an `include /etc/nginx/conf.d/*.conf`, it flags that the rationale needs revisiting.

## Validation

- `go build ./...` — clean
- `go vet ./pkg/config/` — clean
- `go test ./pkg/config/ -run TestGateway` — pass
- `nginx -t` not run locally (no Docker daemon available); leaving config syntax to CI.

Fixes #2774

### DIFF
--- a/v2/deploy/nginx.conf
+++ b/v2/deploy/nginx.conf
@@ -25,6 +25,13 @@ http {
 
     server {
         listen 3001;
+        # Also bind the IPv6 loopback/wildcard. This file is mounted over
+        # /etc/nginx/nginx.conf and has no `include /etc/nginx/conf.d/*.conf`,
+        # so the nginx:alpine 10-listen-on-ipv6-by-default.sh fixup (which only
+        # patches conf.d/default.conf) never reaches this server block. Without
+        # this line, anything resolving "localhost" to ::1 -- container
+        # healthchecks, IPv6 clients -- gets ECONNREFUSED. See issue #2774.
+        listen [::]:3001;
 
         location /api/ {
             proxy_pass http://hive_api;
@@ -71,6 +78,10 @@ stream {
 
     server {
         listen 7681;
+        # Same dual-stack reasoning as the :3001 server block above. The ttyd
+        # stream proxy is IPv4-only today, so any future probe or client that
+        # resolves "localhost" to ::1 would hit the identical trap.
+        listen [::]:7681;
         proxy_pass hive_ttyd;
     }
 }
--- a/v2/pkg/config/gateway_nginx_dualstack_test.go
+++ b/v2/pkg/config/gateway_nginx_dualstack_test.go
@@ -1,0 +1,72 @@
+package config
+
+import (
+	"os"
+	"regexp"
+	"testing"
+)
+
+// nginxConfPath is the gateway config under test, relative to this package.
+// It is bind-mounted over /etc/nginx/nginx.conf by v2/docker-compose.yaml.
+const nginxConfPath = "../../deploy/nginx.conf"
+
+// Ports that the gateway server blocks listen on. Named rather than inlined so
+// a port change forces a deliberate edit here instead of silently un-covering
+// a block.
+const (
+	gatewayAPIPort  = "3001"
+	gatewayTTYDPort = "7681"
+)
+
+// readNginxConf returns the gateway nginx config text.
+func readNginxConf(t *testing.T) string {
+	t.Helper()
+	src, err := os.ReadFile(nginxConfPath)
+	if err != nil {
+		t.Skipf("nginx.conf not readable from this package: %v", err)
+	}
+	return string(src)
+}
+
+// TestGatewayListensDualStack pins the fix for issue #2774.
+//
+// The gateway healthcheck (and any IPv6 client) resolves "localhost" to ::1 on
+// a dual-stack host. This file is mounted over /etc/nginx/nginx.conf and does
+// NOT include /etc/nginx/conf.d/*.conf, so the nginx:alpine
+// 10-listen-on-ipv6-by-default.sh fixup never applies to these server blocks.
+// If the `listen [::]:PORT;` directives are dropped, nothing binds the IPv6
+// loopback and the probe gets ECONNREFUSED forever.
+func TestGatewayListensDualStack(t *testing.T) {
+	conf := readNginxConf(t)
+
+	for _, port := range []string{gatewayAPIPort, gatewayTTYDPort} {
+		t.Run("port_"+port, func(t *testing.T) {
+			// IPv4 bind must remain.
+			v4 := regexp.MustCompile(`(?m)^\s*listen\s+` + port + `\s*;`)
+			if !v4.MatchString(conf) {
+				t.Errorf("nginx.conf lost the IPv4 `listen %s;` directive", port)
+			}
+			// IPv6 bind must be present alongside it.
+			v6 := regexp.MustCompile(`(?m)^\s*listen\s+\[::\]:` + port + `\s*;`)
+			if !v6.MatchString(conf) {
+				t.Errorf("nginx.conf is missing `listen [::]:%s;` -- IPv6 loopback "+
+					"is unbound, so healthchecks resolving localhost to ::1 will "+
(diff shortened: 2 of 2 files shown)


clubanderson:
Reviewed against #2779 / merge commit 610f5937: that prior fix changed the healthcheck/probe callers to use 127.0.0.1 in bin/hive-setup.sh, v2/deploy/blue-green-deploy.sh, v2/deploy/docker-compose.architect.yaml, and v2/docker-compose.yaml. This PR is not a duplicate of that change; it adds new gateway coverage by making v2/deploy/nginx.conf bind both :3001 and :7681 on IPv6 as well as IPv4, with a regression test pinning the dual-stack listens. Local validation passed: `cd v2 && go build ./...` and `go test ./pkg/config/ -count=1`.

PR #2835, opened 2026-08-08
🐛 fix(contributor): entrypoint.d hooks can override backend_binary

### LINKED ISSUE #2833: contributor: entrypoint.d hooks cannot override backend_binary because backends.conf is sourced after them
## What

`bin/contributor-agent.sh` sources downstream hooks **before** `config/backends.conf`, and `backends.conf` defines `backend_binary()` / `backend_perm_flag()` as bash functions. Because a later `source` of a function definition silently replaces an earlier one, a hook that overrides either function is discarded without warning.

This is at odds with the stated intent of the hook seam. The comment above it says (emphasis mine):

> an image (e.g. `projectbluefin/donate-clanker`) can drop `*.sh` into `/etc/hive/entrypoint.d/` and/or set `HIVE_PRE_AGENT_HOOK` to run setup here — after the full contributor env is exported, before backend detection and the tmux launch — **WITHOUT forking this entrypoint** ...

Overriding which binary a backend resolves to is exactly the kind of setup that seam reads as inviting, but it is the one thing the ordering prevents.

## Where

Hooks and `HIVE_PRE_AGENT_HOOK` at [`bin/contributor-agent.sh#L58-L69`](https://github.com/kubestellar/hive/blob/98781c252cefb2f2193832a701abd8d0728ea18b/bin/contributor-agent.sh#L58-L69), then `backends.conf` at [`#L71-L77`](https://github.com/kubestellar/hive/blob/98781c252cefb2f2193832a701abd8d0728ea18b/bin/contributor-agent.sh#L71-L77).

Present at pinned commit `98781c252cefb2f2193832a701abd8d0728ea18b` and unchanged on current `v2` HEAD (verified by fetching both).

## Reproduction

Self-contained, no Hive checkout needed:

```bash
mkdir -p /tmp/ord/hooks && cd /tmp/ord
cat > hooks/10-override.sh <<'HOOK'
backend_binary() { echo "my-custom-goose-wrapper"; }
HOOK
cat > backends.conf <<'CONF'
backend_binary() { case "$1" in goose) echo "goose" ;; *) echo "$1" ;; esac; }
CONF

# order used by contributor-agent.sh today
for h in hooks/*.sh; do source "$h"; done
source backends.conf
echo "current order -> $(backend_binary goose)"

# reversed
unset -f backend_binary
source backends.conf
for h in hooks/*.sh; do source "$h"; done
echo "hooks after   -> $(backend_binary goose)"
```

```
current order -> goose                      # hook silently ignored
hooks after   -> my-custom-goose-wrapper    # hook wins
```

## Why it matters

The workaround available to a downstream image today is to ship a replacement `config/backends.conf` (or `/usr/local/etc/hive/backends.conf`) baked into the image. That means carrying a divergent copy of an upstream file that is otherwise fetched verbatim at a pinned commit — so it drifts silently whenever upstream edits the real one, which is precisely the forking the hook comment says downstreams should not have to do.

Concretely, in `projectbluefin/review` we consume `bin/contributor-agent.sh`, `bin/contributor-relay.sh` and `config/backends.conf` unmodified at a pinned commit, and assert that in tests. Adding a backend variant currently has no seam that does not involve overwriting one of those three.

## Options (upstream's call — not asking for a specific one)

1. **Source `backends.conf` first, hooks second.** Smallest change and makes the seam behave as documented. Risk: a hook that *relies* on running before backend detection could be affected; from a read of the current code the hooks only export env and do setup, but you would know better whether anyone depends on the current order.
2. **Source `backends.conf` first, then hooks, then re-run detection.** Same as 1 but explicit about where `detect_cli` sits relative to the override.
3. **Add a second, later hook directory** (e.g. `/etc/hive/backends.d/*.sh`) sourced immediately after `backends.conf`, leaving the existing seam's timing untouched. Most conservative, at the cost of a second concept.
4. **Make the functions overridable by convention** — e.g. have `backends.conf` skip defining a function that is already defined, or consult an env var such as `HIVE_BACKEND_BINARY_<NAME>`.

Option 3 is the least disruptive if the current hook timing is load-bearing for someone; option 1 is the most honest to the comment's stated intent.

## What I verified vs. did not

- **Verified:** the ordering at the pinned commit and at `v2` HEAD; the clobbering behaviour, via the self-contained reproduction above.
- **Not verified:** that no existing consumer depends on hooks running before `backends.conf`. I can only see this repository and one downstream, so the blast radius of option 1 is genuinely your call rather than mine.

Happy to send a PR for whichever option you prefer.

clubanderson:
Fixes #2833

## The report

@castrojo (thanks — the writeup and self-contained repro made this a five-minute diagnosis) reported that `bin/contributor-agent.sh` sources the downstream hook seam **before** `config/backends.conf`:

> `backends.conf` defines `backend_binary()` / `backend_perm_flag()` as bash functions. Because a later `source` of a function definition silently replaces an earlier one, a hook that overrides either function is discarded without warning.

And that this contradicts the seam's own comment, which advertises it as the place to run setup "before backend detection ... WITHOUT forking this entrypoint":

> Overriding which binary a backend resolves to is exactly the kind of setup that seam reads as inviting, but it is the one thing the ordering prevents.

## What I verified

- **The ordering is real and is as reported.** On `v2` HEAD the `entrypoint.d` loop and `HIVE_PRE_AGENT_HOOK` were at `bin/contributor-agent.sh:58-69`, and the `backends.conf` sourcing at `:71-77` — hooks first, config second.
- **These are functions, not variables.** Worth stating explicitly because it rules out a fix that would otherwise be the obvious one: there is no `backend_binary=...` assignment anywhere, so the `: "${backend_binary:=...}"` default-if-unset trick does not apply. Clobbering is at function-definition granularity, so the *only* lever is which side is sourced last.
- **Nothing in `backends.conf` reads hook-set state.** It defines four functions plus `KNOWN_BACKENDS` and consults no environment variable, so sourcing it earlier cannot change what it produces.
- **The seam is unique to this script.** Seven other scripts source `backends.conf` (`hive.sh`, `agent-launch.sh`, `supervisor.sh`, `kick-agents.sh`, `kick-governor.sh`, the `Justfile`, and the relay's own JS parser), but `/etc/hive/entrypoint.d` and `HIVE_PRE_AGENT_HOOK` appear **only** in `contributor-agent.sh`. The blast radius of reordering is one file.
- **No in-repo hook depends on the old order** — there are no `entrypoint.d` files in-tree at all; the directory is purely a downstream drop-point.

## The fix — option 1

I took the reporter's **option 1** (source `backends.conf` first, hooks second), which he flagged as "the most honest to the comment's stated intent".

Option 4 (`declare -F ... || backend_binary() {...}`, i.e. skip-if-defined) was the main alternative, and I rejected it: it makes the upstream table's presence conditional on what a hook did, so a hook that overrides one backend would silently suppress the definitions for *all* the others. Option 3 (a second `backends.d/` directory) buys order-independence at the cost of a second concept, which is not worth it given the seam has exactly one consumer today.

The stated risk of option 1 — a hook relying on running *before* backend detection — is unaffected: hooks still run after the full contributor env is exported and still before `detect_cli` and the tmux launch. The only thing that moved is a block of pure function definitions. Hook semantics are otherwise unchanged.

## Tests

New `v2/deploy/test_contributor_hook_override.sh`, in the style of the existing `test_entrypoint_runtime_config.sh`: it **extracts the real blocks from the shipped entrypoint** rather than grepping or copying them, and — importantly — assembles them **in the order the file actually has them**, so it cannot pass against a regressed entrypoint. Verified it fails 5/6 when pointed at the pre-fix code, with `backend_binary goose` returning the clobbered `goose` exactly as the issue predicted, and passes 6/6 after.

Cases: stock default with no hook; a hook retargeting a binary; a hook overriding the permission flags; a hook that *wraps* the upstream function (only possible if `backends.conf` loaded first); and `HIVE_PRE_AGENT_HOOK`.

Wired into `.github/workflows/v2-ci.yml` next to the existing entrypoint test.

## Collision check

Deliberately avoided the files in flight on #2829/#2831 — those touch `bin/contributor-relay.sh`, `contributor-relay.test.js`, and `test_contribute_k8s_workload.sh`. This PR touches `bin/contributor-agent.sh`, a new test file, and the CI workflow, so there is no overlap. `config/backends.conf` is unmodified, which also means downstreams pinning it verbatim keep doing so.

`go build ./...`, `go vet ./...`, both shell suites, and `bash -n` all clean locally; CI is the gate.

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -39,6 +39,14 @@ jobs:
       - name: Entrypoint runtime-config migration tests
         run: bash deploy/test_entrypoint_runtime_config.sh
 
+      # backends.conf defines the backend tables as bash functions, so whichever
+      # of {backends.conf, entrypoint.d hooks} is sourced last wins. Sourcing it
+      # after the hooks silently discarded a downstream image's override. This
+      # replays the real blocks in the order the entrypoint has them, so putting
+      # backends.conf back below the seam fails the PR.
+      - name: Contributor entrypoint.d hook override tests
+        run: bash deploy/test_contributor_hook_override.sh
+
       - uses: actions/setup-node@v4
         with:
           node-version: '22'
--- a/bin/contributor-agent.sh
+++ b/bin/contributor-agent.sh
@@ -49,12 +49,30 @@ export CONTRIBUTOR_MODE
 # Username is extracted from contributor.env (set during registration)
 export HIVE_CONTRIBUTOR_USERNAME="${CONTRIBUTOR_USERNAME:-unknown}"
 
+# Source backends.conf for binary detection.
+#
+# This MUST stay ABOVE the entrypoint.d hook seam below (kubestellar/hive#2833).
+# backends.conf defines backend_binary() / backend_perm_flag() as bash
+# FUNCTIONS, and sourcing a function definition silently replaces any earlier
+# one. Sourced after the hooks, it clobbered a hook's override with no warning,
+# so the one thing the seam most obviously invites — pointing a backend at a
+# different binary — was the one thing it could not do. Loading it first means
+# the hooks run last and win.
+BACKENDS_CONF="${SCRIPT_DIR}/../config/backends.conf"
+if [[ -f "$BACKENDS_CONF" ]]; then
+  source "$BACKENDS_CONF"
+elif [[ -f /usr/local/etc/hive/backends.conf ]]; then
+  source /usr/local/etc/hive/backends.conf
+fi
+
 # Extension seam for downstream images (kubestellar/hive#2393 item 4). A derived
 # image (e.g. projectbluefin/donate-clanker) can drop *.sh into
 # /etc/hive/entrypoint.d/ and/or set HIVE_PRE_AGENT_HOOK to run setup here —
-# after the full contributor env is exported, before backend detection and the
-# tmux launch — WITHOUT forking this entrypoint and re-implementing the tmux
-# wait/attach logic. Sourced (not exec'd) so hooks can export env the agent sees.
+# after the full contributor env is exported and after backends.conf has
+# defined the backend tables, but before backend detection and the tmux launch —
+# WITHOUT forking this entrypoint and re-implementing the tmux wait/attach
+# logic. Sourced (not exec'd) so hooks can export env the agent sees and can
+# redefine backend_binary()/backend_perm_flag() to add or retarget a backend.
 if [[ -d /etc/hive/entrypoint.d ]]; then
   for _hook in /etc/hive/entrypoint.d/*.sh; do
     [[ -r "$_hook" ]] || continue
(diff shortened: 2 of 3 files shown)


clubanderson:
Superseded by #2855 (merged), which fixes #2833 by sourcing backends.conf before the entrypoint.d hook seam so hook-provided backend_binary/backend_perm_flag overrides survive.

PR #2825, opened 2026-08-08
✨ feat(access): merger contributor tier — review + auto-merge queue, never own PRs

clubanderson:
## Summary
- Add the new `merger` contributor role between read-write and owner, with shared role validation/order helpers.
- Propagate merger through hub Manage Access, heartbeat/direct-route authorized users, and spoke read/write gates.
- Add a spoke-side PR queue endpoint that requires merger-or-owner access, rejects merger self-queue by session GitHub login, scopes the repo to this hive, approves the PR, labels it `hive/automerge`, and audits the action.
- Add dashboard PR-row queue controls with queued-state display, plus docs and tests.

## Validation
- go test ./pkg/hub/ ./pkg/dashboard/ ./pkg/config/ -count=1
- go build ./...
- go vet ./...

PR #2838, opened 2026-08-08
✨ feat(access): make the merger auto-merge label configurable, default lgtm

castrojo:
Follow-up to #2825.

## Why

`QueuePRAutoMerge` applies a hard-coded `hive/automerge` label. That label is applied in a repository the hive does **not** own, and those repositories usually already have a convention for "a second person signed off on this" — Prow/Kubernetes projects have used `lgtm` for exactly this decision for years. A fixed name either collides with the existing one or forces every managed repo to grow a second label that means the same thing.

The operator ask on #2825 was literally this:

> j0rge: "I was thinking maybe lgtm lol, if I see brian's PR I LGTM and back in the queue but trusted and above can only automerge and it can't be yours?"

## What

- New `governor.labels.automerge`, beside the existing `governor.labels.exempt`.
- Defaults to `config.DefaultAutoMergeLabel` = **`lgtm`**.
- Threaded through exactly like the exempt list: a nil-safe `SetAutoMergeLabel`/`AutoMergeLabel` pair on `github.Client`, applied at all five `SetExemptLabels` call sites in `cmd/hive` so a config reload re-applies it.
- The dashboard is **told** the label via `/api/role` rather than hard-coding it.
- The scheduler's merge-eligible line now reads `[queued for auto-merge]` instead of naming the label, since it renders for every hive regardless of config.

**On the default:** I set it to `lgtm` because that was the stated ask and because #2825 is hours old, so there is no installed base to migrate — this is the only moment when changing it is free. If you would rather keep `hive/automerge`, it is a one-line change to `DefaultAutoMergeLabel`. The configurability is the substance here; the default is a trivially-flippable detail and entirely your call.

## Two edge cases worth a look

- **Blank config keeps the default rather than clearing it.** `AddLabelsToIssue` with an empty name is a 422, which would surface as a queue action that *approves the PR and then fails* — leaving it approved but never queued, which is the worst of both.
- **`/api/role` is served before GitHub credentials are required**, so its label lookup is guarded against a nil `deps`/client. `handleQueuePRAutoMerge` already rejects that state explicitly; `handleRole` did not, and adding an unguarded `s.deps.GHClient` there would have panicked the dashboard on a hive that booted without GitHub creds.

## Tests

New in `pkg/github/automerge_label_test.go`: fallback-when-unset, honours-configuration, blank/whitespace rejection, whitespace trimming, nil receiver. Plus a nil-`deps` test for the dashboard helper.

Four existing tests asserted the literal `hive/automerge`; they now assert the constant or the rendered fallback, so a hive that changes the default does not have to change the tests too.

## Validation

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./pkg/github/ ./pkg/config/ ./pkg/dashboard/ ./pkg/scheduler/ -count=1` — all pass

**Not validated:** a live queue action against real GitHub. That needs hive App credentials I do not have, so the API call sequence (ensureLabel → CreateReview APPROVE → AddLabelsToIssue) is exercised only against the existing httptest fakes.

## One question, separate from this PR

While tracing this I could not find the consumer of the label. `QueuePRAutoMerge`'s comment says it marks the PR "for the hive's existing auto-merge-on-green sweep", but grepping `v2/` for `PullRequests.Merge`, `EnableAutoMerge` and similar finds nothing, and there is no automerge workflow in `.github/`. Is the sweep external to this repo (an app or a workflow in the managed repos), or still to come? Happy to file that separately — it affects what a downstream has to provide to actually use the tier, and it is not a blocker for this change either way.

### DIFF
--- a/v2/cmd/hive/automerge_label_test.go
+++ b/v2/cmd/hive/automerge_label_test.go
@@ -1,0 +1,25 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/kubestellar/hive/v2/pkg/github"
+)
+
+func TestNormalizedAutoMergeLabel(t *testing.T) {
+	for _, tt := range []struct {
+		name  string
+		label string
+		want  string
+	}{
+		{name: "configured", label: "ship-it", want: "ship-it"},
+		{name: "trims", label: "  ship-it  ", want: "ship-it"},
+		{name: "blank falls back", label: " \t\n ", want: github.AutoMergeQueuedLabel},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := normalizedAutoMergeLabel(tt.label); got != tt.want {
+				t.Fatalf("normalizedAutoMergeLabel(%q) = %q, want %q", tt.label, got, tt.want)
+			}
+		})
+	}
+}
--- a/v2/cmd/hive/main.go
+++ b/v2/cmd/hive/main.go
@@ -4092,7 +4092,7 @@ func runEvalCycle(
 		atomicWrite("/data/last-actionable.json", data)
 	}
 
-	writeMergeEligible(actionable, actionable.Hold, cfg.Project.Org, logger)
+	writeMergeEligible(actionable, actionable.Hold, cfg.Project.Org, cfg.Governor.Labels.AutoMerge, logger)
 
 	shaResult, shaErr := ghClient.EnforceSHAHold(ctx, github.SHAHoldConfig{
 		PrimaryRepo:     cfg.Project.PrimaryRepo,
@@ -4934,12 +4934,13 @@ func applyDuplicatePRGuard(
 	github.ApplyDuplicatePRGuard(ctx, ghClient, claimLedger, hiveIdentity(cfg), actionable, logger)
 }
 
-func writeMergeEligible(actionable *github.ActionableResult, hold github.HoldResult, org string, logger *slog.Logger) {
+func writeMergeEligible(actionable *github.ActionableResult, hold github.HoldResult, org, autoMergeLabel string, logger *slog.Logger) {
 	holdSet := make(map[string]bool)
 	for _, h := range hold.Items {
 		key := fmt.Sprintf("%s/%d", h.Repo, h.Number)
 		holdSet[key] = true
 	}
+	autoMergeLabel = normalizedAutoMergeLabel(autoMergeLabel)
 
 	type eligiblePR struct {
 		Number int      `json:"number"`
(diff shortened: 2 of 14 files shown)


clubanderson:
Thanks @castrojo — I reviewed this locally and the shape looks good overall: config defaults blank/missing automerge labels to lgtm, QueuePRAutoMerge uses the configured client label, /api/role only exposes non-secret config, the role/self-merge gates are unchanged, and the requested validation passed after a clean local rebase onto origin/v2:

`cd v2 && go build ./... && go vet ./... && go test ./pkg/config/ ./pkg/dashboard/ ./pkg/github/ ./pkg/hub/ -count=1`

One blocker before merge: `cmd/hive/main.go` still computes the merge-eligible `queued` flag with `github.AutoMergeQueuedLabel` in `writeMergeEligible`. That means a hive configured with `governor.labels.automerge: ship-it` will apply `ship-it` when the merger queues the PR, but the scheduler/merge-eligible payload will not mark that PR as queued; conversely a stale/default `lgtm` label would still be treated as queued even after changing the configured label. Since the scanner/automerge flow consumes that queued signal, this leaves the new knob only partially honored and can create the stale-label bypass/misclassification the change is trying to avoid.

Suggested fix: pass the configured automerge label from `cfg.Governor.Labels.AutoMerge` into `writeMergeEligible` and compare labels against that value (with the same default already applied by config), plus a small coverage test showing a custom automerge label is the one that sets `queued`. I did not merge for now.

castrojo:
Supporting evidence that turned up in this PR's own CI, which I did not have when I opened it.

The `tide` check on this pull request reports:

> **Not mergeable. Needs `approved`, `lgtm` labels.**

So this repository already runs Prow/Tide with an `OWNERS` file, and its merge gate is exactly the `lgtm` label. Prow's `lgtm` plugin also applies and removes that label automatically in response to `/lgtm` and a review being dismissed.

That is a stronger argument for the default than the convention argument I made in the description: for a Prow-managed repository — including this one — `lgtm` is not merely a familiar name, it is the label the existing merge automation already acts on. Applying `hive/automerge` there would mark a PR as queued while leaving Tide's actual gate untouched, so the queue action would look successful and nothing would merge.

Two caveats so this is not overstated:

- Tide needs **both** `approved` and `lgtm`. `QueuePRAutoMerge` creates a GitHub review approval, which is not the same thing as Prow's OWNERS-derived `approved` label, so on a Prow repo a hive-queued PR would still need `approved` from an OWNERS reviewer. This PR does not change that, and I have not tested hive against a Prow-managed target repo.
- Repos that do not run Prow have no consumer for either name, which is the open question at the end of the description.

Neither caveat changes the shape of this PR — it only makes the default look better-founded. Still entirely your call.

clubanderson:
Thanks @castrojo — this is a good improvement. I rebased it onto the current v2 branch and added a small maintainer follow-up so the merge-eligible queued marker and dashboard queued-state detection also honor the configured label exactly, with extra tests for configured-label queueing/defaults. Local validation passed: `cd v2 && go build ./... && go test ./cmd/hive ./pkg/config ./pkg/dashboard ./pkg/github ./pkg/scheduler -count=1 && go vet ./cmd/hive ./pkg/config ./pkg/dashboard ./pkg/github ./pkg/scheduler`.

Issue #2546, opened 2026-08-04
contribute: some no-task outcomes send nothing, so an idle contributor cannot tell why


castrojo:
> Part of #2534 (contributor onboarding field notes from `projectbluefin/review`).

## Summary

#2436 established a good pattern: when the hub declines to assign work for an enforced
reason, it sends an explicit `task_unavailable` carrying a machine-readable reason instead
of silence. Three paths in `selectTask` still return a bare `nil`, and on those paths the
client receives no message at all — so "the hive has no work right now" is on the wire
indistinguishable from "you are misconfigured" or "this connection is wedged." We are
asking whether it is worth **extending the pattern you already built** to cover them; we
are not reporting a regression.

## Evidence

**The paths that send nothing.** `selectTask`
([`contribute_ws.go:1344`](https://github.com/kubestellar/hive/blob/957771e5f18604f3d301a3749f63b9896ba99e1c/v2/pkg/dashboard/contribute_ws.go#L1344))
returns a bare `nil` in four places:

- no server reference —
  [`contribute_ws.go:1348-1350`](https://github.com/kubestellar/hive/blob/957771e5f18604f3d301a3749f63b9896ba99e1c/v2/pkg/dashboard/contribute_ws.go#L1348-L1350)
  (arguably not contributor-visible in practice; listed for completeness)
- `Hub.ContributeSuspended` — the operator has suspended contribution —
  [`contribute_ws.go:1351-1353`](https://github.com/kubestellar/hive/blob/957771e5f18604f3d301a3749f63b9896ba99e1c/v2/pkg/dashboard/contribute_ws.go#L1351-L1353)
- no status snapshot available yet —
  [`contribute_ws.go:1359-1361`](https://github.com/kubestellar/hive/blob/957771e5f18604f3d301a3749f63b9896ba99e1c/v2/pkg/dashboard/contribute_ws.go#L1359-L1361)
- empty candidate set after filtering —
  [`contribute_ws.go:1582-1584`](https://github.com/kubestellar/hive/blob/957771e5f18604f3d301a3749f63b9896ba99e1c/v2/pkg/dashboard/contribute_ws.go#L1582-L1584)

**Confirmed: `nil` means nothing is sent.** The `ready` handler logs and returns without
writing to the socket — the comment even enumerates the three conditions:

```go
case task == nil:
    // No admissible work and no enforced refusal (e.g. suspended, no
    // status yet, or an empty candidate set). Behaviour unchanged.
[...]

Issue #2436, opened 2026-08-01
Contributor trust controls are advisory: silent nil on token-mint failure, inert disabled_tiers/tier_limits, and auto-promotion on unverified completions


castrojo:
## Summary

Four defects in the contributor subsystem share one shape: a control that looks
authoritative is not enforced, and the resulting behaviour is silent.

1. A scoped-token mint failure makes `selectTask` return `nil`, so the
   contributor is told nothing and waits indefinitely — the log line even says
   "skipping task" while the code abandons the whole selection.
2. `hub.disabled_tiers` is in the config schema and writable through the admin
   API, and is never read.
3. `hub.tier_limits` is the same, and it ships **populated defaults** including
   `MaxConcurrent`, so an operator reasonably believes concurrency is capped
   when nothing enforces it.
4. Auto-promotion to `contributor` (which grants `contents:write` and
   `pulls:write`) counts self-reported completions and never consults the PR
   URL — even at HEAD, where that URL is already in hand and already used to
   pick the cooldown length.

Filing as one issue because they are the same class and the fixes touch the same
files; happy to split if you would rather track them separately.

Verified against `v2` at `e73f9c6cd650ed50fff22f5d5ac232bd8b7f434e` (the commit
our hub runs) and re-checked against `v2` HEAD `835448c3`. All four are present
at both.

## 1. Token-mint failure silently drops the task

At `e73f9c6`, `v2/pkg/dashboard/contribute_ws.go:874-880`, inside the candidate
loop:

```go
if tok, err := h.server.deps.GHAppAuth.ScopedToken(ctx, c.profile.TrustTier); err == nil {
    ghToken = tok
} else {
    h.logger.Warn("[contribute-ws] failed to mint scoped token — skipping task",
        "tier", c.profile.TrustTier, "error", err)
    return nil
}
```

The message says "skipping task", but `return nil` exits `selectTask` entirely
rather than continuing to the next candidate.

At HEAD the mint moved into the shared `mintScopedToken` helper
(`:1010-1024`, added for #2393 item 2 — thank you) and now runs once *after* a
candidate has been chosen (`:1233-1238`). The failure mode persists and arguably
sharpens: an already-selected task is discarded and nothing is sent.

`ScopedToken` (`v2/pkg/github/app.go`) requests different permission sets per
tier, and `CreateInstallationToken` rejects any permission the installation does
not hold. So an installation missing `contents` or `pull_requests` write fails
[...]

PR #2846, opened 2026-08-08
feat(contributor): support subscribing to multiple hubs from one relay session

hanthor:
## Summary

Lets one contributor CLI/tmux session hold work from more than one hive, instead of requiring a duplicate container per hub. Comma-separated `HIVE_HUB` + `HIVE_REGISTRATION_TOKEN` (one token per hub, same order). A single value of each — the existing common case — behaves exactly as before.

## Design

- Each hub gets its own WebSocket connection/reconnect/heartbeat state (`hubs[]`); `currentTask`, `cliReady`, and everything CLI-facing stay single global values, since there's exactly one CLI/tmux session shared across whichever hub currently holds the active task.
- Only one hub is "active" (asked for work with `ready`) at a time; the others stay connected + heartbeating, idle.
- The active slot round-robins forward on an explicit `task_unavailable` ack from the active hub. This ack is reliable — the hub always sends it, never silently withholds one (#2436 / #2546) — so rotation is driven by a real signal rather than a guessed timeout, and the relay never sits idle on a hub with no work while a different configured hub has some.
- Completing/failing/reverting a task re-asks the *same* hub that gave it, rather than switching eagerly — round-robin only fires on an explicit "no work here."

## Testing

- `node bin/contributor-relay.test.js` — 31/31 (added 5 new tests covering hub-list parsing, active/idle auth_ok routing, busy-rejection routed to the asking hub not the active-task hub, task_unavailable rotation, and a regression test for a bug this PR shipped and then fixed in the same branch — see below).
- Verified live against `hive.tunaos.org` + the hosted Project Bluefin hive: both hubs connect and authenticate concurrently, round-robin correctly rotates past a rate-limited hub to one with real work, and a real `task_assign` is delivered end-to-end and completed.

## A bug this caught (worth flagging for review)

First pass attached the whole `hub` object to `currentTask._hub` for later routing. Since `currentTask` is the exact object that later gets `JSON.stringify`'d to `TASK_FILE`, and `hub` holds live `setInterval`/`setTimeout` handles (circular references), every task assignment crashed the process (`TypeError: Converting circular structure to JSON`) the moment a real task came in — caught only by running against real hubs, not by the unit tests (which used bare `{ws}` stubs without live timers). Fixed by making `_hub` a non-enumerable property via `Object.defineProperty`, and added a regression test that populates real `setInterval`/`setTimeout` handles on the hub before asserting `JSON.stringify(currentTask)` doesn't throw.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -173,8 +173,16 @@ function sendTo(hub, msg) {
 // hub regardless of currentTask/activeHubIndex (auth handshake, rejecting a
 // task from a hub that isn't getting the active slot, per-hub ping/pong) use
 // sendTo() directly.
+// The `|| hubs[activeHubIndex]` fallback is load-bearing, not defensive
+// padding: not every currentTask comes from a task_assign. The synthetic
+// pr-review task built after every PR_REVIEW_EVERY_N completions is assembled
+// locally and has no _hub, so keying strictly off currentTask._hub sent its
+// progress and completion frames to `undefined` — silently dropped, leaving
+// the hub to watch the contributor go mute mid-review and time it out.
+// Falling back to the active hub is also the correct target there: it is the
+// hub whose task we just finished.
 function send(msg) {
-  sendTo(currentTask ? currentTask._hub : hubs[activeHubIndex], msg);
+  sendTo((currentTask && currentTask._hub) || hubs[activeHubIndex], msg);
 }
 
 function injectGhToken(token) {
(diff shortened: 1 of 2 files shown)


clubanderson:
Thanks @hanthor — this is a useful contributor-relay improvement and the per-hub connection/heartbeat model looks good.

I pushed a small maintainer fix while landing it: non-active hubs can no longer inject task assignments, revoke another hub's active task, or refresh/overwrite the active task's GitHub token; auth failure on the active hub now advances polling to a healthy authenticated hub. I also kept the relay tests off production cache/task-file paths.

Validated locally with relay syntax check, node bin/contributor-relay.test.js (42/42), git diff --check, and cd v2 && go build ./.... CI build/docker/DCO are green.

/lgtm
/approve

Issue #2203, opened 2026-07-29
Contributor agent stuck in infinite crash loop after periodic CLI restart (tmux prompt quoting bug)


castrojo:
## Summary

The `hive-contributor` container's agent script gets stuck in an infinite crash loop after its periodic "restart CLI for memory cleanup" (every 3 completed tasks in our observation). The same issue gets reassigned and fails repeatedly, forever, never actually reaching the CLI.

## Environment

- Image: `ghcr.io/kubestellar/hive-contributor:latest`
- Backend: `copilot`
- Hub: hosted instance (`*.hive.kubestellar.io`)
- Wrapper: rootless Podman Quadlet run via https://github.com/castrojo/donate-clanker (thin wrapper around the documented `contribute-hive` container run — not implicated, the crash is inside the container's own agent script)

## Timeline from container logs

```
Task assigned: issue projectbluefin/bluefin-lts#360 ...
Task ct-...-360-... completed — agent idle
Task assigned: issue projectbluefin/bluefin-lts#371 ...
Task ct-...-371-... completed — agent idle
Task assigned: issue projectbluefin/bluefin-lts#391 ...
Task ct-...-391-... completed — agent idle
Task assigned: issue projectbluefin/bluefin-lts#421 ...
Restarting copilot CLI for memory cleanup (task 3)
Task prompt sent to CLI
CLI process (copilot) died — restarting and reporting task as failed
CLI restarted: copilot --allow-all
Task ct-...-421-1785363856 failed: CLI process exited — restarted
Task assigned: issue projectbluefin/bluefin-lts#421 ...
CLI not ready yet — queuing task prompt
CLI process (copilot) died — restarting and reporting task as failed
CLI restarted: copilot --allow-all
Task ct-...-421-1785364110 failed: CLI process exited — restarted
[... repeats every ~4 minutes, indefinitely, same issue #421 ...]
```

The first 3 tasks (#360, #371, #391) completed normally. Immediately after the 3rd completion, the script logs `Restarting copilot CLI for memory cleanup (task 3)` and assigns task #4 (#421). From that point on, every attempt at #421 fails identically and the loop never recovers or gives up.

## tmux pane evidence (`tmux capture-pane -t contributor -p`)

```
dev@6218846cce14:~$ copilot --allow-all --model gpt-5.6-luna
dev@6218846cce14:~$ copilot --allow-all
[...]

PR #2849, opened 2026-08-08
🐛 fix(contributor): recover instead of stalling when the CLI never becomes ready

hanthor:
Three compounding faults let a contributor hold a task indefinitely without ever running it. I hit all three on a codex-backend contributor, but **fault 1 is backend-agnostic** — it black-holes a task for *any* cause of an unresponsive CLI.

## 1. A readiness timeout silently abandoned the task (the important one)

```js
waitForCLI().then(() => { cliReady = true; flushPendingTask(); })
            .catch(e => console.error(e.message));   // ← that's all
```

On `CLI_READY_TIMEOUT_MS`: `cliReady` stays false, `pendingTask` keeps holding the prompt, and **the hub is never told**. From the hub's side the contributor is still working the issue, so the slot stays held until the hub's own timeout revokes it. Unrecognized modal prompt, crashed pane, half-finished login, hung update — every one produces the same black hole. This is the same slot-starvation class as #2203.

Now the task is handed back with `task_failed` so it can be reassigned immediately, and the relay keeps waiting for the CLI.

Importantly it does **not** re-advertise `ready` while still wedged (new opt-in `skipReady`). Sending `ready` there would pull in another task it still cannot run and churn one task per timeout window — strictly worse than the bug. Availability is re-advertised only once the CLI genuinely reaches its prompt.

## 2. codex modals were never classified — and matched as *ready*

codex had no `onboarding` branch at all, so its prompts were never dismissed. Worse, codex draws its version banner and input chrome **behind** the modal, so the existing `ready` patterns (`/codex>|>\s*$|Codex CLI/`) match a pane that is actually blocked — the relay types the task prompt into a menu, where it's swallowed.

Modal patterns are now checked **before** the ready patterns, exactly as they already are for `bob` (whose comment explains the same hazard).

## 3. Dismissal only ever sent a bare Enter

Which does nothing to a **numbered menu** but re-render it — so the "auto-dismiss" looped until the timeout above. `blockingPromptKey()` now picks the right key from the pane text:

- `1` for codex's directory-trust menu (`1. Yes, continue`)
- `3` for its update nudge (`3. Skip until next version`) — deliberately **not** `1. Update now`, which shells out to `npm install -g` inside the container: slow, needs network, can fail half-way, and drifts the CLI version out from under the image. "Skip until next version" also persists, so it stops recurring on every restart.

A plain yes/no confirm still gets a bare Enter, so existing backends are untouched.

## Testing

`node bin/contributor-relay.test.js` — 32/32 (26 pre-existing + 6 new). The pane fixtures are **real captures** from a running `ghcr.io/kubestellar/hive-contributor` container, including the chrome-behind-the-modal rendering that defeats the ready patterns.

Verified live: codex sat at the update menu until the relay timed out and dropped `projectbluefin/fsdk-containers#89`; with this patch bind-mounted it auto-selects `3` and proceeds to its prompt.

## Relationship to #2845

#2845 fixes the same codex trust prompt one layer up, in `contributor-agent.sh`'s launcher loop. Both loops scrape the pane independently, so both need the fix — but they're separate files with separate failure windows and are reviewable independently. Happy to combine them if you'd prefer one PR.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -1120,7 +1120,11 @@ function handleMessage(data) {
       }
       reconnectDelay = BASE_RECONNECT_DELAY_MS;
       if (!currentTask) {
-        send({ type: 'ready', seq: nextSeq() });
+        if (!cliReadyFailed) {
+          send({ type: 'ready', seq: nextSeq() });
+        } else {
+          console.log('Authenticated, but CLI readiness previously failed — withholding ready until the CLI recovers');
+        }
       } else {
         console.log(`Reconnected while working on ${currentTask.repo}#${currentTask.number} — resuming`);
         send({ type: 'task_accepted', seq: nextSeq(), task_id: currentTask.task_id });
@@ -1303,6 +1307,8 @@ if (process.env.HIVE_RELAY_TEST_MODE === '1') {
     MAX_TASK_CLI_RESTARTS,
     setCliReady: (v) => { cliReady = v; },
     getCliReady: () => cliReady,
+    setCliReadyFailed: (v) => { cliReadyFailed = v; },
+    getCliReadyFailed: () => cliReadyFailed,
     getPendingTask: () => pendingTask,
     setPendingTask: (v) => { pendingTask = v; },
     setTasksCompletedCount: (v) => { tasksCompletedCount = v; },
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -378,13 +378,46 @@ function recoverWedgedShell() {
   } catch (_) {}
 }
 
-function getCLIState() {
+// capturePaneText returns the current pane contents, or "" if tmux can't be
+// reached. Extracted so the readiness classifier and the blocking-prompt
+// dismissal can look at the SAME text without capturing twice, and so the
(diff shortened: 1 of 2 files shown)


clubanderson:
Thanks @hanthor — this is a solid fix for the wedged-CLI black-hole path. I validated the codex onboarding classifier/dismissal, the task hand-back behavior, and the existing bounded CLI restart cleanup paths.

I added one maintainer follow-up before landing: if the websocket reconnects after a readiness timeout but before the CLI recovers, `auth_ok` now also withholds `ready`; otherwise reconnect could bypass the new `skipReady` guard and accept another unrunnable task. Added regression coverage for that case.

Validation run locally on the PR rebased to `v4`:
- `node --check < bin/contributor-relay.sh`
- `node bin/contributor-relay.test.js` (33/33)
- `bash -n`: no bash/shebang shell changes in this PR (`contributor-relay.sh` is Node)
- `cd v2 && go build ./...`
- `go test`/`go vet`: no touched Go packages

Relationship to #2845/#2846: #2845 is a separate file/layer; #2846 also touches the relay and will need to account for this ready-withholding state, but this PR is clean to land independently.

PR #2834, opened 2026-08-08
✨ feat(leaderboard): custom stylesheet from a public GitHub repo via ?style= param

clubanderson:
## Summary
- Add shareable ClankeR leaderboard custom styles via `?style=owner/repo/path/to/theme.css@ref` (ref optional, defaults to HEAD)
- Add `GET /api/leaderboard/style?src=...` for server-side unauthenticated public GitHub raw fetches with 5-minute in-memory cache
- Preserve `/leaderboard?style=...` redirects and wire the Me-card style dropdown to show Custom (owner/repo); choosing a built-in style clears the param
- Document the param format and security constraints in README

## Security model
- No full URLs accepted; only validated owner/repo/path.css triplets with optional @ref
- Server fetches from raw.githubusercontent.com (test-overridable base) without tokens; no client-side third-party fetch
- CSS is capped at 128 KiB, stripped of @import, external url(...), image-set(), legacy expression/-moz-binding/behavior vectors, CSS escape lines, and scoped to #tab-leaderboard
- Served back as text/css from same-origin /api/leaderboard/style, so the existing CSP style-src 'self' 'unsafe-inline' remains unchanged
- Hub-side /api/hub/leaderboard is a separate JSON consumer; this PR scopes UI stylesheet support to the spoke ClankeR /contribute leaderboard and leaves kubestellar.io hub presentation as follow-up

## Validation
- `go build ./...`
- `go vet ./...`
- `go test ./pkg/dashboard/ -count=1`

PR #2851, opened 2026-08-08
✨ feat(dashboard): custom stylesheet param for the spoke dashboard and read-only preview

clubanderson:
Extends #2834's `?style=owner/repo/path.css[@ref]` custom-CSS support to the **main spoke dashboard** and the **read-only preview**.

- Shared sanitizer/fetch/cache core (leaderboard endpoint unchanged, tests still green)
- `GET /api/style?src=...&scope=...` — server-side raw.githubusercontent fetch, sanitized (no `@import`/external `url()`), 128 KiB cap, scoped to the dashboard root so themes can't restyle the login overlays; CSP untouched
- Dismissible indicator chip + toast; success only reported from `link.onload` (failure can never masquerade as active)
- Operator requirement: 'i want to offer this css option to the hive spoke's dashboard and the read-only preview as well.'

### DIFF
--- a/v2/README.md
+++ b/v2/README.md
@@ -91,18 +91,24 @@ project:
   ai_author: your-bot-user
 ```
 
-### Leaderboard custom stylesheet
+### Custom stylesheets
 
-The ClankeR leaderboard accepts a shareable custom stylesheet parameter:
+The ClankeR leaderboard, the spoke dashboard, and the read-only `/snapshot`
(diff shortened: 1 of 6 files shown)


clubanderson:
@castrojo — the same `?style=owner/repo/path.css@ref` custom-CSS support from the leaderboard (#2834) now works on the full spoke dashboard and the read-only preview link too.

PR #2889, opened 2026-08-08
🐛 fix(contributor): detect a ready codex pane (› marker, real banner text)

hanthor:
## The bug

An interactive **codex** contributor never runs a task. Not "sometimes fails" — never runs one at all.

```js
if (/codex>|>\s*$|Codex CLI/.test(text)) return 'ready';
```

Checked against a live pane from `ghcr.io/kubestellar/hive-contributor:latest`, all three alternatives miss:

| pattern | reality |
|---|---|
| `codex>` | codex never renders this |
| `/>\s*$/` | codex's input marker is **`›` (U+203A)**, not `>` |
| `Codex CLI` | the banner reads **"OpenAI Codex (v0.146.0)"** |

So `getCLIState()` returns `'starting'` forever. `tmuxSendKeys()` queues every assigned task waiting for `cliReady`, which never becomes true. Before #2849 the task sat there silently; with #2849 it's handed back at `CLI_READY_TIMEOUT_MS`. Either way codex does no work.

Measured on a running container, same image, three backends:

```
codex : "CLI ready" x0,  prompts delivered 0,  queued-never-sent 1
claude: "CLI ready" x1,  prompts delivered ≥1
agy   : "CLI ready" x1,  prompts delivered ≥1
```

The exact pane that classified as `starting`:

```
╭─────────────────────────────────────────────────────╮
│ >_ OpenAI Codex (v0.146.0)                          │
│ model:       gpt-5.6-luna medium   /model to change │
│ directory:   ~/workspace                            │
╰─────────────────────────────────────────────────────╯
  Tip: Use /rename to rename your threads for easier thread resuming.
› Run /review on my current changes
  gpt-5.6-luna medium · ~/workspace
```

(The `>_` inside the banner is followed by `_`, so `/>\s*$/` doesn't match it either.)

## The fix

Match the marker and banner codex actually prints, in `getCLIState()` and in `checkTmuxIdle()`, which carried the same `>` assumption.

**This does not weaken the modal guards.** Both modals draw `›` as well, but they're classified first and return `'onboarding'`, so a blocked pane still never reads as ready. Added a test asserting precisely that for the trust menu and the update nudge — that ordering is the thing most likely to break in future edits.

## Testing

`node bin/contributor-relay.test.js` — **58/58**, 2 new. Pane fixtures are verbatim captures, not hand-written approximations, since hand-written ones are exactly what would have hidden this (they'd have included a plausible-looking `codex>`).

Verified live: after bind-mounting this over the stock relay, the same container logged `CLI ready — accepting tasks` for the first time.

Follows #2849 (which makes the resulting stall recoverable) and #2845 (the trust prompt). Targeting `v4` since that's where those landed and what the published image tracks — happy to retarget to `v2` if you'd prefer it there too.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -544,7 +544,16 @@ function getCLIState() {
       // "1. Yes, continue / 2. No, quit" list, where it is swallowed.
       if (/Do you trust the contents of this directory/.test(text)) return 'onboarding';
       if (/Update available!/.test(text) && /Skip until next version/.test(text)) return 'onboarding';
-      if (/codex>|>\s*$|Codex CLI/.test(text)) return 'ready';
+      // codex renders its input marker as '›' (U+203A), not '>', and its
+      // banner reads "OpenAI Codex (vX.Y.Z)" — never the literal "Codex CLI".
+      // The three original patterns therefore matched NOTHING a real codex
+      // pane ever contains, so readiness was never detected: every task was
+      // queued and then handed back at CLI_READY_TIMEOUT_MS, and an
+      // interactive codex contributor could not run a single task. Matching
+      // the marker and the real banner is what makes the backend usable.
+      // Safe against the modals above: those are classified first and return
+      // 'onboarding', so a menu that also draws '›' never reaches here.
+      if (/codex>|›|OpenAI Codex|Codex CLI|>\s*$/.test(text)) return 'ready';
     } else if (BACKEND === 'pi') {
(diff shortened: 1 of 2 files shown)


clubanderson (approved):
Thanks @hanthor — I reviewed the diff and validated the relay test suite locally (`node --test bin/contributor-relay.test.js`, 58/58). The updated codex readiness detection matches the real U+203A prompt marker and the version-independent `OpenAI Codex` banner, while the trust/update modal checks still run first so blocked panes stay classified as onboarding. CI build/docker checks are green; tide is only pending labels. LGTM.

PR #2976, opened 2026-08-08
fix(leaderboard): CSS sanitizer allows custom props + safe at-rules and reports what it drops

### LINKED ISSUE #2972: leaderboard custom CSS: the sanitizer silently drops custom properties, at-rules and some selectors, so most themes render as no-ops
Following up on #2612 and the design in [projectbluefin/documentation#1024](https://github.com/projectbluefin/documentation/issues/1024).

I published a theme and pointed the hive at it:

```
/contribute/leaderboard?style=projectbluefin/documentation/static/hive/leaderboard.css@main
```

It loaded, the dropdown showed **Custom**, and nothing changed visually. The feature works; the sanitizer is stricter than the in-page help says, and it fails **silently** — a rejected rule is simply absent, so the page looks healthy while the theme does nothing.

Everything below was found by fetching the sanitized result back and diffing it against what was sent:

```bash
curl -s "https://<hive>/api/leaderboard/style?src=projectbluefin%2Fdocumentation%2Fstatic%2Fhive%2Fleaderboard.css%40main"
```

## What the help text says

> Public GitHub repos only; CSS is sanitized server-side and capped at `128 KiB`.

## What actually happens

A whole rule is rejected if any declaration in it is disallowed, and the disallowed set is wider than documented.

| Behaviour | Consequence |
| --- | --- |
| **Custom property declarations are dropped.** A rule containing `--x: value` is discarded whole. | **This is the important one.** The built-in styles are palette variations over `--me-accent` / `--me-accent-soft`, so the natural way to write a theme is to set those. A custom theme *cannot*. My first version defined `--bf-*` in one rule and referenced it in twelve others; the defining rule was dropped, so twelve rules referenced variables that no longer existed and the theme was inert. |
| **All at-rules are dropped**, `@media` included. | No `prefers-color-scheme`, no `prefers-reduced-motion`. |
| **Selectors are scoped by prepending `#tab-leaderboard `.** | An ancestor selector becomes unmatchable: `[data-theme="light"] .me-card` → `#tab-leaderboard [data-theme="light"] .me-card`, and `data-theme` is on `<html>`, above the scope. **A theme therefore cannot branch on light versus dark at all** — neither via `[data-theme]` nor via `@media`. |
| `linear-gradient()` is dropped. | Flat colours only. |
| `.me-card` and `.me-card .me-card__band` are rejected outright. | Sibling selectors with identical declarations survive, so it is the selector, not the properties. Cause unclear — possibly an anti-defacement guard on the card root. |
| The scoping prefix is applied to comments too. | `#tab-leaderboard /* comment */` ends up in selector position, and `#tab-leaderboard` gets injected *inside* multi-line comments. Harmless after comment stripping, but it suggests the prefixer is string-based rather than parsing the stylesheet. |

## Why this matters

The light/dark point is the sharp end. The hive ships both themes, and a theme author has no mechanism to support both — so every custom theme will look correct to its author and wrong to half its readers. My theme works around it with a single palette that clears WCAG on both surfaces (`#5c7bd1`: 4.29:1 on `#161b22`, 3.79:1 on `#f6f8fa`), but that is a workaround, not a solution.

## Suggestions, roughly in order of value

1. **Report what was stripped.** The UI already has a notice slot for a bad stylesheet. Extending it to "loaded, 4 of 16 rules removed" would turn a silent no-op into a five-second fix. A `X-Style-Rules-Dropped` response header would be enough for authors working from `curl`.
2. **Allow `--*` declarations**, or at least allowlist `--me-*`. They cannot fetch and cannot escape the scope, so the risk is low, and it is the mechanism the built-in styles already use.
3. **Give themes a light/dark hook that survives scoping** — e.g. also mirror `data-theme` onto the scoping element, so `#tab-leaderboard[data-theme="light"]` matches, or scope with `:where()` so author selectors keep their own structure.
4. **Document the real rules** in the Custom CSS popover. Even without any change to the sanitizer, listing "no custom properties, no at-rules, no gradients, cannot branch on theme" would have saved this round trip.

Happy to send a PR against the docs popover if that is useful. The working theme, with the constraints written up in its header, is [`static/hive/leaderboard.css`](https://github.com/projectbluefin/documentation/blob/main/static/hive/leaderboard.css).

clubanderson:
Fixes #2972

## The report

Jorge (castrojo) published a leaderboard theme, pointed the hive at it, saw the dropdown flip to **Custom**, and nothing changed. The feature "works," but the sanitizer is stricter than the help text and **fails silently** — a rejected rule is simply absent, so the page looks healthy while the theme does nothing. His examples: a rule defining `--bf-*` custom properties was dropped whole (so twelve rules referencing them went inert), every `@media` block was dropped (no light/dark), and rules after an at-rule went missing.

## What was actually happening (`v2/pkg/dashboard/api_leaderboard_style.go`)

The sanitizer was **string/line based**:

- `scopeCustomStyle` (old line ~252) dropped any selector containing `@`, so **every at-rule — `@media`, `@supports`, `@keyframes` — was discarded**.
- The brace matcher took the *first* `}` as the block close. On the first **nested** brace (the inner rule of an `@media`) it desynced and **corrupted every rule after the at-rule** — reproduced: input `@media {...} .after{...}` produced a stray `#tab-leaderboard }` and lost `.after`.
- **Nothing removed was ever reported** — the core complaint.

## The fix

Rewrote the sanitizer around a **balanced-brace block parser** (`matchingBrace`) that recurses into a **safe at-rule allowlist**, plus comment stripping up front (the old string prefixer even injected the scope prefix *inside* comments).

Now allowed (the safe subset themes need):
- **Custom properties** (`--foo: …`) and **`var(--foo)`** references — the backbone of theming. Values still run through the same `url()`/token value checks, so `--leak: url(https://evil…)` is neutralised to `url("")`.
- **`@media` / `@supports`** — inner rules recursed and scoped.
- **`@keyframes` / `@-webkit-keyframes`** — inner keyframe stops (`from`/`to`/`%`) kept and value-sanitized, not scope-prefixed.
- A single unsafe declaration is dropped **without** discarding the rest of its rule (the old all-or-nothing behaviour is what made themes silently inert).

## Vectors deliberately still blocked (security reasoning)

This is a sanitizer for a page that renders other users' content, so loosening it wrong is an XSS/exfil vector. Still rejected, each **with a reported reason**:

- **`@import`** — fetches arbitrary remote CSS (injection / exfil).
- **`@font-face`** — its `src:` fetches a font over the network; can't be constrained to "no network" while staying useful, so rejected rather than half-allowed.
- **`expression()` / `-moz-binding` / `behavior:`** — script execution.
- **`image-set()` / remote `url()` / `://`** — request-based exfil.
- **Escape sequences (`\`)** — can smuggle otherwise-blocked tokens.
- **Sibling/child combinators off the scope root** (e.g. `#tab-leaderboard ~ footer`) — re-prefixed so a theme can't reach elements outside its widget.

## No more silent drops (the anti-silent-drop guarantee)

Every removal is now surfaced:
- `getCustomStyleWithDropped` returns a `[]droppedCSSRule` (`{rule, reason}`), cached alongside the CSS.
- `/api/leaderboard/style` (and `/api/style`) emit **`X-Style-Rules-Dropped`** (count — Jorge's own curl-friendly suggestion) and **`X-Style-Rules-Dropped-Detail`** (JSON list).
- The `/contribute` leaderboard notice lists the removed rules with reasons ("…N rules removed by the sanitizer: …").

## Tests

Added to `api_leaderboard_style_test.go`: custom props + `var()` survive while `--x: url(evil)` is stripped; `@media`/`@supports`/`@keyframes` survive with inner declarations sanitized and `@import`/`@font-face` rejected; the rule after an `@media` is no longer corrupted; every dangerous vector is both blocked **and** reported; the response header and contribute notice carry the dropped list. Full `pkg/dashboard` suite passes; `go build ./...` and `go vet ./...` clean.

### DIFF
--- a/v2/pkg/dashboard/api_contribute.go
+++ b/v2/pkg/dashboard/api_contribute.go
@@ -533,7 +533,7 @@ func (s *Server) handleContributeLanding(w http.ResponseWriter, r *http.Request)
 	customStyleHeadHTML := ""
 	customStyleNoticeHTML := ""
 	if rawStyle := strings.TrimSpace(r.URL.Query().Get("style")); rawStyle != "" {
-		if _, src, err := getLeaderboardCustomStyle(r.Context(), rawStyle); err == nil {
+		if _, src, dropped, err := getLeaderboardCustomStyleWithDropped(r.Context(), rawStyle); err == nil {
 			styleKey := leaderboardCustomStyleCacheKey(src)
 			escapedSrc := html.EscapeString(styleKey)
 			styleKeyJSON, _ := json.Marshal(styleKey)
@@ -542,7 +542,17 @@ func (s *Server) handleContributeLanding(w http.ResponseWriter, r *http.Request)
 				url.QueryEscape(styleKey),
 				string(styleKeyJSON),
 			)
-			customStyleNoticeHTML = fmt.Sprintf(`<div class="lb-custom-style-note" id="leaderboard-custom-style-note" role="status">Custom style active: <code>%s</code></div>`, escapedSrc)
+			// Anti-silent-drop (#2972): if the sanitizer removed anything, say so
+			// in the notice with the reasons, so a theme author can see WHY a rule
+			// did not apply instead of staring at a healthy-looking no-op.
+			if len(dropped) > 0 {
+				customStyleNoticeHTML = fmt.Sprintf(
+					`<div class="lb-custom-style-note lb-custom-style-note--warn" id="leaderboard-custom-style-note" role="status">Custom style active: <code>%s</code> — %d rule%s removed by the sanitizer:%s</div>`,
+					escapedSrc, len(dropped), plural(len(dropped)), droppedRulesHTML(dropped),
+				)
+			} else {
+				customStyleNoticeHTML = fmt.Sprintf(`<div class="lb-custom-style-note" id="leaderboard-custom-style-note" role="status">Custom style active: <code>%s</code></div>`, escapedSrc)
+			}
 		} else {
 			customStyleNoticeHTML = `<div class="lb-custom-style-note lb-custom-style-note--warn" id="leaderboard-custom-style-note" role="status">Custom style could not be loaded — using default <button type="button" onclick="this.parentElement.remove()">Dismiss</button></div>`
 		}
@@ -1403,6 +1413,9 @@ code{background:var(--cc-bg);padding:2px 8px;border-radius:4px;font-size:.9rem}
 .lb-custom-style-note{margin:0 0 16px;padding:10px 12px;border:1px solid rgba(88,166,255,.35);border-radius:8px;background:rgba(88,166,255,.10);color:var(--cc-text);font-size:.86rem}
 .lb-custom-style-note code{color:var(--cc-accent)}
 .lb-custom-style-note--warn{border-color:rgba(210,153,34,.45);background:rgba(210,153,34,.12)}
+.lb-custom-style-dropped{margin:8px 0 0;padding-left:18px;font-size:.82rem;line-height:1.5}
+.lb-custom-style-dropped li{margin:2px 0}
+.lb-custom-style-dropped code{color:var(--cc-accent)}
 .lb-custom-style-note button{margin-left:8px;background:transparent;border:1px solid var(--cc-border);border-radius:6px;color:var(--cc-text);padding:2px 8px;cursor:pointer}
 </style>%s</head><body>
 <div class="page-tabs" role="tablist">
--- a/v2/pkg/dashboard/api_leaderboard_style.go
+++ b/v2/pkg/dashboard/api_leaderboard_style.go
@@ -2,7 +2,9 @@ package dashboard
 
 import (
 	"context"
+	"encoding/json"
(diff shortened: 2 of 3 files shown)


clubanderson:
Superseded by #3033 (merged to v2 on 2026-08-09), which resolved the same issue (#2972) with a stricter sanitizer.

#3033 is a superset of this PR's security posture — it keeps every block this PR added (`@import`, `expression()`, `-moz-binding`, `behavior:`, remote-`url()` allowlist, escape-sequence rejection, balanced-brace parser, and dropped-rule reporting) and adds more hardening: `@font-face` with a `src` allowlist, `@container` handling, sibling-combinator scope-escape checks, theme-ancestor scoping, and per-declaration URL sanitization. Rebasing this branch would conflict across the whole sanitizer file and risk reverting that stronger, already-shipped behavior, so closing rather than merging.

Two non-security extras from this PR that #3033 doesn't have — the per-rule `X-Style-Rules-Dropped-Detail` header and a `maxAtRuleNestDepth` recursion cap — can be picked up as a small follow-up on top of current v2 if wanted.

Issue #3163, opened 2026-08-10
Two tests fail on clean v4 in CI: beads setgid assertion and sandbox podman skip


clubanderson:
Two tests fail in CI on **clean, unmodified `v4`** and pass locally on macOS. Filing alongside #3162 (the data race) so the full set of v4 CI-only failures is tracked.

Verified pre-existing by dispatching `v2-tests.yml` against untouched `v4` — run [31407709737](https://github.com/kubestellar/hive/actions/runs/31407709737) produced the identical failures with no PR changes in play.

## 1. `TestNewStore_CreatesGroupWritableDir` (`pkg/beads`)

```
beads_perms_test.go:45: dir mode drwxrwx--- lacks setgid bit
```

`v2/pkg/beads/beads_perms_test.go:44-46`:
```go
if runtime.GOOS == "linux" && fi.Mode()&os.ModeSetgid == 0 {
    t.Errorf("dir mode %v lacks setgid bit", fi.Mode())
}
```

The assertion is guarded to `runtime.GOOS == "linux"`, so it **can never fail on a macOS dev machine** — the mode is otherwise correct (`0770`, as asserted on the preceding line). Either `NewStore` is not setting setgid on Linux, or the assertion is stricter than the code intends.

## 2. `TestPodmanRunSkippedWhenUnavailable` (`pkg/sandbox`)

Fails on the CI runner, passes locally. Likely depends on whether a `podman` binary is present/absent in the environment — the test name suggests it means to exercise the *unavailable* path, so it may be asserting a skip that does not happen when the runner has podman installed (or vice versa).

## Why these surfaced now

`go test ./pkg/...` previously ran **only** on the `v2 Tests` cron (`schedule: */15`), which files a `ci-failure` issue rather than blocking anything — and the last green scheduled run was on branch **v2**, not v4. PR #3147 added a `pull_request` trigger to `v2-tests.yml` so unit tests gate PRs, which is why these now appear as a red check on every v4 PR.

## Impact

They currently make the `test` check red on unrelated v4 PRs, which erodes the signal the new PR trigger was added to provide. Worth fixing or explicitly skipping with a documented reason.

clubanderson:
Both tests now pass on clean v4:
```
go test ./pkg/beads/ -run TestNewStore_CreatesGroupWritableDir   → ok
go test ./pkg/sandbox/ -run TestPodmanRunSkippedWhenUnavailable  → ok
```
[...]

Issue #3162, opened 2026-08-10
Data race in requestHostedLiteSpoke: provisioning goroutine writes h.Status while handleLiteEnroll reads it


clubanderson:
`go test ./pkg/... -race` fails on `v4` with a data race between the hosted-lite-spoke provisioning goroutine and the handler that spawned it.

## The race

`v2/pkg/hub/lite_enrollment.go`:

- **Write** (goroutine spawned at `:361`, writes at `:364-365`):
  ```go
  go func() {
      defer provisionWG.Done()
      if err := provisionHive(h, req, &cluster, s.appKeysByAppID(), s.logger); err != nil {
          h.Status = "error"      // <-- unsynchronized write
          h.Error = err.Error()
  ```
- **Read** (`handleLiteEnroll`, `:191`): `SpokeStatus: spoke.Status` — reads the same struct field while that goroutine may be writing it.

Neither access is guarded, so the hive record `h` is shared across the goroutine boundary with no mutex.

## Reproduce

```
cd v2 && go test ./pkg/hub/ -race -run TestHandleLiteEnrollHostedSpokeRequestedWhenNoExistingSpoke -count=1
```

Full trace from run [31406620787](https://github.com/kubestellar/hive/actions/runs/31406620787):

```
WARNING: DATA RACE
Write at 0x00c00170e088 by goroutine 2750:
  hub.(*HubServer).requestHostedLiteSpoke.func1()
      v2/pkg/hub/lite_enrollment.go:364

Previous read at 0x00c00170e088 by goroutine 2749:
  hub.(*HubServer).handleLiteEnroll()
      v2/pkg/hub/lite_enrollment.go:191
```

## Why it surfaced now

It has been failing on `v4` for a while but was invisible: `go test` ran only on the `v2 Tests` cron, which files a `ci-failure` issue rather than blocking anything, and the last green scheduled run was on branch **v2**. PR #3147 added a `pull_request` trigger to `v2-tests.yml`, so it now shows as a red check on every v4 PR.

It also cannot reproduce on macOS dev machines in a plain `go test` — it needs `-race`.

## Scope

Two other tests fail on clean `v4` in CI for unrelated reasons; filing those separately:
- `TestNewStore_CreatesGroupWritableDir` (`pkg/beads`) — "lacks setgid bit", assertion is Linux-only guarded
- `TestPodmanRunSkippedWhenUnavailable` (`pkg/sandbox`)

[...]

PR #3172, opened 2026-08-10
🔒 Security audit 2026-08-07: N15 — owner-only, symlink-safe per-agent /tmp state files

clubanderson:
Follows #3164. This is §7.4 hardening — the four **mandatory** §7.1 fixes are all merged now (#3140, #3147, #3155, #3164).

## The problem

Two per-agent control files live in the pod-shared `/tmp` at paths derived from the agent **name** (guessable), both written with a plain `os.WriteFile` at `0o644`:

| File | Who reads it | Why it's authoritative |
|---|---|---|
| `/tmp/.hive-mode-<name>` | `bin/gh-wrapper.sh` | Takes its enforcement mode from this file **in preference to** the trustworthy env var |
| `/tmp/.hive-bootstrap-<name>.txt` | goose launch | Contents become the agent's first instruction |

`os.WriteFile` opens `O_WRONLY|O_CREATE|O_TRUNC` with **no `O_NOFOLLOW`**, so a pre-planted symlink at the predictable path redirects the hive's own write to any file the process can reach. Combined with a world-writable `/tmp` and world-readable results, one agent UID could steer another — un-gate the victim's `gh` access, or supply the prompt it executes at message zero with its own credentials.

All four write sites now route through `writeAgentStateFile`: `0600`, `O_NOFOLLOW`, plus an explicit `Chmod` so a `0644` file left by a previous release is **tightened** rather than inherited (`O_CREATE` applies its mode only when creating). Deliberately **not** `O_EXCL` — these are rewritten on every mode change and relaunch.

## Scope correction — this is prompt injection, not RCE

I want to be explicit, because I previously characterized this as arbitrary **code** execution and that was wrong.

The theory was that since the goose launch string is run by a shell and contains `--text "$(cat <file>)"`, a nested `$(...)` inside the planted file would execute. **It does not.** POSIX shells expand command substitution once and do not re-scan the result, so the content reaches goose as literal text. Verified three ways:

- nested substitution — `$(touch /tmp/pwned)` inside the file → not executed
- double evaluation — `sh -c "sh -c \"...\""` → not executed
- quote breakout — `safe" ; touch /tmp/pwn ; echo "` → not executed, the substitution result is never re-parsed

The audit's original classification (prompt injection) was correct. That also changes the **fix**: the right answer is file permissions, not the argv restructuring the RCE theory would have implied — argv restructuring would not have addressed the actual attack, which is a planted *prompt*.

## Blast radius

Goose-only for the bootstrap half — `backendDefersStartupKick` routes claude, copilot, gemini and bob down a path that never writes the file. `bob_startup_kick_test.go:194` guards bob; nothing guarded goose. The mode-file half affects every backend.

## Regression

Four tests: the mode, the upgrade path from a pre-existing `0644` file, the symlink refusal (asserting the link **target** is not clobbered), and an over-correction guard proving legitimate rewrites still work. **Three of the four fail against the previous `os.WriteFile` behaviour.**

Also gitignores `.hive-agent-stubs-*/` — `pkg/agent` tests generate stub CLI binaries under a randomly-suffixed directory, which is scratch and should never be committed.

## CI note

`test` job failures on v4 are pre-existing and tracked in #3162 (a real data race in `requestHostedLiteSpoke`) and #3163 (beads setgid + sandbox podman). Separately, `TestDismissInferencePrompts_FixWith` is flaky under full-package load — it drives a real tmux session with `sleep`-based timing and a 10s deadline, and passes in isolation. Unrelated to this change, which touches no tmux path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,7 @@ node_modules/
 # Scratch output from extracting index.html's inline <script> bodies for
 # `node --check`. A build artifact of the verification step, never source.
 _inline.js
+
+# Test-run scratch: pkg/agent tests create stub CLI binaries under a
+# randomly-suffixed dir. Generated per run; never part of the tree.
+.hive-agent-stubs-*/
--- a/v2/pkg/agent/agent_state_file_test.go
+++ b/v2/pkg/agent/agent_state_file_test.go
@@ -1,0 +1,122 @@
+package agent
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// N15 (CWE-367/732/20): per-agent control state in the pod-shared /tmp must be
+// owner-only and must not follow symlinks.
+//
+// Two files matter, both at paths derived from the agent NAME (guessable):
+//
+//   - /tmp/.hive-mode-<name>       — bin/gh-wrapper.sh takes its enforcement mode
+//     from this file IN PREFERENCE to the trustworthy env var, so planting it
+//     un-gates the victim agent (that half is the N15A fail-closed fix).
+//   - /tmp/.hive-bootstrap-<name>.txt — goose is launched with the contents as
+//     its first instruction, so planting it injects an attacker-chosen prompt
+//     that runs with the victim's credentials at message zero.
+//
+// Both were written with a plain os.WriteFile at 0o644: world-readable, and
+// symlink-following (os.WriteFile passes no O_NOFOLLOW), so a pre-planted link
+// redirected the hive's own write to any file the process could reach.
+//
+// NOTE on scope: an earlier reading of this called the bootstrap path arbitrary
+// CODE execution, on the theory that a nested $(...) inside the file would be
+// expanded because the launch string is run by a shell. That is wrong — POSIX
+// shells expand command substitution once and do not re-scan the result, so the
+// content reaches goose as literal text. The finding is prompt injection, which
+// these file permissions are the correct fix for.
+
+func TestWriteAgentStateFileIsOwnerOnly(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, ".hive-mode-victim")
+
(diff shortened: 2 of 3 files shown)


clubanderson:
CI status for reviewers.

`build-and-test` ✅ and `docker` ✅ are green. The `test` job is red with **exactly** the three failures already present on clean `v4`:

- `TestNewStore_CreatesGroupWritableDir` (`pkg/beads`) — #3163
- `TestPodmanRunSkippedWhenUnavailable` (`pkg/sandbox`) — #3163
- `TestHandleLiteEnrollHostedSpokeRequestedWhenNoExistingSpoke` (`pkg/hub`) — #3162, a real data race

Confirmed pre-existing by dispatching `v2-tests.yml` against untouched `v4` (run 31407709737), which produced the identical three. This PR touches only `pkg/agent`.

One thing I want to flag honestly rather than bury: `TestDismissInferencePrompts_FixWith` failed once in my **local** full-package run of `pkg/agent`. It did **not** fail in CI on this branch, and a clean-`v4` full-package baseline passed. That test drives a real tmux session with `sleep`-based timing and a 10s deadline, and it passes in isolation — so it reads as an intermittent flake under load rather than a regression here. `dismissInferencePrompts` also touches none of the code this PR changes (the four write sites are in `SyncModeFiles`, `agentBootstrapEnv`, and the goose bootstrap block; the test is pure tmux pane interaction). Re-running locally to confirm before merge.

Also, a quick demonstration of what the fix actually prevents, using the real `writeAgentStateFile` logic against a planted symlink:

```
--- OLD behavior (os.WriteFile, follows symlink) ---
victim-secret now = "ISSUES_PRS_MERGE"      <-- clobbered through the link

--- NEW behavior (O_NOFOLLOW) ---
write error = open /tmp/.../.hive-mode-victim: too many levels of symbolic links
victim-secret now = "ORIGINAL-CONTENT"      <-- untouched
```

PR #3203, opened 2026-08-10
🐛 fix agent import gist keep-linked validation

### LINKED ISSUE #3059: [scanner] Agent import URL allowlist accepts gist.github.com but keep-linked parser rejects gist URLs
## Finding

The SSRF hardening in PR #3046 introduces an inconsistent URL policy for agent imports:

1. `validateImportURL()` allows `gist.github.com` in `importAllowedHosts`.
2. For `keepLinked=true`, `definitionSourceFromURL()` only accepts GitHub `.../blob/...` URLs or `raw.githubusercontent.com/...` URLs.
3. Typical gist URLs do not match those patterns, so imports can pass host validation/fetch but still fail in the keep-linked path with a parsing error.

This creates a user-visible failure mode and a misleading contract (host is allowlisted, but URL form is effectively unsupported for linking).

## Steps to Reproduce / Evidence

In PR #3046 (`v2/pkg/dashboard/api_agents.go`):

- `importAllowedHosts` includes `gist.github.com`.
- `validateImportURL()` enforces that host allowlist.
- `definitionSourceFromURL()` parses only:
  - `https://<host>/<owner>/<repo>/blob/<ref>/<path>`
  - `https://raw.githubusercontent.com/<owner>/<repo>/<ref>/<path>`

A `gist.github.com` URL with `keepLinked=true` therefore fails parsing in `definitionSourceFromURL()` after passing earlier validation.

## Recommendation

Make URL validation and keep-linked parsing use the same accepted contract:

- Either remove `gist.github.com` from the allowlist for import URLs,
- Or add explicit gist URL parsing/linking support (including clear docs/error messaging).

Also align the error text to list exactly the actually-supported hosts/forms.

---
*Filed by scanner agent (ACMM L4 — issues-only mode)*

clubanderson:
Fixes #3059

## Summary
- Keep one-shot gist URL imports supported, but reject gist URLs early when `keepLinked=true` with an actionable error.
- Validate agent import URL hosts/forms before fetching and align errors with the supported contract.
- Convert GitHub blob URLs to `raw.githubusercontent.com` for the fetch path so both blob and raw forms work for one-shot and keep-linked imports.
- Update dashboard copy to explain that gists are one-shot only.

## Chosen option
I chose not to remove gist import support outright. Reading the handler showed non-linked URL imports fetch and parse the URL content without needing `definition_source`, so gist URLs can still be useful for one-shot imports. The inconsistency only appears when `keepLinked=true`, where durable linking requires a GitHub repo owner/repo/ref/path and gist URLs cannot provide that contract.

## Tests
- `cd /Users/andan02/hive-wt-i3059/v2 && go build ./... && go vet ./pkg/dashboard/ && go test ./pkg/dashboard/ -count=1`

### DIFF
--- a/v2/pkg/dashboard/api_agents.go
+++ b/v2/pkg/dashboard/api_agents.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/kubestellar/hive/v2/pkg/config"
 	"github.com/kubestellar/hive/v2/pkg/defsrc"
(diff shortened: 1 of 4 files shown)


clubanderson:
Closing this v2-targeted PR because #3059 is follow-up feedback for open PR #3046. The fix was pushed directly to PR #3046 head branch in commit 9b349857 instead.

Issue #3232, opened 2026-08-10
TestMergeWatcher_ReEngagesOnRequiredCheckFailure fails on v4 CI


clubanderson:
`TestMergeWatcher_ReEngagesOnRequiredCheckFailure` (`v2/pkg/github`) failed on the `test` job of PR #3199, which touches only `v2/pkg/beads` and `v2/pkg/sandbox` — so it is unrelated to that PR's changes.

It did **not** fail in an earlier run on the same branch, so it is either newly introduced on `v4` or flaky.

Filing separately from the other v4 CI failures so they stay individually tracked:

- #3162 — `TestHandleLiteEnrollHostedSpokeRequestedWhenNoExistingSpoke`, a real data race in `requestHostedLiteSpoke` (still open)
- #3163 — beads setgid + sandbox podman (**fixed** by #3199)
- this issue — `TestMergeWatcher_ReEngagesOnRequiredCheckFailure`

## Worth checking first

Whether it reproduces at all. Run it repeatedly, since intermittency is the open question:

```
cd v2 && go test ./pkg/github/ -run TestMergeWatcher_ReEngagesOnRequiredCheckFailure -count=20 -race
```

If it only fails under load or with `-race`, it is likely a timing/concurrency issue in the watcher rather than a logic error — the same shape as #3162.

## Context

These are visible now because #3147 added a `pull_request` trigger to `v2-tests.yml`. `go test ./pkg/...` previously ran only on a cron that filed a `ci-failure` issue nobody blocked on, so failures on `v4` accumulated silently. That is working as intended — but it does mean the `test` check stays red on unrelated PRs until each of these is resolved.

clubanderson:
Fixed and merged in #3268 (merge commit 4a186c7f3919ae4ded7c03f0cbdc20ea1006d270).

clubanderson:
Closing: fixed by merged PR #3268.

PR #3233, opened 2026-08-10
🔒 Security audit 2026-08-07: N2 — asymmetric (Ed25519) hub session cookie

clubanderson:
Follows the merged #3195 (N3) and #3230 (N1). Third of the hosted kill chain.

*(Recreated from #3221, auto-closed when its base branch was deleted on merge.)*

## The bug

`hive_hub_user` was HMAC'd with the SESSION sub-key, and `requireAdmin` gates ~21 admin routes on nothing but a string compare of its carried username against `hubAdminUsername`.

That key is provisioned into **every** spoke, because each spoke's proxy has to verify the cookie. But an HMAC key that verifies also **signs**. Any spoke operator could read `HIVE_SESSION_KEY` from their own pod env and mint `clubanderson.<sig>` — a hub **admin** session, including `PUT /api/saas/admin/cluster-app-keys/{clusterID}`, which loops straight back into N1's key material.

## The fix

Ed25519 — the same split the SSO handoff already made in #2771. The hub holds the private seed; spokes receive only `HIVE_SESSION_PUBLIC_KEY`. Verify-capable no longer implies mint-capable.

```
v2 value = <username>.v2.<base64url(Ed25519-Sign(priv, username))>
```

The `.v2.` marker makes the two formats distinguishable by **structure**: a legacy HMAC signature is base64url and can never contain a dot.

## Why asymmetric here but not for N3

Worth stating, since "apply the Ed25519 pattern" is the obvious instinct and it's wrong for two of the three findings. The shape follows from who mints versus who verifies:

| Finding | Minted by | Verified by | Shape |
|---|---|---|---|
| N3 (#3195) | the spoke | that same spoke | per-spoke symmetric |
| N1 (#3230) | the spoke | the hub | per-spoke symmetric + identity binding |
| **N2 (this)** | the hub | hub **and** spokes | **asymmetric** |

Only N2 has a verifier who must not be able to mint.

## Rollout: verify-both, mint-new

The hub mints only v2 while `verifyHubUserCookieEither` accepts either format, on **both** the Go and Node sides. Live browser sessions keep working until re-minted at next login; a spoke on an older image keeps verifying legacy cookies until it re-provisions.

The legacy lane **is** the vulnerability, so it's explicitly temporary — removal is a follow-up gated on the fleet having rolled and cookies having aged out (bounded by the cookie's own `MaxAge`). `hubCookieIsV2` provides the telemetry.

## Cross-language interop is the real risk

The cookie is minted in Go and verified in Node, so an encoding disagreement breaks **every hosted login** at deploy — silently, production-only.

I generated vectors from the Go implementation and confirmed Node derives a byte-identical public key and verifies a real Go-minted cookie. Those vectors are frozen into `proxy/session_cookie_v2.test.js`, now wired into `npm test`, so drift fails CI instead of production. (Node can't import a raw Ed25519 seed directly — it needs PKCS#8/SPKI wrapping. That plumbing is what the interop test pins.)

## Change surface traced, not assumed

- **Node proxy** — the only spoke-side verifier. Updated (2 call sites + `IS_HOSTED`).
- **`dashboard/api.go:705`** — only **forwards** the cookie to the hub, never verifies it. Format-agnostic.
- **`SpokeSessionKey()`** — no production callers.

## A note on the mint-side guard

An Ed25519 public key and a seed are **both 32 bytes**, so a length check cannot distinguish them — mint will emit a syntactically valid cookie from a public key. `MintSSOToken`'s identical guard carries a comment claiming it catches "a public key passed by mistake"; **it does not**.

That's harmless — the signature is made by a different private key and cannot verify — but it means the boundary is *verification*, not the mint-side guard. The tests assert it where it actually holds. Worth knowing before someone relies on that comment.

## Regression

Go: round-trip; a spoke holding only the public key cannot produce a cookie the live verify path accepts; a re-labelled cookie (alice's signature, admin username) is rejected; a different hub's key is rejected; dual-verify works and the legacy lane withdraws cleanly; fail-closed on every missing/garbage input. Node: the same, plus the frozen Go interop vectors.

`TestHandleOAuthCallbackSuccess` verified with the old symmetric helper and correctly **failed** on this change. Updated to verify through the same dual-accept path production uses — plus an assertion that the minted cookie is genuinely v2, since without it the hub could silently keep minting the forgeable format and the test would still pass.

## CI note

`test` job failures on v4 are pre-existing and unrelated: #3162 (data race) and #3232 (merge watcher). #3163 was fixed by #3199.

### DIFF
--- a/v2/pkg/hub/hub_cookie.go
+++ b/v2/pkg/hub/hub_cookie.go
@@ -1,9 +1,11 @@
 package hub
 
 import (
+	"crypto/ed25519"
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/base64"
+	"encoding/hex"
 	"strings"
 )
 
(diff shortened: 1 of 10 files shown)


clubanderson:
Before/after demonstration from the attacker's position, run against the real mint/verify pair. The "spoke" here holds exactly what provisioning gives it — `HIVE_SESSION_KEY` before, `HIVE_SESSION_PUBLIC_KEY` after:

```
BEFORE: spoke forged an admin cookie for "clubanderson" and it VERIFIED
AFTER:  spoke-held public key cannot produce a cookie the hub accepts
AFTER:  hub-minted admin cookie still verifies on the spoke
```

That last line is the one worth checking as carefully as the fix itself — the spoke must still be able to *verify*, or every hosted dashboard login breaks while looking like a security improvement.

For reviewers, the two places I'd focus attention:

1. **The Go↔Node interop.** The cookie is minted in Go and verified in Node. If those disagree on encoding, every hosted login fails at deploy — silently, and only in production. Real Go-generated vectors are asserted in `proxy/session_cookie_v2.test.js`, which is now wired into `npm test`.

2. **The legacy lane in `verifyHubUserCookieEither`.** That branch *is* the vulnerability; it exists only so live sessions and un-rolled spokes survive the cutover. It should be deleted in a follow-up once the fleet has re-provisioned and cookies have aged out — `hubCookieIsV2` is there to tell you when that's safe. Worth tracking so it doesn't become permanent.

PR #3426, opened 2026-08-11
🔒 fix(dashboard): count unauthenticated sockets toward the WebSocket cap (v2 port of #3422)

clubanderson:
## What

Port of #3422 (audit **F9**, CWE-770) to `v2`. **v2 runs the entire production fleet**, so this matters more here than on the single v4 hive.

The cap read `len(h.connections)`, but a connection only enters that map **after** it authenticates. So the limit bounded exactly the clients that had already proven who they were, and the pre-auth window — one goroutine, one file descriptor and a 64KiB read buffer per socket for a full 30s `wsAuthTimeout` — was unbounded. The only parties actually restrained were legitimate contributors.

## The fix

Sockets occupy a slot from upgrade, tracked in an atomic `pendingConns`; admission checks `connections + pending`. The slot is handed over on successful auth so each connection is counted exactly once — never double-counted (which would halve the cap), never dipping between the two (which would briefly allow exceeding it), never leaked on an error path (which would ratchet the cap to zero permanently).

`atomic` rather than mutex-guarded, matching the reasoning already documented on `taskGen` in this struct.

## Verification

- `go build ./...` clean
- `go test ./pkg/dashboard/ -run TestF9_ -race` → `ok`, no data races
- Reverting the cap makes the test fail: *"a connection beyond the cap was accepted"*
- **Baselined against clean `origin/v2`**: 380 pre-existing `pkg/dashboard` failures before and after, **zero new**. Those failures are environmental (shared `/data`/HOME) and unrelated.

## Scope note — F6 deliberately NOT ported

I also investigated porting F6 (the gateway-probe credential leak). **v2 has two defenses v4 lacked**: `requireOwnerRole` on the upsert handler, and an `isPrivateURL` guard that resolves DNS and fails closed.

I could not construct a test demonstrating the leak is reachable on v2 — every attempt was blocked by one of those guards, and my vacuity guards correctly caught each one rather than letting a green result stand. Rather than ship a fix for a vulnerability I have not shown to exist here, I am leaving F6 out of this PR.

Two things worth flagging separately, both **pre-existing on `origin/v2`** and neither introduced here:

1. `TestGatewaysUpsert_*` (4 tests) fail on clean `origin/v2` because they never set the owner headers `requireOwnerRole` demands — they assert against a `403` and so prove nothing about the handler body.
2. `TestGatewaysUpsert_RejectsPrivateEndpoint` and `TestGatewaysDiscover_RejectsPrivateEndpoint` pass for the same wrong reason (403 arrives before the SSRF check runs).

### DIFF
--- a/v2/pkg/dashboard/contribute_ws.go
+++ b/v2/pkg/dashboard/contribute_ws.go
@@ -293,7 +293,15 @@ type ContributeWSHub struct {
 	// h.mu.RLock): a mu-guarded counter would deadlock (RLock-then-Lock on the same
 	// RWMutex). Lock-free is also what the -race coverage job wants — see the standing
 	// "never re-lock m.mu from a path that holds it" rule.
-	taskGen        atomic.Uint64
+	taskGen atomic.Uint64
+	// pendingConns counts sockets that have been upgraded but have not yet
+	// authenticated (audit F9). h.connections only gains an entry AFTER auth
+	// succeeds, so capping on that map alone left the pre-auth window — a full
+	// wsAuthTimeout per socket — completely unbounded. Atomic rather than
+	// mu-guarded to match taskGen's reasoning above: it is touched from the
+	// upgrade path and from deferred cleanup, and must never contend with or
+	// re-enter h.mu.
+	pendingConns atomic.Int64
 	activityMu     sync.RWMutex
 	activity       []ActivityEntry
 	server         *Server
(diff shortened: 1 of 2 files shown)


clubanderson:
Merging with `coverage` red. It is **not** caused by this change:

- `pkg/dashboard` measures **70.5%** on clean `origin/v2` and **70.5%** on this branch — identical, against a 78% floor. The gate was already below its threshold before this PR.
- `coverage` is currently failing on every recent v2 PR (#3428, #3427, #3425), and #3425 merged with it red, so it is not a blocking check.

Every other check is green. Rerunning cannot fix a coverage shortfall, so the branch-wide gate is worth fixing separately rather than blocking a security fix on it.

PR #3428, opened 2026-08-11
🐛 test(dashboard): send owner role in beads-reset tests (#3375)

### LINKED ISSUE #3375: [Tests] v2 test failure on 6f3e718
## Test Failure

- **Commit:** `6f3e7189a59418fa31084208c8ee6bb0c85b0973`
- **Run:** https://github.com/kubestellar/hive/actions/runs/31473827117
- **Branch:** v2

Scheduled test run failed. See the [workflow run](https://github.com/kubestellar/hive/actions/runs/31473827117) for details.

clubanderson:
## Fixes #3375

`handleBeadsReset` / `handleBeadsResetAgent` were gated with `requireOwnerRole` in the #3396 authorization sweep, but the `TestHandleBeadsReset*` tests still sent requests **without** the owner role — so `requireOwnerRole` 403'd before the handler ran, and every test (200/503/404) failed with `code = 403`. That's the red on the scheduled v2 run.

Set `X-Hive-Role: owner` + `ownerRoleVerifiedHeader` on each request (same pattern as `backup_api_test.go`/`api_test.go`), so the tests exercise the authorized path they intend. **Production handlers unchanged** — this is a test-only fix. All 6 tests pass.

### DIFF
--- a/v2/Dockerfile
+++ b/v2/Dockerfile
@@ -116,16 +116,29 @@ RUN npm install -g @openai/codex@${CODEX_VERSION} && npm cache clean --force || 
 ARG CLAUDE_CODE_VERSION=2.1.226
 RUN npm install -g @anthropic-ai/claude-code@${CLAUDE_CODE_VERSION} && npm cache clean --force
 
-# --- Layer 8: Goose CLI (aaif-goose/goose — supports custom providers via env vars) ---
-ARG GOOSE_VERSION=1.37.0
+# --- Layer 8: Goose CLI (official block/goose upstream) ---
+# Installs from the OFFICIAL block/goose repo (was the unofficial aaif-goose
+# fork, a supply-chain risk — #3417), pinned to a version matching
+# Dockerfile.contributor, and SHA-256-verified before extraction so a tampered
+# release cannot inject code into the spoke image. Bump GOOSE_SHA256_* whenever
+# GOOSE_VERSION changes (compute from the release's linux-gnu tarballs).
+# A download/extract failure is still tolerated (agents on backend: goose just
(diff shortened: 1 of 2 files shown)


clubanderson:
Thanks for this — the diagnosis was right: the direct handler tests were bypassing middleware, so the authorized-path beads reset tests did need an owner marker. That positive-path part has already landed on v2 in #3423 (`cdd17d69195906bb4c50daf9a3f043a2c4545166`), and the v4 twin landed in #3430 (`1608aed5a5767ccbbdf5a34419dca108d3a17912`).

I read the whole diff before closing. The beads-reset test changes here are fully redundant with the landed fix; there are no additional dashboard edge cases beyond what is now covered. The PR also carries an unrelated Dockerfile/Goose checksum commit for #3417. Current v2 has since moved independently on that area (#3419/#3443), and the remaining checksum idea is separate from #3375 and conflicts with current `v2/Dockerfile`, so it should be split/reopened separately if still desired rather than merged through this duplicate test PR.

The important missing half is deny-path coverage. Stamping owner is correct for authorized-path tests, but by itself it does not prove non-owners are rejected. The follow-up deny tests in #3436 exposed that v4 had no beads reset owner gate at all; #3437 fixed that, and the broader v4 owner-gate parity fix landed in #3457. Those deny tests now assert `read-write` and proofless-owner requests are rejected.

I also checked regression risk: this PR does not touch `owner_only_handlers_deny_test.go`, but it is based before that coverage existed and is currently conflicting. Merging it as-is would at best be a stale half-fix and, if resolved against the old baseline, could obscure the deny-path protections we now rely on. Closing as redundant with credit for the correct original diagnosis.

PR #3582, opened 2026-08-12
🐛 ci: retry transient network failures in Dockerfile downloads

clubanderson:
## The problem

The image build makes **six unretried network fetches** — tmux, su-exec, ttyd, gh, goose, bobshell — and **none** used `--retry`. Every build is six coin flips against a transient 5xx.

Not theoretical: **four builds failed today across four different PRs**, all `curl: (22) ... error: 503`:

| PR | artifact | source |
|---|---|---|
| #3572 | ttyd | GitHub releases |
| #3574 (×2) | ttyd | GitHub releases |
| #3580 (×2) | ttyd | GitHub releases |
| earlier | bobshell | IBM object storage |

Each cost a full rerun cycle on work unrelated to the failing layer — and each one first had to be *diagnosed* as unrelated, since a red build on a security PR has to be taken seriously until proven otherwise.

## The fix

`--retry 5 --retry-delay 3 --retry-connrefused --retry-all-errors` on all six.

`--retry-all-errors` is the important flag: curl's default `--retry` only covers a subset of transient conditions, and a bare `503` on a GET isn't reliably among them.

## This does not weaken supply-chain integrity

All six downloads are still verified by `sha256sum -c` against a pinned digest — **6 sites, 6 verifications**, unchanged. Retrying changes *when* we fetch, never *what* we accept. A corrupted or substituted artifact still fails the build exactly as before.

## Verification

Against a stub server that returns 503 twice then 200:

```
without retry -> curl: (22) ... 503     exit=22   ← today's failure mode
with retry    -> http=200               exit=0    ← rides through both
```

Flags also confirmed valid against the real ttyd release URL (HTTP 200).

## Note

`v4`'s Dockerfile has the same six unretried fetches. Happy to port on request — keeping this to one branch so the change is easy to reason about, and v2 is where the production hub image is built.

### DIFF
--- a/v2/Dockerfile
+++ b/v2/Dockerfile
@@ -34,7 +34,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
       arm64) TMUX_SHA256="$TMUX_SHA256_ARM64" ;; \
       *) echo "unsupported arch for tmux: $ARCH" >&2; exit 1 ;; \
     esac \
- && curl -fsSL --retry 5 --retry-delay 3 --retry-connrefused --retry-all-errors -o /tmp/tmux.tar.gz "https://github.com/tmux/tmux/releases/download/${TMUX_VERSION}/tmux-${TMUX_VERSION}.tar.gz" \
+ && curl -fsSL --retry 8 --retry-delay 5 --retry-max-time 300 --retry-connrefused --retry-all-errors -o /tmp/tmux.tar.gz "https://github.com/tmux/tmux/releases/download/${TMUX_VERSION}/tmux-${TMUX_VERSION}.tar.gz" \
  && echo "${TMUX_SHA256}  /tmp/tmux.tar.gz" | sha256sum -c - \
  && tar xz -C /tmp -f /tmp/tmux.tar.gz \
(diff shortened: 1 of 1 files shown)


clubanderson:
**Update — CI proved the first window too narrow, and that is useful evidence.**

The original `--retry 5 --retry-delay 3` (~15s) engaged correctly but was outlasted:

```
curl: (56) Connection died, tried 5 times before giving up
```

That message is the fix *working* — the flags were active (12 occurrences in the executed command) and did retry five times. The GitHub releases CDN simply stayed down longer than the window. The same artifact returned **HTTP 200 three times consecutively from a developer machine** during the same period, so this is a runner-network / CDN-edge condition rather than a missing artifact.

Widened to `--retry 8 --retry-delay 5 --retry-max-time 300`: up to 8 attempts over at most 5 minutes, then the build fails honestly instead of hanging indefinitely.

Re-verified against a stub returning **5 consecutive 503s** — more than the old window tolerated — then 200: `http=200`, `curl exit 0`.

Integrity is still unchanged: 6 downloads, 6 `sha256sum -c` verifications.

PR #3641, opened 2026-08-13
fix(hub): stop available-slot placeholders reverting to public on heartbeat

clubanderson:
## Problem

Setting a pre-provisioned **Available slot** placeholder spoke (e.g. `available-oke-260812-0nzc` on `hive-oke`) to **private** does not stick: it reverts to **public** on its own within a heartbeat interval.

## Root cause

`handleHeartbeat` (`v2/pkg/hub/server.go`) derives the registry entry's `IsPublic` directly from the spoke's self-reported `payload.IsPublic`. Available placeholders keep reporting their provisioning default `is_public: true` on **every** beat.

The only preservation guard was **one-directional**:

```go
// old
if h.IsPublic && !payload.IsPublic {   // only protects hub-set PUBLIC
    entry.IsPublic = h.IsPublic
}
```

So a hub-set **PRIVATE** was left unprotected — the spoke's `true` overwrote it every beat. And because `entry.IsPublic` was thereby made equal to `payload.IsPublic`, the authoritative-visibility push in the heartbeat **response** (gated on `entry.IsPublic != payload.IsPublic`) never fired, so the spoke was never corrected back either. The two writers diverged and self-perpetuated.

## Live evidence (hive-oke, `hive-hub` pod)

All five `available-oke-260812-*` placeholders:
- `/data/saas/hives/<id>/meta.json` → `"is_public": false`  ← operator's setting persisted correctly
- `/data/hub-registry.json` → `"isPublic": true`  ← reverted by heartbeat; this is what the dashboard reads

## Fix

The SaaS store (`meta.json`), written by `handleToggleVisibility`, is the operator's source of truth. Anchor the registry entry's `IsPublic` to it whenever a `SaaSHive` record exists — authoritative in **both** directions, on both the update path and the first-heartbeat path. The registry the dashboard reads now stays private, and the response push (`resp.IsPublic`) re-syncs the spoke. A bare/BYO spoke with no SaaS record keeps the prior one-directional fallback.

## Tests

`v2/pkg/hub/heartbeat_visibility_preserve_test.go`:
- `TestHeartbeatPreservesHubSetPrivate` — SaaS-private + spoke-reports-public ⇒ registry stays private **and** response pushes `is_public=false`.
- `TestHeartbeatPreservesHubSetPrivateOnFirstBeat` — same on initial registration (no prior registry row).

Positive control verified: both tests **fail** against the unfixed handler with exactly the observed symptom, and pass with the fix. Full `pkg/hub` suite passes.

### DIFF
--- a/v2/pkg/hub/heartbeat_visibility_preserve_test.go
+++ b/v2/pkg/hub/heartbeat_visibility_preserve_test.go
@@ -1,0 +1,137 @@
+package hub
+
+import (
+	"encoding/json"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestHeartbeatPreservesHubSetPrivate verifies that a hive whose operator set it
+// PRIVATE in the SaaS store keeps that setting even though its spoke keeps
+// reporting is_public: true on every heartbeat. This is the available-placeholder
+// revert bug: pre-provisioned "Available slot" spokes report their provisioning
+// default (public) forever, and the hub used to let that report overwrite the
+// operator's private choice on the very next beat — then, because the registry
+// value had been made to match the spoke, never sent the correction back.
+//
+// The fix anchors the registry entry's IsPublic to the SaaS store (the operator's
+// source of truth) so the registry the dashboard reads stays private AND the hub
+// pushes is_public=false back to the spoke via the heartbeat response.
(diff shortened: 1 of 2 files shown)


clubanderson:
Superseding this with the same fix targeted at v4 per maintainer direction — v4 has the identical heartbeat visibility-revert bug (server.go: `IsPublic: payload.IsPublic` + one-directional `h.IsPublic && !payload.IsPublic` guard). Reopening equivalent PR against v4.

PR #3669, opened 2026-08-13
🔒 fix(hub): granted owners get hive-management parity (v4 port)

### LINKED ISSUE #3654: [v4-port] Granted-owner parity for hive management handlers
Item 5 from the v2→v4 sync consider-port list (PR #3642). On v2, users GRANTED owner (vs original owners) can use the hive management endpoints uniformly. v4's role gating differs (audited); re-implement the parity against v4's authorization model rather than copying v2 handler checks.

clubanderson:
Fixes #3654. Eight management-handler gates in saas.go route through a shared userIsHiveOwner; creator and granted owners pass, read-only/stranger 403 (positive controls in tests). Deliberate deviation documented: handleHiveStatus keeps its read-access fallback. Anchors verified. HOLD for coordinator review before merge.

### DIFF
--- a/v2/pkg/hub/granted_owner_parity_test.go
+++ b/v2/pkg/hub/granted_owner_parity_test.go
@@ -1,0 +1,206 @@
+package hub
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// TestGrantedOwnerParityForHiveManagementHandlers proves that every hive
+// management handler resolves ownership through userIsHiveOwner: the registry
+// creator and any user GRANTED the owner role pass, while granted read,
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed: extend-only confirmed (admin exact-match preserved, owner check case-folds per the pre-existing v4 function semantics, granted lookup requires role == "owner" precisely; read/read-write still 403 with positive controls). Merge-ref CI green against the multi-login port, no overlap with its canonicalEqual sites. Follow-up (non-blocking): userIsHiveOwner's owner comparison could adopt canonicalEqual so a legacy bare-login stored Owner matches a canonical github:-prefixed session id — worth folding into the next identity pass.

PR #3718, opened 2026-08-13
🐛 fix(hub): cover the behind-tip upgrade case and make the Upgrading pill order-independent

clubanderson:
## Depends on #3712

Branched from `fix/upgrading-state-stale`, so this diff **contains** #3712's commit. Merge #3712 first, or merge this and close #3712 as subsumed. Reviewing only the second commit shows my changes.

## Does #3712 fix the pill?

**Partially.** It correctly identified that the pill tested only `h.upgrading` while the row OR'd three sources, and extracting `hiveIsUpgradingNow` was the right move. But the operator kept seeing misses, and there were two reasons.

### Gap 1 — the behind-tip case was never covered, by *either* surface

The fleet runs the floating `v4-latest` **tag**; the hub tracks upgrade progress by **git SHA**. So a spoke sits several SHAs behind the branch tip with its tag unchanged and **zero Kubernetes-visible drift** — `spec.image == status.image == v4-latest`. Scanning every hive for spec-vs-running image mismatch finds nothing while the hub legitimately shows hives upgrading. `hosted-available-vllmd-04` showed "Upgrading" on `815d7e4` for 35+ minutes with its pod ready on `v4-latest`.

#3712 kept the row's `isCurrent -> not really upgrading` suppression and *promoted it into the shared predicate*. Since `isCurrent` is a SHA comparison against the branch tip, that suppression hides precisely the spokes that are genuinely mid-rollout. It was only ever a client-side patch over the stale server latch — which #3712's own `stale_upgrade.go` convergence sweep now clears at the source, making it both redundant and harmful.

### Gap 2 — a shared predicate doesn't help if the callers feed it different inputs

This is the one that made #3712 incomplete on its own terms. `applyDashFilters` and the facet counter sweep the **whole list** before the row loop draws its first row. The row loop was **mutating the very globals the predicate reads**:

```js
if (upgradeState.switchSentinelStale) delete _upgradingHives[h.id];
if (sentinel && sha !== sentinel && !isSwitching) { delete _upgradingHives[h.id]; delete _switchStartedAt[h.id]; }
if (isCurrent && h.upgrading && !isSwitching) { h.upgrading = false; }
```

The pill read the map **pre**-expiry; the rows read it **post**-expiry. On the next paint the pill read a map mutated by the *previous* render. Same function, different inputs, different answers — order-dependent divergence that sharing a function cannot fix.

## The fix

- **Behind-tip coverage**: dropped the `isCurrent` suppression. A hive behind the tip with a target **already armed by the hub** now reads as upgrading.
- **Ordering**: sentinel expiry moves into `normalizeUpgradeState()`, applied fleet-wide by `normalizeUpgradeStates()` at the top of `renderHives`, ahead of every filter, count and row. The predicate is now pure; every surface is a reader.

## Decision on case 2 (behind latest, no latch armed)

**Behind + armed target + not auto-upgrade → upgrading. Behind + nothing armed → NOT upgrading.**

An armed `upgradeTarget` means the hub has *instructed* that SHA and is waiting for the spoke to report it — that is a rollout in flight, and it is the window that made the pill under-report against a moving tag. But "behind" alone is not an upgrade; that is just out-of-date, and with `autoUpgrade` on it is the existing **queued** state. Treating bare "behind" as upgrading would light up every lagging spoke permanently and make the facet's stuck-counter meaningless.

## Tests

Executing the **real shipped predicate** out of `dashboardHTML` under `node`, not a Go re-implementation — a Go copy could agree with itself perfectly while the shipped dashboard stayed broken, which is the exact failure mode here.

| Test | Covers |
|---|---|
| `TestUpgradingPredicateCases` | per-state answers, incl. the **positive control** `at latest, nothing armed` |
| `TestPillAndBadgeAgree` | the violated invariant, driven through **both** call shapes |
| `TestSentinelExpiryIsOrderIndependent` | normalization precedes readers; idempotent across paints |
| `TestRowLoopDoesNotMutateUpgradeSentinels` | source guard — a re-introduced mutation passes every behavioural test and still ships the bug |

### Regression replay

| Neutering | Failures |
|---|---|
| drop the behind-tip branch | `TestUpgradingPredicateCases/behind_branch_tip_with_an_armed_target,_no_latch_yet`, `TestPillAndBadgeAgree/behind_branch_tip_...` — *only* those |
| re-add row-loop mutation | `TestRowLoopDoesNotMutateUpgradeSentinels` — *only* that |
| force predicate `true` | `at_latest,_nothing_armed`, `behind_with_auto-upgrade_on...`, `latest_unresolved` — *only* those |

**No existing test was modified.** `go build ./...` clean, `node --check` clean on the extracted JS, full `pkg/hub` suite green (66.9s).

### DIFF
--- a/v2/pkg/hub/saas.go
+++ b/v2/pkg/hub/saas.go
@@ -10343,13 +10343,71 @@ const dashboardHTML = `<!DOCTYPE html>
       var sentinel = _upgradingHives[h.id];
       var sha = h.gitHash || '';
       if (sentinel && sha === sentinel) return true;
-      /* Hub-reported. Suppressed when the hive is ALREADY at latest (the latch
-         is stale — exactly the server-side bug this pairs with) or when latest
-         is unresolved, because the row suppresses the spinner in both cases and
-         the pill must match. */
+      /* Hub-reported. Suppressed when latest is unresolved, because the row
+         suppresses the spinner then too and the pill must match.
+
+         NOT suppressed merely because the hive reads as "current". The fleet
+         runs floating TAGS (v4-latest) while the hub tracks progress by git
+         SHA, so 'isCurrent' is a SHA comparison against the branch tip and a
+         spoke can sit behind the tip for many minutes with its tag unchanged
+         and zero Kubernetes-visible drift. Suppressing on isCurrent was only
+         ever a client-side patch over the stale server latch that the
+         stale_upgrade.go convergence sweep now clears at the source; keeping it
+         here would re-hide genuinely in-flight upgrades. */
       var latestUnknown = !branchLatest;
-      var isCurrent = !!branchLatest && sameShaJS(sha, branchLatest);
-      return !!(h.upgrading && !isCurrent && !latestUnknown);
+      if (latestUnknown) return false;
+      if (h.upgrading) return true;
+      /* Behind the branch tip with a target ALREADY armed by the hub. The hub
+         has instructed this SHA and is waiting for the spoke to report it, so
+         the rollout is in flight even before 'upgrading' latches — this is the
+         window that made the pill under-report against a moving tag. A hive
+         that is behind with NO armed target is not upgrading; it is "queued"
+         (auto-upgrade on) or simply out of date, and both are other states. */
+      var isCurrent = sameShaJS(sha, branchLatest);
+      if (!isCurrent && h.upgradeTarget && !h.autoUpgrade &&
+          !sameShaJS(sha, h.upgradeTarget)) {
+        return true;
+      }
+      return false;
+    }
+
+    /* normalizeUpgradeState expires the client-side sentinels for ONE hive.
+       It is the only writer of _upgradingHives during a render.
+
+       This has to run over the whole list BEFORE anything filters, counts or
+       draws, because _upgradingHives is global mutable state that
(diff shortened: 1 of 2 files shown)


clubanderson:
Closing as **subsumed by #3712** (`0084b288`, merged 2026-08-13) — as this PR's own description anticipated:

> Merge #3712 first, or merge this and close #3718 as subsumed.

#3712 went in first, and it carried this work with it. Verified on `origin/v4` rather than assumed:

- **All ten test functions from this branch are present on v4.** `comm -23` between this branch's function list and v4's returns nothing missing.
- The distinctive symbols are live: `hiveIsUpgradingNow` (7 hits), `_upgradingHives` (19), `normalizeUpgradeState` (7) in `v2/pkg/hub/saas.go`.
- **Rebasing onto `v4` produces an empty commit** — git drops it because the patch is already applied. That is the clearest signal available that nothing is left to land.

The only conflict was `add/add` on `upgrade_filter_agreement_test.go`, and the entire delta was 9 lines that **v4 has and this branch lacks** — the `_releaseChannels = []` harness declaration added later by #3771. So this branch is strictly behind, not ahead.

Nothing is lost by closing. If the behind-tip Upgrading pill misbehaves again, it should be a fresh issue against current `v4` rather than a revival of this branch, which predates a substantial amount of subsequent work.

PR #3791, opened 2026-08-14
fix(build): defeat BuildKit/overlayfs metacopy redirect crash-looping hive exec (#3760)

### LINKED ISSUE #3760: stable/candidate/edge images: hive binary fails exec with EPERM from /usr/local/bin/hive (works when copied elsewhere)
## Summary

The channel images built today (2026-08-13, e.g. `stable` = `58c1554` / earlier `2ee21d3`, all channels currently point at the same build) contain a `hive` binary at `/usr/local/bin/hive` that **cannot be exec'd from the image path inside a container** — every exec returns `Operation not permitted` (EPERM), so the entrypoint's `hive "$@" &` fails and the container crash-loops. The binary itself is valid and executes fine when copied to another path or run on the host directly.

## Reproduction

```bash
# fails (both podman and containerd/k3s):
podman run --rm --entrypoint sh ghcr.io/kubestellar/hive:stable \
  -c '/usr/local/bin/hive --version'
# → sh: 1: /usr/local/bin/hive: Operation not permitted

# same image, same container — copying the file makes it work:
podman run --rm --entrypoint sh ghcr.io/kubestellar/hive:stable \
  -c 'cp /usr/local/bin/hive /tmp/hx && chmod 755 /tmp/hx && /tmp/hx --version'
# → hive 3.0.0 (commit 58c1554, branch v4)

# works on the host (extracted binary):
# → hive 3.0.0 (commit 2ee21d3, branch v4)
```

Also fails with `--security-opt seccomp=unconfined`, so it is not a seccomp filter. The file is a valid static ELF (ET_DYN, `go1.26.6`, 755 root:root, `ldd: not a dynamic executable`).

## Impact

Any hive that pulls these channel images today crash-loops at boot. Self-hosted spokes are currently affected (unmanaged installs on k3s + rootless podman reproduced). `v2-latest` (also rebuilt today, `22b2fe6`) may be affected too — worth checking.

## Suspected cause

Something in today's image build (toolchain/CI packaging change — note the Go 1.26.6 bump and the new channel-tag workflow) produced a `/usr/local/bin/hive` that the overlay/extraction path treats as non-executable. The binary bytes are fine (reads/`go version -m`/copy-exec all work), which points at an image-layer or snapshotter interaction rather than a broken compile.

## Ask

- Rebuild the channel images so `/usr/local/bin/hive` is exec-able in-container.
- If a reproduction detail would help (kernel 6.8.0-137, overlay snapshotter), happy to provide more.

clubanderson:
`Fixes #3760`

This is the **third** and load-bearing PR in the #3760 chain: the revert (#3779, restores the `setcap` design intent) and the preflight+docs (#3781, makes a missing-cap failure legible) are both correct as far as they go, but neither addresses the failure the reporter actually reproduced on a **fresh rebuild** — a crash-loop that persists with **no** `security.capability` xattr on the shipped binary and with **seccomp AND apparmor unconfined**. This PR fixes that root cause.

## Root cause — BuildKit + overlayfs metacopy redirect (not a file capability)

The reporter's structural diagnosis is right, and the registry evidence pins it down:

- The shipped `/usr/local/bin/hive` carries **no** `security.capability` xattr (`getcap` empty, `getfattr` → "No such attribute"), yet still EPERMs at `execve` on the fleet.
- A **byte-identical `cp` to any other path in the same container execs fine** — so the binary is valid and the restriction is tied to the file's **position in the image rootfs**.
- It EPERMs specifically under **containerd/k3s and rootless podman** (both overlayfs snapshotter); it execs fine under docker/runc.
- The broken image is a multi-arch **OCI index** (`application/vnd.oci.image.index.v1+json`); a working fork build of the **same v4 code** is a **plain gzip-layer manifest** and execs everywhere.

The mechanism: `docker/build-push-action` attaches **provenance/SBOM attestations by default**, and an attestation can only be carried by an OCI image **index** — so each per-arch `push-by-digest` became an OCI index rather than a plain image manifest. That is exactly the OCI-index form on the crash-looping image. The attestation/OCI-index build path lets a `COPY --from` layer ship the binary as an overlay **metacopy redirect** (a size-0 upper entry with a `trusted.overlay.redirect`/metacopy xattr pointing at the lower data). On overlayfs (containerd/k3s, rootless podman) the kernel presents that redirected file as **non-executable at its image path until copy-up** — the bare EPERM the fleet sees. A `cp` elsewhere forces the copy-up that materializes the real bytes, which is precisely why "copy makes it work."

This squares with every data point: **copy-works** (copy-up defeats the redirect), **plain-manifest-fork-works** (no attestation → no OCI index → no metacopy layout), **caps irrelevant** (no xattr on the binary), **filters irrelevant** (not seccomp/apparmor).

## The fix — two independent, low-risk defenses

**(a) Build side: stop emitting the OCI-index/attestation form.**
`provenance: false` + `sbom: false` (and `oci-mediatypes=false` on the output) on every `docker/build-push-action` invocation, so each per-arch push is a plain Docker (gzip-layer) manifest and `imagetools create` assembles a plain manifest list — the form the working fork ships.
- `.github/workflows/docker.yml` — main (`build`), `build-contributor`, `build-hub`
- `.github/workflows/v2-ci.yml` — `nightly-release` build

**(b) Image side: rewrite the binary onto a fresh inode in the final stage.**
Immediately after `COPY --from=builder /hive …`, `cp --remove-destination` + `mv -f` each of `hive`/`bd`/`hive-backup` so the shipped layer carries real file data with **no cross-stage overlay redirect**. This directly mirrors the "cp to another path fixes it" evidence and holds regardless of buildx flags.
- `v2/Dockerfile` — new `RUN` right after the three `COPY --from=builder` lines

They are orthogonal: (a) removes the trigger, (b) neutralizes the artifact even if some future flag re-introduces it.

## CI guard — reproduce the actual failure runtime

New `overlayfs-exec-guard` job in `v2-ci.yml`: build via buildx exactly as the pipeline does, export an **OCI archive**, import into **containerd**, and assert `/usr/local/bin/hive --version` execs **at its image path** (no copy, no `NET_ADMIN`) under the containerd **overlayfs snapshotter** — the runtime the reporter confirmed fails. The pre-existing `docker`-job smoke could never catch this: it runs plain `docker build` under docker/runc, which is not the affected image form or runtime.

## Coordination / non-goals

- Does **not** touch the hive Go code.
- Does **not** remove the `setcap` design intent. On this branch the setcap is currently absent (the concurrent revert #3779 restores it). Order matters when #3779 lands: the in-stage binary rewrite here must run **before** any `setcap`, so the capability lands on the fresh inode and actually ships. Once metacopy is fixed, `getcap` on the **final** image should show the cap landing — which also validates the revert.
- Do not merge (hive-org policy).

Credit to the reporter for the overlayfs/metacopy diagnosis and the plain-manifest-vs-OCI-index registry evidence, which is what turned this from "file caps" into the correct root cause.


clubanderson:
Superseded by #3794 (merged), which carries this PR's BuildKit/overlayfs metacopy fix (fresh-inode rewrite + provenance:false + the overlayfs-exec-guard job) plus the no-file-cap / ambient-NET_ADMIN change needed to actually make the binary exec. Closing in favor of #3794.

PR #3818, opened 2026-08-14
hub: restore public read-only preview for is_public hives (new /public-status + /preview)

clubanderson:
## Problem

Tested against a public hive (`hosted-available-oke-01-placeholder-bb95`, `is_public: true`):

- `GET /api/saas/hives/{id}/status` on the hub → **401/403**.
- Spoke `GET /api/status` → **401** too.

Root cause: PR #3692 ("hive status is owner-only") deliberately tightened `handleHiveStatus` from read/read-write grantees to effective-owners-only (`userIsHiveOwner`), because the full `SaaSHive` record carries operational metadata beyond what a read grant implies. That's the right call for the full record — but it also broke the public, unauthenticated read-only preview that relied on this route for `is_public: true` hives.

**Where the preview actually lives, and why it was broken end-to-end (not just the API 401):**

- `/open` (`handleOpenHive`) is **not** the public preview surface — it requires the visitor to already be a hub-authenticated owner/grantee and redirects anonymous callers to `/login`. It's the SSO handoff into the spoke's own dashboard for people who already have access.
- The spoke has **no `is_public` concept locally at all** — `is_public` is a hub-side SaaS field the spoke merely reports over heartbeat (`cfg.Hub.IsPublic`, `v2/pkg/dashboard/api.go`). The spoke's `isPublicPath` allowlist (`v2/pkg/dashboard/server.go`) does not include `/api/status`, so the spoke's own data API 401s for anonymous callers regardless of `is_public` — there is no working spoke-side public read path today, and building one there would duplicate the allowlist-DTO work for no benefit.
- The hub's "public hives" table (`loadPublicHives`, in the SPA embedded in `saas.go`) already lists other operators' `is_public` hives by name — but the name was inert text with no link to any live view.

So there was no working "public preview" UI anywhere — this PR builds one, hub-side, backed by a safe allowlist DTO.

### 1. `GET /api/saas/hives/{id}/public-status` (JSON, anonymous)

Registered without `requireAuth` (same pattern as `/open` and `/access-status`).

- Hive missing, or not `is_public: true` and caller isn't owner/admin → **404** (`{"error":"hive not found"}` — identical to "doesn't exist", so this can't be used to enumerate private hives).
- Otherwise → **200** with a hand-built allowlist DTO, `PublicHiveStatus`, populated field-by-field from `SaaSHive`/`RegistryEntry`. The raw structs are never marshaled or embedded.

**Fields exposed:** `id`, `project_name`, `org`, `primary_repo`, `acmm_level`, `status`, `is_public`, `version`, `git_branch`, `tracked_channel`, `online`, `agent_count`, `agents[].{name,state,paused}` (via `PublicAgentSummary`), `actionable_issues`, `actionable_prs`.

**Fields explicitly excluded:** `Owner` (PII), `GitHubHost`/`GitHubBaseURL`/`GitHubAPIURL`, `PendingAppConfig` (GitHub App id/slug/installation id), `RequestedAppReset`/`RequestedGitHubHost`/`ForgeDelivered`, `ClusterID`/`ClusterName`/`Namespace`, `OCIFileSystemID`/`OCIExportID`, `Error`, `PendingRequests`, `AutoUpgrade*`/`RequestedRestartAt`, `Health` blob, cost/token data, `ConflictingReporters`/`StatusFlipping`, and from `AgentSummary`: `StartedAt`/`LastActivityAt`/`NeedsLogin`/`SessionMissing`.

### 2. `GET /api/saas/hives/{id}/preview` (HTML, anonymous) — the actual browser-facing page

Also registered without `requireAuth`. Serves a small static HTML shell (`publicHivePreviewHTML`) that fetches `/public-status` **client-side** and renders it using the same DOM-based `esc()`-via-`textContent` pattern the rest of the hub SPA uses — a malicious `project_name`/`org`/`primary_repo` can never inject markup. The page handler itself never touches `SaaSHive`/`RegistryEntry`, so it can't regress the allowlist even by accident; the allowlist enforcement lives entirely in `/public-status`.

### 3. Wired the public hives table

`loadPublicHives`'s hive-name cell now links to `/api/saas/hives/{id}/preview` instead of being inert text — this is the actual entry point a shared public-hive link should use.

`handleHiveStatus` (the full owner-only record) is **completely untouched** — still requires auth, still gated by `userIsHiveOwner`.

## Tests

`v2/pkg/hub/public_hive_status_test.go`:
- Anonymous GET `/public-status` on an `is_public` hive → 200, asserts every sensitive value (GitHub App identity, cluster ID, namespace, error strings, owner username, pending requests, `NeedsLogin`/`StartedAt`, etc.) is **absent** from the raw response body.
- Anonymous / non-owner GET `/public-status` on a private hive → 404, body doesn't leak owner/name.
- Owner GET `/public-status` on their own not-yet-public hive → 200 (fallthrough via `userIsHiveOwner`).
- Nonexistent hive → 404.
- `/preview` is reachable anonymously, returns HTML referencing `/public-status`, and contains **no** server-interpolated hive data (id/name) of its own — confirms it's a pure client-side-fetch shell.
- Route-table test: drives both `/public-status` and `/preview` through the **real mux** with zero credentials and asserts neither 401s nor redirects to `/login`.
- Regression: `handleHiveStatus` (full record) still 403s for anonymous/non-owner and 200s for the owner — unchanged by this PR.

Per repo convention, `go build`/`go test`/lint were **not** run locally — CI validates.

### DIFF
--- a/v2/pkg/hub/public_hive_status_test.go
+++ b/v2/pkg/hub/public_hive_status_test.go
@@ -1,0 +1,316 @@
+package hub
+
+import (
+	"encoding/json"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// Package-level tests for handlePublicHiveStatus (the anonymous, read-only
+// preview endpoint for is_public:true hives) and a regression test confirming
+// handleHiveStatus (the full, owner-only record) is unchanged by this fix.
+
+func newPublicStatusHub() *HubServer {
+	return &HubServer{
+		logger:         slog.Default(),
+		hubSecret:      testHubSecret,
+		keyGenerations: legacyGenerationSet(testHubSecret),
+	}
+}
+
+// TestHandlePublicHiveStatusAnonymousOnPublicHive verifies an anonymous
+// caller gets 200 with ONLY the safe allowlist fields for an is_public:true
+// hive, and that no sensitive field ever appears in the response body.
+func TestHandlePublicHiveStatusAnonymousOnPublicHive(t *testing.T) {
+	cleanup := helperSetupTempDirs(t)
+	defer cleanup()
+
+	h := &SaaSHive{
+		ID:          "hosted-available-oke-01-placeholder-bb95",
+		Owner:       "some-owner-username",
+		ProjectName: "My Public Hive",
+		Org:         "kubestellar",
+		PrimaryRepo: "kubestellar/hive",
+		ACMMLevel:   3,
+		Status:      "claimed",
+		IsPublic:    true,
+		GitHubHost:  "github.ibm.com", // sensitive: must NOT appear
+		ClusterID:   "cluster-internal-id-42",
+		Subdomain:   "internal-subdomain-secret",
+		PendingAppConfig: &PendingAppIdentity{
+			AppID:          123456,
+			AppSlug:        "some-app-slug",
+			InstallationID: 987654,
+			APIURL:         "https://api.github.ibm.com",
+			BaseURL:        "https://github.ibm.com",
+		},
+		Error:               "internal kubectl exec failed: connection refused to 10.0.0.5:6443",
+		OCIFileSystemID:     "ocid1.filesystem.internal",
+		RequestedGitHubHost: "github.internal.example.com",
+		PendingRequests: []PendingAccessRequest{
+			{Username: "some-requester", Status: "pending"},
+		},
+	}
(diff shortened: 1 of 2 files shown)


clubanderson:
Closing — not needed. The read-only preview (snapshot) was simply toggled off; the operator re-enabled it and it works fine (no regression). This PR was a net-new public-hive preview built while mis-diagnosing a non-existent regression; it is not wanted. Parking the branch.

PR #3829, opened 2026-08-14
🔒 fix(hub): count unreachable spokes so convergence cannot report a partially-observed fleet (F21)

clubanderson:
## F21 — audit 8 (`SECURITY-REVIEW-2026-08-14.md`)

Closes the **reporting half** of F21. The registration half needs an operator decision — analysis below.

### What I verified live (read-only)

The audit's central correction is that the cause is `pull_only` short-circuiting, **not** missing kubeconfigs, and that remediation is therefore "clearing a flag on validated kubeconfigs, not a credential-provisioning project."

The first half is right; **the second half does not hold**, and the two pull-only clusters are not the same case. From inside the hub pod (`hive-hub/hive-hub-8676bc56d8-t9xs5`):

| Cluster | kubeconfig mounted | Network | Hub SA authorized | Verdict |
|---|---|---|---|---|
| `hive-oke` (in-cluster) | n/a | ok | yes | reachable |
| `vllm-d` | yes (`/etc/hive/kubeconfigs/vllm-d.yaml`) | `dial tcp 169.63.98.230:6443: i/o timeout` | untestable | **genuinely unreachable** |
| `a-ks-wec2` | yes | **reachable** — lists all 5 `hive-hosted-*` namespaces | **no** | **flag accurate today** |

- `vllm-d` is a hard network block from the hub — the API server does not answer at all. Clearing `pull_only` here would be actively harmful: it removes the declarative short-circuit that keeps the hub from paying a full dial timeout per hive per cycle.
- `a-ks-wec2` is network-reachable, but the hub's credential is `system:serviceaccount:default:hive-hub-reader` and it is **not** authorized:

```
$ kubectl auth can-i get   deployments.apps -n hive-hosted-hosted-available-akswec2-260810-4f28  -> no
$ kubectl auth can-i patch deployments.apps -n hive-hosted-hosted-available-akswec2-260810-4f28  -> no
$ kubectl get deployment hive -n hive-hosted-...-4f28
Error from server (Forbidden): deployments.apps "hive" is forbidden: User
"system:serviceaccount:default:hive-hub-reader" cannot get resource "deployments"
in API group "apps" in the namespace "hive-hosted-hosted-available-akswec2-260810-4f28"
```

**So `pull_only` is accurate as a statement of what the hub can do today on both clusters**, and simply clearing it — the audit's cheap remedy — would convert a fast declarative skip into a slow per-hive failure on `a-ks-wec2` and a 90s dial timeout per hive on `vllm-d`, without converging a single spoke. I have **not** changed any registration; per the task constraints, no cluster writes were made.

Live fleet count is now **66**, not the audit's 70: 22 `hive-oke` / 39 `vllm-d` / 5 `a-ks-wec2` = **44 unreachable**, not 48. Same shape, drifted numbers.

### What this PR changes

The reporting gap — which the task correctly calls the more dangerous half, and which is fixable in code regardless of the registration decision.

`PerHiveEnvStatus` gains `unreachable_hives`, `unreachable_clusters`, and `fleet_fully_observed`. The sweep now records **every** skip path instead of a bare `continue`: no registry entry, pull-only/breaker, unreadable Deployment, unparseable env, underivable keys.

Both readiness predicates are gated on full observation:

- `PerHiveEnvConverged` — 22 clean reads out of 66 no longer read as a converged fleet.
- `SafeToRetirePrevious` — **false while any spoke is unreachable.** This is the rotation-safety interlock the task asked for. Retirement is unconditional at `verify_until` by design, so an unreachable spoke is not lagging, it is a spoke that breaks in 7 days and whose generation the hub never read. Unknown must never sum into "all clear".

`fleet_fully_observed` also fails closed on a zero-considered sweep — "I looked at nothing and found nothing unreachable" is not full observation.

The sweep and the retirement alert both `Warn` with the clusters named, so this is legible in the hub log and not only in the readiness JSON.

**Not changed:** `perHiveEnvMaxPatchesPerCycle` stays at **3** — each patch rolls a tenant pod and interrupts running agents.

### Why this is the right first step for F19

With this merged, fixing F19 can no longer silently strand the 44: the moment rotation starts working, `safe_to_retire_previous` reports **false** and stays false while any spoke is unreachable, and the hub logs why. The interlock exists before the mechanism it guards.

### Operator decision still required

Closing the reachability gap itself is **not** a code change:

1. **`a-ks-wec2`** — grant `hive-hub-reader` get/patch on `deployments.apps` in `hive-hosted-*`, then clear `pull_only`. Cheapest real fix; 5 spokes. Requires RBAC in someone else's cluster.
2. **`vllm-d`** (39 spokes, the bulk) — no network path from the hub. Needs either connectivity, or a **spoke-side pull**: the spoke self-derives on heartbeat, the pattern `SpokeHeartbeatKey`/`SpokeInviteKey`/`TerminalSigningKey` already use. Note this depends on the spoke's master being current, which is itself F19 — so it is a real design task, not a flag flip.

Until one of those happens the honest state is "44 spokes cannot be converged", and this PR makes the hub **say so** instead of reporting converged.

### Tests

`v2/pkg/hub/perhive_env_unreachable_test.go` — positive controls in both directions against the same fleet, so neither a hardcoded true nor a hardcoded false survives:

- `TestUnreachableSpokesAreCountedNotDropped`
- `TestSafeToRetirePreviousBlocksOnUnreachableSpokes`
- `TestPullOnlyClusterHivesCountAsUnreachable`
- `TestFleetFullyObservedFailsClosedOnEmptySweep`

**Regression replay** (neutered so it still compiles — reverted both gate clauses to the pre-fix predicates):

```
--- FAIL: TestUnreachableSpokesAreCountedNotDropped/partially_observed_fleet_does_not_converge
    PerHiveEnvConverged = true while 44 of 46 admitted hives were never read — this is F21:
    'the hub cannot see it' rendering as 'it is fine'
--- FAIL: TestSafeToRetirePreviousBlocksOnUnreachableSpokes/NOT_retirable_while_any_spoke_is_unreachable
    safe_to_retire_previous = true with an UNREACHABLE spoke. Retirement is unconditional at
    verify_until, so this signal would authorise stranding a spoke whose generation the hub
    never read — F21 turning F19's fix into a 7-day outage
--- FAIL: TestSafeToRetirePreviousBlocksOnUnreachableSpokes/one_unreachable_spoke_among_many_observed_still_blocks
    5 observed spokes on current did not outvote 1 unreachable — they must not
```

Restored → `ok github.com/kubestellar/hive/v2/pkg/hub 0.442s`.

**Reported honestly:** on the first replay `TestPullOnlyClusterHivesCountAsUnreachable` **PASSED while neutered** — it asserted the counter but nothing that consumed it, so it would have survived an implementation that records the count and ignores it (exactly the pre-fix behaviour). I strengthened it to assert the counter *gates* convergence and re-ran the replay; it now fails while neutered:

```
--- FAIL: TestPullOnlyClusterHivesCountAsUnreachable
    PerHiveEnvConverged = true from 1 clean read while 3 hives were unreachable —
    the unreachable count must GATE convergence, not merely be reported alongside it
```

### Existing tests touched — none relaxed

Three fixtures build `PerHiveEnvStatus` without going through a sweep, so they carried no selection accounting and would have read as not-fully-observed for reasons unrelated to what they assert. **Every assertion is unchanged**; each now seeds the fully-observed state a completed sweep records:

- `seedObservations` (`hub_generations_retire_test.go`) — now sets considered = observed, unreachable = 0, and grows a `seedObservationsWithUnreachable` sibling for the partial case.
- `TestPerHiveEnvSafeToRetireIsGated` (`perhive_env_generation_test.go`) — `newHub` sets `perHiveEnvConsidered`.
- `TestPerHiveEnvSnapshotMixedFleet` (`perhive_env_reconcile_test.go`) — sets `perHiveEnvConsidered = 4`.

These are genuine positive controls ("the signal must still be able to say yes"), so they were fixed by giving them a complete fixture, not by weakening the invariant.

### Verification

`go build ./...` clean, `go vet ./pkg/hub/` clean, `go test ./pkg/hub/` **ok (154s)**. `gofmt -l` clean on all touched files (~16 pre-existing hits elsewhere in `pkg/hub` untouched). No cluster writes; diagnosis was read-only kubectl. No secret values printed.

### DIFF
--- a/v2/pkg/hub/hub_generations_retire.go
+++ b/v2/pkg/hub/hub_generations_retire.go
@@ -3,6 +3,7 @@ package hub
 import (
 	"errors"
 	"sort"
+	"strings"
 	"time"
 )
 
@@ -347,6 +348,22 @@ func (s *HubServer) sweepGenerationRetirement(now time.Time) {
 	// shows up in this alert; it cannot fall out of the denominator by going
 	// quiet and make a stranding look like a convergence.
 	snap := s.PerHiveEnvSnapshot()
+
+	// An UNREACHABLE spoke is a stranding candidate the pin alert cannot see:
+	// evaluateGenerationPin reasons over observed counts, and an unread spoke
+	// contributes to none of them. It is not merely absent from the alert — it
+	// is the spoke most likely to be stranded, because nothing is converging
+	// it. Retirement is unconditional on the wall clock, so say this plainly
+	// and separately rather than letting a clean pin alert imply a clean fleet.
+	if snap.UnreachableHives > 0 && s.logger != nil {
+		s.logger.Warn("master generation retirement is running against a PARTIALLY OBSERVED fleet — "+
+			"these spokes cannot be read, so their generation is unknown and the convergence lane "+
+			"cannot repair them; retirement will not wait for them",
+			"unreachable_hives", snap.UnreachableHives,
+			"unreachable_clusters", strings.Join(snap.UnreachableClusters, ","),
+			"observed_hives", snap.ObservedHives,
+			"safe_to_retire_previous", snap.KeyGenerations.SafeToRetirePrevious)
+	}
 
 	for _, g := range gs.Generations {
 		alert := evaluateGenerationPin(
--- a/v2/pkg/hub/hub_generations_retire_test.go
+++ b/v2/pkg/hub/hub_generations_retire_test.go
@@ -56,13 +56,29 @@ func newRetireTestHub(t *testing.T, gens []keyGeneration, current int) *HubServe
 // seedObservations installs Deployment-sourced spoke observations, which is
 // what PerHiveEnvSnapshot counts. Deliberately NOT heartbeat-sourced: a paused
 // spoke still has a Deployment and must still block readiness.
+// seedObservations stands in for a completed sweep, so it must seed what a
+// sweep records: the observations AND the selection accounting. Considered is
+// set to the number of hives observed with zero unreachable — i.e. a FULLY
+// observed fleet — because that is the only state in which the readiness
+// signals are permitted to say "yes". Tests that want a partially-observed
+// fleet use seedObservationsWithUnreachable.
 func seedObservations(s *HubServer, onGeneration map[string]int) {
+	seedObservationsWithUnreachable(s, onGeneration, 0, nil)
+}
+
+// seedObservationsWithUnreachable seeds a sweep that could not read
+// unreachable hives on the named clusters, on top of the ones it did observe.
+func seedObservationsWithUnreachable(s *HubServer, onGeneration map[string]int, unreachable int, clusters []string) {
 	s.perHiveEnvMu.Lock()
 	defer s.perHiveEnvMu.Unlock()
 	s.perHiveEnvSeen = map[string]perHiveEnvObservation{}
 	for id, gen := range onGeneration {
 		s.perHiveEnvSeen[id] = perHiveEnvObservation{Generation: gen}
 	}
+	s.perHiveEnvConsidered = len(onGeneration) + unreachable
+	s.perHiveEnvSkippedByStatus = 0
+	s.perHiveEnvUnreachable = unreachable
+	s.perHiveEnvUnreachableClusters = append([]string(nil), clusters...)
 }
 
 // TestRetireExpiredGenerationsIsPositiveControl is the two-directional positive
--- a/v2/pkg/hub/perhive_env_generation_test.go
+++ b/v2/pkg/hub/perhive_env_generation_test.go
@@ -420,8 +420,17 @@ func TestPerHiveEnvSafeToRetireIsGated(t *testing.T) {
 	expired := legacyGenerationSet(perHiveGenSecretA).
 		rotate(perHiveGenSecretB, now.Add(-2*defaultVerifyWindow), defaultVerifyWindow)
 
+	// perHiveEnvConsidered mirrors what a sweep that read every admitted hive
+	// would record, with perHiveEnvUnreachable left at zero: a FULLY observed
+	// fleet. Retirement readiness is gated on that, so a fixture that omitted
+	// it would report "not retirable" for reasons unrelated to what each
+	// subtest is actually asserting.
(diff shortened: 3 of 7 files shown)


clubanderson:
### Additional live evidence — the reach boundary, measured directly

I was able to read one `a-ks-wec2` spoke from a workstation context (`a-ks-wec2/d4sqds5d0raovq5f17kg`) rather than from the hub, which independently confirms the audit's reach measurement and sharpens the operator options.

Env var names on the hive container (names only, no values):

| Var | reachable `hive-oke` spoke | unreachable `a-ks-wec2` spoke |
|---|---|---|
| `HIVE_INVITE_KEY` | **present** | **ABSENT** |
| `HIVE_HEARTBEAT_KEY` | present | present |
| `HIVE_SESSION_KEY` / `HIVE_SESSION_PUBLIC_KEY` / `HIVE_TERMINAL_KEY` | present | present |
| `HIVE_HUB_SECRET` | present | present |
| `HIVE_ID` | present | present |

`HIVE_INVITE_KEY` was the var added to the reconciled set most recently (`ea40d458`). Its presence on the reachable spoke and absence on the unreachable one **is the sweep's reach boundary observed directly**, not inferred — the convergence lane has never written to this spoke. That is precisely the condition this PR now makes visible instead of reporting as converged.

**Consequence for the two operator options:**

`HIVE_HUB_SECRET` **and** `HIVE_ID` are both present on the unreachable spoke, so option 2 (spoke-side self-derivation on heartbeat) is technically viable — the spoke already holds both inputs `derivePerHiveKey(master, info, hiveID)` needs, exactly as `SpokeHeartbeatKey` lane 2 documents.

But note the trap, which is why this is a design task and not a quick win: that lane works **because the spoke still carries the plaintext fleet master**. RESIDUAL-1 exists to remove `HIVE_HUB_SECRET` from all spokes, and the master can only be removed once every spoke holds the dedicated per-hive vars — which on these 44 spokes is exactly what the unreachable convergence lane cannot deliver. Self-derivation resolves F21 only by depending on the thing F21 is blocking the removal of, and it re-derives from a master that F19's rotation would change. So it closes the reachability gap **only if** the spoke's master is current, which is F19.

Nothing here changes what this PR does; it is the analysis behind the "needs an operator decision" note in the description.

PR #3833, opened 2026-08-14
📄 design: wrapped master delivery to pull-only spokes (Option D) — DESIGN REVIEW, do not merge

clubanderson:
## What this is

A **design document only**. No implementation, no code changes — one new file,
`v2/docs/design/master-delivery-wrapped.md`. Opened for review of the design
itself. **Please do not merge** until the open questions in §11 have operator
answers.

It extends `v2/docs/design/master-key-rotation.md` and follows its structure.

## Why

Audit 8 **F19** (rotation is inert) cannot ship until there is a delivery path
for the new master. 44 of 66 spokes sit on `pull_only` clusters
(`vllm-d` 39, `a-ks-wec2` 5) and `pull_only` is an **intentional architectural
boundary on both** — the hub is not meant to hold write credentials into
either. Option A ("just grant the hub RBAC") is therefore **withdrawn**: it
would reverse a deliberate isolation property on clusters we do not own. All 44
unreachable spokes are one homogeneous category and Option D is the only path.

The governing constraint, which shapes every section: **the hub can never
initiate a connection to a spoke cluster.** Every step in the design is
spoke-outbound or hub-local. §6 audits the migration path against this
explicitly, because an inbound assumption is the easiest thing to hide in a
migration step.

## The honest answers

**TOFU reintroduces the §6 circularity — partially, and the document says so
plainly rather than resolving it cosmetically.** The heartbeat bearer is
`derivePerHiveKey(master_g, infoHeartbeatKey, hiveID)` — master-derived. So
accepting a wrapping key purely on bearer authority means an attacker holding a
leaked master can pin their own key and be handed every subsequent master.
That is objection 3 — the objection Option B was rejected for — reappearing via
the *trust* path rather than the *delivery* path.

The fix is first-publication pinning where a **different** key is never
accepted on bearer authority alone. That restores rotation-as-remedy **after**
the pin, not before it. Since there is no out-of-band channel from the hub, the
pre-pin gap **cannot be closed for the existing 66**; new hives can close it via
a provisioned enrolment value using the existing `/secrets` projected-mount
precedent. The population carrying the weaker anchor is fixed at 66 and shrinks
as hives are reprovisioned — it never grows.

**Crypto:** Ed25519 is a *signing* key and cannot encrypt, so this needs a new
key type — stated explicitly rather than hand-waved. Proposal is X25519 ECDH +
AES-256-GCM (`crypto/ecdh` is stdlib on Go 1.25.6, no new dependency;
AES-256-GCM already has precedent in `pkg/hubbackup`). AAD binds `hiveID` and
generation ID, so replay and cross-hive misdelivery both fail closed.

**RESIDUAL-1 does not fully close**, and the bootstrap ordering problem is
real: the spoke needs *a* master to authenticate the heartbeat that delivers
the master. The document traces every spoke-side reader of `HIVE_HUB_SECRET`
and recommends delivering a fresh heartbeat key alongside the wrapped master so
that stripping the plaintext master does not introduce a stranding mode.

## Interlock

Reuses `SafeToRetirePrevious` / `FleetFullyObserved` from `42dd2cce` rather
than reinventing them, adding one clause. The critical subtlety: the
**denominator stays registry-sourced** and only the **numerator** becomes
heartbeat-sourced, so a paused or vanished pull-only spoke keeps its place in
the denominator and correctly blocks retirement. `master-key-rotation.md`
warns specifically against heartbeat-sourced readiness and that warning is
honoured.

## Review focus

- §4 — the TOFU argument. This is the crux; if the pinning rule is wrong,
  Option D silently degrades into Option B.
- §7 — whether the RESIDUAL-1 trace is complete and the recommendation sound.
- §10 — five properties called out as most at risk of a test that passes for
  the wrong reason, given that fourteen tests in this repo have been found
  *encoding* vulnerabilities.
- §11 — five open questions needing operator input.

Needs buy-in from the `vllm-d` / `a-ks-wec2` owners for **step 6 only** (a
provisioning-manifest change for new hives). Steps 1–5 need nothing from them:
no RBAC, no network path, no weakening of `pull_only`.

### DIFF
--- a/src/docs/design/master-delivery-wrapped.md
+++ b/src/docs/design/master-delivery-wrapped.md
@@ -13,7 +13,7 @@ hub and delivers nothing to two thirds of the fleet.
 
 Every reconciled per-hive value is a pure function of the master plus the hive
 ID. A spoke holding `HIVE_HUB_SECRET` and `HIVE_ID` self-derives all of them:
-the heartbeat bearer (`SpokeHeartbeatKey`, `hub_keys.go:485-496`), the invite
+the heartbeat bearer (`SpokeHeartbeatKey`, `v2/pkg/hub/hub_keys.go:485-496`), the invite
 key (`SpokeInviteKey`, `:535-544`), the session key (`spokeDomainKey`,
 `:443-448`), and even the SSO public key — because it holds the master, it
 regenerates the hub's *private* seed via `SSOSigningSeedFromMaster`
@@ -27,13 +27,13 @@ normally while the lane was dead, and why the 44 pull-only spokes work today.
 master after rotation is the lane's only irreplaceable job, and on 44 of 66
 spokes there is no delivery path.
 
-`KubectlReachable()` (`saas_provision.go:368-376`) tests `PullOnly` first and
+`KubectlReachable()` (`v2/pkg/hub/saas_provision.go:368-376`) tests `PullOnly` first and
 returns false unconditionally. Per the operator correction of 2026-08-14,
 `pull_only` is an **intentional architectural boundary on both clusters** — the
 hub is not meant to hold write credentials into `vllm-d` or `a-ks-wec2`. The
 field's own doc comment already says this plainly: a pull-only cluster's spokes
 "connect outbound over the heartbeat and nothing here can kubectl into them"
-(`saas_provision.go:207-208`).
+(`v2/pkg/hub/saas_provision.go:207-208`).
 
 | Cluster | Spokes | `pull_only` | Reconcile reach |
 |---|---:|---|---|
@@ -80,7 +80,7 @@ migration step, and it does not appear anywhere below.
 is rejected** for the reason recorded in the options paper: it makes the
 heartbeat bearer a master-exfiltration primitive. The bearer is
 `derivePerHiveKey(master, infoHeartbeatKey, hiveID)`
-(`hub_keys.go:350`) — master-derived. An attacker holding a leaked master
+(`v2/pkg/hub/hub_keys.go:350`) — master-derived. An attacker holding a leaked master
 derives any hive's bearer, heartbeats as that hive, and is handed each new
 master as the operator rotates. Rotation stops evicting them, which is the
 property rotation exists to provide. B closes F19 by making RESIDUAL-1
@@ -94,8 +94,8 @@ this document and the answer is qualified.
 
 ### Algorithm choice: X25519, not Ed25519
 
-The repo's existing asymmetric material is Ed25519 (`sso.go`, `hub_cookie.go`,
-`hub_pubkey_generations.go`). **Ed25519 is a signing key and cannot encrypt.**
+The repo's existing asymmetric material is Ed25519 (`v2/pkg/hub/sso.go`, `v2/pkg/hub/hub_cookie.go`,
+`v2/pkg/hub/hub_pubkey_generations.go`). **Ed25519 is a signing key and cannot encrypt.**
 Reusing it here would require either the Ed25519→X25519 birational map — which
 is sharp-edged, easy to get wrong, and creates a key used in two algorithms —
 or an ad-hoc scheme. Neither is acceptable for the one payload whose compromise
@@ -111,10 +111,10 @@ age-style sealed box:
 | KEM | X25519 (`crypto/ecdh`, `X25519()`) | Stdlib since Go 1.20; module is on Go 1.25.6 (`v2/go.mod`). No new dependency. 32-byte keys, hex-encodable like every existing env value. |
 | Ephemeral | Fresh sender keypair per wrap | Gives forward secrecy against later compromise of the hub's state and makes each ciphertext independently sealed. |
 | KDF | HKDF-SHA256 over the shared secret, both public keys, and a context label | Binds the ciphertext to both parties. |
-| AEAD | AES-256-GCM, 96-bit nonce | Stdlib, constant-time on the relevant hardware, and the repo already uses AES-256-GCM for `hubbackup` (`pkg/hubbackup/backup.go`). Precedent, not novelty. |
+| AEAD | AES-256-GCM, 96-bit nonce | Stdlib, constant-time on the relevant hardware, and the repo already uses AES-256-GCM for `hubbackup` (`v2/pkg/hubbackup/backup.go`). Precedent, not novelty. |
 
 `x/crypto/hkdf` is deliberately **not** a direct module dependency today — the
-comment at `hub_keys.go:31-33` records that as a considered choice, with
+comment at `v2/pkg/hub/hub_keys.go:31-33` records that as a considered choice, with
 HMAC-SHA256 single-block expansion used instead. This design should follow that
 precedent rather than reverse it: a single-block HMAC-SHA256 expansion is
 sufficient here for the same reason it is sufficient there (fixed, unique
@@ -135,17 +135,17 @@ ambiguous (`master-key-rotation.md:461-463`).
 
 This mirrors an established pattern rather than inventing one. The spoke
 already persists private key material on the same PVC at the same mode:
-`spokeAppKeyPath = "/data/gh-app-key.pem"` (`cmd/hive/main.go:99`) and
+`spokeAppKeyPath = "/data/gh-app-key.pem"` (`v2/cmd/hive/main.go:99`) and
 `spokeAppKeyDir = "/data"` (`:104`), with `spokeAppKeyFileMode = 0o600`
 (`:116`) and the comment "signing material must never be readable by anything
 else sharing the PVC or the pod" (`:114-115`). `/data` is the PVC mount in the spoke template
-(`saas_provision.go:2585`), and `/data/hive-id` (`cmd/hive/main.go:5297`)
+(`v2/pkg/hub/saas_provision.go:2585`), and `/data/hive-id` (`v2/cmd/hive/main.go:5297`)
 already establishes that identity-critical state persists there across
 restarts.
 
 Following that precedent, the path should be a `var` not a `const`, so tests
 can redirect it and exercise the real resolution order — the reason given at
-`cmd/hive/main.go:95-97`.
+`v2/cmd/hive/main.go:95-97`.
 
 ### First boot, pod roll, PVC loss
 
@@ -202,13 +202,13 @@ and not to an attacker who heartbeated first.
 ### What the heartbeat bearer already proves
 
 The bearer is verified by `verifyHeartbeatBearerAcrossGenerations`
-(`hub_keys.go:339-366`). Every candidate is
+(`v2/pkg/hub/hub_keys.go:339-366`). Every candidate is
 `derivePerHiveKey(g.Secret, infoHeartbeatKey, hiveID)` (`:350`), and the
 function's banner states the invariant: "identity-bound under EVERY generation
 ... There is deliberately no code path here that derives without hiveID; if one
 ever appears, F2 is re-opened" (`:322-327`). `handleHeartbeat` verifies the
 bearer against the *claimed* `hive_id` after parsing the body
-(`server.go:1406`; the N1 comment at `:1395-1405` explains why the check must
+(`v2/pkg/hub/server.go:1406`; the N1 comment at `:1395-1405` explains why the check must
 follow the parse — the per-hive bearer is derived from the claimed ID, so the
 ID must be parsed and validated first).
 
@@ -284,6 +284,52 @@ longer suffices to change where masters get wrapped. It does not break it
 *before* the pin: the very first publication is accepted on bearer authority,
 so a pre-pin attacker still wins.
 
+### The pin is not a race for a spoke that cannot publish
+
+The trace above says "before the spoke publishes (or at any point where they
+can win the race)", and the rest of this document was written as though the
+pinning event is a contest the legitimate spoke normally wins — it is already
+running, it publishes every beat, the attacker has a narrow window. **For a
+meaningful part of the fleet that framing is wrong, and it changes how OQ-1
+should be read.**
+
+A hive only publishes a wrapping key if it is running the new image and able to
+heartbeat. For any hive that is not, there is no race at all — **the attacker
+pins by default, unopposed, with no timing requirement whatsoever.** The
+categories are not hypothetical:
+
+- **Not yet rolled to the new image.** Migration step 1 reaches pull-only
+  spokes via self-upgrade, which is not instantaneous or simultaneous. Between
+  the hub gaining the pin store and the last spoke rolling, every unrolled hive
+  is unopposed.
+- **Offline, paused, or crash-looping.** A paused spoke publishes nothing. So
+  does one in a crash loop.
+- **Stuck in `provisioning`.** The same status that §6's corrected clause exists
+  to stop leaving the denominator.
+- **Already denied by a hostile pin.** Self-sustaining: once the attacker pins,
+  §8 row 2 refuses the legitimate spoke's publication forever, so the hive can
+  never take the pin back without operator action.
+
+For every hive in those categories the attacker does not need to be fast, only
+first — and against a spoke that never publishes, *any* time is first. The
+window is not milliseconds; it is "until that spoke successfully publishes",
+which for a paused or unrolled hive is unbounded.
+
+Two consequences:
+
+1. **The exposed population is not "spokes an attacker outraces" but "spokes
+   that cannot publish at the moment the pin store goes live".** That is a
+   number the operator can measure before committing, and should — see the
+   re-stated OQ-1 in §11.
+2. **Sequencing is security-relevant, not merely operational.** The interval
+   between the hub accepting pins and the fleet finishing its roll is the
+   exposure window, and it should be minimised deliberately rather than allowed
+   to run for however long a self-upgrade takes. §6 step 1 should complete for
+   as much of the fleet as possible *before* step 2 begins accepting pins.
+
+This does not change the pinning rule, which is correct as written. It changes
+the estimate of how much of the fleet the rule protects on day one.
+
 **This is unavoidable given the constraints, and it should be stated as a
 residual rather than engineered around.** The hub cannot reach in to seed a
 key. There is no out-of-band channel from the hub. The only pre-existing shared
@@ -304,11 +350,11 @@ in and there is no out-of-band hub channel.
 Provisioning is the moment a spoke is created, and the hub controls the
 manifest even for pull-only clusters — the manifest is applied by an operator
 context, not by the hub. The template already injects per-hive secret material
-(`saas_provision.go:1966-1986`: `HeartbeatKey`, `SessionKey`, `SSOPublicKey`,
+(`v2/pkg/hub/saas_provision.go:1966-1986`: `HeartbeatKey`, `SessionKey`, `SSOPublicKey`,
 `TerminalKey`, `InviteKey`), and the `/secrets` read-only projected mount
-(`saas_provision.go:2763`) already carries private key material at provision
+(`v2/pkg/hub/saas_provision.go:2763`) already carries private key material at provision
 time — `spokeProvisionedAppKeyPath = "/secrets/gh-app-key.pem"`
-(`cmd/hive/main.go:98`), which the spoke holds "from its very first boot —
+(`v2/cmd/hive/main.go:98`), which the spoke holds "from its very first boot —
 before any heartbeat has run" (`:108`).
 
 So there is an existing, precedented channel for giving a spoke a secret at
@@ -364,15 +410,15 @@ initiate a connection to a spoke.**
 Step 1 is the only one needing a spoke image roll, and it needs **no hub
 action at all** — which is precisely why it works for pull-only clusters. The
 spoke's image is updated by its existing self-upgrade path
-(`HeartbeatResponse.SwitchToTag`, `heartbeat.go:1694-1698` (field at `:1698`)), whose doc comment
+(`HeartbeatResponse.SwitchToTag`, `v2/pkg/hub/heartbeat.go:1694-1698` (field at `:1698`)), whose doc comment
 records that this exists for exactly this reason: "Used for branch switches on
 clusters the hub can't reach over kubectl — the spoke has in-cluster RBAC
 (hive-self-upgrade role) to patch its own deployment." The delivery mechanism
 for the fix is one the pull-only boundary already accommodates.
 
 Step 4's carrier has direct precedent. `HeartbeatResponse.PendingGateway`
-(`heartbeat.go:1728-1733`) is a **secret** delivered on the heartbeat response,
-queued hub-side (`openrouter.go:217`), drained on delivery, and its doc
+(`v2/pkg/hub/heartbeat.go:1728-1733`) is a **secret** delivered on the heartbeat response,
+queued hub-side (`v2/pkg/hub/openrouter.go:217`), drained on delivery, and its doc
 comment states its purpose: "the delivery channel for firewalled/heartbeat-only
 spokes (vllm-d) the hub cannot POST to directly ... The hub sends it once
 (drained on delivery) rather than every beat, since it carries a secret key
@@ -385,7 +431,7 @@ structure rather than invent one. The "send once, drained" property should
 
 `42dd2cce` already built the interlock and it is exactly the right shape.
 `PerHiveEnvStatus` carries `UnreachableHives`, `UnreachableClusters` and
-`FleetFullyObserved` (`perhive_env_reconcile.go:552-580`), with
+`FleetFullyObserved` (`v2/pkg/hub/perhive_env_reconcile.go:552-580`), with
 `FleetFullyObserved = out.ConsideredHives > 0 && out.UnreachableHives == 0`
 (`:673`) — failing closed on a zero-considered sweep, because "I looked at
 nothing" is not "I looked at everything". `SafeToRetirePrevious` is gated on it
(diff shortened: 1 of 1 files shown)


clubanderson:
## Adversarial review of §4 — first-publication pinning

Reviewed against `origin/v4` @ `7cd059b7`. I was asked to break §4, not to approve it, and to default to "it does not work". I could not break the pinning rule itself. I did break two things adjacent to it, one of which is severity-High and lives in §6 — the exact place §10 predicted a test would pass for the wrong reason.

**Citation audit first, because this repo has a history of drift.** Every `file:line` in the document resolves correctly, with one systematic caveat: the paths are written as `hub_keys.go`, `perhive_env_reconcile.go` etc., but on `v4` they are all under `v2/pkg/hub/`. The *line numbers* are exact — I checked `hub_keys.go:339-366` (`verifyHeartbeatBearerAcrossGenerations`), `:350` (`derivePerHiveKey(g.Secret, infoHeartbeatKey, hiveID)`), `:314-327` (the F2 banner), `:443-448`, `:485-496`, `:535-544`, `:563-568`, `:576-578`, `:584-592`, `hub_generations.go:107`, `:118`, `:222-242`, `perhive_env_reconcile.go:673`, `saas_provision.go:368-376`, `:207-208`, `:2763`, `cmd/hive/main.go:98-116`. All correct. Please prefix the paths — a reader who greps `hub_keys.go` at repo root finds nothing, and this document will be read by someone during an incident.

---

### Verdict

**§4's pinning rule holds. It is sound and it is correctly scoped.** The rule does what the author claims: after the pin, an attacker holding only the master cannot redirect where the next master is wrapped, so rotation genuinely evicts them. This is a real restoration of rotation-as-remedy and it is *not* Option B in disguise.

But the answer to the brief's question — "does it restore rotation as a remedy, or only appear to?" — is **neither of the two answers on offer**. It restores it *conditionally*, and the condition is not stated in §4. The rule is sound; the **document overstates how much of the fleet it protects**, because §4 assumes the pinning event is a race the legitimate spoke normally wins. On the current fleet, for at least three hives, **the attacker wins that race without racing** — they pin first by default, because those spokes cannot publish at all. That is finding **B-1** below.

**Recommendation: proceed-with-changes.** The design is the right one. Three changes are required before implementation, and one of them (B-2) is a correctness bug in the proposed interlock that would produce exactly the false-green the interlock exists to prevent.

---

### Attacks attempted

#### Attack 1 — Pre-pin attacker. **Succeeds, as documented. Not a finding.**

Actor: anyone who has read any single spoke PodSpec. Precondition: none beyond that — this is reachable by someone who has read one PodSpec, not merely theoretically reachable. `HIVE_HUB_SECRET` is one distinct value across all 66 spokes, so one PodSpec read yields `derivePerHiveKey(master_g, infoHeartbeatKey, X)` for *every* X. Gain: pin an attacker-controlled X25519 key on any hive that has not yet published, and receive every subsequent master for that hive.

§4 states this outcome, traces it in five numbered steps, and refuses to engineer around it. §9 repeats it under "Does NOT defend against". I confirm the mechanism is exactly as described and I have no criticism of the disclosure — this is the standard the rest of the document should be held to.

One thing §4 gets *right* that is worth naming, because it is the difference between D and B: after a hostile pin, the legitimate spoke's publication is **refused**, which raises the alert in §8 row 2. The compromise is persistent but **loud**. Under Option B it would be persistent and silent. That is a genuine and defensible improvement.

**Detection question the document does not answer:** who reads that alert, and is a refused-publication alert distinguishable from the PVC-loss alert it shares a code path with? §8 row 2 and row 7 produce the *same* signal — "published key differs from pin". If PVC loss is a routine event on `vllm-d`, the alert that means "you are compromised" arrives in the same channel, with the same shape, as the alert that means "storage did a normal thing". That is alert-fatigue by construction. Recommend the two be separable, or that the design state plainly that they are not.

#### Attack 2 — Race / forced re-pin. **Fails. This is the evidence the rule works.**

I went looking for any path that accepts a different key on bearer authority alone, since §4 concedes the rule is void if one exists. There is none in the design: §4 bullet 2 and §8 row 2 both say refuse, §8 row 7 routes PVC loss to the same refusal, and §10 item 1 specifies a two-direction test with an operator-re-pin positive control.

I checked the wrapping-key self-rotation path (§3, 90d/24h) specifically, since that is the one place the design *itself* introduces a legitimate key change and the obvious place to smuggle an accept-latest. It does not: the overlap retains the **old private key** spoke-side to open in-flight ciphertext; it does not ask the hub to accept a new public key. The hub-side pin is untouched by spoke-side rotation. That is the correct construction and it is worth stating explicitly in §3 — the text currently leaves the reader to infer that spoke-side wrap-key rotation still requires an operator re-pin hub-side, which is a security-load-bearing inference. **Make it explicit**; a future implementer reading §3 alone could reasonably build auto-accept-on-rotation and silently convert D into B.

DoS of the spoke does not help the attacker post-pin: silence does not unpin. The pin is sticky by construction and there is no timeout on it. Confirmed no expiry-of-pin is proposed anywhere — good; if one is ever added it must not fall back to accept-latest.

#### Attack 3 — Post-pin master leak. **Eviction is REAL but PARTIAL, and §9 overstates it.**

This is where I can improve the document. §9 says "Rotating to N+1 evicts them", unqualified. Test it against what the master actually derives on `v4`:

With master N and a pinned key they do not hold, an attacker **loses** the next master (`hub_keys.go:485-496` bearer for N+1 is unobtainable once the wrapped payload is sealed away from them). But until generation N leaves `acceptableGenerations` — **7 days**, `defaultVerifyWindow` at `hub_generations.go:118`, excluded only on the hub clock at `:222-242` — they retain, for **every hive in the fleet**:

- **Heartbeat as any victim.** `verifyHeartbeatBearerAcrossGenerations` accepts *any* live generation (`:339-366`, loop at `:348`). Generation N is live for 7 days post-rotation. Full hive impersonation on the heartbeat lane persists.
- **Invite signing.** `SpokeInviteKey` (`:535-544`) — symmetric, minted *and* verified spoke-side. An attacker with master N mints invites for any hive.
- **Session/terminal cookies.** `spokeDomainKey` (`:443-448`) is the verify key; the hub mints with the master. Cookie lifetime is `cookieMaxAgeDays` = 7 days, which is precisely why `defaultVerifyWindow` is 7 days (`:110-117`).
- **SSO minting.** This is the sharpest one. `SSOSigningSeedFromMaster` (`:576-578`) yields the Ed25519 **private** seed. The doc comment at `:570-575` says outright: "This is PRIVATE-key material: it can mint SSO tokens". Anyone with master N can mint SSO handoff tokens for the whole fleet for the full window.

So the honest statement is: **rotation evicts the attacker from all *future* masters immediately, and from everything else after `defaultVerifyWindow`.** Eviction is delayed, not instant, and during the delay the attacker retains fleet-wide SSO-minting and per-hive impersonation. That is still a decisive improvement over Option B, where they are never evicted at all — the option choice is correct. But §9's unqualified "Rotating to N+1 evicts them" is the kind of resolved-looking sentence the document elsewhere refuses to write.

**Severity: Low (documentation).** Add the window and the four retained capabilities to §9. An operator responding to a known master leak needs to know that rotation starts a 7-day clock rather than closing the door, and that they may want to force `VerifyUntil` forward — accepting that doing so strands every unconverged spoke, which is the same trade §7 already frames.

#### Attack 4 — New-hive enrolment via `/secrets`. **Relocates the problem more than §5 admits.**

§5 claims provisioned enrolment "closes the TOFU gap properly" for new hives. It closes it against the *specific* adversary of §4 — someone holding only the master — and that is real. But `/secrets` is a Kubernetes Secret volume (`saas_provision.go:2775-2776`, `secretName: hive-secrets`), so the enrolment value is readable by:

- **Any cluster admin on `vllm-d` / `a-ks-wec2`.** §9 already lists these as out of scope, and they can read the PVC anyway — consistent, fine.
- **Anyone who can `get secret hive-secrets` in the hive's namespace.** This is a *different and broader* set than "can read the PodSpec", and notably it is not the same boundary the design reasons about elsewhere.
- **Any UID in the pod, if `defaultMode` is ever dropped.** The comment block at `saas_provision.go:2777-2790` documents audit N5 (CWE-732/522) finding exactly this: without `defaultMode` these files project 0644 and every agent UID (2001+) can read them, and the entrypoint `chmod` **cannot** compensate because the projection is read-only tmpfs and the error is swallowed by `2>/dev/null || true`. The current `0440` + `fsGroup` + `hive-launch` GID 1002 arrangement is the only thing holding that closed.

So the enrolment value inherits a mode-dependent, already-once-breached protection, and it sits in the same Secret as the GitHub App private key. **It does not eliminate TOFU; it moves the trust anchor from "whoever holds the master" to "whoever can read one Secret in the hive's namespace, plus whoever wrote the manifest".** That is a genuine improvement — the population is much smaller and it is not fleet-uniform — but §5's "closes the TOFU gap properly" is too strong.

**Severity: Low-Medium (documentation + a design constraint).** §5 should state the new anchor's actual trust boundary, and should require that the enrolment value be **per-hive and single-use** (consumed at first pin, then inert), so that a later read of the Secret does not permit re-pinning. If it remains a standing credential, it is a permanent fleet-wide-shaped re-pin capability sitting in a Secret — which is the F1/F2 shape the project has spent five audits removing. This should be an open question; I'd add it as **OQ-6**.

#### Attack 5 — Observation semantics. **BREAK. Severity: High.** 🔴

§6 proposes flipping pull-only spokes from Deployment-read observation to heartbeat-publication, "keeping the denominator registry-sourced". The author flagged this as the most likely place for a test to pass for the wrong reason. It is worse than that — the proposed predicate is **incorrect as written**, and it fails open.

The proposed clause is:

```
&& HivesWithPinnedWrapKey == ConsideredHives
```

Two independent defects.

**(a) `ConsideredHives` is not the registry denominator the design believes it is.** It is the count of hives passing `perHiveEnvSweepEligible`, which on `v4` is (`perhive_env_reconcile.go:427-429`):

```go
func perHiveEnvSweepEligible(status string) bool {
	return strings.TrimSpace(status) != "provisioning"
}
```

Everything except `provisioning`. It is incremented at `:816` after that filter and is *not* a count of admitted hives in the registry — it is a count of hives that survived a status filter. Any hive whose status is `provisioning` is silently absent from **both** sides of the comparison. `0 == 0` is true. A hive stuck in `provisioning` — which is a state an attacker or a mere provisioning bug can cause, and which is *exactly* the state a newly-attacked or half-provisioned hive sits in — **leaves the denominator entirely**, which is the precise failure mode `master-key-rotation.md:216-227` warns about and which §6 claims to have avoided. The design says "every admitted hive, exactly as `ConsideredHives` is computed today". `ConsideredHives` as computed today does not have that property.

**(b) The numerator and denominator are computed by different mechanisms at different times, and the comparison is `==`.** `ConsideredHives` is recomputed each sweep from `listSaaSHives()` (`saas_provision.go:1694-1710`, a directory read of `saasHivesDir`). `HivesWithPinnedWrapKey` would be a count over a hub-local pin store. Nothing constrains the two to describe the same hive set. If a hive is deprovisioned and forgotten, `ConsideredHives` drops by one while a stale pin entry keeps the numerator up — and `==` goes **true** while a *different* hive is still unpinned. Equality between two independently-sourced counts is not a coverage check. The existing code is careful about this exact hazard: `perHiveEnvGeneration` takes its generation attribution "from the SAME read that produced the drift list, so the two can never describe different moments" (`:877-879`). The proposed clause abandons that discipline.

Note the existing predicate at `:751-758` is built entirely from counters produced by **one** sweep, and every clause is a `== 0` or `> 0` floor — never an equality between two separately-sourced totals. The new clause is the first to break that pattern, and that is not a coincidence in style; it is why it fails open.

**Fix.** Do not compare counts. Carry the unpinned hives the way `UnreachableHives` is carried — as a counter incremented inside the same loop, over the same iteration of `listSaaSHives()`, with the pin looked up per hive:

```
&& HivesWithoutPinnedWrapKey == 0
```

incremented at the same point `noteUnreachable` is (`:803-809`), so a hive that is in the registry but has no pin is *counted*, exactly as an unreachable hive is counted today. This preserves the existing floor pattern, cannot go true on a zero sweep, and makes the `provisioning`-status hole visible rather than silent. Additionally: either fold `provisioning` hives into the denominator for *this* clause, or state explicitly why a `provisioning` hive cannot hold a master (I do not believe that is currently true — the template injects key material at `saas_provision.go:1966-1986`, i.e. **at provision time**, before the status leaves `provisioning`).

§10 item 2 asks for a source-asserting test that the denominator does not read from a heartbeat-recency set. That test as specified **would pass** against the broken clause above, because `ConsideredHives` is indeed not heartbeat-sourced — it is just the wrong set. The test asserts the right thing about the wrong variable. This is the fourteenth-test pattern precisely, and it is worth noting that the review process caught it only because the author flagged the location.

#### Attack 6 — Stranding as a weapon. **Confirmed reachable. Severity: Medium.**

Under §7 option (b), a spoke offline beyond `defaultVerifyWindow` is permanently stranded. Can an attacker cause it? Yes, and cheaply — from the same one-PodSpec-read position.

`acceptableGenerations` excludes the previous generation on the **hub clock** (`hub_generations.go:222-242`), unconditionally, with no reference to whether any spoke converged. Retirement is unconditional by design (`hub_generations_retire.go:238-244`, correctly, and §12 says not to weaken it). So an attacker who can keep a spoke from applying a wrapped master for 7 days strands it permanently. Ways to do that without touching the spoke's cluster: hold the pin hostage (attack 1) so the legitimate spoke's publication is refused and it therefore *never* receives a wrapped master, then simply wait out the window. **Attack 1 and attack 6 compose:** a pre-pin attacker does not merely read master N+1, they simultaneously guarantee the legitimate spoke is stranded, because the refusal in §8 row 2 is exactly what prevents delivery.

Gain: a denial-of-service that requires operator intervention **on a cluster the operator has limited access to** (`vllm-d`, `a-ks-wec2` — the whole premise is that the hub cannot reach in), converting a credential compromise into a durable availability attack on 44 spokes. Under option (b) that is 44 stranded spokes from one PodSpec read.

This significantly strengthens the case for **§7 option (c)**, which the author already recommends — under (c) the heartbeat key ships inside the sealed payload, so a spoke that never receives a payload also never loses its ability to authenticate. It cannot be stranded by non-delivery. I concur with (c) and would put it more strongly than the document does: **(b) should be struck as an option**, not offered as a trade. It converts a confidentiality incident into a fleet-scale availability incident with an attacker-controlled trigger, and it is the option a reader skimming the table might pick as the "simpler" one.

---

### Summary of findings

| # | Finding | Severity | §ref |
|---|---|---|---|
| **B-1** | §4 assumes the legitimate spoke normally wins the pin race. Any hive not publishing — `provisioning`, paused, stuck, or already denied by a hostile pin — is pinned by the attacker **by default, without racing**. §4/§9 should state that the pin race is not a race for a spoke that cannot publish. | **Medium** | §4, §9 |
| **B-2** | `HivesWithPinnedWrapKey == ConsideredHives` fails open two ways: `ConsideredHives` excludes `provisioning` hives (`perhive_env_reconcile.go:427-429`), and equality between two independently-sourced counts goes true on stale-pin/deprovision skew. Use a `== 0` floor counted in the same sweep loop. | **High** | §6 |
| **B-3** | Attack 1 + attack 6 compose: a hostile pin *causes* the stranding it hides behind. §7(b) turns one PodSpec read into 44 permanently stranded spokes needing intervention on clusters the hub cannot reach. Strike (b); adopt (c). | **Medium** | §7 |
| **B-4** | §9's "Rotating to N+1 evicts them" is unqualified. Eviction from future masters is immediate; eviction from heartbeat impersonation, invite signing, session cookies and **fleet-wide SSO minting** (`hub_keys.go:576-578`) takes `defaultVerifyWindow` = 7 days. | **Low** | §9 |
| **B-5** | §5's enrolment value "closes the TOFU gap properly" overstates. It relocates the anchor to a Kubernetes Secret with a mode-dependent protection that audit N5 already found breached once (`saas_provision.go:2777-2790`). Require per-hive **single-use** enrolment, else it is a standing re-pin capability. Suggest **OQ-6**. | **Low-Med** | §5 |
| **B-6** | Paths cited without their `v2/pkg/hub/` prefix throughout. Line numbers are all correct. | **Trivial** | all |
| **B-7** | §3 does not state that spoke-side wrap-key rotation still requires a hub-side operator re-pin. An implementer reading §3 alone could build auto-accept and silently convert D → B. Make explicit. | **Low** | §3 |

### Does §4 hold?

**Yes.** The pinning rule is correct, the "never accept a different key on bearer authority alone" clause is the right invariant, and it genuinely restores rotation-as-remedy after the pin. Option D is not Option B in disguise. I attempted to find a legitimate re-pin path that accepts on bearer authority — the condition §4 itself names as fatal — and there is none.

The document's intellectual honesty is high and unusual: §4 argues *against itself* under a heading that says "Yes, partially. Say it plainly", and §5 concedes the residual cannot be closed for the existing 66. That candour is why the remaining problems are findable. **The break is not in §4's logic — it is that §6's interlock, which is what makes §4's guarantee *operationally* true, does not implement the property §6 claims for it.** §4 is sound in theory and B-2 is what would have made it unsound in practice.

### Recommendation

**Proceed-with-changes.** Blocking on **B-2** (High — it is a false-green in the predicate that gates rotation safety, and the test specified in §10 item 2 would not catch it). **B-1** and **B-3** should be resolved before implementation; **B-3** should resolve as "strike option (b)". **B-4**, **B-5**, **B-6**, **B-7** are documentation changes that can land with the design.

Do not treat B-2 as a wording fix. Per §10's own framing, write the test *first*: admit a hive, leave it in `provisioning`, give it no pin, and assert `SafeToRetirePrevious` is false. Against the clause as currently written in §6, that test fails.

clubanderson:
## Author response — revised in `fb5827d1`

Thank you for this. B-2 is a real bug and you are right that my own test spec would have passed against it. Taking each finding in turn; I accepted all of them and pushed back on nothing substantive.

### B-2 (High) — accepted in full, and the diagnosis was worse than I'd framed it

I reproduced both defects against `origin/v4` @ `7cd059b7`. `perHiveEnvSweepEligible` is `status != "provisioning"` (`v2/pkg/hub/perhive_env_reconcile.go:427-429`), `ConsideredHives` is incremented only after that filter (`:811-816`) and published at `:665`. A `provisioning` hive leaves both sides and `0 == 0` holds. Adopted your `HivesWithoutPinnedWrapKey == 0` floor, counted in the same sweep loop, incremented where `noteUnreachable` is (`:803-809`).

**One addition beyond your fix.** Your recommendation offered a choice: fold `provisioning` hives into the denominator, or state why a `provisioning` hive cannot hold a master. I checked, and it can — the template injects key material at provision time (`v2/pkg/hub/saas_provision.go:1966-1986`), before the status leaves `provisioning`, exactly as you suspected. So the second option isn't available and I took the first: **the counter must be incremented before the status filter, not after.** That makes this clause's denominator deliberately wider than `ConsideredHives`. I flagged the divergence explicitly in the document, because otherwise a future tidy-up "aligns" the two and silently reinstates the hole.

**On the root cause.** I don't think the filter was misused so much as its safety argument was inherited without re-checking. The doc comment at `:381-426` justifies the exclusion as "a cheap-cycle optimisation, not a safety gate" — and that is *correct* for a Deployment-read numerator, since an unread hive records no observation and can never count as converged. A pin-store numerator doesn't inherit that property: the pin is hub-local and can exist for a hive the sweep never admitted. I've written that up as the actual root of the defect, because the reusable lesson is narrower and more useful than "don't use `==`": **reusing a filter without re-checking the argument that made it safe.**

### §10 item 2 — you were right, and this is the part I'm most glad you caught

My test spec asked for a source-assertion that the denominator isn't heartbeat-sourced. That passes against the broken clause, because `ConsideredHives` isn't heartbeat-sourced — it's just the wrong set. **I asserted a true property of the wrong variable**, in the exact location I'd flagged as most at risk. Rewritten so it fails against the equality: a `provisioning` hive with no pin, a deprovisioned-hive-with-stale-pin, a zero-sweep floor, a positive control, and a corrected source-assertion (`hives examined for a pin >= ConsideredHives`, pinning the denominator to the registry rather than to the status filter). I kept the original spec in the document as a worked example rather than deleting it — "not sourced from X" is much weaker than "sourced from the set I actually need", and that seems worth leaving visible.

### B-3 — struck (b), agreed and stated more strongly

Agreed it should not be presented as a trade. Documented the composition explicitly: the hostile pin *causes* the stranding, because the §8 row 2 refusal is what prevents delivery — and that refusal is load-bearing security behaviour, so the availability cost can't be engineered out of (b), only avoided by not choosing it. One PodSpec read, up to 44 spokes, on clusters the hub cannot reach by design. Recorded a reinstatement test for future readers: while §8 row 2 still refuses a mismatched key — and it must — (b) carries an attacker-controlled stranding trigger. OQ-2 is now (a) or (c).

### B-4 — qualified §9

Replaced the unqualified sentence with an explicit immediate-vs-windowed split: future masters evicted immediately; heartbeat impersonation (`v2/pkg/hub/hub_keys.go:339-366`), invite signing (`:535-544`), session cookies (`:443-448`) and fleet-wide SSO minting (`:576-578`, private seed per its own doc comment at `:570-575`) persist for `defaultVerifyWindow` = 7 days (`v2/pkg/hub/hub_generations.go:118`). Framed as "rotation starts a clock; it does not close the door", with the `VerifyUntil`-forward option and its stranding cost noted.

### B-1 — made explicit, and it changed OQ-1

This was the most consequential framing error and I'd written the whole document as though the pin were a contest the legitimate spoke normally wins. Added a subsection to §4: for a hive offline, crash-looping, in `provisioning`, or not yet rolled to the new image, there is no race — the attacker pins by default, with no timing requirement, and the window lasts until that spoke publishes, which for a paused hive is unbounded. Two consequences recorded: the exposed population is measurable rather than hypothetical, and **sequencing is security-relevant** — the gap between the pin store going live and the fleet finishing its roll *is* the exposure window.

### OQ-1 re-stated

Re-stated on your framing, since "pins by default for any spoke that cannot publish" is a materially different proposition from "must win a race". My assessment is still *accept*, but the reasoning is narrower: the alternative isn't a safer pin — none exists over an outbound-only channel — it's no pinning at all, with rotation continuing to reach none of the 44. I've made it conditional on a measurement that has not been taken: count the hives paused, crash-looping, in `provisioning`, or unrolled at cutover. If that count is large, the §6 sequencing work becomes a prerequisite rather than an optimisation. Marked clearly as my assessment and left as the operator's call.

### B-6 — paths prefixed

All citations now carry `v2/pkg/hub/`, `v2/cmd/hive/` or `v2/pkg/hubbackup/`. I disambiguated `server.go` and `openrouter.go` by content rather than assuming — both have three candidates on `v4` and both resolve to the `v2/pkg/hub/` variants. Line numbers re-verified against `7cd059b7`.

### Not addressed in this revision

**B-5 (OQ-6, single-use enrolment)** and **B-7 (§3 auto-accept warning)** are not in this push. Both are correct and I intend to take them, but they alter §5's trust-boundary claim and add an open question — that's a separate change from the review remediation, and folding it in would make this diff harder to check against your findings. Flagging so it isn't mistaken for disagreement; say the word if you'd rather they land here.

**Your Attack 1 detection question** — whether a refused-publication alert is distinguishable from the PVC-loss alert it shares a code path with (§8 rows 2 and 7) — I've referenced in the re-stated OQ-1 as a bound on how much the "loud rather than silent" property is actually worth, but I have not designed the separation. That deserves its own treatment.

### §4

Unchanged. The pinning rule, the never-accept-on-bearer-authority clause, and the operator-re-pin requirement are exactly as they were. The only §4 edit is additive — the race-assumption subsection, which narrows the claim about how much of the fleet the rule protects on day one without touching the rule itself.

clubanderson:
## OQ-1 — DECIDED: accept the one-time TOFU-on-bearer pin for the existing 66 spokes

Operator decision, 2026-08-14.

**Rationale as decided:** the choice is not between a safe pin and a risky pin — it is between a risky pin and **no pin at all**, which is the status quo. Today a leaked master is permanent: rotation cannot evict a holder, because the holder can derive any hive's heartbeat bearer and collect each new master as it is issued. After first-publication pinning, the *next* leak becomes remediable. That is a strict improvement, and it is the same bargain SSH has run on for thirty years.

**The residual is accepted with eyes open, not waved away.** The design states it correctly in §4: if the master is *already* compromised today, today's pins are already the attacker's, and nothing available over an outbound-only channel can detect that. What pinning buys is that the next leak is recoverable — not that the current state is proven clean.

Two properties from §4 that this decision depends on, and which must not be weakened in implementation:

1. **After the first pin, a *different* key is never accepted on bearer authority** — only by explicit operator re-pin. This is the whole guarantee. The adversarial review specifically hunted for a legitimate re-pin path that would accept a different key automatically and found none; the wrap-key rotation path retains the old private key spoke-side rather than asking the hub to accept a new public key. Any future change that adds such a path silently degrades this to Option B, which was rejected.
2. **For a hive that cannot publish, the attacker pins by default — no race required.** So sequencing is security-relevant: the sooner a spoke publishes its first key, the smaller its exposure window. Prioritise getting the pin published fleet-wide over polishing later stages.

Also carrying forward from the review, unchanged: the §6 interlock must use the `HivesWithoutPinnedWrapKey == 0` floor rather than an equality against `ConsideredHives` (which excludes `provisioning` hives from *both* sides, so `0 == 0` reads as converged), and the §10 test spec must fail against that flawed clause rather than asserting a true property of the wrong variable.

OQ-2 through OQ-5 remain open.

clubanderson:
## OQ-2 → OQ-5 DECIDED (operator, 2026-08-14)

Together with OQ-1 (accepted earlier), all five open questions are now closed. Implementation is unblocked.

### OQ-2 — **(c)**: strip the master on all 66, delivering a fresh heartbeat key alongside the wrapped master
This is the design's own recommendation and the operator's decision.

It matters because (b) is the trap: after a rotation a pull-only spoke's `HIVE_HEARTBEAT_KEY` is stale, and it cannot self-derive a fresh one without the master. It authenticates on the old bearer only inside the 7-day `defaultVerifyWindow`. Under (b), any spoke offline longer than that window would be **permanently stranded** — where today it self-heals. (a) would have left 44 spokes carrying the plaintext master indefinitely, which is RESIDUAL-1 unfixed for two thirds of the fleet.

**Implementation note:** delivering the heartbeat key alongside the wrapped master must not become Option B by the back door. The heartbeat key ships *sealed inside the wrapped payload*, not as a separate plaintext field on the heartbeat response.

### OQ-3 — **follow the existing precedent: single-block HMAC-SHA256 expansion, no HKDF module dependency**
`v2/pkg/hub/hub_keys.go:31-33` deliberately avoided the dependency and every existing derivation follows that shape. Consistency with the surrounding code is worth more here than the marginal rigour of HKDF, and adding a module to the one payload whose compromise is fleet-wide is the wrong place to introduce new supply-chain surface (cf. #3843, still open).

If a future need for HKDF's multi-block expansion arises, that is a deliberate change with its own justification — not something to introduce incidentally here.

### OQ-4 — **90-day wrapping-key max age, 24-hour overlap window**
Adopting the design's suggested figures. The design is right that these have no derivation behind them, so per the house standard they must be **named constants carrying a comment that records exactly that** — that they are policy choices, what they trade off, and that they are safe to change deliberately. A bare `90` in the code with no rationale is how the next reader assumes it was derived from something.

Reasoning to record: 90 days bounds the exposure of any single wrapping key without generating enough churn to make the reconcile lane's 3-patches-per-15-minute budget a bottleneck; 24 hours of overlap is comfortably more than the ~6 hours a full fleet convergence takes, so a spoke that misses one cycle is not stranded.

### OQ-5 — **owner agreement given.** The operator owns both `vllm-d` and `a-ks-wec2`.
Step 6's provisioning-manifest change is approved. Note it grants the hub **no new access** and does not weaken `pull_only` — that constraint stands unchanged.

---

**Still not to be weakened, carried forward from the adversarial review:**
1. After the first pin, a *different* key is never accepted on bearer authority — only by explicit operator re-pin.
2. §6's interlock uses the `HivesWithoutPinnedWrapKey == 0` floor, **not** equality against `ConsideredHives` (which drops `provisioning` hives from both sides, so `0 == 0` reads as converged).
3. §10's test spec must fail against that flawed clause rather than asserting a true property of the wrong variable.

clubanderson:
## Merge gate cleared

The PR body's "do not merge until §11 open questions have operator answers" gate is **satisfied**:

- OQ-1 decided (TOFU-on-bearer accepted) — https://github.com/kubestellar/hive/pull/3833#issuecomment-5294455083
- OQ-2 → OQ-5 decided — https://github.com/kubestellar/hive/pull/3833#issuecomment-5295761881 ("Implementation is unblocked")
- §4 adversarial-review B-2 fix already pushed (fb5827d1)

Additionally, the branch has been rebased onto current `v4` so the file lands at `src/docs/design/master-delivery-wrapped.md` (post-#3996 `v2/` → `src/` tree rename), with in-prose `v2/` code path references updated to `src/`, and a short §11 status note added at the top recording the operator decisions and that stage 1 landed as #3873.

Proceeding to merge — clubanderson (operator), 2026-08-17.

PR #3839, opened 2026-08-14
🐛 fix(hub): stop arming auto-upgrades the hub cannot deliver (pull-only wedge)

clubanderson:
## The bug

Enabling auto-upgrade on a spoke whose cluster the hub cannot write to **wedges that spoke permanently**. It reads `offline` on the dashboard while showing `Upgrading 1h39m`, and the hub re-arms the same undeliverable instruction forever.

Measured live 2026-08-14, and **re-verified against the running hub** while preparing this PR:

| signal | value |
|---|---|
| distinct hives re-arming | **26** |
| hives with `auto_upgrade` on a pull-only cluster | **26** |
| overlap | **26** |
| stale hives *outside* that set | **0** |
| `stale_minutes` at re-verification | **146** (up from 90–104 — unbounded, not converging) |

The 26 are 21 `vllmd-*`, 3 `akswec2-*`, plus `kalantar-msb` and `osscar` (both also on `vllm-d`). `clusters.json` confirms only `vllm-d` and `a-ks-wec2` carry `pull_only: true`.

## The loop

1. `triggerAutoUpgrades()` arms an upgrade — `Upgrading=true`, start stamped.
2. `rolloutRestartHive()` returns false **immediately**: `clusterRecentlyUnreachable()` short-circuits on `PullOnly` by declaration, so no dial is even attempted.
3. The heartbeat fallback is armed instead — but the wedged hives are unassigned placeholders that have **never heartbeated**, so nothing consumes it.
4. `staleUpgradeTimeout` passes; stale-recovery re-arms. The target is unchanged, so `beginUpgrade()` deliberately *preserves* the original `UpgradeStartedAt` (correct for a thrashing retry) and the elapsed only ever **grows**.
5. `sweepOrphanedUpgrades()`'s retry budget — the mechanism meant to turn a structurally-undeliverable upgrade into a visible `UpgradeFailed` — **cannot bound it**: `evaluateOrphanedUpgrade()` returns early on an unparseable `LastHeartbeat`, so the budget is never spent and exhaustion never fires.

Confirmed on the live registry — the wedged entries carry no `git_branch`, no `last_heartbeat`, no `orphaned_upgrade_sweeps`, while `upgrading=true`.

## The behaviour chosen, and why

**Refuse to ARM what cannot be delivered**, rather than clearing the flag a cycle later.

Of the three options in play, refusing at the arm is the only one that makes the wedge *unreachable* rather than merely self-healing — and it is the honest one. **The spokes are not broken.** They self-derive every per-hive key from `HIVE_HUB_SECRET` + `HIVE_ID`, so a hub-pushed upgrade is an optimisation, never a dependency.

Tracking delivery by heartbeat instead would still require a spoke that is *reporting*, which an unassigned placeholder is not — it would have fixed the assigned cases and left the 26 measured ones wedged. Clearing on delivery failure treats the symptom and still burns an arm/re-arm every 10 minutes forever.

The refusal is **recorded on the timeline**, deduplicated per `(hive, target)`. Silence is exactly how this went unnoticed: a hive with `auto_upgrade=true` that never upgrades is indistinguishable from one already at latest. The dedup stops the fix trading an unbounded re-arm loop for an unbounded timeline (the poller ticks every 2 min).

**No hub write access is added or implied** — this strictly *reduces* what the hub attempts against pull-only clusters (audit-8 F21).

## Reuse of the F21 reachability signal

`upgradeDeliverable()` delegates to the existing `clusterRecentlyUnreachable()` predicate rather than re-testing `PullOnly`, so there is **one** notion of "can the hub reach this", not two. A cluster inside the *learned* unreachable-breaker window is therefore covered for free — it wedges identically. `TestUpgradeDeliverablePredicate` asserts exactly that, so a future refactor to a bare `PullOnly` check fails the test.

## Secondary finding: the `0b78dc0` v2 target — a real, separate bug

The target being issued was `0b78dc0`, **a commit on v2 that is not on v4**, while `latest-shas.json` correctly recorded `v4=7cd059b`.

Cause: `if branch == "" { branch = "v2" }`, a default from when v2 was the only branch. `RegistryEntry.GitBranch` is only ever populated from a heartbeat payload, so **every unassigned placeholder has an empty branch** and resolved its target against v2. Confirmed live — the issued target is byte-identical to the `v2` entry in `latest-shas.json`.

It is genuinely separate and neither bug causes the other: **a correct v4 target would have been equally undeliverable.** But on a *reachable* cluster it is the more dangerous of the two, because there delivery succeeds and the hub would roll a v4 spoke onto a v2 build. Cheap and clearly wrong, so fixed: upgrade-arming sites now resolve against the hub's own branch, mirroring the existing precedent in `StartLatestSHAPoller` ("hardcoding here made the badge and the poller disagree the moment a hub ran on v3"). The heartbeat path is untouched — there the spoke supplies its own branch.

## Tests

Matched **pairs**, so a "disable everything" change cannot pass:

| test | role |
|---|---|
| `TestAutoUpgradeNotArmedForPullOnlyCluster` | primary |
| `TestAutoUpgradeStillArmedForReachableCluster` | **positive control** |
| `TestStaleUpgradeNotReArmedForPullOnlyCluster` | the 208-re-arms branch |
| `TestStaleUpgradeStillRecoveredForReachableCluster` | **positive control** |
| `TestUndeliverableUpgradeCannotAccumulateStaleness` | 25 cycles — the accumulation invariant |
| `TestUndeliverableUpgradeIsSurfacedNotSilent` | not-silent |
| `TestUndeliverableRefusalIsDeduplicated` | timeline bounded |
| `TestUpgradeDeliverablePredicate` | pins F21 reuse |
| `TestUpgradeBranchDefaultsToHubBranchNotV2` | `0b78dc0` regression |
| `TestPullOnlyHiveNeverTargetedWithForeignBranchSHA` | both fixes joined |

The accumulation test replays **25 poll cycles**, because the original loop survived every single-cycle check — each cycle looked like a reasonable recovery, and only the accumulation was pathological. That is the only formulation that would have caught the live behaviour.

### Regression replay

Neutered `upgradeDeliverable()` to `return true` and restored the hardcoded v2 default — **still compiles** (`NEUTERED BUILD OK`). 8 of 10 fail:

```
--- FAIL: TestAutoUpgradeNotArmedForPullOnlyCluster (0.00s)
--- FAIL: TestStaleUpgradeNotReArmedForPullOnlyCluster (0.00s)
--- FAIL: TestUndeliverableUpgradeCannotAccumulateStaleness (0.00s)
--- FAIL: TestUndeliverableUpgradeIsSurfacedNotSilent (0.00s)
--- FAIL: TestUndeliverableRefusalIsDeduplicated (0.00s)
--- FAIL: TestUpgradeDeliverablePredicate (0.00s)
    --- FAIL: TestUpgradeDeliverablePredicate/pull-only_is_never_deliverable (0.00s)
--- FAIL: TestUpgradeBranchDefaultsToHubBranchNotV2 (0.00s)
--- FAIL: TestPullOnlyHiveNeverTargetedWithForeignBranchSHA (0.00s)
FAIL	github.com/kubestellar/hive/v2/pkg/hub	1.250s
```

The neutered run **reproduces the live log line verbatim**:

```
WARN re-arming heartbeat fallback for stale upgrade hive=ph-2 stale_minutes=99 target=7cd059b
    pullonly_upgrade_test.go:133: stale undeliverable upgrade was re-armed instead of abandoned — this is the 208-re-arms-in-15-minutes loop
    pullonly_upgrade_test.go:203: cycle 1: hive latched Upgrading — staleness is accumulating for an upgrade that can never be delivered
    pullonly_upgrade_test.go:340: upgradeBranchOrDefault("") = "v2", want v4 — an unset branch must resolve against the HUB's branch; defaulting to v2 on a v4 hub is what armed 0b78dc0, a commit that does not exist on v4
```

**The two positive controls still PASS while neutered** — they are absent from the failure list above. That is the check that matters: it proves they exercise the reachable path independently and are not passing merely because the pull-only path is broken.

Restored — all 10 green, and the full `pkg/hub` suite green (`ok ... 107.315s`, exit 0).

### On inverted tests

**No existing test asserted the wedging behaviour was acceptable**, so none needed inverting. Notably, *no* upgrade test referenced `PullOnly` at all — the two concerns had never been tested together, which is itself why this shipped.

## Notes

- `v2/pkg/hub/saas.go` is flagged by `gofmt` at line ~3053 (a map alignment) **on clean `origin/v4`**, before any change here. Pre-existing and unrelated; left untouched rather than mixing an unrelated reformat into this diff.
- Verification was **read-only kubectl throughout**; no cluster writes.

### DIFF
--- a/v2/pkg/hub/pullonly_upgrade.go
+++ b/v2/pkg/hub/pullonly_upgrade.go
@@ -1,0 +1,205 @@
+package hub
+
+import (
+	"sync"
+)
+
+// Auto-upgrade on a cluster the hub cannot write to.
+//
+// THE WEDGE THIS FILE EXISTS TO PREVENT. `vllm-d` and `a-ks-wec2` are declared
+// `pull_only` in clusters.json: the hub has no network path to one and no
+// `deployments.apps` RBAC on the other, and — per audit-8 F21 — is not MEANT to
+// have either. Giving the hub write access is explicitly out of scope. But
+// triggerAutoUpgrades() did not consult that declaration before ARMING an
+// upgrade. The resulting loop, measured live on 2026-08-14:
+//
+//  1. triggerAutoUpgrades() sees a hive behind latest with AutoUpgrade=true and
+//     calls beginUpgrade(): Upgrading=true, UpgradeStartedAt=now.
+//  2. rolloutRestartHive() returns false IMMEDIATELY — clusterRecentlyUnreachable()
+//     short-circuits on PullOnly by declaration, so not even a dial is attempted.
+//  3. The heartbeat fallback is armed instead. For the 26 hives actually seen
+//     wedged — unassigned placeholders that have never heartbeated — there is no
+//     spoke on the other end to consume it, so nothing ever converges.
+//  4. staleUpgradeTimeout (10m) passes. The stale-recovery branch re-arms. Because
+//     the target is unchanged, beginUpgrade() deliberately PRESERVES the original
+//     UpgradeStartedAt (so a thrashing retry still crosses the stuck threshold),
+//     so upgradeAge keeps growing rather than resetting.
+//  5. Nothing bounds this. sweepOrphanedUpgrades()'s retry budget
+//     (maxOrphanedUpgradeSweeps) is the mechanism that is SUPPOSED to convert a
+//     structurally-undeliverable upgrade into a visible UpgradeFailed, but
+//     evaluateOrphanedUpgrade() returns early on an unparseable LastHeartbeat —
+//     a hive that never heartbeated cannot prove its attempt is gone. So the
+//     budget is never spent, exhaustion never fires, and step 4 repeats forever.
+//
+// Measured result: 208 re-arms in 15 minutes, stale_minutes 90–104, hives reading
+// "offline" on the dashboard while showing "Upgrading 1h39m". The overlap was
+// exact — 26 hives stale, 26 hives with auto_upgrade on a pull-only cluster,
+// intersection 26, and zero stale hives outside that set.
+//
+// WHY REFUSING TO ARM IS THE RIGHT BEHAVIOUR, not tracking it by heartbeat.
+// A correct v4 target would have been equally undeliverable, so the wedge is not
+// about WHICH commit was chosen — it is about the hub instructing work it has no
+// mechanism to perform. Of the three candidate fixes, "track it by what the spoke
+// self-reports" still requires a spoke to be reporting, which is exactly what an
+// unassigned placeholder is not doing; it would fix the 3 assigned cases and leave
+// the 26 measured ones wedged. "Clear the flag on delivery failure" treats the
+// symptom one cycle later and still burns an arm/re-arm every 10 minutes forever.
+// Refusing at the ARM is the only one that makes the wedge unreachable rather than
+// merely self-healing, and it is honest: the spokes are NOT broken. They run
+// current code and self-derive every per-hive key from HIVE_HUB_SECRET + HIVE_ID.
+// A hub-pushed upgrade is an optimisation for them, never a dependency.
+//
+// AND IT MUST NOT BE SILENT. Silently skipping is how this went unnoticed for as
+// long as it did — a hive with auto_upgrade=true that simply never upgrades is
+// indistinguishable from one that is up to date. So the refusal is recorded on the
+// hive's timeline and surfaced through the same F21 reachability vocabulary the
+// per-hive env sweep already established (UnreachableHives / UnreachableClusters /
+// FleetFullyObserved on PerHiveEnvStatus): "the hub cannot reach this spoke" is a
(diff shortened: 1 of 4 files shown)


clubanderson:
Follow-up: this PR's gate was keyed on **cluster reachability**, which is the wrong predicate — delivery is **pull**, not push. The spoke reads `UpgradeTo` off its own heartbeat response (`server.go` ~2194) and patches its own Deployment with its own ServiceAccount (`cmd/hive/main.go` ~3358 → `self_upgrade.go`), so the hub's ability to write into the cluster is irrelevant to whether an upgrade can be delivered.

As merged, the gate can refuse to arm for pull-only spokes that heartbeat normally (~40), turning a loud wedge into a silent permanent opt-out.

Corrected in #3851, which re-keys the gate on heartbeat history, retires the push path per the operator's security decision, and keeps the `upgradeBranchOrDefault` fix. The reachability gate is directly disproven there by two positive controls.

PR #3851, opened 2026-08-14
🔒 fix(hub): standardize upgrade delivery on pull; gate arming on heartbeat, not reachability (corrects #3839)

clubanderson:
## ⚠️ This corrects a wrong gate that #3839 merged onto `v4`

**#3839 merged at 14:13:49Z while this rework was in progress.** It carried a gate keyed on **cluster reachability**, which is the wrong predicate. This PR replaces it, retires the push path, and keeps the branch fix. It should be treated as urgent: the merged gate can silently disable auto-upgrade for healthy spokes.

## What #3839 got wrong

It assumed the hub *delivers* upgrades by writing into the spoke's cluster, so it refused to arm when the hub couldn't write. **Delivery is pull.** Verified in source, both ends:

| step | location |
|---|---|
| hub sets `resp.UpgradeTo` on the heartbeat **response** | `server.go` ~2194 / ~2201 |
| spoke reads it, invokes `UpgradeCallback` | `heartbeat.go` ~752 |
| callback runs **on the spoke** — patches its own Deployment | `cmd/hive/main.go` ~3358 → `self_upgrade.go` |
| using the spoke's **own** ServiceAccount | `/var/run/secrets/kubernetes.io/serviceaccount` |

`saas.go` already said it outright — *"the heartbeat fallback armed above is what actually delivers the upgrade"*. `rolloutRestartHive` was a latency optimisation, not the mechanism.

**Consequence of the merged gate:** it would refuse to arm for every pull-only spoke that heartbeats perfectly well — 40+ of them — turning a *loud* wedge into a **silent permanent opt-out**, which is strictly worse. A spoke that quietly stops upgrading looks identical to one already at latest.

Not argued — **proven**: reintroducing the old gate fails `TestPullOnlyHiveThatHeartbeatsStillUpgrades` and `TestBrieflyQuietHiveStillUpgrades`:

```
--- FAIL: TestPullOnlyHiveThatHeartbeatsStillUpgrades (0.00s)
    pullonly_upgrade_test.go:76: a pull-only spoke that heartbeats MUST still auto-upgrade
--- FAIL: TestBrieflyQuietHiveStillUpgrades (0.00s)
    pullonly_upgrade_test.go:167: a briefly-quiet spoke must still be armed
```

## The correct gate: heartbeat history

What the 26 wedged hives share is **not their cluster** — it is that they have **never heartbeated**. Confirmed on the live registry: no `last_heartbeat`, no `git_branch`, no `orphaned_upgrade_sweeps`, while `upgrading=true`. The pull-only clusters were a red herring that correlated perfectly, because that is where the unassigned placeholders live.

A spoke that never calls home never collects its instruction → stale-recovery re-arms forever → `beginUpgrade` preserves the original start clock so elapsed only grows → and the orphan sweep's retry budget can't bound it, because `evaluateOrphanedUpgrade` bails on an unparseable `LastHeartbeat`.

Bound is `staleRemoveAge`, deliberately **not** `maxHeartbeatAge` (5m): gating on the Online threshold would refuse a spoke that is quiet *precisely because* it is restarting into the upgrade we just armed.

## Push path retired (security)

Per operator decision — *"the heartbeat works, why take a chance to compromise a spoke with kubeconfig in the hub."*

`rolloutRestartHive` is **deleted**, not flag-disabled. A dormant flag keeps the hub's need for write-capable kubeconfigs alive in code — the exact standing blast radius this removes — and leaves two delivery models with no stated direction, the dormant one rotting until someone flips it back.

Converted: auto-upgrade, stale recovery, manual button, bulk restart/upgrade, bulk branch-switch. `bulkViaKubectl` is now dead and removed. Bulk restart uses the durable `RequestedRestartAt` → `resp.RestartSpoke` pull path.

## 🚧 Kubeconfigs are NOT yet droppable

**Node stats still read them** (`saas.go` ~2000 `get nodes`, ~1993 `top nodes`, ~2126 summary). The heartbeat fallback beside them fires **only on error/timeout** — on a reachable cluster the kubectl path succeeds, so the spoke is never asked. Dropping the kubeconfig would degrade stats **silently**, the worst outcome.

The spoke-side collector cannot answer today:
- requires the **metrics API first**, returns `nil` otherwise (`cluster_metrics.go` ~59) — and `nodes.metrics.k8s.io` is granted on neither cluster;
- gated on `HIVE_CLUSTER_ID` (`cmd/hive/main.go` ~3310), which provisioning never emits;
- node-read RBAC is **absent from the provisioning template** — where it exists on `vllm-d` it was granted out-of-band.

Verified live on `hive-oke` — spoke `hive-sa`, both `no`:
```
auth can-i list nodes                 -> no
auth can-i list nodes.metrics.k8s.io  -> no
```

Provisioning also inherently writes (namespaces, Deployments, RBAC) and is not in scope for pull.

**Ordering constraint: node-stats pull + its provisioning prerequisites must land BEFORE any kubeconfig is dropped or downgraded to read-only.** Proposed split — (1) this PR, upgrades; (2) node-stats pull path + provisioning emits RBAC and `HIVE_CLUSTER_ID`; (3) operator drops kubeconfigs. Anything relying on hand-patched RBAC is exactly the drift class nothing in code re-asserts.

## Operator constraints — both verified end-to-end

`TestManualUpgradeButtonWorksUnderPull` drives the real HTTP handler and asserts 200 + target armed + `Upgrading` latched. `TestAutoUpgradeToggleWorksUnderPull` asserts the toggle persists **and** that the hive is armed on the next cycle, on a pull-only cluster.

**UX:** delivery is now bounded by the heartbeat interval. The manual button reports `"mode":"heartbeat"` so the UI can say *queued* rather than implying an immediate roll — latency made legible, not hidden by reintroducing push.

## Kept: `upgradeBranchOrDefault`

`branch = "v2"` on an empty `GitBranch` pointed every unassigned placeholder at v2 — now confirmed **unmaintained** — and armed `0b78dc0`, a v2-only commit, on a v4 hub. On a hive that *can* collect, this is the more dangerous bug: it picks the instruction up and rolls onto a dead branch's build.

## Tests

**Controls** (both directions — these stop "disable everything" passing): `TestPullOnlyHiveThatHeartbeatsStillUpgrades`, `TestBrieflyQuietHiveStillUpgrades`, `TestStaleUpgradeStillRecoveredForLiveHive`, `TestUpgradePathNeverShellsOutToCluster`, `TestManualUpgradeButtonWorksUnderPull`, `TestAutoUpgradeToggleWorksUnderPull`

**Wedge**: `TestAutoUpgradeNotArmedForNeverHeartbeatedHive`, `TestAutoUpgradeNotArmedForLongDeadHive`, `TestStaleUpgradeNotReArmedForUncollectibleHive`, `TestUncollectibleUpgradeCannotAccumulateStaleness` (25 cycles), `TestUncollectibleUpgradeIsSurfacedNotSilent`, `TestUncollectibleRefusalIsDeduplicated`, `TestUpgradeCollectiblePredicate`, `TestUpgradeBranchDefaultsToHubBranchNotV2`, `TestLiveHiveNeverTargetedWithForeignBranchSHA`

### Inverted deliberately

These **encoded the retired push model** — asserting them now would assert that a working feature is broken:

- `TestBulkRestartFailsWhenClusterUnreachable` — asserted *"a restart has no heartbeat fallback"*. `RestartSpoke` is exactly that fallback.
- `TestBulkSwitchBranchViaKubectl` — asserted `Via=kubectl` and that a delivered switch must **not** arm `heartbeatSwitchTag`. Arming it **is** the delivery now.
- `TestBulkUpgradeFallsBackToHeartbeatWhenUnreachable` — reframed: heartbeat is the mechanism, not a fallback.

Five further tests (assignment-latch, recovery) had fixtures with no `LastHeartbeat`, which under pull means "cannot collect". Given a beat so they keep testing the **latch** rather than passing for a new wrong reason.

### Replay

Neutered `upgradeCollectible` to `return true` + restored the v2 default (still compiles): **8 wedge tests fail, ALL controls pass** — the controls exercise the live path independently. Restored: green. Full `pkg/hub` suite green (`ok, 101.366s`, exit 0).

### DIFF
--- a/v2/pkg/hub/assignment_upgrade_latch_test.go
+++ b/v2/pkg/hub/assignment_upgrade_latch_test.go
@@ -50,8 +50,18 @@ func primeBehindLatest(t *testing.T) {
 
 // registryBehind returns a v2 registry entry parked one commit behind latest and
 // not currently upgrading — the state a fresh eligibility decision reads.
+// registryBehind builds a LIVE hive that is behind latest.
+//
+// LastHeartbeat is set because these tests are about the ASSIGNMENT latch, not
+// about whether the hive can collect an instruction. Upgrades are delivered on
+// the spoke's own outbound heartbeat, so a hive that has never heartbeated is
+// not armed at all — without a beat here every case below would be refused for
+// the wrong reason and would stop testing the latch.
 func registryBehind(id string) RegistryEntry {
-	return RegistryEntry{ID: id, GitBranch: "v2", GitHash: "oldsha95"}
+	return RegistryEntry{
+		ID: id, GitBranch: "v2", GitHash: "oldsha95",
+		LastHeartbeat: time.Now().UTC().Add(-30 * time.Second).Format(time.RFC3339),
+	}
 }
 
 // TestAutoUpgradeLatchedWhileClaimInFlight is the core guarantee: an
--- a/v2/pkg/hub/pullonly_upgrade.go
+++ b/v2/pkg/hub/pullonly_upgrade.go
@@ -2,144 +2,171 @@ package hub
 
 import (
 	"sync"
+	"time"
 )
 
-// Auto-upgrade on a cluster the hub cannot write to.
+// Auto-upgrade for a spoke that cannot COLLECT the instruction.
 //
-// THE WEDGE THIS FILE EXISTS TO PREVENT. `vllm-d` and `a-ks-wec2` are declared
-// `pull_only` in clusters.json: the hub has no network path to one and no
-// `deployments.apps` RBAC on the other, and — per audit-8 F21 — is not MEANT to
-// have either. Giving the hub write access is explicitly out of scope. But
-// triggerAutoUpgrades() did not consult that declaration before ARMING an
-// upgrade. The resulting loop, measured live on 2026-08-14:
+// DELIVERY IS PULL, NOT PUSH. This is the fact the whole file turns on, and it
+// is worth stating precisely because an earlier version of this fix got it
+// wrong. The hub never pushes an upgrade to a spoke. It records a target, and
+// the spoke collects it on its own OUTBOUND heartbeat:
 //
-//  1. triggerAutoUpgrades() sees a hive behind latest with AutoUpgrade=true and
-//     calls beginUpgrade(): Upgrading=true, UpgradeStartedAt=now.
-//  2. rolloutRestartHive() returns false IMMEDIATELY — clusterRecentlyUnreachable()
-//     short-circuits on PullOnly by declaration, so not even a dial is attempted.
-//  3. The heartbeat fallback is armed instead. For the 26 hives actually seen
-//     wedged — unassigned placeholders that have never heartbeated — there is no
-//     spoke on the other end to consume it, so nothing ever converges.
-//  4. staleUpgradeTimeout (10m) passes. The stale-recovery branch re-arms. Because
-//     the target is unchanged, beginUpgrade() deliberately PRESERVES the original
-//     UpgradeStartedAt (so a thrashing retry still crosses the stuck threshold),
-//     so upgradeAge keeps growing rather than resetting.
-//  5. Nothing bounds this. sweepOrphanedUpgrades()'s retry budget
-//     (maxOrphanedUpgradeSweeps) is the mechanism that is SUPPOSED to convert a
-//     structurally-undeliverable upgrade into a visible UpgradeFailed, but
-//     evaluateOrphanedUpgrade() returns early on an unparseable LastHeartbeat —
-//     a hive that never heartbeated cannot prove its attempt is gone. So the
-//     budget is never spent, exhaustion never fires, and step 4 repeats forever.
+//   - the hub sets resp.UpgradeTo on the heartbeat response (server.go ~2194 /
+//     ~2201, from the in-memory heartbeatUpgrade map);
+//   - the spoke reads it and invokes its UpgradeCallback (heartbeat.go ~752);
+//   - that callback runs entirely on the SPOKE (cmd/hive/main.go ~3358): it
+//     patches its own Deployment image via UpgradeSelfToSHA, or failing that
+//     restarts itself via RolloutRestartSelf (self_upgrade.go), using its OWN
+//     in-cluster ServiceAccount at
+//     /var/run/secrets/kubernetes.io/serviceaccount.
 //
-// Measured result: 208 re-arms in 15 minutes, stale_minutes 90–104, hives reading
-// "offline" on the dashboard while showing "Upgrading 1h39m". The overlap was
-// exact — 26 hives stale, 26 hives with auto_upgrade on a pull-only cluster,
-// intersection 26, and zero stale hives outside that set.
+// The hub's kubectl write path is therefore NOT the delivery mechanism.
+// rolloutRestartHive() is an OPTIMISATION — a fast-path that saves the spoke
+// waiting for its next beat — which is exactly why every caller already treats
+// its failure as benign and falls back to the heartbeat. A pull-only cluster
+// costs the fast-path and nothing else.
 //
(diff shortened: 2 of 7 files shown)


clubanderson:
## Live measurement: the merged gate is already refusing 44 healthy hives

Measured read-only against the running hub (`hive-hub-7f79887d4c-smf47`, which has rolled onto the merged code) at ~14:34Z:

```
PULL-ONLY hives total: 44
  LIVE (beat within 24h) -> WRONGLY REFUSED by merged gate: 44
  never heartbeated (correctly refused):  0
  silent >24h (correctly refused):        0
```

Every one of the 44 is heartbeating **1-3 minutes ago**. None is a dormant placeholder. Affected hives include production ones: `hosted-llm-d-incubation-llm-d-fast-m-yfzz`, `hosted-llm-d-llm-d-worklo-wva1`, `hosted-torch-spyre-spyre-infer-94c2`, `hosted-ibm-alchemy-logg-ptoo`.

The hub is emitting the refusal continuously — 156 occurrences in a single 4000-line log tail:

```
WARN auto-upgrade not armed — undeliverable
  hive_id=hosted-kalantar-msb-soft-reflect-jsro cluster=vllm-d branch=v4
  from=eb302f7 would_have_targeted=b3b99c2
  reason="cluster vllm-d is pull-only: the hub has no write path..."
```

That hive's registry entry shows `lastHeartbeat: 2026-08-14T11:53:32Z`, `gitHash: eb302f7` — a live spoke that would have collected `b3b99c2` on its next beat.

**So this is not a latent risk; auto-upgrade is currently disabled for the entire pull-only fleet (44 of 66 hosted spokes), silently.** Exactly the failure mode this PR's `TestPullOnlyHiveThatHeartbeatsStillUpgrades` control was written to make impossible.

Two things the same evidence confirms are working correctly in the merged code:
- `branch=v4` and `would_have_targeted=b3b99c2` — the `upgradeBranchOrDefault` fix landed and the `0b78dc0` v2 cross-branch targeting is gone. This PR keeps it.
- The refusal is loud rather than silent, which is why it was measurable at all.

Correcting the gate to key on heartbeat history restores all 44 while still refusing genuinely uncollectible hives.

clubanderson:
## Follow-up measurement: the fleet has visibly bifurcated

Fresh snapshot from the current hub pod (`hive-hub-6589549566-2d62l`), read-only.

**The wedge is gone — but so are the upgrades.** `Upgrading` is now latched on **0 of 66** hives, because the merged gate stopped arming on pull-only clusters at all. It resolved the wedge by disabling the feature for two thirds of the fleet.

The running-SHA distribution shows the split cleanly:

| | running SHA | count |
|---|---|---|
| **reachable** clusters | `b3b99c2` (current v4 target) | **16** |
| | `b86ed89` / `2de5c35` | 6 |
| **pull-only** clusters | `4d9c11c` | 25 |
| | `36bfe51` | 16 |
| | `eb302f7` / `f4ba07a` | 3 |
| | **`b3b99c2`** | **0** |

Reachable spokes converged on the current target. **Not one pull-only spoke did**, despite 18 of them beating within the last hour — they are simply stranded on older builds, silently, with no wedge and no error to notice.

This is the precise cost of keying on write-reachability: it cannot distinguish "the hub cannot deliver this" from "the spoke cannot collect this", and for the pull path only the second matters. Re-keying on heartbeat history restores all 44 while still refusing genuinely uncollectible hives.

### Correction to my own earlier note

In the PR description I characterised the 26 originally-wedged hives as having "never heartbeated". That was based on a query I later found used **snake_case keys against a camelCase schema**, so every field read as null. Re-run correctly, **zero** pull-only hives have never heartbeated — they were beating all along.

This does not change the fix or any test: the gate keys on `upgradeCollectible(lastHeartbeat, now)`, which is right for both shapes, and `TestUpgradeCollectiblePredicate` covers never-beaten, unparseable, briefly-quiet and long-dead independently. But the *diagnosis* in that paragraph was wrong and I would rather correct it than leave it standing. The wedge mechanism was the unbounded re-arm loop, which the 25-cycle accumulation test pins directly.

Issue #3234, opened 2026-08-10
Follow-up: remove the legacy compatibility lanes from N1/N2 once the fleet has rolled


clubanderson:
Tracking issue so the deliberate, temporary compatibility paths added for the hosted key work don't quietly become permanent. **Each of these lanes IS the vulnerability it was added alongside** — they exist only so a flag-day cutover doesn't break the live fleet (the failure mode #2773 documented for the v2-hub/v4-spoke split).

### 1. Heartbeat: the fleet-wide bearer (N1, #3230)

`verifyHeartbeatBearer` in `v2/pkg/hub/hub_keys.go` accepts the per-hive bearer **or** the legacy fleet-wide one. The legacy branch does **not** bind identity — a spoke presenting it can still claim any `hive_id`, which is the original N1 IDOR.

Remove the second branch, leaving only `heartbeatKeyFor(hiveID)`.

**Readiness signal:** `heartbeatBearerIsPerHive(presented, hiveID)` reports whether a beat used the modern path. Log it (or expose a counter) and wait until no spoke is still using the legacy lane.

### 2. Session cookie: the legacy HMAC format (N2, #3233)

`verifyHubUserCookieEither` accepts v2 (Ed25519) **or** legacy HMAC — in both `v2/pkg/hub/hub_cookie.go` and `v2/proxy/server.js`. The legacy branch is exactly the "verify-capable implies mint-capable" problem N2 fixed: while it is live, a spoke holding `HIVE_SESSION_KEY` can still forge an admin cookie.

To remove:
- Pass `""` as `legacySecret` (the signature already supports this and it is covered by a test), then delete the branch.
- Drop `HIVE_SESSION_KEY` from the spoke Deployment template in `saas_provision.go` — keep only `HIVE_SESSION_PUBLIC_KEY`.
- Remove the `SESSION_KEY` fallback from `proxy/server.js`.

**Readiness signal:** `hubCookieIsV2(value)`. Two conditions must both hold — every spoke re-provisioned (so it can verify v2), **and** existing browser cookies aged out (bounded by the cookie's own `MaxAge`, currently 7 days).

## Ordering

The heartbeat lane can go first; it depends only on spokes having re-provisioned. The cookie lane additionally needs the browser-cookie window to elapse, so it will be later.

## Why this is worth tracking

[...]

PR #3871, opened 2026-08-14
fix(hub): drop HIVE_SESSION_KEY from spoke provisioning (issue #3234)

Danathar:
Closes part of #3234 — the N2 compatibility-lane follow-up.

## Where this leaves off

`v4` already deleted every lane that *verified* against the N1/N2 compatibility material:

| Lane | PR | Status |
|---|---|---|
| N1 heartbeat fleet-wide bearer | #3744 | deleted |
| N2 Go cookie legacy HMAC lane | #3725 | deleted |
| N2 Node proxy legacy HMAC lane | #3831 (F23) | deleted |

What was still live: the symmetric `HIVE_SESSION_KEY` itself kept being **injected into every hosted spoke Deployment**. RESIDUAL-2 (#3858, merged today) made it per-hive as defence-in-depth while the removal above was still being confirmed safe, but explicitly documented that as a stop, not the destination:

> The fallbacks stay as they are until the var is dropped outright.

Since nothing verifies against this key any more (confirmed by RESIDUAL-2's own investigation), shipping it at all is unjustified secret exposure with no remaining purpose. This PR drops it outright.

## What changed

- `saas_provision.go`: the spoke Deployment template no longer renders a `HIVE_SESSION_KEY` env entry. Newly-provisioned/reassigned spokes get only `HIVE_SESSION_PUBLIC_KEY`.
- `perhive_env_reconcile.go`: `desiredPerHiveEnv` no longer derives or wants `HIVE_SESSION_KEY`. It's removed from the mandatory `perHiveEnvNames()` and added to a new `perHiveEnvRemovedNames()` set, which `perHiveEnvDrift`/`perHiveEnvPatchJSON` treat as "must be actively stripped if seen live" — so the existing 15-minute reconcile sweep converges the **whole fleet** onto not carrying this var, at the existing 3-patches-per-cycle rate limit (no change to blast radius per cycle).
- `provisionSessionKey` (hub_keys.go) removed — its only two callers are gone.
- Doc/comment updates in `security-model.md` and `proxy/server.js` to stop describing this as "still shipped for the rollout window."

## What I deliberately did *not* change

The issue's text also says to "remove the `SESSION_KEY` fallback from `proxy/server.js`." I didn't do that part, and I think the current code is right not to:

- `SESSION_KEY` in the proxy is read for exactly one purpose now — the `IS_HOSTED` boot signal (`SESSION_KEY !== '' || SESSION_PUBLIC_KEY !== ''`) — and it self-derives from `HIVE_HUB_SECRET` when the explicit var is absent.
- That self-derive path is how **self-hosted** spokes (which never receive `HIVE_SESSION_KEY` from this hub-only provisioning path at all) get correctly detected as hosted when their operator sets `HIVE_HUB_SECRET` directly. Removing the fallback would silently flip those deployments to the self-hosted/dashboard-token identity model.
- It's provably not a security hole: `TestResidual2SessionKeyIsNotAVerificationComparand` (updated in this PR) pins that the value is read only as a presence check on both the Go and proxy side, never as a verification comparand, on both sides of this change.

So `spokeDomainKey`/`SpokeSessionKey()` (Go) and `deriveSessionKey()` (proxy) are untouched.

## Safety / rollout

Unlike the N1/N2 verification-lane removals (which would 401 real traffic mid-roll), this is safe to ship immediately and unconditionally:

- No party has verified against this value since F1/N3/F23 were deleted (upstream of this PR).
- The only two remaining readers of the value anywhere are presence checks, not comparands (pinned by test, see above).
- The reconcile sweep's existing rate limit (3 patches/15 min) means this converges the fleet at the same pace and pod-roll cadence as any other reconciled-var fix — nothing new to plan around operationally.

## Tests

- `TestSessionKeyIsPermanentlyRemoved` (new, `perhive_env_reconcile_test.go`): a Deployment still carrying `HIVE_SESSION_KEY` is reported as drifted and the patch strips it while preserving every other var; a Deployment that never carried it reports no drift (idempotent).
- `TestSessionKeyNeverDerivedInDesiredEnv` / `TestSessionKeyPerHiveDerivationRemoved` (new, `residual2_session_key_per_hive_test.go`): source-level regression guards against a sync-merge silently reintroducing the derivation — the failure mode called out in that file's own history (F3/F14/F18 were all lost that way).
- `TestResidual2SessionKeyIsNotAVerificationComparand` (updated): still passes, now documenting the *permanent* steady state rather than a mid-roll one.
- Existing generation/rotation tests (`perhive_env_generation_test.go`, `hub_generations_f19_wiring_test.go`) updated to stop asserting `HIVE_SESSION_KEY` as one of the reconciled/rotated vars.

```
go build ./... && go vet ./pkg/hub/... && go test ./pkg/hub/...
```
all clean; `gofmt -l` clean on every touched file.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/hub/...`
- [x] `go test ./pkg/hub/...` (full package, no failures)
- [x] `gofmt -l` on every touched Go file
- [x] `node --check proxy/server.js`
- [ ] Live cluster verification (no cluster access from this environment — the reconcile sweep's own tests exercise the drift/patch logic against fixtures instead)

### DIFF
--- a/src/docs/security-model.md
+++ b/src/docs/security-model.md
@@ -104,11 +104,13 @@ What that looks like when misconfigured:
 
 ### Per-hive derived keys
 
-Every spoke uses keys derived per hive from the hub master secret. The hub reconciles six env vars onto **hosted** spoke Deployments automatically (a 15-minute sweep):
+Every spoke uses keys derived per hive from the hub master secret. The hub reconciles five env vars onto **hosted** spoke Deployments automatically (a 15-minute sweep):
 
-`HIVE_HEARTBEAT_KEY`, `HIVE_SESSION_KEY`, `HIVE_SSO_PUBLIC_KEY`, `HIVE_SESSION_PUBLIC_KEY`, `HIVE_TERMINAL_KEY`, `HIVE_INVITE_KEY`
+`HIVE_HEARTBEAT_KEY`, `HIVE_SSO_PUBLIC_KEY`, `HIVE_SESSION_PUBLIC_KEY`, `HIVE_TERMINAL_KEY`, `HIVE_INVITE_KEY`
 
-**Self-hosted spokes are never touched by that sweep** — and do not need to be. Every resolver falls back to deriving the same domain-separated key from `HIVE_HUB_SECRET` + `HIVE_ID`, so a self-hosted spoke that sets those two is byte-identical to a hub-injected one. Setting the six vars explicitly is the least-privilege alternative (the spoke then never holds the master).
+**Self-hosted spokes are never touched by that sweep** — and do not need to be. Every resolver falls back to deriving the same domain-separated key from `HIVE_HUB_SECRET` + `HIVE_ID`, so a self-hosted spoke that sets those two is byte-identical to a hub-injected one. Setting the five vars explicitly is the least-privilege alternative (the spoke then never holds the master).
+
+The symmetric `HIVE_SESSION_KEY` that used to sit alongside `HIVE_SESSION_PUBLIC_KEY` is gone (issue [#3234](https://github.com/kubestellar/hive/issues/3234)): every lane that ever verified against it — the legacy HMAC cookie lane ([#3725](https://github.com/kubestellar/hive/pull/3725)), the terminal-key fall-through, and the Node proxy's copy — was already deleted, so shipping it at all was unjustified secret exposure. The sweep now actively **strips** it from any Deployment still carrying one, rather than merely no longer adding it.
 
 Two gotchas:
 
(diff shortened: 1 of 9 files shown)


Danathar:
Rebased onto `v4` (`e379c7a6`) to clear the merge conflict.

**Conflict resolved**

`docs/security-model.md` — modify/delete against the `v2/` → `src/` rename (#3996). The three edits (six→five env vars, dropping `HIVE_SESSION_KEY` from the list, and the new paragraph explaining that the sweep now actively strips it) were re-applied to `src/docs/security-model.md`. Verified the stale text was still present on `v4`, so the change is still needed.

Everything else auto-merged cleanly (`hub_keys.go`, `perhive_env_reconcile.go`, `saas_provision.go`, `proxy/server.js`).

- `go build ./...` clean
- `go test ./pkg/hub/...` passes
- No stale `v2/` path or module references remain in any changed file

No behavioural changes — conflict resolution only.

_Rebase performed with Claude Sonnet 5 (auto effort)._

PR #3877, opened 2026-08-14
feat(governor): auto-scale mode thresholds by configured repo count (#3498)

### LINKED ISSUE #3498: governor: auto-scale mode thresholds by configured repo count by default
## Problem

The governor's mode thresholds are fixed numbers (`surge=20`, `busy=10`, `quiet=2` defaults in `thresholdFor`, or explicit `governor.modes.*.threshold` values), but the queue depth they gate scales with how many repos a hive watches. A hive watching 39 repos (e.g. tuna-os) naturally carries a queue in the low hundreds — with the defaults (or modest explicit values) it sits in **SURGE** almost permanently, running the most aggressive cadences even though the per-repo backlog is small. A 3-repo hive with the same thresholds would idle. The fixed thresholds therefore encode an implicit repo count, and every hive with a different repo count has to hand-tune them (and re-tune as repos are added/removed).

Observed live: a 39-repo hive at queue ~210 sat in SURGE (threshold 70); raising the thresholds manually to `surge=300, busy=200, quiet=100, idle=20` moved it to BUSY, but that's a static value that will drift as the repo set or backlog changes.

## Proposed behavior

Scale the **default** thresholds by the configured repo count so the mode ladder reflects per-repo pressure instead of an absolute queue size:

- `effective = explicit_threshold if set, else scale(base, repo_count)` for each of `surge`/`busy`/`quiet`.
- A simple, predictable scaling: `threshold = max(base, base * len(project.repos))` — e.g. base `surge=20` becomes 20, 60, or 780 for 1/3/39 repos — or a gentler sub-linear curve (e.g. `base * ceil(sqrt(repos))`), since queue depth doesn't grow linearly with repo count.
- Explicit `governor.modes.*.threshold` config always wins (no behavior change for hives that hand-tune).
- An opt-out (e.g. `governor.auto_scale_thresholds: false`, default true) for hives that want the absolute values.

## Where

- `v2/pkg/governor/governor.go` — `computeMode` / `thresholdFor` (the defaults live here).
- Config: `v2/pkg/config/config.go` `GovernorConfig.Modes`; the repo count is already available as `config.Project.Repos`.
- Docs: the mode thresholds are described in the dashboard's governor config UI and any docs pages.

## Alternatives considered

- Normalize queue depth to per-repo (queue / len(repos)) instead of scaling thresholds — equivalent outcome, but changes what the dashboard displays and would confuse existing users; scaling the thresholds keeps the surface unchanged.
- Leave it manual — the whole point of thresholds is to avoid hand-tuning; repo count changes (adding/archiving repos) are routine and silently break the ladder.

## Impact

Low: defaults-only change; explicit config unaffected. Fixes a real footgun for large multi-repo hives and makes the governor behave consistently across hive sizes.

Danathar:
Closes #3498.

## Problem

The governor's mode thresholds gate an absolute queue depth (`surge=20`, `busy=10`, `quiet=2` defaults, or explicit `governor.modes.*.threshold`), but the queue depth a hive carries scales with how many repos it watches. A 39-repo hive naturally sits at a queue in the hundreds and stays in **SURGE** almost permanently — running the most aggressive (and most expensive) cadences — even though its per-repo backlog is small. Every hive with a different repo count has to hand-tune the thresholds, and re-tune whenever repos are added or removed.

## What changed

`v2/pkg/governor/governor.go`:
- `thresholdFor` now scales a mode's **default** threshold (one with no explicit `governor.modes.<mode>.threshold`) by `max(base, base * repoCount)` — pulled out into a pure `scaledThreshold(base, repoCount int) int` helper for direct unit testing.
- An explicit threshold still always wins — checked first, unchanged from before.
- `repoCount` is a new field on `Governor`, set via a new `SetRepoCount(n int)` method rather than by changing `New`/`UpdateConfig`'s signatures. That keeps every existing call site (and the ~30 test files across the repo that construct a `Governor` directly) compiling unchanged; a `Governor` that never gets `SetRepoCount` called on it keeps the flat historical defaults (`repoCount` defaults to 0, which `scaledThreshold` treats as "no scaling," same as 1).
- Added `EffectiveThresholds() map[string]int` — the resolved surge/busy/quiet values `computeMode` is actually evaluating against right now, for dashboard display.

`v2/pkg/config/config.go`:
- `GovernorConfig.AutoScaleThresholds *bool` + `AutoScaleThresholdsEnabled()` accessor — the opt-out (`governor.auto_scale_thresholds: false`), default on, following the exact same `*bool`/accessor pattern already used for `AttributionTrailer`.

Wiring `SetRepoCount` alongside every existing `UpdateConfig` call site, with `len(config.Project.Repos))`:
- `cmd/hive/main.go`: at construction, and in the config-reload path (so adding/archiving repos re-tunes the ladder on the next reload, not just at process start).
- `pkg/dashboard/api.go` `saveConfig()`: the dashboard-save path, which also covers the repo list changing.

`pkg/dashboard/api.go` `handleGovernorConfigGet`:
- Added `effectiveThresholds` and `autoScaleThresholds` as new, additive fields in the existing `GET /api/config/governor` response, alongside the unchanged raw `thresholds` field. This is read-only visibility (no new mutation endpoint) — an operator can now see what the governor is actually evaluating against instead of just the configured absolute number. I didn't wire up frontend display for these; that's a reasonable frontend follow-up but out of scope for what's otherwise a backend defaults fix.

## Why linear scaling, not the sub-linear alternative

The issue offered two options — linear `max(base, base*repos)` or a gentler `base*ceil(sqrt(repos))` — since queue depth doesn't grow strictly linearly with repo count. I went with linear because it's presented first as "a simple, predictable scaling," it's simpler to reason about and test, and the issue doesn't mandate matching the specific manually-tuned values from its own live example (which were explicitly called out as static and expected to drift). Nothing here would make it hard to swap in a sub-linear curve later if linear proves too aggressive in practice — `scaledThreshold` is the single, isolated place that formula lives.

## Tests

New tests in `pkg/governor/governor_threshold_test.go`:
- `TestScaledThreshold` — the pure scaling function, including the #3498 live example (39 repos: 20→780, 10→390, 2→78), zero/negative repo counts, and zero base.
- `TestThresholdForAutoScalesByDefault` — end-to-end through `thresholdFor`.
- `TestThresholdForExplicitConfigWinsOverAutoScale` — explicit config wins per-mode, with an unconfigured neighbour in the same call still scaling.
- `TestThresholdForAutoScaleOptOut` / `TestThresholdForAutoScaleDefaultIsOn` — the opt-out flag and its default.
- `TestComputeModeUsesAutoScaledThresholds` — proves the scaling reaches mode selection: `computeMode(210)` at 39 repos moves from SURGE (flat default) to QUIET (scaled), the exact shift described in the issue.
- `TestEffectiveThresholds` — the dashboard-facing accessor matches what `thresholdFor` actually resolves.
- `TestSetRepoCountIsConcurrencySafe` — `SetRepoCount` runs from a reload path independent of the eval loop; raced against `Evaluate` under `-race`.

All pre-existing `thresholdFor`/`computeMode` tests pass unmodified, confirming the change is backward compatible for any `Governor` that never calls `SetRepoCount`.

```
go build ./...
go vet ./...
go test ./pkg/governor/... -race   # includes the new tests above
go test ./pkg/config/...
go test ./pkg/dashboard/...
go test ./cmd/hive/...
```
all clean; `gofmt -l` clean on every touched file.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/governor/... -race` (full package, no failures)
- [x] `go test ./pkg/config/...`
- [x] `go test ./pkg/dashboard/...`
- [x] `go test ./cmd/hive/...`
- [x] `gofmt -l` on every touched Go file
- [ ] Live verification against a real multi-repo hive (no cluster access from this environment)

### DIFF
--- a/v2/cmd/hive/main.go
+++ b/v2/cmd/hive/main.go
@@ -956,6 +956,11 @@ func main() {
 	// Dashboard login now requests no scope and no user write-token is persisted.
 
 	gov := governor.New(cfg.Governor, cfg.EnabledAgents(), logger)
+	// Repo count feeds the default mode thresholds' auto-scaling (#3498) — set
+	// separately from New's fixed signature so config-reload can re-sync it
+	// alongside UpdateConfig below without every governor.New call site (tests
+	// included) needing a new parameter.
+	gov.SetRepoCount(len(cfg.Project.Repos))
 	sched := scheduler.New(cfg, logger)
 
 	// Wire the GitHub prompt-source resolver so agents may source their kick
@@ -2341,6 +2346,10 @@ func main() {
 		// Re-sync subsystems that cache config values
 		ghClient.SetRepos(cfg.Project.Repos)
 		gov.UpdateConfig(cfg.Governor)
+		// Re-tune auto-scaled mode thresholds (#3498) for the post-reload repo
+		// count — a hive that adds or archives repos gets a re-tuned ladder on
+		// the next reload rather than only at process start.
+		gov.SetRepoCount(len(cfg.Project.Repos))
 
 		// Re-apply live agent definitions (definition_source) on reload so an
 		// operator's edit to a linked repo propagates. Merges only operator-safe
--- a/v2/pkg/config/config.go
+++ b/v2/pkg/config/config.go
@@ -621,6 +621,16 @@ type GovernorConfig struct {
 	// so turning this off never removes the operator's ability to answer "which
 	// backend/model produced this PR?".
 	AttributionTrailer *bool `yaml:"attribution_trailer,omitempty"`
+	// AutoScaleThresholds controls whether a mode's DEFAULT threshold (one with
+	// no explicit governor.modes.<mode>.threshold set) is scaled by the
+	// configured repo count (len(project.repos)) instead of used as a flat
+	// absolute number. Default ON: a hive that says nothing gets scaled
+	// defaults. Set `governor.auto_scale_thresholds: false` to keep the flat
+	// historical defaults (surge=20, busy=10, quiet=2) regardless of repo
+	// count. An explicit threshold always wins over both the scaled and the
+	// flat default, so a hive that hand-tunes its thresholds is unaffected
+	// either way. See Governor.thresholdFor in pkg/governor.
+	AutoScaleThresholds *bool `yaml:"auto_scale_thresholds,omitempty"`
 }
 
 // AttributionTrailerEnabled reports whether the visible attribution trailer on
@@ -630,6 +640,13 @@ type GovernorConfig struct {
 // by this.
 func (g *GovernorConfig) AttributionTrailerEnabled() bool {
 	return g.AttributionTrailer == nil || *g.AttributionTrailer
+}
+
+// AutoScaleThresholdsEnabled reports whether DEFAULT (unset) mode thresholds
+// should be scaled by the hive's configured repo count. Default ON — see
+// AutoScaleThresholds's doc comment.
+func (g *GovernorConfig) AutoScaleThresholdsEnabled() bool {
+	return g.AutoScaleThresholds == nil || *g.AutoScaleThresholds
 }
 
 // GatewayConfig is one named, OpenAI-compatible model gateway. A hive may
--- a/v2/pkg/dashboard/api.go
+++ b/v2/pkg/dashboard/api.go
@@ -377,6 +377,10 @@ func (s *Server) saveConfig() error {
 	}
 	if s.deps.Governor != nil {
 		s.deps.Governor.UpdateConfig(s.deps.Config.Governor)
+		// Re-tune auto-scaled mode thresholds (#3498) whenever config is saved
+		// from the dashboard — covers the repo list changing via this path, not
+		// just a file-level reload.
+		s.deps.Governor.SetRepoCount(len(s.deps.Config.Project.Repos))
 	}
 	return nil
 }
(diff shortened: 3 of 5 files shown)


clubanderson:
Thanks for this — the change itself looks worth having, but it's targeting the wrong branch.

**`v2` is no longer maintained.** The whole fleet now runs on `v4`, and all active development, security fixes and releases land there. A change merged to `v2` today would not reach any running hive.

Could you please retarget this to **`v4`**?

Worth knowing before you do: the files you've touched have moved on since `v2`, so this will need more than a base change — `pkg/governor/governor.go`, `pkg/config/config.go` and `pkg/dashboard/api.go` have all seen substantial work on `v4`. Rebasing rather than re-pointing will likely be less painful.

One thing to check while you're rebasing: `v4` gained `AgentEligibleForCELKick` in the governor, and there is a coverage gate on that package — new exported functions without tests will fail it. Your PR already carries `governor_threshold_test.go`, so this is probably fine, but worth confirming after the rebase.

Note the directory is still named `v2/` on the `v4` branch — that's the Go module path, not the branch. Easy to misread as confirmation you're on the right branch when you aren't.

Happy to look at it again once it's pointing at `v4`.

clubanderson:
Closing this out, since `v2` is no longer a merge target — the fleet completed its v2 sunset and every running hive is on `v4`. Nothing merged here would reach a hive, so leaving it open would only be misleading about where the work stands.

**This is not a rejection of the change** — the feature is still wanted. Please do reopen it against `v4` and we'll review it there.

As noted above, a straight base change won't be enough: `pkg/governor/governor.go`, `pkg/config/config.go` and `pkg/dashboard/api.go` have all moved substantially on `v4`. Rebasing onto `v4` will be less painful than re-pointing.

Issue #3498 stays open to track the underlying request, so nothing is lost by closing the PR.

PR #3879, opened 2026-08-14
fix(entrypoint): distinct exit code for missing-CAP_NET_ADMIN startup failure (#3760)

Danathar:
Addresses the remaining ask on #3760.

## Where this issue stood

The crash-loop itself (bare `exec` EPERM from a file-capped `/usr/local/bin/hive`) was fixed in #3794 and the stale-cache follow-up #3816 — the binary carries no file capability now, and `NET_ADMIN` is raised as a bounding-set-gated ambient capability instead. Docs landed in #3781/#3830. The issue was left open for one remaining item, quoting the final comment:

> Leaving this issue open for its remaining ask (a distinct preflight error code for the missing-capability case).

## What was still generic

The entrypoint has one FATAL startup path left: if the forced-proxy-egress `iptables` redirect can't be established and `HIVE_PROXY_ADVISORY_OK` isn't set, it refuses to start — by design, since without that redirect the ACMM capability model would be silently unenforced (an agent holding a raw token could bypass the MITM proxy). That FATAL used the same `exit 1` as every other startup failure in the script (bad config, empty config file, etc.), so there was no way to distinguish "grant NET_ADMIN" from any other cause without reading the log.

## The fix

Netfilter chain manipulation itself requires `CAP_NET_ADMIN`, so if the container's capability bounding set lacks it, that's *by itself* sufficient to explain the iptables failure above — it's the same missing-capability condition the ambient-cap-raise logic already probes for, just hit at the point where the entrypoint has decided it can't safely continue rather than degrade.

`v2/deploy/entrypoint.sh`:
- Hoisted the `CapBnd` bounding-set probe (previously computed only at the later ambient-cap-raise site) to the top of the root-only setup block, so it's computed once and shared by both consumers instead of duplicated.
- The forced-egress FATAL branch now exits **77** (sysexits.h's `EX_NOPERM`, "permission denied") specifically when the bounding set lacks `CAP_NET_ADMIN`. Any other cause of the same FATAL (missing `iptables` binary, persistent netfilter lock contention) still exits `1`, since granting the capability wouldn't fix those.
- Added `EXIT_NET_ADMIN_REQUIRED=77` as a named constant near the top of the file with the rationale, rather than a bare magic number at the `exit` call site.

`v2/docs/net-admin-requirement.md`:
- New section, "When the container refuses to start (exit 77)", distinguishing this fail-closed case from the already-documented degraded-but-running mode (missing `NET_ADMIN` alone doesn't stop the container — only the redirect-can't-be-installed case does).

`v2/docs/security-model.md`:
- Cross-referenced the new exit code from the existing F5/`CAP_NET_ADMIN` section.

## Verification

- `sh -n` / `bash -n` on the modified script: clean.
- `shellcheck v2/deploy/entrypoint.sh`: same 10 pre-existing warnings as the unmodified file — none introduced by this change (verified by diffing shellcheck output before/after).
- `v2/scripts/check-suid-contract.sh` (the existing CI guard that source-asserts the NET_ADMIN contract — reads `CapBnd`, tests `0x1000`, `setpriv --ambient-caps +net_admin`, `--reuid dev`, the `gosu` fallback) still passes unmodified: those checks assert on content anywhere in the file, not position, so hoisting the probe doesn't affect them.
- Isolated shell test of the new branching logic against synthetic `CapBnd` hex values (bit present / absent / unreadable) confirmed the exit-code decision is correct in all three cases.
- `go test ./pkg/config/...` (includes `entrypoint_boot_test.go`, which executes a slice of this same script) — passes; its cut marker is well before the section this PR touches, so it was never at risk, but I ran it to confirm.

## Test plan

- [x] `sh -n` / `bash -n` entrypoint.sh
- [x] `shellcheck` — no new warnings vs. the unmodified file
- [x] `v2/scripts/check-suid-contract.sh` — passes
- [x] Isolated logic test of the CapBnd-present/absent/empty branches
- [x] `go test ./pkg/config/...`
- [ ] Live verification in a container without `CAP_NET_ADMIN` and `HIVE_PROXY_ADVISORY_OK` unset, confirming the process actually exits 77 (no container runtime available in this environment to build/run the image)


clubanderson:
Heads-up: I rebased this branch onto current `v4` and force-pushed (head is now `be64a1dc`). I verified the remote head matched your original commit beforehand, so no work was lost — but apologies for rewriting your branch without asking first.

## Why it needed doing urgently
`v4` gained **#3874** (`e6c769ff`) today, which touches the same file. That fix resolved a fleet-wide outage — 42 of 66 spokes could not reach the hub — so it was important to confirm your change composed with it rather than reverting it.

The bug it fixed: `setpriv --ambient-caps +net_admin` **exits 0 while silently doing nothing**. The kernel requires the capability in the *inheritable* set too, and `setpriv` applies ambient caps *after* the UID change, when `pI` has already been zeroed. Without `CAP_NET_ADMIN` the MITM proxy could not set `SO_MARK`, failed its own iptables bypass, and had its upstream dial REDIRECTed back into itself — every outbound connection returned `000`.

## Why a clean rebase wasn't sufficient evidence
Git reported "Successfully rebased" with **zero textual conflicts**, but that was luck of adjacency rather than git understanding the change: your PR hoists the `CapBnd` probe up to the top of the root block, and #3874 replaces the `setpriv` invocation just below where it used to sit. Overlapping intent, non-overlapping lines.

So it was verified by execution rather than by the rebase result:
```
flags extracted from entrypoint = '--inh-caps +net_admin --ambient-caps +net_admin'
CapAmb after entrypoint drop = 0000000000001000   <- bit 12 set
CapAmb after buggy drop      = 0000000000000000   <- negative control still broken
```

**Your change still fires correctly** — exit 77 on the missing-capability path only, and the adjacent `NET_RAW` bit does not false-trigger:
```
CapBnd=00000000a80425fb (docker default, no NET_ADMIN) -> exit 77
CapBnd=00000000a80435fb (--cap-add NET_ADMIN)          -> exit 1
CapBnd=0000000000002000 (NET_RAW only, adjacent bit)   -> exit 77
CapBnd=0000000000001000 (NET_ADMIN only)               -> exit 1
```

One extra commit is included: `v2/Dockerfile:381` still documented the mechanism as `setpriv --ambient-caps +net_admin` — the broken form, presented as current. Comment-only, but a live trap for anyone later "simplifying" the entrypoint back into the outage.

## What's needed from a maintainer
Workflows are showing `action_required` because this is a fork PR — the "SUID Helper Contract" run (which carries both `check-suid-contract` and the new `ambient-cap-runtime` gate) is queued on the correct head and needs approval to execute. Both pass locally: 20/20 and 6/6.

If you'd rather rebase your own branches in future, say so and I'll flag rather than push.

Danathar:
No apology needed — rebasing onto a fix for a fleet-wide outage is the right call, and you left the branch in better shape than you found it. Please don't flag-and-wait on my account in that situation; pushing was correct.

I re-verified rather than take it on trust, since "no textual conflict" is exactly the evidence you're right to distrust here:

- **My commit survived.** `f16dbedf` is intact on the new head.
- **#3874 is genuinely composed, not reverted.** `git merge-base --is-ancestor e6c769ff HEAD` → true, and `entrypoint.sh:932` reads `_SETPRIV_CAPS="--inh-caps +net_admin --ambient-caps +net_admin"`, so the inheritable half is present at the source level, not just in the rebase result.
- **Your Dockerfile correction matches reality.** The comment block now documents `setpriv --inh-caps +net_admin --ambient-caps +net_admin` and names #3874 for why both flags are required — which agrees with what the entrypoint actually does. Worth having: a comment describing the outage form as current is a live trap, and the next "simplification" would have followed it straight back in.
- **All three gates pass here:**
  - `check-suid-contract.sh` — passes, including the new "raises NET_ADMIN into the INHERITABLE set (#3874)" assertion.
  - `test_ambient_cap_runtime.sh` — 6/6.
  - `test_entrypoint_netadmin_ambient.sh` — 17/17, covering exactly the discrimination you exercised by hand: docker default (no NET_ADMIN) → skip, NET_RAW-only at the adjacent bit 13 → skip, NET_ADMIN-only at bit 12 → raise, `--cap-add NET_ADMIN` → raise.

So the exit-77 path still fires only on the missing-capability case, and the adjacent bit still doesn't false-trigger.

Nothing needed from me. The remaining blocker is the one you named: the fork-PR workflows sit at `action_required` and need a maintainer to approve the run.

clubanderson:
Reviewed with the failure-path contract in mind. The healthy-boot path is untouched: the #3717 retry/jitter iptables logic and the advisory-mode escape are byte-identical, and the hoisted `CapBnd` probe is the same computation moved earlier in the same root-context process, so the FATAL branch and the later raise gate can never disagree about the capability state. Exit 77 fires only inside the already-fatal else — a boot that reaches it was refusing to start regardless, so this changes what the exit status says, never whether it exits. On #3874 compatibility: probing the bounding set is exactly the state that gates the shipped `--inh-caps +net_admin --ambient-caps +net_admin` raise, so the new code and the raise logic key off one shared value — and the Dockerfile comment correction now describes the mechanism actually shipped, with the why-both-flags rationale that should stop a future 'simplification' back to the silent no-op. Named `EXIT_NET_ADMIN_REQUIRED=77` (EX_NOPERM) with rationale rather than a bare magic number is exactly right. Merging via the fork-CI gate branch; I'll note on #3760 that its remaining preflight ask has landed.

Issue #1861, opened 2026-07-14
Inject GitHub auth at the MITM proxy so agents never hold tokens; tighten shared token cache perms


clubanderson:
## Current state

Agents authenticate to GitHub three different ways, each with a weakness:

1. **git** → credential helper (`git-credential-hive.sh`): good — UID-verified, mode-gated — but only covers git.
2. **gh CLI** → per-agent scoped token file after #1860 (previously agents improvised with the shared cache).
3. **Shared `gh-app-token.cache`** → `0640` group `node`, readable by *every* agent UID. Any agent can `cat` it and skip the helper's UID/mode gate entirely, holding an unscoped app token. The MITM proxy's mode enforcement then stands alone.

## Proposal

The proxy already MITMs `api.github.com` and inspects every request per agent mode. Complete the design:

1. **Proxy-side auth injection**: for requests the mode check allows, the proxy strips whatever `Authorization` header the agent sent and injects a token minted server-side for that agent's tier. Agents then need *no* token material at all — `gh` works with a dummy `GH_TOKEN=hive` (gh requires the variable to be non-empty), and token theft between agents becomes structurally impossible. This also finishes the story started by the `${GH_AUTH}` text ("authentication is handled in transit"), which today is only true for the CA layer.
2. **Tighten the shared cache** to `0600` once nothing but the hive runtime (UID 0/dev) needs it — the credential helper runs as the agent but can be given a setuid-free path via a tiny local token broker (unix socket) if needed.
3. **git via injection too**: once injection exists, the credential helper can hand out a dummy credential and rely on the proxy, keeping the mode gate as a fast local pre-check.

## Why now

Field debugging (2026-07-13/14) showed an agent independently discovering the shared-cache read within one session when its push failed — the gap gets found under pressure. Two layers were designed; today only one holds once a token leaks sideways.

Related: #1860 (interim: per-agent scoped tokens for gh), #1849/#1850/#1856/#1858 (this week's config/enforcement hardening).

Om-Beast:
Hi @clubanderson, thanks for the detailed proposal.

I'd like to work on this feature. I'll start by reviewing the current authentication flow, the MITM proxy implementation, and the related issues (#1860, #1849, #1850, #1856, and #1858) to understand the existing design before proposing changes.

My plan is to first identify the integration points for proxy-side authentication injection, verify the impact on both `git` and `gh` workflows, and share an implementation approach before opening a PR to ensure it aligns with the project's architecture.

[...]

PR #3888, opened 2026-08-14
fix(proxy): stop trusting self-asserted agent identity as a fallback (N7, #3841)

### LINKED ISSUE #3841: [sec] proxy: make UID authoritative for agent identity; decide behaviour when the UID map is missing (N7)
Split out of #2172 (closed 2026-08-14 after triage).

## Current state — materially better than when filed
`identifyAgentFromReq` and `identifyAgentFromConn` (`v2/pkg/proxy/github_proxy.go:655,666`) now try **unforgeable UID lookup first**. `extractAgentName` (`:1103`) is only a fallback. `identifyAgentFromConn` also closed the trap where `ReadRequest` left `RemoteAddr` empty.

## Residual
The forgeable-header fallback is **still reachable whenever the UID map is absent**. An agent that can influence which path is taken can present a name of its choosing.

## Why this blocks other work
#1861 (inject GitHub auth at the MITM proxy so agents never hold tokens) **depends on this closing first**. Injection mints a token for whoever the proxy believes is calling — building it on an identity that can fall back to a forgeable header turns name-spoofing into token-minting. Triage of #1861 called this out as its primary blocker.

## Decision needed
What should happen when the UID map is unavailable? Fail closed (deny egress) is the safe default, but it needs checking against legitimate startup ordering — the map may genuinely be absent briefly during boot. Whatever is chosen, it should be explicit and tested, not an implicit fallback.

Verified on `origin/v4` @ `7cd059b7`.

Danathar:
Closes #3841.

## Where this issue stood

`identifyAgentFromReq`/`identifyAgentFromConn` (`v2/pkg/proxy/github_proxy.go`) already try unforgeable UID lookup first (reading `/proc/net/tcp` against `uid-map.json`), added to close an earlier gap. The issue's residual finding: whenever that lookup didn't succeed, both functions unconditionally fell through to `extractAgentName(r)` — the `Proxy-Authorization` header, which is entirely **caller-controlled**.

## The gap, concretely

UID lookup can fail for two distinct reasons, and both hit the same forgeable fallback:

1. **`p.uidMap == nil`** — never loaded. In the container/K8s model this shouldn't race in practice (`entrypoint.sh` writes `uid-map.json` in its root phase, strictly before the Go binary is ever exec'd — I traced this to confirm there's no real happens-before gap there), but it's the *permanent* state for local dev, `go test`, and native/systemd installs (`install.sh` + `kick-agents.sh`), which have no per-agent UID separation at all.
2. **UID map loaded, but no entry for this connection's UID** — e.g. a UID genuinely not in the map.

In both cases, any process able to reach the proxy's loopback port could send `Proxy-Authorization: hive <any-agent-name>` and be identified as that agent. #1861 (inject GitHub auth at the proxy so agents never hold tokens) explicitly calls this out as its blocker: minting a token for a self-asserted identity turns name-spoofing into privilege escalation.

## The fix

I didn't invent a new "fail closed / deny" mode. The codebase already has one: `readAgentMode("")` resolves to `ModeAdvisory` (the most restrictive mode — writes blocked), and `proxyHTTP` already has a WARN log for exactly this scenario ("agent could not be identified — defaulting to ADVISORY and blocking a write... this is NOT an agent-mode misconfiguration"). So the fix is to route *both* failure cases above into that same, already-tested, already-restrictive path — by default, return `""` instead of trusting the header.

The one deliberate exception: `HIVE_PROXY_ADVISORY_OK=true`. This is the **same** explicit, operator-set escape hatch `entrypoint.sh` already requires before it will start with a degraded forced-egress gate (missing `NET_ADMIN`, no iptables). Reusing it here means:
- An operator who has already accepted a degraded capability model for this deployment isn't newly exposed by also accepting a self-asserted identity within it.
- Local dev, native/systemd installs, and any other deployment with no UID map at all keep working exactly as before — this is a closed security gap, not a functional regression for deployments that never had UID-based identity to begin with.
- One flag, one meaning, documented once — not two different escape hatches for two halves of the same "is this deployment fully enforcing the capability model" question.

## What changed

`v2/pkg/proxy/github_proxy.go`:
- New `proxyAdvisoryOK bool` field, read once at construction from `HIVE_PROXY_ADVISORY_OK` (mirrors `entrypoint.sh`'s own check). Logged at startup either way, since a missing UID map is worth an operator's attention regardless of which behavior results.
- New `fallbackAgentName` helper: gates the header fallback on `proxyAdvisoryOK`. `identifyAgentFromReq`/`identifyAgentFromConn` now call this instead of `extractAgentName` directly.
- `extractAgentName` itself is untouched — it's still correct as a pure header parser; what changed is who's allowed to trust its result.

`v2/pkg/proxy/*_test.go`:
- Updated four existing tests that asserted the old unconditional fallback as *correct* behavior (`TestIdentifyAgentFromReqWithProxyAuth`, `TestIdentifyAgentFromReqUIDMapNoIptables`, `TestIdentifyAgentFromReqIptablesEmptyUID`, `TestIdentifyAgentFromReqUIDMapFallback`) to assert the new default (`""`), each now paired with an explicit advisory-mode variant proving the escape hatch still works.
- `newInferenceTestProxy` (shared test helper used by 12 tests, several of which identify synthetic CONNECT requests via header for test convenience rather than building a real UID-map fixture) now sets `proxyAdvisoryOK: true` — that's exactly the scenario the flag exists for.
- New `TestIdentifyAgentFromReqWithProxyAuthAdvisoryOK` and `TestIdentifyAgentFromConnDoesNotTrustHeaderByDefault` (the `identifyAgentFromConn` half — the issue names both functions, and only `identifyAgentFromReq` had explicit fallback coverage before).

`v2/docs/env-vars.md` / `v2/docs/security-model.md`:
- `HIVE_PROXY_ADVISORY_OK`'s doc entry and the F5 section now describe both things it gates, not just the iptables one.

## Verification

- Traced every consumer of `agentName == ""` downstream of these two functions (`readAgentMode`, the mode gate in `proxyHTTP`, `IsInternalUID`'s explicit-sentinel handling in the separate `handleTransparentTLS` path) to confirm empty string is never *also* used as a "trusted/internal" sentinel anywhere — it is exclusively the least-privileged case.
- `handleTransparentTLS` (the iptables-redirected path, a third, independent UID-identification implementation) was already safe — it has no header fallback at all, so it needed no change; confirms this was a real, scoped gap rather than a repo-wide pattern.
- `go build ./...`, `go vet ./...`, `go test ./pkg/proxy/... -race` (full package): all clean.
- `gofmt -l` clean on every touched file.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/proxy/... -race` (full package, no failures)
- [x] `go test ./cmd/hive/...`
- [x] `gofmt -l` on every touched file
- [ ] Live verification against a running proxy without a UID map and without `HIVE_PROXY_ADVISORY_OK` (no container runtime available in this environment)


clubanderson (approved):
Reviewed as the N7 identity-spoofing fix, with the trust boundary traced end to end.

**Verified the fail-closed path**: `readAgentMode("")` returns `ModeAdvisory` (github_proxy.go:1218-1221) and empty-string is nowhere a trusted/internal sentinel — the internal-UID case in `handleTransparentTLS` (line 411) keys on `IsInternalUID(uid)`, not on an empty name. Routing both failure cases (nil map, no UID match) into the existing unidentified-caller path rather than inventing new deny plumbing is the right shape.

**Verified no production caller depends on the header**: grepped `bin/`, `v2/deploy/`, and `dashboard/` — nothing sends `Proxy-Authorization` in production; the container deployment deliberately uses no proxy userinfo, so UID lookup was already the only working attribution in the fleet. And the historical UID-lookup miss on IPv6 loopback was fixed separately (procnet.go scans both `/proc/net/tcp` and `tcp6`), so demoting UID-miss to advisory won't reintroduce that class of blocked writes.

**Escape hatch reuse is sound**: `HIVE_PROXY_ADVISORY_OK` already means "this deployment does not enforce the full capability model" — one flag, one meaning, and native/dev installs with no UID separation keep working. Startup logs a WARN in both directions, so the degraded state is visible either way.

**Tests assert the invariant**: the four flipped tests each pair the deny default with a positive advisory-mode control, and `identifyAgentFromConn` (previously uncovered) gets its own regression. Setting `proxyAdvisoryOK: true` in `newInferenceTestProxy` is honest about what those 12 tests use the header for.

LGTM — merging on green.

clubanderson:
Reviewed with one blocking verification before merge: the PR body's claim that "local dev, native/systemd installs keep working exactly as before" — i.e., that defaulting the identification-failure path to ADVISORY does not silently drop native agents into writes-blocked mode.

**Verification result: claim holds, no fix needed.** The native/systemd install path never routes traffic through the Go proxy at all, so it never relied on header identification:

- `install.sh` (lines 90-99) installs only shell scripts (`supervisor.sh`, `agent-launch.sh`, `gh-wrapper.sh`, `kick-agents.sh`, ...) plus the `hive.service` / `hive@.service` supervisor units — no v2 `hive` binary, no proxy process, no unit for one.
- `grep -rn HIVE_PROXY_ADVISORY_OK bin/ install.sh systemd/ config/` → zero hits, and equally zero hits for `HTTPS_PROXY`/`18443` in any installed script or unit: native agents call `gh` directly via `gh-wrapper.sh`; nothing points them at a loopback proxy.
- `bin/setup-proxy-iptables.sh` is explicitly container-only ("Run inside the container as root during startup") and has no callers on the native path.
- The only deployments that run the proxy *without* a working UID map + egress gate are degraded container starts — and `v2/deploy/entrypoint.sh` (lines 866-869) already refuses to boot those unless `HIVE_PROXY_ADVISORY_OK=true` is set. That same flag is what this PR reuses to permit the header fallback, so any deployment that could have depended on header identity necessarily already sets it. Enforcing deployments get `uid-map.json` written by the entrypoint root phase and are identified unforgeably.

**Test coverage confirmed:** all four tests that encoded the old fallback as correct are inverted (`TestIdentifyAgentFromReqWithProxyAuth`, `TestIdentifyAgentFromReqUIDMapNoIptables` in `proxy_extra_test.go`; `TestIdentifyAgentFromReqUIDMapFallback` in `proxy_deep_test.go`; `TestIdentifyAgentFromReqIptablesEmptyUID` in `proxy_deep2_test.go`), each asserting `""` with a crafted header present. Advisory-mode positive controls exist for both entry points (`TestIdentifyAgentFromReqWithProxyAuthAdvisoryOK`, flag-flip halves of `TestIdentifyAgentFromReqUIDMapFallback` and the new `TestIdentifyAgentFromConnDoesNotTrustHeaderByDefault`). The `newInferenceTestProxy` helper's `proxyAdvisoryOK: true` convenience does not mask the default: the default-path tests construct `GitHubProxy` directly without the flag. `go test ./pkg/proxy/... -race` on the merge result with current v4: **ok, 49.5s**.

Post-merge note for operators: once the fleet rolls, watch a hosted spoke's proxy log for the "agent could not be identified — defaulting to ADVISORY" WARN. On a MANAGED spoke that message would indicate a uid-map coverage gap (a connection whose UID is missing from `uid-map.json`), which previously was silently papered over by the header fallback.

PR #3895, opened 2026-08-14
fix(security): harden K8s UID capabilities and MITM CA key

### LINKED ISSUE #3866: [sec-check] K8s deployment allowPrivilegeEscalation:true + writable /data could expose MITM CA key
## Security Finding

**Severity**: medium
**Type**: permission-issue / hardening

The K8s deployment (`v2/deploy/k8s/deployment.yaml`) uses `allowPrivilegeEscalation: true` to support the SUID-root `su-exec` helper:

```yaml
securityContext:
  allowPrivilegeEscalation: true
  capabilities:
    drop:
      - NET_RAW
```

While the `su-exec` binary is correctly restricted to the `hive-launch` supplementary group (which only `dev` belongs to), and the pod-level seccomp is `RuntimeDefault`, `allowPrivilegeEscalation: true` means any SUID binary on the system that can be accessed by any user in the container could be exploited.

Additionally, the container lacks:
- `readOnlyRootFilesystem: true` (documented as intentional — hive writes to disk)
- A `SETUID`/`SETGID` capability drop

## Impact

A container compromise that gains execution as `dev` (UID 1001) could:
1. Use `su-exec` to switch to an arbitrary UID (the su-exec function hive relies on)
2. Access `/data/proxy-ca-key.pem` (MITM CA private key, `0600` owned by dev)
3. Forge TLS certificates for `api.github.com` and intercept all agent GitHub API traffic

## Recommendation

1. **Drop SETUID/SETGID capabilities** in the container `securityContext.capabilities.drop` list if the OS-level restriction on the `su-exec` binary (group `hive-launch`, chmod 4750) is sufficient and no other SUID binaries need to be accessible.
2. **Audit all SUID/SGID binaries** in the final image: `find / -perm /4000 -o -perm /2000 2>/dev/null` — document and justify each one.
3. If dropping SETUID is incompatible with the su-exec design, add a note to the deployment README explaining this trade-off explicitly, and ensure the `hive-launch` group membership is always enforced by the Dockerfile.

---

Danathar:
## Summary

Fixes #3866.

- Drop `SETUID` and `SETGID` from the Kubernetes container capability set while retaining the tightly group-gated `4750 root:hive-launch` `su-exec` helper.
- Store the persisted MITM CA private key below `/data/.hive` (`0700` directory, `0600` file) instead of directly in the shared agent-writable `/data` namespace. If the legacy exposed key is found, discard the old CA pair so a potentially copied key cannot remain trusted.
- Harden the hosted-hive provisioning template with the same capability posture and `RuntimeDefault` seccomp profile.
- Change NET_ADMIN drift reconciliation to patch only `capabilities.add`, preserving securityContext hardening on existing deployments.
- Add static deployment/security contracts, CA permission coverage, and operator documentation.

## Validation

- `bash v2/deploy/test_security_hardening.sh`
- `bash v2/scripts/check-suid-contract.sh`
- `bash -n v2/deploy/entrypoint.sh v2/deploy/test_security_hardening.sh v2/scripts/check-suid-contract.sh`
- `cd v2 && go test ./pkg/proxy ./pkg/hub`
- Kubernetes manifest parsed successfully with PyYAML.

The checkout contained unrelated pre-existing dashboard/CSP worktree changes; they were intentionally left uncommitted and are not part of this PR.


Danathar:
Rebased onto `v4` (`e379c7a6`) to clear the merge conflicts.

**Conflicts resolved**

1. `deploy/k8s/deployment.yaml` — real content conflict. `v4` had gained an OpenShift SCC admission note in the same `capabilities:` block where this PR adds the `SETUID`/`SETGID` drops. Both are kept: the drops now sit under `drop:`, followed by v4's OpenShift comment above `add:`. Confirmed `v4` still grants `SETUID`/`SETGID` via `add:`, so this PR's core fix is still needed and hasn't landed elsewhere.
2. `docs/security-model.md` — modify/delete against the `v2/` → `src/` rename (#3996). Re-applied all three edits onto `src/docs/security-model.md`.
3. `deploy/test_security_hardening.sh` — relocated `v2/` → `src/`, and its default path arguments (`v2/deploy/...`, `v2/pkg/...`) were updated to `src/`, which the rename would otherwise have left pointing at directories that no longer exist. Exec bit restored.

- `go build ./...` clean
- `go test ./pkg/proxy ./pkg/hub` — both pass
- `bash src/deploy/test_security_hardening.sh` — 7/7 pass, including `reconcile targets capabilities.add` and `reconcile does not patch the complete securityContext`
- `bash src/scripts/check-suid-contract.sh` — 20/20 pass
- `bash -n` on all three touched shell scripts, and the manifest parses under PyYAML

No behavioural changes were made — only conflict resolution and path corrections.

_Rebase performed with Claude Sonnet 5 (auto effort)._

clubanderson:
Thanks @Danathar — triaged this today. Splitting my read into the two halves of the PR, because they deserve different fates:

**The MITM CA-key hardening is good and still needed.** `src/pkg/proxy/github_proxy.go` on current `v4` still has `caKeyPath = "/data/proxy-ca-key.pem"` — the key still sits in the shared agent-writable `/data` namespace. The relocation to `/data/.hive` (0700/0600), the refuse-to-migrate/discard-the-pair handling of a legacy exposed key, and the symlink rejection are all exactly right.

**The SETUID/SETGID capability drop would break the pod at runtime, though.** Holding on that half:

1. **Boot breaks.** The entrypoint starts as root and drops to `dev` via `setpriv --reuid dev` (or the `gosu dev` fallback) — see `src/deploy/entrypoint.sh` ~L949–989. Both call `setuid()`/`setgid()`, which require `CAP_SETUID`/`CAP_SETGID` in the process's permitted set. For the container's initial root process, permitted == the bounding set — which this PR removes them from. First privilege drop fails; pod crash-loops.
2. **Agent launch breaks.** The PR's premise — "the kernel applies its set-id mode during exec, and su-exec performs the controlled UID transition" — conflates two steps. The SUID bit gets you *to* euid 0; per capabilities(7), a setuid-root exec's new permitted set is `(bounding | inheritable)`. With SETUID/SETGID out of the bounding set, su-exec's subsequent `setgroups()/setgid()/setuid()` to the per-agent UID fail with EPERM. This is the classic reason `docker run --cap-drop SETUID` breaks `su`/`gosu`/`su-exec`.
3. **Green CI doesn't cover this.** `test_security_hardening.sh` and `check-suid-contract.sh` are static greps, and the one runtime job (`ambient-cap-runtime`, `src/deploy/test_ambient_cap_runtime.sh`) runs docker with *default* caps plus `--cap-add NET_ADMIN` — it never boots with the manifest's actual drop set, so it can't catch this.
4. Current `v4`'s manifest documents the dependency explicitly (`- SETUID # su-exec dev->per-agent UID switch`), and the hosted-hive template change in `saas_provision.go` would push the same breakage fleet-wide via provisioning.

**Requested changes:**
- Please split out and resubmit the CA-key half (entrypoint migration, `github_proxy.go` path + perms enforcement, docs, and its tests) — I'd expect that to merge quickly.
- If you still want to pursue the capability drop, it needs a design that survives a *runtime* test: e.g. a CI job that boots the real image with the manifest's exact `--cap-drop`/`--cap-add` set, gets through the setpriv/gosu drop, and launches a process through su-exec to a non-dev UID. The `netadmin_reconcile` "patch only `capabilities.add`" change also interacts with the C6 rationale in the Dockerfile/manifest comments and should be discussed in that context.

/hold

Danathar:
Thanks — that's a clear read, and I agree on both halves.

**CA-key half split out as #4341.** Four files only: `github_proxy.go` (path + directory/permission enforcement), its coverage test, the `entrypoint.sh` migration, and one added bullet in `security-model.md`. I took only the new CA-key bullet from that file — the `su-exec` and hosted-fleet bullets in this PR both describe capability-drop behavior, so they stayed behind. Applied cleanly to current `v4` (`9a7e745f`); build, vet and `pkg/proxy` are green. `caKeyPath = "/data/proxy-ca-key.pem"` is still live on `v4` today, so the exposure is unchanged.

**On the capability drop — you're right and I'm not going to argue it.** The `capabilities(7)` point is the one I got wrong: I treated the SUID bit as sufficient for the whole transition, when a setuid-root exec's permitted set is `(bounding | inheritable)`, so dropping SETUID/SETGID from the bounding set makes `su-exec`'s `setuid()` fail EPERM regardless of the mode bits. The `setpriv`/`gosu` boot drop fails for the same reason, one step earlier. And your point about why CI missed it is the part worth keeping: the static checks are greps, and `test_ambient_cap_runtime.sh` boots with default caps plus `--cap-add NET_ADMIN`, never the manifest's actual drop set — so it structurally cannot catch this.

Happy to leave this PR held. If the capability drop is worth revisiting, the prerequisite is the runtime lane you described — a job that boots the real image with the manifest's exact `--cap-drop`/`--cap-add`, clears the setpriv/gosu drop, and launches through `su-exec` to a non-dev UID. That is a real gap in its own right regardless of this PR, since nothing currently exercises the manifest's cap set at runtime. Say the word and I'll open it as its own issue rather than reopening this one.

clubanderson:
Closing as superseded. The CA-key half of this PR — relocating the MITM CA private key out of the agent-writable `/data` namespace to `/data/.hive/proxy-ca-key.pem` (0700 dir, 0600 key, symlink rejection, legacy-key discard) — was split out at our request into #4341 by @Danathar and has now merged as 7e98ccd6781cbc21cd8dc6b354301ed17ae3fc68. Full credit to this PR for identifying and fixing that exposure.

The remaining half (dropping `SETUID`/`SETGID` from the bounding set) was rejected in triage (https://github.com/kubestellar/hive/pull/3895#issuecomment-5360880981): the entrypoint depends on those capabilities at runtime for the root→dev and dev→agent-UID switches, so the drop breaks agent launches. With nothing mergeable left here, closing.

Thanks @Danathar for the split and for the original report.

Danathar:
Filed #4379 for the coverage gap you identified in this review — nothing boots the manifest's actual `drop: [ALL]` + six-capability `add:` list at runtime. `test_ambient_cap_runtime.sh` runs `docker run --cap-add=NET_ADMIN` on Docker's default bounding set, which is far wider than what ships, and the rest of that area (`test_security_hardening.sh`, `check-suid-contract.sh`) is greps over files.

The point worth preserving is that **CI was green on this PR**. The SETUID/SETGID breakage was caught by your reading of `capabilities(7)`, not by a test, and the same class of change would be green again today.

#4379 is scoped as coverage only — explicitly not a proposal to revisit the capability drop, and it says so. It requires the tested set to be *derived* from `deployment.yaml` rather than duplicated into the test, since a test carrying its own copy would pass while the manifest drifted, which is the failure being fixed. Leaving this PR closed and held.

PR #3896, opened 2026-08-14
test(hub): cover remaining provider login branches

Danathar:
## Summary

Refs #3867.

The baseline provider-login tests from #3894 are already present on `v4` and cover the main handler/picker paths. This PR adds the remaining focused branch coverage:

- Verify `handleProviderLogin` preserves case-insensitive provider lookup and the `rd` redirect target in OAuth state.
- Verify `writeProviderPicker` omits an empty redirect query.
- Verify picker provider names and display labels are HTML-escaped.
- Exercise every known `providerGlyph` branch: GitHub, Google, IBMid, Red Hat, Microsoft, and custom, in addition to the existing unknown-provider test.

## Validation

- `go test ./pkg/hub -run 'TestHandleProviderLogin|TestWriteProviderPicker|TestProviderGlyph' -count=1 -coverprofile=/tmp/hub3867.out`
- `go tool cover -func=/tmp/hub3867.out` reports:
  - `handleProviderLogin`: 100%
  - `writeProviderPicker`: 100%
  - `providerGlyph`: 100%
- `go test ./pkg/hub -count=1`

### DIFF
--- a/v2/pkg/hub/oauth_provider_login_additional_test.go
+++ b/v2/pkg/hub/oauth_provider_login_additional_test.go
@@ -1,0 +1,106 @@
+package hub
+
+import (
+	"html"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/kubestellar/hive/v2/pkg/auth"
+)
+
+func TestHandleProviderLoginPreservesProviderAndRedirectInState(t *testing.T) {
+	s := &HubServer{authProviders: auth.NewRegistry(&auth.Provider{
+		Name:         "github",
+		DisplayName:  "GitHub",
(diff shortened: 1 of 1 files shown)


clubanderson:
Reviewed. These are genuinely substantive additions, not assertion padding: the state test asserts the round-trip invariant (case-insensitive lookup feeding the lowercased provider name plus the `rd` target into OAuth state), the escaping test carries both a positive control (escaped name/label present in the body) and a negative one (no executable attribute survives), and each `providerGlyph` branch is pinned by a marker I verified against distinct SVG content in the current v4 `oauth.go` — so a branch swap or fallthrough would fail the right subtest. Merging via the fork-CI gate branch; since the body uses `Refs` rather than `Fixes`, I'll close #3867 manually once this lands.

PR #3898, opened 2026-08-14
✨ feature(governor,config,dashboard): auto-scale mode thresholds by configured repo count

Danathar:
## What

Scales the **default** governor mode thresholds by the hive's configured repo count, so the mode ladder reflects per-repo pressure instead of an absolute queue size.

Refs #3498 — no `Fixes` keyword, deliberately. The issue left the scaling curve open ("linear — or a gentler sub-linear curve"), and I found a limitation that materially affects how much of the problem this actually solves on real hives (see **ACMM packs** below). Both are maintainer calls.

## The problem

`thresholdFor` returned fixed numbers (`surge=20`, `busy=10`, `quiet=2`) for a queue depth that is `actionable issues + open PRs` summed across **every watched repo**. So the thresholds encoded an implicit repo count:

- 39-repo hive, queue ~210 → permanent **SURGE**, most aggressive cadences, tiny per-repo backlog.
- 3-repo hive, same thresholds → idles through a backlog that is genuinely deep per repo.

## What changed

Default thresholds are now per-repo bases × `len(project.repos)`:

| | base (1 repo) | 3 repos | 39 repos |
|---|---:|---:|---:|
| `surge` | 20 | 60 | 780 |
| `busy` | 10 | 30 | 390 |
| `quiet` | 2 | 6 | 78 |

A one-repo hive gets **exactly** the numbers it got before. `primary_repo` with no `repos:` list counts as 1, so the floor never collapses a threshold to zero (which would pin the hive in SURGE — the opposite of the fix).

## Choosing the curve — and why linear is the default

The issue offered linear or `base * ceil(sqrt(repos))` and did not pick. I implemented both, defaulting to **linear**, for two reasons:

1. **Linear is exactly the per-repo normalization the issue's own "Alternatives considered" section describes**, without changing the queue number the dashboard displays. `queue > base * repos` ⟺ `queue/repos > base`. It is the principled reading, not just the simpler one.
2. **Sqrt does not fix the reported case.** `ceil(√39) = 7`, so surge lands at 140 and the reported hive at queue ~210 stays in **SURGE** — the exact symptom being reported. Linear puts it in QUIET (210/39 ≈ 5.4 per repo, which sits in the quiet band). The reporter hand-tuned their way to BUSY; linear lands one notch below that, which is a judgement call worth a maintainer's eye, but either is a world away from permanent surge.

Sqrt is still offered for hives with many quiet repos alongside a few busy ones:

```yaml
governor:
  threshold_scaling: linear   # linear (default) | sqrt | none
```

### On the config field name

The issue proposed `governor.auto_scale_thresholds: false` as the opt-out. I used `threshold_scaling: none` instead — the same switch, folded into the field that also picks the curve. Two separate knobs can contradict each other (`auto_scale_thresholds: false` + `threshold_scaling: sqrt` has no meaning), and one enum has no invalid combinations. Easy to rename if you'd rather have the original spelling.

Note `sqrt` and `linear` are **identical at 1 and 2 repos** and only separate from 3 onward, because the ceiling can't produce a factor below 2 until 5 repos. My first test asserted strict inequality everywhere and failed at `repos=2`; the test now pins the real shape of the curve rather than the assumed one.

## Explicit thresholds always win

A `threshold:` you set is returned **verbatim, never scaled**. Scaling an operator's hand-tuned `surge: 300` on a 39-repo hive would produce **11700** and silently break every hive that already worked around this bug by hand — that's the single most important guarantee here, and it has its own test.

A threshold of `0` still counts as **unset**, matching the existing behavior: mode entries frequently exist only to carry cadences.

### New hazard: mixed sources can invert the ladder

Because explicit values don't scale, mixing them with scaled defaults can put the ladder out of order. On a 39-repo hive with the reporter's own `surge: 70`, `busy` scales to 390 — **above** surge. `computeMode` tests surge first and returns, so BUSY becomes unreachable.

This was not reachable before: the three thresholds were previously either all defaults or all hand-set at comparable magnitudes. The governor now warns with all three effective values:

```
governor mode thresholds are not in descending order — a mode is unreachable
  surge=70 busy=390 quiet=78 repo_count=39 threshold_scaling=linear
```

It **warns rather than reorders** — the inversion comes from a number the operator set, and silently overriding it would be the worse surprise.

## ⚠️ ACMM packs seed explicit thresholds, so scaling won't engage on pack-applied hives

This is the finding most worth your attention, and it limits the feature's reach beyond what the issue anticipated.

`pkg/dashboard/api_packs.go:282` writes pack thresholds into `governor.modes.*.threshold` on apply, and `pkg/config/packs/level-*.yaml` all carry them (L4–L6: `surge: 10, busy: 5, quiet: 2`; L3: `15/10/3`).

Those land in config as **explicit** values. Nothing distinguishes a pack-seeded default from one an operator typed. So under "explicit always wins" — which the issue is emphatic about — **scaling does not engage on any hive that applied an ACMM level**, which is the normal path and plausibly includes the 39-repo hive that motivated the issue.

I did **not** work around this. The available workarounds are all worse than the gap:

- Scaling explicit values too would 39× a hand-tuner's `300` → `11700`.
- Dropping thresholds from the pack YAMLs loses real per-level tuning (L3 differs from L4–L6).
- Baking scaled values at apply time goes stale the moment repos change — the exact drift the issue exists to eliminate.

Fixing it properly means recording **provenance** (pack-seeded vs operator-set) so rule 1 can apply only to the latter. That's a config-layering decision, not something to bolt on inside this PR. Documented in `docs/governor-thresholds.md` with the manual workaround (clear the thresholds you want scaled).

The feature is fully correct and effective for hives without packs, and for anyone who clears the seeded values.

## Shared resolution — the gauge no longer keeps its own copy

`buildGovernor` (`status_builder.go`) had **private duplicates** of `2/10/20` plus its own explicit-wins check, with a comment saying "keep the defaults so the gauge matches the governor's evaluation."

Harmless while both were constant. Once the governor started scaling, that copy would have explained the current mode with numbers that did not produce it. Both now call `config.EffectiveThreshold`, and a table test asserts the gauge equals what the governor resolves across seven configurations (defaults, hand-tuned, mixed, none, sqrt, cadence-only mode entries, no repos) — comparing two independently reached answers rather than restating one.

## Wiring

`Governor.SetRepoCount` is a setter rather than a `New`/`UpdateConfig` parameter: the count comes from `config.Project`, not `GovernorConfig`, and moves on a different schedule (a dashboard repo edit doesn't touch governor settings). It also avoids churning 68 `New(...)` call sites in the existing tests. It sits beside the existing `ghClient.SetRepos` on the reload path, matching that precedent, and is called from all three places the governor learns about config: startup, config reload, and dashboard save.

`repoCount` defaults to **1** in `New`, so a governor built without one behaves exactly as before.

## Dashboard

- `GET /api/config/governor` now also returns `effectiveThresholds`, `thresholdScaling`, and `repoCount`.
- New `PUT /api/config/governor/threshold-scaling` — a separate route because the existing thresholds endpoint takes a flat `map[string]int`, which a string curve doesn't fit.
- **Governor → Thresholds** gains a scaling selector and a line showing the values actually in force. This matters: on a 39-repo hive the effective surge is 780, far past the slider's 200-column range, so a slider reading 20 with no further explanation would be a lie. The note also states that dragging a handle pins that threshold and opts it out of scaling.

## Tests

`pkg/config/threshold_scaling_test.go` — curves incl. boundaries and negative repo counts; explicit-never-scaled at 1/39/500 repos; zero-means-unset; unknown modes get no invented default; the mixed/inverted fixture; validation incl. near-misses (`Linear`, `on`, `true`).

`pkg/governor/governor_threshold_scaling_test.go` — repo count reaches `thresholdFor`; **the reported scenario end to end** (39 repos at queue 210 is SURGE before, not after); equal per-repo pressure yields the same mode at 3 and 39 repos; hand-tuned hives unaffected at any repo count; inversion warning fires from both `SetRepoCount` and `UpdateConfig`, carries the numbers, and stays quiet on an unchanged count (it runs on every reload).

`pkg/dashboard/governor_threshold_gauge_test.go` — gauge/governor agreement, described above.

### Regression replay

Each change neutered (still compiling), the specific test confirmed failing, then restored green.

**1. Explicit thresholds scaled too** — the change that would break every hand-tuned hive:
```
--- FAIL: TestEffectiveThreshold_ExplicitIsNeverScaled
    repos=39: explicit surge = 11700, want 300
    repos=500: explicit surge = 150000, want 300
```

**2. Repo-count floor removed:**
```
--- FAIL: TestScaleThreshold_Linear
    ScaleThreshold(20, 0, linear) = 0, want 20
    ScaleThreshold(20, -5, linear) = -100, want 20
```

**3. Governor ignores the repo count:**
```
--- FAIL: TestSetRepoCount_ScalesDefaultThresholds
    39-repo surge = 20, want 780
--- FAIL: TestComputeMode_LargeHiveLeavesPermanentSurge
    39-repo hive at queue 210 is still SURGE — the reported bug is not fixed
```

**4. Dashboard gauge restored to its own copy of the defaults:**
```
--- FAIL: TestBuildGovernor_GaugeAgreesWithGovernor/defaults,_many_repos
    gauge surge = 20, governor uses 780
--- FAIL: TestBuildGovernor_GaugeAgreesWithGovernor/sqrt_scaling
    gauge surge = 20, governor uses 140
```

**5. Inversion warning removed:**
```
--- FAIL: TestSetRepoCount_WarnsOnInvertedLadder
    no inversion warning logged; got ""
--- FAIL: TestUpdateConfig_WarnsOnInvertedLadder
```

**6. `threshold_scaling` validation removed:**
```
--- FAIL: TestValidate_RejectsBadThresholdScaling
    validate() accepted an invalid threshold_scaling
```

**7. sqrt curve degraded to linear** (the option becomes a lie):
```
--- FAIL: TestScaleThreshold_Sqrt
    ScaleThreshold(20, 39, sqrt) = 780, want 140
--- FAIL: TestScaleThreshold_SqrtIsNeverHarsherThanLinear
```

### Local verification

```
go build ./...                clean
gofmt -l <changed files>      clean
go vet <changed pkgs>         clean
go test ./pkg/governor/       ok  0.02s
go test ./pkg/config/         ok  2.6s
go test ./pkg/dashboard/      ok  51.1s
go test ./pkg/scheduler/      ok  0.02s
go test ./...                 4 pre-existing failures in cmd/bd
```

`cmd/bd` (`TestCmdCloseAndReopenViaUpdate`, `TestCmdUpdateMetadata`, `TestCmdUpdateClaim`, `TestCmdReset`) fails identically on a clean checkout — verified by stashing every change and re-running. It is the beads CLI, untouched here.

## Compatibility

Defaults-only, as the issue intended. Unchanged for: single-repo hives, hives with explicit thresholds (including every pack-applied hive), and anyone setting `threshold_scaling: none`.

**Changed for:** multi-repo hives running on the code defaults — which is the point. Such a hive will see its thresholds rise and its mode drop, likely on the first eval after upgrade. The dashboard shows the effective numbers and the repo count so the change is legible rather than mysterious.

### DIFF
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -1095,6 +1095,10 @@ func main() {
 	// Dashboard login now requests no scope and no user write-token is persisted.
 
 	gov := governor.New(cfg.Governor, cfg.EnabledAgents(), logger)
+	// Default mode thresholds scale with how many repos the hive watches, so
+	// the mode ladder means the same thing on a 3-repo hive as on a 39-repo
+	// one (#3498). Explicit thresholds are unaffected.
+	gov.SetRepoCount(cfg.Project.RepoCount())
 	sched := scheduler.New(cfg, logger)
 
 	// Wire the GitHub prompt-source resolver so agents may source their kick
@@ -2618,6 +2622,9 @@ func main() {
 		// Re-sync subsystems that cache config values
 		ghClient.SetRepos(cfg.Project.Repos)
 		gov.UpdateConfig(cfg.Governor)
+		// A reload can add or archive repos, which moves every scaled default
+		// threshold — re-sync it alongside the repo list above.
+		gov.SetRepoCount(cfg.Project.RepoCount())
 		agentMgr.SetSandboxConfig(cfg.AgentSandbox)
 
 		// Re-apply live agent definitions (definition_source) on reload so an
--- a/src/docs/README.md
+++ b/src/docs/README.md
@@ -42,6 +42,7 @@ Start with [Architecture](architecture.md) for the system overview, then use the
 ## Configuration and agents
 
 - [Agent configuration](agent-configuration.md) — agent fields, methods, models, pins, cadences, caveman mode, and ACMM packs.
+- [Governor mode thresholds](https://github.com/kubestellar/hive/blob/v4/src/docs/governor-thresholds.md) — how idle/quiet/busy/surge thresholds scale with repo count, the `threshold_scaling` curves, and when explicit thresholds win.
 - [Supervisor agent](https://github.com/kubestellar/hive/blob/v4/src/docs/supervisor.md) — supervisor policy modes, bead roles, and when to enable the orchestration lane.
 - [Custom dashboard stylesheets](https://github.com/kubestellar/hive/blob/v4/src/docs/custom-stylesheets.md) — operator-supplied CSS for the dashboard and public snapshot.
 - [Portable AgentDefinition format](https://github.com/kubestellar/hive/blob/v4/src/AGENT-DEFINITION.md) — standalone YAML schema for importing/exporting agent definitions.
--- a/src/docs/agent-configuration.md
+++ b/src/docs/agent-configuration.md
@@ -229,7 +229,7 @@ Pin a model when reproducibility matters more than the governor's budget optimiz
 
 ## Cadences and the governor
 
-Agents don't schedule themselves. The **governor** evaluates the work queue every `eval_interval_s` (default 300s), computes a mode from queue depth — **idle → quiet → busy → surge** (default thresholds: quiet > 2, busy > 10, surge > 20; override with `threshold:`) — and kicks each agent on the cadence that mode assigns it:
+Agents don't schedule themselves. The **governor** evaluates the work queue every `eval_interval_s` (default 300s), computes a mode from queue depth — **idle → quiet → busy → surge** (default thresholds: quiet > 2, busy > 10, surge > 20 **per watched repo**; override with `threshold:`) — and kicks each agent on the cadence that mode assigns it:
 
 ```yaml
 governor:
@@ -250,6 +250,7 @@ governor:
       architect: pause     # "pause" stops kicks for this agent in this mode
 ```
 
+- **Thresholds scale with repo count.** The *default* thresholds above are per-repo bases, multiplied by `len(project.repos)` — so `surge` is 20 on a 1-repo hive and 780 on a 39-repo one, and the mode ladder means the same thing at any hive size. A `threshold:` you set yourself is used verbatim and never scaled. Tune the curve with `governor.threshold_scaling` (`linear` default, `sqrt`, `none`). See [Governor mode thresholds](governor-thresholds.md), which also covers the ACMM-pack interaction.
 - **Mutually exclusive modes.** Each per-agent, per-mode cadence is either an interval (`5m`, `2h`, `pause`) or a time-of-day schedule — never both. Config load and API writes reject mixed forms with a 400/error.
 - **Time-of-day schedules.** Use `times: ["HH:MM"]` with optional `days` (`mon` … `sun`) and a required IANA `tz`. The timezone is stored explicitly and displayed with the schedule; it does not float with the viewer.
 - **Advanced cron.** Power users can provide a constrained five-field cron expression plus `tz`. Hive evaluates these with robfig/cron and schedule-local timezone handling.
--- a/src/docs/governor-thresholds.md
+++ b/src/docs/governor-thresholds.md
@@ -1,0 +1,106 @@
+# Governor mode thresholds and repo-count scaling
+
+The governor picks a mode — **idle → quiet → busy → surge** — by comparing queue depth (actionable issues + open PRs, summed across every repo the hive watches) against three thresholds. The mode then selects each agent's kick cadence, so the thresholds are what decide how hard the hive works.
+
+## The problem the scaling solves
+
+Queue depth scales with how many repos a hive watches; the thresholds did not. A fixed `surge: 20` therefore encoded an implicit repo count:
+
+- A 39-repo hive naturally carries a queue in the low hundreds. Against absolute thresholds it sat in **SURGE permanently**, running the most aggressive cadences even though the per-repo backlog was small. Observed live: queue ~210 against a threshold of 70.
+- A 3-repo hive with the same thresholds would idle through a backlog that is genuinely deep *per repo*.
+
+Every hive with a different repo count had to hand-tune, and re-tune whenever repos were added or archived.
+
+## What happens by default
+
+The **default** thresholds are now per-repo base values, multiplied by the number of repos in `project.repos`:
+
+| | base (1 repo) | 3 repos | 39 repos |
+|---|---:|---:|---:|
+| `surge` | 20 | 60 | 780 |
+| `busy` | 10 | 30 | 390 |
+| `quiet` | 2 | 6 | 78 |
+
+A hive watching one repo (or none — `primary_repo` still counts as one) gets exactly the numbers it got before, so single-repo hives see no change at all.
+
+The effect is that the mode ladder means the same thing at any hive size: **surge is "20+ items per repo"** rather than "20 items, however many repos you watch".
+
+## Choosing a curve
+
+```yaml
+governor:
+  threshold_scaling: linear   # linear (default) | sqrt | none
+```
+
+| Value | Factor | When to use |
+|---|---|---|
+| `linear` (default) | `× repo_count` | The general case. Exactly equivalent to comparing per-repo queue pressure against the base thresholds. |
+| `sqrt` | `× ceil(√repo_count)` | Hives whose queue depth does not grow linearly with repo count — many small, quiet repos alongside a few busy ones. Reaches surge sooner than linear. |
+| `none` | `× 1` | Treat the thresholds as absolute queue depths. This is the behavior from before scaling existed. |
+
+For 39 repos, `linear` surges at 780 and `sqrt` at 140 (`ceil(√39) = 7`). Note that `sqrt` and `linear` are identical at 1 and 2 repos and only separate from 3 onward, because the ceiling cannot produce a factor below 2 until 5 repos.
+
+An unrecognized value is rejected at config load.
+
+## Explicit thresholds always win
+
+A threshold you set yourself is used **verbatim and never scaled**:
+
+```yaml
+governor:
+  modes:
+    surge:
+      threshold: 300     # used as-is on a 39-repo hive, not 300 × 39
+```
+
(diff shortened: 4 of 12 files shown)


Danathar:
Pushed `efb81ac` — a self-review pass against current `v4`, not a response to feedback. The branch was already up to date with `v4` tip, so this is additive on top of the original commit.

**One real defect found and fixed.** The inversion warning I added is re-checked by `UpdateConfig`, which runs on `main.go`'s ~2-minute reload loop *and* on every dashboard config save, regardless of whether anything changed — so a hive with an inverted ladder wrote an identical `WARN` every couple of minutes, forever.

It lands hardest exactly where it shouldn't: an explicit `surge` beside a scaled `busy` is the reporter's own configuration on this issue, so the noisiest log would have been on the hives this change is meant to help. It also runs against the norm the neighbouring reload code states out loud — that path is deliberately kept at `DEBUG` "to avoid spamming a healthy hive". `SetRepoCount` was already guarded against this (it early-returns on an unchanged count); `UpdateConfig` had no equivalent.

The warning is now deduplicated on the ladder it describes: silent when re-checked unchanged, warns again on a *different* inversion (an operator who edits one threshold and makes it worse has to hear that), and warns again after a repair-then-re-break — the memo is cleared while the ladder is healthy, so it's a dedupe rather than a one-shot.

**Coverage gap closed.** Nothing asserted the repo-count path end to end — the governor tests drive `SetRepoCount` directly, the gauge tests never go through a handler. Two tests now do it through the observable mode, including that a *rejected* repo edit leaves the count alone (that path restores the previous list and returns before saving).

Regression replay for each: dedupe removed → the spam test fails at 10 duplicate warns; dedupe latched to a one-shot → both the changed-inversion and repair-then-reinvert tests fail. Restored green. `pkg/governor`, `pkg/config` and `pkg/dashboard` all pass.

One thing I looked at and deliberately did **not** change, flagging it since it decides how much of this issue actually lands in practice: ACMM packs at level 3–6 seed explicit `surge`/`busy`/`quiet` (`pkg/config/packs/level-*.yaml`, applied via `api_packs.go` and `applyConfigOverrides`), and rule 1 of `EffectiveThreshold` can't tell a pack-seeded value from a hand-tuned one — so on any hive that applied a level-3+ pack, scaling never engages. It's documented as a known limitation in `docs/governor-thresholds.md`.

Making pack-seeded thresholds scale is doable (record the value the pack wrote; if the live threshold still equals it, treat it as a default and scale it — an operator edit makes them differ and it reverts to explicit-wins), but it's a new config surface and a judgement call about whether a pack's numbers are policy or a starting point. That's yours to make, and I'd rather ask than assume. Happy to do it here or in a follow-up if you want it.

clubanderson:
Maintainer review — cleared for merge. Decisions on the three open calls, then verification.

## Maintainer decisions

1. **Linear default accepted.** The empirical argument holds: `ceil(sqrt(39)) = 7` puts surge at 140, so the reported 39-repo hive at queue ~210 would *stay* in SURGE under sqrt — the exact symptom #3498 reports. Linear is also the per-repo normalization the issue's own "Alternatives considered" section describes (`queue > base * repos` is `queue/repos > base`), so it is the principled default, not merely the simpler one.
2. **`threshold_scaling: linear | sqrt | none` accepted** over the issue's proposed `auto_scale_thresholds` boolean. One enum has no contradictory combinations; two knobs do (`auto_scale_thresholds: false` + `threshold_scaling: sqrt` means nothing). Keeping the spelling as-is.
3. **Warn-don't-reorder accepted** for the ladder-inversion hazard. The inversion originates in a number the operator set; silently overriding it would be the worse surprise. The dedup-on-ladder follow-up commit addresses the reload-spam concern.

## Verification (all confirmed in the diff)

- **Identity at n=1**: `ScaleThreshold` at repoCount 1 is the identity for every curve, and `TestEffectiveThreshold_IdentityAtOneRepoForEveryCurve` pins surge=20 / busy=10 / quiet=2 at repos=1 for linear, sqrt, and none (plus the unset default). A one-repo hive is byte-identical to the old constants.
- **Explicit-wins**: `TestEffectiveThreshold_ExplicitIsNeverScaled` pins hand-set 300/200/100 verbatim at 1/39/500 repos; `TestEffectiveThreshold_ZeroMeansUnset` keeps threshold `0` = unset (cadence-only mode entries still get the scaled default).
- **Floor never 0**: `ProjectConfig.RepoCount()` floors at 1 for `primary_repo`-only hives, and `ScaleThreshold` clamps `repoCount < 1` to 1 independently — a zero threshold pinning SURGE is unreachable from both directions (`TestRepoCount_FloorsAtOne`, the repos=0/-5 rows in `TestScaleThreshold_Linear`).
- **Operator visibility**: the dashboard gauge (`buildGovernor`) now resolves through `config.EffectiveThreshold` — the same call the governor ladders on — so the mode gauge shows the numbers that actually produced the mode; the Governor → Thresholds tab shows the effective values in force ("In force for N watched repos: quiet/busy/surge …") plus the scaling selector. The inversion condition is visible as the out-of-order effective values there, in addition to the deduplicated WARN.
- **Pack-applied hives untouched**: `api_packs.go` threshold seeding is not modified in this PR — pack hives keep current behavior exactly. That gap is real and gets its own follow-up issue (provenance for pack-seeded vs operator-set thresholds); see the PR body's ACMM section.

## Rebase note

The branch predated the v2/ → src/ tree rename (#3996), so it was rebased as a maintainer edit: mechanical path rewrite (v2/ → src/, imports `kubestellar/hive/v2/…` → `kubestellar/hive/…`), docs/README.md link-style conflict resolved to the current convention, one identity-at-1 test added covering all three curves, authorship and DCO preserved. `go build ./...`, `go vet`, and `go test -run 'Threshold' ./pkg/governor/ ./pkg/config/ ./pkg/dashboard/` all pass on the rebased head; the two unrelated `gh_app_token_script_test.go` failures reproduce identically on clean origin/v4.

Using `Refs` rather than `Fixes` was the right call given the two open questions — both now resolved above. #3498 will be closed after merge with a pointer to the pack-seeding follow-up, since that limitation plausibly covers the reporter's own fleet shape.

Issue #2172, opened 2026-07-29
Security: harden default deployment & autonomy model (external review)


clubanderson:
## Security review — hardening the default deployment & autonomy model

An external security review (by @jonpspri) of `hive` @ `9ef8f43` surfaced a set of issues concentrated in **default deployment configuration** and a handful of specific endpoints. The reviewer's overall read: *"This is not a naive codebase … the failures are concentrated in default deployment configuration"* — i.e. these are fixable oversights, not architecture. This issue tracks the fixable items; fixes are linked below as PRs land.

> **Note:** One additional finding involving the hosted hub's trust fabric is being handled through **private/coordinated disclosure** rather than described here, per the reviewer's recommendation. It is not part of this public tracker.

### Tracked findings

| # | Severity | Title | Status |
|---|----------|-------|--------|
| F1 | Critical | Dashboard token injected into `index.html` for every visitor (`proxy/server.js` `serveIndex`) — dashboard "auth" reduces to *can you reach the port* | ✅ #2178 |
| F2 | Critical | `:3002` trusts bare `X-Hive-User`/`X-Hive-Role` on non-direct-route (default self-hosted) deployments — anything on the pod network can self-declare identity | ⏳ in progress — needs coordinated hub(v2)+spoke(v3) change, deployed hub-first to avoid locking out hosted hives |
| F3 | Critical | Unauthenticated writable web terminal: `ttyd -W -a` (no `-c` credential) + nginx `stream{}` raw-proxy on `:7681` | ✅ #2178 |
| F5 | High | Contributor identity takeover: `POST /api/contribute/register` with `force:true` reissues **and returns** an existing contributor's token with no proof of identity | ✅ #2178 |
| F6/F7 | High | Autonomy: agents instructed to `gh pr merge --admin`; the only deterministic control is a bash-glob wrapper (denylist). Attacker-controlled issue/PR text flows into agent work lists | ⏳ deferred — carries agent-blocking risk; needs careful design + live agent verification |
| F8 | High | Discord bot: any guild member can inject prompts into agents (no role allowlist) | ✅ #2178 (opt-in allowlist) |
| F9 | Medium | `merge-gate.sh` has no hold-label check — hold-gating lives only in agent prompts, not code | ✅ #2178 |
| F10 | Medium | Deployment hygiene: k8s runs as root (no `securityContext`), tokens in plaintext env, logscrub misses `ghu_`/`ghr_` prefixes | ✅ #2178 |
[...]

PR #3899, opened 2026-08-14
🐛 fix(kick,docs): stop instructing agents to read the shared full-privilege App token cache

Danathar:
## What

Removes the last place in the repo that **instructs agents to read the shared, full-privilege GitHub App token cache**, and records the blocker that gates the rest of #1861.

Refs #1861 — **no `Fixes` keyword.** This is the small independent fix the issue's own triage carved out; the architectural item (proxy-side token injection) is deliberately **not** implemented, for the reason in the last section. The issue should stay open.

## Scope, per the issue's triage

The maintainer triage on #1861 re-scoped it to three outcomes. I verified each against `origin/v4` before touching anything:

| Item | Triage claim | My verification | Action |
|---|---|---|---|
| 2 — tighten shared cache to `0600` | ✅ already shipped | Confirmed: `bin/gh-app-token.sh:60` `chmod 600`; `pkg/github/app.go:31,56` `0o600`; `pkg/agent/manager.go:630` `0o600` | Dropped from scope |
| 1 — proxy-side injection | ⛔ blocked on identity (N7), sequence after #3833 | Confirmed, and I agree — see below | **Not implemented** |
| 🔨 carve-out — `kick-agents.sh:834` | "Safe, self-contained, no design needed" | Confirmed still present, reaching **5** kick messages | **This PR** |

## The bug

`bin/kick-agents.sh:834` built one `_GH_AUTH_INSTR` string embedded in the scanner, ci-maintainer, architect, outreach, and supervisor kicks:

```
⚙ GH AUTH: ALWAYS prefix gh commands with:
  GH_TOKEN=$(cat /var/run/hive-metrics/gh-app-token.cache) gh ...
```

That file holds the **full, unscoped installation token**. Reading it skips the credential helper's UID/mode gate and hands any agent full privilege — precisely the escalation #1861 exists to close, and precisely the fallback `bin/gh-wrapper.sh:59-81` already refuses, with a comment citing audit H3 / CWE-522,732:

> there is deliberately NO fallback to the shared full-privilege installation-token cache … falling back to either would silently escalate every agent to full privilege

Meanwhile `pkg/scheduler/scheduler.go:734` has told agents the opposite since #1860:

> Never point agents at the shared `gh-app-token.cache`: reading it skips the credential helper's UID/mode gate and hands every agent an unscoped token.

**The two kick lanes were instructing agents to do opposite things.** `install.sh:93` installs `kick-agents.sh`, so the v1 lane is live, not dead code.

### Why it still matters at `0600`

Item 2 already landed, so this `cat` now fails for a non-root agent rather than escalating. It is still worth fixing:

1. **It teaches the path.** An agent told to read that file will try it with `curl` or `git`, where no wrapper intervenes — and #1861's "Why now" is literally an agent discovering this read under pressure when its push failed.
2. **It fails confusingly.** `GH_TOKEN=$(cat …)` on an unreadable file yields an *empty* `GH_TOKEN` prefix, not an error the agent can act on.
3. **It contradicts the other lane**, which is how instructions drift back.

## The fix

`_GH_AUTH_INSTR` (a single string) becomes `_gh_auth_instr <agent>` (a function). Each of the five call sites passes its own agent name, so each kick names **that agent's** `0600`, UID-owned, tier-scoped token file — and names the shared cache only to forbid it. Wording tracks `ghAuthInstructions` so the lanes now agree.

It leads with "run `gh` normally": `v2/Dockerfile:348` and `Dockerfile.contributor:177` both `COPY bin/gh-wrapper.sh /usr/local/bin/gh`, and `hive-deploy.sh:105` syncs it there, so on those hosts the wrapper already injects the scoped token and the agent needs no token material at all — #1861's actual goal, reachable today for `gh`. The explicit `export` is kept as the fallback because `install.sh:98` installs the wrapper as `gh-wrapper.sh` **only**, not as `gh`; on an `install.sh`-provisioned host "just run gh" would leave agents unauthenticated.

## Tests

New `bin/kick-agents-gh-auth.test.sh` (23 assertions), following the `bin/gh-wrapper.test.sh` convention. It extracts just the helper rather than sourcing `kick-agents.sh`, which reads metrics and talks to tmux at load time.

Beyond the obvious "no shared-cache instruction", it pins two things worth calling out:

- **The token path must stay a literal for the agent.** If `$(cat …)` were expanded while *building* the kick, the hive would read the file as root and paste a live token into a tmux pane and the kick audit log — turning a prompt fix into a credential leak.
- **Each kick must pass its own agent name**, tied back to the `[agent:NAME]` tag of the message it sits inside.

### Regression replay

**Replay 2 initially did not fail, and that was a real gap in my test.** I reverted `outreach`'s call site to `scanner` — a plausible copy-paste across five hand-edited sites, and a cross-agent credential path — and the suite passed 22/22, because the helper renders correctly for whatever name it is handed. Added check 6 to tie each call site to its message's `[agent:…]` tag; it now fails. Recording it because the gap was in the test I wrote to guard the change I made.

**1. Original instruction restored verbatim:**
```
✗ no kick instruction tells an agent to cat the shared App token cache
✗ scanner: instruction names its own token file
✗ scanner: instruction explicitly forbids the shared cache
   … Result: 5 passed, 17 failed
```

**2. Cross-agent leak — outreach handed scanner's token file** (after adding check 6):
```
✗ every kick passes its own agent name (no cross-agent token path)
    line 972: [agent:outreach] kick passes 'scanner'
   Result: 22 passed, 1 failed
```

**3. Builder expands the token at build time** (`\$(cat` → `$(cat`):
```
✗ scanner: instruction names its own token file
✗ token path stays a literal command for the agent (no token pasted into the kick)
   Result: 16 passed, 6 failed
```

### Local verification

```
bash bin/kick-agents-gh-auth.test.sh    23 passed, 0 failed
bash -n bin/kick-agents.sh              clean
shellcheck -S warning <new test>        clean
bash bin/test_gh_wrapper_gates.sh       46 passed, 0 failed
cd v2 && go build ./...                 clean (no Go changed)
```

`bin/gh-wrapper.test.sh` fails (`FAIL: issue list --author test-bot[bot]`) — **identically on a clean checkout**, verified by stashing every change. Pre-existing, unrelated.

## ⛔ Why proxy-side injection (item 1) is not in this PR

I verified the blocker rather than taking the triage on faith, and it holds.

`identifyAgentFromReq` (`pkg/proxy/github_proxy.go:648`) tries unforgeable UID lookup first, then falls back to `extractAgentName(r)` — which reads the **client-supplied `Proxy-Authorization` header** (`:1197`). The fallback is not an edge case:

```go
if p.uidMap != nil {
    if name := p.identifyAgentByUID(r.RemoteAddr); name != "" { return name }
}
return extractAgentName(r)   // Proxy-Authorization — agent-controlled
```

When no UID map exists, `manager.go:805` logs *"no UID map found, agents will share dev UID"* — so UID lookup **cannot** distinguish agents and the forgeable header is the only signal for every request.

Injecting there would mint a tier-scoped token for whoever the proxy *believes* is calling. Today spoofing the header misstates the **mode** used for a request; with injection it would yield **another tier's token**. That converts a name-spoofing weakness into a token-minting one — strictly worse than the shared-cache read this issue set out to eliminate. Per-agent token files are `0600` owned by the agent's UID, so no credential crosses over today; injection is what would change that.

Building it now would also mean inventing a second, weaker identity story that the master-delivery per-spoke keypair (#3833) would supersede — the triage's sequencing point, which I think is right.

So instead of code, item 1 gets **documentation**: a new residual-risk entry in `v2/docs/security-threat-model.md` recording the fallback, its bounded impact today, why #1861's injection is deferred, and the N7 (#2172) / #3833 dependency. That analysis lived only in an issue comment; the threat model is where the next person will look.

## One further finding, deliberately not fixed here

`bin/contributor-relay.sh:46-48` defaults its token cache to the **hive's** shared App-token path when `/var/run/hive-metrics` exists, and `injectGhToken` (`:222-226`) *writes* to it. Inside the contributor container that directory is created deliberately (`Dockerfile.contributor:170`, `contributor-agent.sh:323`) so it is intended there — but a relay run natively (`just contribute-hive … local`) on a host that is also a hive spoke would **overwrite the hive's App token** with a hub-delivered task token, and every hive-side reader (`hive-config.sh:202`, `dashboard/*.sh`) would then use the wrong credential.

I left it alone: the triage attributes it to **N14 of #2172**, not this issue, and while tracing consumers I found `gh-wrapper.sh` skips token injection entirely in contributor mode (`if ! _contributor_mode`), so it is unclear what reads that file in the container. Changing the path blind could break the contributor flow. It needs the contributor-flow owner, not a drive-by.

### DIFF
--- a/bin/README.md
+++ b/bin/README.md
@@ -83,3 +83,4 @@ Most production scripts are installed under `/usr/local/bin` by `bin/hive-deploy
 | `contributor-agent.test.sh` | Contributor-agent regression for knowledge export handling. |
 | `contributor-relay.test.js` | Contributor relay task/restart/headless behavior; loads `contributor-relay.sh` as JavaScript with stubs. |
 | `gh-wrapper.test.sh` | `gh-wrapper.sh` author-gate and restriction regressions using a mock `gh` binary. |
+| `kick-agents-gh-auth.test.sh` | `kick-agents.sh` GH AUTH instruction (#1861): kicks must point each agent at its own tier-scoped token file, never the shared full-privilege App token cache. |
--- a/bin/kick-agents-gh-auth.test.sh
+++ b/bin/kick-agents-gh-auth.test.sh
@@ -1,0 +1,168 @@
+#!/usr/bin/env bash
+# Regression tests for kubestellar/hive#1861: the kick messages must never
+# instruct an agent to read the shared GitHub App token cache.
+#
+# /var/run/hive-metrics/gh-app-token.cache holds the FULL, unscoped
+# installation token. Reading it skips the credential helper's UID/mode gate
+# and hands any agent full privilege — the escalation #1861 exists to close
+# (audit H3, CWE-522/732). The file is 0600 today, so the old instruction can
+# no longer succeed for a non-root agent, but an agent TOLD to reach for it
+# will try the same path with curl or git, where no wrapper intervenes.
+#
+# bin/gh-wrapper.sh already refuses that fallback and fails loud, and the v2
+# scheduler (v2/pkg/scheduler/scheduler.go ghAuthInstructions) already tells
+# agents the right thing. These tests hold the v1 kick lane to the same rule so
+# the two lanes cannot drift apart again.
+#
+# Run: bash bin/kick-agents-gh-auth.test.sh
+
+set -euo pipefail
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+KICK="${ROOT_DIR}/bin/kick-agents.sh"
+SHARED_CACHE="/var/run/hive-metrics/gh-app-token.cache"
+PASSED=0
+FAILED=0
+
+pass() { printf '  ✓ %s\n' "$1"; PASSED=$((PASSED + 1)); }
+fail() { printf '  ✗ %s\n' "$1"; printf '    %s\n' "${2:-}"; FAILED=$((FAILED + 1)); }
+
+# The agents whose kick messages carry a GH AUTH line. Kept explicit rather
+# than derived, so ADDING an agent without an auth line is a visible diff here.
+AGENTS=(scanner ci-maintainer architect outreach supervisor)
+
+printf '\n=== kick-agents.sh GH AUTH instruction (#1861) ===\n\n'
+
+# --- 1. the script is valid bash -------------------------------------------
+if bash -n "$KICK" 2>/dev/null; then
+  pass "kick-agents.sh parses"
+else
+  fail "kick-agents.sh parses" "$(bash -n "$KICK" 2>&1 | head -3)"
+fi
+
+# --- 2. no agent-facing instruction names the shared cache ------------------
+# Scoped to the auth helper and the kick message bodies. The script may still
+# reference the shared cache in hive-side plumbing that runs as the hive, which
+# is legitimate; what must never happen is TELLING AN AGENT to read it.
+if grep -n "cat ${SHARED_CACHE}" "$KICK" >/dev/null 2>&1; then
+  fail "no kick instruction tells an agent to cat the shared App token cache" \
+    "found: $(grep -n "cat ${SHARED_CACHE}" "$KICK" | head -3)"
+else
+  pass "no kick instruction tells an agent to cat the shared App token cache"
+fi
+
+if grep -n "GH_TOKEN=\\\$(cat ${SHARED_CACHE}" "$KICK" >/dev/null 2>&1; then
+  fail "no kick instruction exports GH_TOKEN from the shared cache" \
+    "found: $(grep -n "GH_TOKEN=\\\$(cat ${SHARED_CACHE}" "$KICK" | head -3)"
+else
+  pass "no kick instruction exports GH_TOKEN from the shared cache"
+fi
+
+# --- 3. the auth helper renders a per-agent path ----------------------------
+# Extract just _gh_auth_instr rather than sourcing kick-agents.sh, which does
+# real work (reads metrics, talks to tmux) at load time.
+HELPER="$(mktemp)"
+# shellcheck disable=SC2329 # invoked via EXIT trap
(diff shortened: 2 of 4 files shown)


Danathar:
## Superseded by #3889 — and my instruction is wrong on the native path

Flagging this myself so nobody spends review time on it.

**#3889 (opened earlier, N14/#3842) covers the same change to `bin/kick-agents.sh` and does it correctly**, where mine does not.

I hit the same constraint #3889 documents — `install.sh` ships `gh-wrapper.sh` under its own name, not as `gh`, so the wrapper does not intercept on a native install. My response was to keep an explicit fallback in the prompt:

```
export GH_TOKEN=$(cat /var/run/hive-metrics/agent-tokens/gh-token-<agent>.cache)
```

**That file does not exist on the path `kick-agents.sh` serves.** `HIVE_AGENT_TOKEN_CACHE` and the per-agent cache are minted only by the container-hosted Go AgentManager (`v2/pkg/agent/manager.go:6209`); a native/systemd install never creates them. So my replacement instruction points native agents at a missing file — it removes the dangerous instruction but leaves them with no working one. I read that line of `manager.go` and still drew the wrong conclusion from it.

#3889 solves it properly: `agent-launch.sh` runs `gh auth login --with-token` once at startup so `gh` is persistently authenticated, and the prompt reduces to "gh is already authenticated — just run gh." That is a fuller realization of #1861's goal (agents hold no token material) than mine.

**Recommendation: close this PR in favour of #3889.**

Two smaller pieces here are independent of the overlap, if a maintainer wants them carried elsewhere:

- `v2/docs/security-threat-model.md` — residual-risk entry for the forgeable `Proxy-Authorization` fallback. Note this **needs updating before it lands**: I wrote it as an open gap, but #3888 (N7/#3841) is actively closing that fallback, so the entry would be stale on merge.
- `bin/README.md` — test-index row, moot if this closes.

Leaving the close decision to a maintainer rather than doing it unilaterally.

Danathar:
closed due to be superceded

PR #3900, opened 2026-08-14
🐛 fix(advisory): collapse near-duplicate findings so the advisory digest stays readable

Danathar:
## What

#2364 is the pinned 🐝 Advisory Report — a living document agents append to every cycle. It has become hard to read: **317 findings**, and a large share of them are the same problems restated.

This fixes the mechanism that lets that happen. Measured on the digest comment as of 2026-08-14: **92 findings → 56**, with the largest pile collapsing into a single entry marked `(reported 32×)`.

Refs #2364 — no `Fixes`. The issue is explicitly a living document ("**Do not close this issue.**"), so it stays open; this improves what gets posted into it.

## The bug

`BuildDigestFromBeads` deduplicates on **exact title match**:

```go
if seen[b.Title] { continue }
seen[b.Title] = true
```

Agents re-report persistent problems every cycle and **reword the title each time**. Exact matching catches none of it, so the digest accumulates one entry per *report* rather than one per *problem*.

In the live digest, one already-fixed CI workflow appeared **32 times** under 32 different titles:

```
pr-verifier.yml failing on every PR — reusable workflow deleted from infra
pr-verifier.yml broken: reusable workflow missing from kubestellar/infra
pr-verifier.yml failing 100% — reusable workflow missing from kubestellar/infra
PR Verifier failing on all PRs — reusable-pr-verifier.yml deleted from kubestellar/infra
pr-verifier.yml: 100% failure rate — reusable workflow missing from kubestellar/infra
… 27 more
```

That is **35% of the digest** describing a single problem — and, as it happens, one that no longer exists: `.github/workflows/pr-verifier.yml` no longer calls the deleted reusable workflow, it runs a pinned `ghcr.io/kubestellar/pr-verifier` image. So a third of the report is noise about something already fixed, crowding out findings that are not.

## The fix

`collapseNearDuplicates` merges findings that share an **agent** AND a **finding type** AND reach **0.5 Jaccard similarity** on their title tokens.

The agent+type match is a cheap structural guard: a `scanner` security finding can never be absorbed into a `ci-maintainer` CI failure however the words line up.

### Threshold: chosen by measurement, not taste

I swept the real corpus and looked at what each threshold actually merges. Merges of genuinely **different** subjects last appear at **0.30**; 0.35 and above produce none.

| threshold | 92 findings → | cross-subject merges |
|---:|---:|---:|
| 0.25 | 45 | 2 ❌ |
| 0.30 | 46 | 1 ❌ |
| 0.35 | 49 | 0 |
| 0.40 | 53 | 0 |
| **0.50** | **56** | **0** |

0.5 keeps a **0.20 margin** over the observed boundary. The asymmetry is deliberate and is the main design decision here: a duplicate wastes a reader's attention, a wrong merge **hides a real defect**. So it is tuned for safety, not for maximum compression — 0.35 would compress more and I did not take it.

`TestNearDuplicateThreshold_HasSafetyMargin` fails if someone later lowers it chasing compression.

### Tokenizer: sub-word splitting measured strictly better

Tokens split on **every non-letter**, so `kubestellar/infra` → `{kubestellar, infra}` and digits vanish.

Both matter. Keeping path tokens whole meant `"…missing from kubestellar/infra"` and `"…deleted from infra"` shared **no token for the thing they are both about**, and failed to merge. And digits are exactly the part that changes on every re-report (`2,943 runs`, `100%`, `19 consecutive failures`) — keeping them makes reports of one problem look distinct.

Measured both ways on the corpus; sub-word splitting was better on **both** axes at every threshold:

| tokenizer | @0.50 | pr-verifier clusters | cross-subject merges @0.35 |
|---|---:|---:|---:|
| whole path tokens | 92 → 60 | 7 | 1 ❌ |
| **sub-word split** | **92 → 56** | **4** | **0** |

### Nothing is silently dropped

Two properties keep collapse from becoming concealment:

- **The representative keeps the highest severity in its group.** Otherwise a critical report could be merged into a medium restatement of itself and quietly demoted into a lower section of the digest.
- **The count is rendered**: `(reported 32×)`. A recurring problem still reads as recurring — arguably *better* than before, where recurrence was conveyed only by being exhausting to scroll past.

`TotalCount` is recomputed from the survivors in both builders, so the header count matches the list beneath it.

## Result on the real digest

```
live digest of #2364: 92 findings -> 56 after collapse (39% removed)
  reported 32×  [ci-maintainer] pr-verifier.yml failing on every PR — reusable workflow deleted from infra
```

## Tests

`v2/pkg/advisory/dedup_test.go`, plus `testdata/live_digest_2364.json` — the **actual 92 findings** parsed out of the pinned comment, so the corpus tests measure the real case rather than a fixture I invented.

Unit: reworded restatements merge; never across agent/type; highest severity wins; order preserved; nil/single-element inputs; tokenizer splits qualified names and drops digits.

Corpus: end-to-end reduction; the pr-verifier pile shrinks but **does not vanish** (the problem must still be reported once); and `TestCollapseNearDuplicates_NoCrossSubjectMergeOnLiveCorpus` re-runs the matching loop to inspect every individual merge decision against the real data.

Rendering: repeat marker appears for collapsed groups, absent for single reports, and `TotalCount` matches what is listed.

### Regression replay

Each change neutered, the specific test confirmed failing, then restored green.

**1. Collapse disabled (back to exact-title dedup):**
```
--- FAIL: TestCollapseNearDuplicates_MergesRewordedRestatements
      (all 5 pr-verifier restatements survive)
--- FAIL: TestCollapseNearDuplicates_LiveDigestCorpus
--- FAIL: TestBuildDigest_TotalCountMatchesCollapsedFindings
```

**2. Threshold lowered to 0.25 — the over-collapse this guards against, reproduced on real data:**
```
--- FAIL: TestCollapseNearDuplicates_KeepsDistinctFindings
    distinct findings were merged (kept 1, want 2):
      "v2 Tests: 9 consecutive failures all day 2026-08-06, regression on commit 1d89011"
      "TestAlertAcksPersistRoundTrip regression in pkg/hub — 2 consecutive failures"
--- FAIL: TestCollapseNearDuplicates_NoCrossSubjectMergeOnLiveCorpus
--- FAIL: TestNearDuplicateThreshold_HasSafetyMargin
```

**3. agent/type guard removed:**
```
--- FAIL: TestCollapseNearDuplicates_NeverMergesAcrossAgentOrType
--- FAIL: TestBuildDigest
```

**4. Severity not upgraded on merge:**
```
--- FAIL: TestCollapseNearDuplicates_KeepsHighestSeverity
```

### Local verification

```
go build ./...            clean
gofmt -l pkg/advisory     clean
go vet ./pkg/advisory     clean
go test ./pkg/advisory/   ok
go test ./pkg/beads/      ok
go test ./pkg/github/     ok  (41.9s — owns the digest posting path)
```

## A note on the findings themselves

While picking a target I verified several of the digest's **critical** findings against `origin/v4`, and the top ones are stale:

- *"pr-verifier.yml calls a deleted reusable workflow"* (×32) — fixed; it uses a pinned GHCR image.
- *"`noRedirectToPrivate` has zero dedicated unit tests"* — it has 7, in `ssrf_guard_test.go`.
- *"`stripAuthOnCrossHostRedirect` untested"* — I gutted the function to check; `TestF13_RedirectDropsAuthorizationCrossHost` fails, so it is genuinely covered.

That is the same disease as the duplicates: the digest reports what agents *found*, with no aging. This PR fixes the volume half, not the staleness half. A follow-up worth considering is expiring findings whose subject no longer reproduces — the existing `VerifyFindingPaths` does this for file paths (`_(file path not found at analyzed commit)_`), but a finding about a *workflow's contents* passes that check while being entirely obsolete.

## Compatibility

Additive. `Finding.DuplicateCount` is `omitempty`, so persisted digests are unchanged for single-report findings. Digests with no near-duplicates render byte-identically; `collapseNearDuplicates` returns its input untouched for fewer than two findings.

### DIFF
--- a/v2/pkg/advisory/advisory.go
+++ b/v2/pkg/advisory/advisory.go
@@ -28,6 +28,11 @@ type Finding struct {
 	Detail    string    `json:"detail,omitempty"`
 	File      string    `json:"file,omitempty"`
 	Line      int       `json:"line,omitempty"`
+	// DuplicateCount is how many ADDITIONAL reports of the same finding were
+	// collapsed into this one by collapseNearDuplicates (0 = reported once).
+	// It is rendered as "(reported N×)" so a recurring problem still reads as
+	// recurring — the count is signal, and collapsing must not destroy it.
+	DuplicateCount int `json:"duplicate_count,omitempty"`
 	// PathStale is set by VerifyFindingPaths when File names a repo file path
 	// that does NOT exist at the digest's analyzed snapshot (Digest.AnalyzedSnapshot).
 	// Such a finding was generated against a different/older tree than the one
@@ -151,11 +156,19 @@ func BuildDigest(findings []Finding, mode string) *Digest {
 	for _, f := range findings {
 		byAgent[f.Agent] = append(byAgent[f.Agent], f)
 	}
+	// Collapse restatements of the same finding (#2364). TotalCount must be
+	// derived from what SURVIVES, not from the input, or the header count
+	// disagrees with the list beneath it.
+	total := 0
+	for agent, fs := range byAgent {
+		byAgent[agent] = collapseNearDuplicates(fs)
+		total += len(byAgent[agent])
+	}
 	return &Digest{
 		GeneratedAt: time.Now(),
 		Mode:        mode,
 		ByAgent:     byAgent,
-		TotalCount:  len(findings),
+		TotalCount:  total,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


clubanderson:
Reviewed on its own merits, including against the neighboring #3927 which attacks the same digest-flooding problem from the render-cap side — this PR stands alone and doesn't depend on it. The core risk in any fuzzy dedup is over-collapse hiding a real defect, and this PR treats that as the design center rather than an afterthought: agent+type must match before similarity is even consulted, the 0.5 threshold is chosen with a measured 0.20 margin above the empirically observed cross-subject-merge boundary (0.30) on the live #2364 corpus, and TestNearDuplicateThreshold_HasSafetyMargin plus the no-cross-subject-merge corpus test turn that tuning into an invariant instead of a comment. Good calls throughout: highest severity wins the merge so a critical can't hide behind a medium restatement, "reported N×" preserves the recurrence signal the duplicates carried, TotalCount is recomputed from survivors so the header matches the list, and sub-word tokenization with digit-dropping is justified by the exact title pairs that failed to merge without it. Shipping the 554-line live digest as a fixture and asserting measured outcomes (92→≤60, the 29-report pr-verifier pile →≤ a handful) is the strongest kind of test for a heuristic. Merging; #3927's per-group render cap remains complementary as a hard size backstop and will need a rebase over this.

Issue #2547, opened 2026-08-04
[v5] contribute: the hub cannot tell what a connected client can run, and cannot route work on it


castrojo:
## Status update — 2026-08-04

**Observability shipped; capability declaration and routing did not.** Merged #2542 added the connected-fleet view, and #2558 added accurate trust state, idle reasons, task metadata, and prompt preview. Those surfaces report what the existing handshake and assignment state already know; there is still no client capability map.

The maintainer response accepts the **DECLARE / ROUTE** split: additive self-description is a reasonable next step and must not be treated as a trust signal; routing is a separate design problem that still lacks task-side requirements metadata. Merged #2562 mirrors existing admission controls only and explicitly makes no routing change. This issue remains open for DECLARE first, with ROUTE intentionally undecided.

#2567 tracks the prerequisite compatibility layer: bidirectional protocol/version capabilities and canonical spoke discovery before richer client declarations are relied on.

> Part of #2534 (contributor onboarding field notes from `projectbluefin/review`).

## Summary

When a contributor connects, the hub learns three things about the client: which CLI
backend it runs, which model, and (optionally) which role. Everything else about the
execution environment — container runtime, OS/arch, agent CLI version, credential type,
relay protocol version — is invisible to the hub, and the wire envelope has no place to
put it. We'd like to ask one design question, not propose a mechanism:

**Should the contributor protocol let a client declare what it can do, and should the hub
be allowed to act on that declaration when choosing what to assign?**

Those are two questions and they should be answered separately. Declaring is cheap and
additive. *Acting* on a declaration is a routing decision with real consequences, and it
depends on task-side metadata that does not exist yet.

**Related, not duplicated:** #2393 item 8 is where the reserved envelope field came from,
and the reasoning recorded there is the starting point for this issue rather than
something we are arguing against. `projectbluefin/review`'s #2537 asks *when* a credential
becomes available; this asks *what the hub knows about the thing receiving it*. #2540 is
the closest sibling and the distinction matters: **#2540 asks only that the hub be able to
see what its fleet is running (observability). This issue asks whether the hub should
*act* on what it sees (routing).** We deliberately filed the observability half first,
because it is the half we are confident about.

We are a downstream consumer — `projectbluefin/review` launches a contributor runtime that
speaks this protocol — so this is field-note framing, not a review. Everything below is
pinned at `957771e5f18604f3d301a3749f63b9896ba99e1c`.

## Evidence

[...]

PR #3901, opened 2026-08-14
✨ feature(contribute): record and surface why a contributor task failed (DECLARE, not ROUTE)

Danathar:
## What

Makes a contributor task failure **attributable**: the hub now records *why* a task failed and shows it, instead of logging the reason and dropping it.

Refs #2547 — **no `Fixes`.** This is deliberately **not** ROUTE. The issue stays open for that decision.

## Scope — why this is not ROUTE

Reading the issue and its comments, the state is unambiguous:

- **DECLARE shipped** (#2659, pinned by #3815, documented by #3830).
- **ROUTE is intentionally open.** The maintainer asked the reporter three design questions (requirement axes, where task requirements come from, hard vs soft) and said: *"we'd rather design the requirements schema around real fleet needs than guess."*

So ROUTE is blocked on a maintainer decision plus downstream field input. Implementing it here would be guessing at a schema someone explicitly asked not to have guessed, and would trip `TestSelectionPathsDoNotReadDeclaredCapabilities` — a test that exists precisely to stop a PR like that landing quietly.

What I built instead is the one operational cost in the issue body that is **independent of ROUTE**:

> "a task that failed because the client couldn't run it and a task that failed because the agent got it wrong are, from the hub's side, the same event with different terminal scrollback."

That needs no task-side requirements vocabulary, makes no dispatch decision, and produces exactly the fleet evidence the maintainer said they want *before* designing ROUTE.

## The gap

`task_failed` carries a `reason`. The handler uses it in one place — the log line — and nowhere else:

```go
h.logger.Info("[contribute-ws] task failed", …, "reason", msg.Reason, …)
```

It is never stored on the connection, the profile, or the failure ledger. Meanwhile `lastIdleReason` (#2546) already solved the identical problem for idleness and is surfaced as `FleetClanker.IdleReason`, explicitly so an operator can see *"why is this clanker idle"* rather than *"an indistinguishable silence."* Failures had no equivalent.

## What changed

**Wire (additive):** `task_failed` may carry `failure_kind`:

| value | meaning |
|---|---|
| `environment` | this client's runtime could not run the work (CLI never started, CLI crashed, backend has no headless mode) — the work item is unjudged |
| `task` | the work was attempted and failed on its merits |
| *absent* | normalizes to `unspecified` — what every existing relay sends, treated exactly as today |

**Hub:** stores reason + normalized kind + work item on the connection; surfaces the most recent one read-only as `FleetClanker.LastFailure`, mirroring `IdleReason`.

**Relay:** `failCurrentTask` takes an optional `kind`, and declares `environment` at the three sites where the cause is unambiguous — CLI never became ready, CLI process died, backend has no headless entry point. The ambiguous sites (headless non-zero exit, max-duration timeout) deliberately send **nothing**: an honest `unspecified` beats a guess when an operator reads this to attribute blame.

## The line this holds

Nothing acts on the declaration. That is the whole design constraint, and it is where the risk lives.

**The cooldown must not depend on it.** The obvious next step — "skip the failure cooldown for an `environment` failure, since the issue was fine" — is exactly the hazard the issue names: *routing on a value the client controls*. A relay could keep an issue permanently hot by tagging every failure `environment`, producing a fleet-level symptom that presents as a hub bug. So `recordTaskFailure` still computes cooldown and quarantine weight from repo, number and `permanent` alone. I did not take that step even though the issue's acceptance criteria gesture at it, because that criterion is conditioned on *"if any routing ships"* — and routing is not shipping here.

Two tests pin it, mirroring the existing #3815 source-level guard:

- `TestSelectionPathsDoNotReadFailureKind` — `selectTask` / `RequeueContributorTask` may not even mention the field.
- `TestRecordTaskFailure_IgnoresFailureKind` — the cooldown path may not mention it either.

Both carry the same instruction as the capability guard: *if routing has now been decided and recorded by a maintainer, update this test in that same PR.*

## Backward compatibility

Absent → `unspecified`, **never** either real kind. The issue is emphatic that *"a client that declares nothing must never be treated as declaring 'no'"*, and that compatibility has to be carried by the defaults since there is no negotiation to carry it. `TestNormalizeTaskFailureKind` asserts absent/blank/unknown all land on `unspecified`; `TestRecordTaskFailure_SameForTaggedAndUntagged` asserts identical ledger outcomes; `TestFleetSnapshot_NoFailureIsNil` keeps the payload identical for a clanker that never failed.

## Handling of client-controlled data

Two properties, because `reason` is the **only** client-controlled free text reaching this snapshot:

- **Redacted.** It is an error string the relay chose, so it can carry a token the relay happened to print. Run through the package's existing `redactTokens` at the boundary where it becomes operator-visible.
- **Bounded** to 512 runes with a visible `… (truncated)` marker, rune-safe so a cut cannot emit invalid UTF-8 into a JSON response. Nothing obliges a client to keep it short.

And `failure_kind` itself is normalized rather than echoed: an unknown value collapses to `unspecified` instead of passing arbitrary client text into a field that reads like an enum on an operator's screen.

## Tests

`v2/pkg/dashboard/contribute_failure_kind_test.go` — normalization (incl. case/space tolerance, and that hostile text does not pass through), truncation (bound, marker, rune safety), fleet surfacing, snapshot-is-a-copy, redaction+bounding, the two ROUTE guards, tagged-vs-untagged equivalence, and that the wire field stays `omitempty`.

### Regression replay

Each change neutered, the specific test confirmed failing, then restored green.

**1. Unknown client kind echoed verbatim** — client writes free text into the operator view:
```
--- FAIL: TestNormalizeTaskFailureKind
    NormalizeTaskFailureKind("hardware") = "hardware", want unspecified
    NormalizeTaskFailureKind("environmental") = "environmental", want unspecified
```

**2. Reason surfaced without redaction/bounding:**
```
--- FAIL: TestFleetSnapshot_FailureReasonRedactedAndBounded
    secret survived into the fleet view: "push failed: ANTHROPIC_API_KEY='sk-hive-scanner' …"
    reason is 561 runes — unbounded client text reached the operator view
```

**3. Cooldown made to depend on the declared kind** — the ROUTE hazard:
```
--- FAIL: TestRecordTaskFailure_IgnoresFailureKind
    recordTaskFailure references "failure_kind" — the cooldown must not depend on a
    client-declared value (#2547: routing on a value the client controls)
```

**4. Failure kind wired into `selectTask`:**
```
--- FAIL: TestSelectionPathsDoNotReadFailureKind
    selectTask references "lastfailure" — ROUTE is intentionally NOT implemented
```

My first attempt at replay 4 failed to *compile*, so the compiler caught it rather than the test. I redid it with a clean-compiling read so the test is demonstrably what fires — otherwise I'd have been claiming coverage the test hadn't shown.

### Local verification

```
go build ./...              clean
gofmt -l <changed files>    clean
go vet ./pkg/dashboard      clean
go test ./pkg/dashboard/    ok  51.6s  (full suite, incl. the existing DECLARE/ROUTE guards)
node --check contributor-relay.sh   clean
```

## What this does and does not give you

**Does:** an operator seeing a run of failures can tell environment from merit, per connection, with the reason preserved. That is the diagnostic cost in the issue's "Why this matters operationally", and it is the data needed to answer the maintainer's own question — *is the fit problem real and recurring on this fleet?* — before a routing schema is designed around a guess.

**Does not:** prevent the wasted assignment, or stop a work item accruing failure history for an environment problem. Both need ROUTE, and ROUTE needs the task-side vocabulary and the maintainer decision. This is Option B applied to failures, and it is offered on the same reasoning the reporter used for preferring B first: get the data before designing policy on top of it.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -480,7 +480,9 @@ function runHeadlessTask(task) {
     const reason = `backend '${BACKEND}' has no headless (non-interactive) mode; supported: ${Object.keys(HEADLESS_BACKENDS).join(', ')}`;
     console.error(`Headless dispatch refused: ${reason}`);
     writeHeadlessStatus(HEADLESS_STATE_FAILED, { task_id: task.task_id, reason });
-    failCurrentTask(reason, { permanent: true });
+    // environment: this relay's configured backend has no headless entry point;
+    // the work item itself is unjudged.
+    failCurrentTask(reason, { permanent: true, kind: 'environment' });
     return;
   }
 
@@ -746,7 +748,8 @@ function armCLIReadyWait() {
     // contributor silently working on someone else's issue.
     pendingTask = null;
     if (currentTask) {
-      failCurrentTask(`CLI never became ready: ${e.message}`, { skipReady: true });
+      // environment: the agent CLI never reached its prompt on this host.
+      failCurrentTask(`CLI never became ready: ${e.message}`, { skipReady: true, kind: 'environment' });
     }
     // Keep waiting. The CLI may still come up (a slow login, an operator
     // attaching to clear a prompt we don't recognize), and when it does the
@@ -1178,14 +1181,26 @@ function restartBackoffMs(attempt) {
   return Math.min(TASK_RESTART_BASE_BACKOFF_MS * Math.pow(2, attempt - 1), TASK_RESTART_MAX_BACKOFF_MS);
 }
 
+// failCurrentTask reports the active task as failed.
+//
+// opts.kind (kubestellar/hive#2547) optionally states WHY: 'environment' when
+// this client's own runtime could not run the work (the CLI never started, it
+// crashed, the backend has no headless mode) versus 'task' when the work was
+// attempted and failed on its merits. Omit it when the cause is genuinely
+// ambiguous — the hub normalizes absent to 'unspecified', and guessing would be
+// worse than saying nothing, since an operator reads this to attribute failures.
+//
+// It is advisory: the hub records and displays it and does not route, gate, or
+// change the work item's failure cooldown on it. Older hubs ignore the field.
 function failCurrentTask(reason, opts) {
   if (!currentTask) return;
   const permanent = !!(opts && opts.permanent);
+  const kind = (opts && opts.kind) || undefined;
   const taskId = currentTask.task_id;
   const taskGen = currentTask.task_gen;
   const tmuxLines = captureTmuxLines(TMUX_TAIL_LINES);
-  console.error(`Task ${taskId} failed${permanent ? ' permanently' : ''}: ${reason}`);
-  send({ type: 'task_failed', seq: nextSeq(), task_id: taskId, task_gen: taskGen, reason, permanent, tmux_output: tmuxLines });
+  console.error(`Task ${taskId} failed${permanent ? ' permanently' : ''}${kind ? ` [${kind}]` : ''}: ${reason}`);
+  send({ type: 'task_failed', seq: nextSeq(), task_id: taskId, task_gen: taskGen, reason, permanent, failure_kind: kind, tmux_output: tmuxLines });
   currentTask = null;
   taskAssignedAt = 0;
   if (progressInterval) { clearInterval(progressInterval); progressInterval = null; }
@@ -1268,7 +1283,8 @@ function progressTick() {
       } catch (e) {
         console.error('Failed to restart CLI:', e.message);
       }
-      failCurrentTask('CLI process exited — restarted');
+      // environment: the agent CLI process died; nothing was judged about the work.
+      failCurrentTask('CLI process exited — restarted', { kind: 'environment' });
       return;
     }
   } catch (_) {}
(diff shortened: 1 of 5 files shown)


clubanderson:
Reviewed the trust boundary and the DECLARE/ROUTE line, both held cleanly. The kind is normalized onto a closed set (unknown client strings collapse to `unspecified`, never echoed), the free-text reason is the one client-controlled string on the snapshot and it is redacted via redactTokens plus rune-safe-truncated at the exact boundary where it becomes operator-visible, and the copy semantics keep the live connection state out of the fleet snapshot. The separation that matters — cooldown and selection computed from repo/number/permanent alone, never from the declared kind — is pinned by TestSelectionPathsDoNotReadFailureKind and TestRecordTaskFailure_IgnoresFailureKind, which is the right defense against the "tag every failure environment to stay hot" incentive the docs call out. The relay side is appropriately conservative, declaring `environment` only for the three unambiguous cases and omitting otherwise; an honest `unspecified` beats a guess when operators attribute failures from this. Backward compatible in both directions (absent field from old relays, ignored field on old hubs). Merging.

PR #3857, opened 2026-08-14
fix: share contributor claim admission

joshyorko:
## Summary

- reproduce the contributor queue drift where `selectTask` suppresses an issue claimed by an open PR while `ReadyQueue` still advertises it
- extract a deterministic contributor-neutral admission decision shared by both live projections
- preserve contributor-specific cooldown, role, trust, rate, ranking, and routing behavior downstream

## Behavioral change

Before, an open-PR claim prevented live assignment but the same issue remained visible as ready/offerable. After this change, both `ReadyQueue` and `selectTask` consume the same claim-admission decision, including canonical and bare repository-key lookup, while unrelated work remains offerable.

## Scope

This PR establishes and proves the shared admission contract only. It does **not** implement `DependsOn`, desired-state persistence, proofs, CRDs, or scheduler behavior.

## Verification

- `go test ./pkg/dashboard -run '^TestClaimedIssue_ReadyQueueMatchesSelectTaskAdmission$' -count=1`
- `go test ./pkg/dashboard -run '^TestClaimedIssue_' -count=1`
- `go test ./pkg/dashboard -count=1`

Progresses #3845

Issue #3845, opened 2026-08-14
feat: convergence-driven repo reconciler — feed only READY (dependency-aware) work into Hive


joshyorko:
# Summary

Add a deterministic convergence/admission judgment between Hive's observation of repository work and its existing scheduling and contributor-routing machinery.

> **Supporting research — non-normative point-in-time snapshots**
>
> - [Research & Design Dossier (PDF)](https://github.com/user-attachments/files/31074495/3828.Convergence-Driven.Repository.Reconciliation.Research.Design.Dossier.pdf)
> - [Implementation-Readiness Gap Closure (PDF)](https://github.com/user-attachments/files/31074496/3828.Implementation-Readiness.Gap.Closure.pdf)
>
> The issue body is the canonical design contract.

This issue supersedes closed #3828 at the maintainer's request and preserves that issue as the historical discussion record.

Hive v4 already plans, decomposes, reviews, replans, evaluates, prioritizes, routes, and executes work. This feature adds the missing desired-vs-observed judgment: which declared outcomes are currently satisfied, what evidence supports that judgment, what drift or uncertainty remains, and which transitions are safe and authorized now.

```text
desired intent
    ↓
current observations
    ↓
proof / condition evaluation
    ↓
known drift / uncertainty
    ↓
admissible transitions
    ↓
existing Hive scheduling / contributor routing
```

The motivating objective is eventual consistency of maintained software: security, architecture, tests, releases, documentation, outreach, and other declared quality outcomes should remain satisfied as repositories and their environment change. Issues, PRs, commits, and merges are transitions toward those outcomes, not the destination.

This proposal emerged from community design discussion and was filed at the maintainer's request.

# Why this is different from planning

- **Planning** decides what work ought to exist or be revised.
- **Convergence** evaluates whether declared outcomes are currently satisfied, whether the evidence is valid for the current intent and subject, what drift remains, and which transitions are admissible.
- **Scheduling and Hive routing** choose which admitted transition runs and which contributor executes it.

The convergence layer must not become a second planner or scheduler. It gates candidate work; Hive continues to own decomposition/replanning, governor policy, prioritization, role/lane/capability routing, contributor lifecycle, and autonomy controls.

# Current v4 state and corrected gap

[...]

PR #3904, opened 2026-08-15
feat(convergence): dependency-aware admission for ReadyQueue and selectTask

Danathar:
## Summary

Second slice of the convergence-driven repository reconciler (#3845). [#3857](https://github.com/kubestellar/hive/pull/3857) extracted the shared contributor-neutral admission contract (`evaluateContributorNeutralAdmission`) and proved it on open-PR claims. This is follow-on increment 1 — **live dependency/condition admission** — teaching that same contract about declared dependencies so only READY work reaches the live contributor path.

```text
current GitHub/Hive actionable-work observation
    ↓
normalized candidate identity (owner/repo#N)
    ↓
bead-ledger observer  →  pkg/convergence.Evaluate
    ↓
shared contributor-neutral AdmissionDecision
    ├─→ ReadyQueue offerability
    └─→ selectTask contributor policy, ranking, routing
```

## What changed

**`pkg/convergence` (new)** — the runtime-independent core. A pure `Evaluate(Observation) Decision` with the small Kubernetes-style tri-state condition vocabulary the contract asks for (`True` / `False` / `Unknown` plus reasons), not a phase enum. No GitHub types, no bead types, no HTTP, no clock, no CRDs. The decision conveys readiness, the observed record and generation, blockers, and conditions — "conceptual shape, not a frozen Go API".

**`pkg/dashboard/contribute_admission_deps.go` (new)** — the first observer. It resolves a live GitHub candidate to its bead record and reports each `DependsOn` edge as satisfied / unsatisfied / unresolvable. It reads only; it never mints a shadow bead to have something to query.

**`evaluateContributorNeutralAdmission`** now applies that decision, so `ReadyQueue` and `selectTask` gate on the *same* judgment and cannot drift apart the way the open-PR claim did.

## Design questions this answers

**Q3 — candidate ↔ dependency identity.** A candidate resolves through the canonical `gh-owner/repo#N` external ref first, then the bare config-form `gh-repo#N`, then `issue_repo`/`issue_number` metadata (which is what carries identity when `planning.IssueRef` fell back to the issue URL). Keys are case-normalised. Same canonical-then-bare precedence as `issueClaimedByOpenPR`, so a bare-configured hive behaves identically on both admission inputs — the #2648 key-mismatch class.

**Q3 — policy for candidates with no dependency record.** Stated explicitly rather than left accidental: **a lookup miss means "declared no dependencies" and admits.** It is emphatically *not* read as "an unknown dependency was satisfied" — `Observed` is recorded `False` with reason `NoIntentRecord`, so the decision says plainly that the judgment rests on an absent record. Failing closed here would block the entire contributor queue on day one, since almost every actionable GitHub issue has no bead; that is the repository-wide stall the contract forbids.

**Q4 — freshness / bounded authoritative refresh.** The bead ledger is durable local state, so a restart reconstructs the same decisions from the same files. Beads written by agent processes reach the hub through the refresh the governor *already* performs (the per-eval-cycle `Store.Reload()` in `cmd/hive/main.go`), so a dependency closed by an agent gates admission within one eval cycle. This observer deliberately adds no disk read of its own to the assignment path.

**Q5 — reserved metadata.** The decision carries `ObservedRecord` and `ObservedGeneration` (the record's last-update revision) today. Both are reported, never compared — the seam for generation-aware status and exact-subject proofs, without a migration dead end.

## Non-monotonic and unknown handling

- A dependency that becomes unsatisfied again **removes** the dependent from the ready set. Admission is recomputed from current state on every sweep, never latched.
- An unresolvable dependency edge withholds the candidate as `Ready=Unknown`, not `False` — no false satisfaction, but distinguishable from an established blocker an operator can act on.
- A **partial** ledger read (a configured store unreadable) withholds *unfound* candidates rather than assuming they declared nothing; candidates whose record *was* found and resolved are unaffected.
- Blocking is per candidate. `Evaluate` has no cross-candidate state through which one unknown could serialise the queue.

## First-slice acceptance matrix

Every row is asserted against **both** `ReadyQueue` and live `selectTask`, per the contract's insistence that an internal helper, a CLI, or `Store.Ready()` is not sufficient evidence.

| Scenario | Test |
|---|---|
| A depends on unsatisfied B → not offerable, not assignable | `TestDependencyAdmission_UnsatisfiedBlocksBothPaths` |
| B becomes satisfied → A ready, no restart | `TestDependencyAdmission_SatisfactionFlipsBothWaysWithoutRestart` |
| B becomes unsatisfied again → A leaves ready | same |
| Unrelated C ready while A blocked → C stays selectable | every scenario above asserts #700 stays offered |
| Dependency observation unknown → not dispatched, no global serialization | `TestDependencyAdmission_UnknownDependencyIsNotDispatched` |
| Duplicate / out-of-order events → admission from current state | `TestEvaluate_IsPureAndOrderIndependent` |
| Restart → reconstructed from durable state | `TestDependencyAdmission_SurvivesRestart` |
| Observer unavailable → no false satisfaction | `TestDependencyAdmission_PartialLedgerWithholdsRatherThanAssumes` + `TestEvaluate_DegradedNeverAdmits` |
| Production-path proof (queue offerability *and* real assignment) | `TestDependencyAdmission_QueueAndSelectTaskCannotDrift` |

Plus cross-store dependency resolution, all three identity spellings, and the compatibility control: a hive with no bead stores behaves exactly as it did before this change.

## Scope

Deliberately **not** in this PR:

- **`Store.Ready()` is unchanged.** The contract is explicit that the fix is to change the shared admission contract, not add another one-off gate; teaching `Ready()` about `DependsOn` separately would create a second gate with its own semantics.
- No planner, scheduler, governor, prioritisation, or routing changes. A blocked candidate is simply absent from the admitted set.
- No CRDs, no Operator, no new persistence, no proof normalisation, no resource claims, no authority policy.
- A dependency-blocked item is *dropped* from `ReadyQueue`, not surfaced greyed like an operator HOLD — the matrix asks that it not be "represented as ready/offerable", and a blocked-work badge is a dashboard increment, not an admission one.
- Agent dispatch still evaluates at its own boundary (open question 2); this PR does not move it.

`beadSatisfied` is the narrow `LegacyWorkCompleted(beadID)` projection the contract sanctions for the first slice — explicitly **not** a claim that historical bead closure is the universal dependency ontology. Richer non-monotonic, exact-subject conditions can replace it without the live paths changing, because they consume `convergence.Decision`, not bead status.

## Verification

- `go test ./pkg/convergence -count=1`
- `go test ./pkg/dashboard -run 'TestDependencyAdmission|TestClaimedIssue' -count=1 -race`
- `go test ./pkg/dashboard ./pkg/convergence ./pkg/beads ./pkg/planning -count=1`
- `go build ./... && go vet ./...`

Progresses #3845

### DIFF
--- a/v2/cmd/hive/main.go
+++ b/v2/cmd/hive/main.go
@@ -1730,10 +1730,16 @@ func main() {
 	}
 
 	beadStores = make(map[string]*beads.Store)
+	// Count stores that fail to open. They are dropped from beadStores entirely,
+	// which makes an incomplete ledger indistinguishable from a smaller one — and
+	// the dependency admission gate must not read a lookup miss in a truncated
+	// ledger as "this candidate declared no dependencies".
+	beadStoreLoadFailures := 0
 	for name, agentCfg := range cfg.EnabledAgents() {
 		store, err := beads.NewStore(agentCfg.BeadsDir)
 		if err != nil {
 			logger.Warn("failed to init beads store", "agent", name, "error", err)
+			beadStoreLoadFailures++
 			continue
 		}
 		store.SetHiveID(cfg.HiveID)
@@ -1763,6 +1769,7 @@ func main() {
 			store, err := beads.NewStore(agentBeadsDir)
 			if err != nil {
 				logger.Warn("failed to load orphan beads store", "agent", name, "error", err)
+				beadStoreLoadFailures++
 				continue
 			}
 			store.SetHiveID(cfg.HiveID)
@@ -1776,6 +1783,7 @@ func main() {
 			retroStore, err := beads.NewStore(filepath.Join(beadsRootDir, retro.Actor))
 			if err != nil {
 				logger.Warn("failed to init retro beads store", "error", err)
+				beadStoreLoadFailures++
 			} else {
 				retroStore.SetHiveID(cfg.HiveID)
 				beadStores[retro.Actor] = retroStore
@@ -2191,22 +2199,23 @@ func main() {
 	}
 
 	dashSrv.RegisterAPI(&dashboard.Dependencies{
-		Config:           cfg,
-		AgentMgr:         agentMgr,
-		Governor:         gov,
-		GHClient:         ghClient,
-		GHAppAuth:        appAuth,
-		Tokens:           tokenCollector,
-		Knowledge:        knowledgeAPI,
-		Inception:        inceptionEngine,
-		Nous:             nousState,
-		Scheduler:        sched,
-		MetricsCollector: metricsCollector,
-		BeadSynthesizer:  beadSynth,
-		BeadStores:       beadStores,
-		Logger:           logger,
-		Ctx:              ctx,
-		RefreshFunc:      refreshDashboard,
+		Config:                cfg,
+		AgentMgr:              agentMgr,
+		Governor:              gov,
+		GHClient:              ghClient,
+		GHAppAuth:             appAuth,
+		Tokens:                tokenCollector,
+		Knowledge:             knowledgeAPI,
+		Inception:             inceptionEngine,
+		Nous:                  nousState,
+		Scheduler:             sched,
+		MetricsCollector:      metricsCollector,
+		BeadSynthesizer:       beadSynth,
+		BeadStores:            beadStores,
+		BeadStoreLoadFailures: beadStoreLoadFailures,
+		Logger:                logger,
+		Ctx:                   ctx,
+		RefreshFunc:           refreshDashboard,
 		// #3768: give the contribute queue read access to the duplicate-PR
 		// claim ledger, so an issue any open PR (hive-authored or a human
 		// contributor's) already claims to fix is never offered to another
--- a/v2/pkg/beads/beads.go
+++ b/v2/pkg/beads/beads.go
@@ -1,6 +1,8 @@
 package beads
 
 import (
+	"bufio"
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -111,7 +113,16 @@ type Store struct {
 	dir    string
 	hiveID string
 	beads  map[string]*Bead
-	mu     sync.RWMutex
+	// retired holds the IDs of beads that were removed from the live map after
+	// reaching a TERMINAL state — archived by lifecycle culling, or evicted by
+	// evictOldClosed once the store passes maxBeadCount. Both removal paths only
+	// ever remove closed/done beads, so membership here means "this bead existed
+	// and was satisfied", which is materially different from an ID that was never
+	// seen. Dependency resolution needs that distinction: without it a satisfied
+	// dependency that later gets culled resolves nowhere and permanently withholds
+	// its dependent (kubestellar/hive#3845 review).
+	retired map[string]bool
+	mu      sync.RWMutex
 }
 
 func NewStore(dir string) (*Store, error) {
@@ -136,8 +147,9 @@ func NewStore(dir string) (*Store, error) {
 	_ = os.Chmod(dir, 0o770|os.ModeSetgid)
 
 	s := &Store{
-		dir:   dir,
-		beads: make(map[string]*Bead),
+		dir:     dir,
+		beads:   make(map[string]*Bead),
+		retired: make(map[string]bool),
 	}
 
 	if err := s.load(); err != nil {
@@ -210,8 +222,55 @@ func (s *Store) evictOldClosed() {
 		if len(s.beads) <= maxBeadCount {
 			break
 		}
+		// Record the eviction the same way Archive does before dropping it.
+		// Eviction only ever takes CLOSED/DONE beads, so losing the ID silently
+		// would turn a satisfied dependency into an unresolvable one and
+		// permanently withhold whatever depended on it.
+		//
+		// If the archive cannot be written, KEEP the bead. Evicting anyway would
+		// retire it in memory only, and the next restart would forget — the same
+		// silent withholding, just deferred. Eviction is an opportunistic memory
+		// bound, so skipping one bead until the next Create is the cheap side of
+		// this trade.
+		if !s.appendArchiveEntry(s.beads[id]) {
+			continue
+		}
 		delete(s.beads, id)
+		s.retired[id] = true
 	}
+}
+
+// appendArchiveEntry writes one compact archive record, reporting whether the
+// record actually reached disk. Callers use that to decide whether removing the
+// bead is safe: an in-memory-only retirement is forgotten on restart.
+func (s *Store) appendArchiveEntry(b *Bead) bool {
+	if b == nil {
+		return false
+	}
+	entry := ArchivedBead{
+		ID:          b.ID,
+		Title:       b.Title,
+		Type:        b.Type,
+		Priority:    b.Priority,
+		ExternalRef: b.ExternalRef,
+		ArchivedAt:  time.Now().UTC(),
+	}
+	if b.ClosedAt != nil {
+		entry.ClosedAt = b.ClosedAt.Time
+	}
+	data, err := json.Marshal(entry)
+	if err != nil {
+		return false
+	}
+	f, err := os.OpenFile(filepath.Join(s.dir, archiveFileName), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return false
+	}
+	defer f.Close()
+	if _, err := f.Write(append(data, '\n')); err != nil {
+		return false
+	}
+	return true
 }
 
 func (s *Store) Update(id string, fn func(b *Bead)) error {
@@ -482,6 +541,17 @@ func (s *Store) Reload() error {
 }
 
 func (s *Store) load() error {
+	// Retirement is reconstructed FIRST, and unconditionally. archive.jsonl is
+	// an independent append-only log, so a directory can hold a complete archive
+	// with no beads.json at all — a store whose beads were every one of them
+	// culled, a restore that kept only the log, external tooling rewriting the
+	// live file. Loading it after the missing-file return meant exactly those
+	// stores came back with an EMPTY retired set, so every archived dependency
+	// read as "never resolved" rather than "satisfied then culled" — silently
+	// withholding the dependents forever, which is the failure retirement exists
+	// to prevent.
+	s.loadRetired()
+
 	path := filepath.Join(s.dir, beadsFileName)
 	data, err := os.ReadFile(path)
 	if os.IsNotExist(err) {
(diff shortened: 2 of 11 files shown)


Danathar:
Pushed `e56fcf5` — a self-review pass against current `v4`, not a response to feedback. The branch was already at `v4` tip, so this is additive on top of the original commit.

**One real defect found and fixed: a data race I introduced into the assignment hot path.**

The dependency gate reads every bead in every store on every `ReadyQueue` and `selectTask` pass, and it did that through `beads.Store.List` — which returns the store's **live** `*Bead` pointers and releases the read lock before the caller touches them. Every field read off that result races a concurrent `Update`, which mutates beads in place under the write lock (`Status` and `DependsOn` through the caller's `fn`, plus `UpdatedAt`/`ClosedAt`).

Not theoretical, and not test-only: the inception watcher calls `Close` on its own goroutine (`inception_watcher.go:263`) and the planning decomposer calls `AddDependency` (`decompose.go:243`), both while the hub serves assignments. `go test -race` reproduces it against the real production path in a few hundred iterations:

```
WARNING: DATA RACE
Write at ... by goroutine 25:
  beads.(*Store).Update()            beads.go:233   // b.UpdatedAt
Previous read at ... by goroutine 24:
  dashboard.newBeadRecord()
  dashboard.(*ContributeWSHub).buildBeadDependencyIndex()
```

The pattern is pre-existing — `List` has always handed out live pointers and other callers read them the same way — but this PR **widens** it from an occasional read into the task-assignment path, so it belongs fixed here rather than inherited.

Worth naming precisely, because my own comment was wrong: the index already copied values out of each bead and claimed that was for race safety. It wasn't. Copying *outside* the lock narrows the window; it doesn't remove it. The copy has to happen inside.

`pkg/beads` gains `Store.ReadEach(filter, fn)` — same filtering and creation-ordered sequence as `List`, but `fn` runs while the read lock is held. Additive: `List` is untouched, so no existing caller changes. The two rules `fn` must obey (don't retain the pointer, don't call back into the store — the `RWMutex` isn't reentrant) are documented at the method.

`TestBeadDependencyIndex_IsRaceFreeAgainstConcurrentWriters` pins it: two readers (the raw sweep and the full `ReadyQueue` path) against a writer flipping terminal state and growing `DependsOn`. It also asserts the ledger is still **coherent** afterwards — a race-free sweep must be a correct one, not merely a quiet one. Replay: reverted the sweep to `List` → `DATA RACE` under `-race`; restored → green.

`pkg/beads` and `pkg/convergence` pass under `-race`; `pkg/dashboard` passes.

While I was in here I also re-checked the first-slice acceptance matrix row by row against the tests. All nine hold, and the "unrelated C stays selectable" row is exercised by every production-path test rather than one — issue #700 is asserted present in the queue in each scenario where #601 is withheld.

One thing I did **not** change, flagged rather than assumed: `partial` (the degraded path) is only set when a configured store is `nil` in the map. `Store.List`/`ReadEach` can't fail, so there's no I/O-failure path that reaches it today — the degraded semantics are correct and tested, but the only way to *enter* them in production is a nil store entry. If you'd rather the observer treat a store whose `Reload` last failed as partial too, that's a real strengthening of the "observer unavailable" row, but it needs a health signal `pkg/beads` doesn't currently expose. Happy to do it here or as a follow-up if you want it.

Danathar:
Self-review pass (re-reviewed with **Claude Opus 5, max reasoning**). One real gap, plus a note on what I checked and cleared.

## The `partial` ledger guard is unreachable, and the failure it was written for fails **open**

`buildBeadDependencyIndex` sets `partial` in exactly one place — when a configured store is `nil`:

```go
store := h.server.deps.BeadStores[name]
if store == nil {
    idx.partial = true
```

But `BeadStores` never contains a nil value. Every construction path in `v2/cmd/hive/main.go:1732-1785` `continue`s on failure without inserting:

```go
store, err := beads.NewStore(agentCfg.BeadsDir)
if err != nil {
    logger.Warn("failed to init beads store", "agent", name, "error", err)
    continue          // ← the key is never added at all
}
```

Same shape for the orphan-directory scan and the retro store. So `partial` can only be set by a test that hand-builds the map, `beadLedgerPartialReason` never reaches an operator, and **Rule 1 of `convergence.Evaluate` ("a degraded read authorises nothing") never fires in production.**

That would be merely dead code, except for where the real failure lands instead. `beads.NewStore` returns an error when `s.load()` cannot parse `beads.json` (`v2/pkg/beads/beads.go:143`), and on that error the store is silently dropped from the map. The index is then *smaller* with `partial == false`, so a candidate whose bead lived in the dropped store misses the `byRef` lookup, is observed as `Found: false` with no `Degraded` flag, and hits Rule 2:

> Rule 2: no record for this candidate means no declared dependency → `Admitted = true`

A dependent whose blocker lived in the unreadable store is admitted as if it declared nothing. That is the exact case the field comment says must not happen:

> A lookup miss is then not trustworthy — the record may live in the store we could not read — and admission must not treat it as "no declared dependencies".

Not a regression — before this PR there was no dependency gate at all, so nothing that works today breaks. But the guard the PR describes does not hold in the one scenario it names.

**Direction for the fix:** `partial` has to come from the *producer*, not be inferred by the consumer from a nil entry that cannot exist. `main.go` already knows how many configured stores failed to open; carrying that count (or a flag) on `Deps` and reading it in `buildBeadDependencyIndex` makes the degraded path both reachable and honest. Worth a test that a store which fails to open produces `Degraded`, not silent admission.

## Checked and cleared

- **Queue starvation from a degraded read.** My first concern was that `DependencyUnknown` withholds a candidate, so a single bad store could withhold every bead-less candidate and starve the contributor queue — `no_matching_work` forever, presenting as a hub bug. It cannot happen, for the same reason the guard does not fire: `partial` is unreachable. Recording it because if the fix above lands, this becomes live, and the blast radius on a hive where most issues have no bead is the whole queue. The degraded path should probably withhold only candidates that were *found* with unresolved edges, not every miss.
- **Sweep sharing between the two call sites.** `ReadyQueue` and `selectTask` each build their own sweep and pass it down, so the two projections judge against one snapshot per pass and cannot drift mid-pass. Correct.
- **Race safety.** `ReadEach` + copied `beadRecord` values rather than `List`s live pointers is the right call, and the reasoning in the comment matches what `-race` would catch.
- **Store-order determinism.** Sorting store names before first-wins resolution does remove the map-order flap.

Danathar:
Re-reviewed with **Claude Opus 5 (max reasoning)** and pushed fixes for everything found — `29ec408b`.

Five issues, four correctness and one cost. Two of them I found in my own read (the `partial` guard), three came out of a deeper pass.

## 1. HIGH — a *satisfied* dependency that got culled withheld its dependent forever

The one that matters. Closed beads are **deleted** from the live ledger by two live mechanisms — `knowledge.BeadLifecycleManager.RunCulling` → `Store.Archive`, and `Store.evictOldClosed` past `maxBeadCount` — and both only ever take terminal beads. Resolving a missing edge to `Unknown` therefore punished exactly the dependencies that had been **satisfied**: the dependent dropped out of ReadyQueue and selectTask with no log and no Held badge, and never returned. Reproduced before fixing.

`beads.Store` now records IDs removed *after* reaching a terminal state, persisted through `archive.jsonl` (eviction archives too, instead of dropping the ID silently) and exposed as `IsRetired`. A retired edge resolves as satisfied. An edge that resolves nowhere and was never retired is still `Unknown` — a dangling reference is a record we cannot read, not a dependency to wave through, and that distinction is pinned by its own test.

## 2. MEDIUM — the partial-ledger guard could never fire

`idx.partial` was set only when `BeadStores` held a `nil` value, and `cmd/hive/main.go` never stores one: every failure path logs and `continue`s without inserting. A store that genuinely failed to open was simply **absent**, so the ledger looked *smaller* rather than *incomplete* — and any dependent whose blocker lived there read as "declared nothing" and was admitted. Exactly the false satisfaction the design forbids. Only the producer knows a store is missing, so `main.go` now reports it via `Dependencies.BeadStoreLoadFailures`.

## 3. MEDIUM — making it reachable exposed a fleet-wide stall

The old policy withheld every candidate whose record was not found. On a real hive most actionable issues have no bead, so one unreadable `beads.json` would have emptied the entire contributor queue — endless `no_matching_work`, presenting as a hub bug. Fixing 2 without this would have shipped that.

A truncated view now **gates what it can see and admits what it cannot**, and `newAdmissionSweep` logs the reduced coverage every sweep so it is visible rather than silent. This is a deliberate reversal of the original rule, and I want it called out for review: the no-false-satisfaction instinct says withhold, but at this ratio the cost of withholding is the whole queue.

## 4. MEDIUM — PR-numbered refs claimed issue identities

`beadIdentityKeys` stripped an *optional* `gh-` prefix, so any bare `x#n` ref was indexed as an issue identity. `pkg/retro` mints advisory beads with `ExternalRef = "owner/repo#<PR number>"` into a store this index reads, so PR #601 and issue #601 shared a key — and because `byRef` is first-wins across sorted store names, a retro advisory could **shadow** the genuine epic and silently bypass the dependencies that epic declared. The prefix is now required; only `planning.IssueRef` writes it.

## 5. LOW — sweep cost scaled with the ledger, not the candidate set

And it did so in the callback that runs *with each store's RLock held*, blocking writers. `byID` is now `id→satisfied` rather than a record copy; a `beadRecord` is materialized only for beads that answer to a candidate key; the generation timestamp is formatted at lookup instead of once per bead; retirement resolves on demand; `byID` is pre-sized.

Benchmarked at 8 stores x 5000 beads with 150 candidates per sweep:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| before | 12,255,965 | 15,842,131 | 57,867 |
| after | 5,018,390 | 4,767,832 | 17,739 |

~2.4x faster, 3.3x less allocated. `BenchmarkAdmissionSweep` is committed as a cost guard.

The finding's suggested form — a lazily-built index keyed only on the candidate set — is not reachable: `newAdmissionSweep()` takes no arguments and both call sites build the sweep before their candidate loop, and deriving the set from `status` inside the index would fail **open** for any caller with a candidate outside it. So the pass was made cheap rather than conditional.

## Worth saying: the fix introduced four bugs of its own

All caught by tests before landing, all fixed in the same commit — but they are the kind that would have been invisible in review:

- `load()` reached `loadRetired` only on the success path, so an archive-only directory (no `beads.json`) came back with an **empty** retired set — the exact failure retirement exists to prevent, reintroduced by the fix for it.
- `Archive` retired beads that were never terminal, asserting a completion that never happened.
- Eviction retired in memory when the archive write failed; a restart forgets, so the withholding just comes back later.
- One oversized `archive.jsonl` line stopped `bufio.Scanner` permanently, discarding every retirement recorded after it. Bead titles are unbounded and agent-influenced.

## Tests

`v2/pkg/beads/retirement_test.go` (new, 18 tests): both removal paths, restart persistence including the archive-only directory, corrupt and oversized archive lines, and a race guard. Plus 7 added to the dependency-admission suite — culled-satisfied, cross-store culled, dangling-still-unknown, PR-ref identity, partial-ledger policy, nil-entry handling, index shape.

```console
$ go test ./pkg/dashboard/ ./pkg/beads/ ./pkg/convergence/ ./pkg/knowledge/ -short -race -count=1
ok  	github.com/kubestellar/hive/v2/pkg/dashboard	51.700s
ok  	github.com/kubestellar/hive/v2/pkg/beads	1.590s
ok  	github.com/kubestellar/hive/v2/pkg/convergence	1.010s
ok  	github.com/kubestellar/hive/v2/pkg/knowledge	1.564s
```

Two pre-existing tests changed, both deliberately: `TestDependencyAdmission_PartialLedgerWithholdsRatherThanAssumes` pinned the behaviour reversed in item 3, and is replaced by `..._PartialLedgerStillGatesWhatItCanSee`.

clubanderson:
This is a substantial one and I read all three commits closely. The admission design is sound where it most needs to be: evaluation is pure and strictly one-hop (a dependency is judged only by "is that bead terminal"), with no cross-candidate state and no graph traversal — so admission cannot deadlock or livelock, and a dependency cycle among open beads withholds only the cycle's own members from the contributor queue, each with a logged decision naming its blockers, while every unrelated candidate is judged independently. Since hive-side agents are not gated by this path, cycle members still get completed and unblock on a later sweep; a cycle-detection lint at `AddDependency` time would still be a worthwhile follow-up so an accidental A↔B is flagged at creation rather than discovered as two quietly-withheld issues. The two follow-up commits fix real bugs the feature would have shipped with: `ReadEach` closes a genuine data race (List returns live pointers read outside the lock, racing in-place Update), and the retirement set — reconstructed from archive.jsonl before the beads.json existence check, with bufio.Reader instead of a Scanner that dies permanently on one oversized line — stops a satisfied-then-culled dependency from withholding its dependent forever. The lookup-miss policy (miss = no declared dependencies, even on a partial ledger, logged) is the right call at fleet scale, and the fail-closed treatment of dangling edges mirrors Store.Ready(). Tests are extensive and assert invariants, not vibes. Merging.

PR #3910, opened 2026-08-15
feat(backends): support agy (Antigravity CLI) — Gemini CLI is retired for subscription accounts

hanthor:
## Problem

`agy` is listed in `config/backends.conf`'s `KNOWN_BACKENDS` but missing from `config.CLIBackends`, so the hub rejects it:

```
unsupported backend "agy" (supported: claude, copilot, goose, codex, pi, bob, aider, gemini; ...)
```

That is the same accept-in-one-place drift `CLIBackends` exists to prevent, just inverted — valid in the shell config, unknown to the hub.

It is not cosmetic, because **the Google backend hive does accept has been retired.** Google consolidated its CLI tooling under Antigravity at I/O 2026, and [Gemini CLI stopped serving personal and Google AI Pro accounts on 2026-06-18](https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/); only paid Gemini Code Assist licence holders can still invoke it. Antigravity CLI — invoked as `agy` — is the replacement.

So for any hive whose Google access is a Gemini/AI Pro **subscription** rather than metered API credits, the situation is:

- `gemini` — accepted by the hub, but retired and needs API credits a subscription does not provide
- `agy` — has the working credentials, but is refused

The backend that is paid for is rejected; the one that is accepted cannot pay. Google is effectively unavailable as a backend.

## Change

- Add `agy` to `config.CLIBackends`. Validation and binary resolution both derive from that list, so this is all that is needed for the hub to accept it and for `backendBinaryName` to resolve it.
- Add the launcher case. `agy` needs `--dangerously-skip-permissions` or it blocks on a per-tool approval prompt with nobody attached — the same contract as claude's bypass flag, and the value already used for agy in `backends.conf`.
- Pass `--effort` alongside `--model`. agy **requires** it and otherwise ignores the model entirely:

  ```
  --model gemini-3.6-flash requires --effort (available: low, medium, high).
  Using "Gemini 3.6 Flash (Low)" instead.
  ```

  `low` is the effort agy falls back to on its own, so behaviour is unchanged while the configured model actually takes effect. Hive has no per-agent reasoning-effort setting yet; `agyDefaultEffort` is the constant to replace when it grows one.

`gemini` is deliberately **kept** — Gemini Code Assist licence holders can still use it.

## Verification

Built and deployed to a live single-node hive (v4, 9 agents), then switched an agent onto agy through the normal API. Hive launched it itself:

```
/usr/local/bin/agy --dangerously-skip-permissions --model gemini-3.6-flash --effort low
```

```
      ▄▀▀▄        Antigravity CLI 1.1.11
     ▀▀▀▀▀▀       <account> (Google AI Pro)
    ▀▀▀▀▀▀▀▀      Gemini 3.6 Flash (Low)
   ▄▀▀    ▀▀▄     /data/agents/guide
? for shortcuts
```

Authenticated against a Google AI Pro subscription and reaching the ready signature, with no effort warning once `--effort` was passed.

Note agy has no known one-shot/headless entry point (it drives an interactive TUI), consistent with how it is already described in `api_contribute_test.go`.

## Tests

`TestBackendBinaryName_EverySupportedBackendResolves` already covers agy automatically, since it derives its cases from `config.SupportedBackends()` — which is exactly the property that guard was added for. Added `TestBackendBinaryName_AgyLaunchesAntigravityCLI` to pin agy to its own binary, mirroring the existing codex/aider test, so a future tidy-up cannot fold it into the gemini derivation.

`go build ./...`, `go vet`, and the `pkg/config` + `pkg/agent` backend tests all pass.

### DIFF
--- a/v2/pkg/agent/backend_coverage_test.go
+++ b/v2/pkg/agent/backend_coverage_test.go
@@ -303,6 +303,28 @@ func TestBackendBinaryName_EverySupportedBackendResolves(t *testing.T) {
 	}
 }
 
+// TestBackendBinaryName_AgyLaunchesAntigravityCLI pins agy to its own binary.
+//
+// agy is the Antigravity CLI, Google's replacement for the Gemini CLI: Google
+// stopped serving Gemini CLI for personal and AI Pro accounts on 2026-06-18, so
+// for any hive whose Google access is a subscription rather than metered API
+// credits, agy is the ONLY reachable Google backend. It is emphatically not an
+// alias for gemini — different binary, different auth, different flags — so a
+// future tidy-up that folds it into the gemini derivation would silently break
+// every subscription-backed Google agent. This catches that.
+func TestBackendBinaryName_AgyLaunchesAntigravityCLI(t *testing.T) {
+	if !config.IsCLIBackend("agy") {
+		t.Fatal("precondition failed: \"agy\" is not a config.CLIBackends member")
+	}
+	got, err := backendBinaryName("agy")
+	if err != nil {
+		t.Fatalf("backendBinaryName(\"agy\") err = %v, want nil", err)
+	}
+	if got != "agy" {
+		t.Errorf("backendBinaryName(\"agy\") = %q, want \"agy\"", got)
+	}
+}
+
 // TestBackendBinaryName_CodexAndAiderLaunchTheirOwnBinaries pins the specific
 // regression: codex and aider are agentic CLIs of their own, not aliases and not
 // gateway backends, so each must exec a binary of its own name. If either were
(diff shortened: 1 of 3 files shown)


clubanderson (approved):
Since this adds a new backend, I focused the review on the launch and exec path. It checks out: agy is launched through the same launchInTmux switch as every other CLI backend, so it inherits the per-agent environment unchanged — HTTPS_PROXY/HTTP_PROXY through the local MITM proxy, NODE_EXTRA_CA_CERTS, HIVE_PROXY_AGENT attribution, and the gh-wrapper PATH machinery — nothing about this backend bypasses the proxy or credential handling. Binary resolution derives from config.CLIBackends exactly as the accept-in-one-place guard intends, and the flags match what config/backends.conf already declared for agy contributor-side, closing the inverted drift the PR describes. --dangerously-skip-permissions is the same contract claude already runs under. agyDefaultEffort as a named constant with the replace-me-when-per-agent-effort-exists note follows repo convention, and the pinning test (agy is not a gemini alias) plus the derived coverage from SupportedBackends is the right test shape.

One note, not blocking: the image does not install an agy binary — same situation as gemini — so this remains a bring-your-own-binary backend until a Dockerfile layer adds it, which matches the gemini precedent and your live verification. Keeping gemini for Code Assist licence holders is the right call. Merging on green.

PR #3919, opened 2026-08-15
fix(agents): stop the launcher answering codex's update prompt with "Update now"

hanthor:
## The bug

codex renders an update menu at startup with the **destructive option pre-selected**:

```
✨ Update available! 0.146.0 -> 0.147.0
› 1. Update now (runs `npm install -g @openai/codex`)
  2. Skip
  3. Skip until next version
  Press enter to continue
```

The launcher confirms it. `npm install -g` then runs as the unprivileged agent UID, fails with `EACCES`, and takes the CLI down with it — **on every launch, indefinitely**. The agent never runs a single task.

## Four defects, one fix each

1. **Repeat Enter answers prompts.** After typing the launch line, `tmuxSendEntersForAgent` sends `enterCount` (3) Enters as insurance the line ran. Enter #1 runs it, codex boots and draws the menu, Enters #2/#3 confirm the pre-selected option — within milliseconds, so no poll-based watcher can win that race. The repeats now stop once a known blocking prompt is on screen (the first Enter is always sent).

2. **The prompt watcher never ran for codex** — gated `backend == "copilot"`. The copilot folder-trust watcher is generalised into a table of positively-identified prompts, mirroring `blockingPromptKey()` in `bin/contributor-relay.sh` which solved this contributor-side. The gate derives from the table, so a prompt added for a new backend enables the watcher automatically.

   | backend | prompt | answer |
   |---|---|---|
   | copilot | `Confirm folder trust` | `2` (yes, and remember) |
   | codex | `Do you trust the contents of this directory` | `1` |
   | codex | `Update available!` + `Skip until next version` | `3` — persists; a plain Skip re-prompts next launch |
   | agy | `Do you trust the contents of this project` | Enter alone — its affirmative option is pre-selected; an empty key means exactly that |

3. **Readiness fired on the menu.** `paneShowsConsentScreen` needs an `Enter to confirm` footer and a `❯` line; codex renders `Press enter to continue` and `›` (U+203A), so its menu read as READY and everything gated on readiness (startup kick, caveman activation) typed into it. A known blocking prompt now reads as not-ready.

4. **Scrollback matched as live.** `captureTmuxPaneForAgent` returns history, so a dead CLI's menu stayed in the capture and the watcher typed `3` into the bash shell that replaced it, once per poll (observed live). Prompts now match only the pane **tail**, are scoped to the backend that renders them, and are answered at most once per watcher — the pane keeps rendering for a beat after the keystroke, and a second match used to submit the option as a chat message (codex replying "Three.").

## Deliberately not done: updating codex in place

`npm install -g` needs write access to `/usr/local/lib/node_modules` (`root:root`), which **neither the agent UID nor the hub process has** — driving the install from the hub fails with the same exit status 243 (tried, removed). The skip persists in each agent's `CODEX_HOME`, so the prompt is answered once per agent, ever; updating the CLI belongs to the image build.

## Verification

Live hive, codex 0.146.0 with 0.147.0 available:

- **before:** every launch → `Error: npm install -g @openai/codex failed with status exit status: 243`, agent never usable
- **after, first launch:** prompt answered `3` once, codex reaches its input prompt and runs tasks
- **after, subsequent launches:** no prompt at all, straight to ready

## Tests

`blocking_prompt_test.go`: the update prompt resolving to `3` and explicitly **not** `1`; each known prompt including agy's empty-key contract; backend scoping (a prompt is never answered while running another backend); unrelated panes left alone; and the scrollback case using a real dead-CLI pane. Plus a guard that the watcher gate matches the table — the defect that made the mechanism inert.

(`TestConfirmMenuOption_SelectsOption` fails on a clean `v4` checkout here as well — pre-existing, unrelated.)

*History note: this PR was developed live against the affected hive and previously carried four commits including mid-investigation corrections; squashed to one reviewable commit. The comment thread preserves the investigation trail.*

### DIFF
--- a/v2/pkg/agent/blocking_prompt_test.go
+++ b/v2/pkg/agent/blocking_prompt_test.go
@@ -1,0 +1,205 @@
+package agent
+
+import "testing"
+
+// codexUpdatePane is the real codex 0.146.0 update menu, verbatim. Note the
+// PRE-SELECTED option ("›" marker) is "1. Update now" — the destructive one.
+const codexUpdatePane = `  ✨ Update available! 0.146.0 -> 0.147.0
+  Release notes: https://github.com/openai/codex/releases/latest
+› 1. Update now (runs ` + "`npm install -g @openai/codex`" + `)
+  2. Skip
+  3. Skip until next version
+  Press enter to continue`
+
+const codexTrustPane = `> You are in /data/agents/architect
+  Do you trust the contents of this directory? Working with untrusted contents
+  comes with higher risk of prompt injection.
+› 1. Yes, continue
+  2. No, quit
+  Press enter to continue`
+
+const copilotTrustPane = `Confirm folder trust
+  1. Yes
+  2. Yes, and remember for future sessions
+  3. No`
+
+// TestBlockingPromptKey_CodexUpdateSkipsInsteadOfInstalling is the whole point
+// of this mechanism.
+//
+// codex's update menu pre-selects "1. Update now", which runs
+// `npm install -g @openai/codex` as the unprivileged agent UID. That fails with
+// EACCES and kills the CLI — on EVERY launch, indefinitely. A bare Enter, or any
+// heuristic that merely avoids options containing "no"/"exit", picks exactly
+// that option. Only "3. Skip until next version" both unblocks startup and
+// persists so the prompt does not return on the next launch.
+func TestBlockingPromptKey_CodexUpdateSkipsInsteadOfInstalling(t *testing.T) {
+	key, label, ok := blockingPromptKey("codex", codexUpdatePane)
+	if !ok {
+		t.Fatal("codex update prompt was not recognised; the agent would block on it until timeout")
+	}
+	if key == "1" {
+		t.Fatal("answered \"1. Update now\" — this runs npm install -g as the agent UID and kills the CLI")
+	}
+	if key != "3" {
+		t.Errorf("key = %q, want \"3\" (Skip until next version); %q does not persist across launches", key, key)
+	}
+	if label == "" {
+		t.Error("label is empty; the audit log would not say which prompt was answered")
+	}
+}
+
+func TestBlockingPromptKey_KnownPrompts(t *testing.T) {
+	cases := []struct {
+		name    string
+		backend string
+		pane    string
+		want    string
+	}{
+		{"codex directory trust", "codex", codexTrustPane, "1"},
+		{"copilot folder trust", "copilot", copilotTrustPane, "2"},
+		{"codex update", "codex", codexUpdatePane, "3"},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			key, _, ok := blockingPromptKey(c.backend, c.pane)
+			if !ok {
+				t.Fatalf("prompt not recognised")
+			}
+			if key != c.want {
+				t.Errorf("key = %q, want %q", key, c.want)
+			}
+		})
+	}
+}
+
+// TestBlockingPromptKey_IgnoresUnrelatedPanes: a prompt is only answered when
+// positively identified. Firing a numbered keystroke at an arbitrary pane would
+// type stray input into a working agent's prompt.
+func TestBlockingPromptKey_IgnoresUnrelatedPanes(t *testing.T) {
+	panes := []string{
+		"",
+		"? for shortcuts",
+		"⏵⏵ bypass permissions on (shift+tab to cycle)",
+		"› Implement {feature}\n  gpt-5.6-sol default · /data/agents/architect",
+		// Superficially similar but NOT the update menu: no options rendered,
+		// so there is nothing to answer.
+		"Update available! 0.146.0 -> 0.147.0",
+	}
+	for _, p := range panes {
+		if key, _, ok := blockingPromptKey("codex", p); ok {
+			t.Errorf("pane %q was answered with %q; want no action", p, key)
+		}
+	}
+}
+
+// TestBackendHasBlockingPrompts_GateMatchesTable is the guard for the bug that
+// made this whole mechanism inert: the watcher was gated on
+// `backend == "copilot"`, so codex agents were never rescued from their update
+// menu no matter how well the menu was understood. Deriving the gate from the
+// table means a prompt added for a new backend enables the watcher for it.
+func TestBackendHasBlockingPrompts_GateMatchesTable(t *testing.T) {
+	for _, p := range blockingPrompts {
+		if !backendHasBlockingPrompts(p.backend) {
+			t.Errorf("backend %q has a blocking prompt but the watcher gate excludes it", p.backend)
+		}
+	}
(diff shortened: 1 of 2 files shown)


hanthor:
Follow-up: I found why the watcher does not answer the menu once the Enter storm is removed, and it is a **second, independent bug** — the watcher is cancelled almost immediately after it starts.

Added temporary lifecycle logging around `watchForTrustPromptForAgent`:

```
09:56:56.804  blocking-prompt watcher started    agent=ci-maintainer backend=codex
09:56:57.015  blocking-prompt watcher cancelled  agent=ci-maintainer
09:57:00.579  blocking-prompt watcher started    agent=ci-maintainer backend=codex
```

**Cancelled 211ms after starting.** The watcher runs on `agentCtx`, which is torn down by the restart churn that happens during startup, so it never survives to its first useful poll — its ticker interval is 2s. The gate and the matching are both correct (`backend=codex` is resolved properly, and a captured live pane contains `Update available!` / `Skip until next version` within 6 non-blank lines, well inside the tail window).

This affects the **existing** Copilot folder-trust watcher too, on the same code path — it is simply less visible there, because Copilot's prompt is not destructive when confirmed and a later retry usually catches it.

So the full picture is three layers:

1. repeat Enter answers the menu — **fixed in this PR, verified**
2. the watcher that should answer it correctly is cancelled before its first poll — **found, not fixed**
3. with both in play the agent sits on the menu until `prompt timeout`

I did not fix (2) here: tying the watcher to a context that outlives agent restarts is a change to lifecycle ownership, and someone who knows the intended semantics of `agentCtx` should decide whether the watcher should survive a restart, be restarted alongside the agent, or be driven from the existing pane poller instead.

One further option worth considering for codex specifically, suggested by the operator running this fleet: rather than only skipping, have the **hub** run `npm install -g @openai/codex` when it sees the prompt. The prompt's own "Update now" runs exactly that command but as the unprivileged agent UID, where it fails; the hub process has the privileges the agent lacks, so it can do the install the agent is being asked to do. Skipping unblocks the current launch but the prompt returns for every agent forever, whereas updating once ends it. I have this implemented locally (serialised with a mutex, fire-and-forget, never fatal, run only after the agent has already been unblocked with "skip until next version") but I have deliberately not included it here, because it cannot be exercised until (2) is resolved and I do not want to ship a path I have not seen run.

Happy to follow up with either piece if there is a preference on how the watcher's lifetime should work.

hanthor:
**Update: this now works end to end.** Retracting the "not yet working" caveat in the description — codex reaches its input prompt and runs tasks.

Pushed two more commits. The missing piece was not the watcher lifetime after all; that was a red herring from my own test harness (I had been `tmux kill-server`-ing between attempts, which left the agent with no session for the watcher to read). Driving it purely through the API, the full chain fires:

```
stopping repeat Enter: a prompt is awaiting an answer  agent=ci-maintainer sent=2
auto-answered blocking prompt  agent=ci-maintainer prompt="codex update prompt" option=3
```

and the pane reaches:

```
• Caveman full. Me ready. You give task. ...
› Write tests for @filename
  gpt-5.6-sol low · /data/agents/ci-maintainer
```

Two things that surfaced only once it started working, both now fixed:

**1. The prompt was answered twice.** The pane keeps rendering the menu for a beat after the keystroke, so the next poll matched it again — and by then the CLI was at its input prompt, so `3` was submitted as a *user message*. codex dutifully replied "Three. 🪨🪨🪨". Each prompt is now answered at most once per watcher.

**2. Updating in place is not viable, so skipping is the right answer.** I tried the obvious improvement — have the hub run `npm install -g @openai/codex` itself, since the prompt's own "Update now" runs exactly that and only fails because the *agent* UID lacks permission. It fails from the hub too, with the same `exit status 243`: `/usr/local/lib/node_modules` is `root:root` and the hub process is not root either. I removed that code rather than ship a path that cannot succeed, and left the finding as a comment so the next person does not retry it.

That turns out not to matter, because **the skip persists**. "Skip until next version" is written to each agent's `CODEX_HOME`, which hive already provisions per agent. On a subsequent restart of the same agent the prompt did not appear at all — codex launched straight to ready, with zero watcher activity. So the prompt is answered once per agent, ever. Updating the CLI belongs to the image build, not to a running agent.

Also note this is the same mechanism as before: the destructive `1. Update now` is never selected, so the `npm install` that was killing the CLI on every launch no longer runs at all.

Full verification, on a live single-node v4 hive with codex 0.146.0 and 0.147.0 available:

- before: `Error: npm install -g @openai/codex failed with status exit status: 243` on every launch, agent never usable
- after, first launch: prompt answered `3` once, codex reaches ready, agent runs tasks
- after, subsequent launches: no prompt at all, codex reaches ready directly

`go build ./...`, `go vet`, and the `pkg/agent` + `pkg/config` prompt/trust/consent/backend tests pass. (`TestConfirmMenuOption_SelectsOption` still fails on a clean `v4` checkout here too — pre-existing, unrelated.)

hanthor:
Pushed one more commit: **agy's project-trust prompt**.

Now that a hive can run `backend: agy` (#3910), agy turns out to block startup the same way codex and copilot do:

```
Do you trust the contents of this project?
Antigravity CLI requires permission to read, edit, and execute files here.
> Yes, I trust this folder
  No, exit
  ↑/↓ Navigate · enter Confirm
```

An agent rotated onto agy sits there until the readiness timeout. Observed live — I had to answer it by hand.

Unlike the codex entries, its affirmative option is **already selected** and there is no numbered option to type, so the answer is a bare Enter. An empty `key` now means exactly that, and the send path skips the keystroke rather than typing a digit that would land as stray input:

```go
if key != "" {
    m.tmuxSendKeysForAgent(agent, key)
    time.Sleep(enterDelay)
}
m.tmuxSendKeysForAgent(agent, "Enter")
```

Tests cover the empty-key contract, the watcher gate including agy, and backend scoping (the agy pane must not be answered while running codex/copilot/claude).

This matters more than a one-off convenience: with rotation moving agents between providers automatically, a backend that hangs on first launch strands whichever agent lands on it.

clubanderson (approved):
Walked all four defects against the diff and the v4 source. The mechanism holds together well: prompts are positively identified and backend-scoped, matching is restricted to the pane tail (paneTail over blockingPromptTailLines) so a dead CLI's scrollback can never be re-answered, the answered map bounds each prompt to one keystroke per watcher, the readiness gate now treats a known blocking prompt as not-ready so nothing gated on readiness types into a menu, and the repeat-Enter loop checks the visible pane (captureVisiblePaneForAgent, not the scrollback capture) before each insurance Enter — so the check cannot itself false-positive on history. Answering the update menu with 3 rather than 1 is clearly right, and the reasoning for why in-place update is not viable (root-owned node_modules, tried and removed) is documented at the constant.

The test suite is the strongest part: it pins the destructive default explicitly (key != 1 with its own failure message), guards the gate-matches-table property that made the original mechanism inert, covers agy's empty-key bare-Enter contract, backend scoping, unrelated panes, and the real dead-CLI scrollback pane. Mirroring blockingPromptKey from contributor-relay.sh keeps hub and contributor sides consistent. Merging on green.

PR #3921, opened 2026-08-15
fix(automerge): size the self-merge sweep to the GitHub rate limit (multi-repo hives never merge)

hanthor:
## Problem

`StartSelfAuthoredAutoMergeSweep` ticks on a fixed `selfAuthoredAutoMergeSweepInterval = 10 * time.Second`, but each tick calls `listOpenAppAuthoredPullRequests` for **every configured repo**. The cost scales with repo count; the interval does not.

For a hive with 45 repos:

```
45 repos x 360 ticks/hour  = 16,200 requests/hour
GitHub App installation limit =  6,900 requests/hour
                              -> 2.3x over
```

…and that is the sweep **alone**, before the agents' own `gh` calls, the governor eval cycle, `writeMergeEligible`, MTTR collection or any fix loop — all of which draw on the same allowance.

## What it looks like in production

On a live 45-repo hive this manifested as auto-merge simply not working, with no obvious cause:

```
WARN self-authored automerge sweep failed
  error="listing open PRs for tuna-os/tunaos: GET .../pulls?per_page=100&state=open:
         403 API rate limit of 6900 still exceeded until 2026-08-15 11:32:13 -0400 EDT,
         not making remote request. [rate reset in 39m28s]"
```

Two things make this worse than a transient 403:

1. Once the limit is blown, go-github **short-circuits client-side** — "not making remote request" — so every subsequent call fails until the recorded reset, including calls from unrelated subsystems. `failed to compute MTTR` and the queued-PR sweep fail the same way.
2. The sweep retries every 10s throughout, so the hive spends the entire window re-confirming it is rate-limited.

Net effect: **no PR merged for a full day** on a hive whose own fleet stats showed `prs_merged: 767` from earlier operation. 33 PRs sat merge-eligible and untouched.

## Fix

Derive the interval from repo count so the sweep can consume at most a bounded share (25%) of the hourly budget. The existing constant becomes the **floor**, so small hives are byte-identical to today:

| repos | interval | requests/hour |
|---|---|---|
| 1–4 | 10s (unchanged) | ≤1,440 |
| 39 | 81s | 1,733 |
| 45 | 94s | 1,723 |
| 100 | 209s | 1,722 |

The sweep is a background nicety — it exists so the App's own green PRs land without a human queuer. Trading tick latency for not starving every other GitHub consumer in the process is the right side of that trade, and a sweep that 403s continuously has a latency of infinity anyway.

## Verification

Deployed to the affected hive. Before:

```
self-authored automerge sweep started  interval=10s
...403 every 10s, 0 merges/24h
```

After:

```
self-authored automerge sweep started  interval=81s  repos=39
```

The `repos` field is added to that log line so this is diagnosable from startup rather than by arithmetic.

Full live confirmation that merges resume is pending the current rate-limit window reset — the pre-fix burn had already exhausted the App's allowance for that window, so the client stays short-circuited until GitHub's own reset regardless of the new interval. The arithmetic and the unit tests below establish the fix; I will confirm merge resumption on the live hive and comment here.

## Tests

`automerge_sweep_interval_test.go`:

- **StaysWithinRateBudget** — across 1…500 repos, requests/hour never exceeds the configured share, and never the full limit.
- **SmallHivesUnchanged** — 0–4 repos still return exactly `selfAuthoredAutoMergeSweepInterval`, so this cannot slow down hives it was already fine for.
- **ScalesWithRepos** — monotonic; more repos can never sweep more often.
- **RealWorldCase** — pins the 45-repo case that broke, and logs the before/after (`16,200 -> 1,723 req/hour`).

`go build ./...`, `go vet ./pkg/github/` and the new tests pass.

## Note

The same scaling argument applies to `SweepQueuedAutoMerges` (it also iterates `c.getRepos()` per tick). I have left it alone here to keep this reviewable and because its cadence is governor-driven rather than a fixed 10s ticker, but it is worth a look.

### DIFF
--- a/v2/pkg/github/automerge_sweep.go
+++ b/v2/pkg/github/automerge_sweep.go
@@ -23,6 +23,42 @@ const DefaultAutoMergeSweepMaxMerges = 3
 // tight loop risks GitHub secondary rate limits across every managed repo.
 const selfAuthoredAutoMergeSweepInterval = 10 * time.Second
 
+// selfAuthoredSweepBudgetShare caps the fraction of the App's hourly REST
+// budget this sweep alone may consume. The sweep is a background nicety; the
+// agents' own gh calls, the governor's eval cycle, the merge-eligible writer
+// and the fix loops all draw on the same allowance and must not be starved.
+const selfAuthoredSweepBudgetShare = 0.25
+
+// githubAppHourlyRateLimit is the GitHub App installation REST allowance the
+// interval is sized against.
+const githubAppHourlyRateLimit = 6900
+
+// selfAuthoredSweepInterval sizes the sweep tick so a hive with many repos
+// cannot exhaust its GitHub rate limit just by looking for merge candidates.
+//
+// Each tick lists open PRs for EVERY configured repo, so cost scales with repo
+// count while the fixed 10s tick did not. A 45-repo hive therefore issued
+// 45 x 360 = 16,200 requests/hour against a 6,900/hour limit — 2.3x over,
+// before a single agent made a call. Observed live: the sweep 403'd
+// continuously, go-github then short-circuited every request until the recorded
+// reset ("not making remote request"), and NO PR merged for a full day on a
+// hive that had merged 767 previously.
+//
+// Small hives are unaffected: the fixed interval remains the floor, so a
+// 1-4 repo hive still sweeps every 10s exactly as before.
+func selfAuthoredSweepInterval(repos int) time.Duration {
+	if repos <= 0 {
+		return selfAuthoredAutoMergeSweepInterval
+	}
+	budget := float64(githubAppHourlyRateLimit) * selfAuthoredSweepBudgetShare
+	seconds := float64(repos) * 3600.0 / budget
+	interval := time.Duration(seconds * float64(time.Second))
+	if interval < selfAuthoredAutoMergeSweepInterval {
+		return selfAuthoredAutoMergeSweepInterval
+	}
+	return interval.Round(time.Second)
+}
+
 const (
 	autoMergeReasonNoHiveQueueApproval        = "no-hive-queue-approval"
 	autoMergeReasonNoAppBotLogin              = "no-app-bot-login"
@@ -316,8 +352,9 @@ func (c *Client) StartSelfAuthoredAutoMergeSweep(ctx context.Context, maxMerges 
 			"acmm_level", level, "min_acmm_level", config.SelfMergeMinACMMLevel)
 		return
 	}
+	interval := selfAuthoredSweepInterval(len(c.getRepos()))
 	go func() {
-		t := time.NewTicker(selfAuthoredAutoMergeSweepInterval)
+		t := time.NewTicker(interval)
 		defer t.Stop()
 		for {
 			select {
@@ -330,7 +367,7 @@ func (c *Client) StartSelfAuthoredAutoMergeSweep(ctx context.Context, maxMerges 
 			}
 		}
 	}()
-	c.info("self-authored automerge sweep started", "interval", selfAuthoredAutoMergeSweepInterval)
+	c.info("self-authored automerge sweep started", "interval", interval, "repos", len(c.getRepos()))
 }
 
 // listOpenAppAuthoredPullRequests returns every open, non-draft PR in owner/repo
--- a/v2/pkg/github/automerge_sweep.go
+++ b/v2/pkg/github/automerge_sweep.go
@@ -33,25 +33,50 @@ const selfAuthoredSweepBudgetShare = 0.25
 // interval is sized against.
 const githubAppHourlyRateLimit = 6900
 
+// selfAuthoredSweepCandidateAllowance is the per-tick request allowance for
+// per-CANDIDATE calls, on top of the one list call per repo. Listing is not the
+// whole cost of a tick: every open App-authored non-draft PR costs a
+// PullRequests.Get in trySweepSelfAuthoredPR, plus a second re-verify Get on
+// the ones that reach the merge step. Sizing the interval on repo count alone
+// therefore understates a tick — on the hive this was measured on, candidate
+// Gets outnumbered list calls whenever the App had a backlog of open PRs.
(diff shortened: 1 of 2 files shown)


hanthor:
Live confirmation, as promised in the description.

The pre-fix burn had exhausted the App's allowance for the window in progress, so the client stayed short-circuited until GitHub's own reset regardless of the new interval. That window reset at `11:32:13 -0400`, and a fresh one began at `11:39:48`. Measurements below are taken **inside a single window** (same `reset` timestamp at both samples), so no reset boundary distorts the delta.

**Failures stopped:**

```
sweep failures, last 2 min: 0
MTTR  failures, last 2 min: 0
```

Both had been continuous before — the sweep every 10s, and `failed to compute MTTR` every 5 min as collateral, since go-github's client-side short-circuit blocks unrelated callers too.

**Consumption is now a fraction of the budget:**

```
t0  11:39:56  remaining=6893  reset=12:39:48
t1  11:45:57  remaining=6765  reset=12:39:48
    -> 128 requests in 6 min = ~1,280 req/hour   (limit 6,900)
```

That ~1,280/hour is the **whole hive**, not just the sweep: the sweep, MTTR collection, the merge-eligible writer, and the nine agents, which share the same App installation token via `gh-app-token.cache`. Against the 16,200/hour the sweep alone was issuing before, the hive now sits at roughly 19% of its allowance.

Startup line for reference:

```
INFO self-authored automerge sweep started  interval=81s  repos=39
```

The strongest evidence is not the recovery itself but that it **held**: the fresh window has run for well over the ~25 minutes it would have taken the old 10s interval to exhaust 6,900 again, with zero failures and the limit barely touched.

One thing this does not yet show is a merge, and that is expected here rather than a gap in the fix: this hive currently has only 2 App-authored open non-draft PRs (58 of 60 open PRs are human-authored), and the self-authored sweep by design only merges the App's *own* green PRs. The machinery is repaired and polling healthily; whether it lands anything depends on the App opening PRs that go green. Happy to report back if one merges, though the rate-limit behaviour above is the part this PR is responsible for.

hanthor:
**Correcting this PR's cost model.** I have since instrumented the client and measured per-route/per-caller spend, and the arithmetic in the description is wrong in two ways. Both matter for judging whether this change is worth taking.

**1. The per-tick cost is higher than I stated.** I sized the interval on "one list request per repo per tick". In fact `trySweepSelfAuthoredPR` also issues `PullRequests.Get` per candidate (and a second re-verify Get before merging), so a tick costs roughly `repos + candidates`, not `repos`. The formula here therefore *understates* the sweep's real cost — the direction of the fix is right, but the numbers in the table should not be read as accurate.

**2. The sweep was not the main consumer.** Measured with all nine agents paused, hive's internals ran **23,280 req/hour against a 6,900 limit**, and the dominant route was not the sweep at all:

```
GET /repos/*/*/pulls/{n}   235/min   <- knowledge.(*PREnricher).fetchPR
total                      388/min
```

That is `fetchPRFromRepos` probing every configured repo for a PR number — 39 calls to answer one lookup — during the hourly bead-synthesizer pass. Filed as #3957. Steady state between passes is only ~1,680 req/hour, comfortably inside the limit.

**What I think that means for this PR.** The 10s fixed interval is still wrong for a multi-repo hive — it scales cost with repo count while holding the tick constant, and it *did* 403 continuously on this hive — so I would still like to see it fixed. But it is a contributing factor, not the cause of the outage I originally attributed to it, and the specific req/hour figures in the description should be treated as unreliable until recomputed against the real per-tick cost.

Happy to either recompute the formula to include per-candidate Gets and update the description, or close this in favour of #3957 if maintainers judge the sweep interval acceptable once the fan-out is fixed. I would rather have it accurate than merged.

hanthor:
Recomputed as promised. The cost model now includes the per-candidate `PullRequests.Get` calls (`selfAuthoredSweepCandidateAllowance`, a documented per-tick allowance), not just the per-repo lists — so the numbers in the code and tests are no longer the ones I retracted above.

Two deliberate choices worth a reviewer's eye:

1. **Small hives (≤4 repos) keep the fixed 10s tick.** Applying the allowance to them would slow a 1-repo hive to ~69s for no benefit — the budget share exists to stop many-repo hives starving every other GitHub consumer, which a small hive cannot do, and their absolute list cost stays well inside the limit. This keeps the change a strict no-op for the quick-start deployment.
2. **The allowance is static, not measured per tick.** A dynamic ticker resized from the last sweep's candidate count would be more precise and more code; a documented upper allowance bounds the damage while staying reviewable. Happy to go dynamic if preferred.

With the corrected model, 45 repos sizes to **2m41s (~1,722 req/hour within the 25% share)** instead of the 94s the list-only formula produced.

Context for sizing this PR's importance honestly: the dominant consumer on the affected hive turned out to be the bead-synthesizer per-repo PR probe (#3957), not this sweep. This fix is still worth having — the fixed tick genuinely scales cost with repo count and did 403 continuously here — but it is a contributing factor, not the root cause of the outage originally described.

clubanderson (approved):
Reviewed the diff and re-derived the arithmetic: at 45 repos the old fixed 10s tick issues 16,200 list calls/hour against a 6,900/hour installation allowance, and the new interval lands the sweep inside a 25% share including the per-candidate Get allowance. Every constant is named with its rationale in a comment — selfAuthoredSweepBudgetShare, githubAppHourlyRateLimit, selfAuthoredSweepCandidateAllowance, selfAuthoredSweepSmallHiveRepos — which is exactly this repo's convention, and the honest framing of the candidate allowance as an allowance rather than a measurement is appreciated. Small hives keep the 10s floor byte-identical, and the tests pin the budget bound, monotonicity, small-hive invariance, and the 45-repo outage case specifically. Adding repos to the startup log line makes this diagnosable without arithmetic.

One note for the future, not blocking: githubAppHourlyRateLimit is a static assumption about the installation allowance, which GitHub scales with installation size. As a sizing heuristic that is fine — especially with #3943 handling recovery when reality disagrees — and it is a single constant to revisit if that ever drifts. Merging on green.

PR #3924, opened 2026-08-15
🐛 fix: guard mode file reads in git-credential-hive.sh and gh-wrapper.sh so an unreadable file cannot kill them

### LINKED ISSUE #3679: [sec-check] gh-wrapper.sh crashes on unreadable mode file — agent mode enforcement silently bypassed
## Security Finding

**Severity**: high
**Type**: unsafe-pattern / security enforcement bypass

The `bin/gh-wrapper.sh` script runs under `set -euo pipefail` and reads the per-agent mode file without error handling:

```bash
MODE_FILE="/tmp/.hive-mode-${AGENT_NAME_GW}"
if [ -f "$MODE_FILE" ]; then
  AGENT_MODE="$(cat "$MODE_FILE")"   # crashes on EACCES
else
  AGENT_MODE="${HIVE_AGENT_MODE:-}"
fi
```

On the live system the mode files are created as `0600 dev:node`:

```
-rw-------. 1 dev node 14 /tmp/.hive-mode-sec-check
```

Agent UIDs (e.g. `hive-sec-check`, uid=2008, group=node) can **see** the file (`-f` is true) but cannot **read** it. Under `set -e`, the failing `cat` inside a command substitution causes the wrapper to exit with error 1 immediately — before any mode gate, before any repo restriction, before `exec "$REAL_GH"`.

**Result**: every `gh` invocation by any agent fails with exit 1 and no output. The ACMM mode gates (ISSUES_ONLY / NO_GITHUB / ISSUES_AND_PRS hold) are **never evaluated**.

## Impact

- All agents are locked out of the `gh` CLI, breaking automated workflows.
- The mode-based security enforcement (which constrains what agents can do on GitHub) silently fails for the mode-file-based hot-reload path.

## Recommendation

Add a safe fallback in `bin/gh-wrapper.sh`:

```bash
MODE_FILE="/tmp/.hive-mode-${AGENT_NAME_GW}"
if [ -f "$MODE_FILE" ]; then
  AGENT_MODE="$(cat "$MODE_FILE" 2>/dev/null)" || AGENT_MODE="${HIVE_AGENT_MODE:-}"
else
  AGENT_MODE="${HIVE_AGENT_MODE:-}"
fi
```

Also verify the deployed hive binary creates mode files as `0644` (as specified in `v2/pkg/agent/manager.go` `0o644`) — the live files are `0600` which suggests the binary may not match the source.

---

### LINKED ISSUE #3881: 🐛 bug: git-credential-hive.sh dies under set -e when the agent mode file is unreadable, blocking all agent pushes
## Problem

`bin/git-credential-hive.sh` runs with `set -euo pipefail` and reads the hot-reloadable mode file like this:

```sh
MODE_FILE="/tmp/.hive-mode-${AGENT}"
if [ -f "$MODE_FILE" ]; then
  AGENT_MODE="$(cat "$MODE_FILE")"
```

The guard only checks existence (`-f`), not readability. When `/tmp/.hive-mode-<agent>` exists but is not readable by the agent's UID (see companion issue about umask-filtered mode-file permissions), `cat` fails and `set -e` kills the helper mid-script. Git receives no credential, and with `GIT_TERMINAL_PROMPT=0` in the agent environment every `git push` fails immediately.

## Impact

Every push-capable agent (ACMM L4/L5 modes) is unable to push branches or create hold PRs. Agents report the failure in their own words, e.g. *"Push/hold PR creation is blocked because the App Git credential is unavailable."* Observed live on a hosted hive at ACMM L5 after an L3→L4→L5 transition; a pod restart does not help because the mode files are recreated with the same permissions on every boot.

## Prior art

#3681 (`a244990e`) fixed this exact pattern in `bin/gh-wrapper.sh` by guarding with `[ -f ... ] && [ -r ... ]` and falling back to `$HIVE_AGENT_MODE`. `git-credential-hive.sh` has the same unguarded read, and because it is `set -e` the failure is fatal there rather than a graceful fallback.

## Fix

Mirror the #3681 guard: check `-r` in addition to `-f`, and tolerate a failing `cat` so the helper falls back to the `HIVE_AGENT_MODE` env var instead of dying. Mode enforcement is unchanged — the same mode value is enforced, just sourced from the env fallback when the file is unreadable.

bketelsen:
## Summary

Port of #3883 / #3884 (and #3681) from `v2` to `v4` — neither reader-side guard ever landed on `v4`.

- Both enforcement scripts run under `set -euo pipefail` and read the per-agent mode file `/tmp/.hive-mode-<agent>` guarded only by `-f`. On `v4` the Manager writes that file owner-only (`agentStateFileMode = 0600`, #3172) while the scripts run as the per-agent UID, so `cat` fails with `EACCES`, `set -e` kills the script mid-flight, and:
  - `bin/git-credential-hive.sh` returns no credential → with `GIT_TERMINAL_PROMPT=0` every push from every push-capable agent fails (#3881);
  - `bin/gh-wrapper.sh` exits 1 before any mode gate, repo restriction, or `exec "$REAL_GH"` runs → every agent `gh` call fails with no output (#3679 — closed by the `v2`-only #3681, so still live on `v4`).
- Guard both reads with `-r` in addition to `-f` and tolerate a failing `cat`, falling back to the `HIVE_AGENT_MODE` env var the Manager exported. Mode enforcement is unchanged: the same mode value is enforced, just sourced from the session env when the file cannot be read.
- Scope note: my original #3883 touched only `git-credential-hive.sh` because `gh-wrapper.sh` had already been fixed on `v2` by #3681. That fix is also missing on `v4`, so this PR carries the identical one-line guard for both scripts rather than leaving `gh` broken while fixing `git push`.
- Writer-side companion (make the files readable by the agent UID and self-heal running pods — port of #3885): opened separately.

## Related issues

Fixes #3881
Fixes #3679

## Testing

- [ ] `cd v2 && go build ./...`
- [ ] `cd v2 && go test ./...`
- [x] Other / not run (explain): shell-only change, no Go code touched. `bash -n` passes on both scripts. Behavioural check of the extracted mode-read block under `set -euo pipefail` with a `chmod 000` mode file and `HIVE_AGENT_MODE=ISSUES_AND_PRS`:

  ```
  --- BEFORE (origin/v4) ---
  cat: .../modefile-test: Permission denied
  cred: exit=1 (helper died, no AGENT_MODE)
  cat: .../modefile-test: Permission denied
  gh: exit=1 (helper died, no AGENT_MODE)
  --- AFTER ---
  cred: exit=ok AGENT_MODE=ISSUES_AND_PRS
  gh: exit=ok AGENT_MODE=ISSUES_AND_PRS
  --- AFTER, readable file (mode file still wins) ---
  cred: exit=ok AGENT_MODE=ISSUES_ONLY
  gh: exit=ok AGENT_MODE=ISSUES_ONLY
  ```

## Contributor checklist

- [x] PR targets `v4` — `v2` is no longer the maintained line for this work; CONTRIBUTING.md on `v4` still says `v2` and may want an update.
- [x] Docs, examples, and policies are updated when behavior changes. (No behavior change beyond the fallback; rationale is in the script comments.)

### DIFF
--- a/bin/gh-wrapper.sh
+++ b/bin/gh-wrapper.sh
@@ -366,10 +366,15 @@ fi
 
 # ── Mode-based enforcement (hot-reloadable via mode file) ──
 # Read mode from file first (updated by Manager on mode change), fallback to env var.
+# -r as well as -f: this script runs under `set -e`, so a mode file that exists
+# but is unreadable by the agent UID (owner-only perms, #3679) would kill the
+# wrapper before any mode gate or repo restriction ran, failing every gh call
+# with exit 1 and no output. Fall back to the env var instead — the same mode
+# value the Manager exported.
 AGENT_NAME_GW="${HIVE_AGENT:-${HIVE_AGENT_ID:-unknown}}"
 MODE_FILE="/tmp/.hive-mode-${AGENT_NAME_GW}"
-if [ -f "$MODE_FILE" ]; then
-  AGENT_MODE="$(cat "$MODE_FILE")"
+if [ -f "$MODE_FILE" ] && [ -r "$MODE_FILE" ]; then
+  AGENT_MODE="$(cat "$MODE_FILE" 2>/dev/null || true)"
 else
   AGENT_MODE="${HIVE_AGENT_MODE:-}"
 fi
--- a/bin/git-credential-hive.sh
+++ b/bin/git-credential-hive.sh
@@ -68,10 +68,14 @@ fi
 
 # ── Mode-based enforcement: block git push for agents without push capability ──
 
-# Read mode from file first (hot-reloadable), fallback to env var
+# Read mode from file first (hot-reloadable), fallback to env var.
+# -r as well as -f: this script runs under `set -e`, so a mode file that exists
+# but is unreadable by the agent UID (owner-only perms, #3679/#3881) would kill
+# the helper mid-script and block every push with no credential and no error.
+# Fall back to the env var instead — the same mode value the Manager exported.
 MODE_FILE="/tmp/.hive-mode-${AGENT}"
-if [ -f "$MODE_FILE" ]; then
-  AGENT_MODE="$(cat "$MODE_FILE")"
+if [ -f "$MODE_FILE" ] && [ -r "$MODE_FILE" ]; then
+  AGENT_MODE="$(cat "$MODE_FILE" 2>/dev/null || true)"
 else
   AGENT_MODE="${HIVE_AGENT_MODE:-}"
 fi


clubanderson:
Reviewed with the trust path in mind — this wrapper/credential pair is the merge gate behind the proxy hard-deny, so the bar here is "does the guard fail closed."

It does, for the case that matters: an agent cannot *induce* the fallback. The mode file is hive-owned in sticky `/tmp`, so the agent UID can neither chmod it unreadable nor replace it — the `-r` branch only ever fires on a writer-side perms bug (the 0600 state from #3172), which is exactly the state where today's behavior is "script dies, everything blocked with no error." Falling back to `HIVE_AGENT_MODE` (the Manager's own export) is strictly better than that, and the behavioral check in the PR body showing the file still winning when readable is the right invariant.

One small seam, non-blocking: inside the new branch, `cat ... || true` on a racing failure yields an *empty* `AGENT_MODE` rather than the env fallback — downstream that drops to the ACMM-level path instead of the exported mode. `[ -n "$AGENT_MODE" ] || AGENT_MODE="${HIVE_AGENT_MODE:-}"` after the cat would close it. Since a readable-but-empty file already produces the same state today and the race is not agent-forcible, fine to leave for a follow-up.

Merging; the writer-side companion (#3925) is going in alongside so the fallback becomes a rarely-exercised path rather than a load-bearing one.

Issue #3175, opened 2026-08-10
[scanner] N15 helper remains race-prone: path-based Chmod after O_NOFOLLOW write can be symlink-swapped


clubanderson:
## Finding

The new `writeAgentStateFile` hardening in PR #3172 still has a TOCTOU window after secure open/write: it closes the file descriptor, then calls `os.Chmod(path, 0o600)` on the pathname. In shared `/tmp`, another process can replace that pathname with a symlink between close and chmod.

This means the "symlink-safe" claim is incomplete: while content write is protected by `O_NOFOLLOW`, the post-write permission-tightening step is still path-racy.

## Steps to Reproduce / Evidence

Code path in PR #3172 (`v2/pkg/agent/manager.go`):

1. `os.OpenFile(path, O_WRONLY|O_CREATE|O_TRUNC|O_NOFOLLOW, 0o600)`
2. `f.Write(...)`
3. `f.Close()`
4. `os.Chmod(path, 0o600)`

The race exists between (3) and (4): pathname can be swapped in world-writable `/tmp`.

## Recommendation

Apply permissions via file descriptor before close (`f.Chmod(0o600)` / `syscall.Fchmod`) and avoid a path-based `Chmod` after releasing the FD. Keep `O_NOFOLLOW` for open-time protection.

---
*Filed by scanner agent (ACMM L4 — issues-only mode)*

clubanderson:
Verified still present on v4 @ 43f0c2231ce6d362783e8f726c50277fbd8b4848: v2/pkg/agent/manager.go:5730 — `writeAgentStateFile` still calls `os.Chmod(path, agentStateFileMode)` on the pathname after `f.Close()` (close at 5726, chmod at 5730). A comment (5727-5729) argues the path chmod is acceptable because the earlier `O_NOFOLLOW` open established the path is not a symlink, but that check is not atomic with the later chmod: in shared /tmp the pathname can be swapped for a symlink between close and chmod, and `os.Chmod` follows symlinks — the exact TOCTOU this issue describes. The fix remains an fd-based `f.Chmod(agentStateFileMode)` before close.

clubanderson:
Fixed in PR #3812 (merged to v4 as 329463acbe278b31b031b1774a245ec9f27b8419).

[...]

PR #3925, opened 2026-08-15
🐛 fix: make /tmp/.hive-mode-* readable by the agent UID and self-heal existing pods

### LINKED ISSUE #3679: [sec-check] gh-wrapper.sh crashes on unreadable mode file — agent mode enforcement silently bypassed

### LINKED ISSUE #3882: 🐛 bug: agent mode files /tmp/.hive-mode-* can be created unreadable (umask-filtered) and nothing ever repairs them
## Problem

The Manager writes the per-agent GitHub-mode files with:

```go
os.WriteFile(modeFile, []byte(mode.String()), 0o644)
```

in both `SyncModeFiles` (`pkg/agent/manager.go:4873`) and `agentEnvPairs` (`:5094`). Two problems with that:

1. `os.WriteFile`'s mode argument is filtered by the process umask at creation time. Under a restrictive umask the file is created 0600, and the readers that actually enforce the mode — `gh-wrapper.sh`, `git-credential-hive.sh` — run as per-agent UIDs and get `Permission denied`.
2. For a file that already exists, the mode argument is ignored entirely, so rewriting on level change or kick never repairs bad permissions. A pod restart doesn't help either: the files are recreated the same way every boot.

Observed live on a hosted hive at ACMM L5: `cat /tmp/.hive-mode-quality` → `Permission denied` from the agent's shell. Combined with the unguarded read in `git-credential-hive.sh` (#3881) this blocked every agent push.

## Fix

- `os.Chmod(modeFile, 0o644)` after each successful write in both call sites, so perms are explicit regardless of umask and repaired on existing files.
- Add the mode files to the permissions watcher's 10s tick so a **running** pod self-heals without operator intervention (same pattern as the `.bob` state-dir remediation), restricted to files owned by the hive uid — /tmp is sticky/world-writable, so the watcher must never act on a foreign or symlinked entry (CWE-59).

Related: #3881 (the reader-side guard in `git-credential-hive.sh`).

bketelsen:
## Summary

Port of #3885 (merged on `v2`) to `v4`. On `v4` the shape of the fix is different because of #3172, so this needs a careful read — @clubanderson, this touches your N15 change and I want to be explicit about why.

**The failure.** On `v4` the Manager writes the per-agent control files (`/tmp/.hive-mode-<agent>`, `/tmp/.hive-bootstrap-<agent>.txt`) through `writeAgentStateFile` at a hard-coded `agentStateFileMode = 0600` (#3172, audit N15). But their consumers do not run as the hive uid: `bin/gh-wrapper.sh` and `bin/git-credential-hive.sh` `cat` the mode file as the per-agent UID (`hive-<agent>`), and goose is launched with `--text "$(cat /tmp/.hive-bootstrap-<name>.txt)"` from the agent's own tmux session. Owner-only therefore means `EACCES` for every reader: under `set -e` both scripts die mid-flight, every agent `gh` call fails with exit 1 and every push has no credential. That is exactly what #3679 reported from a live system (`-rw------- dev node /tmp/.hive-mode-sec-check`, uid 2008 cannot read; the issue even noted the binary "may not match the source" — it was the `v4` build), and what #3881/#3882 hit again on a hosted hive at ACMM L5.

**Why 0644 does not undo N15.** The N15 hole was symlink-following on write (`os.WriteFile` without `O_NOFOLLOW`), and that fix — plus chmod through the open descriptor (#3175) — is fully retained. World-*read* was never the problem: `/tmp` is sticky, so an agent UID cannot unlink, rename or replace a hive-owned file there, and 0644 grants write to nobody but the owner. The contents are not secret either — the mode is the agent's own enforcement level, and the bootstrap prompt is what the agent prints in its own pane. Owner-only can never work for a file whose consumer is a different uid by design; it just silently degrades to "no gh, no push".

**The change**, mirroring #3885:

- `agentStateFileMode` 0600 → 0644 (owner-write, world-read). Because `writeAgentStateFile` already fchmods on every write, files left 0600 by the #3172 build (or umask-narrowed, #3882) are repaired on the next level change or kick, and group/world-writable files are still tightened. The N15 rationale comments are updated to match.
- The permissions watcher re-widens mode files on its 10-second tick (`fixModeFiles`), so an already-running pod self-heals with no operator intervention — same remediation pattern as the `.bob` state-dir fix. `/tmp` is sticky and world-writable, so the scan opens each path `O_NOFOLLOW` (a planted symlink is refused — CWE-59), stats and chmods on that same descriptor (no path-swap window — CWE-367, the same discipline `writeAgentStateFile` follows per #3175), skips anything that is not a regular file owned by the hive uid, and only ever adds read bits.

If you would rather keep the read narrower than world, 0640 with the group forced to `NodeGID` would also work for the per-agent-UID model — say so and I'll rework it; I went with 0644 to match what was merged on `v2`.

## Related issues

Fixes #3882
Fixes #3679

Reader-side companion (`-r` guards in `git-credential-hive.sh` / `gh-wrapper.sh`, port of #3883/#3884/#3681): #3924.

## Testing

- [x] `cd v2 && go build ./...`
- [x] `cd v2 && go test ./...` — full `./pkg/agent` suite passes (`ok ... 334s`); every other package passes except two that fail identically without this change and are unrelated to it (`pkg/policies` assumes a `master` default branch under a `init.defaultBranch=main` git config; `v2/test` needs a live hive at `192.168.4.85:3003`). Note for anyone reproducing: `TestStart_*`/`TestRestart*` in `tmux_coverage_test.go` fail with `File name too long` when the checkout path is deep — the tmux socket dir is created relative to the package dir — so run from a short path.
- [x] Other / not run (explain): tests are fail-before/pass-after — with only `manager.go` reverted, `TestWriteAgentStateFileIsOwnerWriteWorldRead` and `TestWriteAgentStateFileRepairsExistingMode` fail (`mode = 600, want 644`); the new `permissions_modefile_test.go` covers widening of a 0600 file, idempotence on 0644, planted-symlink skip and non-regular skip. `TestWriteAgentStateFileRefusesSymlink`, `...ChmodsViaDescriptor` and `...OverwritesRegularFile` are unchanged and still pass.

## Contributor checklist

- [x] PR targets `v4` — `v2` is no longer the maintained line for this work; CONTRIBUTING.md on `v4` still says `v2` and may want an update.
- [x] Docs, examples, and policies are updated when behavior changes. (Behavior change is the file mode + self-healing perms; documented on `agentStateFileMode`, `writeAgentStateFile`, `ModeFileGlob`, `fixModeFiles`.)

### DIFF
--- a/v2/pkg/agent/agent_state_file_test.go
+++ b/v2/pkg/agent/agent_state_file_test.go
@@ -5,11 +5,14 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 	"testing"
 )
 
 // N15 (CWE-367/732/20): per-agent control state in the pod-shared /tmp must be
-// owner-only and must not follow symlinks.
+// owner-WRITE-only and must not follow symlinks. It must remain world-readable,
+// because its consumers run as the per-agent UID, not the hive uid — see
+// agentStateFileMode and #3679/#3881/#3882.
 //
 // Two files matter, both at paths derived from the agent NAME (guessable):
 //
@@ -20,9 +23,14 @@ import (
 //     its first instruction, so planting it injects an attacker-chosen prompt
 //     that runs with the victim's credentials at message zero.
 //
-// Both were written with a plain os.WriteFile at 0o644: world-readable, and
-// symlink-following (os.WriteFile passes no O_NOFOLLOW), so a pre-planted link
-// redirected the hive's own write to any file the process could reach.
+// Both were written with a plain os.WriteFile at 0o644, symlink-following
+// (os.WriteFile passes no O_NOFOLLOW), so a pre-planted link redirected the
+// hive's own write to any file the process could reach. The 0644 itself was
+// not the hole: /tmp is sticky, so an agent UID cannot replace a hive-owned
+// file there, and world-read is REQUIRED — gh-wrapper.sh, git-credential-hive.sh
+// and goose's `--text "$(cat ...)"` all read these files as the agent UID. The
+// original N15 fix tightened to 0600 and locked every agent out of gh and git
+// push (#3679/#3881); the mode is now 0644 with the symlink hardening kept.
 //
 // NOTE on scope: an earlier reading of this called the bootstrap path arbitrary
 // CODE execution, on the theory that a nested $(...) inside the file would be
@@ -31,9 +39,16 @@ import (
 // content reaches goose as literal text. The finding is prompt injection, which
 // these file permissions are the correct fix for.
 
-func TestWriteAgentStateFileIsOwnerOnly(t *testing.T) {
+// TestWriteAgentStateFileIsOwnerWriteWorldRead pins the mode: only the hive uid
+// may write (no group/other write bits — that is what would let one agent steer
+// another), and everyone may read (the per-agent UID that consumes the file is
+// not the hive uid). The write must be exact regardless of the caller's umask.
+func TestWriteAgentStateFileIsOwnerWriteWorldRead(t *testing.T) {
 	dir := t.TempDir()
 	path := filepath.Join(dir, ".hive-mode-victim")
+
+	restore := syscall.Umask(0o077) // the restrictive umask that produced 0600 in prod (#3882)
+	defer syscall.Umask(restore)
 
 	if err := writeAgentStateFile(path, []byte("ISSUES_ONLY")); err != nil {
 		t.Fatalf("writeAgentStateFile: %v", err)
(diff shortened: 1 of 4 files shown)


clubanderson:
Read this against the N15 history before anything else, since it deliberately relaxes an audit fix. The reasoning holds:

- The N15 hole was symlink-following on write, and that is fully retained (O_NOFOLLOW + fchmod-through-the-descriptor). World-*read* was never the finding; owner-only for a file whose designed consumer is a different UID was just a broken state dressed as hardening.
- The property that actually matters — the agent UID must not be able to flip its own mode — survives: 0644 grants write to nobody but the hive uid, and sticky `/tmp` blocks unlink/rename/replace. Checked that `fixModeFiles` can't be abused either: O_NOFOLLOW at open, regular-file + hive-uid ownership checks and the chmod all on one descriptor, and the only mutation it can ever make is adding read bits. A planted symlink or directory at a matching name is a no-op, and the tests pin all of that.
- Test quality is above the bar for this area: the umask pinning in `TestWriteAgentStateFileIsOwnerWriteWorldRead` reproduces the exact #3882 failure shape, and the repair test covers both directions (0600 widened, 0666 tightened).

One observation, not a blocker: the watcher glob covers only `.hive-mode-*`, so a 0600 `.hive-bootstrap-*` file heals only on the next `writeAgentStateFile` rewrite. That is fine in practice — goose reads it at launch, which immediately follows a fresh write — but worth remembering if a bootstrap-read-at-tick consumer ever appears.

Merging together with the reader-side guards in #3924.

PR #3926, opened 2026-08-15
fix(contributor): stop the relay defaulting its token write to the hub's full-privilege cache path

Danathar:
## What

The contributor relay writes the hub-delivered, **task-scoped** GitHub token to `GH_TOKEN_CACHE` (`injectGhToken`, fed by `task_assign.github_token` / `token_refresh`). Whenever `/var/run/hive-metrics` exists, that path defaulted to `/var/run/hive-metrics/gh-app-token.cache` — the exact filename of the hub's **full-privilege** installation-token cache (`bin/gh-app-token.sh`, root-owned `0600` since audit H3 / `69256ecf`).

On a host running both hive-hub components and a contributor relay (native install), that collision means:

- **relay as root** → silently overwrites the hub's installation-token cache with a short-TTL repo-scoped contributor token, corrupting every hub-side consumer (`bin/hive-merge.sh:90`, entrypoint's `HIVE_GITHUB_TOKEN` at `v2/deploy/entrypoint.sh:1039`) until the next refresh;
- **relay as any other uid** → `fs.writeFileSync` throws `EACCES`, uncaught in the ws `'message'` handler (`injectGhToken` only try/caught the `mkdir`), crashing the relay on **every** `task_assign` that carries a token — a crash loop on assignment, before `task_accepted` is even sent;
- either way, `detectCapabilities()` misreported the hub's own cache as this relay's credential (`credential_type: 'app'`).

## Fix

1. Default to a **relay-owned filename** (`contributor-gh-token.cache`) in the same directory. The contributor container behaviour is unchanged — it owns `/var/run/hive-metrics` (`v2/Dockerfile.contributor:170-171`) and, verified by repo-wide search, **nothing reads the relay-written file by path** (the wrapper in contributor mode injects no token; the inline cred helper bakes `GH_TOKEN` at container start), so no consumer needed updating. The `HIVE_GH_TOKEN_CACHE` override is preserved.
2. `injectGhToken` now degrades loudly instead of throwing — a failed token write logs and continues (the agent can still work with its own `GH_TOKEN`); it must never kill the relay mid-assignment.

Two regression tests added (`bin/contributor-relay.test.js`, 68/68 passing): the default path must never be the hub's shared cache filename, and a `task_assign` with an unwritable cache path must still produce `task_accepted`.

## Context

Part of the #1861 constellation and the concrete contributor-relay half of the **N14 re-verification** asked for in #3842 (full verification writeup posted there). Complements #3888 (N7) and #3889 (kick-agents N14 instruction fix) — no overlap with either.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -43,8 +43,19 @@ const MODEL = process.env.AGENT_MODEL || process.env.GOOSE_MODEL || '';
 const REASONING_EFFORT = process.env.AGENT_REASONING_EFFORT || '';
 const AGENT_ROLE = (process.env.HIVE_AGENT_ROLE || '').trim();
 const TMUX_SESSION = process.env.HIVE_AGENT_SESSION || 'contributor';
+// Where the hub-delivered, task-scoped token is written (injectGhToken). This
+// deliberately does NOT default to /var/run/hive-metrics/gh-app-token.cache:
+// that filename is the hub's FULL-privilege installation-token cache
+// (bin/gh-app-token.sh, root-owned 0600 since audit H3). A relay started on a
+// host that also runs hive-hub components (native install) would either
+// clobber that cache with a short-lived repo-scoped token (relay running as
+// root) or die on EACCES trying (any other uid — the write was uncaught), and
+// detectCapabilities() would misreport the hub's own cache as this relay's
+// credential. Distinct filename, same directory, so the contributor container
+// (which owns /var/run/hive-metrics — v2/Dockerfile.contributor) behaves as
+// before. See kubestellar/hive#1861 / #3842 (audit N14).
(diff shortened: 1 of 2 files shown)


clubanderson:
Verified both halves. The default-path collision is real: `gh-app-token.cache` is the hub's root-owned 0600 full-token cache, so a relay on a native-install host either clobbered it as root or crash-looped on EACCES — and I confirmed nothing inside the contributor container reads the relay's write path by name except the relay itself (the wrapper in contributor mode uses the personal `GH_TOKEN`, and `detectCapabilities` reads the same constant), so the rename is behavior-preserving for containerized relays while an explicit `HIVE_GH_TOKEN_CACHE` override still wins. The try/catch around `injectGhToken` is the right degradation: `task_assign` calls it before `task_accepted`, so an unwritable path was a crash loop per assignment, not a one-off. Both tests are positive controls — the basename assertion catches a regression to the hub path, and the file-as-parent-directory fixture makes the write fail regardless of uid and asserts the task is still accepted. Merging.

Issue #2875, opened 2026-08-08
[scanner] ioscan semantic classifier can flood advisory beads with duplicate high-priority entries


clubanderson:
## Finding

PR #2868 wires ioscan classifier advisories directly to bead creation without deduplication or rate limits. In `v2/pkg/scheduler/ioscan_enforce.go`, `recordClassifierAdvisory` emits an advisory on every `warn`/`redact`/`block` action. In `v2/cmd/hive/main.go`, `SetAdvisoryFunc` persists each advisory via `store.Create(title, beads.TypeAdvisory, beads.PriorityHigh, agentName, "")`.

This means repeated kicks over the same untrusted text (or repeated classifier warnings on noisy repositories) will create a new high-priority bead each time, even when the underlying signal is identical.

## Steps to Reproduce / Evidence

1. Review PR #2868 diff:
   - `v2/pkg/scheduler/ioscan_enforce.go` adds `recordClassifierAdvisory(...)` and calls it on classifier warn/redact/block paths.
   - `v2/cmd/hive/main.go` adds `sched.SetAdvisoryFunc(...)` that always creates a new advisory bead.
2. There is no content-hash key, cooldown window, or "already reported" guard in either path.
3. Classifier execution budget resets per kick build (`resetClassifierBudget()`), so the write path can trigger repeatedly across scheduler cycles.

## Recommendation

Add dedupe and throttling before bead creation, for example:

- use a stable key `(rule, category, normalized segment hash, action)` and skip if recently reported,
- downgrade repeated duplicates to audit-only logs,
- optionally lower default severity for classifier warnings until confidence tuning stabilizes.

---
*Filed by scanner agent (ACMM L4 — issues-only mode)*

clubanderson:
Rechecked current origin/v2 before making changes. The reported flood path is no longer live in v2 HEAD:

- `v2/docs/ioscan.md:3` states ioscan is not present in v2 HEAD: no ioscan package, no `ioscan:` config block, and no canary/fail-mode implementation in the kick path.
- `v2/docs/ioscan.md:29` tells operators not to enable ioscan in v2 configs unless it is reintroduced.
[...]

PR #3927, opened 2026-08-15
fix(advisory): stop near-duplicate findings from flooding the digest past GitHub's comment limit

### LINKED ISSUE #2364: 🐝 Hive Advisory Report
This issue collects advisory findings from Hive agents.

At lower ACMM levels, some agents work in advisory mode — they analyze code and post findings here but do not create issues or PRs. At higher levels, designated agents (e.g. quality) can open issues and PRs directly, while other agents remain advisory-only.

The governor posts periodic digest comments summarizing what advisory agents found.

**Do not close this issue.** It is a living document.

Danathar:
## The symptom, on the advisory issue itself

The digest comment on #2364 (2026-08-15 15:08 EDT) ends with:

> ⚠️ *Digest truncated: 174962 → 65196 characters (GitHub limit: 65536)*

It reports **334 findings** and renders CRITICAL and HIGH; the 136 medium and 50 low findings never appear, because `truncateDigest` (`v2/pkg/github/advisory.go:209`) trims from the bottom. **37 of the rendered entries are one signal** — the pr-verifier workflow calling a deleted reusable workflow — re-filed by the `ci-maintainer` agent:

```
pr-verifier.yml broken: reusable workflow missing from kubestellar/infra
pr-verifier.yml failing 100% — reusable workflow missing from kubestellar/infra
PR Verifier failing on all PRs — reusable-pr-verifier.yml deleted from kubestellar/infra
pr-verifier.yml fails on every PR — reusable workflow deleted upstream
…
```

## Fix — a render cap, plus a narrow dedup improvement

**1. Per-(agent, finding-type) render cap** of 5 per severity section, newest kept, remainder collapsed into an explicit line:

```
- _…plus 32 more [ci-failure] findings from ci-maintainer, collapsed to keep lower-severity sections within GitHub's comment limit_
```

**This is the load-bearing half.** I extracted the 37 real titles from the live digest and ran them through the new renderer: **37 in → 5 rendered + 1 collapse line, 1,055 chars** for that group. The summary table and section headers keep the **true** counts, so nothing vanishes silently — a reader sees that 37 exist and 32 were collapsed, instead of today's behavior where the entire lower-severity half is silently gone.

**2. Normalized dedup key** in `BuildDigestFromBeads`: lowercase, digit runs folded to `#`, punctuation dropped, replacing the exact-title `seen[b.Title]` check. This catches the mechanical variants (`"failing (run #3279)"` vs `"failing (run #3291)"`).

**Measured honestly: on this real dataset the dedup layer alone collapses 37 titles to only 36 distinct keys.** These are genuine human-style paraphrases ("broken:", "failing 100% —", "fails on every PR:"), not digit drift, so normalization barely touches them. I kept it because it is strictly better than exact-title matching and costs nothing, but I do not want the PR to claim it solves the flood — the cap does. Deliberately no stemming or token-similarity matching: that is the false-positive mode called out in the #2875 review, and merging two findings that differ in *words* is worse than rendering both.

## Scope / verification

- `v2/pkg/advisory/advisory.go` only; `go build ./...` clean, `go vet` clean, `go test ./pkg/advisory/ ./pkg/github/` pass (the latter covers `truncateDigest` and the digest post path).
- Three regression tests added: key normalization (merge and must-not-merge directions, plus the all-symbol edge case), bead-level dedup of run-number variants, and the render cap (collapse line present, newest kept, other finding-types unaffected, header count uncapped).
- Deliberately **not** `Fixes #2364` — that issue is a standing living document ("Do not close this issue"), and #3925 removed `help wanted` from it for the same reason.

Related: #2875 (closed) fixed the ioscan *creation*-side bead flood. This is the render side, which is where paraphrase floods have to be handled — no creation-side guard can recognize that "PR Verifier broken" and "pr-verifier.yml fails on every PR" are the same finding.


clubanderson:
Reviewed against #3900, which merged first and covers the dedup half: `collapseNearDuplicates` now runs in both `BuildDigest` and `BuildDigestFromBeads`, so `normalizedFindingKey` is largely subsumed (kept anyway — it's a harmless cheap pre-filter, exactly as this PR's own honest measurement predicted). The render cap is the complementary, load-bearing half and is not redundant: #3900's collapse took the live corpus 92→56 findings, which still leaves a 174k-char digest far past the 65,536 limit, and only a render-layer bound guarantees the medium/low sections survive truncation for paraphrases below the 0.5 similarity threshold.

I resolved the merge conflict on this branch (merge commit 4163462) keeping both layers, with one semantic fix: the cap test's `variant A..H` titles tokenize identically under #3900's Jaccard collapse and would have merged into a single finding before the renderer ever saw them, failing the test. The fixture now uses eight distinct-worded titles (pairwise Jaccard 0.11 vs the 0.5 threshold) — which is also truer to the cap's purpose, since it exists precisely for restatements the dedup layer cannot safely merge. Merging on green via the fork-CI gate branch.

Issue #2567, opened 2026-08-04
contribute: discovery and WebSocket auth expose no API/protocol version or canonical spoke URL


castrojo:
> Part of #2534 (contributor onboarding field notes from `projectbluefin/review`).

## Summary

The contributor path now evolves quickly, but discovery and WebSocket authentication do not identify the protocol/API version, supported message set, or canonical spoke API base. A downstream client can authenticate successfully without knowing whether the selected Hive supports `token_refresh`, the newer `task_unavailable` reasons, prompt preview metadata, or future capability declaration.

This is adjacent to #2547 but distinct: #2547 asks what a **client** can run and whether Hive may route on it. This asks how both sides know which **protocol and deployed server capabilities** they are speaking.

## Evidence

At `v2` commit [`08518560`](https://github.com/kubestellar/hive/commit/08518560e14cc583215765831cf72487d432d1c1):

- `auth_response` carries registration token, backend, model, and role; `auth_ok` returns contributor identity, trust tier, permissions, and role. Neither side sends a protocol version or capability set ([`contribute_ws.go:1013-1105`](https://github.com/kubestellar/hive/blob/08518560e14cc583215765831cf72487d432d1c1/v2/pkg/dashboard/contribute_ws.go#L1013-L1105)).
- The relay sends the same unversioned handshake ([`contributor-relay.sh:704-723`](https://github.com/kubestellar/hive/blob/08518560e14cc583215765831cf72487d432d1c1/bin/contributor-relay.sh#L704-L723)).
- The past 24 hours added multiple contributor API/protocol behaviors (#2456, #2542, #2558, #2562) while maintained branches and deployed spokes moved independently.

The public topology also has two surfaces:

- [`hive.kubestellar.io/api/contribute/status`](https://hive.kubestellar.io/api/contribute/status) is Hub discovery and returns the selected Hive identity.
- Spoke-only APIs such as [`/api/contribute/fleet`](https://hosted-projectbluefin-knuckle-gjvq.hive.kubestellar.io/api/contribute/fleet) live on the selected Hive’s dashboard route, not the Hub root.

#2520 improved namespace/vanity-route identity, but the discovery response still does not provide a canonical dashboard/API URL or deployed SHA/capability identifier.

[...]

PR #3930, opened 2026-08-15
✨ feature(contribute): let both sides detect an incompatible protocol peer (#2547 compatibility criterion)

Danathar:
## What

Makes an incompatible contributor-protocol peer **detectable from both sides**, instead of only discoverable by watching it misbehave.

Refs #2547 — **no `Fixes`.** This is the peer-compatibility acceptance criterion, not ROUTE. ROUTE stays open and untouched.

## The criterion this closes

> Either the protocol gains a way for both sides to detect an incompatible peer, or it is documented that compatibility is managed out of band and how.

DECLARE (#2659/#2600, pinned by #3815, documented by #3830) shipped the *declaration*. #2567 gave each side a version to **state**. Neither side ever **compared** them, so the sentence from the issue body is still literally true today:

> with no version on either side, the only way to learn that an old relay is talking to a new hub is to watch it misbehave

The ops row renders the client's declared `proto 1.1` — with nothing to compare it against.

## The drift already happened

This is the part that convinced me it was worth doing:

| | hub | relay |
|---|---|---|
| #2600 | `1.1` | `1.1` |
| #2671 (`credential_after_accept`) | **`1.2`** | `1.1` |

`bin/contributor-relay.sh` carried an explicit `Keep in step with contributorProtocolVersion` comment. Nothing enforced it, and the in-tree relay has under-declared itself ever since — while `v2/docs/contributor-relay.md` said "Since protocol 1.2 the relay self-reports…". A comment cannot hold that invariant.

## What changed

**One comparison, one vocabulary, both sides.** `MAJOR.MINOR` → `unknown` / `current` / `older` / `newer` / `incompatible` / `malformed`. The verdict always describes *the peer*, so the same `older` means "the hub is older" on the relay and "the client is older" on the hub; every message names both versions explicitly so it cannot be misread.

**Hub.** `FleetClanker.Protocol` carries hub version, peer version, verdict and operator prose. The ops row shows `protocol: client 1.1 · hub 1.2 · older than this hub`; the hub log records the verdict at connect. Derived per snapshot from the capability the client already declares — no new connection state, and no new field on the wire in either direction.

**Relay.** Compares the hub's advertised version instead of only logging it, and prints one advisory line per hub on the contributor's own terminal. That is the half a hub log cannot deliver: whoever runs the relay is usually not whoever reads the hub.

```
Protocol incompatible: hub 2.0 differs from this relay 1.2 in MAJOR version — the wire
contract differs and behaviour is undefined; consider updating the relay. Continuing
normally — this is advisory and nothing is gated on it.
```

**The drift, fixed and pinned.** The relay moves to `1.2` — it already handles the `token_refresh` credential delivery that version denotes — and the two constants are now pinned in step from *both* sides (`TestRelayProtocolVersionMatchesHub` in Go, and a Go-source-reading test in `contributor-relay.test.js`), so the next bump either happens on both sides or fails the build.

## The line this holds

**Nothing is gated, in either direction.** A drifted or even majorly-incompatible relay authenticates, is admitted, and receives exactly the work it received before. A relay that sees a hub version it does not recognise keeps working normally.

A declared version is self-reported client text. Acting on it would be *routing on a value the client controls* — the failure mode the issue names, where a mis-declaring client produces a fleet-level symptom that presents as a hub bug. So this stays on the same side of the #3815 line as DECLARE, and is pinned the same way:

- `TestProtocolCompatIsNotReadBySelection` — `selectTask` / `RequeueContributorTask` may not even mention the comparison. Carries the same instruction as the capability guard: *if routing has now been decided and recorded by a maintainer, update this test in that same PR.*
- `TestWS_IncompatiblePeerIsReportedNotRejected` — a client declaring a **major** mismatch, the loudest verdict reachable, still gets `auth_ok` and is served.
- Relay side: `warnOnProtocolDrift` may not contain `process.exit` / `throw`, and a behavioural test asserts a mismatch leaves `authFailed`, `authenticated` and the socket untouched. A relay that stopped on a version difference would strand its own contributor over an advisory signal.

## Backward compatibility

A client that declares nothing reads as `unknown`, is **not** flagged as a mismatch, and renders nothing. The issue is emphatic that silence must never be read as a claim of incompatibility, and that compatibility has to be carried by the defaults since there is no negotiation to carry it. `TestWS_UnversionedPeerIsQuiet` pins the whole shape: nil capabilities, verdict `unknown`, `mismatch` false, empty peer.

Matching versions also render nothing, so a healthy fleet stays exactly as quiet as it is today — and the warning means something when it does appear.

## Handling of client-controlled data

`relay_protocol_version` is the one client-controlled string reaching this payload:

- **Bounded** to 32 runes, rune-safe. Nothing obliges a relay to keep it short.
- **Never interpolated into the operator prose.** `Detail` is server-authored and names only the hub's own constant; the client's value travels separately in `Peer` and is escaped at render. A hostile version string cannot dress itself up as a hub statement on an operator's screen.
- **Strictly parsed.** `1.2.3`, `v1.2`, `1`, `1.x` land on `malformed` rather than being coerced into a comparison that would report a confident, wrong verdict.

## Docs

`v2/docs/contributor-relay.md` gains a **Protocol compatibility** section: what is compared, what each side shows, what MAJOR vs MINOR means, and — stated plainly — that nothing is gated on it, with a pointer to the server-side controls that *are* enforced if an operator actually needs to shape the queue.

## Tests

`v2/pkg/dashboard/contribute_protocol_compat_test.go` (11 tests) and 3 new tests in `bin/contributor-relay.test.js`. Full `pkg/dashboard` suite green (51s); `node bin/contributor-relay.test.js` 69/69.

### Regression replay

Each change neutered, the named test confirmed failing, then restored green.

| # | Neutered | Test that caught it |
|---|---|---|
| 1 | relay back to `1.1` | `TestRelayProtocolVersionMatchesHub` |
| 2 | `Detail` built from raw client text | `TestPeerProtocolCompat_HandlesHostileDeclaredVersion` |
| 3 | length bound removed (554 runes through) | same |
| 4 | `unknown` counted as a mismatch | `TestWS_UnversionedPeerIsQuiet`, `TestPeerProtocolCompat_MismatchFlag` |
| 5 | hub sends `auth_failed` on incompatible | `TestWS_IncompatiblePeerIsReportedNotRejected` |
| 6 | `classifyPeerProtocol` read inside `selectTask` | `TestProtocolCompatIsNotReadBySelection` |
| 7 | ops row stops rendering the line | `TestOpsPage_ProtocolMismatchIsVisibleAndAdvisory` |
| 8 | relay `process.exit(1)` on incompatible | `TestRelayComparesHubProtocolVersion` |
| 9 | relay drops socket / fails auth on drift | relay: "never changes relay behaviour" |
| 10 | relay warns on every reconnect | same (dedupe assertion) |
| 11 | relay version drifts (JS half) | relay: "declares the same protocol version" |

No task-side requirements vocabulary, no dispatch change, no capability read anywhere in selection. The three ROUTE design questions in #2547 are still open for the maintainer and the downstream reporter; nothing here presumes an answer to any of them.

Separately open from me and independent of this: #3901 (failure attribution), which touches `contribute_protocol.go` and `contribute_ws.go` in different places. Either can merge first; I'll rebase the other.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -123,9 +123,16 @@ const TASK_UNAVAILABLE_RETRY_MS = 30000;
 // RELAY_PROTOCOL_VERSION is the contributor-protocol version this relay speaks
 // (kubestellar/hive#2567). It is DECLARED to the hub in auth_response (additive,
 // optional — an older hub simply ignores it) and the hub advertises its own
-// version + capability set back on auth_ok. Keep in step with
-// contributorProtocolVersion in v2/pkg/dashboard/contribute_protocol.go.
-const RELAY_PROTOCOL_VERSION = '1.1';
+// version + capability set back on auth_ok.
+//
+// MUST equal contributorProtocolVersion in v2/pkg/dashboard/contribute_protocol.go:
+// the hub and this relay ship from the same tree, so they speak the same version
+// by construction. That was previously only a comment, and it drifted — #2600
+// shipped both at 1.1, #2671 bumped the hub to 1.2 for credential_after_accept
+// (handled by the token_refresh case below) and left this at 1.1, so the relay
+// under-declared itself for months with nothing to notice. It is now pinned by
+// TestRelayProtocolVersionMatchesHub, which fails the build on the next drift.
+const RELAY_PROTOCOL_VERSION = '1.2';
 
 // Per-task CLI-crash retry budget. Issue #2203: a task whose CLI kept dying was
 // reassigned by the hub and failed identically forever (5+ times in ~20min),
@@ -161,6 +168,9 @@ const hubs = rawHubList.map((url, i) => ({
   reconnectTimer: null,
   authenticated: false,
   authFailed: false,
+  // #2547: set once we have reported a contributor-protocol difference with
+  // this hub, so a reconnect loop does not repeat the same advisory line.
+  protocolDriftReported: false,
 }));
 // Index into hubs[] of the hub we are currently soliciting work from (sent it
 // the last 'ready'), or that owns currentTask. Round-robins forward on an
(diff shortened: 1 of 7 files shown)


Danathar:
Rebased onto v4 (`0910c95f`). The conflict was with #3948, which I also
authored — it landed in the same files while this was open.

One resolution is worth flagging for review, since it changed this PR's diff:
both PRs independently added a `waitForClanker` test helper, so the merge
compiled to a redeclaration. I dropped the copy in
`contribute_protocol_compat_test.go` and this PR now uses the one #3948 added
in `contribute_capability_bounds_test.go`. That helper returns `FleetClanker`
where mine returned `*FleetClanker`; all three call sites here only read
`fc.Protocol` / `fc.Capabilities`, so it's a drop-in with no behaviour change.
That accounts for the test file being 19 lines shorter than before the rebase.
The remaining conflicts were additive on both sides and kept both halves.
Everything else is line-identical to the pre-rebase diff.

Locally on the rebased tree: `bin/contributor-relay.test.js` 75/75,
`go test ./pkg/dashboard/` passes, `go vet` and `go build ./...` clean.

Two notes on the current state:

- The `needs-rebase` label looks stale — GitHub reports the PR mergeable.
  GitHub's API was throwing intermittent 503s around the time I pushed, which
  may be why the sweep hasn't re-evaluated.
- `verify PR contents` failed during *Set up job* with HTTP 429 fetching
  `docker/login-action`, before any repo code ran. I don't have rights to
  rerun it; it should clear on a rerun or the next push.

clubanderson:
Reviewed with the "old client ↔ new hub must not brick" criterion up front, and it holds on both sides: the hub logs/annotates and still admits every verdict including MAJOR-incompatible and malformed (pinned by TestWS_IncompatiblePeerIsReportedNotRejected and TestProtocolCompat_NeverGates), the relay warns once per hub and changes nothing about what it sends or requests, and an unversioned peer reads as `unknown` and stays silent — which is exactly what every pre-#2567 relay looks like. The refusal to gate is well-reasoned beyond compatibility: the version is client-controlled text, so routing on it would hand the client the keys. Good hygiene details: the verdict `Detail` interpolates no client text, the declared version is rune-bounded and escaped at render, and the operator line is amber-not-red with an explicit "still served" note so it can't be misread as a fault state. Catching the real drift (relay stuck declaring 1.1 since #2671 bumped the hub to 1.2, against an unenforced "keep in step" comment) and pinning it with TestRelayProtocolVersionMatchesHub closes the exact hole that let this criterion rot. Merging.

PR #3931, opened 2026-08-15
🔒 fix(scheduler): stop handing every agent a live GitHub token in its kick prompt (#1861)

Danathar:
## What

Stops the kick prompt handing every container-hosted agent a live GitHub App token. This is the **container-path twin of #3889**, and the part of #1861 that reaches the issue's stated goal — *"agents then need no token material at all"* — without waiting on proxy-side injection.

Refs #1861 — **no `Fixes`.** Item 1 (injection) stays deferred exactly as the 2026-08-14 triage recorded.

## The instruction

Every kick prompt (`${GH_AUTH}`, rendered for supervisor / architect / outreach / sec-check and every templated agent) said, verbatim:

```
- gh CLI: export your per-agent token first:
    export GH_TOKEN=$(cat /var/run/hive-metrics/agent-tokens/gh-token-<agent>.cache)
```

That is wrong three separate ways.

**1. It leaks a credential into the agent's own context.** A token the agent types is a token in its transcript and in anything it echoes — one prompt injection from exfiltration. This is precisely the exposure #3842/#3889 removed from the *native-install* prompt. It differs only in blast radius (tier-scoped here, fleet-wide there), not in kind. #1861 exists because agents should hold no token material, and the container path was still handing them some.

**2. It was redundant.** `v2/Dockerfile` installs the wrapper **as** `gh`:

```
COPY bin/gh-wrapper.sh /usr/local/bin/gh
```

and `bin/gh-wrapper.sh:76` already does `export GH_TOKEN="$(cat "$HIVE_AGENT_TOKEN_CACHE")"` before the agent's command runs. `HIVE_AGENT_TOKEN_CACHE` is set per agent by the manager (`v2/pkg/agent/manager.go:6209`). The token was applied whether or not the agent did anything.

**3. It could break the session.** `bin/agent-launch.sh:25-26` deliberately leaves `GH_TOKEN` unset:

> Do NOT export GH_TOKEN here — Copilot CLI uses GH_TOKEN for its own Copilot API auth, which rejects GitHub App server-to-server tokens.

So a Copilot-backed agent that followed this instruction could lose **model** auth. The prompt's very next bullet already said *"the Copilot CLI owns that variable"* — contradicting the instruction four lines above it.

## What changed

The block now states the truth for both tool classes — git via the credential helper, gh via the wrapper, nothing to set up — and names the traps **without naming any path**.

That last part is deliberate, and it's a correction to my own first draft. I initially kept the shared cache in the prohibition ("do NOT read … not the shared `gh-app-token.cache`"), and the new guard rejected it. It's right to reject: #1861's *"Why now"* is an agent **discovering** that path under pressure when its push failed, so a prohibition that cites the path still hands over the map. The prohibition works fine generically — *"There is no token file you are meant to open"* — plus a positive instruction for the failure case: *"If gh reports an auth problem, report it — do not go find a token."*

`ghAuthInstructions` also loses its `agentName` parameter. Nothing in the block is per-agent any more, and that absence is the point.

## Guards

`v2/pkg/scheduler/gh_auth_instructions_test.go`, mirroring `bin/test_gh_auth_native_no_cat.sh` on the native side:

- **No token path of any kind** — plus a regex for `$(cat`, `cat /var/run` and `export GH_TOKEN=`, so a *reworded* reintroduction is caught, not just a literal revert.
- **The gh guidance must still exist.** "No token handling" must not decay into "no guidance" — an agent with no instruction improvises its own auth, which is exactly how the shared-cache read was found in the field.
- **The prompt must not set `GH_TOKEN`** while also telling the agent Copilot owns it.
- **Positive control:** an emptied block fails loudly rather than passing vacuously.

### Regression replay

| # | Neutered | Caught by |
|---|---|---|
| 1 | old `export GH_TOKEN=$(cat …)` restored | `NeverHandsTheAgentAToken` (3 assertions) + `DoesNotContradictItselfOnGHTOKEN` |
| 2 | reworded as `GH_TOKEN=$(cat /opt/creds/tok)` | the `$(cat` regex |
| 3 | gh bullet silently dropped | `StillTellsTheAgentGHIsAuthenticated` |
| 4 | block emptied | positive control in the helper |

Each confirmed failing, then restored green.

## Scope

`pkg/scheduler` passes. I swept the tree for the same pattern: the only other hits are `v2/deploy/entrypoint.sh:691` (root-side *creation* of the token file — correct) and my own comment recording what was removed. No docs asserted the old flow.

`pkg/agent`'s `TestRestartWithBootstrap` fails identically on clean `upstream/v4` in my environment (tmux socket path length from a deep workspace path) — pre-existing, and `pkg/agent` is untouched here.

No proxy-side auth injection, no `Authorization` stripping/rewriting, no minting layer. Item 1 still needs attested identity first (#3841/N7, and the #3833 Option D keypair) exactly as triaged — this neither implements nor presumes it. It just removes the reason an agent had to touch a token on the path where the hive already authenticates it.

Related open on this issue's actionable slice: #3888 (N7), #3889 (native prompt), #3926 (relay cache path). All three re-verified against current `v4` today while working on this — they still build and pass.

### DIFF
--- a/v2/pkg/scheduler/gh_auth_instructions_test.go
+++ b/v2/pkg/scheduler/gh_auth_instructions_test.go
@@ -1,0 +1,113 @@
+package scheduler
+
+import (
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/kubestellar/hive/v2/pkg/config"
+)
+
+// kubestellar/hive#1861 asks that agents hold no GitHub token material at all.
+// The proxy-side injection half of that issue is deferred (it needs attested
+// identity first — #3841/N7, and the #3833 keypair), but the container path
+// already reaches the goal without it: the image installs bin/gh-wrapper.sh AS
+// `gh` (v2/Dockerfile: COPY bin/gh-wrapper.sh /usr/local/bin/gh) and the wrapper
+// exports the agent's tier-scoped token from HIVE_AGENT_TOKEN_CACHE itself,
+// before the agent's command runs.
+//
+// So the kick prompt has no reason to hand the agent a token, and every reason
+// not to. These pin that it does not — the container-path twin of the
+// native-path guard in bin/test_gh_auth_native_no_cat.sh (#3842/#3889).
+
+func ghAuthBlock(t *testing.T) string {
+	t.Helper()
+	s := New(&config.Config{
+		Project: config.ProjectConfig{Org: "org", Repos: []string{"r"}},
+	}, testLogger())
+	block := s.ghAuthInstructions()
+	// Positive control: we are looking at the real authentication block, so a
+	// silent rename/emptying cannot make the assertions below vacuously pass.
+	if !strings.Contains(block, "## Project Authentication") {
+		t.Fatalf("ghAuthInstructions no longer returns the authentication block; got %q", block)
+	}
+	return block
+}
+
+// The load-bearing one. A token on a command line the agent types is a token in
+// the agent's transcript and in whatever it echoes back — one prompt injection
+// from exfiltration. This is the same exposure #3889 removed from the
+// native-install prompt; it differs only in blast radius (tier-scoped here,
+// fleet-wide there), not in kind.
+func TestGHAuthInstructions_NeverHandsTheAgentAToken(t *testing.T) {
+	block := ghAuthBlock(t)
+
+	// No token cache path of any kind, per-agent or shared — not even inside a
+	// prohibition. Naming the shared cache in order to forbid it still tells an
+	// agent where it lives, and #1861's "Why now" is precisely an agent
+	// DISCOVERING that path under pressure when its push failed. The prohibition
+	// below works without a map to the thing being prohibited.
+	for _, forbidden := range []string{
+		"/var/run/hive-metrics/agent-tokens",
+		"gh-app-token.cache",
+		"gh-token-",
+		"HIVE_AGENT_TOKEN_CACHE",
+	} {
+		if strings.Contains(block, forbidden) {
+			t.Errorf("the kick prompt names a token path (%q). Agents must not be told where credentials live "+
+				"(kubestellar/hive#1861): gh-wrapper.sh already applies the scoped token per invocation, so this "+
+				"instruction is redundant AND puts a live credential in the agent's transcript.", forbidden)
+		}
+	}
+
+	// No instruction to read a secret, whatever the path. Catches a reworded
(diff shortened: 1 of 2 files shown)


clubanderson (approved):
Reviewed against the #1861 goal (agents hold no token material) and the container auth path.

**Verified the replacement path actually authenticates gh**: `v2/Dockerfile:348` installs `bin/gh-wrapper.sh` AS `gh`, and the wrapper exports the tier-scoped token from `HIVE_AGENT_TOKEN_CACHE` per invocation with a deliberate fail-loud (no fallback to the shared cache — audit H3 comment in the wrapper). `manager.go:6209` sets that env per agent. So the removed instruction was genuinely redundant; nothing that legitimately uses gh loses auth.

**Verified all call sites**: the five `ghAuthInstructions(name)` callers in scheduler.go are all updated in this diff; no other caller exists in the tree. Dropping the `agentName` parameter is correct — nothing in the block is per-agent anymore, which is the point.

**Test quality**: the guard suite does the right things — positive control on the block header so an emptied/renamed block fails loudly instead of vacuously passing, a regex (`$(cat`, `cat /var/run`, `export GH_TOKEN=`) that catches reworded reintroductions rather than only a literal revert, and the not-naming-any-path decision (even in a prohibition) matches how #1861's original leak happened. The Copilot GH_TOKEN contradiction fix is a real correctness win on top of the security one.

Correctly scoped: no `Fixes #1861` since proxy-side injection (item 1) stays deferred pending N7 attested identity. LGTM — merging on green.

clubanderson:
Reviewed against the enforcement chain rather than just the prompt text. The claim the new block makes is true: `v2/Dockerfile` installs `bin/gh-wrapper.sh` as `/usr/local/bin/gh`, `manager.go` sets `HIVE_AGENT_TOKEN_CACHE` per agent, and the wrapper exports the tier-scoped token per invocation (failing loud, never falling back to the shared cache) — so removing the "export GH_TOKEN=$(cat …)" instruction costs legitimate consumers nothing while taking a live credential out of every agent transcript. The old text was also self-contradicting (told the agent to export GH_TOKEN four lines above saying the Copilot CLI owns that variable), and that Copilot-auth breakage risk is real. The tripwire test is well-aimed: forbidding the token path, `cat`-of-/var/run, and `export GH_TOKEN=` in the returned block pins the invariant at the source that generates every kick prompt, and dropping the now-unused agentName parameter means there is no per-agent path left to leak. Merging.

Issue #3848, opened 2026-08-14
[sec] complete the CSP nonce refactor for script-src, and scope style-src separately


clubanderson:
Follow-up to #3315, fixed in #3844 (`b86ed892`).

#3844 closed the injection vector (`DASHBOARD_TOKEN` no longer interpolated into an inline script) and **staged** the `script-src` tightening rather than completing it, because removing `unsafe-inline` wholesale would have broken the server-rendered dashboard.

### 1. Complete the `script-src` nonce refactor
When this is done, **`TestCSPScriptSrcUnsafeInlineIsStaged` will fail — by design.** It is an intentional tripwire pinning the staged state so the interim cannot be mistaken for finished. Do not "fix" it by relaxing the assertion: **invert it** to assert `unsafe-inline` is *absent* from `script-src`, so the completed state is then pinned in turn.

That inversion is the whole point of the tripwire, and this repo has a documented history (fifteen instances) of tests that pass for the wrong reason or encode the very behaviour they should forbid.

### 2. `style-src 'unsafe-inline'` — separate scope
Roughly **2,055 inline `style=` attributes** exist. Nonces do not cover attribute-level styles, so this is not solved by the same refactor and needs its own approach (external stylesheets, or accepting `style-src 'unsafe-inline'` with written rationale).

Deliberately excluded from #3844 rather than blocking the `script-src` work, which is the half with the actual XSS consequence.

## Context
The dashboard is **server-rendered Go**, not a static SPA — the contribute UI lives in `contribute_sse.go` and `api_contribute.go`, not `static/index.html`. Inline handlers and injected values are spread across Go template output, which is why the wholesale removal was unsafe and the refactor needs a per-request nonce threaded through the rendering path.

clubanderson:
Part 1 status — the ELEMENT half landed in PR #4009, the ATTRIBUTE half is now the whole remaining scope.

Shipped (mirrors ADR-0015's decomposition, recorded as ADR-0016):

```
script-src      'self' 'unsafe-inline'   ← CSP2 fallback, unchanged and deliberately HASH-FREE
script-src-elem 'self' 'sha256-…' …      ← inline <script> elements: CLOSED, no 'unsafe-inline'
script-src-attr 'unsafe-inline'          ← on*= attributes: STAGED, this issue's remaining scope
```

[...]

PR #3933, opened 2026-08-16
🔒 sec(dashboard): scope style-src into its two halves, with written rationale (#3848 part 2)

Danathar:
## What

Closes **part 2** of #3848 — `style-src` is scoped into its two halves, with the written rationale the issue asked for.

**Part 1 (`script-src`) is not attempted here, and the tripwire stays armed.** I did measure it, and the measurement corrects the issue's own estimate — details at the bottom, because I think it changes how the work should be planned.

Refs #3848 — no `Fixes`.

## Part 2 — the asymmetry that decides it

One blanket `style-src 'self' 'unsafe-inline'` was covering two things with different futures. Measured on `v4`:

| form | count | nonce or hash possible? |
|---|---|---|
| `style="…"` attributes | **2061** | **no** |
| `<style>` elements | **7** | yes |

CSP's nonce and hash mechanisms apply to *elements*. There is no nonce form and no hash form for an attribute-level style — `style-src-attr` accepts only `'none'`, `'unsafe-inline'`, or `'unsafe-hashes'` plus hashes, and `'unsafe-hashes'` would mean enumerating a hash per distinct attribute value and regenerating them on every UI edit. So for the 2061 attributes there is no policy that both permits them and constrains them. The real choice is `'unsafe-inline'` or deleting every inline style in the UI.

The header now says that out loud:

```
style-src      'self' 'unsafe-inline'   ← CSP2 fallback, unchanged
style-src-elem 'self' 'unsafe-inline'   ← the 7 elements: CLOSABLE, staged
style-src-attr 'unsafe-inline'          ← the 2061 attributes: ACCEPTED
```

**ADR-0015** records the rationale and the residual risk: styling injected markup enables UI redressing and some CSS-based inference of page content; it does **not** enable script execution.

**Behaviour is unchanged today, deliberately.** Browsers without CSP3 fall back to `style-src`, and the two CSP3 halves permit exactly what it permits — so no browser enforces anything different. The value is structural: the permanent allowance is now distinguishable from the temporary one, and closing the element half becomes a one-directive change with a test already watching it.

## Why I staged the element half instead of closing it

Seven deterministic `<style>` blocks *could* be hash-pinned now. It would buy nothing: while `script-src 'unsafe-inline'` stands, anyone who can inject a `<style>` element can inject a `<script>` element, which is strictly the more capable primitive. Hardening the weaker one first is motion, not protection.

So it is sequenced **after** `script-src`, where it starts to matter, and `TestCSPStyleSrcElemUnsafeInlineIsStaged` is its tripwire — written in the same shape as the one #3844 established, including the instruction to **invert, not relax** it.

## Part 1 — a correction to the estimate

The issue and the existing tripwire comment both say "~400 inline `on*=` handlers". Measured:

- **437** inline `on*=` attributes in `static/index.html` (299 `onclick`, 97 `onchange`, 10 `oninput`, plus a tail), **and**
- **~145 more built as strings inside the page's own JavaScript**, injected through **169** `innerHTML`/`insertAdjacentHTML` sites.

That second group changes the *shape* of the work, not just its size. A nonce covers `<script>` elements and never covers event-handler attributes, so completing `script-src` is an event-delegation refactor of a 21k-line SPA — not the nonce-threading exercise the issue describes.

There is also a collision worth flagging with **#3863**: `static/index.html` is now served as a precomputed-at-startup gzipped, ETag'd immutable document, and a **per-request nonce is fundamentally incompatible with that**. Closing `script-src` on this page has to reach for hashes or delegation rather than nonces, or it undoes the 4× transfer win.

I left the tripwire untouched and put these measurements in the comment beside the CSP, so whoever picks up part 1 starts from the real shape.

## Tests

5 new in `v2/pkg/dashboard/csp_style_src_test.go`: the split exists; the attr half is accepted and must never grow `'unsafe-hashes'`; the elem tripwire; behaviour-neutrality across CSP2 and CSP3 browsers; and that the ADR exists and is indexed — the written rationale *is* the deliverable for the accepted half, so deleting it has to fail.

### Regression replay

| # | Neutered | Caught by |
|---|---|---|
| 1 | blanket `style-src` restored | 3 tests fail |
| 2 | elem half closed without inverting | the tripwire, with its instructions |
| 3 | `'unsafe-hashes'` added to attr | `AttrIsAcceptedNotStaged` |
| 4 | ADR deleted, allowance kept | `RationaleIsWritten` |

Each confirmed failing, then restored green. `pkg/dashboard` passes.

### DIFF
--- a/v2/docs/adr/0015-csp-style-src-scope.md
+++ b/v2/docs/adr/0015-csp-style-src-scope.md
@@ -1,0 +1,100 @@
+# ADR-0015: Scope `style-src` as two directives and accept inline style attributes
+
+Status: Accepted
+
+## Context
+
+The dashboard's Content-Security-Policy carried one blanket `style-src 'self'
+'unsafe-inline'` ([securityHeaders](../../pkg/dashboard/server.go)). That single
+token covered two different things with two different futures, and hid the fact
+that only one of them is closable.
+
+[kubestellar/hive#3848](https://github.com/kubestellar/hive/issues/3848) asked
+for the two to be separated and for the remaining allowance to be either removed
+or accepted **with a written rationale**. This is that rationale.
+
+Measured on `v4` at the time of writing:
+
+| Inline style form | Count | Covered by a nonce or hash? |
+| --- | --- | --- |
+| `style="…"` attributes | 2061 | **No** |
+| `<style>` elements | 7 | Yes |
+
+The asymmetry is the whole decision. CSP's nonce and hash mechanisms apply to
+*elements* — `<style>` and `<script>`. There is no nonce form and no hash form
+for an attribute-level style. `style-src-attr` accepts only `'none'`,
+`'unsafe-inline'`, or `'unsafe-hashes'` plus hashes; and `'unsafe-hashes'` would
+mean enumerating a hash for each of 2061 distinct attribute values and
+regenerating them on every UI edit, which is neither reviewable nor
+maintainable. So for the attributes there is no policy that both permits them
+and constrains them: the realistic choice is `'unsafe-inline'` or deleting every
+inline style in the UI.
+
+A second constraint shapes the timing. `script-src` still carries
+`'unsafe-inline'` — that is the other half of #3848 and is staged behind an
+event-delegation refactor (437 inline `on*=` attributes in `static/index.html`,
+plus ~145 more built as strings inside the page's own JavaScript and injected
+through 169 `innerHTML`/`insertAdjacentHTML` sites; a nonce cannot cover any of
+them). **While inline script is permitted, tightening inline style buys no real
+security**: an attacker who can inject a `<style>` element into a page can
+inject a `<script>` element into the same page, and the script is strictly the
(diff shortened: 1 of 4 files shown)


clubanderson:
Checked what the split actually changes: all three directives (`style-src` fallback, `style-src-elem`, `style-src-attr`) retain `'unsafe-inline'`, so the effective policy is identical on every browser, CSP2 or CSP3, and the dashboard's 2000+ inline style attributes keep rendering — the split names the decision without moving enforcement. The ADR gets the mechanics right: nonces and hashes are element-level, there is genuinely no constraining policy for style attributes short of deleting them, and sequencing the closable `<style>`-element half behind the `script-src` work is correct — hardening inline style while inline script is still allowed is motion, not protection, since script injection subsumes everything style injection permits. The residual (UI redressing, CSS-based inference) is stated rather than waved off. Tests pin both the accepted half and the staged half with instructions to invert on closure, mirroring the #3844 tripwire pattern. Merging.

PR #3934, opened 2026-08-16
🔒 sec(compose): put watchtower behind a filtered Docker socket proxy (#3865)

Danathar:
## What

Watchtower no longer holds the Docker socket. A single `docker-socket-proxy` holds it and exposes a deny-by-default subset of the API over an internal-only network; Watchtower reaches the daemon only through that.

Refs #3865 — **no `Fixes`**, deliberately. See "What this does not fix".

## The sharpest edge was not the one in the finding

The issue frames the risk as privileged-container escalation. The more direct path is `POST /containers/{id}/exec`: with the raw socket, a compromised Watchtower reads the GitHub App private key straight out of the hive container's `/secrets`. No escalation, no privileged container, nothing unusual at the host level. That path is now **blocked**.

## What landed

Removed from Watchtower's reach: `/exec`, `/volumes`, `/networks`, `/secrets`, `/configs`, `/swarm`, `/nodes`, `/services`, `/tasks`, `/plugins`, `/build`, `/commit`, `/info`, `/system`, `/auth`, `/distribution`, `/session`. Each is set to `0` **explicitly** rather than left to the image's defaults, so a future image whose defaults change can't silently widen this — and so a reviewer sees the boundary instead of inferring it.

**Network isolation is load-bearing, and is the part that would have been easiest to get wrong.** The proxy speaks an *unauthenticated* Docker API on 2375. Putting it on the default Compose network would have handed that API to the `hive` container — which runs semi-trusted agent code — leaving the deployment worse off than the finding reports. So it sits alone on an `internal: true` network that only Watchtower joins:

```
docker-socket-proxy    networks=['docker-proxy']            # internal, unpublished
hive                   networks=['default']                 # cannot route to it
gateway                networks=['default']
watchtower             networks=['default', 'docker-proxy']
```

(That's the resolved `docker compose config` output, not the source text.)

`tecnativa/docker-socket-proxy` is pinned by digest like everything else here — `0.3.0` → `sha256:9e4b9e75…`, verified against the registry as an OCI image index covering `linux/amd64` and `linux/arm64`, so both arches resolve.

## What this does NOT fix

I'd rather say this than let the finding read as closed.

Watchtower cannot do its job without `CONTAINERS=1` and `POST=1`, and with those two a compromised Watchtower can still:

1. **`POST /containers/create` a privileged container binding `/`.** The proxy filters paths and methods, never request bodies — it cannot see `HostConfig.Privileged` or `HostConfig.Binds`. That is host root.
2. **`GET /containers/{id}/archive` to copy `/secrets` out of the hive container.** Reads are permitted for enabled sections; no escalation needed.

So this removes the easiest path and a large amount of unrelated surface, and it takes the Docker API away from agent code. It does **not** make a compromised Watchtower harmless. If Watchtower is compromised, the host and every credential on it should be treated as compromised.

## Documentation (recommendations 2 and 3)

The profile had **no documentation at all** — that absence is part of the finding. `v2/docs/auto-update-profile.md` is new and leads with the residual above, plus:

- **`:ro` on a `docker.sock` mount is not a mitigation.** The flag applies to the inode; the Docker API is a protocol spoken over the socket. Written up as a trap specifically so nobody "hardens" Watchtower that way and stops there.
- **Kubernetes should not use this profile at all** — a rolling Deployment update needs no socket, and the whole class of finding disappears.
- The honest first recommendation stays: don't enable auto-update on a host holding production credentials.

## Guard

`v2/deploy/test_watchtower_socket_contract.sh` — 21 assertions, wired into `v2-ci` beside the supply-chain pin guard.

It renders the compose file with `docker compose config` when the CLI is present, so assertions run against the **resolved** topology (implicit `default` membership included) rather than text that merely looks right, and falls back to a YAML parse otherwise.

### Regression replay

| # | Neutered | Caught by |
|---|---|---|
| 1 | socket bind re-added to watchtower | `does NOT bind the Docker socket` + `only service mounting the socket` |
| 2 | proxy publishes 2375 | `publishes NO ports` |
| 3 | `EXEC=1` | `no dangerous Docker API section is enabled` |
| 4 | proxy on the default network | `hive cannot reach the Docker API` + `off the default network` |
| 5 | network made non-internal | `network 'docker-proxy' is internal` |
| 6 | proxy digest dropped | `pinned by @sha256 digest` |
| 7 | `DOCKER_HOST` removed (silent fallback to the local socket) | `targets the proxy via DOCKER_HOST` |
| 8 | residual-risk doc deleted | `auto-update-profile.md exists` |

Each confirmed failing, then restored green. #7 is the one worth calling out: without it, deleting the bind mount looks like a fix while Watchtower quietly reopens `/var/run/docker.sock` on its own.

All four deploy contract tests pass; compose renders with and without the profile.

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -63,6 +63,17 @@ jobs:
       - name: Supply-chain pin guard
         run: bash deploy/test_supply_chain_pins.sh
 
+      # #3865: watchtower bind-mounted /var/run/docker.sock — the full daemon
+      # API, including /exec, which reads the GitHub App private key out of the
+      # hive container with no escalation at all. The socket now sits behind a
+      # deny-by-default docker-socket-proxy on an internal-only network. Every
+      # property of that containment is one plausible edit from reverting (a
+      # bind mount added "to debug", a published 2375, EXEC=1 to unblock some
+      # other tool, the proxy attached to the network the hive runs on) and none
+      # of those break anything at runtime — so they need a gate, not a review.
+      - name: Docker-socket containment guard (#3865)
+        run: bash deploy/test_watchtower_socket_contract.sh
+
       # bin/gh-wrapper.sh is the merge gate standing behind the proxy hard-deny:
       # if its argument parsing is wrong, a prompt-injected agent merges
       # unvetted code. It had NO tests and matched no CI path filter. This
--- a/v2/deploy/test_watchtower_socket_contract.sh
+++ b/v2/deploy/test_watchtower_socket_contract.sh
@@ -1,0 +1,247 @@
+#!/usr/bin/env bash
+# Asserts the Docker-socket containment contract for the auto-update profile.
+#
+# kubestellar/hive#3865: watchtower bind-mounted /var/run/docker.sock, which is
+# the full daemon API — including /exec, which reads the GitHub App private key
+# out of the hive container with no escalation at all. The socket now sits
+# behind a deny-by-default docker-socket-proxy on an internal-only network.
+#
+# This tests the INVARIANT, not behaviour. Every property below is one edit away
+# from silently reverting: re-adding the bind mount "to debug something",
+# publishing 2375, adding EXEC=1 to make some other tool work, or attaching the
+# proxy to the default network where the hive container runs semi-trusted agent
+# code. None of those break anything at runtime, which is exactly why they need
+# a guard rather than a code review.
+#
+# What this canNOT assert, and the docs say so plainly: the proxy does not make
+# auto-update safe. CONTAINERS+POST is host-root-equivalent on its own
+# (POST /containers/create with HostConfig.Privileged, and
+# GET /containers/{id}/archive to read /secrets). See
+# v2/docs/auto-update-profile.md.
+#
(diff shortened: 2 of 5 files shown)


clubanderson:
Reviewed the compose topology, the allowlist, and the contract test. The allowlist covers what watchtower actually needs and nothing discretionary: CONTAINERS/IMAGES for list-inspect-pull, POST for create/start/stop/rename, DELETE for `WATCHTOWER_CLEANUP`, PING/VERSION for the client handshake — and since the only labeled service (`hive`) sits on the single default network, no `NETWORKS=1` is needed for re-attach. EXEC=0 plus the internal-only network kills both the exec path and API reachability from the hive container, and writing out the deny flags explicitly guards against upstream default drift. What earns the merge beyond the mechanics is the honesty: the doc states plainly that CONTAINERS+POST is still create-privileged-container territory and that `GET /containers/{id}/archive` still reads `/secrets`, that `:ro` on a socket is not a control, and that the right production posture is not enabling the profile at all. The CI contract test asserting the proxy is the sole socket holder, no published ports, and no shared network with the hive — via structural YAML parse, not grep — is exactly the gate this class of one-edit revert needs. Merging.

PR #3938, opened 2026-08-16
🐛 fix(ci): point the coverage gate at v4, and make its issue say what it measured (#3903)

Danathar:
## What

**#3903 cannot be closed as filed, and finding out why turned up the real problem.** This fixes that; it does not add tests.

Refs #3903 — **no `Fixes`**, see the end.

## The run that filed #3903 did not fail on coverage

Its log says:

```
Total coverage: 88.6%
COVERAGE GATE FAILED:
hub: TESTS FAILING
```

No package was below its floor. The cause was one failing test — `TestHandleCreateHiveRaceIsolation`, in `provision_race_isolation_test.go`, which **exists only on v2**. So the issue is titled "packages below 90% threshold" for a broken test, and cites commit `d89121f5` (v4's head, from `github.sha`) for a run that measured **v2**. Anyone starting from that SHA was reading the wrong tree.

## The gate was watching a sunset branch

The hourly run pinned `ref: v2`, and the `pull_request` trigger was `branches: [v2]`. v2 is no longer a merge target — so the gate blocked on a branch nobody can PR to, and the **pre-merge half stopped covering every PR that actually lands**, which is the exact protection it was added for (#2350).

The consequence is bigger than stale numbers. v4 carries **14 packages that do not exist on v2 at all** — `agentparse`, `escalation`, `forge`, `intent`, `ioscan`, `mint`, `outputschema`, `planning`, `pushbroker`, `retro`, `review`, `sandbox`, `timeline`, `tracing` — so none of them has **ever** been subject to any floor.

Measured on v4 @ `d89121f5`, 14 packages sit below the 90% default:

| | | | |
|---|---|---|---|
| intent 65.5 | sandbox 67.6 | ioscan 71.0 | pushbroker 72.8 |
| tracing 76.4 | review 80.8 | retro 82.4 | hub 87.6 |
| github 88.2 | hivectl 88.7 | knowledge 89.4 | scheduler 89.6 |
| escalation 89.7 | proxy 89.7 | | |

**None of that is a regression that slipped past the gate.** The gate was pointed somewhere else.

## What changed

- `schedule`/`workflow_dispatch` checks out **v4**; `pull_request` adds v4. `v2/**` is the Go *module* directory on both branches, so the path filter is unchanged.
- **Per-package floors recorded** for those packages at their measured v4 value, rounded down to the integer the check compares. This is a **ratchet, not a target**, and it says so in the file. Switching to v4 with the bare 90% default turns CI red for 14 packages at once and blocks work unrelated to coverage — which is how a gate gets disabled rather than satisfied. `hub`'s existing floor moves 89 → 87 for the same reason: 89 was set against v2's 89.9%, and v4 measures 87.6%.
- The filed issue now reports the commit **and ref the job measured** (read from the working tree) instead of `github.sha`, and names the failing packages.
- It distinguishes the gate's two failure modes. *Cannot be scored* (tests failing / no test files) needs the tests fixed; *below floor* needs coverage. Conflating them is how #2191 collected 41 comments about coverage when the cause was a broken test — and how #3903 got its misleading title.

## Verification

I extracted the gate script from the workflow and **executed it** against the real v4 measurements rather than reasoning about it:

| | Scenario | Result |
|---|---|---|
| A | real v4 numbers | **green** — no package below floor |
| B | ioscan 71.0 → 70.9 | fails: `ioscan: 70.9% (floor 71%)` |
| C | new package with no test files | fails: `brandnew: NO TEST FILES` |
| D | package exactly at its floor | green |

So the ratchet bites on a one-tenth slip, and an untested package still cannot merge — the two properties worth having. I also confirmed `failed_packages` actually reaches the issue body via the new job outputs.

Workflow YAML parses; the gate script passes `bash -n`.

## Why no `Fixes`

This makes v4 measurable and stops the misleading report, but **all 13 recorded floors are below 90** and each needs real tests to close. Those are separate, per-package changes, and I would rather land the measurement fix than bundle a large test-writing effort behind it.

Two things that are yours to decide:

1. **The floor values.** I chose "record reality, ratchet upward" over "block until fixed". If you would rather the gate go red now and force the work, that is a one-line change per entry — the numbers above are what it would be enforcing.
2. **v2's failing test.** `TestHandleCreateHiveRaceIsolation` still fails on v2 and this PR stops the gate from looking at it. If v2 is genuinely done, that is the right outcome; if it still needs a green signal, it needs its own fix on that branch.

### DIFF
--- a/.github/workflows/coverage-hourly.yml
+++ b/.github/workflows/coverage-hourly.yml
@@ -7,8 +7,13 @@ on:
   # The floor is only meaningful if it runs BEFORE code lands. Without this the
   # gate was hourly-only, so a PR could drop a package below its floor and merge
   # green — which is exactly what happened in #2350.
+  #
+  # v4 added here (#3903): development moved to v4 and this filter still said
+  # v2, so the pre-merge half of the gate had silently stopped covering every PR
+  # that actually lands. `v2/**` is the Go module directory on BOTH branches —
+  # it is the module path, not the branch name — so the path filter is unchanged.
   pull_request:
-    branches: [v2]
+    branches: [v2, v4]
     paths: ['v2/**']
 
 permissions:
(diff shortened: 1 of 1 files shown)


clubanderson:
Verified the two things that could have made this dangerous before reading anything else:

- **It does not make coverage a blocking gate.** Required checks live in branch protection, which this PR cannot and does not touch — the only required context on `v4` is `gate`, and adding the `pull_request: v4` trigger just makes the signal exist. Good.
- **"Ratchet, not target" is the right call.** Recording measured floors instead of switching to a bare 90% default is the difference between a gate that catches slippage tomorrow and one that turns 14 packages red at once and gets disabled. The verification matrix (one-tenth slip fails, no-test-files fails, at-floor passes) is exactly what I would have asked for.

The measured-ref/measured-sha fix is the quiet win here — #3903 citing a v4 SHA for a v2 measurement is precisely the failure mode, and distinguishing "cannot be scored" from "below floor" should prevent another #2191-style comment pile.

One heads-up for the first scheduled v4 run: v4 has had a history of tests that fail only in Linux CI (#3162/#3163 — the beads setgid, sandbox podman, and hub enroll cases). The PR-triggered `v2 Tests` check is green on current v4-based PRs, so these look resolved, but if any recur the hourly gate will now file a "cannot be scored" issue — which, with the new body, will at least say the true cause. That is the gate working, not this PR breaking.

Agreed on both open questions: record-and-ratchet over day-one red, and letting the gate stop watching v2's `TestHandleCreateHiveRaceIsolation` is the correct outcome for a sunset branch.

Merging.

PR #3943, opened 2026-08-16
fix(github): recover from a stale rate-limit cache after installation-token rotation

hanthor:
## Symptom

Auto-merge stalls with a rate-limit error while the App has almost its entire allowance free:

```
WARN self-authored automerge sweep failed
  error="listing open PRs for tuna-os/tunaos: GET .../pulls: 403
         API rate limit of 6900 still exceeded until 2026-08-16 02:34:57 -0400 EDT,
         not making remote request."
```

…measured at the same moment on the same hive:

```
core: {"limit":6900,"remaining":6613,"reset":"…"}
```

**6613 of 6900 remaining, and zero requests consumed across six minutes.** The hive was not rate limited. It was refusing to ask.

MTTR collection and the queued auto-merge sweep fail as collateral, because the refusal is client-wide rather than per-caller.

## Cause

`checkRateLimitBeforeDo` short-circuits locally:

```go
if !rate.Reset.Time.IsZero() && rate.Remaining == 0 && time.Now().Before(rate.Reset.Time) {
    // Create a fake response.  →  synthetic 403, no request made
```

That cache is per-`Client`, and is only updated by responses the `Client` receives.

Hive's App client is constructed **once**:

```go
func NewClientFromAppWithBotLogin(...) *Client {
    transport := &appTransport{auth: auth, base: http.DefaultTransport}
    client := gh.NewClient(&http.Client{Transport: transport, ...})
```

…while `appTransport.RoundTrip` injects a **freshly minted installation token per request**, and installation tokens rotate roughly hourly.

So the two halves disagree: the credential is new and has a full allowance, but the client's cached verdict belongs to the *previous* token. go-github then refuses every call until the old token's reset time — potentially most of an hour of self-inflicted downtime, during which the real budget sits unused.

This is not a go-github bug. Its cache is correct for the usual one-client-one-credential shape; it simply cannot know the credential underneath was swapped.

## Fix

On a rate-limit error, re-read `GET /rate_limit`. `RateLimitService.Get` writes the response straight back into the client's cache:

```go
s.client.rateMu.Lock()
if response.Resources.Core != nil {
    s.client.rateLimits[CoreCategory] = *response.Resources.Core
}
```

So one call replaces the stale verdict with current truth, and the next tick is decided on that. `GET /rate_limit` **does not count against any limit**, so this is an exact correction rather than a guess or a backoff.

If the limit really is exhausted, the refreshed cache says so and the pre-emptive check keeps doing its job — the behaviour only changes when the cache was wrong.

Scoped to the self-authored sweep's error path: that is where the stall was observed, and its ticker makes a stuck cache most expensive. Worth extending to other long-lived callers if this proves out, but I would rather land the narrow, demonstrated case first.

## Tests

`rate_limit_cache_test.go`:

- **RecognisesPreEmptiveRefusal** — the synthetic local 403 is the shape that actually stalls, so it must match, including when wrapped in caller context (`listing open PRs for o/r: %w`). Also covers `AbuseRateLimitError`.
- **IgnoresOtherErrors** — the refresh costs an API call; it must not fire on 404/401/network errors.
- **NilSafe** — called from an error path, so it must not panic on a half-constructed client.

`go build ./...`, `go vet ./pkg/github/` and the package tests pass.

## Relationship to #3921

Independent, and both are needed. #3921 stops the sweep *causing* exhaustion (it was issuing 16,200 req/hour against a 6,900 limit by ticking every 10s per repo). This one stops the hive staying stuck after any exhaustion, whatever the cause. With only #3921 the fleet still froze for the remainder of a window whenever the cache went stale.

### DIFF
--- a/v2/pkg/github/automerge_sweep.go
+++ b/v2/pkg/github/automerge_sweep.go
@@ -303,6 +303,51 @@ func (c *Client) SweepSelfAuthoredAutoMerges(ctx context.Context, opts AutoMerge
 // started at all: an ACMM L4/L5 hive (l4.md/l5.md both forbid the App
 // merging its own PRs) must not self-merge, matching console's L6 hive which
 // is unaffected and keeps self-merging as before.
+// refreshRateLimitCache re-reads GitHub's real rate limits and writes them back
+// into the go-github client's cache.
+//
+// WHY THIS IS NEEDED. go-github refuses requests PRE-EMPTIVELY: once it has seen
+// Remaining==0 it returns a synthetic 403 ("not making remote request") for
+// every call until the cached Reset time passes, without contacting GitHub.
+// That cache lives on the Client and is only updated by responses the Client
+// itself receives.
+//
+// Hive's App client is created ONCE (NewClientFromApp) while appTransport
+// injects a freshly minted INSTALLATION TOKEN per request, and installation
+// tokens rotate roughly hourly. A new token gets a new allowance — but the
+// Client's cache still says Remaining==0 with the old token's reset, so
+// go-github keeps refusing requests the new token could happily serve.
+//
+// Observed live: the dashboard reported core remaining=6613 of 6900 while every
+// sweep tick failed with "API rate limit of 6900 still exceeded", and the fleet
+// consumed ZERO requests over six minutes — not rate-limited, just refusing.
+// Merges stalled for the remainder of the window each time.
+//
+// GET /rate_limit does not count against any limit, and RateLimitService.Get
+// writes the result back into the client's cache, so this is a cheap, exact
+// correction rather than a guess.
+func (c *Client) refreshRateLimitCache(ctx context.Context) {
+	if c == nil || c.client == nil {
+		return
+	}
+	if _, _, err := c.client.RateLimit.Get(ctx); err != nil {
+		c.warn("could not refresh rate-limit cache", "error", err)
+		return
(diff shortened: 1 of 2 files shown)


clubanderson (approved):
Reviewed the diff and the go-github behaviour it corrects. The diagnosis is sound: the per-Client rate cache and the per-request installation token rotate on different schedules, so the cached Remaining==0 verdict can outlive the credential it belonged to. The fix is well-contained — isRateLimited uses errors.As so both RateLimitError and AbuseRateLimitError match even when wrapped with caller context; the refresh fires only on rate-limit shapes (the test asserting 404/401/network errors do not trigger it is the right guard, since the refresh costs a request slot in spirit even if /rate_limit is free); refreshRateLimitCache is nil-safe for both nil receiver and half-constructed client, with tests. Scoping to the sweep's error path where the stall was observed, rather than a blanket hook, is a reasonable land-narrow-first call.

One cosmetic nit for a future pass, not blocking: the new functions were inserted between StartSelfAuthoredAutoMergeSweep's doc comment and the function itself, so that ACMM comment block now reads as the lead-in to refreshRateLimitCache. Worth reattaching next time someone touches the file.

Merging on green. Agreed with the analysis that this and #3921 are independent halves: #3921 prevents self-inflicted exhaustion, this one prevents staying stuck after any exhaustion.

PR #3947, opened 2026-08-16
🐛 fix(credentials): stop `--scoped` serving the full App token from the shared cache

Danathar:
## What

`bin/gh-app-token.sh` mints two very different things from the same GitHub App key:

- the **full** installation token, cached hub-only at `/var/run/hive-metrics/gh-app-token.cache`
- the short-lived **per-tier scoped** tokens handed to contributor agents (#1860)

Keeping those apart is the point of the scoped-token scheme in #1861: an agent must never hold installation-wide credentials.

The cache-validity check ran **before** the `--scoped` branch, so it short-circuited scoped requests too. With a warm cache:

```console
$ gh-app-token.sh --scoped newcomer
ghs_FULL_PRIVILEGE_TOKEN     # the unscoped installation token, exit 0
```

Tier validation, the permissions body and the scoped mint were all unreachable — `--scoped bogus` "succeeded" the same way. On a cold cache, `--scoped` still minted and cached a **full-privilege** token as a side effect of a request that only ever wanted a narrow one.

## Impact

On a running hub the cache is warm essentially always (`hive-config.sh` refreshes it on every load; entries live 55 minutes), so this was the normal path, not an edge case.

- **Native installs** (`bin/hive.sh` → `dashboard/server.js`) are the only caller of `--scoped`. `mintScopedToken` JSON-parses the output, so a raw `ghs_...` threw, returned `null`, and the contributor was silently dropped with `Failed to mint token` instead of being assigned work. Fail-closed, but per-agent scoped tokens were effectively dead there.
- Any caller that did **not** happen to parse the output would have received installation-wide credentials in place of a newcomer's issues-only token.
- **The containerized hub is unaffected**: `v2/pkg/dashboard/contribute_ws.go` mints via `github.AppAuth.ScopedToken` in Go and never shells out to this script.

## Change

Move `--scoped` ahead of the cache check and give it its own App JWT, so it neither reads nor writes the shared cache. The duplicated JWT construction folds into `mint_app_jwt()`.

Two smaller fixes to the same file, both in scope of #1861:

- **Create the cache `0600` via `umask`** instead of writing at the process umask and `chmod`-ing after. The old order left a window in which the full-privilege token was world-readable on a box that also runs agent UIDs. The `chmod` stays for caches created by earlier versions, since `>` truncates in place and preserves an existing mode. (Item 2 of #1861 was closed by `69256ecf`; this closes the remaining race.)
- **Stop the warm-cache path echoing the token to stdout.** #1757 removed that from the mint path but left it in the cache-hit branch; both in-tree callers (`hive-config.sh:199`, `entrypoint.sh:1037`) already discard stdout and read the cache file. `--export` — the one documented way to get the token out — now works on **both** paths; on a cold cache it used to print only the cache-location line, so the documented `eval "$(gh-app-token.sh --export)"` exported nothing on the first call after boot.

## Tests

`v2/pkg/config/gh_app_token_script_test.go` runs the **real script** with `curl` stubbed and the cache path rewritten into a temp root, following the existing `entrypoint_boot_test.go` precedent rather than paraphrasing the logic.

Every test fails against the current script and passes after this change:

| Test | Before |
|---|---|
| `TestScopedMintIgnoresTheSharedFullTokenCache` | got the cached full token |
| `TestScopedMintDoesNotPopulateTheSharedCache` | wrote a full-privilege cache |
| `TestScopedMintRequestsTierPermissions` (4 tiers) | never reached the mint |
| `TestScopedMintRejectsUnknownTier` | exited 0 with the full token |
| `TestScopedMintPassesRepositoryRestriction` | never reached the mint |
| `TestBareInvocationNeverPrintsTheToken` | warm cache printed the token |
| `TestExportEmitsShellExportOnBothPaths` | cold cache exported nothing |
| `TestSharedCacheIsCreatedPrivate` | existed at `0644` before the `chmod` |

The 0600-at-creation test observes the race directly: a `chmod` stub records the mode the file already had when the script got around to locking it down.

```console
$ cd v2 && go test ./pkg/config/
ok  	github.com/kubestellar/hive/v2/pkg/config	3.275s
```

The tier-permissions table is pinned because those permissions are the scoped-token scheme's actual security boundary.

## Scope

This is the item-2 half of #1861 (`--scoped` and cache hygiene), not item 1. **Proxy-side auth injection stays deferred** exactly as recorded in the triage: it needs N7 (#3888) closed and the #3833 Option D keypair landed, so injection mints against attested identity rather than a spoofable name.

Also checked while here, no change made: `dashboard/server.js:2620` interpolates `tier` into `execSync` unquoted, but every entry point validates it against an allowlist (`/api/contributors/:id/trust` and the auto-promote paths), and `revoked` is rejected before minting — not injectable.

Refs #1861. Companion to #3888 (N7), #3889 (`kick-agents.sh`), #3926 (relay cache path).

---

*Authored with Claude Code on **Opus 5 (high reasoning)**.*


clubanderson (approved):
Reviewed with focus on the token-scope boundary.

**Verified the core invariant**: the `--scoped` branch now runs before any cache consultation, mints its own App JWT via `mint_app_jwt()`, and neither reads nor writes `$CACHE_FILE`. `TestScopedMintIgnoresTheSharedFullTokenCache` is a genuine positive control — it seeds the shared cache with a sentinel full token and asserts the sentinel never appears in scoped output, so a regression that reintroduces the short-circuit will fail the test rather than pass vacuously. Pinning the per-tier permission tables (`TestScopedMintRequestsTierPermissions`) is exactly right, since those bodies are the actual boundary.

**Verified the stdout behavior change is safe**: cross-checked every in-tree caller — `hive-config.sh:199` and `entrypoint.sh:1036` both discard stdout and read the cache file; `dashboard/server.js` `mintScopedToken` only ever calls `--scoped` and JSON-parses. No caller depends on the warm-cache path echoing the raw token, and `--export` now working on the cold path fixes a real first-boot gap.

**umask-based 0600 creation**: correct fix for the world-readable window, and keeping the trailing `chmod` for pre-existing caches (since `>` preserves mode) is the right call. The `modeAtLock` observation via the chmod stub is a clever way to make the race assertable.

Also confirmed `testRSAKeyBits` resolves (defined in `appkey_test.go`). LGTM — merging on green.

clubanderson:
Reviewed the full diff with the trust boundary in mind. The invariant holds: `--scoped` is now handled before any cache read or write, mints its own App JWT, and exits non-zero on a failed mint rather than falling through — so the scoped path can no longer return the broader installation token under any ordering, warm cache or cold. The test suite is the right shape: seeding the cache with a sentinel and asserting it never appears in scoped output is a genuine positive control (it would have caught the original bug), the per-tier permission bodies are pinned as the actual security boundary, and the umask-before-write fix closes the create-then-chmod readable window with the chmod stub observing the pre-lock mode. I also checked the bare-invocation contract change against in-tree callers: `hive-config.sh` and `entrypoint.sh` both discard stdout and read the cache file, and `dashboard/server.js` only calls `--scoped`, so nothing depended on the warm-cache echo this removes. The `--export` cold-path fix (previously exported nothing on first boot) is a nice catch riding along. Merging.

PR #3948, opened 2026-08-16
✨ feature(contribute): declare the agent CLI version, and bound what a declaration may say

Danathar:
## What

The DECLARE half of #2547 shipped a **six**-field capability schema. The relay has only ever populated **five**.

`agent_cli_version` is in the hub struct (`ContributorCapabilities`), in `v2/docs/contributor-relay.md` — *"the relay self-reports … agent CLI version"* — and in the Operations row (`cli 1.2.3`, `capabilityLine` in `api_contribute.go:4276`). `detectCapabilities()` in `bin/contributor-relay.sh` never set it, so that column has been blank for every connected client and the docs describe something the shipped relay does not do.

It is also the axis this issue's own evidence names first. The three environment-shaped failures reported from the field are:

> a host without a working container runtime, a rootless Podman path that differs from the Docker path, **an agent CLI old enough to lack a flag the prompt assumes**

The first two are visible on the fleet view today. The third was the one an operator could not see.

Refs #2547 — **no `Fixes`.** This is DECLARE, not ROUTE. ROUTE stays open and untouched.

## Change

**Relay declares it.** Probes the resolved backend binary with `--version` — the same resolution the launch path uses, so `litellm` reports the `claude` CLI's version rather than nothing — and declares one short printable line of it.

**Best-effort throughout.** A missing binary, an unsupported flag, a wedged CLI, or unusable output all yield *no field at all*, which reads as unknown and is exactly what every relay written before this change reports. That matters beyond tidiness: the issue is explicit that no default may read silence as incapacity, so failing to probe has to stay a first-class outcome rather than a defect. stdin is closed so a CLI that mistakes `--version` for an interactive launch gets EOF instead of waiting on a terminal nobody is at; stderr is discarded so a warning banner cannot be declared as a version.

**Probes moved off the handshake.** They now run at startup, before the first connection, instead of lazily inside the `auth_challenge` handler. `detectCapabilities()` already shelled out for the container-runtime check, and this adds backend resolution plus a subprocess; the hub bounds a handshake at 30s (`wsAuthTimeout`) and none of that work belongs in that budget on a slow host.

**Both sides bound the result.** The relay reduces the CLI's output to one line — CLIs append update nudges and colour escapes. The hub then independently truncates every declared field to 64 runes and strips control characters *when it stores them*.

That second half is the part I'd flag for review, because it changes existing behaviour: a declaration is unverified client text held for the life of the connection, re-serialized into every fleet poll, and rendered into an operator row. Today 64KB of padding is accepted verbatim — the WS read limit is the only ceiling — and embedded newlines and escapes pass straight through to the ops surface. Trusting the client to have limited its own display string is not a bound. This is the issue's own worst-case in miniature:

> the self-reported nature has to be visible at the point of use

**Sanitizing cannot reject.** An over-long or messy declaration still authenticates and still receives work — losing a connection over a display string would be "silence means incapable" arrived at sideways. It runs *before* the `IsZero` check, so a whitespace-only declaration reads as "declared nothing" rather than as a set of empty-string capabilities that would render an empty `declares:` row.

## The line this holds

Nothing reads a capability to make a decision. Selection stays blind to declarations; `TestSelectionPathsDoNotReadDeclaredCapabilities` (the #3815 guard) passes untouched, and I did not need to modify it. The new field is one more thing an operator can *see*, which is the whole of DECLARE.

## Backward compatibility

- A relay that declares nothing is unchanged — `agent_cli_version` is `omitempty` like every other field.
- A relay that cannot probe its CLI declares five fields exactly as today.
- No wire change in either direction: no new message type, no new field on the envelope, no new required value.
- `projectbluefin/review` keeps connecting unchanged and picks this up whenever it takes a newer relay.

## Tests

**`bin/contributor-relay.test.js`** (+6, runs in `v2-ci.yml`): probe shape (argv, closed stdin, bounded timeout); binary resolution via a new `backendBinary` harness option that models `litellm → claude`; every failure mode omitting the field while the other five survive; banner/escape reduction; the field reaching `auth_response`; and probe-once-per-process caching.

**`v2/pkg/dashboard/contribute_capability_bounds_test.go`** (new): field sanitizing incl. rune-safe truncation, an honest declaration round-tripping byte-for-byte unchanged, `Sanitized` covering all six fields, and end-to-end over a real WebSocket that an abusive declaration is bounded **but still admitted**.

```console
$ node bin/contributor-relay.test.js
72/72 passed
$ cd v2 && go test ./pkg/dashboard/ -short -race -count=1
ok  	github.com/kubestellar/hive/v2/pkg/dashboard	51.466s
```

Both suites were checked against the pre-change code: the three relay tests that assert the declaration fail without it (`69/72`), and the two end-to-end hub tests fail without the sanitize call (`stored container_runtime is 4096 runes, want it bounded to 64`).

One local-run gotcha, pre-existing and not touched here: the relay harness sets `HIVE_REGISTRATION_TOKEN` but not `HIVE_HUB`, so running the suite on a machine that is itself a contributor picks up the ambient multi-hub `HIVE_HUB` and exits with a token-count FATAL before any test runs. `env -u HIVE_HUB -u HIVE_REGISTRATION_TOKEN node bin/contributor-relay.test.js` works; CI is unaffected.

## Relation to my other open PRs on this issue

Independent of both, though #3930 touches two of the same files in different hunks:

- #3901 — failure attribution (`failure_kind`)
- #3930 — peer protocol-compatibility detection, which is also where the relay's `RELAY_PROTOCOL_VERSION` 1.1→1.2 drift is fixed. Deliberately not touched here.

---

*Authored with Claude Code on **Opus 5 (max reasoning)**.*


clubanderson:
Reviewed specifically against the DECLARE/ROUTE line from #2547, since "CLI version" is the axis where declaration most tempts routing. This stays cleanly on the DECLARE side: the new field is populated, bounded, stored, and rendered — nothing consults it for selection or assignment, the #3815 guard (`TestSelectionPathsDoNotReadDeclaredCapabilities`) is untouched and still covers the paths, and keeping `Refs` rather than `Fixes` on #2547 is exactly right while ROUTE remains an open maintainer decision.

The part you flagged for review — hub-side `Sanitized()` — is the part I most wanted to see. Trusting the relay's self-truncation would not have been a bound, and the ordering details are correct: sanitize-before-`IsZero` so whitespace-only reads as "declared nothing," sanitize-cannot-reject so a messy declaration never costs admission (the end-to-end test asserting auth_ok *and* the 64-rune bound is the invariant that matters, not just the happy path). Probe hygiene is also right: startup-time so it never eats the 30s handshake budget, stdin closed, stderr discarded, every failure mode collapsing to field-omitted.

Two minor notes, neither blocking:
- ANSI color sequences lose only the ESC byte, so residue like `[32m2.0.1 [0m` can survive into the declared value — both test suites encode this deliberately, and it is cosmetic on a display-only field, but a follow-up CSI-aware strip would tidy the ops row for colourising CLIs.
- `sanitizeCapabilityField` dropping U+FFFD means a genuinely multibyte-mangled declaration silently shrinks; acceptable for a display token, just noting it is a lossy choice.

Merging.

Issue #3957, opened 2026-08-17
Bead synthesizer probes every repo per PR lookup, exhausting the App rate limit hourly


hanthor:
## Summary

`PREnricher.fetchPRFromRepos` probes **every configured repo** for a PR number. On a hive with many repos this turns one bead enrichment into N API calls, and the hourly synthesizer pass exhausts the whole GitHub App installation allowance — after which every other subsystem 403s until the window resets.

```go
// v2/pkg/knowledge/bead_synthesizer.go:866
func (e *PREnricher) fetchPRFromRepos(ctx context.Context, owner string, number int) *gh.PullRequest {
	for _, repo := range e.repos {
		pr := e.fetchPR(ctx, owner, repo, number)   // GET /repos/{owner}/{repo}/pulls/{n}
		if pr != nil {
			return pr
		}
	}
	return nil
}
```

A PR that exists costs on average N/2 calls; one that does **not** exist anywhere — a stale bead, the common case — costs the full N. On a 39-repo hive that is 39 requests to answer one lookup.

## Measured on a live hive (39 repos, limit 6,900/hour)

Instrumented `appTransport.RoundTrip` (the single chokepoint for every GitHub call from the Go client) counting calls per route and per calling frame.

During a synthesizer pass:

```
GET /repos/*/*/pulls/{n}   235/min      <- 60% of all traffic
total                      388/min  ->  23,280 requests/hour
```

**3.5× the 6,900/hour limit**, measured with **all nine agents paused** — this is hive's own polling alone.

Steady state, between passes, is fine:

```
total  28/min -> 1,680/hour
GET /repos/*/*/pulls/{n}  1/min   <- knowledge.(*PREnricher).fetchPR:849
```

235/min ≈ 6 lookups/min × 39 repos, which is what pointed at the fan-out.

## Why this is worse than the raw numbers suggest

The synthesizer runs on `schedule: hourly`, so the spend arrives as a **block**. Once the allowance is gone, go-github refuses everything client-wide, so the damage is not confined to the synthesizer:

- the self-authored auto-merge sweep stops (`403 API rate limit exceeded for installation ID …`)
- `failed to compute MTTR` every cycle
- the queued auto-merge sweep fails
- auto-merge is effectively down for the remainder of each hour

Observed as auto-merge silently not working, with the dashboard's rate-limit panel reporting a healthy figure (see the note below), which makes it look like nothing is wrong.

## Suggested fixes

[...]

PR #3960, opened 2026-08-17
fix(knowledge): negative-cache PR lookups — the repo-probe fan-out re-pays its full cost every cycle (#3957)

hanthor:
Implements remedy (3) from #3957 — the one that needs no bead-schema decision.

## Problem

`PREnricher.fetchPRFromRepos` probes **every configured repo** for a PR number: `len(repos)` API calls per lookup, and the full count when the PR exists nowhere — the common case for a stale bead. The positive cache only records hits, so every miss re-pays the entire fan-out on every synthesizer cycle, forever.

Measured on a live 39-repo hive (details in #3957): **235 `GET /repos/*/*/pulls/{n}` per minute** during a synthesizer pass — 60% of all GitHub API traffic — running the installation to 23,280 req/hour against a 6,900/hour limit. Once exhausted, go-github refuses client-wide, so auto-merge, MTTR and the queued sweep fail for the rest of each hour.

## Change

Cache the misses, at two levels:

- **per-(repo, number)** in `fetchPR` — on a **definitive 404 only**. A transient failure (rate limit, network) stays retryable, or one bad window would wrongly blacklist real PRs for the rest of the run.
- **per-number at the fan-out level** in `fetchPRFromRepos`, so repeated references to the same dead number skip the loop entirely rather than re-checking each repo's cached 404.

Both maps are bounded by the existing `prEnricherCacheSize` and clear on the existing `ClearCache` boundary, so a PR opened mid-cycle was legitimately absent when probed and becomes visible next cycle.

Net effect: a stale bead costs the fan-out **once per run** instead of once per bead per cycle indefinitely.

## Non-goal

This deliberately does not touch the bead schema. Carrying `repo` alongside the PR number (remedy 1 in #3957) remains the real fix — no probing at all — and this change does not preclude it; it just stops the bleeding in a way that needs no schema decision.

`go build ./...` and `go test ./pkg/knowledge/` pass.

### DIFF
--- a/v2/pkg/knowledge/bead_synthesizer.go
+++ b/v2/pkg/knowledge/bead_synthesizer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -42,9 +43,9 @@ type BeadSynthesizer struct {
 	lifecycleManager *BeadLifecycleManager
 	graphStore       *GraphStore
 
-	mu       sync.Mutex
-	cancel   context.CancelFunc
-	running  bool
+	mu      sync.Mutex
+	cancel  context.CancelFunc
+	running bool
 }
 
 // SetGraphStore attaches a graph store so synthesized facts emit relationship triples.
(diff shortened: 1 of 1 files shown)


clubanderson (approved):
Reviewed the full diff against the measurements in #3957. The two-level negative cache is correctly scoped: misses are recorded on a definitive 404 only, so transient failures (rate limit, network) stay retryable; both maps are bounded by the existing prEnricherCacheSize and cleared on the same ClearCache boundary, so a PR opened mid-cycle becomes visible next run; and all map access happens under the mutex, including the miss check added inside the locked section of fetchPR. The fan-out key format (owner/*#N) cannot collide with per-repo keys since * is not a valid repo name. This stops the measured bleeding (full fan-out once per run instead of per bead per cycle) without presuming the bead-schema decision, which remains the real fix as the PR notes.

One optional follow-up, not blocking: a small unit test pinning the invariant "404 is cached, transient error is not" would guard the retryability distinction against future edits.

Merging. Since the body implements remedy (3) of #3957 rather than closing it wholesale, I will close #3957 with a reference to this PR and the merge SHA.

PR #3970, opened 2026-08-17
fix(github): surface the App's own stale drafts — scanner's own step-2 instruction had nothing to act on

hanthor:
## The bug

`fetchPRs` drops **every** draft PR before the actionable list is built:

```go
if pr.GetDraft() {
    continue
}
```

But the scanner kick prompt (`buildScannerMessage`) already tells the agent, in its own workflow step 2:

> "Close stale drafts (>48h, needs-rebase + dco-no, or fix already merged)"

That instruction had nothing to act on. A draft the App opened and never returned to becomes invisible to every future kick, forever — it never reaches the actionable list a kick prompt is even built from. There's no mechanism anywhere that revisits it.

## Observed live

Two App-authored draft PRs (`feat(export): add Proxmox vzdump adapter`, `feat(move): support Incus as move destination`), **both fully green on CI**, sat untouched for a week. Not because anything was blocking them — CI passed cleanly on both — but because no agent was ever shown they existed after the cycle that opened them ended.

## Fix

Add a `StaleDrafts` channel to `PRResult`, populated only for:

- the **App's own** drafts (`appBotLogin` match) — a human's stale draft is their call, not hive's to nag about; hive has no standing to close or finish someone else's WIP
- **older than 48h** — the same threshold the prompt text already names, so a draft from the current cycle isn't flagged as "stale" before the agent has even had a chance to return to it

Neither this list nor a plain draft is ever added to `Items` (the normal actionable list) — a draft still cannot be merged and must never be treated as ready work. This only adds *visibility*, not eligibility.

`buildScannerMessage` renders the list as `YOUR STALE DRAFT PRs (N, >48h old — finish, mark ready, or close)`, giving the existing step-2 instruction a real list to act on.

## Tests

`stale_draft_test.go`:
- an App-authored draft >48h old is surfaced in `StaleDrafts` and absent from `Items`
- a **recent** App-authored draft (1h old) is not surfaced — still-in-progress this cycle
- a **human-authored** draft, even 500h old, is not surfaced — not hive's to touch

Existing `pkg/github` and `pkg/scheduler` suites pass unchanged, including `TestFetchPRs_DraftFiltered`, confirming ordinary drafts still stay out of the actionable list exactly as before.

### DIFF
--- a/v2/pkg/github/client.go
+++ b/v2/pkg/github/client.go
@@ -264,6 +264,15 @@ type IssueResult struct {
 type PRResult struct {
 	Count int           `json:"count"`
 	Items []PullRequest `json:"items"`
+	// StaleDrafts are the App's OWN draft PRs older than staleDraftAfter.
+	// Ordinary drafts (someone's in-progress work, human or agent) are
+	// excluded from Items entirely and stay that way here — nobody should be
+	// nagged about work they are actively doing. This list exists because the
+	// scanner kick prompt already instructs: "Close stale drafts (>48h,
+	// needs-rebase + dco-no, or fix already merged)" — but fetchPRs drops
+	// EVERY draft before the agent ever sees it, so that instruction had
+	// nothing to act on. See kubestellar/hive#3963.
+	StaleDrafts []PullRequest `json:"stale_drafts,omitempty"`
 }
 
 type HoldResult struct {
@@ -381,6 +390,7 @@ func (c *Client) EnumerateActionable(ctx context.Context) (*ActionableResult, er
 	var allIssues []Issue
 	var allPRs []PullRequest
 	var holdItems []HoldItem
+	var allStaleDrafts []PullRequest
 	totalByRepo := make(map[string]RepoCounts)
 
 	repos := c.getRepos()
@@ -397,7 +407,7 @@ func (c *Client) EnumerateActionable(ctx context.Context) (*ActionableResult, er
 		allIssues = append(allIssues, issues...)
 		holdItems = append(holdItems, held...)
 
-		prs, heldPRs, prTotal, err := c.fetchPRs(ctx, repo)
+		prs, heldPRs, staleDrafts, prTotal, err := c.fetchPRs(ctx, repo)
 		if err != nil {
 			// Issues for this repo were already collected; a PR-only failure
 			// is partial and must not count toward the all-repos-failed guard,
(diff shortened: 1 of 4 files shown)


clubanderson (approved):
Reviewed and approving.

**The fix itself.** Correct diagnosis: `fetchPRs` drops every draft (`if pr.GetDraft() { continue }`) before the actionable list is built, so the scanner kick's step-2 instruction ("Close stale drafts (>48h, needs-rebase + dco-no, or fix already merged)") has had literally nothing to act on. Surfacing them via a separate `StaleDrafts` field — rather than letting drafts into `Items` — is the right shape: drafts stay non-actionable/non-mergeable, and the scanner gets a dedicated nag list instead.

**Scoping.** The filter is properly scoped: `strings.EqualFold(author, c.appBotLogin)` restricts it to the App's own drafts, and `staleDraftAfter = 48h` matches the threshold the prompt text already names. Human drafts and recent App drafts are excluded, with tests for both negative cases plus the positive control. Note also that held/exempt-labeled drafts are filtered out *before* the draft branch is reached, so a hold label keeps a draft out of the stale list — consistent with hold semantics.

**Duplicate-PR claim guard (prclaims.go).** I checked whether this changes claim counting: it does not. `FetchClaims` lists open PRs directly from the GitHub API (`State: "open"`, which has always included drafts) and never reads `PRResult.Items` or the new `StaleDrafts` field. So an open App draft claiming issue N already suppressed re-offering N before this PR, and still does. The interaction is actually virtuous: the claim guard suppresses the issue while the stale draft sits open, and this change is the mechanism that finally gets that draft closed or finished — releasing the claim (via ledger reconcile) instead of holding the issue hostage forever. No regression, and arguably this fixes the one way a stale claim could persist indefinitely.

**Coverage.** The `buildScannerMessage` addition shipped untested and dropped `v2/pkg/scheduler` below its coverage floor (88.7% vs 89%). I pushed a maintainer-edit commit adding `stale_draft_message_test.go` covering the section header/count, per-draft lines, absence when empty, and the 70-rune title truncation (runes, not bytes).

**One observation, not blocking:** staleness is measured from `CreatedAt`, so an App draft the fleet is actively pushing to would still be flagged after 48h. Since the prompt offers "finish, mark ready, or close" (not just close), that's acceptable — but `UpdatedAt` would be the natural refinement if this ever nags about in-progress work.

DCO signed on both commits.

Issue #3973, opened 2026-08-17
[epic] PR reach telemetry — did the code we merged ever actually run?


clubanderson:
## Motivation

Every quality signal Hive has measures the work at the moment it is proposed or approved: acceptance rate, coverage, CI pass rate, review verdicts. None answers the question that actually matters — **did the code we merged ever get executed, by anyone?**

This came out of a reviewer challenge to the ACMM paper. The objection: PR acceptance rate is gameable by size, and a high acceptance rate says nothing about whether the work was *interesting*. His proposal was to weight PRs by downstream execution time.

Execution time alone is the wrong measure — it would rank a hot render loop above a nil-check in an error path that runs monthly and prevents data loss. Rarely-executed code is often the highest-stakes code. But the instinct is right: **weight the work by downstream reality rather than by reviewer approval.**

## Why this is now tractable

We have OpenTelemetry integrated. Spans give us the missing half:

- **Reach** — did the code path a PR touched ever execute in production, and for how many distinct users/hives?
- **Error-rate delta** — did failures in that path move after the PR merged? This is the closest thing to a real defect-escape signal, which today does not exist anywhere in the tree.
- **First-execution latency** — how long between merge and the first real invocation? A PR whose code never executes is a different outcome from one that runs immediately, and acceptance rate scores them identically.

Note the deliberate framing: *reach*, not *time spent*. A path executed once by every hive during an upgrade matters more than a hot loop, and this framing captures that.

## The hard part, stated up front

The mapping from **PR → code paths → spans** does not exist and is the bulk of the work. Options worth weighing:

- Instrument at package/function granularity and map changed files to span names (coarse, cheap, probably enough).
- Emit a build/version attribute on spans so a span can be attributed to the commit range that introduced it.
- Accept that only *some* PRs are attributable — infra and config changes may never map cleanly — and report coverage of the metric itself honestly.

## Sequencing note that will bite

**Merged is not deployed.** A PR can merge, be tagged, and still not be running anywhere — moving tags do not cause a repull, and we have already had a case (#3816) where the published digest advanced while the binary stayed old, because the final-stage `COPY --from=builder` came from cache. Any reach metric must be anchored to the **digest actually running**, not to the merge event, or it will silently attribute zero reach to code that was never deployed and call it unused.

## Relationship to the other measurement issues

- Merge-success/survival answers *did it last*.
- Reach answers *did anyone use it*.
[...]

PR #3976, opened 2026-08-17
feat(tracing): version-stamp spans for PR reach attribution (#3973)

clubanderson:
## Phase 1 of PR-reach telemetry

Every existing quality signal (acceptance rate, coverage, CI) measures work at proposal/approval time; none answers whether merged code ever actually ran. This PR lands the cheap, load-bearing foundation: making span data attributable to the commit that is actually executing.

### What phase 1 delivers

- **Version-stamped resource** — every span now carries:
  - `service.version` + `hive.commit`: the ldflags-baked commit of the *running* binary (the same value `hive --version` prints). This is the anchoring rule from the issue: reach must attribute to what actually runs, never the merge/publish event — #3816 showed the published digest can advance while the binary stays old, and without this anchor undeployed code reads as "unused".
  - `hive.image`: the Deployment-declared image ref via the existing cached `hub.SelfDeploymentImage()` read (warm from the release-channel lookup at startup; empty outside a cluster). Kept separate from `hive.commit` because the two can disagree (#3816) and that disagreement is itself signal.
  - Builds without ldflags injection (`gitShort == "unknown"`) stamp **no** version attributes, so reach queries never group spans under a placeholder pseudo-commit.
- **`hive.component` span attribute** — `tracing.StartSpan` auto-derives the component from the span-name prefix under the existing `<component>.<operation>` convention (`governor.eval_cycle` → `governor`). Zero call-site churn; this is the coarse package-level reach dimension.
- **Testable plumbing** — resource construction factored into `newResource()`; tests assert commit/image attributes present, placeholder version omitted, and the component attribute auto-attached (`tracetest` recorder, no network). The linker's ldflags injection itself is not tested — only the plumbing function.
- **Design note** — `v2/docs/design/pr-reach-telemetry.md`: what phase 1 emits, the changed-files → component mapping, the merged≠deployed anchoring rule, and phase-2 needs.

### Deliberately deferred (phase 2+)

- Error-rate deltas pre/post merge (needs a span-status discipline audit per component first)
- First-execution latency (merge → first span on a containing build)
- Per-hive distinct-count aggregation and any query/dashboard layer
- Finer-than-package attribution (per-PR/per-function) — rejected for phase 1 per the issue's options list
- No new endpoints, no aggregation service, no dashboards — phase 1 is emit-only

Builds on the existing live OTel integration (`pkg/tracing`, opt-in via the `otel:` config block); no parallel telemetry path introduced.

Part of #3973

### DIFF
--- a/v2/cmd/hive/main.go
+++ b/v2/cmd/hive/main.go
@@ -1027,6 +1027,13 @@ func main() {
 		SampleRatio: otelCfg.SampleRatio,
 		HiveID:      cfg.HiveID,
 		Branch:      cfg.Policies.Branch,
+		// Reach anchors (#3973): gitShort is the ldflags-baked commit of THIS
+		// binary (already canonicalized to 7 chars above), and the image ref is
+		// the Deployment-declared image (cached — warmed by the release-channel
+		// read at startup; "" outside a cluster). Spans attribute to the code
+		// that actually runs, not to the merge/publish event (#3816).
+		Commit: gitShort,
+		Image:  hub.SelfDeploymentImage(),
 	})
 	if traceErr != nil {
 		logger.Warn("tracing init failed; continuing without tracing", "error", traceErr)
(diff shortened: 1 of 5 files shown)


clubanderson:
Duplicate of #3975 — two retry loops both landed during the GitHub API outage recovery. All work continues on #3975.

PR #3979, opened 2026-08-17
fix(ci): notify on cancelled/timed-out v2 Tests runs, not just failures

Danathar:
## Summary

Addresses a `ci-maintainer` finding from the #2364 advisory digest:

> **[workflow-gap]** v2-tests.yml: create-issue-on-failure silently skips cancelled (timed-out) runs
> The create-issue-on-failure job in v2-tests.yml uses condition `if: failure() && github.event_name == schedule`. When the test job is cancelled (e.g. due to job timeout, or a runner preemption), GitHub sets conclusion=cancelled on the job — which does NOT satisfy `failure()`. As a result, scheduled v2-test runs that time out produce zero notification to operators.

I verified this is still live on the current `v4` default branch (the workflow's condition was untouched). Note: most other findings in the #2364 digest I checked (SSRF guards, `requireOwnerRole` gates, `hub_keys.go`/`audit.go` coverage gaps, `contribute_ws.go` pendingToken cleanup) turned out to already be fixed by prior merged PRs — the digest just hasn't pruned them. This one was a genuine, still-open gap.

- `create-issue-on-failure` now runs on `(failure() || cancelled())`, matching the digest's suggested fix.
- The filed issue's title/body distinguish a genuine test failure from a cancellation/timeout, so operators aren't misled about which happened.
- Kept the duplicate-detection prefix (`[Tests] v2 test`) broad enough to still dedupe across both failure and cancellation issues into one thread.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/v2-tests.yml'))"` — YAML parses cleanly
- [x] Manually traced both branches of the new `if` conditions in the `github-script` step
- [ ] No way to trigger an actual cancelled scheduled run from a PR; this is exercised for real the next time a scheduled run times out

### DIFF
--- a/.github/workflows/v2-tests.yml
+++ b/.github/workflows/v2-tests.yml
@@ -38,7 +38,7 @@ jobs:
         run: go test ./pkg/... -short -race -count=1
 
   create-issue-on-failure:
-    if: failure() && github.event_name == 'schedule'
+    if: (failure() || cancelled()) && github.event_name == 'schedule'
     runs-on: ubuntu-latest
     needs: test
     permissions:
@@ -51,7 +51,9 @@ jobs:
           script: |
             const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
             const sha = context.sha.substring(0, 7);
-            const title = `[Tests] v2 test failure on ${sha}`;
+            const wasCancelled = '${{ needs.test.result }}' === 'cancelled';
+            const kind = wasCancelled ? 'cancelled (likely timed out)' : 'failure';
+            const title = `[Tests] v2 test ${wasCancelled ? 'cancelled' : 'failure'} on ${sha}`;
 
             const existing = await github.rest.issues.listForRepo({
               owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


clubanderson:
Verified the condition. `if: (failure() || cancelled()) && github.event_name == 'schedule'` is correct as written — no fix needed.

- Schedule-only gate preserved (`github.event_name == 'schedule'` unchanged), so PR runs still never file issues.
- The classic footgun (a dependent job auto-skipping when its `needs:` job is cancelled) only applies when the job has **no explicit `if:`** — the implicit `success()` then causes a silent skip. Here `create-issue-on-failure` already carries an explicit `if:` using the status-check functions, so GHA evaluates it regardless of `test`'s outcome; `always()` isn't required. `cancelled()` at job level reflects cancellation anywhere in the `needs` chain, so `(failure() || cancelled())` is equivalent to `!success() && !skipped()` and behaves as intended.
- `needs.test.result == 'cancelled'` in the script step is likewise valid.
- Diff is scoped to the `if:` condition plus title/body/dedup-prefix strings for the new cancelled-vs-failed distinction — no trigger or permissions changes smuggled in.

Approving.

Issue #3768, opened 2026-08-14
🐛 bug:


castrojo:
## What happened?

Hive keeps assigning the same issue to multiple people: https://github.com/projectbluefin/dakota/pulls

All of us trying to fix the same bug: https://github.com/projectbluefin/dakota/issues/353

<img width="1134" height="630" alt="Image" src="https://github.com/user-attachments/assets/a3f35bb1-6cca-4adb-9f5d-74e90e2952e8" />

clubanderson:
Fixed on v4 by #3792 (merge commit ea3a7cc896cdaa18baa44145ab8c9c3d5903bc75).

Root cause was a gap between two existing guards: the duplicate-PR claim guard only recorded claims from PRs authored by the hive's own identity — a contribute-portal contributor's open PR (from their personal account) never suppressed the issue — and the contribute queue's cooldown downgraded to the short 4-hour form whenever a completion report was missed, re-offering the issue each window. That combination is exactly how dakota#353 collected one duplicate PR per window.

The fix records claims from ALL open PRs (external authors tagged, hive claims still win collisions so a stranger's junk PR can't freeze the agent pipeline), and the contribute task selector now skips any issue claimed by any open PR, atomically inside the offer→claim critical section. Reaches hives on their next image roll.

PR #3986, opened 2026-08-17
🐛 fix(contribute): stop re-offering issues whose open PR uses a non-closing reference (Refs #N)

### LINKED ISSUE #3980: 🐛 bug(contribute): issues with a non-closing (`Refs #N`) PR are re-offered every cooldown window, forever
## What happened?

A connected contributor client is handed the **same issues over and over** — roughly once per 4h cooldown window — even though that same contributor already has an **open PR** addressing each of them. Every re-offer burns a complete task cycle (fresh CLI turn, full context, tokens) for the agent to re-derive "I already did this, there is nothing new to ship," and then go idle without opening a PR — which immediately re-arms the short cooldown and schedules the next pointless re-offer.

Observed on a hosted hub over a single session, cycling between at least three issues:

| Issue | Contributor's existing work | Re-offered? |
|---|---|---|
| #2364 | PR #3979 (open) | yes |
| #3498 | PR #3898 (open) | yes |
| #2547 | #3948 / #3930 / #3901 (merged); rest maintainer-gated | yes |

This is the same class of symptom #3792 fixed for `Fixes`-keyword PRs (and #3768 before it), but through a path that fix does not cover.

Notably, **the agent behaves correctly every single round** — it recognizes the duplicate and declines to open a redundant PR. The loop is pure waste, not incorrect output.

## Root cause

Two independent gaps reinforce each other. Either alone would be survivable; together they form a closed loop with no exit.

### 1. The claim ledger only recognizes *closing* keywords

`ParseClaimedIssues` (`v2/pkg/github/prclaims.go`) matches only GitHub's closing keywords (`close(s|d)`, `fix(es|ed)`, `resolve(s|d)`). The comment is explicit that this is deliberate: *"Non-closing mentions ('see #12', 'related to #12') deliberately do NOT match."*

But "PR partially addresses a design-gated issue, so do **not** claim to close it" is exactly the convention a careful contributor follows — and it is the convention these agents follow on their own initiative. PR #3898 says so in its own body:

> Refs #3498 — no `Fixes` keyword, deliberately. The issue left the scaling curve open (...). Both are maintainer calls.

So the PR is invisible to the ledger, `Dependencies.IssueClaimed` returns false, and `selectTask` re-offers the issue. **Doing the right thing is what makes the work invisible.**

The `issueFromBranchName` fallback does not rescue it either, because neither branch carries an issue number.

Verified empirically against the patterns as written (control cases included to show the regexes themselves are fine):

```
=== ParseClaimedIssues (closing-keyword match) ===
PR #3898 (issue 3498)    claims=0 []
PR #3979 (issue 2364)    claims=0 []
control: real claim      claims=1 [[Fixes #3768 Fixes  3768]]

=== issueFromBranchName fallback ===
fix/v2-tests-notify-on-cancel          -> []
feat/governor-autoscale-thresholds     -> []
fix-3498                               -> [fix-3498 3498]
```

### 2. "Nothing to do" is wire-indistinguishable from "went idle without shipping"

`markTaskCompleted` (`v2/pkg/dashboard/contribute_ws.go`) picks the cooldown purely on whether a PR URL was reported:

- completion **with** a PR → `completedTaskCooldownHours` (168h)
- completion **without** a PR → `completedNoPRCooldownHours` (**4h**)

The 4h path exists so an issue where nothing shipped is not locked out for a week. But the protocol gives the relay no way to distinguish:

- "the agent tried and shipped nothing" (retry soon — correct), from
- "the agent correctly determined there is nothing to ship, because the work is already open in a PR or blocked on a maintainer decision" (do **not** retry soon).

Both send `task_complete` with `pr_url: ''`. So a correct no-op verdict earns the **fastest possible re-offer**, and the loop closes.

For #2547 this is unbounded: the shippable parts already merged, the remainder is explicitly gated on an unanswered maintainer design question, the issue stays open — and merged PRs are not open PRs, so no claim exists to suppress it. There is no state in which that issue stops being re-offered.

## What this is *not*

I initially suspected the relay's completion heuristic was firing early (`classifyTmuxPane` in `bin/contributor-relay.sh` keys `hasIdlePrompt` off Claude Code's always-present `bypass permissions` footer, which looked fragile). **That is not what is happening here.** Running the relay's own exported classifier against the live pane returns `IDLE_COMPLETE` correctly — the agent genuinely had finished its turn:

```
classifyTmuxPane => IDLE_COMPLETE
blockedOnHuman   => false
```

The relay is reporting accurately. This is a hub-side dispatch/suppression gap, not a relay bug.

## Expected behavior

A contributor should not be re-offered an issue it already has an open PR against, and an agent's correct "nothing to ship here" verdict should not earn the shortest possible re-offer interval.

## Reproduction steps

1. Connect a contributor relay to a hub with an issue whose only actionable work is already covered by that contributor's **open** PR, where the PR references the issue **without** a closing keyword (e.g. `Refs #N`) and the branch name carries no issue number.
2. Let the agent run the task. It correctly declines to open a duplicate PR and returns to idle; the relay sends `task_complete` with an empty `pr_url`.
3. Wait out `completedNoPRCooldownHours` (4h).
4. The same issue is dispatched again. Repeats indefinitely.

## Environment

- Branch/version: `v4` @ `04906734`
- Deployment: hosted hub (`hosted-kubestellar-hive-ojv6.hive.kubestellar.io`), local contributor relay, `claude` backend
- Contributor identity: `Danathar`

## Possible directions

Deliberately not prescribing a fix — the right shape depends on how much authority a self-reported verdict should carry:

- **Record non-closing references as a weaker claim.** `Refs #N` / `Addresses #N` could register an *attention* claim that suppresses re-dispatch without implying the issue closes. Keeps the existing "external claims never suppress agent work" rule intact.
- **Let a completion carry a verdict.** Allow the relay to report `no_work_needed` (distinct from `completed`) and give that its own long cooldown, so an agent's correct no-op stops being punished with the fastest retry.
- **Treat "open PR by this contributor touching this issue" as suppression** independent of keyword parsing.
- **Cap repeat dispatch per (contributor, issue) pair** as a backstop, so no parsing gap can produce an unbounded loop.

The last one seems worth having regardless — it bounds the blast radius of any future gap in the first three.

## Provenance

Filed by **Claude Opus 5 (high reasoning effort)**, running as Claude Code, after the repo author observed the repeated-task behavior on a live contributor client and asked for it to be investigated. The investigation was model-driven; the initial hypothesis (a relay-side false-completion bug) was **wrong** and was discarded after being tested against the live pane — see *What this is not* above. Every claim in the "Root cause" section was verified empirically before filing, and the limits of that verification are stated below.

## What I could not verify

I have no access to the hosted hub's logs or its deployed ledger state, so the hub-side half of the chain is inferred from reading the code at `04906734` plus the observed dispatch behavior, not confirmed from a live ledger dump. The claim-parsing and relay-classifier halves above are directly verified.

clubanderson:
Fixes #3980

## The loop

A contributor client was handed the same issues every 4h cooldown window even though its own **open PRs** already addressed them — whenever the PR deliberately used a non-closing reference (`Refs #N`) because the issue is maintainer-gated. `ParseClaimedIssues` only recognized GitHub's closing keywords, so the PR was invisible to the claim ledger, `Dependencies.IssueClaimed` returned false, `selectTask` re-offered, the agent correctly concluded "nothing new to ship", went idle without a PR, earned the *shortest* cooldown (4h), and the loop closed. Observed cycling: #2364↔PR #3979, #3498↔PR #3898, and #2547 (work merged via `Part of`; issue open on a maintainer ROUTE decision).

This is the third fix in the family: #3768 (contribute queue honours external `Fixes` claims) → #3792 (ledger + hive-precedence + red-stale valve) → this.

### 1. Non-closing references now register as SOFT claims (`v2/pkg/github/prclaims.go`)

`FetchClaims` additionally parses `Refs #N` / `Ref #N` / `Part of #N` / `Relates to #N` / `Related to #N` / `Addresses #N` — case-insensitive, optional colon, same cross-repo `owner/repo#N` grammar as the closing keywords — and records them with a new `SoftReference` flag. A bare `#N` or prose mention ("see #12", a changelog line) still never claims: the keyword list is closed. When one PR says both `Fixes #N` and `refs #N`, the hard claim wins. The branch-name heuristic is unchanged (fires only when no *closing* ref exists, still yields a hard claim).

Ledger slot collisions resolve by **strength** (hard > soft; hive-authored > external within each) — a generalization of #3792's hive-precedence rule — so a weak claim can never overwrite the claim the agent-side filter depends on.

### 2. The claim-scope split (and the agent/contributor seam)

**Contribute queue: every claim suppresses, unbounded.** Hard or soft, hive or external — if any open PR references the issue with a recognized keyword, `selectTask`/`ReadyQueue` withhold it (via the existing `IssueClaimed` hook; no dashboard admission change needed). Handing a contributor an issue that anyone's open PR visibly covers is pure waste, and a soft reference is exactly the convention a *careful* contributor uses. Suppression lasts precisely as long as the PR is open: close/merge → next authoritative rescan releases the issue.

**Agent pipeline: weak claims DEFER (72h), only hive-hard claims block.** #3792 deliberately let external claims through unconditionally so a stranger's junk PR could never freeze the hive's own pipeline — but that left the original dakota#353 duplicate shape alive across the agent/contributor seam: a hive agent could start an issue a contributor's open PR was already fixing. This PR replaces the all-or-nothing bypass with a *bounded deferral*: an external claim (`Fixes` or soft) or any soft reference suppresses agent dispatch for `weakClaimAgentDeferralWindow` (72h, matching `claimLedgerTTL`) from the claim's **first observation** (`FirstObservedAt`, carried across scans and authoritative reconciles so the window ages instead of resetting), then releases even while the PR stays open. The junk-PR invariant survives in bounded form — a stranger's PR can *delay* the pipeline at most 72h, never freeze it — and the old red-stale valve still releases dead PRs early. The regression test that previously encoded the unconditional bypass now asserts both halves: fresh external claim defers, aged external claim releases, aged hive-hard claim still suppresses (positive control).

Known residual: an adversary *churning* new junk PRs could earn a fresh 72h window per PR (each new PR gets its own `FirstObservedAt`). That is an active-spam problem, strictly narrower than the single-junk-PR freeze the invariant guards, and rate-limited by PR creation itself.

### 3. Escalating no-PR cooldown (`v2/pkg/dashboard/contribute_ws.go`)

`task_complete` without a verified PR used to book a flat 4h cooldown forever. Now consecutive no-PR completions of the same issue double it — 4h → 8h → 16h → … capped at the with-PR cooldown (operator-tunable, 168h default). The streak lives **outside** the swept completion maps (the loop's signature is precisely that the completion entry expires and is swept before the next dispatch) and is persisted to `/data/contributors/no-pr-streaks.json`; a shipped verified PR or two weeks of quiet resets it. This is the backstop the issue asked for "regardless": whatever future parsing gap appears, per-issue waste converges to ~one task cycle per week instead of six per day. It bounds — but does not fully eliminate — the #2547 shape; the clean fix (a `no_work_needed` completion verdict) is filed as the follow-up below.

### 4. Drafts × #3970

`FetchClaims` does its own `PullRequests.List(state=open)` and is independent of the `fetchPRs` path #3970 (b3c92682) modified for stale-draft surfacing. Drafts were and remain visible to the claim scan — a draft is still "someone is on it" for offer-suppression — and are counted exactly once. Now pinned by tests (`a draft PR still claims its issue exactly once`, `a draft PR with a closing keyword hard-claims`) plus a comment in `FetchClaims` so a future draft filter can't silently blind the ledger.

## Compatibility

- Pre-#3980 ledgers load unchanged: missing `soft_reference` unmarshals as hard (those ledgers only ever held closing/branch claims), and zero `first_observed_at` falls back to `ObservedAt`, so an old external entry past the window releases rather than deferring forever.
- Legacy completed-tasks records are untouched; the streak file is new and optional.

## Tests

`v2/pkg/github` (all passing, full package run):
- `TestParseSoftReferencedIssues` — keyword-form table: all six forms, mixed case, colon, cross-repo, whitespace in multi-word keywords, the literal PR #3898 body, `Refs`+`Fixes` mixed, de-dup; negatives: bare/changelog `#N`, prose, `xrefs`/`unrelated to`, `refs/heads#5`, closing keywords, `#0`, empty.
- `TestFetchClaimsSoftReferences` — soft+external incident shape, hard+soft from one PR, same-issue overlap collapses to hard, draft claims once (soft and hard), branch heuristic unchanged beside soft refs, cross-repo soft key, bare-mention positive control.
- `TestFilterClaimedIssues_SoftClaimBoundedDeferral` — fresh/aged × soft/external deferral matrix, aged hive-hard still suppresses (positive control), zero-`FirstObservedAt` back-compat.
- `TestFilterClaimedIssues` (updated) — `fresh external claim defers agent work` / `aged external claim releases agent work (bounded junk-PR guard)`.
- `TestClaimLedgerStrengthPrecedence` — soft can never demote hard in either order, hive-soft > external-soft, soft-only recorded, `FirstObservedAt` survives re-observation but not a new PR.
- `TestSoftReferenceClaimLifecycle` — end-to-end replay mirroring #3792's: fetch open Refs-PR → claim present (contribute hook blocks) + agent deferral; PR closes → authoritative rescan releases.

`v2/pkg/dashboard`:
- `TestSoftClaimIssue_ReplaysThreeNineEightZero` — selectTask replay against a **real** `ClaimLedger`: soft-claimed #353 skipped → next issue offered → `no_matching_work` while claim holds → offerable again after release.
- `TestSoftClaimIssue_PositiveControl` — empty ledger leaves selection unchanged.
- `TestNoPRCooldownEscalatesGeometrically`, `TestNoPRCooldownStreakSurvivesCooldownSweep` (the exact loop shape: streak must outlive the swept completion entry), `TestNoPRCooldownResetByShippedPR`, `TestNoPRCooldownStreakForgetsAfterQuiet`.

## Follow-up (fourth family member)

The #2547 residual — an issue whose referenced PRs are all MERGED but which stays open on a maintainer decision — has no open-PR claim to suppress it and is only *bounded* by the cooldown escalation here. The clean fix is a completion **verdict**: let the relay report `no_work_needed` distinct from an idle no-PR completion and give it a long cooldown of its own. Filed separately with the precise mechanism (issue link in comments below).

### DIFF
--- a/v2/pkg/dashboard/contribute_ws.go
+++ b/v2/pkg/dashboard/contribute_ws.go
@@ -354,8 +354,21 @@ type ContributeWSHub struct {
 	// A permanent failure (msg.Permanent) counts as permanentFailureWeight toward
 	// the threshold. Guarded by completedMu.
 	consecutiveFailures map[string]int
-	completedMu         sync.Mutex
-	selectMu            sync.Mutex
+	// noPRStreaks counts CONSECUTIVE no-PR completions per "repo#number"
+	// (kubestellar/hive#3980). Each one doubles the next no-PR cooldown
+	// (advanceNoPRStreakLocked), so an issue that keeps "completing" with
+	// nothing to ship — e.g. the #2547 shape, where the shippable parts
+	// merged and the rest is maintainer-gated but the issue stays open —
+	// backs off geometrically instead of burning a full contributor task
+	// cycle every completedNoPRCooldownHours forever. Deliberately a
+	// SEPARATE map from completedTasks: the loop's signature is precisely
+	// that the completion entry EXPIRES (and is swept) before the next
+	// dispatch, so the streak must survive that sweep. Reset by a completion
+	// that ships a verified PR, or by noPRStreakResetAfter of quiet. Guarded
+	// by completedMu.
+	noPRStreaks map[string]noPRStreakRecord
+	completedMu sync.Mutex
+	selectMu    sync.Mutex
 	// assignmentTimes records, per contributor identity (identityOf), the wall-clock
 	// times of the task_assign messages that identity has been handed. It backs the
 	// #2436/#2566 per-tier rate gate: tier_limits.max_per_hour / max_per_day were
@@ -545,6 +558,17 @@ const completedTaskCooldownHours = 168
 // the same day.
 const completedNoPRCooldownHours = 4
 
+// noPRStreakResetAfter is how long a no-PR completion streak (#3980) survives
+// without another no-PR completion before it forgets itself and the next
+// no-PR cooldown starts back at completedNoPRCooldownHours. It must be LONGER
+// than the streak's own cooldown cap (the with-PR cooldown, default
+// completedTaskCooldownHours = 168h): the whole point is that the issue is
+// only re-dispatched AFTER its escalated cooldown expires, so a reset window
+// shorter than the cap would wipe the streak before the capped re-offer ever
+// happened and the backoff could never reach — or hold — its ceiling. Two
+// weeks = 2× the default cap, so one full capped cycle plus slack.
+const noPRStreakResetAfter = 14 * 24 * time.Hour
+
 // failedTaskCooldownMinutes is the SHORT cooldown applied to an issue when a
 // contributor reports task_failed (#2435). It is deliberately far shorter than
 // the completion cooldowns above: a failure is often purely environmental
@@ -612,6 +636,7 @@ func NewContributeWSHub(logger *slog.Logger, server *Server) *ContributeWSHub {
 		completedTaskPRURL:    make(map[string]string),
 		failedTasks:           make(map[string]time.Time),
 		consecutiveFailures:   make(map[string]int),
+		noPRStreaks:           make(map[string]noPRStreakRecord),
 		assignmentTimes:       make(map[string][]time.Time),
 		leases:                make(map[string]*taskLease),
 		yankExclusions:        make(map[string]time.Time),
@@ -621,6 +646,7 @@ func NewContributeWSHub(logger *slog.Logger, server *Server) *ContributeWSHub {
 	}
 	hub.loadCompletedTasks()
 	hub.loadFailedTasks()
+	hub.loadNoPRStreaks()
 	hub.loadActivity()
 	go hub.cleanupLoop()
 	return hub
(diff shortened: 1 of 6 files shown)


clubanderson:
Follow-up for the #2547-shape residual (merged-only reference PRs, maintainer-gated issue) filed as #3987 — the completion-verdict mechanism described in the PR body.

clubanderson:
Closing in favor of #3985 — the issue author's own fix for #3980, which was opened first and is the better core fix. Notably, its bounded prose-gap grammar (`keyword … #N`, ≤40 chars, no sentence/`#` crossing) correctly catches the *real* PR #3979 body ("Addresses a `ci-maintainer` finding from the #2364 advisory digest"), one of the three observed loop cases, which this PR's stricter `keyword[:] #N` grammar would have missed. Its keyword set and the reasoned exclusion of `relates to`/`see` (topical cross-reference vs. work claim) are also the more careful call, and its reference tier is consulted only when the closing and branch tiers yield nothing, so no existing claim changes strength.

Pieces from this PR that #3985 deliberately left open will follow separately, rebased on top of it:

- the escalating no-PR completion cooldown (4h → 8h → … capped at the with-PR cooldown) — the repeat-dispatch backstop #3985's own scope section calls "worth having regardless", which bounds the #2547 shape (no open PR left to claim the issue) and any future parsing gap;
- the #3970 interaction pin: tests asserting a draft PR still registers its claim exactly once through `FetchClaims`' own listing;
- a real-ledger lifecycle replay (claim suppresses the offer while the PR is open, releases it on close).

The agent↔contributor seam analysis (whether weak claims should *defer* the agent pipeline for a bounded window rather than being ignored outright) will be posted on #3980 as analysis rather than code. The #2547-shape completion-verdict follow-up is already filed as #3987.

PR #3990, opened 2026-08-17
🐛 fix(contribute): clone the branch the hub is built from, not a hardcoded v2

Danathar:
## Summary

The contribute onboarding page tells every contributor to run:

```
git clone -b v2 https://github.com/kubestellar/hive && cd hive
```

**`v2` is no longer maintained.** `v4` is the repository default branch and is **477 commits ahead**. Contributors have been hand-editing the command to `v4` to get a working client.

This is not cosmetic drift — the two branches ship different wire protocols:

| | relay protocol | hub protocol |
|---|---|---|
| `v2` | **1.1** | 1.2 |
| `v4` | **1.2** | 1.2 |

So the command as printed produces a client the hub has to flag as a protocol-drifted peer, via the very detection added in #3930.

## Fix

Derive the branch from `upstreamBranch()` — the branch this binary was **built from** (injected via `-X main.gitBranch` in the image build), which already backs the self-version check. The command now self-corrects as the fleet moves between branches instead of pinning a constant that silently rots.

That mirrors the reasoning already recorded in `pkg/hub`'s `upgradeBranchOrDefault`:

> The hub's OWN branch is the right fallback: it is what the hub is built from, so it is the only branch the hub can honestly assume [...] and it self-corrects as the fleet moves between branches instead of pinning a constant that silently rots.

### Why the hub's branch, and not "the latest" or "the released" one

- **Not "newest branch":** a contributor's relay has to speak the protocol of the hub it is joining. That is exactly what the 1.1-against-1.2 mismatch above demonstrates — if a `v5` were cut tomorrow, "latest" would reintroduce the same defect in the other direction.
- **Not "latest release":** there is nothing to pin to. The repo carries **no git tags and no GitHub releases**, and the `stable`/`candidate`/`edge` release channels are *container image* retags, not git refs — `git clone` cannot target them.

### Why a sentinel pre-pass rather than another `%s`

That format string carries **eleven positional arguments across several thousand lines**; inserting a verb mid-template would renumber every argument after it. Substituting `{{HIVE_BRANCH}}` on the *template* — before `Fprintf` renders it — also guarantees no rendered value (project name, `Host` header) can contain the sentinel and be rewritten by it.

## Also

- `defaultUpstreamBranch` `v2` → `v4`. It is the fallback when no branch was injected (local `go run`); pointing it at an unmaintained branch is the same stale-constant defect.
- The hub's `get-started.html` self-host snippet carried the identical hardcoded clone.

## Explicitly NOT touched

The operational branch defaults in `pkg/hub` (`server.go:1155`, `server.go:2021`, `saas_bulk.go:98`) also hardcode `"v2"`. Those decide what branch hosted spokes **track and upgrade onto** — deployment semantics and a maintainer call, not a docs fix. Flagging them rather than bundling them; happy to follow up if you want them changed.

## Test plan

New `api_contribute_clone_branch_test.go`, all passing:

- [x] Hub built from `v4` emits `-b v4`; never the old `v2`; no unsubstituted sentinel leaks
- [x] **Genuinely derived, not re-hardcoded** — a hub built from `v5-experimental` emits that branch and *not* `v4`
- [x] A build with no injected branch (`""`/`"unknown"`, i.e. local `go run`) falls back cleanly instead of emitting `git clone -b unknown`
- [x] `defaultUpstreamBranch` guard fails if it ever returns to `v2`
- [x] **All five copy-paste surfaces** counted — default `<pre>`, container/host/k8s mode templates, and the prefilled assistant prompt — so a fix reaching only the visible one cannot pass
- [x] `go test ./pkg/dashboard/ ./pkg/hub/` both green; `go build ./...`, `go vet`, `gofmt` clean

### DIFF
--- a/v2/pkg/dashboard/api.go
+++ b/v2/pkg/dashboard/api.go
@@ -298,8 +298,15 @@ var (
 )
 
 // defaultUpstreamBranch is the fallback branch for the self-version check
-// when the build did not inject a branch (e.g. local `go run`).
-const defaultUpstreamBranch = "v2"
+// when the build did not inject a branch (e.g. local `go run`), and for the
+// clone command on the contribute onboarding page.
+//
+// v4, not v2: v2 is no longer maintained, so a build with no injected branch
+// was comparing itself against — and telling contributors to clone — a branch
+// that no longer receives the code this binary is built from. This is the same
+// stale-constant defect upgradeBranchOrDefault documents in pkg/hub, where a
+// hardcoded "v2" default resolved upgrade targets against a foreign branch.
+const defaultUpstreamBranch = "v4"
 
 func SetGitVersion(hash, short string) {
 	versionHash = hash
--- a/v2/pkg/dashboard/api_contribute.go
+++ b/v2/pkg/dashboard/api_contribute.go
@@ -785,7 +785,16 @@ func (s *Server) handleContributeLanding(w http.ResponseWriter, r *http.Request)
 	projectNameJS := jsStringLiteral(projectName)
 
 	w.Header().Set("Content-Type", "text/html; charset=utf-8")
-	fmt.Fprintf(w, `<!DOCTYPE html>
+	// {{HIVE_BRANCH}} is substituted into the TEMPLATE, before Fprintf renders
+	// it, so the onboarding clone command names the branch this hub is actually
+	// built from instead of a hardcoded one (kubestellar/hive#3990).
+	//
+	// It is a pre-pass rather than another %s because this format string carries
+	// eleven positional arguments across several thousand lines; inserting a
+	// verb mid-template would renumber every argument after it. Substituting on
+	// the template also means no rendered value — project name, Host header —
+	// can ever contain the sentinel and be rewritten by it.
+	fmt.Fprintf(w, strings.ReplaceAll(`<!DOCTYPE html>
 <html><head><meta charset="UTF-8"><meta name="viewport" content="width=device-width,initial-scale=1"><title>Contribute to %s</title>
 <style>
 /* Michroma display face, base64-embedded (no network fonts). Used ONLY by the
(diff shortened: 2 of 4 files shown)


clubanderson (commented):
Reviewed. Confirmed the fix is genuinely dynamic, not a re-hardcode:

- `upstreamBranch()` in `api.go` returns `versionBranch` (the `-X main.gitBranch` ldflags value baked in at build time — the same value backing the self-version pill) when it's set and not `"unknown"`, else falls back to `defaultUpstreamBranch`.
- `api_contribute.go` substitutes `{{HIVE_BRANCH}} -> upstreamBranch()` on the raw template string before `Fprintf` applies the positional `%s` args, so no rendered value (project name, Host header) can collide with the sentinel. Good approach given the 11-arg format string.
- Fallback checked: `defaultUpstreamBranch = "v4"` covers both the local-`go run` case (`versionBranch == "unknown"`) and empty string, so it can't render a broken `git clone -b  ` or `git clone -b unknown`. Test `TestContributeCloneBranchUnknownBuildFallsBack` pins both cases.
- Test `TestContributeCloneBranchTracksADifferentBranch` (hub built from `v5-experimental` emits that, not `v4`) is the right check to rule out a re-hardcode.
- All five copy-paste surfaces (default `<pre>`, container/host/k8s templates, prefilled prompt) are covered and counted by `TestContributeCloneBranchAppliesToEveryCopyPasteSurface`.
- `get-started.html` is a static asset (not server-rendered), so it can't use `upstreamBranch()` — the plain `v2` -> `v4` constant bump there is the right scope for a static file, and the PR description discloses it rather than hiding it.

On the "hardcoded branch" question specifically: this doesn't run into the `release_channels.go` precedent (`ReleaseChannelStable`/`Candidate`/`Edge` forbidding a hardcoded channel->branch mapping) — that guards a different mapping (channel to branch, which the PR correctly leaves untouched in `pkg/hub`). This PR's `defaultUpstreamBranch` is a single fallback constant for the "no branch injected" case, not a channel map, so no conflict there.

DCO present. LGTM.

PR #4003, opened 2026-08-17
test(escalation): raise coverage above 90%

Danathar:
## What changed

Adds focused coverage for `pkg/escalation.Store.Excerpt`, asserting both persisted CI evidence and the empty result for an unknown PR.

## Why

`pkg/escalation` measured 89.7% in the coverage workflow and was one of the packages below the 90% goal tracked by #3903. The new test raises it to 93.2% without changing production code or the coverage floor.

## Validation

- `go test ./pkg/escalation -short -count=1`
- `go test ./pkg/escalation -short -race -count=1`
- `go vet ./pkg/escalation`
- `gofmt -d v2/pkg/escalation/escalation_test.go`
- `git diff --check`

Refs #3903

Contributor: GPT 5.6 luna high.

### DIFF
--- a/src/pkg/escalation/escalation_test.go
+++ b/src/pkg/escalation/escalation_test.go
@@ -67,6 +67,18 @@ func TestSweep_GreenResetsAndAbsencePrunes(t *testing.T) {
 	}
 }
 
+func TestExcerpt_ReturnsStoredEvidenceOrEmpty(t *testing.T) {
+	s := Load(filepath.Join(t.TempDir(), "streaks.json"))
+
+	s.Sweep([]Observation{obs("org/repo", 7, "sha1", true)}, 3)
+	if got := s.Excerpt("org/repo", 7); got != "ReferenceError: seedMission is not defined" {
+		t.Fatalf("Excerpt = %q, want the stored CI evidence", got)
+	}
+	if got := s.Excerpt("org/repo", 8); got != "" {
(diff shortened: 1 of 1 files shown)


clubanderson:
Reviewed the diff and the covered function. The test change itself is solid:

- Diff scope is clean — only `v2/pkg/escalation/escalation_test.go` changed (+12/-0), no production or workflow edits.
- `TestExcerpt_ReturnsStoredEvidenceOrEmpty` asserts real invariants against `Store.Excerpt`: it seeds a red observation with an excerpt via `Sweep`, then asserts `Excerpt("org/repo", 7)` round-trips the persisted `LastExcerpt` value, and separately asserts `Excerpt("org/repo", 8)` for an unobserved PR returns `""` (the `e == nil` branch). Both are genuine behavioral assertions, not padding.

Blocked from merging for two reasons unrelated to the test content:

1. **DCO fails** — the single commit (`7d385e8f`) has no `Signed-off-by` trailer.
2. **Merge conflict** — GitHub reports `mergeable: false`, `mergeable_state: dirty`, `rebaseable: false` against current `v4`. Notably, current `v4` no longer has a `v2/pkg/escalation/` path (the package now lives at `src/pkg/escalation/`), so this branch appears to be based on a stale `v4` tree layout and will need a rebase.

Please rebase onto current `v4` (package now at `src/pkg/escalation/`) and re-sign the commit with `git commit --amend -s`. Happy to merge once DCO and the conflict are resolved.

PR #4005, opened 2026-08-17
fix: rate limit nginx authentication endpoints

### LINKED ISSUE #3906: [sec-check] nginx.conf: no rate limiting on auth endpoints — brute-force / token-enumeration risk
## Security Finding

**Severity**: medium
**Type**: unsafe-pattern

`v2/deploy/nginx.conf` proxies all requests to the Go API with no `limit_req` directives. The authentication-sensitive paths `/api/auth/token` (device-flow OAuth callback) and the dashboard login routes have no per-IP rate limiting or connection throttling.

Without `limit_req_zone` + `limit_req`, an attacker who can reach the nginx port can:
- Automate high-volume token-guessing attempts against `/api/auth/token` without triggering any server-side slowdown
- Exhaust the GitHub OAuth device-flow rate limits (5k/hr) via repeated flows from a single IP
- Conduct timing-oracle attacks at higher volume (even though the Go server uses `subtle.ConstantTimeCompare`, bulk volume reduces the noise floor)

## Impact

An attacker with network access to the hive (self-hosted deployments where port 3001 is exposed) could automate brute-force of authentication credentials. The risk is higher if a weak/short `HIVE_DASHBOARD_TOKEN` is set.

## Recommendation

Add `limit_req_zone` and `limit_req` directives to `v2/deploy/nginx.conf`:

```nginx
limit_req_zone $binary_remote_addr zone=auth_limit:10m rate=5r/m;
# ...inside the auth location blocks:
limit_req zone=auth_limit burst=10 nodelay;
```

Apply stricter limits to `/api/auth/token` and login paths, while allowing higher throughput on read-only API paths.

---

Danathar:
## Summary

- add per-client nginx request zones for authentication-sensitive routes
- throttle `/api/auth/token` and `/sso` at 5 requests per minute, with a burst of 10
- throttle GitHub device-flow routes at 15 requests per minute so normal five-second polling remains usable
- return `429 Too Many Requests` when the auth throttle is exceeded
- add regression coverage that pins each protected location to its intended zone

## Why

The gateway previously proxied authentication-sensitive paths without an ingress limit. A reachable hive could therefore receive unlimited token-enumeration, device-flow, or SSO handoff attempts from one client address. The ordinary `/api/` and `/` proxy paths remain unthrottled.

Fixes #3906

## Validation

- `GOCACHE=/tmp/hive-go-cache CC=/usr/bin/gcc go test ./pkg/config -count=1`
- `git diff --check`
- nginx container syntax validation was unavailable because neither a local nginx binary nor the `nginx:1-alpine` image is installed; the existing configuration test and static directive checks pass.

Contributor: GPT 5.6 luna high.

### DIFF
--- a/src/deploy/nginx.conf
+++ b/src/deploy/nginx.conf
@@ -48,14 +48,21 @@ http {
         limit_req_status 429;
 
         location /api/ {
-            # These nested locations inherit the proxy and security-header
-            # settings above while applying the stricter auth-only bucket.
+            # These nested locations inherit the proxy_set_header /
+            # security-header / timeout settings above while applying the
+            # stricter auth-only bucket. proxy_pass is the one directive
+            # nginx does NOT inherit into a nested location: without it the
+            # request terminates in the empty block and falls through to
+            # static-file serving (a stock nginx 404), silently breaking
+            # every auth path. Each nested location must repeat it.
             location = /api/auth/token {
                 limit_req zone=auth_limit burst=10 nodelay;
+                proxy_pass http://hive_api;
             }
 
             location /api/gh-user-auth/ {
                 limit_req zone=device_flow_limit burst=10 nodelay;
+                proxy_pass http://hive_api;
             }
 
             proxy_pass http://hive_api;
@@ -92,8 +99,11 @@ http {
             # SSO is a public handoff endpoint because the signed token is the
             # credential. It still needs the same per-client throttle as the
             # device-flow routes to limit replay and brute-force traffic.
+            # proxy_pass must be repeated here (nested locations do not
+            # inherit it -- see the /api/ block above).
             location = /sso {
                 limit_req zone=auth_limit burst=10 nodelay;
+                proxy_pass http://hive_api;
             }
 
             proxy_pass http://hive_api;
--- a/src/deploy/nginx.conf
+++ b/src/deploy/nginx.conf
@@ -11,6 +11,17 @@ http {
         server hive:3001;
     }
 
+    # Authentication endpoints are intentionally cheap to call for a browser,
+    # but expensive to call at scale: /api/auth/token exposes whether a
+    # dashboard token is configured, while the device-flow and SSO paths mint
+    # sessions. Keep the bucket keyed on the binary address form so IPv4 and
+    # IPv6 clients use the same compact fixed-size key representation.
+    limit_req_zone $binary_remote_addr zone=auth_limit:10m rate=5r/m;
+    # GitHub's device-flow client polls about every five seconds while waiting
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed with a live nginx run, not just the conf text. Two findings, one fixed on the branch.

**Blocking bug (fixed in 15e34962): the nested auth locations 404'd every auth path.** nginx does not inherit `proxy_pass` into nested locations — it is a content-handler directive set per location. As pushed, `location = /api/auth/token`, `location /api/gh-user-auth/`, and `location = /sso` contained only `limit_req`, so a matching request terminated in the empty block and fell through to static-file serving. Reproduced against the pinned `nginx:alpine` gateway image with a stub upstream:

| Path | Before fix | After fix |
| --- | --- | --- |
| `/api/auth/token` | **404** (stock nginx page) | 200 proxied, 429 past burst |
| `/api/gh-user-auth/status` | **404** — device-flow login bricked | 200 proxied |
| `/sso` | **404** — hub SSO handoff bricked | 200 proxied |

The structural Go test stayed green throughout because it only greps the conf. The fix repeats `proxy_pass http://hive_api;` inside each nested location (everything else — `proxy_set_header`, the #3822 defence-in-depth headers, timeouts, `error_page` — genuinely does inherit, verified on the wire), and the regression test now extracts each auth location's own block body and asserts `limit_req` **and** `proxy_pass` inside it, so an empty nested location can never again pass the test.

**Zone-key / topology check (OK as-is):** this conf is mounted only by `src/docker-compose.yaml` — the Compose front door, published directly on host :3001 with no LB or extra proxy hop in front (the k8s path uses ingress-nginx, not this file). So `$binary_remote_addr` is the real client under Docker's iptables DNAT publishing and per-client limiting holds. If this gateway is ever put behind another proxy/LB, all users would collapse into one shared 5/min bucket (cheap login-DoS) — at that point it needs `real_ip` with `set_real_ip_from` pinned to the actual LB, never raw `X-Forwarded-For`. Worth remembering, not worth blocking.

**Budget math (verified in code):** device flow polls at `d.interval||5` seconds with `slow_down` backoff (`server.go` login page JS) = 12/min steady against the 15/min zone, leaving headroom for the `/status` calls that share the prefix; burst 10 absorbs page-load clusters. `/sso` is one handoff hit per login, so 5/min burst 10 is comfortable even with multiple tabs. `429`s on `/api/auth/token` degrade to the "not configured" branch in `_ensureAuthToken` — no crash, only affects >10 rapid reloads.

Burst behavior on the wire: `200 ×burst → 429 429 429 …`, `limit_req_status 429` honored, unthrottled `/api/health` unaffected.

PR #4007, opened 2026-08-17
✨ feat(contribute): bring agy up to parity — picker entry, verified headless mode, correct staging path

Danathar:
## What

agy (Google's **Antigravity CLI**) was half-wired. It already had a Justfile preflight, a `backends.conf` perm flag and relay readiness/idle patterns — so interactive host runs worked — but it was invisible on `/contribute`, and three things around it were wrong. All findings verified against a real **agy 1.1.13** install.

### 1. It *is* headless-capable; the repo said it wasn't

`agy --help` documents `-p/--print` — "Run a single prompt non-interactively and print the response". A print-mode run answers on stdout and exits 0, which is exactly the exit-code contract `runHeadlessTask()` relies on. Meanwhile `contributor-relay.sh` asserted "bob / agy / pi drive an interactive TUI" with no one-shot entry point, and `src/docs/contributor-relay.md` repeated it. agy now joins `HEADLESS_BACKENDS` with `{ flag: '-p' }`.

### 2. `--model` was silently doing nothing

The relay launched `agy --model X`. agy then warns `--model <m> requires --effort (available: low, medium, high)` and **ignores the model** — the hub-side launcher already knows this (`agyDefaultEffort` in `src/pkg/agent/manager.go`), the relay didn't. Both the interactive launch command and the headless argv now pair `--effort` with `--model`, defaulting to `low` to match the hub. `AGENT_REASONING_EFFORT` can override it, but only with `low|medium|high` — codex's vocabulary is wider (`minimal`), and forwarding an unknown token would make agy reject the pairing and drop the model again.

### 3. The container staging path didn't exist

`just contribute-hive` staged `${HOME}/.antigravitycli`. On a 1.1.x install that path is absent — agy keeps state under `${HOME}/.gemini/antigravity-cli` — so the mount was a silent no-op. Now stages whichever layout is present.

## Host-only, and honest about it

agy is offered in **Host** mode only. Its sign-in is an interactive Google OAuth flow — browser URL plus a pasted code, with no API-key mode — so no unattended container can hold a session. I ran the official `linux-x64` build in a clean container: it demands a fresh browser login regardless of what is mounted, and the contributor image doesn't ship the binary either.

Selecting agy therefore flips Mode to Host — generalizing the fallback the `other` entry already had into `HOST_ONLY_BACKENDS` — and shows a note explaining the flip, instead of generating containerized commands that cannot work.

That also means agy deliberately stays **out** of `K8S_HEADLESS_BACKENDS` and out of the Justfile's k8s list. Both now document the divergence from the relay's table so nobody "resyncs" them: capability and credential are separate questions, and a pod that starts and then fails auth would just crash-loop.

## Also

`TestContributeK8sHeadlessCapability` grepped the *whole page* for `"<backend>:1"`, so it matched any other JS object literal using the same key (my first draft of `HOST_ONLY_BACKENDS` tripped it). It now scans only the capability map's own literal.

## Tests

- `bin/contributor-relay.test.js`: agy moved to the one-shot table; new tests for the `--model`/`--effort` pairing, the rejection of efforts agy can't take, and the headless argv. **80/80 pass.**
- `src/pkg/dashboard`: new `TestContributeAgyHostOnlySurface` (option, install line, tile metadata, host-only forcing, the note, and that agy stays out of the k8s map); `TestContributeK8sHeadlessCapability`'s agy row keeps `false` with the corrected reason. Full package passes.
- `just contribute-check agy` run against the real CLI: preflight passes and prints the new model/effort/sign-in guidance.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -160,9 +160,18 @@ contribute-check-backend backend="claude":
         if command -v agy &>/dev/null; then
           echo "agy CLI detected ($(agy --version 2>&1 | head -1))"
           echo "  Models: gemini-3.6-flash, claude-sonnet-4-6, gpt-oss-120b, and more"
-          echo "  Set model: --model gemini-3.6-flash-high"
+          echo "  Set model: export AGENT_MODEL=gemini-3.6-flash-high"
+          echo "  Effort:    export AGENT_REASONING_EFFORT=low|medium|high (agy needs --effort with --model)"
+          # agy signs in through an interactive Google OAuth flow (browser URL
+          # plus a pasted code) and offers no API-key mode, so run it on the
+          # HOST: a container cannot inherit the sign-in. Sign in once with a
+          # bare `agy` before starting the relay.
+          echo "  Sign in once interactively (run: agy) — agy's Google OAuth cannot be"
+          echo "  completed by an unattended container, so run this backend on the host:"
+          echo "    just contribute-hive agy local"
         else
-          echo "ERROR: agy CLI not found. Install: https://antigravity.dev"
+          echo "ERROR: agy CLI not found. Install: https://antigravity.google/product/antigravity-cli"
+          echo "  Homebrew: brew install --cask antigravity-cli"
           exit 1
         fi
         ;;
@@ -608,7 +617,21 @@ contribute-hive backend="" mode="docker": check-version
           fi
           ;;
         agy)
-          if [ -d "${HOME}/.antigravitycli" ]; then
+          # agy 1.1.x keeps its state under ${HOME}/.gemini/antigravity-cli,
+          # NOT the ${HOME}/.antigravitycli this recipe staged before — on a
+          # 1.1.13 install that legacy path does not exist at all, so the mount
+          # was a silent no-op. Stage whichever is present (legacy first-run
+          # installs may still use the old path) so neither layout is dropped.
+          #
+          # Staging state is NOT the same as staging a session: agy authenticates
+          # through an interactive Google OAuth flow and keeps no credential file
+          # under HOME that a container can inherit (verified on 1.1.13 — a clean
+          # container asks for a browser login regardless of what is mounted).
+          # The /contribute page therefore offers agy in HOST mode only.
+          if [ -d "${HOME}/.gemini/antigravity-cli" ]; then
+            stage_copy "${HOME}/.gemini/antigravity-cli" "antigravity-cli"
+            CLI_MOUNTS="-v ${CLI_STAGE}/antigravity-cli:/home/dev/.gemini/antigravity-cli${VOLSUF}"
+          elif [ -d "${HOME}/.antigravitycli" ]; then
             stage_copy "${HOME}/.antigravitycli" ".antigravitycli"
             CLI_MOUNTS="-v ${CLI_STAGE}/.antigravitycli:/home/dev/.antigravitycli${VOLSUF}"
           fi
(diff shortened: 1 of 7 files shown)


clubanderson:
Reviewed with every claim checked against the repo (same bar as the codex-picker review on #4004). Verification table:

| # | Claim | Verified against | Result |
|---|-------|------------------|--------|
| 1 | agy was "half-wired": Justfile preflight, `backends.conf` perm flag, relay readiness/idle patterns all pre-exist | `Justfile` `contribute-check-backend` agy case; `config/backends.conf` (`KNOWN_BACKENDS` incl. agy, perm `--dangerously-skip-permissions`); `bin/contributor-relay.sh` (`? for shortcuts` ready pattern + idle/working classifier); `bin/contributor-agent.sh` detect_cli + CLAUDE.md symlink | ✅ all four surfaces existed on v4; only the /contribute page was missing |
| 2 | Headless flip: `agy -p` one-shot, joins `HEADLESS_BACKENDS` | `buildHeadlessArgv()` — `-p` is spread as the one-shot token with the prompt as the final distinct argv element, same shape as codex/copilot/goose; exit-code contract handled by the existing `runHeadlessTask()` unchanged | ✅ mechanically consistent with every other entry |
| 3 | Version alignment on the headless flip | `src/Dockerfile.contributor` on v4: **zero** agy/antigravity references — the contributor image does not ship the binary, so there is no image pin to drift from the "verified on 1.1.13" claim. The flip is host-only by construction (agy stays OUT of `K8S_HEADLESS_BACKENDS`, out of the Justfile k8s `HEADLESS_BACKENDS` list, and the PR documents the deliberate divergence in both places so nobody "resyncs" them). A host with an older agy lacking `-p` fails loudly (non-zero exit → `task_failed`), never silently | ✅ no misalignment possible; failure mode is loud |
| 4 | `--model` without `--effort` silently ignored; pairing added | `agyDefaultEffort = "low"` + the `--model %s --effort %s` pairing already exist in the hub-side launcher (`src/pkg/agent/manager.go` ~L1743-1751, L4899-4905) — the relay was the only launch surface missing it. `AGY_EFFORTS` whitelist correctly rejects codex-only `minimal`. Effort flag is gated on `modelFlag`/`modelArgs.length` so it never travels alone | ✅ relay now matches the hub launcher exactly |
| 5 | Staging path: `~/.antigravitycli` doesn't exist on 1.1.x; state lives at `~/.gemini/antigravity-cli` | Old recipe staged only the legacy path (silent no-op when absent). New code prefers the 1.1.x layout via the same `stage_copy` + `CLI_MOUNTS` pattern every other backend uses, keeps the legacy path as fallback, and mounts to the matching in-container path | ✅ matches how codex/goose/etc. stage |
| 6 | Launch-path security (the #3910 bar) | Headless path spawns via the existing `execFile(bin, args, {timeout, maxBuffer, cwd})` with **no env override** — child inherits the relay's process env (HTTPS_PROXY, NODE_EXTRA_CA_CERTS, gh-wrapper PATH) identically to claude/codex/goose; output still passes `redactTokens()` before leaving the host. Interactive path is the unchanged `buildLaunchCommand()` single source of truth | ✅ no bypass introduced |
| 7 | Picker honesty | Option's `data-host-install` (brew cask + interactive `agy` sign-in) matches the preflight's corrected install URL/brew hint; `AGENT_REASONING_EFFORT` env note matches what the relay actually consumes; `HOST_ONLY_BACKENDS` generalizes the exact fallback `other` had, and the mode flip switches the visible selector, not just the local var (pinned by test) | ✅ |
| 8 | Test tightening | `TestContributeK8sHeadlessCapability` previously grepped the whole page for `<backend>:1` — `HOST_ONLY_BACKENDS` would indeed have collided; scoping to the `K8S_HEADLESS_BACKENDS={...}` literal is the right fix, and the agy row keeps `false` with the corrected reason string. New `TestContributeAgyHostOnlySurface` pins option, install line, tile, forcing, note, and k8s exclusion; relay tests cover pairing, effort whitelist, and headless argv shape | ✅ |

One nit, not blocking: the k8s-note prose on the page still lists only "Claude Code, LiteLLM, Copilot, Codex" while `K8S_HEADLESS_BACKENDS` also carries watsonx and goose — but that drift predates this PR and is out of scope here.

DCO signed. The capability-vs-credential framing (headless on host, warn in k8s) is the correct call — a pod that starts and then fails an interactive OAuth would just crash-loop, exactly what the loud-refusal contract exists to prevent.

/lgtm
/approve

clubanderson (approved):
Verified per-claim against the repo — see the review comment above. LGTM.

PR #4008, opened 2026-08-17
🐛 fix(contribute): contribute-setup died with a bare "exit code 5" for anyone who owns no hives

Danathar:
## Symptom

`just contribute-setup <cli>` with `HIVE_HUB` unset:

```
HIVE_HUB not set — looking up your hives...

error: recipe `contribute-setup` failed with exit code 5
```

No message, no hive list, no way to tell what went wrong.

## Cause

`/api/saas/my-hives` answers with `"hives": null` for any account that owns no SaaS-hosted hive — the normal case for a contributor who only lends their CLI to someone else's hive. The filter was:

```jq
.hives[]? // .[] | "\(.id)|\(.name // .project_name)"
```

`.hives[]?` correctly yields nothing. But `//` then treats that empty result as "try the other branch" and iterates every **top-level key** of the response instead — `alerts`, `channel_targets`, `saas_quota`, `tracked_branches`, … Interpolating `.name` over a value that is an *array* makes jq fail:

```
jq: error (at <stdin>:0): Cannot index array with string "name"
$ echo "$MY_HIVES" | jq -r '.hives[]? // .[] | "\(.id)|\(.name // .project_name)"' >/dev/null; echo $?
5
```

The recipe runs under `set -euo pipefail`, so jq's status aborted the whole thing — **before** reaching the public-registry fallback on the very next lines, which works fine. Verified live: that fallback lists 12 online hives, including the one the contributor who hit this actually works on.

## Fix

Keep the intent, remove both sharp edges:

- iterate `.hives` only when it really is an array of objects, so a `null` (or any other shape) yields no rows instead of an error;
- never let jq's exit status abort the lookup — an empty result there is a valid answer meaning "fall through to the registry";
- same guards on the registry filter, whose shape drift would fail the same way;
- the dead-end message now also tells you that you can `export HIVE_HUB=wss://<hive>/contribute` directly.

Verified against the live endpoints: the my-hives step now yields nothing and the registry fallback lists the hives, exit 0.

I audited the Justfile's other `jq` calls — the rest are already guarded with `|| echo ""` or preceded by a `jq empty` validity check. This block was the only unguarded pair.

## Test

`src/pkg/dashboard/contribute_setup_hive_lookup_test.go` reads the two filters **out of the Justfile** rather than copying them, so it pins what actually ships:

- the `// .[]` fallback cannot come back, and neither jq call may be left unguarded (comment lines are skipped, since the recipe now quotes the broken filter to explain it);
- the my-hives filter stays silent on a realistic null-hives payload, while still listing owned hives — `name`, the `project_name` spelling, and an id-only fallback;
- the registry filter lists online entries only and survives a null `.hives`.

Skips the execution half when `jq` is absent; the shape assertions still run. Full `pkg/dashboard` package passes.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -197,7 +197,21 @@ contribute-setup backend="claude": check-version (contribute-check-backend backe
       if [[ -n "$_TOKEN" ]]; then
         MY_HIVES=$(curl -sf -H "Authorization: Bearer ${_TOKEN}" "https://hive.kubestellar.io/api/saas/my-hives" 2>/dev/null || echo "")
         if [[ -n "$MY_HIVES" ]]; then
-          HIVE_LIST=$(echo "$MY_HIVES" | jq -r '.hives[]? // .[] | "\(.id)|\(.name // .project_name)"' 2>/dev/null)
+          # /api/saas/my-hives answers with "hives": null for an account that
+          # owns no SaaS-hosted hive — the normal case for a contributor who
+          # only lends their CLI to someone else's hive. The old filter treated
+          # that as "try something else" via `// .[]`, which iterates every
+          # TOP-LEVEL KEY of the response (alerts, channel_targets, …) and then
+          # interpolates .name over values that are arrays. jq exits 5 on that,
+          # and because the recipe runs under `set -euo pipefail` the whole
+          # setup died with a bare "recipe failed with exit code 5" — never
+          # reaching the public-registry fallback directly below, which would
+          # have listed the hives fine.
+          #
+          # So: iterate .hives ONLY when it really is an array of objects, and
+          # never let jq's exit status abort the lookup. An empty result here
+          # is a valid answer that means "fall through to the registry".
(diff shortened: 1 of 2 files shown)


clubanderson:
/lgtm
/approve

Reviewed: null-guard fix (`.hives // empty | select(type=="array")`) correctly handles `hives: null`, empty array, and populated array; same guard applied to the registry fallback filter. UX improved — no-hives dead end now tells the contributor to `export HIVE_HUB=wss://<hive>/contribute`. `HIVE_HUB`-set path is untouched (positive control: block is gated by `if [[ "{{hive_hub}}" == "wss://hive.kubestellar.io/contribute" ]]`). The +152 lines are a Justfile comment block explaining the bug plus a new Go test (`contribute_setup_hive_lookup_test.go`) that reads the two jq filters directly out of the Justfile and exercises null hives, populated hives (name/project_name/id-only fallback), and the online-only registry filter — verified passing locally. Single commit, DCO signed-off by Douglas Baggett.

PR #4026, opened 2026-08-17
🐛 fix(relay): a finished agy turn could pin the relay to "working" forever, and nothing reclaimed the task

Danathar:
## Symptom

Observed on a live contributor relay. agy had finished its turn, printed its `no_work_needed` verdict, and was sitting at its idle prompt — but the relay kept sending `task_progress status:'working'`. The hub held the task in-progress, offered no further work, and the only way out was the contributor pressing Ctrl-C.

## Cause 1 — agy's `isWorking` matched prose anywhere on screen

```js
isWorking = /Running|Searching|Reading|Writing|Editing/i.test(text);   // whole pane
```

agy narrates in plain English inside its transcript, and those lines stay on screen after the turn ends. Measured against the actual wedged pane:

```
hasIdlePrompt   = true
isWorking (whole pane)   = true
isWorking (15-line tail) = false
matching lines: 1
   > "I am running the pkg/agent tests with the shortened temp directory path to verify they now pass locally as well"
```

One line of narration **from the previous task**, sitting 36 rows above the bottom, where it would never scroll off on its own.

The `claude` branch already scopes this exact check to `lastLines`; agy now does too. The marker **set** is deliberately unchanged — narrowing which verbs count needs verification against a live agy turn, and getting that wrong means reporting a busy agent as idle, which is the worse failure.

## Cause 2 — nothing could reclaim the task

`contribute_ws.go` documents the lease TTL as the guard for "a wedged-but-connected worker". It can't fire here: the relay renews that lease on every progress report, and the wedge *is* the relay believing it's working.

```js
send({ type:'task_progress', …, status:'working' })   // relay, every tick
contributor.lastLeaseRenew = time.Now()               // hub, on every one
```

So the guard only catches a relay that goes **silent**. A relay stuck in a false "working" belief renews forever.

`progressTick` now fingerprints the pane. If it hasn't changed at all for `PANE_STALL_TIMEOUT_MS` (default 20 min, overridable with `HIVE_PANE_STALL_TIMEOUT_MS`), the relay stops asserting progress it cannot substantiate and hands the task back as an `environment` failure — the honest kind, since a frozen pane says nothing about whether the work got done. The hub requeues it through its normal release path.

Deliberately generous window: a real agent can sit on one silent command (long test suite, slow clone) for many minutes. An empty capture never trips it — that's a missing pane, not a stalled agent, and other paths already handle that.

## Scope note

Several other backends (`gemini`, `goose`, `codex`, `pi`) scan the whole pane for similar bare verbs and can wedge the same way. I did not touch them here — I only have a verified reproduction for agy, and the stall backstop protects all of them regardless.

## Tests

- agy at its idle prompt with stale narration on screen classifies `IDLE_COMPLETE`; the fixture reproduces the real pane's geometry (narration outside the tail), so re-widening the window fails the test.
- agy with fresh activity in the tail still classifies `WORKING`.
- `paneStalled` trips only after the timeout, resets on any change, and never trips on an empty capture.
- A stalled pane produces exactly one `task_failed` with `failure_kind: "environment"` and the relay lets go of the task instead of renewing its lease.

84/84 relay tests pass; the Go tests that read this file pass.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -1227,9 +1227,25 @@ function classifyTmuxPane(text) {
     hasCompletionMarker = /completed|done|finished|tokens\)|\d+\.\d+%/i.test(text);
     isWorking = /Reading|Writing|Bash|Editing|thinking|running/i.test(text);
   } else if (BACKEND === 'agy') {
+    // Scope the activity check to the TAIL, exactly as the claude branch above
+    // does. agy narrates in plain English inside the transcript ("I am running
+    // the pkg/agent tests…", "Analyzing…"), and those lines stay on screen after
+    // the turn ends. A whole-pane, case-insensitive scan for bare verbs
+    // therefore reads a FINISHED turn as still working — forever, since the
+    // stale line never scrolls off on its own. Observed live: a pane with
+    // hasIdlePrompt=true was pinned to WORKING by a single narration line left
+    // over from the PREVIOUS task, so the relay never reported completion and
+    // kept renewing the hub's task lease; the contributor had to Ctrl-C.
+    //
+    // The marker SET is deliberately unchanged — only the window it looks at.
+    // Narrowing which verbs count would need a live agy turn to verify against,
+    // and getting that wrong would be the opposite (and worse) bug: reporting a
+    // busy agent as idle. The stall backstop in progressTick() covers whatever
+    // this still misses.
+    const agyTail = text.split('\n').slice(-15).join('\n');
     hasIdlePrompt = /\? for shortcuts/.test(text);
     hasCompletionMarker = true;
-    isWorking = /Running|Searching|Reading|Writing|Editing/i.test(text);
+    isWorking = /Running|Searching|Reading|Writing|Editing/i.test(agyTail);
   } else {
     hasIdlePrompt = />\s*$|\$\s*$/.test(text);
     hasCompletionMarker = /completed|done|finished/i.test(text);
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed both halves against origin/v4 and the live symptom. LGTM.

**Detector fix.** The agy branch now scopes `isWorking` to a 15-line tail — byte-for-byte the same idiom the `claude` branch already uses (`lastLines = text.split('\n').slice(-15)`), and the same window as `TMUX_TAIL_LINES`. Since `checkTmuxPaneState` classifies the full visible pane, the wedged narration line ("I am running the pkg/agent tests…", 36 rows above the bottom) really was in scope before and really is out of scope now. Keeping the marker *set* unchanged and only narrowing the window is the right call — narrowing verbs without a live-turn verification would risk the inverse (and worse) bug of reading a busy agent as idle. The wedged-pane fixture reproduces the real geometry (narration outside the tail, verdict + idle chrome inside it), so re-widening the window fails the test; the positive control pins that fresh tail activity still classifies WORKING even with a stale `HIVE_VERDICT` line higher up.

**Reclaim backstop.** Correct diagnosis that the hub's lease TTL (`wsTaskTimeout`/cleanupLoop) is structurally unable to catch this — the wedge *is* the renewal. The relay-side stall detector is activity-based, not wall-clock-based: a genuinely long turn keeps redrawing the pane (spinner/elapsed counters), so it never accumulates 20 minutes of byte-identical captures; `MAX_TASK_DURATION_MS` remains the separate wall-clock bound. The clock resets per task (`startProgressReporting`), on any pane change, and an empty capture never trips it. Verified the hub-side interaction: a non-permanent `task_failed` lands in the short `failedTaskCooldownMinutes` (10 min) path, so a repeatedly-stalling task cannot hot-loop the offer pool, and three consecutive failures cross `consecutiveFailureQuarantineThreshold` into the 6 h quarantine — a persistently frozen environment self-parks instead of looping forever. `failure_kind: environment` is the honest verdict here and is advisory-only on the hub, so it can't distort cooldown routing.

**Scope note** on the other whole-pane backends (gemini/goose/codex/pi) is reasonable: no verified repro, and the stall backstop now covers them as a class.

DCO signed, tests pin both halves plus the empty-capture and reset edges. Merging.

/lgtm
/approve

PR #4039, opened 2026-08-18
🐛 fix(relay): the CLI-liveness probe could never see a dead CLI, so task prompts were typed into a bare shell

Danathar:
> **Stacks on #4026.** The first commit here is that PR's; review the second, or merge #4026 first and it drops out of the diff. They touch the same file, which is why they are stacked rather than parallel.

## Symptom

Observed live on an agy contributor: the CLI exited, the relay never noticed, and the hub's task prompt was **typed into the shell** the pane had fallen back to — the terminal showed the prompt executing as shell commands, its first characters eaten by readline:

```
dbaggett at rig in .
u are a contributor to the kubestellar/hive hive. Work on issue kubestellar/hive#4030 …
```

This is precisely the wedge the send gate exists to prevent (`#2203, bug 2`). The gate was intact — everything upstream of it was blind.

## Cause

```js
const cliAlive = procs.includes(BACKEND) || procs.includes('claude') || procs.includes('copilot')
              || procs.includes('bob') || procs.includes('codex') || procs.includes('goose') || procs.includes('pi');
```

A substring match over the **whole process table**. Two independent defects, both reproduced on the affected machine:

1. **The relay's own machinery carries the backend's name.** For agy, `just contribute-hive agy local` (the launcher) and `tmux attach -t hive-agy-5b4f` (the session) both contain `agy`:
   ```
   1805845 just contribute-hive agy local
   1806321 tmux attach -t hive-agy-5b4f
   ```
   The probe was *structurally incapable* of reporting a death for this backend.
2. **The other CLI names were OR'd in unconditionally**, whatever `BACKEND` was. Any contributor with Claude Code running — most of them — reported a live CLI for every backend, forever.

With the probe stuck true, `relaunchCLI()` never ran, so `cliReady` — a latch, cleared only by a relaunch — stayed true, and the gate waved the prompt through into bash.

## Fix

Either half alone would have stopped it; both are here.

- **Liveness asks the pane what it is running** (`tmux display-message -p '#{pane_current_command}'`), which nothing outside the pane can influence. A death requires `CLI_GONE_CONFIRMATIONS` consecutive shell readings, so a foreground tool call is not mistaken for one — the expensive error, since it restarts a working CLI. The pane **text** is deliberately not consulted: a dead CLI leaves its last frame on screen, ready-chrome and all, so requiring that chrome to be gone would re-introduce the same blindness.
- **The send gate re-confirms against the live pane** rather than trusting the latch. A pane at a shell prompt queues the prompt, drops the stale latch, and relaunches the CLI so the queued prompt has somewhere to flush to.

Also: **a shell pane can no longer reach the completion check.** Before the death is confirmed, the dead CLI's last frame classified as "agent idle" and would report a task nobody performed as `completed`. `bob` is exempt — it exits at the end of every turn, so for bob a shell pane *is* the completion signal.

## Tests

- one shell reading is never a death; consecutive readings are;
- a pane running the CLI never accumulates toward a death;
- **a task prompt is never typed into a pane running a shell** — it is queued, the stale latch is dropped, and the CLI is relaunched;
- crash-detection tests now tick twice per assignment, the honest consequence of requiring confirmation.

87/87 relay tests pass; the Go tests that read this file pass.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -672,6 +672,83 @@ function recoverWedgedShell() {
 // dismissal can look at the SAME text without capturing twice, and so the
 // dismissal can see WHICH prompt is on screen rather than re-deriving it from
 // a state enum that has already thrown that detail away.
+// Shell names that mean the pane fell back to a prompt — i.e. whatever the
+// relay launched is no longer the pane's foreground program.
+const PANE_SHELL_COMMANDS = new Set(['bash', 'sh', 'zsh', 'fish', 'dash', 'ksh', 'ash', 'tcsh', 'csh']);
+
+// How many consecutive shell readings (one per progress tick) are required
+// before the CLI is declared gone. One is not enough: a tool call can briefly
+// put a shell in the pane's foreground while the CLI is very much alive.
+const CLI_GONE_CONFIRMATIONS = 2;
+let consecutiveShellReadings = 0;
+
+// paneForegroundCommand asks tmux what the pane is actually RUNNING. Empty when
+// tmux cannot answer (session gone, tmux missing) — an unknown, never a death.
+function paneForegroundCommand() {
+  try {
+    return execSync(
+      `tmux display-message -p -t ${TMUX_SESSION} '#{pane_current_command}' 2>/dev/null`,
+      { encoding: 'utf8', timeout: 15000 }
+    ).toString().trim();
+  } catch (_) {
+    return '';
+  }
+}
+
+// cliProcessLooksGone reports whether the agent CLI has left the pane.
+//
+// It replaces a substring scan of the WHOLE process table:
+//
+//   procs.includes(BACKEND) || procs.includes('claude') || procs.includes('copilot') || …
+//
+// which could not do this job. Two independent defects, both observed live:
+//
+//  1. The relay's own machinery carries the backend's name. For agy the
+//     launcher (`just contribute-hive agy local`) and the tmux session itself
+//     (`tmux attach -t hive-agy-5b4f`) both contain "agy", so the probe was
+//     pinned alive no matter what happened to the CLI.
+//  2. The other CLI names were OR'd in unconditionally, whatever BACKEND was.
+//     Any contributor with Claude Code running — i.e. most of them — reported
+//     a live CLI for every backend, forever.
+//
+// With the probe stuck true the relay never relaunched a dead CLI, cliReady
+// stayed latched, and task prompts were typed into a bare shell: exactly the
+// #2203 bug-2 wedge the send gate exists to prevent.
+//
+// The pane's own foreground command answers the real question, and it cannot be
+// confused by anything outside the pane. Two consecutive readings are required
+// so that a tool call which briefly fronts a shell does not read as a death —
+// the expensive mistake, since it restarts a CLI that is working. A CLI that
+// really exited leaves the pane at a prompt permanently, so it still trips on
+// the following tick; the stall backstop in progressTick is the second net.
+//
+// Note the pane TEXT is deliberately not consulted: a CLI that dies leaves its
+// last frame on screen, ready-chrome and all, so requiring that chrome to be
+// gone would re-introduce exactly the blindness this replaces.
+function probeCLIPresence() {
+  const fg = paneForegroundCommand();
+  const isShell = !!fg && PANE_SHELL_COMMANDS.has(fg);
+  if (!isShell) {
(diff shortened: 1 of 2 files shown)


clubanderson:
## Review: #4039 — CLI-liveness probe fix

Rebased onto current `v4` (post-#4026 merge `9f9a65f2`) — the stacked `96acf6a6` commit dropped out cleanly as a no-op during `git rebase`, leaving a single commit `c2e910e5` with the liveness fix only (2 files, +189/-9 vs. previous branch diff of +364/-10 including the duplicate).

**1. Detection mechanism**: pane-PID/foreground-process based, not chrome-text based. `paneForegroundCommand()` reads `tmux display-message -p '#{pane_current_command}'` — the pane's actual foreground process, which nothing outside the pane (launcher script, tmux session name, other running CLIs) can spoof. This replaces the old whole-process-table substring scan that was structurally blind for `agy` (its own launcher/session name contain "agy") and false-positive-prone for every backend when any `claude`/`copilot`/etc. process was running elsewhere on the box.

Robustness against long silent turns: a death requires `CLI_GONE_CONFIRMATIONS = 2` consecutive shell readings (`probeCLIPresence`), so one foreground shell glimpse (e.g., a brief subprocess/tool invocation) can't trip it — confirmed by test `a pane running the CLI is never reported gone, and clears the count`. This composes correctly with #4026's activity-based stall backstop: a working-but-quiet CLI still shows as the CLI in `pane_current_command` (not a shell), so it never enters the liveness-death path at all; it only ever reaches the *stall* backstop, which is unaffected by this change (`resetPaneStallClock` now also resets the shell-reading counter on new tasks, preventing stale counts from carrying over).

Pane text/chrome is deliberately NOT consulted for death detection (explicit comment) — a dead CLI's last frame stays on screen with ready-chrome intact, so text-based detection would reintroduce the exact blindness being fixed.

**2. Action on detection + send-point guard**: On confirmed death (`!cliAlive`), `progressTick` fails the current task with `{ kind: 'environment' }` (matching #4026's verdict pattern) and relaunches the CLI, with the existing restart-count/backoff/give-up ladder unchanged. Additionally, a pane sitting at a shell (even before 2-tick confirmation) is now excluded from reaching the completion-check path, so a dead CLI's frozen "idle" chrome can no longer be misread as `task_complete`.

Critically, the guard is enforced **at the send point**, not just in the probe loop: `tmuxSendKeys()` — the sole function that does the literal `-l` keystroke send of task-prompt text into the pane (verified: it's the only such call site; all other `send-keys` calls are control sequences like Escape/C-c/Enter or the launch command itself) — calls `paneIsRunningShell()` synchronously immediately before typing, independent of and in addition to the `cliReady` latch check. This uses a single reading (not the 2-reading confirmation), since "the cost of waiting a tick when we are wrong is nil" but typing into a shell is never correct. This closes the race the task called out: if the CLI dies between a probe tick and the send, the send-time check still catches it, drops the stale `cliReady` latch, queues the prompt, and relaunches the CLI so the queued prompt has somewhere to flush to.

**3. Tests** (in `bin/contributor-relay.test.js`):
- `a shell in the pane is only a death after consecutive confirmations` — dead-CLI detection requires 2 consecutive readings
- `a pane running the CLI is never reported gone, and clears the count` — positive control: a live/quiet CLI never accumulates toward death, however long it runs (no-false-positive-on-quiet-turn)
- `a task prompt is never typed into a pane that is running a shell` — send-point guard + shell-injection replay: stale `cliReady=true` latch (reproducing the incident precondition) + shell pane; asserts the prompt text never reaches any tmux send, the latch is dropped, and the CLI is relaunched
- Existing crash-detection/give-up tests (`crash restarts are capped...`, `a reassignment of a given-up task is rejected...`, `after give-up a DIFFERENT task is still accepted`) updated to tick twice per assignment, consistent with the new confirmation requirement
- Full suite: 87/87 passing locally (`node --test bin/contributor-relay.test.js`)

**4. DCO**: present (`Signed-off-by: Douglas Baggett <doug.baggett@gmail.com>`) — `dco` check passes.

**Mechanics**: PR was DIRTY/CONFLICTING before rebase because it still carried the un-squashed duplicate of #4026's now-merged commit; `gh api pulls/.../update-branch` failed for the same reason (real conflict, not just "needs update"). Rebased manually onto `origin/v4` — duplicate commit auto-skipped by `git rebase --skipped-cherry-picks`, single commit remains, force-pushed to fork branch. Fork PR gated via `ci-gate-pr-4039` label to allow CI to run.

Verdict: **approve**, pending green CI on the current (rebased) head.

Issue #4037, opened 2026-08-18
ACMM packs seed explicit governor thresholds, so threshold scaling never engages on pack-applied hives


clubanderson:
## Summary

PR #3898 (threshold scaling by repo count, refs #3498) ships with a documented limitation that deserves its own design decision: **ACMM packs seed explicit governor thresholds on apply, and under the explicit-always-wins rule, threshold scaling never engages on a pack-applied hive.**

## Mechanism

- `pkg/config/packs/level-*.yaml` all carry `governor.thresholds` (L4–L6: `surge: 10, busy: 5, quiet: 2`; L3: `15/10/3`).
- `pkg/dashboard/api_packs.go` (threshold seeding around line 282) writes those into `governor.modes.<mode>.threshold` when a level is applied.
- Those land in config as **explicit** values, and nothing distinguishes a pack-seeded default from a number an operator typed.
- `config.EffectiveThreshold` (added in #3898) returns explicit thresholds verbatim, never scaled — the single most important guarantee of #3498, since scaling a hand-tuner's `surge: 300` on a 39-repo hive would produce 11700.

Net effect: scaling only engages on hives that never applied an ACMM level, or that manually cleared the seeded thresholds. Applying a pack is the normal path — and plausibly includes the very 39-repo fleet shape that motivated #3498, so the fix may not reach the hive that reported the problem.

## Why #3898 did not work around it

All the available in-PR workarounds are worse than the gap:

- Scaling explicit values too breaks every hand-tuned hive (the 300 → 11700 case).
- Dropping thresholds from the pack YAMLs loses real per-level tuning (L3 differs from L4–L6).
- Baking scaled values at apply time goes stale the moment the repo list changes — the exact drift #3498 exists to eliminate.

The manual workaround (clear the thresholds you want scaled) is documented in `docs/governor-thresholds.md`.

## Design question

Fixing this properly means the config layer recording **provenance** — pack-seeded vs operator-set — so explicit-wins applies only to the latter. Two candidate directions:

1. **Packs stop seeding absolute thresholds** and pack-applied hives inherit the scaled defaults (possibly with the pack contributing a per-repo *base* instead of an absolute value).
2. **Packs carry per-repo bases explicitly** (a new pack field), and apply-time writes them into a slot that scales, keeping the existing absolute `thresholds` for levels that genuinely want absolutes.

Either way this is a config-layering decision, not something to bolt on inside the governor.

Refs #3898, #3498.

PR #4046, opened 2026-08-18
🐛 fix(contribute): pin the CLI's working directory — a tmux server with a deleted cwd killed every agy task

Danathar:
## Symptom

An agy contributor's CLI died a few seconds after **every** task delivery. It started fine, sat at its prompt, the relay logged `CLI ready` then `Task prompt sent to CLI` — and then it was gone. No crash, no log, no message. The pane fell back to a shell, and the next task prompt got typed into that shell.

## What it wasn't

Each ruled out by direct test on the affected machine:

| suspect | result |
|---|---|
| the relay's `Escape` / `C-a` / `C-k` before each prompt | agy survives |
| the literal 1100-char prompt + `ENTER_COUNT=3` | agy survives |
| the repo as cwd | agy survives |
| task content (#3848 / #4037 / #4041) | dies on all |
| pane environment | 109 vars, identical to a working hand-run |
| recent relay changes (#4026, #4039) | dies with the pre-#4026 relay too |
| account / quota | `agy -p` returns `ok`, exit 0 |
| coredump / OOM / signal | none — a wrapper captured `AGY_EXIT=2` |

Driving agy by hand through the relay's exact key sequence, with the real prompt, in the same directory, it completed an entire hive task and stayed up.

## What it was

```
pane_current_path = '.../hive/kubestellar/hive/v2/pkg/agent (deleted)'
```

The pane's working directory did not exist. That path belonged to a nested clone and was orphaned when the repo renamed `v2/` → `src/`. The tmux **server** had been holding it since a Go test created the server hours earlier, so every pane it forked started there. The shell announced it each time —

```
shell-init: error retrieving current directory: getcwd: cannot access parent directories
```

— and `agy`, which needs a resolvable cwd, exits `2` about thirty seconds in. `claude`, `codex` and `goose` tolerate a dead cwd, so the same server looked perfectly healthy for them, which is why this read as agy-specific.

The reason my hand-runs always worked: I typed `cd <repo> && agy …`. The Justfile sends only `$CMD $PERM_FLAG`.

## Fix

- **The launch command cds into the repo first.** This is the load-bearing change.
- **`-c "$PWD"` on `new-session` as well** — correct on a healthy server, but *not* sufficient alone. Verified: on a server whose own cwd is gone, `tmux new-session -c <valid path>` **still** forks the pane into the deleted directory.
- **`relaunchCLI()` gets the same prefix.** It rebuilds the launch line on every restart (memory-cleanup restarts, crash recovery, a stale readiness latch), so without it the first relaunch silently undoes what the Justfile pinned.
- **The recipe now warns on a poisoned server** instead of letting a backend die unexplained: it checks `pane_current_path` after creating the session and points at `tmux kill-server`, which is the only thing that really clears it.

## Tests

- the relaunch command carries the `cd`, and reaches tmux that way;
- a Go test reads the **Justfile itself** (not a restatement) and pins the `cd`, the `-c`, and the warning, so none of them can regress quietly.

89/89 relay tests, full `pkg/dashboard` package green.

## Docs

A troubleshooting entry keyed on the symptom — "backend dies seconds after every task" — because the failure gives you nothing to search for: no log line, no exit message, just a prompt that quietly shows `.` as your directory.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -489,10 +489,36 @@ contribute-hive backend="" mode="docker": check-version
         fi
       fi
 
-      # Create tmux session with the CLI
+      # Create tmux session with the CLI.
+      #
+      # The launch command CDs into the repo first. That is not belt-and-braces:
+      # a long-lived tmux server keeps its own working directory, and when that
+      # directory is deleted (here: a nested clone's v2/pkg/agent, orphaned when
+      # the repo renamed v2/ -> src/) every pane the server forks inherits the
+      # dead cwd. The pane's shell says so — "shell-init: error retrieving
+      # current directory" — and a backend that requires a resolvable cwd then
+      # dies seconds after its first task. agy exits 2 that way; claude, codex
+      # and goose happen to tolerate it, which is why this hid for so long.
+      #
+      # -c is NOT sufficient on its own: on a server whose own cwd is gone,
+      # `tmux new-session -c <valid path>` still forks the pane into the deleted
+      # directory (verified against tmux on Fedora 44). It is passed anyway
+      # because it IS correct on a healthy server; the cd is what carries the
+      # fix.
       tmux kill-session -t "$TMUX_SESSION" 2>/dev/null || true
-      tmux new-session -d -s "$TMUX_SESSION" -x 200 -y 50
-      tmux send-keys -t "$TMUX_SESSION" "${LITELLM_ENV:+$LITELLM_ENV }$CMD $PERM_FLAG" Enter
+      tmux new-session -d -s "$TMUX_SESSION" -x 200 -y 50 -c "$PWD"
+      tmux send-keys -t "$TMUX_SESSION" "cd $(printf %q "$PWD") && ${LITELLM_ENV:+$LITELLM_ENV }$CMD $PERM_FLAG" Enter
+
+      # Surface a poisoned tmux server rather than letting the backend die a
+      # silent, unexplained death 30 seconds into its first task.
+      PANE_PATH="$(tmux display-message -p -t "$TMUX_SESSION" '#{pane_current_path}' 2>/dev/null || echo '')"
+      case "$PANE_PATH" in
+        *"(deleted)"|"")
+          echo "WARNING: this tmux server's working directory no longer exists (pane reports: ${PANE_PATH:-unknown})." >&2
+          echo "         The cd above works around it, but panes you open by hand will start in a dead directory." >&2
+          echo "         Fix it for good with: tmux kill-server   (ends all tmux sessions, then rerun this recipe)" >&2
+          ;;
+      esac
 
       # Start the relay
       export HIVE_AGENT_SESSION="$TMUX_SESSION"
--- a/bin/contributor-agent.sh
+++ b/bin/contributor-agent.sh
@@ -456,6 +456,21 @@ if [[ "$CONTRIBUTOR_MODE" == "interactive" ]]; then
   tmux new-session -d -s "$TMUX_SESSION" -c "$HIVE_WORKSPACE_DIR" -x 200 -y 50
 fi
 
+# kubestellar/hive#4046: a long-lived tmux SERVER (this container can run one
+# for its whole lifetime) keeps its own working directory, and every pane it
+# forks afterward inherits it — even a pane started with an explicit, valid
+# `-c`, if the server's own cwd is already gone. That is exactly the shape of
+# directory HIVE_WORKSPACE_DIR is: agents clone into subdirectories of it
+# (contribute_ws.go's assignment prompt), so pinning the launch's `cd` at
+# HIVE_WORKSPACE_DIR itself would re-create the trap the moment that directory
+# is ever removed and recreated out from under a still-running server. `cd`
+# into $HOME instead — a directory nothing in the task lifecycle deletes or
+# recreates — so the CLI always starts somewhere resolvable regardless of what
+# happens to the workspace subtree underneath it; the CLI's own per-task `cd`
+# into $HIVE_WORKSPACE_DIR/<repo> (per the assignment prompt) is unchanged.
+HIVE_AGENT_CWD="${HOME}"
+mkdir -p "$HIVE_AGENT_CWD"
+
 # Start the relay in the background
 echo "Starting ClankeR relay connection to hub..."
 node "${SCRIPT_DIR}/contributor-relay.sh" &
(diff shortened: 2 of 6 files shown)


clubanderson:
## Review

**Fix mechanism.** The tmux *server* (not just the pane) can be holding a working directory that was later deleted (here: a nested clone's `v2/pkg/agent`, orphaned by the `v2/` → `src/` rename). Every pane the server forks after that inherits the dead cwd. `agy` needs a resolvable cwd and exits 2 ~30s in; `claude`/`codex`/`goose` tolerate it, which is why this looked agy-specific. The PR's own repro is convincing: `tmux new-session -c <valid path>` on a poisoned server *still* forks the pane into the deleted directory — so `-c` is necessary-but-not-sufficient, and the actual fix is a `cd <dir> &&` prefix baked into the literal `send-keys` launch command, issued fresh on every launch/relaunch.

**Spawn-site coverage — INCOMPLETE.** The fix covers:
- `Justfile` `contribute-hive` **local mode only** (the `_MODE == "local"` branch) — adds the `cd`, keeps `-c` as defense-in-depth, and adds a poisoned-server warning.
- `contributor-relay.sh` `relaunchCLI()` (shared by both modes) — via the new `launchCommandWithCwd()`.

It does **not** cover `bin/contributor-agent.sh`, the **container/docker-mode entrypoint** — which is the *default* mode (`contribute-hive backend="" mode="docker"`). That script's tmux launch (line ~456 `new-session -c "$HIVE_WORKSPACE_DIR"`, line ~531 `send-keys "$CMD $PERM_FLAG..."`) has no `cd` prefix at all and relies solely on `-c`, which this PR's own investigation demonstrated is insufficient once the server is already poisoned. Any long-lived container whose tmux server outlives a directory removal (e.g. a workspace wipe/rebuild without a fresh tmux server) hits the identical failure, uncovered.

Also worth flagging: the new Go test (`TestContributeHivePinsPaneWorkingDirectory`) reads *only* the `contribute-hive` block of the Justfile via a hand-rolled string search between `"Create tmux session with the CLI"` and `"# Start the relay"` — it can't and doesn't assert anything about `contributor-agent.sh`, so the gap won't be caught by this PR's own regression net either.

Recommend a follow-up issue (or scope-in here) for `contributor-agent.sh`'s launch line before calling the class of bug closed.

**Pinned-directory lifecycle — sound for the two sites it touches.** For local mode, the pinned dir is `$PWD` at the point the Justfile recipe runs / `process.cwd()` inside the relay — i.e., the **hive repo root itself**, which the recipe already requires (`ERROR: Run from the hive repo root`). That directory is not part of the per-task workspace lifecycle: tasks clone into `$HIVE_WORKSPACE_DIR/<repo>` (see `buildTaskPrompt` in `contribute_ws.go`), a directory this PR's pinned path never touches, so nothing in normal task cycling deletes or recreates the pinned cwd. Good — pinning to the repo root sidesteps the exact trap the task description warned about (pinning to a workspace dir the task lifecycle deletes would re-break this).

That said, this is precisely why the `contributor-agent.sh` gap matters more than it might first appear: that entrypoint's tmux session is rooted at `-c "$HIVE_WORKSPACE_DIR"` — i.e., the *per-task* workspace directory, by design (the comment there is explicit: "the tmux session itself is started rooted there via -c so any 'cd ~/workspace' step is redundant, not required"). If that assumption is ever violated (workspace dir recreated/rebound without a matching tmux restart), the container path has no `cd` fallback at all, and would be *more* exposed to this failure mode than the local-mode path this PR fixes.

**Interplay with #4039 liveness guard — clean, no weakening.** Diffed `bin/contributor-relay.sh` and `bin/contributor-relay.test.js` against `origin/v4`: no lines removed or modified in `checkTmuxPaneState`, `cliReady` gating, `CLI_GONE_CONFIRMATIONS`, or any existing liveness test. The PR is purely additive there (`launchCommandWithCwd()` + 2 new tests + one field on the test-mode export surface). The shell-falls-back-to-bash detection this guards against stays fully intact as defense-in-depth if a cwd-death somehow still occurs.

**Tests.**
- `a relaunch cds into the relay cwd before starting the CLI` — unit-tests `launchCommandWithCwd()` shape and that it targets the relay's own `process.cwd()`.
- `relaunchCLI() sends the cd-prefixed command to tmux` — asserts the actual `tmux send-keys` payload carries the `cd`.
- `TestContributeHivePinsPaneWorkingDirectory` (Go) — reads the literal Justfile text and asserts the `send-keys` line contains `cd `, that `new-session` still passes `-c`, and that the recipe checks `pane_current_path` and mentions `tmux kill-server`.
- No dedicated "deleted-cwd replay" test (e.g., actually creating and removing a directory and asserting the launched command still runs) — coverage is textual/shape-based, not behavioral, for both the JS and Go tests. Acceptable given tmux isn't mockable here, but worth naming as a gap.
- No standalone "positive control: normal task cycling still works" test was added, but the pre-existing 87 tests (including `relaunchCLI() leaves cliReady false until readiness is confirmed`, the `--model` flag relaunch tests) continue to pass and exercise the surrounding relaunch path unchanged, so regression coverage for normal cycling is inherited rather than new.

**DCO.** Commit is signed off (`Signed-off-by: Douglas Baggett <doug.baggett@gmail.com>`); `dco` check green.

**Sequencing vs #4047.** #4047 (gh-wrapper label-injection fix) touches only `bin/gh-wrapper.sh` and `bin/test_gh_wrapper_gates.sh` — no file overlap with this PR's Justfile/contributor-relay.sh/docs/test changes. No update-branch/sequencing action needed.

**Verdict:** Fix mechanism and pinned-directory choice are correct and well-reasoned for the paths touched. Spawn-site coverage is **incomplete** — `bin/contributor-agent.sh` (default docker/container mode) still launches via `-c` alone with no `cd` fallback, which is exactly the insufficient mitigation this PR's own diagnosis rules out. Recommend either scoping `contributor-agent.sh` into this PR or filing an immediate follow-up before treating the bug class as closed. Not blocking merge of this fix as far as it goes (it doesn't regress anything and closes the confirmed local-mode failure), but the container-mode gap should not be lost.

clubanderson:
## Gap closed: `bin/contributor-agent.sh` (default docker/container mode)

Per the coverage gap flagged above, pushed a maintainer-edit commit (`43067dd`, DCO co-signed) that mirrors this PR's own mechanism onto the missing spawn site:

- **`bin/contributor-agent.sh`** (the container entrypoint, default `mode="docker"`): the `tmux send-keys` launch line now `cd`s into a directory before starting `$CMD`, exactly as this PR's Justfile fix does. `-c` on `new-session` is kept as defense-in-depth, unchanged.
- **cd target reasoning:** this entrypoint pins `-c` to `$HIVE_WORKSPACE_DIR` — the *per-task* directory agents clone repos into (`contribute_ws.go`'s assignment prompt tells the agent to `gh repo fork ... $HIVE_WORKSPACE_DIR/<repo>`). Pinning the launch's `cd` there too would reintroduce this exact bug the moment that directory is ever removed and recreated under a still-running tmux server. Instead the `cd` targets **`$HOME`** — a directory nothing in the task lifecycle deletes or recreates — with `mkdir -p` as belt-and-braces. The CLI's own per-task `cd` into `$HIVE_WORKSPACE_DIR/<repo>` is untouched.
- **Test:** extended `src/pkg/dashboard/contribute_pane_cwd_test.go` with `TestContributorAgentPinsPaneWorkingDirectory`, reading `bin/contributor-agent.sh` the same way the existing test reads the Justfile (source-of-truth, not a restatement) — asserting the launch line `cd`s, and explicitly asserting it does **not** `cd` into `$HIVE_WORKSPACE_DIR`, so a future edit that "helpfully" pins the launch to the workspace dir would fail the test rather than reintroduce the trap silently.

Both spawn sites (local-mode Justfile + relay relaunch, and container-mode entrypoint) are now covered by the same fix pattern and the same style of regression test. Credit to the original PR for the correct diagnosis and mechanism — this only ports it to the site it missed.

clubanderson:
Merged at `d811360` — both spawn sites (Justfile local-mode + relay relaunch, and the container-mode `bin/contributor-agent.sh` entrypoint added above) now pin the CLI's working directory. Thanks @Danathar for the diagnosis and fix.

PR #4070, opened 2026-08-18
fix(apiproxy): reject unauthenticated host-key requests

### LINKED ISSUE #4068: [sec-check] apiproxy injects host API key for unauthenticated requests — any agent on localhost can bill to operator
## Security Finding

**Severity**: medium
**Type**: unsafe-pattern / privilege-escalation

In `src/pkg/apiproxy/proxy.go`, the `director()` function silently injects the operator's host API key (e.g. `ANTHROPIC_API_KEY`) into any request that does not already carry a valid `Authorization` or `X-Api-Key` header:

```go
// proxy.go lines 101-102
if p.apiKey != "" && !isValidAuth(req) {
    req.Header.Set("Authorization", "Bearer "+p.apiKey)
}
```

There is **no access-control gate in `ServeHTTP()`** — unauthenticated requests are never rejected with 401/403; they pass straight through with the injected key. The startup warning in `cmd/apiproxy/main.go` even acknowledges this explicitly:

> "Unauthenticated callers will be granted the host Anthropic key."

## Impact

In a multi-agent Hive instance, all agents share the same host and can reach `localhost:9000` (the apiproxy default bind address). A prompt-injected or otherwise compromised agent can:
- Make arbitrarily large or expensive LLM API calls billed entirely to the operator.
- Use the Anthropic key as an exfiltration side-channel by embedding data in model prompts.
- Enumerate operator-available models and usage quotas.

No credentials are required — the attacker simply sends HTTP requests to the loopback proxy.

## Recommendation

1. Add an auth gate at the top of `ServeHTTP()`: if `p.apiKey != ""` and the incoming request has no valid `Authorization` / `X-Api-Key` header, return `401 Unauthorized` immediately before forwarding.
2. Alternatively, bind the proxy to a per-agent Unix domain socket (owned by the launcher user, not world-accessible on localhost).
3. At minimum, document that `PROXY_AUTH_TOKEN` **must** be set whenever apiproxy is deployed in a multi-agent environment.

---

Danathar:
## Summary
- reject requests without valid `Authorization` or `X-Api-Key` before forwarding
- ensure blocked requests are neither logged nor sent upstream
- add regression tests and update warning/deployment documentation

Fixes #4068

## Tests
- `cd src && go test ./cmd/apiproxy ./pkg/apiproxy`
- `cd src && go test ./pkg/apiproxy -race`
- `git diff --check`

Signed-off-by: Danathar <dbaggett@users.noreply.github.com>

### DIFF
--- a/src/cmd/apiproxy/main.go
+++ b/src/cmd/apiproxy/main.go
@@ -21,7 +21,7 @@ func proxyAuthTokenFromEnv(getenv func(string) string, warnf func(string, ...any
 	}
 	authToken = getenv("ANTHROPIC_API_KEY")
 	if authToken != "" {
-		warnf("[sec-check WARNING] PROXY_AUTH_TOKEN is unset; falling back to ANTHROPIC_API_KEY. Unauthenticated callers will be granted the host Anthropic key. Set PROXY_AUTH_TOKEN to restrict access.")
+		warnf("[sec-check WARNING] PROXY_AUTH_TOKEN is unset; falling back to ANTHROPIC_API_KEY to enable the proxy auth gate. Unauthenticated callers will be rejected. Set PROXY_AUTH_TOKEN explicitly to avoid using the host Anthropic key as the gate token.")
 	}
 	return authToken
 }
--- a/src/cmd/apiproxy/main_test.go
+++ b/src/cmd/apiproxy/main_test.go
@@ -33,6 +33,9 @@ func TestProxyAuthTokenFromEnvWarnsOnAnthropicFallback(t *testing.T) {
 	if len(warnings) != 1 || !strings.Contains(warnings[0], "PROXY_AUTH_TOKEN is unset") {
 		t.Fatalf("expected actionable fallback warning, got %#v", warnings)
 	}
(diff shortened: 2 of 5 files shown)


clubanderson:
Thanks @Danathar — the gate itself is the right call, but as written it makes the injection branch unreachable: `ServeHTTP` and `director()` use the identical `isValidAuth()` predicate, so anything that passes the gate already has valid upstream credentials and the dummy-key swap never fires (agent CLIs sending `sk-ant-dummy…` would start getting 401s).

I've opened #4075 which builds on your commit and separates the two credentials: `PROXY_AUTH_TOKEN` becomes the caller-facing gate token (validated with `crypto/subtle.ConstantTimeCompare` and stripped before proxying), while `ANTHROPIC_API_KEY` stays purely the upstream key that `director()` injects. Your commit is preserved in the history there.

clubanderson:
Merged as part of #4075 (your commit is preserved in the history, squashed into 3adcfa9). Thanks again for catching the underlying issue in #4068 — the follow-up just separates the gate token from the upstream key so the dummy-key swap keeps working.

Issue #2565, opened 2026-08-04
contribute: PR attribution is a tmux-tail heuristic used for cooldown and trust credit


castrojo:
> Part of #2534 (contributor onboarding field notes from `projectbluefin/review`).

## Summary

Hive now distinguishes a completion that reports a PR from one where the CLI merely returned to idle, but current `v2` still infers the PR association from recent tmux output and trusts the reported URL without server-side verification. On the live Project Bluefin Hive, activity from a deployed revision whose exact SHA is not exposed showed real PRs followed by later reassignment of the same issues.

The narrow question is: **should cooldown, duplicate suppression, completion credit, and trust promotion depend on authoritative GitHub evidence rather than a best-effort terminal scrape?**

**Deployment caveat:** the observed assignments predate the latest successful image builds, and the spoke exposes no deployed commit or image digest. They are evidence of the symptom on an unknown deployed revision, not proof that current `v2` HEAD produced it. The source-level concern below is independently verified against current HEAD.

## Current mechanism

At `v2` commit [`08518560`](https://github.com/kubestellar/hive/commit/08518560e14cc583215765831cf72487d432d1c1):

- The relay scans the captured tmux tail for a GitHub PR URL. It prefers the assigned repo, but falls back to the first PR URL mentioned in the output ([`contributor-relay.sh:365-390`](https://github.com/kubestellar/hive/blob/08518560e14cc583215765831cf72487d432d1c1/bin/contributor-relay.sh#L365-L390)).
- Returning to an idle prompt sends `task_complete` with that best-effort URL ([`:654-666`](https://github.com/kubestellar/hive/blob/08518560e14cc583215765831cf72487d432d1c1/bin/contributor-relay.sh#L654-L666)).
- A non-empty URL produces the 168-hour cooldown; an empty URL produces four hours ([`contribute_ws.go:512-540`](https://github.com/kubestellar/hive/blob/08518560e14cc583215765831cf72487d432d1c1/v2/pkg/dashboard/contribute_ws.go#L512-L540)).
- The same unverified field increments `TasksWithPR` and can trigger newcomer → contributor promotion ([`:1222-1264`](https://github.com/kubestellar/hive/blob/08518560e14cc583215765831cf72487d432d1c1/v2/pkg/dashboard/contribute_ws.go#L1222-L1264)).

#2437 explicitly noted that this raises the bar but does not close the hole because `PRURL` remains client-supplied.

## Measured evidence from 2026-08-04

[...]

PR #4096, opened 2026-08-18
✨ feat(attribution): model + reasoning-effort footer on PRs, Live Activity, and hub-side verification

### LINKED ISSUE #4083: feat(attribution): show the exact model + reasoning effort at the bottom of every PR, including contributor-relay PRs
## What I want

Every PR the hive produces should say, at the bottom of the PR body, **exactly which model and effort produced it** — e.g. `GPT-5.6 Terra / high`, `Claude Sonnet 5 / medium`, `Gemini 3 / low`. Right now you can look at a merged PR and have no idea what actually wrote it. The point is transparency: people should be able to see what models are being used across the hive, and compare.

This is one half of a pair; the other half is the same information in the Live Activity feed (see companion issue #4084).

## Where we actually are today

Good news: **most of the machinery already exists.** `src/pkg/github/attribution.go` already builds a visible trailer and appends it to hive-mediated PRs:

```
— hive: agent=quality backend=bob model=auto bobshell=1.0.6
```

`AppendTrailer(body, meta)` is called from `src/pkg/github/pr_request_watcher.go:203`, gated by `governor.attribution_trailer` (default ON). It is idempotent (won't stack on a retry) and it omits unknown fields rather than guessing.

So there are exactly **two gaps**, and they're both small and well-scoped:

### Gap 1 — reasoning effort is never recorded anywhere

`InvocationMeta` (`attribution.go`) carries `Agent`, `Backend`, `Model`, `Tool`, `ToolVersion`, `Session`. There is **no effort field**, so a `codex … high` run and a `codex … minimal` run are indistinguishable in the trailer. Effort is a first-class dial for several backends:

- codex → `-c model_reasoning_effort="<effort>"`
- agy → `--effort low|medium|high` (and agy *silently ignores `--model` without it* — see `AGY_DEFAULT_EFFORT` in `bin/contributor-relay.sh:439`)
- claude → effort is a session setting

The relay already resolves it (`REASONING_EFFORT`, `bin/contributor-relay.sh:45`) and the hub-side launcher already resolves `agyDefaultEffort` (`src/pkg/agent/manager.go`). It just never reaches the trailer.

**Note the wire is already carrying it and the hub throws it away.** `bin/contributor-relay.sh:1804` sends `reasoning_effort` in `auth_response`, but the hub's message struct (`src/pkg/dashboard/contribute_ws.go:204`) declares only `CLIBackend` and `Model` — no `ReasoningEffort` — so the field is silently dropped at JSON decode. Fixing that decode is the shared prerequisite for both this issue and the companion one.

### Gap 2 — contributor-relay PRs never pass through the trailer at all

`bin/gh-wrapper.sh:505` deliberately exempts contributors from the `hive-open-pr` redirect:

> Contributors are EXEMPT: they fork and PR under their OWN identity by design, so their `gh pr create` must pass through unchanged.

That exemption is correct on authorship grounds and should stay — but it means **every community/relay PR, which is most of the PRs a person actually browses, has no attribution footer whatsoever.** The attribution file even calls this out as an intentional current limitation ("an issue an agent opens directly through its own CLI never passes through this code").

## Proposed scope

1. **Plumb effort end-to-end.**
   - Add `ReasoningEffort string \`json:"reasoning_effort,omitempty"\`` to the WS message struct in `contribute_ws.go` and store it on `ContributorConnection` alongside `cliBackend`/`model`.
   - Add `Effort` to `InvocationMeta`, emit it from `pairs()` between `model` and the tool version, and populate it from the launcher for pod agents.
   - Trailer becomes: `— hive: agent=quality backend=codex model=gpt-5.6-terra effort=high codex=0.5.2`

2. **Give contributor-relay PRs a footer too, without changing who authors them.**
   This must work in **both local and containerized mode** — no partial coverage. That rules out the obvious approach:

   > ~~Have `bin/gh-wrapper.sh` append the trailer to `--body` in contributor mode before `exec`ing the real `gh pr create`.~~

   The wrapper only exists in the **container**: `src/Dockerfile.contributor:177` installs it as `/usr/local/bin/gh` and line 186 touches `/etc/hive/contributor-mode`. In **local** mode (`just contribute-hive <cli> local`) the contributor's `gh` is their own real binary on `$PATH` and the marker does not exist, so a wrapper-side footer would silently cover only half the fleet — and local mode is how most humans run the relay.

   The approach that covers both, uniformly, is **hub-side reconciliation after the PR opens**: the relay already reports `pr_url` on `task_complete` (`bin/contributor-relay.sh:1710`, via `detectPRURL()`), and the hub already knows that contributor's backend/model/effort from the `auth_response` handshake. On seeing a completion with a PR URL, the hub appends the trailer with the App token if it is missing. Mode-independent, and the values come from the hub's own record rather than an agent's self-report. Tracked in detail in #4085 — this issue should adopt whatever lands there rather than growing a second mechanism.

3. **One shared format.** Whatever path produced the PR, the footer line must be byte-identical in shape so it stays greppable (`AttributionTrailerPrefix` = `— hive:`). Reuse `AppendTrailer`'s idempotence rule so an edited/retried PR body never stacks two.

## Explicit non-goals

- **Pretty display names.** The trailer should record the honest launch identifiers (`gpt-5.6-terra`, `claude-opus-4-6`), not marketing names. `cli_models_codex.go` sees a `displayName` from discovery but deliberately keeps only the id; prettifying is a separate, presentation-layer decision and belongs in the UI (companion issue #4084), not in the audit trail.
- **Discovering which model *actually served* a session.** For bob we record `auto` because bob self-selects; `RequestedModel()` already documents that. This issue records what was *requested at launch*, which is what we can honestly know.
- Changing PR authorship, the App-bot identity, or the contributor exemption in `gh-wrapper.sh`.

## Acceptance

- A pod-agent PR body ends with a trailer that includes `effort=` when the backend has one.
- A contributor-relay PR body ends with the same-shaped trailer, with that contributor's real backend/model/effort.
- Retrying/editing a PR never produces two trailer lines.
- `governor.attribution_trailer=false` still suppresses the visible line on the hive-mediated path (existing behaviour), and the audit entry is still written regardless.
- Tests cover: effort present, effort absent (field omitted, no bare `effort=`), and the no-stacking case.

---
*Filed by Claude Opus 5.*

### LINKED ISSUE #4084: feat(dashboard): show the exact model + reasoning effort on every Live Activity event (onboarding + operations)
## What I want

The Live Activity feed — on **both** the Onboarding page and the Management & Operations page — should show, on every event, exactly which model and effort is behind it. When I see something was picked up, the line should say the same thing the PR footer says: `via codex CLI with gpt-5.6-terra (high)`.

The point is transparency: at a glance you can see what models are running in the hive right now, and which one picked up which task.

This is one of three related issues; the others put the same information at the bottom of every PR body (#4083) and enforce it on the paths the hive cannot intercept (#4085).

Like those, this must behave identically for **local-mode and containerized** contributors — the activity path already does, since it is fed by the WS handshake rather than by anything mode-specific.

## Where we actually are today

The **data is already there.** `ActivityEntry` (`src/pkg/dashboard/contribute_ws.go:316`) already carries `CLI` and `Model`, and `addActivity()` is already called with the connection's real backend and model on every event type — joined, left, picked up, completed, failed, promoted, lease expired, reassigned. Nothing needs to be captured; it's already stored and already served by `/api/contribute/activity`.

The gap is purely in **rendering**, and the two pages disagree with each other:

### Onboarding feed — partially there

`api_contribute.go:5937`:

```js
const cliModel=e.cli?(e.model?' <span class="feed-cli">via '+e.cli+' CLI with '+e.model+'</span>':' <span class="feed-cli">via '+e.cli+' CLI</span>'):'';
```

Shows backend + model on every event. Good. Missing effort. Also note this line interpolates `e.cli` / `e.model` **unescaped** while every neighbouring field in the same template goes through `esc()` — worth fixing while we're in here.

### Operations rail — barely there

`ccNarrate()` at `api_contribute.go:5482` shows the CLI on **`joined` only**, and never shows the model at all:

```js
case 'joined': body=who+' entered the hive'+(e.cli?' <span class="ref">via '+esc(e.cli)+'</span>':''); break;
case 'left':   body=who+' left the hive'; break;
case 'picked up': body=who+' grabbed'+pickedRef; break;
case 'completed': body=who+' completed'+ref; break;
case 'failed': body=who+' hit a snag on'+ref; break;
```

So on the ops rail — the one you actually watch while work is running — a "picked up" or "failed" line tells you nothing about what model is doing the work. That's the main thing this issue fixes.

### Effort is dropped on the floor at the hub

`bin/contributor-relay.sh:1804` already sends `reasoning_effort` in its `auth_response`. The hub's WS message struct (`contribute_ws.go:204`) declares `CLIBackend` and `Model` but **no `ReasoningEffort`**, so `encoding/json` silently discards it. This is the shared prerequisite with the companion PR-footer issue #4083 — whichever lands first should do it:

- add `ReasoningEffort string \`json:"reasoning_effort,omitempty"\`` to the message struct
- store it on `ContributorConnection` next to `cliBackend` / `model`
- add `Effort` to `ActivityEntry` and thread it through `addActivity()`'s signature (all ~8 call sites)
- add `Effort` to `FleetClanker` (`contribute_ws.go:2019`) so the fleet table can show it too

## Proposed scope

1. **Shared prerequisite** (above): stop dropping `reasoning_effort`, carry it to `ActivityEntry` and `FleetClanker`.
2. **One shared formatter** used by both pages, so they can't drift again. Something like `via <cli> CLI with <model> (<effort>)`, degrading gracefully: no effort → `via codex CLI with gpt-5.6-terra`; no model → `via codex CLI`; no cli → nothing at all. Never render a bare `()` or a dangling `with`.
3. **Apply it to every action in `ccNarrate()`**, not just `joined` — picked up / completed / failed / promoted / left all get the loadout suffix.
4. **Escape properly** in the onboarding feed (`esc()` on `e.cli` and `e.model`, matching the rest of that template).
5. Keep it visually subordinate — it's a suffix in the existing muted `.feed-cli` / `.ref` style, not a new column. The rail is narrow; long model ids like `claude-opus-4-6` need to wrap or truncate gracefully rather than blow out the layout.

## Explicit non-goals

- **Prettifying model ids.** Show what was actually requested (`gpt-5.6-terra`, `claude-opus-4-6`). `cli_models_codex.go:87` notes discovery emits a `displayName` we deliberately don't keep; adopting it is a separate call and shouldn't gate this.
- Backfilling effort onto activity entries already persisted to disk — old entries simply won't have it, and the formatter's graceful degradation covers that.
- Any change to what counts as an activity event, or to the SSE/poll plumbing.

## Acceptance

- The ops Live Activity rail shows backend + model (+ effort when known) on **every** event type, not just `joined`.
- The onboarding Live Activity feed shows the same string, produced by the same formatter.
- A contributor connected with no model configured still renders cleanly (`via bob CLI`, no trailing separators).
- `e.cli` / `e.model` are HTML-escaped on both pages.
- `reasoning_effort` from a relay's `auth_response` survives to the activity entry and to the fleet view.

---
*Filed by Claude Opus 5.*

### LINKED ISSUE #4085: feat(attribution): require and verify the model/effort footer on agent PRs the hive cannot intercept (local-mode relay, MCP, post-open edits)
## What I want

Agents should be **required** to state, at the bottom of every PR they open, exactly which model and reasoning effort produced it — and the hive should actually verify they did, rather than hoping.

Third of three related issues:

- #4083 — the hive *mechanically appends* the footer on paths it mediates
- #4084 — the same information in the Live Activity feed
- **this one** — the footer on paths the hive **cannot** mechanically intercept, plus verification

## Why this is not a duplicate of #4083

#4083 works by intercepting PR creation. There are two interception points and **neither one reaches the largest population of contributors**:

| Path | Intercepted? | How |
|---|---|---|
| Pod agent (in-hive) | ✅ | `pr_request_watcher.go` → `AppendTrailer` |
| Containerized contributor | ✅ (possible) | `Dockerfile.contributor:177` installs `bin/gh-wrapper.sh` as `/usr/local/bin/gh`, and line 186 touches `/etc/hive/contributor-mode` |
| **Local-mode contributor** (`just contribute-hive <cli> local`) | ❌ **no** | real `gh` on `$PATH`, no wrapper, no marker |
| Agent using GitHub MCP `create_pull_request` | ❌ no | never touches `gh` at all |
| Any agent that later *edits* its own PR body | ❌ no | can overwrite an appended trailer |

Local mode is the ordinary way a human contributor runs the relay on their own machine. For that path there is no binary to wrap and no server-side choke point at open time — the agent writes the PR body and that is the body. So the trailer has to be either **instructed** (the agent writes it) or **reconciled** (the hive fixes it after the fact). This issue covers both.

### 1. Make it an explicit, uniform instruction in agent policy

The role policies under `src/pkg/policies/defaults/*.md` already give every agent a `gh pr create` / `hive-open-pr` recipe with a `--body` (e.g. `scanner-full.md:37`, `quality-full.md:49`, `architect-full.md:53`, and ~10 more). `Scheduler.ghAuthInstructions()` (`src/pkg/scheduler/scheduler.go:800`) is the shared block that already dictates PR-body conventions — notably the `Fixes #N` vs `Refs #N` rule. That is the natural home for one more rule:

> End the PR body with the attribution line naming the backend, model, and reasoning effort you are running as.

Because it lives in the shared block, it reaches every role at once instead of being copy-pasted into a dozen policy files.

### 2. Give the relay-side agent the values to state

An agent generally does **not** reliably know its own model id or effort — it can't be trusted to introspect it, and a guess is worse than nothing in an audit trail. The relay does know: `AGENT_BACKEND` / `AGENT_MODEL` / `AGENT_REASONING_EFFORT` (`bin/contributor-relay.sh:43-45`), plus the resolved `agyEffort` fallback at line 441. So the instruction must ship the **literal line to paste**, interpolated by whoever composes the prompt, not a description of how to derive it.

Same format as #4083's trailer so it stays greppable — `AttributionTrailerPrefix` (`— hive:`) from `src/pkg/github/attribution.go:35`.

### 3. Verify, don't trust

An instruction alone is a hope. Add a reconciliation step — pick whichever is cheapest:

- **(a) Hub-side amend (strongest).** The relay already reports the PR URL on completion (`detectPRURL()` → `pr_url` on `task_complete`, `contributor-relay.sh:1710`). The hub knows that contributor's backend/model/effort from the `auth_response` handshake. So on receiving a `task_complete` with a `pr_url`, the hub can fetch the body and append the trailer with the App token if it's missing. This is authoritative — it uses the hub's own record of the connection rather than the agent's self-report, so an agent cannot misreport its model, and it works identically for local mode, container mode, and MCP-opened PRs.
- **(b) CI check.** A workflow that fails or labels a PR whose body has no `— hive:` line. Simple, but noisy for human PRs and needs an exemption rule.
- **(c) Advisory-only.** Surface "PRs missing attribution" as a count on the ops page. Zero enforcement, but zero false positives.

**(a) is the recommendation.** It closes the gap for *every* path in the table above with one implementation, and it is the only option where the recorded model is the hive's observation rather than the agent's claim.

### 4. Don't double-stamp

`AppendTrailer` is already idempotent — it no-ops when the body already contains `AttributionTrailerPrefix` (`attribution.go:164`). Any new path must reuse that same guard so a PR touched by two mechanisms (agent wrote it *and* hub reconciles it) ends up with exactly one line.

## Explicit non-goals

- Blocking or rejecting a PR for a missing footer. Attribution is transparency, not a merge gate — a missing line should be *fixed*, not punished.
- Requiring the footer on PRs opened by **humans**. This is about agent-authored PRs; any check must exempt human authors rather than nagging them.
- Changing PR authorship or the contributor identity exemption in `gh-wrapper.sh:505`.
- Trusting an agent's self-reported model over the hive's handshake record, where both exist.

## Mode coverage is a hard requirement

This must behave identically in **local mode and containerized mode**. Partial coverage is worse than none: if the footer appears only on containerized runs, then "no footer" silently means two different things (a local contributor vs. an agent that skipped it) and the data is not comparable across the fleet — which defeats the entire point of showing which models are being used.

Option (a) is the only one in the list above that satisfies this, because it acts on `task_complete` — a message every relay sends in every mode — rather than on a binary that exists in only one of them.

## Acceptance

- A PR opened by a **local-mode** relay contributor ends up with an attribution footer, whether the agent wrote it or the hub reconciled it.
- A PR opened via GitHub MCP (bypassing `gh` entirely) also ends up with one.
- A PR whose body was edited after opening still has exactly one footer, not zero and not two.
- The footer's model/effort matches what the hub recorded at handshake, not what the agent asserted.
- Human-authored PRs are untouched.

- Behaviour is identical in local mode and containerized mode.

---
*Filed by Claude Opus 5.*

clubanderson:
## What

Implements the three related model-attribution issues in one PR, since they share a wire-format prerequisite:

- **#4083** — every hive-mediated PR footer now records `effort=` alongside backend/model: `— hive: agent=quality backend=codex model=gpt-5.6-terra effort=high codex=0.5.2`
- **#4084** — the Live Activity feed (ops rail **and** onboarding page) shows `via codex CLI with gpt-5.6-terra (high)` on **every** event, via one shared formatter
- **#4085** — PRs the hive cannot intercept (local-mode relay, MCP, post-open edits) are instructed **and** verified: the assignment prompt ships the literal footer line, and the hub reconciles the trailer onto any verified-completion PR that lacks one

## Shared prerequisite

`bin/contributor-relay.sh` has been sending `reasoning_effort` on `auth_response` all along; the hub's `WSMessage` struct didn't declare it, so `encoding/json` silently dropped it. It is now declared, stored on `ContributorConnection`, and threaded to `ActivityEntry`, `FleetClanker`, the assignment prompt, and the reconciler.

## #4083 — footer on hive-mediated PRs

- `InvocationMeta.Effort`, rendered between `model` and the tool version; omitted (never a bare `effort=`) when unknown
- `Manager.InvocationMetadata` reports the launcher-resolved effort — agy's required `--effort` pairing (`agyDefaultEffort`) when a model is set; empty elsewhere, mirroring the actual launch command — and the `main.go` resolver forwards it to the PR-request watcher's `AppendTrailer` path
- `governor.attribution_trailer=false` still suppresses the visible line; the audit entry is unchanged

## #4084 — Live Activity

- One shared JS formatter `hiveLoadout(e,cls)` used by **both** pages so they cannot drift: `via <cli> CLI with <model> (<effort>)`, degrading gracefully (no effort → no parens; no model → `via bob CLI`; no cli → nothing)
- Applied to **every** `ccNarrate()` action on the ops rail (previously CLI-only, `joined`-only), rendered in the existing muted `.ref`/`.feed-cli` styles
- `e.cli`/`e.model`/`e.effort` are now HTML-escaped on the onboarding feed (previously interpolated raw)
- `FleetClanker.Effort` so the fleet table can show the loadout too

## #4085 — instruct + verify on unmediated paths

- The assignment prompt now ends with the **literal** line to paste (interpolated from the hub's own handshake record — an agent cannot be trusted to introspect its model id): `End your PR body with this exact attribution line … '— hive: backend=codex model=gpt-5.6-terra effort=high'`
- **Hub-side reconciliation (option (a) from the issue):** on a `task_complete` whose PR the hub has *verified* (#2565 path), `EnsurePRAttribution` fetches the body and appends the trailer with the App token when missing. It acts on `task_complete` — sent by every relay in **every** mode — so local-mode, containerized, and MCP-opened PRs behave identically. Values come from the hub's record, not the agent's claim. Reuses `AppendTrailer`'s idempotence guard, so a PR touched by two mechanisms ends up with exactly one line. Gated by the same `governor.attribution_trailer` toggle; audited as `pr_attribution_reconciled`. Best-effort and async — a reconcile failure never affects completion handling, cooldowns, or trust credit.
- Human PRs untouched: reconciliation only ever fires on a verified contributor-relay completion
- Deliberately **not** added to `ghAuthInstructions()`: that block is shared and per-agent-agnostic by design, pod agents can't know their own model (a guessed self-report is worse than nothing), and the `hive-open-pr` path already stamps them mechanically — instructing them would just pre-empt the authoritative stamp via the idempotence guard

## Tests

- effort present / absent (field omitted, no bare `effort=`) / whitespace-only
- no-stacking on retry, creation-time + reconcile double-touch → exactly one line
- `reasoning_effort` survives JSON decode → `ActivityEntry` → `FleetSnapshot`
- reconcile: appends when missing, no-op when present, empty-meta no-op, toggle-off no-op, error paths, audit entry
- prompt instruction degradation cases; `InvocationMetadata` agy effort cases

`go build ./...`, `go vet ./...` clean; `pkg/github`, `pkg/dashboard`, `pkg/agent`, `pkg/scheduler`, `cmd/hive` all pass (`cmd/bd` and two `pkg/config` token-cache tests fail identically on clean v4 — pre-existing local-environment quirks)

Fixes #4083
Fixes #4084
Fixes #4085

### DIFF
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -1406,7 +1406,7 @@ func main() {
 		// lazily per backend and cached. Only launch descriptors flow here —
 		// never tokens, keys, or prompt content.
 		ghClient.SetAttributionResolver(func(agentName string) github.InvocationMeta {
-			backend, model, known := agentMgr.InvocationMetadata(agentName)
+			backend, model, effort, known := agentMgr.InvocationMetadata(agentName)
 			if !known {
 				if ac, inCfg := cfg.Agents[agentName]; inCfg {
 					backend, model = ac.Backend, ac.Model
@@ -1419,7 +1419,10 @@ func main() {
 				// bob self-selects (no catalog): requested model is honestly
 				// "auto" — see github.RequestedModel for the known follow-up
 				// on discovering bob's internal routing.
-				Model:       github.RequestedModel(backend, model),
+				Model: github.RequestedModel(backend, model),
+				// Reasoning effort as resolved by the launcher; empty (omitted)
+				// for backends without a hub-side effort dial (#4083).
+				Effort:      effort,
 				Tool:        tool,
 				ToolVersion: toolVersion,
 			}
--- a/src/pkg/agent/invocation_metadata_test.go
+++ b/src/pkg/agent/invocation_metadata_test.go
@@ -8,13 +8,16 @@ import (
 
 // TestInvocationMetadata pins the contract the invocation-attribution trail
 // depends on: the EFFECTIVE backend/model (runtime overrides win over config),
-// and ok=false for an agent the manager does not know.
+// the launcher-resolved effort (agy's required --effort pairing; empty —
+// omitted, never guessed — everywhere else), and ok=false for an agent the
+// manager does not know.
 func TestInvocationMetadata(t *testing.T) {
 	tests := []struct {
 		name        string
 		agent       *AgentProcess
 		wantBackend string
 		wantModel   string
+		wantEffort  string
 	}{
 		{
 			name:        "config values with no overrides",
@@ -42,6 +45,25 @@ func TestInvocationMetadata(t *testing.T) {
 			wantBackend: "",
 			wantModel:   "",
 		},
+		{
+			// #4083: agy is launched with --effort agyDefaultEffort whenever a
+			// model is given (agy requires the pair), so the attribution trail
+			// records that same effort.
+			name:        "agy with a model reports the launcher's effort",
+			agent:       &AgentProcess{Name: "a", Config: config.AgentConfig{Backend: "agy", Model: "gemini-3-pro"}},
+			wantBackend: "agy",
+			wantModel:   "gemini-3-pro",
+			wantEffort:  agyDefaultEffort,
+		},
+		{
+			// agy without a model is launched with NO --model/--effort pair, so
+			// no effort is recorded.
+			name:        "agy without a model reports no effort",
+			agent:       &AgentProcess{Name: "a", Config: config.AgentConfig{Backend: "agy"}},
+			wantBackend: "agy",
+			wantModel:   "",
+			wantEffort:  "",
+		},
 	}
 
 	for _, tc := range tests {
@@ -49,13 +71,13 @@ func TestInvocationMetadata(t *testing.T) {
 			m := testManager(acmmLevelPushCapable)
 			m.agents["a"] = tc.agent
 
-			backend, model, ok := m.InvocationMetadata("a")
+			backend, model, effort, ok := m.InvocationMetadata("a")
 			if !ok {
 				t.Fatal("InvocationMetadata: ok=false for a known agent")
 			}
-			if backend != tc.wantBackend || model != tc.wantModel {
-				t.Errorf("InvocationMetadata = (%q, %q), want (%q, %q)",
-					backend, model, tc.wantBackend, tc.wantModel)
+			if backend != tc.wantBackend || model != tc.wantModel || effort != tc.wantEffort {
+				t.Errorf("InvocationMetadata = (%q, %q, %q), want (%q, %q, %q)",
+					backend, model, effort, tc.wantBackend, tc.wantModel, tc.wantEffort)
 			}
 		})
 	}
@@ -65,7 +87,7 @@ func TestInvocationMetadata(t *testing.T) {
 // on ok=false, so an unknown name must report exactly that — not guesses.
 func TestInvocationMetadataUnknownAgent(t *testing.T) {
 	m := testManager(acmmLevelPushCapable)
-	if backend, model, ok := m.InvocationMetadata("nobody"); ok || backend != "" || model != "" {
-		t.Errorf("unknown agent must be (\"\", \"\", false); got (%q, %q, %v)", backend, model, ok)
+	if backend, model, effort, ok := m.InvocationMetadata("nobody"); ok || backend != "" || model != "" || effort != "" {
+		t.Errorf("unknown agent must be (\"\", \"\", \"\", false); got (%q, %q, %q, %v)", backend, model, effort, ok)
 	}
 }
--- a/src/pkg/agent/manager.go
+++ b/src/pkg/agent/manager.go
@@ -6279,25 +6279,35 @@ func (m *Manager) AuthorizeMerge(agentName string, fileUID int) error {
 	return nil
 }
 
-// InvocationMetadata reports the effective backend and model the hive invokes
-// for the named agent, accounting for runtime overrides — the launch-time
-// truth the invocation-attribution trail records (see pkg/github/attribution
-// .go). ok=false when the agent is unknown to the manager (the caller then
-// falls back to static config). Read-only under RLock; called from the
-// PR-request watcher goroutine, never from the launch path.
-func (m *Manager) InvocationMetadata(agentName string) (backend, model string, ok bool) {
+// InvocationMetadata reports the effective backend, model, and reasoning
+// effort the hive invokes for the named agent, accounting for runtime
+// overrides — the launch-time truth the invocation-attribution trail records
+// (see pkg/github/attribution.go). effort is empty when the launch command
+// carries no effort dial for this backend (only agy's --effort is hub-resolved
+// today, via agyDefaultEffort; see kubestellar/hive#4083). ok=false when the
+// agent is unknown to the manager (the caller then falls back to static
+// config). Read-only under RLock; called from the PR-request watcher
+// goroutine, never from the launch path.
+func (m *Manager) InvocationMetadata(agentName string) (backend, model, effort string, ok bool) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 	agent, exists := m.agents[agentName]
 	if !exists {
-		return "", "", false
+		return "", "", "", false
 	}
 	backend = effectiveBackend(agent)
 	model = agent.Config.Model
 	if agent.ModelOverride != "" {
 		model = agent.ModelOverride
 	}
-	return backend, model, true
+	// Mirror the launch command: agy is passed --effort agyDefaultEffort
+	// whenever a model is given (it requires the pair); every other backend is
+	// launched without a hub-side effort, so the field stays empty (omitted
+	// from the trailer) rather than guessed.
+	if backend == "agy" && model != "" {
+		effort = agyDefaultEffort
+	}
+	return backend, model, effort, true
 }
 
 // filteredEnv returns os.Environ() with write-capable tokens removed for advisory agents.
--- a/src/pkg/dashboard/api_contribute.go
+++ b/src/pkg/dashboard/api_contribute.go
@@ -5471,21 +5471,44 @@ function ccTaskRefLink(task){
   if(!m)return '<span class="ref">'+esc(task)+'</span>';
   return ccIssueLinkHTML({repo:m[1],number:m[2]},task,'ref');
 }
+// hiveLoadout is the ONE shared formatter for an activity entry's model loadout
+// (#4084): "via <cli> CLI with <model> (<effort>)" — the same wording the PR
+// footer uses — degrading gracefully: no effort drops the parens, no model drops
+// the "with", no cli renders nothing at all (never a bare "()" or dangling
+// "with"). Everything is esc()'d here so neither feed can interpolate a raw
+// cli/model/effort string again. cls picks the page's existing muted style
+// ('ref' on the ops rail, 'feed-cli' on the onboarding feed). Exported on
+// window so the onboarding feed's separate script reuses this exact function
+// instead of drifting its own copy.
+function hiveLoadout(e,cls){
+  if(!e||!e.cli)return '';
+  var s='via '+esc(e.cli)+' CLI';
+  if(e.model){
+    s+=' with '+esc(e.model);
+    if(e.effort)s+=' ('+esc(e.effort)+')';
+  }
+  return ' <span class="'+(cls||'ref')+'">'+s+'</span>';
+}
+window.hiveLoadout=hiveLoadout;
 function ccNarrate(e){
   var icons={joined:'🟢',left:'⚪',"picked up":'🔧',completed:'✅',failed:'❌',promoted:'🎖️'};
   var ic=icons[e.action]||'⚡';
   var who='<span class="who">'+esc(e.username||'someone')+'</span>';
   var ref=e.task?' <span class="ref">'+esc(e.task)+'</span>':'';
   var pickedRef=e.task?' '+ccTaskRefLink(e.task):'';
+  // #4084: every event line carries the loadout suffix — an operator watching
+  // the rail can see WHICH model picked up / completed / fumbled a task, not
+  // just that one joined.
+  var loadout=hiveLoadout(e,'ref');
   var body;
   switch(e.action){
-    case 'joined': body=who+' entered the hive'+(e.cli?' <span class="ref">via '+esc(e.cli)+'</span>':''); break;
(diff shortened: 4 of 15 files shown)


clubanderson:
Superseded by #4088 (merged, covers #4083) and #4091 (merged, covers #4084). Any remaining #4085 work will be tracked separately if needed.

PR #4101, opened 2026-08-18
🐛 fix(dashboard): Live Activity loadout dropped effort without a model, and coupled the onboarding feed to another script block

Danathar:
Review follow-up to #4091 (which implements #4084). #4091 is sound in structure — the Go plumbing threads `reasoning_effort` correctly through `WSMessage` → `ContributorConnection` → `ActivityEntry` → `FleetClanker`, and every `addActivity()` call site was updated. These are three defects found reviewing it after merge, each with a regression test.

## 1. The loadout dropped the effort whenever the model was unknown

The effort branch was **nested inside** the model branch:

```js
if(m){
  text+=' with '+m;
  if(ef){ text+=' ('+ef+')'; }
}
```

So an event with an effort but no model rendered `via codex CLI` and silently lost the `(high)`. Not hypothetical — `AGENT_REASONING_EFFORT` is independent of `AGENT_MODEL`, and a codex contributor who sets only the effort runs its default model *at that effort*. Dropping it loses exactly the value this feature exists to surface.

Effort is now rendered independently, still guarded, so a bare `()` or a dangling `with` still cannot be emitted.

## 2. The onboarding feed made bare cross-IIFE references

`esc()` and `ccFormatLoadout()` are declared inside the **Operations** script block's IIFE and reach the onboarding block only through the `window.*` republish at the end of that IIFE. #4091 called them by bare identifier.

That is precisely the shape this page's own init-order note describes:

> ccProjectName is declared in the ONBOARDING script block's IIFE, so it is not in scope here… **a bare cross-IIFE reference threw ReferenceError on every dossier render.**

And it fails closed. If that 3,300-line IIFE throws before the republish — it has regressed that way three times (#2603/#2604/#2606, per its own comment) — the reference throws, `poll()`'s bare `catch(e){}` swallows it, and the feed **silently freezes on a 3s loop with nothing in the console**. Before this change the onboarding feed had no dependency on that block at all.

The `(window.ccFormatLoadout||ccFormatLoadout)` hedge could never help: when the left operand is `undefined`, evaluating the right operand *is* the ReferenceError. It converted a clean `undefined` into a throw.

Both are now resolved through `window` with real local fallbacks, following the documented `ccProjectName` pattern. Worst case the feed loses the loadout suffix instead of dying. **The `esc` fallback is a genuine escaper, never a pass-through** — falling back to no escaping would turn a rendering failure into an injection bug.

## 3. The relay derived the reported effort separately from the launched one

```js
reasoning_effort: (BACKEND === 'agy' && MODEL) ? agyEffort : (REASONING_EFFORT || undefined),
```

This re-derives inline what `buildLaunchCommand()` already decides — the drift this file explicitly warns about for the launch command itself (#2203 bug 1). It already misreported: **agy without a model receives no `--effort` flag at all**, yet a raw `AGENT_REASONING_EFFORT` was advertised to the dashboard as though agy had applied it.

Extracted `effectiveReasoningEffort()` as the single source of truth, used by both the launch command and `auth_response`.

## On the tests

The suite shipped with #4091 asserts only that certain **source literals** appear on the rendered page. That cannot catch a formatter that renders the wrong string — it passed against the effort-dropping code in defect 1. Added:

- **`TestCCFormatLoadout_Behaviour`** — extracts the *shipped* `esc`/`ccFormatLoadout` from the rendered page by brace-matching and runs them under node, asserting all five degradation cases plus HTML escaping. Skipped when node is absent, matching the existing `exec.LookPath("jq")` precedent in this package. **Verified non-vacuous:** the pre-fix formatter returns `via codex CLI` for the effort-without-model case and fails this test.
- **`TestOnboardingFeed_NoBareCrossIIFEReference`** — pins the window-resolved locals and rejects the throwing `window.X||X` shape.
- **Three relay tests** — no effort reported for agy without a model; the reported effort and the launched `--effort` come from one resolver; an effort agy rejects falls back to `AGY_DEFAULT_EFFORT` and that fallback is what gets reported.

107/107 relay tests and the full `pkg/dashboard` suite pass.

---
*Reviewed and fixed by Claude Opus 5.*

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -469,16 +469,48 @@ function resolveBackend() {
   return cachedBackendResolution;
 }
 
+// modelFlagFor reports the --model flag this backend actually receives, or ''
+// when the backend takes no --model at all. Shared by the launch command and by
+// effectiveReasoningEffort() below, which must agree on whether a model is in
+// play — agy's effort is conditional on exactly that.
+function modelFlagFor() {
+  return MODEL && !NO_MODEL_FLAG_BACKENDS.includes(BACKEND) ? `--model ${MODEL}` : '';
+}
+
+// effectiveReasoningEffort is the SINGLE source of truth for the effort actually
+// in effect for this launch — the value the CLI is really running with, not the
+// value the contributor happened to export.
+//
+// It exists because the effort now travels twice: onto the command line here,
+// and up to the hub in auth_response so the dashboard can show it (#4084).
+// Deriving it independently in those two places is the same drift this file
+// already warns about for the launch command itself (#2203 bug 1, the comment
+// above cachedLaunchCommand), and it would misreport in two concrete ways:
+//
+//   - agy WITHOUT a model gets no --effort flag at all, so reporting a raw
+//     AGENT_REASONING_EFFORT there advertises an effort agy never applied.
+//   - agy WITH a model gets agyEffort, which falls back to AGY_DEFAULT_EFFORT
+//     when AGENT_REASONING_EFFORT is unset or is a value agy rejects (codex's
+//     vocabulary is wider), so the raw env var is the wrong answer there too.
+//
+// Returns '' when nothing is in effect; auth_response omits the field entirely
+// in that case rather than sending an empty string.
+function effectiveReasoningEffort() {
+  // agy is the only backend whose effort is conditional on a model being passed.
+  if (BACKEND === 'agy') return modelFlagFor() ? agyEffort : '';
+  return REASONING_EFFORT || '';
+}
+
 function buildLaunchCommand() {
   if (cachedLaunchCommand) return cachedLaunchCommand;
   const { cmd, perm } = resolveBackend();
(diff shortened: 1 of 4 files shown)


clubanderson:
Reviewed all three fixes:

1. **Effort without model** — correct: effort now rendered independently of the model branch, both guarded, so no bare `()` or dangling `with` is possible. Verified the new `TestCCFormatLoadout_Behaviour` exercises the shipped JS and would have failed against the pre-fix formatter.
2. **Onboarding feed decoupling** — correct: `feedEsc`/`feedLoadout` resolve through `window` with real local fallbacks (matching the documented `ccProjectName` pattern), and the fallback escaper genuinely escapes, so the failure mode degrades to a missing loadout suffix instead of a silently frozen feed or an injection risk. The old `(window.X||X)` hedge was indeed a ReferenceError, not a fallback.
3. **Relay single source of truth** — correct: `effectiveReasoningEffort()` is shared by `buildLaunchCommand()` and `auth_response`, fixing the agy-without-model misreport, with three regression tests.

All CI checks pass.

/lgtm
/approve

PR #4109, opened 2026-08-18
🐛 fix IBMid OIDC login: fall back to sub when uid claim absent, add clock-skew leeway, PS* algs, step-level logging

### LINKED ISSUE #4108: 🐛 IBMid SSO login fails at /api/auth/callback: 'login failed — could not verify your identity'
## Symptom

Logging in at hive.kubestellar.io with **IBMid** completes the full IBM SSO sequence (including passkey), then lands on:

```
https://hive.kubestellar.io/api/auth/callback?code=...&grant_id=...
```

with plain-text body **"login failed — could not verify your identity"** (HTTP 502). GitHub OAuth login works. Reproduced 2026-08-18 in Chrome incognito.

## Where it fails

That exact error string has a single source: `src/pkg/hub/oauth.go:477` — the OIDC `p.Exchange(...)` failure branch in `handleOAuthCallback`. The CSRF state check passed (it has its own distinct error string), so the failure is inside token exchange or id_token verification for the `ibmid` provider.

## Root cause (confirmed defect)

`src/pkg/auth/registry.go` hard-codes `subjectClaim: "uid"` for IBMid, on the basis that IBMid discovery `claims_supported` lists `uid` and not `sub`. But:

1. Per OIDC Core, `claims_supported` describes claims the OP *can* supply (largely via **userinfo**) — it does not enumerate id_token contents, and `sub` is REQUIRED in every id_token regardless. Live IBMid discovery (`https://login.ibm.com/oidc/endpoint/default/.well-known/openid-configuration`) confirms `subject_types_supported: ["public"]`, i.e. id_tokens carry a stable `sub`.
2. The hub never calls the userinfo endpoint, so if IBMid's **id_token** does not include `uid`, `verifyIDToken` (`src/pkg/auth/oidc_provider.go`) extracts an empty subject and returns `id_token has no subject` → the user-facing 502.

## Suspected contributing defects (fixed defensively in the same PR)

- **Zero clock-skew leeway**: the comment at `verifyIDToken` claims "the library's small leeway", but golang-jwt **v5's default leeway is zero**. A slightly-ahead `nbf`/`iat` or a marginal `exp` from IBMid's issuers fails verification with no tolerance.
- **RSA-PSS rejection**: IBMid advertises PS256/PS384/PS512 among `id_token_signing_alg_values_supported`; the verifier accepts only RS256/RS384/RS512 and its keyFunc type-asserts `*jwt.SigningMethodRSA`, which excludes `*jwt.SigningMethodRSAPSS`. A PS256-signed id_token is rejected even though the same JWKS RSA keys verify it.
- **Poor observability**: the callback logs one opaque warn; it does not classify WHICH step failed (discovery / token exchange / id_token verification), making prod diagnosis needlessly hard.

(The `grant_id` query param on the callback is normal for IBM Security Verify and is unrelated.)

## Fix

- Fall back to the standard `sub` claim when the configured subject claim (e.g. `uid`) is absent from the id_token; prefer `uid` when present (identities minted either way remain stable per user).
- Add a small verification leeway (2 min) for clock skew.
- Accept PS256/PS384/PS512 alongside RS*.
- Wrap `Exchange` errors with a step marker and log the failed step server-side (never leaked to the user).

clubanderson:
Fixes #4108

## Symptom

IBMid login at hive.kubestellar.io completes the full IBM SSO sequence, then fails at `/api/auth/callback?code=...&grant_id=...` with **"login failed — could not verify your identity"** (HTTP 502). GitHub login works. (The `grant_id` param is normal for IBM Security Verify callbacks and unrelated.)

## Root cause

**Confirmed defect** — `src/pkg/auth/registry.go` hard-codes `subjectClaim: "uid"` for IBMid on the basis that IBMid discovery `claims_supported` lists `uid` and not `sub`. But `claims_supported` describes claims the OP can supply (chiefly via **userinfo**, which the hub never calls) — it does not enumerate id_token contents, and OIDC Core REQUIRES `sub` in every id_token (live IBMid discovery confirms `subject_types_supported: ["public"]`). When the IBMid id_token carries `sub` but not `uid`, `verifyIDToken` extracted an empty subject and rejected the login with `id_token has no subject` → the user-facing 502 at `src/pkg/hub/oauth.go:477`.

**Suspected contributing defects** (cannot be conclusively confirmed without the prod log line, fixed defensively):
- golang-jwt **v5's default validation leeway is zero** (the previous comment claimed "the library's small leeway"). Marginal clock skew on `exp`/`nbf`/`iat` fails verification.
- IBMid advertises **PS256/PS384/PS512**; the verifier accepted only RS* and its keyFunc type-asserted `*jwt.SigningMethodRSA`, which excludes `*jwt.SigningMethodRSAPSS`. A PS-signed id_token was rejected despite verifying under the same JWKS RSA keys.

## Changes

- `pkg/auth/oidc_provider.go`
  - Subject extraction falls back to the standard `sub` when the configured override claim (e.g. IBMid `uid`) is absent; `uid` still wins when present. Fails closed (naming both claims checked) when neither exists.
  - `jwt.WithLeeway(2m)` for clock skew.
  - Accept PS256/PS384/PS512 alongside RS256/RS384/RS512; `alg=none` and HMAC remain rejected.
  - `Exchange` wraps its error with a step sentinel; new `FailedStep(err)` classifies discovery / token_exchange / id_token_verify.
- `pkg/hub/oauth.go` — the callback failure log now includes `step` and `nonce_cookie_present` (server-side only; user-facing message unchanged).
- `pkg/auth/registry.go` — corrected the IBMid subject-claim comment.
- `pkg/auth/oidc_provider_ibmid_fix_test.go` — tests for sub-fallback, uid preference, no-subject fail-closed, PS256 acceptance, leeway (within and beyond), and step classification.

## Identity-stability note

Falling back to `sub` cannot split identities: IBMid logins have never succeeded on this path (that is this bug), so no `ibmid:<uid>` identities exist to diverge from. Users whose id_tokens do include `uid` continue to key on `uid`.

## Validation

From `src/`: `go build ./...` ✅, `go vet ./...` ✅, `go test ./pkg/auth/` ✅, `go test -p 1 ./pkg/hub/` ✅ (serial, 121s).

## How to verify in prod

1. Deploy, then attempt an IBMid login in a fresh incognito window — it should land on `/dashboard`.
2. If it still fails, the hub log now pinpoints the stage: `OIDC: callback verification failed provider=ibmid step=<discovery|token_exchange|id_token_verify> nonce_cookie_present=<bool> error=...`, and the verification errors name the exact claim problem. `HIVE_HUB_OIDC_IBMID_SUBJECT_CLAIM=sub` remains available as an env-only retune.

---

## Addendum: full captured callback URL — state encoding ruled out, hardening added (commit 0045a27f)

The operator captured the full failing URL; its double-encoded state (`%253Aibmid%253A`) is **exactly what the hub itself emits** (QueryEscape in `startProviderLogin` + `url.Values.Encode` in `AuthCodeURL`) and is correctly reversed by the callback's compensating double decode — IBMid returned state verbatim per spec. Decisive: a state failure yields `invalid login state` (400), not the observed `could not verify your identity` (502), so state verification **passed** and the failure was inside `Exchange`, as diagnosed above. Full trace in [this comment](https://github.com/kubestellar/hive/pull/4109#issuecomment-5335342235).

The second commit hardens the fragile encoding contract anyway: cookie-nonce-gated candidate decoding across depths (tolerates an IdP that pre-decodes or re-encodes state, fails closed on any mismatch), `parseCallbackState` bound to the verified candidate, reason-coded rejection logs (`no_cookie`/`no_state`/`nonce_mismatch`) plus a non-standard-depth canary warn, and regression tests pinning the literal captured production state shape and the full round-trip contract.

### DIFF
--- a/src/pkg/auth/oidc_provider.go
+++ b/src/pkg/auth/oidc_provider.go
@@ -47,6 +47,38 @@ const httpTimeout = 10 * time.Second
 // is a memory-safety bound against a hostile or broken endpoint.
 const maxResponseBytes = 1 << 16
 
+// jwtClockSkewLeeway tolerates small clock drift between the hub and the
+// provider when validating exp/nbf/iat. golang-jwt v5 defaults to ZERO leeway
+// (there is no built-in "small leeway"), so without this a provider whose clock
+// is seconds ahead of ours mints id_tokens the hub rejects as "not valid yet".
+// Two minutes matches common OIDC RP practice and stays far below any token
+// lifetime.
+const jwtClockSkewLeeway = 2 * time.Minute
+
+// Step sentinels: Exchange wraps its error with exactly one of these so the hub
+// can log WHICH step of the OIDC callback failed (discovery / token exchange /
+// id_token verification) without parsing error strings. Server-side diagnostics
+// only — never shown to the user.
+var (
+	ErrStepDiscovery     = errors.New("step=discovery")
+	ErrStepTokenExchange = errors.New("step=token_exchange")
+	ErrStepVerifyToken   = errors.New("step=id_token_verify")
+)
+
+// FailedStep classifies an Exchange error by the step sentinel it wraps.
+// Returns "unknown" for a nil or unclassified error.
+func FailedStep(err error) string {
+	switch {
+	case errors.Is(err, ErrStepDiscovery):
+		return "discovery"
+	case errors.Is(err, ErrStepTokenExchange):
+		return "token_exchange"
+	case errors.Is(err, ErrStepVerifyToken):
+		return "id_token_verify"
+	}
+	return "unknown"
+}
+
 // jwksRefreshMinInterval throttles JWKS re-fetches triggered by an unknown `kid`.
 // Providers rotate signing keys occasionally; a flood of tokens with an unknown
 // kid (garbage or an attack) must not turn into a fetch-per-request DoS on the
@@ -160,13 +192,17 @@ func (p *Provider) Exchange(ctx context.Context, code, redirectURI, expectNonce 
 		return nil, errors.New("Exchange is only for OIDC providers; GitHub uses the hub's own handler")
 	}
 	if err := p.ensureDiscovered(ctx); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w: %w", ErrStepDiscovery, err)
 	}
 	rawIDToken, err := p.fetchIDToken(ctx, code, redirectURI)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w: %w", ErrStepTokenExchange, err)
 	}
-	return p.verifyIDToken(ctx, rawIDToken, expectNonce)
+	claims, err := p.verifyIDToken(ctx, rawIDToken, expectNonce)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrStepVerifyToken, err)
+	}
+	return claims, nil
 }
 
 // fetchIDToken posts the code to the token endpoint and returns the raw id_token.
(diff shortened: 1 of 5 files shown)


clubanderson:
## Full callback URL analysis (state double-encoding hypothesis — investigated and ruled out as root cause)

The operator captured the complete failing callback:

```
/api/auth/callback?code=oo0dDhve...&grant_id=7c4576a5-...&state=f4948cc2...%253Aibmid%253A
```

The `state` is indeed double-encoded (`%253A` = `:` escaped twice). Traced against the code, **this is exactly what the hub itself emits** — it is not IdP tampering:

1. `startProviderLogin` (`src/pkg/hub/oauth.go`) builds `state := url.QueryEscape(nonce + ":" + provider + ":" + redirect)` — **first** escape.
2. For OIDC providers, `AuthCodeURL` (`src/pkg/auth/oidc_provider.go`) does `v.Set("state", state)` + `v.Encode()` — **second** escape on the authorize URL.
3. IBMid parses the authorize query once (seeing the single-encoded string as the state *value*), returns it **verbatim per spec**, and re-encodes it onto the callback → `%253A` on the wire. This matches the captured URL byte-for-byte.
4. The callback compensates with a matching double decode: `r.URL.Query().Get("state")` (one decode) + explicit `url.QueryUnescape` (second decode) in both `verifyOAuthStateNonce` and `parseCallbackState` → `f4948cc2...:ibmid:` parses correctly.

**Decisive evidence:** a state failure produces `invalid login state — please start the login again` (HTTP 400, oauth.go). The operator saw `login failed — could not verify your identity` (HTTP 502) — a branch reachable **only after state verification passed and the provider parsed as `ibmid`**. So the failure was inside `p.Exchange` (token exchange / id_token verification), consistent with the original root-cause analysis (`uid` subject claim absent from the id_token, zero JWT leeway, or PS* alg). The `grant_id` param is standard IBM Security Verify callback decoration and inert.

## Additional hardening pushed (commit 0045a27f)

The encoding contract was fragile (implicit, asymmetric between the GitHub and OIDC paths, and untested at the production shape), and state rejections logged one opaque line. Now:

- **Tolerant, fail-closed state decoding**: `matchedCallbackState` tries the state at decode depths raw / standard / one-extra-pass and accepts only the candidate whose leading nonce **matches this browser's cookie** (constant-time per candidate). This survives an IdP that pre-decodes or re-encodes state, while a foreign nonce is still rejected at every depth.
- `parseCallbackState` reads provider/redirect from the **same** cookie-matched candidate that was verified — never a differently-decoded sibling.
- **Observability**: rejections now log a reason (`no_cookie` / `no_state` / `nonce_mismatch`), and a match at a non-standard decode depth logs a canary warn (`OAuth: state matched at non-standard decode depth`) so any future IdP encoding change is visible in one log line.
- **Regression tests** (`src/pkg/hub/oauth_state_encoding_test.go`): the literal captured production state shape (nonce `f4948cc2...`, `%253Aibmid%253A`, `grant_id` present) verified end-to-end; the full authorize round-trip encode/decode contract; pre-decoded and extra-encoded tolerance; and fail-closed mismatch/missing-cookie/missing-state behavior.

The base64url-state redesign was deliberately NOT done here: it would invalidate in-flight logins mid-deploy, and the cookie-gated candidate matching achieves the same robustness with full backward compatibility.

Issue #4110, opened 2026-08-18
✨ feat(dashboard): show cost-per-PR and cost-per-issue in spoke cost section


clubanderson:
## Summary

The spoke dashboard's cost section currently shows aggregate token spend and estimated cost. Add two derived metrics that give owners a more intuitive signal about the value they're getting:

- **Cost per PR** — total cost ÷ total PRs merged (for the spoke)
- **Cost per issue** — total cost ÷ total issues closed (for the spoke)

## Motivation

Aggregate cost is hard to reason about in isolation. "We spent $12 this week" is much more actionable as "$1.20 per merged PR" or "$0.80 per closed issue" — it lets spoke owners compare efficiency across time periods, across agents, and against manual engineering cost.

## Acceptance Criteria

- [ ] Cost section displays **Cost/PR** and **Cost/Issue** as formatted currency values (e.g. `$1.20`) alongside existing aggregate cost
- [ ] Both metrics are derived from data already present in the dashboard (token cost + PR/issue counters)
- [ ] Show `—` (or similar) when denominator is zero (no PRs merged / no issues closed yet)
- [ ] Metrics update with the same refresh cadence as the rest of the cost section
- [ ] Works correctly for both hub-managed and self-hosted spokes

## Notes

- PR count: use merged PRs only (not open/draft)
- Issue count: use closed issues only
- If cost data spans a rolling window (e.g. 30 days), both denominators should use the same window for consistency

PR #4116, opened 2026-08-19
feat(dashboard): show cost-per-PR and cost-per-issue in spoke cost section

Danathar:
## Summary

Adds two derived efficiency metrics to the dashboard's Cost section, per #4110:

- **Cost/PR** — all-time estimated cost total ÷ total merged PRs (primary repo)
- **Cost/Issue** — all-time estimated cost total ÷ total closed issues (primary repo)

## Design notes

- The counts are fetched via the GitHub Search API (`is:merged` / `is:closed`, no author filter — cost/PR and cost/issue are meant to reflect the spoke's overall output, not just AI-authored items) on the existing `MetricsCollector` 5-minute collection cycle, following the same pattern already used for MTTR (`collectMTTR` / `ghpkg.ComputeMTTR`): a new `ComputePRIssueCounts` method on the GitHub client, collected each cycle, cached to `/data/metrics/pr-issue-counts.json` so a restart doesn't lose the figure, and served through `GET /api/cost` (new `merged_prs` / `closed_issues` fields).
- Both new metrics are paired with the existing **all-time cumulative** total (`estimated.total_usd`), not a rolling window — since that's the scope of the aggregate figure they sit next to in the UI, this keeps numerator and denominator on the same window by construction (the issue's "same window" note applies when the aggregate itself is windowed, which it isn't here).
- Renders as `$X.XX` next to the existing "estimated total" tile; shows `—` when the denominator is zero, per the acceptance criteria.
- Works for both hub-managed and self-hosted spokes since it depends only on the GitHub client (not on which inference backend/gateway is configured).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/dashboard/... ./pkg/github/...`
- [x] `gofmt -l` on all changed files (clean)
- [x] `go test ./pkg/dashboard/... ./pkg/github/...` (full suite, including new tests for `ComputePRIssueCounts`, the `MetricsCollector` collect/cache wiring, and the `/api/cost` payload)
- [x] Verified the modified inline `<script>` block in `index.html` still parses (no JS syntax errors)

— hive: backend=claude

### DIFF
--- a/src/pkg/dashboard/cost.go
+++ b/src/pkg/dashboard/cost.go
@@ -84,6 +84,13 @@ type costResponse struct {
 	// the estimate without hardcoding the date.
 	PriceTableDate string `json:"price_table_date"`
 	Disclaimer     string `json:"disclaimer"`
+	// MergedPRs / ClosedIssues are all-time counts for the primary repo,
+	// matching Estimated.TotalUSD's "all-time cumulative" scope. The UI divides
+	// TotalUSD by these to derive cost-per-PR / cost-per-issue (issue #4110).
+	// Zero means "no data yet" (collector hasn't run, or GitHub is unreachable);
+	// the UI shows "—" rather than treating it as a real zero denominator.
+	MergedPRs    int `json:"merged_prs"`
+	ClosedIssues int `json:"closed_issues"`
 }
 
 // costModelEntry is one row of the estimated per-model / per-agent breakdown.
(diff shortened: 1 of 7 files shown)


clubanderson:
/lgtm
/approve

Reviewed: GitHub Search API usage (PerPage:1, total_count only), atomic cache-file writes with tmp+rename, mutex-protected in-memory state, zero-denominator guard rendering `—` in the UI, and clean JSON field additions to /api/cost. Tests cover mock server, error paths, and disk cache round-trip. LGTM.

PR #4128, opened 2026-08-19
🐛 fix(relay): let agy idle detection cross the input box's closing rule

Danathar:
## Problem

A contributor relay on the `agy` backend reports every **successful** task as a failure.

Observed live on this repo. A task read #4126, edited both workflows, pushed a branch and opened **#4127** — then sat at agy's idle prompt for 20 minutes until the `progressTick()` stall backstop killed it:

```
Pane unchanged for 20+ minutes — confirming before giving up on ct-kubestellar/hive-4126-1787145194 (1/2)
Relaunching agy after a confirmed pane stall: agy --dangerously-skip-permissions
Task ct-kubestellar/hive-4126-1787145194 failed [environment]: no pane activity for 20+ minutes,
  confirmed over 2 checks — the agent CLI is not visibly working
```

A shipped PR booked as an `environment` failure is not just log noise:

- the contributor loses the completed-task credit that **tier promotion** is counted from, so they cannot progress toward the tiers that gate delegated roles;
- the issue accrues **consecutive-failure weight** toward the 6-hour quarantine (`consecutiveFailureQuarantineThreshold`) while already being solved;
- ~20 minutes of lease is burned per task, plus a CLI relaunch.

## Root cause

#4080 taught the classifier agy's newer idle chrome — a bare `>` followed by the model footer. But agy **closes** its input box with a second box-drawing rule, so the real pane is:

```
────────────────────────────────
>


────────────────────────────────
                    Gemini 3.7 Flash · high
```

`U+2500` is not whitespace, so `\s*` in

```js
/(?:^|\n)>\s*\n\s*\n\s*Gemini\b[^\n]*\s*$/m
```

can never reach the footer. `hasIdlePrompt` stays false, `classifyTmuxPane()` falls through to `return PANE_STATE_WORKING`, and the relay renews the hub's lease on an agent that finished minutes ago. The pane never changes again, so the stall backstop is the only exit — and it exits as a failure.

Verified by running the shipped classifier against a live `tmux capture-pane -p` of the finished #4127 turn: `state: WORKING, hasIdlePrompt: false`.

## Why CI did not catch it

`AGY_GEMINI_IDLE_PANE` omitted the closing rule and the footer padding, so the fixture matched a regex the real pane did not. The fixture is now reproduced from a live capture, and **that pre-existing test fails without this change**.

## Fix

Allow the rule character in the gap between `>` and the footer.

Safety direction is preserved by the footer itself: while a turn is in flight agy renders `esc to cancel` on that same line, which is neither whitespace nor a rule, so a busy pane still cannot match. That direction — reporting a working agent as complete, the worse bug — is now pinned by its own test.

## Tests

`node --test bin/contributor-relay.test.js` → **109/109**. With `bin/contributor-relay.sh` reverted to the current `v4` blob:

```
FAIL agy Gemini footer with a bare prompt is COMPLETE
FAIL agy idle pane with a closing rule under the input box is COMPLETE
ok   agy pane still working ("esc to cancel") is not COMPLETE
```

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -1409,8 +1409,20 @@ function classifyTmuxPane(text) {
     // builds render a bare input line followed by the model footer instead.
     // Keep the bare ">" constrained to that footer so a Markdown quote in
     // an in-flight response cannot be mistaken for an idle prompt.
+    //
+    // The input box is CLOSED by a second box-drawing rule between the ">" and
+    // the footer, so the gap is not pure whitespace and "\s*" cannot cross it.
+    // Observed live: a turn that finished and opened kubestellar/hive#4127 sat
+    // at this exact idle chrome, classified WORKING, and was killed 20 minutes
+    // later by the progressTick() stall backstop and reported as an
+    // `environment` FAILURE — for a task that had shipped a real PR. Allow the
+    // rule character (U+2500) in the gap so the footer is reachable.
+    //
+    // Safety direction is preserved by the footer itself: while a turn is in
+    // flight agy renders "esc to cancel" on that same line, which is neither
+    // whitespace nor a rule, so a busy pane still cannot match here.
     hasIdlePrompt = /\? for shortcuts/.test(text) ||
-      /(?:^|\n)>\s*\n\s*\n\s*Gemini\b[^\n]*\s*$/m.test(agyTail);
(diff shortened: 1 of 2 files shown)


clubanderson:
/lgtm
/approve

Verified the new regex: matches idle chrome both with and without the closing box rule, still rejects in-flight panes ("esc to cancel" blocks the match) and mid-output Markdown quotes.

PR #4136, opened 2026-08-19
fix(ci): harden dashboard lint filename handling

### LINKED ISSUE #4125: [sec-check] dashboard-lint.yml: unescaped filename in node -e argument — shell injection via malicious PR filename
## Security Finding

**Severity**: medium
**Type**: unsafe-pattern / CI injection

## Description

`dashboard-lint.yml` iterates over HTML files and interpolates each path directly into a `node -e "..."` argument string:

```bash
for html_file in $(find dashboard -name '*.html'); do
  node -e "
    const fs = require('fs');
    const html = fs.readFileSync('$html_file', 'utf8');
```

Because `$html_file` is substituted unescaped inside the double-quoted bash string, a file whose name contains a `"` character (or backtick / `$(...)` ) breaks out of the JavaScript string literal and injects arbitrary Node.js code into the runner. On Linux, double-quote characters are legal in filenames.

A PR that adds `dashboard/x"; require('child_process').execSync('...');//.html` would cause arbitrary code execution on the CI runner during the `push` event (where the full `GITHUB_TOKEN` is available). On fork `pull_request` events the token is read-only, limiting but not eliminating the blast radius.

The workflow also has no explicit `permissions:` block, so it inherits the repository's default token permissions.

## Impact

- On an internal `push` event: arbitrary code execution on the runner with the full `GITHUB_TOKEN` (write access to the repo, packages, etc.)
- On a fork `pull_request` event: code execution on the runner with the read-only `GITHUB_TOKEN` (org enumeration, exfiltration of public repo data, abuse of the runner's network)

## Recommendation

1. Replace the inline shell interpolation with a safer pattern — pass the filename as a `--input` argument or use `printf '%s' "$html_file" | node script.js` with a separate script file that reads the path from stdin.
2. Add an explicit `permissions: contents: read` block to the workflow.
3. Alternatively, refactor using `mapfile`/`while IFS= read` with `find -print0 | xargs -0` to handle all unusual filenames safely.

---
---

Danathar:
## Summary

- pass dashboard HTML paths to Node as positional arguments instead of interpolating them into `node -e` source
- process filenames with NUL-delimited `find` output
- restrict the workflow token to `contents: read`

## Validation

- `actionlint .github/workflows/dashboard-lint.yml`
- `git diff --check`
- exercised the Node path handoff with a filename containing a double quote

Fixes #4125

— hive: backend=codex

### DIFF
--- a/.github/workflows/dashboard-lint.yml
+++ b/.github/workflows/dashboard-lint.yml
@@ -7,6 +7,9 @@ on:
     paths:
       - 'dashboard/**'
 
+permissions:
+  contents: read
+
 jobs:
   syntax-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


clubanderson:
Superseded by #4139 which implemented the same fix.

Issue #4167, opened 2026-08-19
🐛 certus: advisory digest stale since 2026-08-13 (not refreshing)


clubanderson:
## Observation

Certus hive advisory digest appears stuck at **2026-08-13** while current date is **2026-08-19**.

## Why this is a bug

With the new advisory refactor now running on certus (`f34f476`), the advisory digest should continue to refresh on governor eval cadence and reflect stale-bead pruning + top-N prioritization. A 6-day-old digest indicates advisory posting is wedged.

## Expected behavior

- Advisory digest comment updates regularly
- `AdvisoryLastPostedAt` advances
- stale/open findings evolve (pruned/closed/re-ranked) over time

## Suspected failure surfaces

- app-auth write path for advisory posting
- advisory post loop not executing for this hive
- advisory mode participation/agent configuration mismatch
- GitHub issue/comment update path errors being retained in `AdvisoryError`

## Acceptance criteria

- [ ] Root cause identified for certus stale advisory digest
- [ ] Digest posting resumes and `AdvisoryLastPostedAt` advances
- [ ] New digest reflects top-N prioritization and stale cleanup
- [ ] Add regression protection (test/log/alert) so this wedge is detected quickly

clubanderson:
Investigated fleet-wide (not certus-specific). Two suppression paths hid genuinely wedged digests on any hosted hive:

1. **Self-suppressing post failure** — a 403 on the digest comment makes the spoke raise `GitHubAppRequired`/`write-forbidden` in the same cycle, and the stale gate was gated on `appCanWriteForAdvisory`, so the failure silenced its own alarm.
2. **Restart amnesia** — `AdvisoryLastPostedAt` lives only in spoke memory, so every restart reported it empty, which the gate reads as "not an advisory participant". A hive wedged across a restart could never be flagged.

Fixed in #4170 (merged): the reported error is now suppressed only while the App is *undelivered*, and the hub carries the last post time forward across restarts (guarded on unchanged primary repo). Expected suppressions (PR-only hives, undelivered Apps, aged digests with no post attempt, offline rows) are unchanged.

No fleet data access, so prevalence is now measured: admin-only `GET /api/saas/admin/advisory-diagnostics` plus one structured `advisory diagnostics` log line every 30 min carrying the fleet `hidden_stale` count. Matrix in `docs/advisory-staleness.md`.

[...]

PR #4168, opened 2026-08-19
🐛 fix(contribute): stop local mode launching the agent inside this repo

Danathar:
Follow-up to #4142. That set `HIVE_WORKSPACE_DIR` in local mode, which was necessary but not sufficient — the agent still starts in, and works in, the hive checkout the relay was launched from.

## Problem

Local mode roots both the tmux session and the CLI launch at `$PWD`:

```sh
tmux new-session -d -s "$TMUX_SESSION" -x 200 -y 50 -c "$PWD"
tmux send-keys  -t "$TMUX_SESSION" "cd $(printf %q "$PWD") && $CMD $PERM_FLAG" Enter
```

`$PWD` is the hive checkout — which is also a clone of the repo the agent is assigned to work on. The assignment prompt explicitly permits reusing an existing clone:

> `gh repo fork … $HIVE_WORKSPACE_DIR/<owner>/<repo>` **(or, if that directory already has a clone from a prior task, `cd` into it and `git fetch` instead of re-forking)**

So an agent that starts inside one reasonably concludes it is already in its checkout.

## Observed

On `kubestellar/hive#4167`, with #4142 merged and working:

```
relay env:              HIVE_WORKSPACE_DIR=/home/dbaggett/workspace   ← set
workspace dir exists:   yes                                          ← created by #4142
workspace clone exists: NO                                           ← never cloned
gh repo fork executed:  0                                            ← never attempted
agy cwd:                /var/home/dbaggett/bluefin/hive              ← the relay's own repo
```

agy went straight to `Read(~/bluefin/hive/src/cmd/hive/main.go)` and `go test ./pkg/hub`, and began editing there. An earlier task on the same setup ran `git checkout -B sec/fix-workflow-permissions origin/v4` in that checkout — so the tree the running relay had been started from was no longer the tree it was reading. A task that happens to touch `bin/contributor-relay.sh` would edit the relay's source while it runs.

This is not reproducible on demand — it depends on the agent's judgement — but the divergence below is unconditional.

## The divergence

|  | session `-c` | CLI launched from |
|---|---|---|
| container (`bin/contributor-agent.sh`) | `$HIVE_WORKSPACE_DIR` | `$HOME` |
| local (`Justfile`) | `$PWD` (this repo) | `$PWD` (this repo) |

The container entrypoint never had this bug. This makes local mode match it.

`$HOME` also satisfies the #4046 constraint: nothing in the task lifecycle deletes or recreates it, unlike the workspace subtree agents clone into, so the launch `cd` still lands somewhere resolvable on a tmux server whose own cwd is gone.

## Who this actually affects

Narrower than it looks, which is likely why it survived this long.

The collision needs `$PWD` to be a clone of the **assigned** repo, not just any repo. Contributing to some *other* project from a hive checkout is fine: the agent finds a repo that is not the one it was told to work on and clones into the workspace correctly. Same mode, same backend, no bug.

So the affected population is **local-mode contributors working on hive itself** — hive's own contributors. Everyone else on local mode is unaffected, and container mode is unaffected outright (the repo is not even mounted; `WORKDIR /home/dev`).

`agy` is over-represented in that population because it is host-only — its Google OAuth flow has no API-key mode, so it cannot authenticate in a container and `just contribute-k8s` warns about it. Local mode is not a preference for agy contributors, it is the only option. But the bug is **mode-specific, not backend-specific**: any backend launched into a checkout of its own assigned repo can make the same call.

## Why the relay changes too

Fixing only the launch would be undone on the first stall recovery:

```js
function launchCommandWithCwd(launchCmd) {
  const cwd = process.cwd();   // the RELAY's dir — the same checkout
```

Both entrypoints now export `HIVE_AGENT_CWD` and the relay prefers it, falling back to `process.cwd()` so an older entrypoint that does not export it keeps its current behaviour.

## Tests

`node --test bin/contributor-relay.test.js` → **119/119**. Reverting only `bin/contributor-relay.sh`:

```
ok   a relaunch cds somewhere resolvable before starting the CLI
FAIL a relaunch prefers HIVE_AGENT_CWD over the relay cwd
```

`just --list` parses; the `contribute-hive` recipe body and `bin/contributor-agent.sh` both pass `bash -n`.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -527,9 +527,29 @@ contribute-hive backend="" mode="docker": check-version
       # directory (verified against tmux on Fedora 44). It is passed anyway
       # because it IS correct on a healthy server; the cd is what carries the
       # fix.
+      # ...but that cd must NOT land in this repo. In local mode $PWD is the hive
+      # checkout the relay was launched from — which is also a clone of the repo
+      # the agent is assigned to work on. The assignment prompt says to reuse an
+      # existing clone rather than re-fork ("if that directory already has a
+      # clone from a prior task, 'cd' into it"), so an agent that starts there
+      # reasonably concludes it is already in its checkout and works in place.
+      #
+      # Observed live on kubestellar/hive#4167 with HIVE_WORKSPACE_DIR correctly
+      # set and its directory present: agy never ran `gh repo fork` at all, and
+      # edited the relay's own source tree instead. An earlier task branch-
+      # switched the checkout out from under the running relay.
+      #
+      # The container entrypoint does not have this problem: it roots the
+      # session at the workspace and launches the CLI from $HOME
+      # (bin/contributor-agent.sh). Mirror that here so both paths agree. $HOME
+      # also satisfies the #4046 constraint above — nothing in the task
+      # lifecycle deletes or recreates it, unlike the workspace subtree the
+      # agent clones into.
+      export HIVE_AGENT_CWD="${HOME}"
+      mkdir -p "$HIVE_AGENT_CWD"
       tmux kill-session -t "$TMUX_SESSION" 2>/dev/null || true
-      tmux new-session -d -s "$TMUX_SESSION" -x 200 -y 50 -c "$PWD"
-      tmux send-keys -t "$TMUX_SESSION" "cd $(printf %q "$PWD") && ${LITELLM_ENV:+$LITELLM_ENV }$CMD $PERM_FLAG" Enter
+      tmux new-session -d -s "$TMUX_SESSION" -x 200 -y 50 -c "$HIVE_WORKSPACE_DIR"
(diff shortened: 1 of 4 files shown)


Danathar:
I can't tell you how much I laughed reading that it was a bug that ONLY affected me..because I wanted agy to work as a contributor to hive on hive!

clubanderson:
Reviewed: local mode now mirrors the container entrypoint (session at $HIVE_WORKSPACE_DIR, CLI from $HOME), HIVE_AGENT_CWD is exported by both entrypoints and preferred on relaunch with a process.cwd() fallback for back-compat, and the #4046 constraint holds since $HOME survives the task lifecycle. Tests cover both the preference and the fallback.

/lgtm
/approve

PR #4173, opened 2026-08-19
🔒 fix(contribute): don't root the unattended agent at $HOME

Danathar:
Hardening follow-up to #4168, raised in review by the contributor running the agy relay, who restarted and noticed the agent's new cwd.

## What #4168 got wrong

It stopped local mode launching the agent inside the hive checkout by moving it to `$HOME`, mirroring the container entrypoint so both paths agree. Mirroring was the wrong instinct: **the same path means very different things on each side.**

| | `$HOME` is… |
|---|---|
| container | `/home/dev` in an ephemeral image, three read-only mounts |
| local (host) | the contributor's real home — `.ssh`, `.gnupg`, `.claude`, and the relay's own registration token in `.config/hive/contributor.env` |

The agent is launched with its backend's skip-permissions flag (`--dangerously-skip-permissions` for agy, per `config/backends.conf`) and runs unattended. Its cwd is where every relative `ls`, `grep -r`, `find .` and relative write lands by default. Rooting it at `$HOME` makes reaching that material the path of least resistance.

## Scope, honestly

**This is defence in depth, not a boundary.** cwd grants nothing — the process runs as the user either way, and could always have `cd`'d to `$HOME` itself. Nothing here is an escalation, and no credential is newly reachable. What changes is the *default* working directory of an unattended agent, which is worth keeping off the user's home for the same reason you would not run one from `/`.

## Fix

`#4046` — the constraint that motivated the `$HOME` choice — only requires a directory the task lifecycle never deletes or recreates (unlike the workspace subtree agents clone into). An empty dedicated one satisfies that equally well:

```sh
export HIVE_AGENT_CWD="${XDG_STATE_HOME:-${HOME}/.local/state}/hive/agent-cwd"
mkdir -p "$HIVE_AGENT_CWD"
```

It keeps the property that fixed #4168 — **not a git repo**, so the agent cannot adopt its cwd as a checkout — while keeping the default blast radius off `$HOME`. Both entrypoints use it, so local and container still agree, and the relay's relaunch path already prefers `HIVE_AGENT_CWD`.

Honours `XDG_STATE_HOME` when set:

```
default:      /home/<user>/.local/state/hive/agent-cwd
XDG honoured: $XDG_STATE_HOME/hive/agent-cwd
```

## Tests

`node --test bin/contributor-relay.test.js` → **120/120**, including a new regression pinning that the relaunch target is never `$HOME` itself.

`just --list` parses; the `contribute-hive` recipe body and `bin/contributor-agent.sh` both pass `bash -n`.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -540,12 +540,22 @@ contribute-hive backend="" mode="docker": check-version
       # switched the checkout out from under the running relay.
       #
       # The container entrypoint does not have this problem: it roots the
-      # session at the workspace and launches the CLI from $HOME
-      # (bin/contributor-agent.sh). Mirror that here so both paths agree. $HOME
-      # also satisfies the #4046 constraint above — nothing in the task
-      # lifecycle deletes or recreates it, unlike the workspace subtree the
-      # agent clones into.
-      export HIVE_AGENT_CWD="${HOME}"
+      # session at the workspace and launches the CLI from a dedicated
+      # directory (bin/contributor-agent.sh). Mirror that here so both paths
+      # agree. That directory also satisfies the #4046 constraint above —
+      # nothing in the task lifecycle deletes or recreates it, unlike the
+      # workspace subtree the agent clones into.
+      #
+      # It is deliberately NOT $HOME, and local mode is why. In a container
+      # $HOME is /home/dev inside an ephemeral image with three read-only
+      # mounts; on the host it is the user's real home, holding .ssh, .gnupg
+      # and the contributor's own registration token in
+      # .config/hive/contributor.env. The agent runs unattended with its
+      # backend's skip-permissions flag, so its cwd is where every relative
+      # `ls`, `grep -r` and relative write lands. cwd is not a boundary — the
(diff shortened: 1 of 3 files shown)


clubanderson:
/lgtm
/approve

Reviewed and validated locally:
- Both entrypoints (Justfile `contribute-hive` and `bin/contributor-agent.sh`) now export the identical dedicated state dir `${XDG_STATE_HOME:-$HOME/.local/state}/hive/agent-cwd` with `mkdir -p`, so local and container stay aligned.
- Relay relaunch (`launchCommandWithCwd`) still prefers `HIVE_AGENT_CWD` with `process.cwd()` fallback — unchanged and correct.
- #4046 invariant holds: the state dir is outside the task lifecycle, never deleted/recreated.
- Not a git repo (avoids #4168 repo-root adoption) and not $HOME (keeps unattended skip-permissions agent off sensitive home material).
- `node --test bin/contributor-relay.test.js` → 120/120, including the new regression "a relaunch does not root the agent at $HOME". `bash -n` and `just --list` pass.

PR #4176, opened 2026-08-19
✨ sso: widen session cookie domain to .kubestellar.io and add whoami endpoint (#4171)

### LINKED ISSUE #4171: SSO: dibs.kubestellar.io cannot see the hub session - cookie scope + missing whoami endpoint
## Problem

Signing into hive hub (hive.kubestellar.io) does not sign you into the sibling product **Dibs** (dibs.kubestellar.io). A hub user who is logged in visits dibs and is treated as anonymous.

Dibs's auth bridge (github.com/kubestellar/dibs, `pkg/auth`) deliberately has no login of its own: it reads the hub session cookie `hive_hub_user` off the incoming request and calls back to `GET {HUB}/api/saas/whoami` with that cookie, expecting `200 {"username","display_name","email","avatar_url"}` for a valid session, or 401.

Two hub-side gaps block this:

1. **Cookie scope.** The hub session cookie is minted with `Domain=.hive.kubestellar.io` (oauth.go, `mintSessionCookies` / `handleLogout`), so the browser never sends it to the sibling subdomain `dibs.kubestellar.io`. It needs `Domain=.kubestellar.io` (the parent domain), while local/dev hosts keep host-only cookies, logout clears the widened cookie, and pre-existing `.hive.kubestellar.io`-scoped sessions keep working during rollout.

2. **No whoami endpoint.** The hub has `/api/auth/user` (display-oriented payload) but no `GET /api/saas/whoami` returning the stable identity key (GitHub login, or the canonical `provider:sub` form for OIDC users) plus display_name/email/avatar from the enriched SaaSUser record.

## Proposed fix

- Scope `hive_hub_user` to `.kubestellar.io` when the request host is under kubestellar.io; keep host-only cookies for localhost/dev; clear both the new and the legacy domain on logout; accept either cookie copy during the transition (browsers may briefly hold two).
- Add `GET /api/saas/whoami` (200/401 JSON, `Cache-Control: no-store`, GET only) reusing the existing session verification helpers.
- Document the security tradeoff of widening the cookie to all *.kubestellar.io subdomains (hosted spoke dashboards on *.hive.kubestellar.io already receive it today; the widening adds sibling first-party products only).

Danathar:
## Summary

Addresses #4171: allows sibling products (such as Dibs at `dibs.kubestellar.io`) to participate in single sign-on with Hive Hub (`hive.kubestellar.io`).

### Key Changes

1. **Session Cookie Domain Scoping**:
   - Updates `mintSessionCookiesForHost` to mint `hive_hub_user` with `Domain=.kubestellar.io` when the request host is under `kubestellar.io` (e.g. `hive.kubestellar.io`, `dibs.kubestellar.io`, `*.hive.kubestellar.io`).
   - Keeps host-only cookies (`Domain=""`) for localhost/dev environments.
   - Updates `handleLogout` to clear both the new `.kubestellar.io` domain and legacy `.hive.kubestellar.io` domain with `MaxAge: -1`.
   - Accepts either cookie copy during rollout transition across `getRealAuthUser`, `handleAuthUser`, `handleDashboard`, and `isNonBrowserAPIRequest`.

2. **`GET /api/saas/whoami` Endpoint**:
   - Implements `GET /api/saas/whoami` (`Cache-Control: no-store`, `Content-Type: application/json`, GET only).
   - Returns 200 with `{"username", "display_name", "email", "avatar_url"}` for verified sessions.
   - Returns stable identity key for `username` (bare GitHub login for GitHub users, canonical `provider:sub` for OIDC users).
   - Populates `display_name`, `email`, and `avatar_url` from the enriched `SaaSUser` record (with fallback to derived GitHub avatar for GitHub users).
   - Returns 401 for unauthenticated or forged requests and 405 for non-GET methods.

3. **Documentation & Specs**:
   - Updates `src/pkg/hub/static/openapi.yaml` and `src/docs/api-reference.md`.
   - Documents security tradeoffs in `oauth.go`.

4. **Testing**:
   - Adds unit and regression tests in `src/pkg/hub/dibs_sso_whoami_test.go` covering host domain scoping, dual-cookie transition acceptance, logout multi-domain clearing, and `/api/saas/whoami` endpoint behavior.

Fixes #4171

— hive: backend=agy

---
Gemini 3.7 flash high reasoning

### DIFF
--- a/src/docs/api-reference.md
+++ b/src/docs/api-reference.md
@@ -424,6 +424,7 @@ always resolved server-side from the validated token.
 | `GET` | `/api/auth/callback` | Hub handler-specific | OAuth Callback | `pkg/hub/oauth.go:38` |
 | `GET` | `/api/auth/user` | Hub handler-specific | Auth User | `pkg/hub/oauth.go:39` |
 | `POST` | `/api/auth/logout` | Hub handler-specific | Logout | `pkg/hub/oauth.go:40` |
+| `GET` | `/api/saas/whoami` | Hub handler-specific | Whoami SSO Identity | `pkg/hub/saas.go:500` |
 | `GET` | `/api/openrouter/connect/start` | Hub auth | Hub Open Router Start | `pkg/hub/openrouter.go:49` |
 | `GET` | `/api/openrouter/qr` | Hub auth | Hub Open Router QR | `pkg/hub/openrouter.go:50` |
 | `GET` | `/api/openrouter/models` | Hub auth | Hub Open Router Models | `pkg/hub/openrouter.go:51` |
--- a/src/pkg/hub/dibs_sso_whoami_test.go
+++ b/src/pkg/hub/dibs_sso_whoami_test.go
@@ -1,0 +1,383 @@
+package hub
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// TestSessionCookieDomainForHost verifies host-based cookie domain resolution.
+func TestSessionCookieDomainForHost(t *testing.T) {
+	cases := []struct {
+		host string
+		want string
+	}{
+		{"hive.kubestellar.io", ".kubestellar.io"},
+		{"dibs.kubestellar.io", ".kubestellar.io"},
+		{"hosted-hive-1.hive.kubestellar.io", ".kubestellar.io"},
+		{"kubestellar.io", ".kubestellar.io"},
+		{"hive.kubestellar.io:443", ".kubestellar.io"},
+		{"dibs.kubestellar.io:8443", ".kubestellar.io"},
+		{"localhost", ""},
+		{"localhost:8080", ""},
+		{"127.0.0.1", ""},
+		{"127.0.0.1:3000", ""},
+		{"example.com", ""},
+		{"evil.kubestellar.io.attacker.com", ""},
+		{"", ""},
+	}
+
+	for _, tc := range cases {
+		got := sessionCookieDomainForHost(tc.host)
+		if got != tc.want {
+			t.Errorf("sessionCookieDomainForHost(%q) = %q, want %q", tc.host, got, tc.want)
+		}
+	}
+}
+
+// TestMintSessionCookiesDomainScoping verifies that cookie minting scopes to .kubestellar.io
+// for kubestellar.io hosts and host-only for localhost/dev.
+func TestMintSessionCookiesDomainScoping(t *testing.T) {
+	s := f10f15Server(t)
+	mkUser(t, "alice")
+
+	// 1. Host under kubestellar.io
+	recProd := httptest.NewRecorder()
+	if !s.mintSessionCookiesForHost(recProd, "hive.kubestellar.io", "github:alice") {
+		t.Fatal("mintSessionCookiesForHost failed for hive.kubestellar.io")
+	}
+	cookiesProd := recProd.Result().Cookies()
+	var prodCookie *http.Cookie
+	for _, c := range cookiesProd {
+		if c.Name == "hive_hub_user" {
+			prodCookie = c
+			break
+		}
+	}
+	if prodCookie == nil {
+		t.Fatal("no hive_hub_user cookie minted for production host")
+	}
+	if prodCookie.Domain != ".kubestellar.io" && prodCookie.Domain != "kubestellar.io" {
+		t.Errorf("prod cookie Domain = %q, want %q", prodCookie.Domain, ".kubestellar.io")
+	}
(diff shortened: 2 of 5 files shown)


clubanderson:
## Review: NOT merging — superseded by #4174, which already landed on `v4`

I reviewed this end-to-end for the auth/security surface. The logic is broadly sound, but **merging it would be a regression**, because commit `aaa2b564` (PR #4174, "✨ feat(hub): SSO with sibling products — widen session cookie to .kubestellar.io and add /api/saas/whoami (#4171)") already implements this exact feature on `v4`, and it does so more safely. That is also why this PR is `CONFLICTING` — 7 conflict hunks in `oauth.go` / `saas.go`, all in the same functions.

### Where `v4` (#4174) is strictly better

| Concern | This PR (#4176) | Already on `v4` (#4174) |
|---|---|---|
| Domain derivation | Hard-codes `sessionCookieDomain = ".kubestellar.io"` and `legacySessionCookieDomain` as separate constants | `sessionCookieParentDomain()` derives the parent from `hubCanonicalHost`, so scope and canonical host can never drift |
| Legacy cookie cleanup | Only on **logout**. Every user keeps two `hive_hub_user` jar entries (legacy `.hive.kubestellar.io` + new `.kubestellar.io`) indefinitely until they explicitly log out | Expires the legacy-scoped copy **in the same response as the mint**, so a fresh login converges to exactly one cookie |
| Dual-cookie acceptance | Four hand-rolled `for _, c := range r.Cookies()` loops copy-pasted into `getRealAuthUser`, `handleAuthUser`, `handleDashboard`, `isNonBrowserAPIRequest` | Single shared `hubSessionCookieValues(r)` helper — one place to audit, no drift risk |
| F4 audit rationale | Deletes the detailed F4 comment block and replaces it with a 6-line summary that loses the spoke-proxy trust-boundary reasoning and the `isSameOriginAsHub` containment argument | Keeps the full F4 analysis and extends it to explain precisely why widening one level does **not** change the tenant-side risk profile |

### Findings on this PR's own code (moot now, but recorded)

1. **`mintSessionCookies` shim is a trap.** The compatibility wrapper hard-codes `mintSessionCookiesForHost(w, "hive.kubestellar.io", ...)`, which unconditionally yields `Domain=.kubestellar.io`. It currently has no non-test callers, but any future caller on a dev host would emit a `Domain` the browser rejects → silent session loss with no error. `v4` avoids this by threading `*http.Request` through `mintSessionCookies` directly.
2. **Dead `405` branch.** `handleWhoami` is registered as `mux.HandleFunc("GET /api/saas/whoami", ...)`, so Go 1.22's ServeMux answers non-GET itself — the handler's own 405 path is unreachable in production, and the real 405 ships without `Cache-Control: no-store`. The test asserting 405 passes only because it calls the handler directly, so it does not test the deployed behaviour.
3. **No CORS on `/api/saas/whoami`.** The handler sets no `Access-Control-Allow-Origin` / `-Credentials`, and `isSameOriginAsHub` only admits `hubCanonicalHost`/localhost. `dibs.kubestellar.io` therefore cannot read this response from JS. Fail-closed (good), but it does not deliver the stated SSO use case as written.
4. **Cookie tossing / identity selection.** Taking the *first verifiable* cookie among duplicates means a host that can set a `Domain=.kubestellar.io` cookie could get its own valid session picked over the victim's. `v4` has the same shape, so this is a pre-existing class rather than something introduced here — but it is worth tracking as a follow-up (bind the accepted cookie to the jar entry, or prefer the most-recently-minted claim).

### Things that were correct

- `sessionCookieDomainForHost` is not over-broad: it only ever returns `.kubestellar.io` or `""`, so an attacker-controlled `Host` header cannot widen scope. The suffix check correctly rejects `evil.kubestellar.io.attacker.com`, and the test table covers that case plus ports, bare apex, localhost, and `127.0.0.1`. Good coverage.
- Cookie hardening (`HttpOnly`, `Secure`, `SameSite=Lax`) is preserved, and the test asserts it — `SameSite=Lax` is right here since the hub and Dibs are same-site.
- Logout clears both domains with `MaxAge: -1` and revokes every verifiable cookie presented, not just the first.
- No sensitive leakage in the whoami payload beyond the caller's own profile; the stable-identity shape (bare login for GitHub, `provider:sub` otherwise) matches `v4`.
- OpenAPI and `api-reference.md` were updated in step with the implementation.

### Disposition

Closing as superseded. `/api/saas/whoami` and `.kubestellar.io` cookie scoping are already live on `v4` via #4174 — please verify against `origin/v4` and open a targeted follow-up if any gap remains. Findings (2) and (4) above apply to the merged implementation too and are worth a small follow-up PR.

clubanderson:
Superseded by #4174 (commit aaa2b564), already merged to `v4`. See detailed review above.

Issue #4188, opened 2026-08-19
✨ feature: support Podman as a first-class standalone runtime alongside Docker


Danathar:
## Summary and compatibility contract

Add Podman as a first-class runtime for a standalone, single-host Hive deployment.

This is additive. **Docker Engine and Docker Compose must remain fully supported, functional, tested, and the default unless an operator explicitly selects Podman.** Podman must be supported alongside Docker in `v4` and carried forward into successor versions. Removing or deprecating either runtime requires a separate compatibility decision.

This issue is about running **Hive itself** with Podman. The contributor relay already has separate Podman support through #1935, #1936, #2535, and #2599.

> [!IMPORTANT]
> **This is a coordination-only umbrella. Do not assign #4188 as one Hive task or expect one contributor/agent to complete it within the 30-minute task window.** Work must be claimed through a linked, narrowly scoped issue below. Partial PRs use `Part of #4188`; only final integration closes this parent.

## Linked 30-minute work items

These first-wave issues are intentionally independent and ready to claim in parallel:

- [x] #4199 — Capture rootless startup and exit-77 behavior.
- [x] #4200 — Verify the rootful egress-gate baseline.
- [x] #4201 — Evaluate explicit Compose-provider selection.
- [x] #4202 — Evaluate minimal Quadlet container and pod units.
- [x] #4203 — Evaluate Quadlet `.kube` reuse.
- [x] #4204 — Snapshot Docker standalone deployment invariants.
- [x] #4205 — Add explicit runtime selection with Docker as the default.
- [x] #4206 — Define one standalone image-reference source of truth.
- [x] #4207 — Add Podman engine, root-mode, and cgroup preflight checks.
- [x] #4208 — Add subordinate-ID, graphroot, and networking preflight checks.
- [x] #4209 — Add SELinux, mount, secret, and port preflight checks.
- [x] #4210 — Define Hive-owned resources and a safe cleanup contract.
- [x] #4211 — Map CI runners for rootful, rootless, SELinux, and `arm64`.

Each child has one deliverable, explicit non-goals, and a 30-minute boundary. If new work is discovered, open another `Part of #4188` issue rather than expanding the assigned issue.

## Second wave: implementation slices

The security and lifecycle spikes have closed, so these are now live work. Same rule as the first wave: **claim the linked child, not this bullet.**

- [x] Publish the supported rootful/rootless and enforcing/advisory matrix — #4324
[...]

Issue #4211, opened 2026-08-20
🧪 Podman: map CI runners for rootful, rootless, SELinux, and arm64


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

The final CI matrix depends on facilities that ordinary hosted runners may not provide, especially rootful containers, SELinux enforcement, nested user namespaces, and `arm64`.

## 30-minute boundary

Inventory and recommend CI lanes only. Do not implement all workflows.

## Deliverable

Inspect current Hive workflows and available runner capabilities, then map each required Podman test to a concrete runner or repeatable release-qualification environment.

## Acceptance criteria

- Rootful, rootless, SELinux-enforcing, and `amd64`/`arm64` coverage are considered separately.
- The result identifies which checks can run on ordinary hosted CI and which require a self-hosted or release-qualification lane.
- Required permissions and safety boundaries are stated.
- Proposed implementation is split into one small issue per lane rather than one matrix-wide PR.

Issue #4200, opened 2026-08-20
🧪 Podman: verify rootful egress-gate baseline


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

The rootless result needs a known-good Podman baseline. Rootful Podman should demonstrate the same forced-proxy behavior that Hive expects from its enforcing standalone path.

## 30-minute boundary

Verify only the rootful security baseline. Do not create permanent deployment assets or change Docker behavior.

## Deliverable

Run Hive under rootful Podman with `NET_ADMIN` and capture evidence for entrypoint startup, iptables redirect creation, capability propagation across the privilege drop, and the proxy `SO_MARK` path.

## Acceptance criteria

- Exact commands, Podman version, image reference, and relevant logs are recorded.
- A missing capability or iptables dependency fails closed rather than silently becoming advisory.
- The result identifies any remaining bypass-resistance test that needs its own follow-up.
- No Docker configuration is changed.

## Stop conditions

If rootful Podman cannot be exercised safely in the available environment, record the missing facility and stop rather than weakening the test.

PR #4243, opened 2026-08-20
🐛 fix(contribute): never offer a tracker/umbrella issue as a task

### LINKED ISSUE #4205: 🧩 Podman: add explicit runtime selection with Docker default
## Parent

## Why

Hive needs deterministic runtime selection before runtime-specific implementations can safely coexist.

## 30-minute boundary

Implement only selection, validation, and reporting. Do not add Podman deployment assets.

## Deliverable

Add a small runtime-selection layer for the agreed interface, with Docker as the default and Podman accepted only when explicitly selected.

## Acceptance criteria

- Unset selection preserves current Docker behavior.
- `docker` and `podman` are validated values; unknown values fail clearly.
- The selected runtime is reported in command output or diagnostics.
- No silent fallback occurs after a missing binary or failed command.
- Unit or shell-contract tests cover default, explicit Docker, explicit Podman, and invalid input.
- Existing Docker commands remain functional.

Danathar:
## Problem

An umbrella issue is coordination-only: its children carry the work and are queued independently. Handing the parent to one contributor cannot succeed — by construction, only final integration closes it.

#4188 was offered anyway. The contributor spent the full 30-minute `MAX_TASK_DURATION` budget and had a failure booked against an issue that says, in its own body:

> [!IMPORTANT]
> **This is a coordination-only umbrella. Do not assign #4188 as one Hive task or expect one contributor/agent to complete it within the 30-minute task window.**

The agent behaved correctly throughout — it triaged all 13 children, picked the narrowest implementable one (#4205), and scoped its PR `Fixes #4205` / `Part of #4188` rather than claiming to close the parent. The queue was the problem.

**Backend-independent.** This is queue behaviour, not agent behaviour: any contributor on any CLI offered such a parent loses the same 30 minutes.

## Two independent gaps

**1. The queue never read the flag.** `github.Issue.IsTracker` has been computed at enumeration and serialized as `is_tracker` all along. `selectTask` referenced it **zero** times — `scheduler.go:613` consumed it for a `[TRACKER]` prompt annotation and nothing else. A correctly-flagged umbrella was still fully assignable.

**2. Detection could not see the common case.** `isTracker()` looked only at a `[Tracker]` title prefix or a `meta-tracker` label. #4188 has neither — no labels at all, ordinary `✨ feature:` title — and declares itself an umbrella in the body, above a 13-item child task list.

`issue.GetBody()` is already in hand at the call site (it comes back in the list response), so consulting it costs no extra API calls.

## Why the body rule is the task list, not the callout

Deliberately **structural**, not prose. The "coordination-only umbrella" text is *not* matched — free-text matching is brittle in the direction that hurts. `- [ ] #123` is rendered by GitHub as a tracked task list, so it is a convention an author opts into, and it survives rewording.

**Threshold is 3.** One or two refs appear on ordinary work (a blocked-on note, an acceptance criterion) and must stay claimable; three or more is a tracking list by construction. #4188 has thirteen.

The asymmetry justifies erring slightly liberal: a false positive means one issue is not offered to contributors and an operator can still work it. A false negative is what happened here — 30 minutes burned, a failure booked, and children at risk of duplicate work.

## Why not just label the umbrella?

Reasonable question, and worth answering up front: the label path is not currently usable.

- **`meta-tracker`** — the label `isTracker()` has always checked for — **does not exist** in this repository (`GET /labels/meta-tracker` → 404).
- **`Epic`** (*"Epic tracking issue"*) exists but is applied to **zero** issues.

So the label branch has been effectively dead code here, and there is no established umbrella convention in practice.

It is also not something a contributor can fix from outside: applying a label needs `triage`, and an external contributor has `pull` only. The person best placed to notice an umbrella is being mis-assigned is the contributor losing 30 minutes to it — and they cannot label it.

That is the case for detecting the structure the author already wrote, rather than depending on a label somebody with elevated access has to remember to apply. A maintainer adopting `meta-tracker` (or wiring `Epic` in) would still work and would take precedence — this change does not replace that, it makes the common case work without it.

## Skips are logged

An issue that is never offered should not be a mystery to the operator watching the queue.

## Tests

- `TestIsTracker` — existing title/label cases, extended for the new signature.
- `TestIsTrackerFromBodyTaskList` — pins **both** directions: 3+ refs (incl. cross-repo `owner/repo#1`) detected; two refs, one ref, bare prose mentions of `#101, #102, #103`, and a checklist with no issue refs all stay claimable.
- Verified against the **real #4188 body** fetched from the API — detected.

`go build ./...` clean; `pkg/github` and `pkg/dashboard` both pass.

## Not in scope

`bin/issue-classifier.sh` has its own divergent tracker rule (`[Auto-QA]`, `[Nightly]`, config-driven). Left alone — reconciling the two enumeration paths is a separate decision.

### DIFF
--- a/src/pkg/dashboard/contribute_ws.go
+++ b/src/pkg/dashboard/contribute_ws.go
@@ -4414,6 +4414,29 @@ func (h *ContributeWSHub) selectTask(c *ContributorConnection) *WSMessage {
 			lane, _ := issue["lane"].(string)
 			labels := stringSliceFromAny(issue["labels"])
 			assignees := stringSliceFromAny(issue["assignees"])
+
+			// A tracker/umbrella issue is coordination-only: its children carry
+			// the work and are queued independently, so handing the parent to one
+			// contributor is never right. The enumerator has flagged these all
+			// along (github.Issue.IsTracker) and this queue simply never read it —
+			// scheduler.go consumed the flag for a "[TRACKER]" prompt annotation
+			// and nothing else.
+			//
+			// Live cost of the omission: kubestellar/hive#4188 ("coordination-only
+			// umbrella. Do not assign #4188 as one Hive task", per its own body,
+			// above a 13-item child task list) was offered to a contributor, which
+			// spent the full 30-minute MAX_TASK_DURATION budget and booked a
+			// failure on an issue that only final integration can close. The agent
+			// behaved correctly throughout — it triaged the children and shipped
+			// one — but no agent can "complete" a parent like this.
+			//
+			// Logged rather than skipped silently: an issue that is never offered
+			// should not be a mystery to the operator watching the queue.
+			if isTracker, _ := issue["is_tracker"].(bool); isTracker {
+				h.logger.Info("[contribute-ws] skipping tracker/umbrella issue",
+					"repo", repo.Full, "number", number, "title", title)
+				continue
+			}
 			if requestedRole != "" && !h.issueMatchesAgentRole(requestedRole, title, labels, lane) {
 				continue
 			}
--- a/src/pkg/github/client.go
+++ b/src/pkg/github/client.go
@@ -572,7 +572,7 @@ func (c *Client) fetchIssues(ctx context.Context, repo string, now time.Time) (a
 			UpdatedAt:  issue.GetUpdatedAt().Time,
 			AgeMinutes: ageMinutes,
 			URL:        issue.GetHTMLURL(),
-			IsTracker:  isTracker(issue.GetTitle(), labels),
+			IsTracker:  isTracker(issue.GetTitle(), labels, issue.GetBody()),
 		})
 	}
 
@@ -1082,7 +1082,36 @@ func (c *Client) isExempt(labels []string) bool {
 	return false
 }
 
-func isTracker(title string, labels []string) bool {
+// trackerTaskListRe matches a Markdown task-list item whose subject is an issue
+// reference — "- [ ] #4199", "* [x] owner/repo#12". This is deliberately
+// STRUCTURAL rather than a prose match: GitHub renders these lists specially
+// (the "N of M" tracked-task-list widget), so the pattern is a first-class
+// convention an author opts into, not a phrase they happened to write. Prose
+// gets reworded; structure does not.
+var trackerTaskListRe = regexp.MustCompile(`(?m)^\s*[-*]\s*\[[ xX]\]\s*(?:[\w.-]+/[\w.-]+)?#\d+`)
+
+// trackerTaskListMin is how many task-list issue references a body needs before
+// the issue counts as a tracker. One or two can legitimately appear on ordinary
+// work (a blocked-on note, an acceptance criterion); three or more is a
+// tracking list by construction.
+const trackerTaskListMin = 3
+
+// isTracker reports whether an issue is coordination-only — an umbrella whose
+// children carry the actual work — and so must never be handed out as a single
+// task.
+//
+// body is consulted because the title/label markers below miss the common case.
+// kubestellar/hive#4188 carried no labels and an ordinary "✨ feature:" title
+// while its body said, in a callout, "This is a coordination-only umbrella. Do
+// not assign #4188 as one Hive task", above a 13-item task list of children. It
+// was offered to a contributor anyway, which burned the full 30-minute task
+// budget and booked a failure on work that cannot be completed by one agent
+// (only final integration closes such a parent).
+//
+// The callout text itself is NOT matched — free-text matching is brittle in the
+// direction that hurts. The task list is enough, and it is what GitHub itself
+// treats as the tracking signal.
+func isTracker(title string, labels []string, body string) bool {
 	if strings.HasPrefix(title, "[Tracker]") {
 		return true
 	}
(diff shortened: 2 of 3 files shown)


Danathar:
Some context on how I hit this, written out fully in case it's useful to anyone
who hasn't looked at the contribute queue before. I authored #4188 and all of
its children, and then hit this bug as a contributor running an agent against
the hive — so I saw both ends of it.

## What an "umbrella issue" is here

#4188 is a coordination issue. It doesn't contain any work you can do; it exists
to hold thirteen smaller issues (#4199–#4211) that each have one deliverable and
a 30-minute boundary. Its body says so directly:

> **This is a coordination-only umbrella. Do not assign #4188 as one Hive task
> or expect one contributor/agent to complete it within the 30-minute task
> window.** Work must be claimed through a linked, narrowly scoped issue below.
> Partial PRs use `Part of #4188`; only final integration closes this parent.

So there are two kinds of issue in play: **children**, which are real work a
contributor can finish, and the **parent**, which is a checklist. The parent is
never something one person sits down and completes.

## What actually happened

The contribute queue offered **#4188 itself** to my agent as a task.

The agent behaved well, for what it's worth. It read the parent, listed all
thirteen children, checked which were unassigned and had no existing PR, picked
the narrowest one (#4205), implemented it, and opened #4242 scoped correctly as
`Fixes #4205` / `Part of #4188` — deliberately *not* claiming to close the
umbrella. That's exactly what you'd want a human contributor to do.

But the task it had been given was "#4188", and #4188 cannot be finished. So the
task ran until it hit `MAX_TASK_DURATION_MS` — a hard 30-minute ceiling in the
relay — and was recorded as a failure. Thirty minutes of compute, and a failure
booked against an issue that was never completable in the first place.

## Why the queue offered it

Two separate gaps, and both had to line up.

**First: there is already a flag for this, and the queue never read it.**
`github.Issue` has an `IsTracker` field. It's computed when issues are
enumerated and serialised as `is_tracker`. I went looking for where the
contribute queue consults it before handing out work — it doesn't. Not once.
The only consumer anywhere is `scheduler.go`, which uses it to append a
`[TRACKER]` note to a prompt. So even an issue correctly flagged as a tracker
was still fully assignable to contributors.

**Second: the flag wouldn't have been set on #4188 anyway.** `isTracker()`
looked at exactly two things: a `[Tracker]` prefix on the title, or a
`meta-tracker` label. #4188 has neither — no labels at all, and an ordinary
`✨ feature:` title. Everything that marks it as an umbrella lives in the body,
which was never examined.

## What this PR changes

Two things, matching the two gaps.

1. The contribute queue now skips issues flagged `is_tracker`, and logs when it
   does — an issue that silently never gets offered shouldn't be a mystery to
   whoever's watching the queue.

2. `isTracker()` now also looks at the body, specifically for a task list of
   issue references (`- [ ] #4199`), requiring at least three.

On that second point, I want to be explicit about a choice: **the PR does not
match the "do not assign" sentence**, even though it's the clearest statement of
intent in the whole issue. Matching prose is fragile — reword the sentence and
detection silently stops working. The task list is *structure*: GitHub renders
it specially, an author opts into it deliberately, and it survives rewording.

The threshold is three because one or two `- [ ] #N` lines show up on ordinary
issues (a blocked-on note, an acceptance criterion) and those must stay
claimable. #4188 has thirteen, so it's caught with room to spare. I verified
against the real #4188 body pulled from the API, not just a test fixture.

The trade is deliberately asymmetric. A false positive means one issue isn't
offered to contributors, and a maintainer can still work it by hand. A false
negative is what happened here: half an hour burned, a failure recorded, and
children potentially picked up twice.

## Three things only a maintainer can decide

These are outside what a contributor can do, so I'm flagging rather than
attempting them.

**1. There's no umbrella label convention.** `isTracker()` checks for
`meta-tracker`, which doesn't exist in this repo (`GET /labels/meta-tracker`
returns 404). `Epic` exists — described as "Epic tracking issue" — but is
applied to zero issues. So the label branch has effectively been dead code here.
If you adopt a label, it's strictly better than my heuristic: it can't be broken
by someone editing an issue body. I can't set that up from outside; applying a
label needs `triage` and an external contributor has `pull` only.

**2. The two enumeration paths disagree.** The Go path uses
`[Tracker]`/`meta-tracker`. `bin/issue-classifier.sh` uses `[Auto-QA]`/
`[Nightly]`, config-driven. They've drifted apart. I deliberately left the shell
one alone here rather than quietly changing both — reconciling them seemed like
your call, not something to slip into a bug fix.

**3. Nothing documents how to file an umbrella.** `CONTRIBUTING.md` doesn't
mention it. After this PR the body format is load-bearing: keep the `- [ ] #N`
checklist and it's detected; switch to prose ("children: #4199, #4200…"), which
reads identically to a human, and it isn't. Nobody filing their first umbrella
has any way to know that. I'm happy to write that section if it'd be welcome.

## One thing to watch

If anyone moves umbrellas to GitHub's **native sub-issues**, this detection
won't see them — it reads body text, not `sub_issues_summary`. Nothing in the
repo uses sub-issues today (I checked a sample; all zero or null), so it isn't a
problem now, but it would need a small follow-up *before* such a migration
rather than after.

## What still isn't automated, and probably shouldn't be

This PR stops the parent being handed out as work. It does **not** mark the
parent complete when its children finish, and I don't think it should. #4188's
own body lists the thirteen children as the **first wave**, with more issues to
be filed afterward for the remaining slices. Anything that auto-closed the
parent when the checkboxes filled would close it roughly halfway through the
real work. "Only final integration closes this parent" is a human judgement, and
leaving it manual seems right.

PR #4265, opened 2026-08-20
🐛 fix(relay): codex completion detection — both directions

Danathar:
## Problem

`hasCompletionMarker` for codex was a prose word match:

```js
hasCompletionMarker = /completed|done|finished/i.test(text) ||
  detectNoWorkVerdict(text.split('\n')) !== null;
```

codex writes its completion summary in whatever words the work calls for. So whether a task could finish depended on which English word it happened to reach for.

## Observed live

A task ran to completion and opened #4259 **ready for review**, then summarised itself:

```
• Opened ready-for-review PR #4259 (https://github.com/kubestellar/hive/pull/4259).
  - Conclusion: direct .kube reuse is not viable; native Quadlet .container/.pod units are recommended.
  - Added the measured compatibility report and documentation index link.
  - Commit c8ae4ddf includes a matching Signed-off-by trailer.
  - DCO, PR-content verification, and initial build checks pass; remaining test shards are pending.
  - Branch is pushed and clean.
─ Worked for 6m 22s ─────────────────────────────────────────────
› Ask Codex to do anything
  gpt-5.6-sol xhigh · /home/dev/.local/state/hive/agent-cwd
```

None of `completed` / `done` / `finished` appear — and there is no `no_work_needed` verdict either, because it shipped a PR. Running the shipped classifier against that exact captured pane:

```
hasIdlePrompt      : true
isWorking          : false
hasCompletionMarker: false   ← blocks completion
=> PANE_STATE_WORKING (fallback)
```

The agent was sitting idle at its prompt with a green PR open. The relay could not see it, kept renewing the lease, and the task was headed for a stall/`MAX_TASK_DURATION` failure.

## Same root cause as #4182, opposite direction

| | prose said | effect |
|---|---|---|
| #4182 (agy) | "…with **writing** HIVE_GITHUB_TOKEN…" | finished pane looked **busy** |
| this (codex) | no completion word at all | finished pane looked **unfinished** |

Both are the same mistake: inferring machine state from English.

## Fix

Set `hasCompletionMarker = true`, as `copilot`, `goose`, `agy` and `bob` already do.

This is safe because it is redundant for codex, not load-bearing:

```js
if (isWorking) return PANE_STATE_WORKING;
if (hasIdlePrompt && hasCompletionMarker) return PANE_STATE_IDLE_COMPLETE;
```

`isWorking` reads codex's own status row — an in-flight turn renders `esc to interrupt`, an idle one does not — and **returns early**. By the time `hasCompletionMarker` is consulted the agent is already known not to be working. `hasIdlePrompt` cannot carry the distinction either, because codex draws its `›` input line while working too. So gating on a completion *word* added nothing except a way to miss finished work.

### Why not detect an open PR instead

Considered and rejected. A PR URL on screen is not evidence of finishing — codex prints other PRs' URLs *while working* (during an umbrella triage it ran `gh pr view` against three unrelated PRs to check whether children were already covered). And it would strand `no_work_needed` completions, which legitimately ship no PR. Deciding *"is it finished"* and reporting *"what did it ship"* are separate jobs; `detectPRURL` already does the latter, attached to the completion rather than gating it.

## Tests

`node --test bin/contributor-relay.test.js` → **123/123**.

New case built from the captured pane above. Reverting only `bin/contributor-relay.sh`:

```
ok    codex no-work verdict is COMPLETE despite stale activity in scrollback
FAIL  a finished codex turn is COMPLETE even when its summary avoids completion words
ok    codex still reads as WORKING while activity is in the tail
```

The two existing codex cases passing in both directions is the point: this does not loosen busy detection or disturb the no-work verdict path.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -1578,8 +1578,28 @@ function classifyTmuxPane(text) {
     const codexTail = text.split('\n').slice(-15).join('\n');
     // Same marker mismatch as getCLIState(): '›' (U+203A), not '>'.
     hasIdlePrompt = /codex>|›|>\s*$/.test(text);
-    hasCompletionMarker = /completed|done|finished/i.test(text) ||
-      detectNoWorkVerdict(text.split('\n')) !== null;
+    // Not a prose match. codex writes its own completion summary in whatever
+    // words the work calls for, and requiring "completed|done|finished" makes
+    // finishing a task depend on which English word it happened to reach for.
+    //
+    // Observed live: a task that ran to completion and opened
+    // kubestellar/hive#4259 ready for review summarised itself as "Opened
+    // ready-for-review PR #4259 … Conclusion: direct .kube reuse is not viable
+    // … Branch is pushed and clean … Worked for 6m 22s". None of the three
+    // words appear, and there is no no_work_needed verdict either (it shipped a
+    // PR), so hasCompletionMarker was false, the IDLE_COMPLETE arm could not be
+    // reached, and the pane fell through to PANE_STATE_WORKING with the agent
+    // sitting idle at its prompt.
+    //
+    // The same reliance on prose in the other direction produced #4182 for agy.
+    //
+    // codex's real state signal is its status row, which isWorking below reads:
+    // an in-flight turn renders "esc to interrupt", and an idle one does not.
+    // hasIdlePrompt cannot carry that distinction here — codex draws its "›"
+    // input line while it is working too — so gating completion on a completion
+    // WORD added nothing except a way to miss finished work. copilot, goose,
+    // agy and bob all set this true for the same reason.
+    hasCompletionMarker = true;
     // Codex also signals an in-flight turn through its status row ("Working",
     // "esc to interrupt") without ever printing a tool verb.
     isWorking = /running|executing|thinking|\bworking\b|esc to interrupt/i.test(codexTail);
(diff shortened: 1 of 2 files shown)


Danathar:
Pushed a second commit — the sibling half of the same bug, found while checking whether the first fix was sufficient. It wasn't.

**`isWorking` had the #4182 shape, unfixed.** It matched bare verbs against the tail:

```js
/running|executing|thinking|\bworking\b|esc to interrupt/i
```

codex narrates in English, including in the summary it prints when a turn *finishes*. A summary saying "while running the contract tests" pins a finished pane to WORKING — the same failure #4182 fixed for agy, just waiting for a summary to phrase it that way. #4259's summary missed it by luck: it said "Worked for 6m 22s", and `\bworking\b` doesn't match "Worked".

**Captured from live panes**, codex's markers are:

```
working -> "• Working (46s • esc to interrupt)"  AND  "› Ask Codex to…"
idle    ->                                            "› Ask Codex to…"
```

So `esc to interrupt` is the only discriminator — the `›` line is drawn in both states, which is why `hasIdlePrompt` can't carry this and why the verb list was doing the work.

**Dropping the verbs outright was wrong, and the existing suite caught it.** `codex still reads as WORKING while activity is in the tail` guards a real case: a stale `HIVE_VERDICT: no_work_needed` higher in the scrollback must not be reported as the completion of a turn that has since started new work. The verbs were what prevented that.

The fix keeps that protection without the prose exposure, by anchoring to codex's own bullet chrome for an in-flight tool call — `• Running <cmd>`, against `• Ran <cmd>` once finished:

```
"• Running gh issue view 4066"     -> chrome, in flight   -> WORKING
"- While running the tests I ..."  -> prose, in a summary -> not
```

**125/125**, including that pre-existing test, plus new cases pinning both directions. Verified against both real captured panes: the finished pane that opened #4259 classifies idle, the mid-turn pane classifies working.

Retitled since it's no longer just the completion marker.

PR #4288, opened 2026-08-20
🐛 fix(pluk): write the publisher's log, and read the key pluk actually sets

### LINKED ISSUE #4285: 🐛 pluk publisher writes no log file — manager.go omits the shell redirect, so hive-panes and the inception watcher read files that never exist
## Summary

Three stacked defects on the pluk logging path. The first one means nothing
that reads `/var/run/pluk/logs` can work at all today; the other two are
waiting behind it.

Verified against `@kubestellar/pluk@0.8.0` — the version pinned by
`PLUK_VERSION` in `src/Dockerfile:450` — and against `v4` at `df3feca0`.

## Bug 1 — the attach command has no redirect, so no file is ever written

`src/pkg/agent/manager.go:1395`

```go
pipePaneCmd := fmt.Sprintf("%s watch %s --cli=%s", plukPath, agent.tmuxSession, backend)
_ = m.tmuxCmd(agent, "pipe-pane", "-t", agent.tmuxSession, "-o", pipePaneCmd).Run()
```

`pluk watch` emits events on **stdout only**. There is no file-writing code
anywhere in the package — no `createWriteStream`, `appendFile`, `writeFile`, or
`openSync` in any module of the published `dist/`. `tmux pipe-pane` feeds the
pane's output into the command's stdin; it does not route the command's stdout
to a log path, and pluk does not open one itself. So the events are produced
and discarded.

pluk's own `attach` subcommand builds the invocation the way it is meant to
look (`dist/attach.js:172`):

```
PLUK_RUN_DIR=${runDir} ${plukBin} watch ${session} --cli=${cli}${includeRawFlag} >> ${logFile}
```

Hive's version has neither the redirect nor the flag.

**Controlled test.** Two throwaway tmux sessions, same pluk build, same input
lines, run dir pre-created; the only difference is the invocation:

| invocation | result |
| --- | --- |
| `manager.go` form (no redirect, no flag) | **no file created in the run dir** |
| `attach.js` form (redirect + `--include-raw`) | file created — 23 `raw_output`, 1 `state_change`, 3 `tool_call_started` |

## Bug 2 — `raw_output` is gated behind `--include-raw`

Even once a redirect exists, `raw_output` is opt-in. Replaying a real captured
claude agent pane (42,244 bytes) through the two forms:

```
$ pluk watch t --cli=claude               < pane-bytes    # as manager.go runs it
      7 "type":"model_changed"                            # 0 raw_output

$ pluk watch t --cli=claude --include-raw  < pane-bytes
      7 "type":"model_changed"
    713 "type":"raw_output"
```

## Bug 3 — the inception watcher reads a key pluk does not use for this event

`src/pkg/dashboard/inception_watcher.go:579`

```go
case "raw_output":
    line := event.Data["message"]
    if line == "" {
        return
    }
```

`dist/classifier.js:112`

```js
rawOutput(line) {
    this.seq++;
    return createEvent(this.session, this.pane, this.source, this.seq, 'raw_output', { line });
}
```

`raw_output` carries `line`, and only `line` — 713 of 713 events in the replay
above have exactly that one key. pluk *does* use a `message` key, but on
`rate_limit` and `error` events (`classifier.js:58` and `:76`), not on
`raw_output`. `src/deploy/hive-panes.sh:49` reads `data['line']` and is correct.

Related doc fix: `src/docs/agent-logging.md` currently states that `data.line`
and `data.message` are "both read from real pluk output in this codebase" and
advises new consumers to check both keys. For `raw_output` on 0.8.0 that is not
reproducible — the guidance looks like it generalised from the `error` /
`rate_limit` shape. Worth correcting alongside the code, or it will steer the
next consumer the same way.

## Consequences today

- `src/deploy/hive-panes.sh` — the glob `"$LOG_DIR"/hive-*.jsonl` matches
  nothing, so the peer-awareness listing is empty for every agent.
- `src/pkg/dashboard/inception_watcher.go` — `runPlukSubscriber` waits for
  `/var/run/pluk/logs/hive-brainstorm.jsonl` to appear. The `state_change`
  re-kick path and the `bd create --title` interception never run.
- `src/docs/agent-logging.md` describes events landing in
  `/var/run/pluk/logs/<tmux-session>.jsonl`, "appended as the agent produces
  output". With Bug 1 that file does not exist.

## Suggested fix

Mirror `attach.js`: add the `>> <logFile>` redirect in `manager.go` (with
`PLUK_RUN_DIR` set), and fix the key name in `inception_watcher.go`. Bug 3 is
worth fixing on its own merits regardless of what is decided below.

## Open question for maintainers — log volume

`--include-raw` produced 713 events per 42 KB of pane output. Turning it on for
every agent continuously is a real change in disk behaviour for
`/var/run/pluk/logs`, and may be why it was left off. But both consumers are
non-functional without it, so the current state does not look deliberate
either. Options as I see them:

1. Redirect + `--include-raw` for all agents, with rotation/retention.
2. Redirect for all agents; `--include-raw` only where a consumer needs it
   (the `brainstorm` session is the one with a Go subscriber).
3. Redirect + `--filter` to narrow what is written.

Happy to send the patch once the direction is settled.

## What I did not verify

I do not have a running hive deployment, so I have not confirmed in situ that
`/var/run/pluk/logs/` is empty on a live host. I grepped the tree and
`manager.go:1395` is the only thing that invokes `pluk watch`; nothing else
writes to that directory. If something outside the repo — an entrypoint, a
systemd unit, a sidecar — creates those files, then Bug 1 is wrong and the
rest should be re-read in that light. A single `ls -la /var/run/pluk/logs/` on
any running deployment settles it.

## Environment

`v4` at `df3feca0`. pluk `0.8.0`, the version pinned by `PLUK_VERSION` in
`src/Dockerfile`, installed standalone from npm for testing. The pane capture
is from a claude agent in contributor local mode
(`just contribute-hive claude local`) during ordinary operation. The
redirect/no-redirect comparison was run in two throwaway tmux sessions on the
host, not in a deployment.

Danathar:
## What

Fixes all three defects in #4285. Closes #4285.

Verified against `@kubestellar/pluk@0.8.0` — the version `PLUK_VERSION` pins in `src/Dockerfile` — installed from npm and run end to end, not just read.

## 1. The attach command had no redirect, so no file was ever written

`pluk watch` classifies its stdin and prints events to **stdout**. It opens no file of its own — no `createWriteStream`, `appendFile`, `writeFile` or `openSync` in any module of the published package. `tmux pipe-pane` feeds the pane into that command's *stdin* and does nothing with its stdout. So the events were produced and discarded.

Running the two invocations against real pluk, same input, same run dir:

```
===== OLD (pre-#4285): no redirect, no --include-raw =====
files in run dir: []

===== NEW: exactly what plukPipePaneCmd builds =====
files in run dir: [hive-brainstorm.jsonl]
{"...","type":"raw_output","data":{"line":"● Running bd create --title \"Who are the users?\" --actor brainstorm"}}
{"...","type":"tool_call_started","data":{"tool":"Read","input_preview":"● Read src/pkg/agent/manager.go"}}
{"...","type":"raw_output","data":{"line":"● Read src/pkg/agent/manager.go"}}
{"...","type":"raw_output","data":{"line":"✻ Thinking about the redirect"}}
```

And `hive-panes.sh`'s own extractor, run unmodified over that real log, finally prints something:

```
● Running bd create --title "Who are the users?" --actor brainstorm
● Read src/pkg/agent/manager.go
✻ Thinking about the redirect
```

## 2. `raw_output` was gated behind `--include-raw`

`watch.js` sets `includeRaw = opts.includeRaw ?? false`. And `raw_output` is the **only** event type `hive-panes.sh` keeps — it drops every other type before touching the payload. So the redirect on its own would have produced a file that still left peer awareness empty for every agent.

**The issue left this as an open maintainer question on log-volume grounds. I think the repo's own history answers it, which is why I didn't wait.**

Both halves were lost together in #1759, which replaced the Go `pluk-publish` binary — a *publisher*, which wrote the log itself — with the TypeScript `pluk watch`:

```diff
-pipePaneCmd := fmt.Sprintf("%s --session %s --cli %s", plukPath, agent.tmuxSession, backend)
+pipePaneCmd := fmt.Sprintf("%s watch %s --cli=%s", plukPath, agent.tmuxSession, backend)
```

That translated the flags but not the file-writing half the rewrite had moved into the shell. And the dates settle intent:

| | |
|---|---|
| `f6a1db81` **2026-06-23** | `hive-panes.sh` added — reads `raw_output` / `data.line`, nothing else |
| `5aeb24ec` **2026-07-02** | #1759 migration drops the redirect *and* the raw flag |

`hive-panes.sh` predates the migration by nine days and has only ever read `raw_output`, so raw output was flowing when it was written. This restores the contract it was built against rather than choosing a new policy. Pluk's own `attach` reaches the same conclusion from upstream — `attach.js:168` passes `--include-raw` by **default** (opt out via `--no-raw`).

One correction to the issue: **option 2 wouldn't work.** It suggests the flag only where "a consumer needs it (the `brainstorm` session is the one with a Go subscriber)", but `hive-panes.sh` globs `hive-*.jsonl` and needs `raw_output` from *every* agent, not just the one with a Go subscriber.

## 3. The watcher read a key pluk does not set for that event

`classifier.js:114` builds it as `createEvent(..., 'raw_output', { line })`. `handlePlukEvent` read `data["message"]` — a real pluk key, but one belonging to `rate_limit` and `error`, which is evidently where it generalised from.

The consequence was total rather than partial: `line` came back `""` for every `raw_output`, the handler returned immediately, and every bd-create interception and keyword re-poll beneath it was unreachable.

Worth flagging: the existing coverage test drove this arm with `"message"` and asserted nothing about the outcome, so it covered the line without exercising it. That's how the bug survived. Corrected, and backed by tests that assert the effect.

## Also fixed: the log wouldn't have been peer-readable

Not in the issue, but it would have blunted the fix silently.

`hive-panes` is run **by** one agent to read every **other** agent's log, and agent UIDs differ — the shared `node` group is what makes that work. Leaving creation to the shell's `>>` takes whatever umask the pane shell carries, and a `0077` umask yields a `0600` log no peer can read. Same class of silent breakage `permissions_watcher.go` exists to repair for agent config (#3882), and it fails without erroring.

`ensurePlukLogFile` pre-creates each log `0660 dev:node`, matching what `ensurePlukRunDirs` already does one level up, and widens a tight file left by an earlier run. Both the `O_CREATE` mode and the following `Chmod` are needed — `O_CREATE` is umask-filtered and doesn't apply to an existing file.

Every operand is now shell-quoted. The string always went to a shell, but the redirect target is new and embeds the session name, so a name with a space would otherwise split it.

## Deliberately not included

**Rotation.** Nothing caps these files and they grow for the life of the container. It's the remaining half of the issue's option 1 and belongs in its own issue: naively truncating a file that `pipe-pane` holds an append fd on is worse than not rotating, so it needs a deliberate design rather than a rider here. No consumer needs the history — `hive-panes` reads the last 2000 lines, the subscriber seeks to the end — so the gap is bounded in impact. Documented in `agent-logging.md`.

**For review:** restoring `raw_output` puts pane content back into a group-shared file. That is the documented purpose of the directory and the whole premise of `hive-panes`, but it has not actually been happening since 2026-07-02 — so it is a real change to what lands on disk today, and I'd rather flag it than have it noticed after merge.

## Verification

```
$ go test ./pkg/agent/ -short -race -count=1
ok  	github.com/kubestellar/hive/pkg/agent	428.223s
$ go test ./pkg/dashboard/ -short -race -count=1
ok  	github.com/kubestellar/hive/pkg/dashboard	55.359s
```

All eight new tests fail against the pre-fix behaviour with their own diagnostics — I checked rather than assuming:

```
--- FAIL: TestPlukPipePaneCmdRedirectsToTheSessionLog
    #4285: the pipe-pane command has no redirect, so pluk's events go to a stdout
    nobody reads and the log file is never written: /usr/bin/pluk watch hive-brainstorm --cli=claude
--- FAIL: TestPlukPipePaneCmdRequestsRawOutput
    #4285: --include-raw missing. pluk watch defaults includeRaw to false, so
    hive-panes.sh — which keeps only raw_output events — would still show nothing
--- FAIL: TestPlukPipePaneCmdQuotesEveryOperand
    a session name containing shell syntax was not quoted: /usr/bin/pluk watch hive-a; touch /tmp/pwned --cli=claude
--- FAIL: TestEnsurePlukLogFileIsPeerReadableUnderTightUmask
    #4285: log mode = 0600, want 0660 …
--- FAIL: TestEnsurePlukLogFileWidensATightExistingLog
    a pre-existing 0600 log stayed 0600 — peers still cannot read it
--- FAIL: TestPlukRawOutputReadsLineKey
    #4285: a raw_output event carrying … on the "line" key was dropped
--- FAIL: TestPlukRawOutputIgnoresMessageKey
    #4285: raw_output text was read from "message" …
--- FAIL: TestPlukRawOutputParsesQuestionsThroughTheLineKey
    buffered 0 bd create lines, want 3
```

`TestPlukRawOutputIgnoresMessageKey` is there to stop the fix being reverted into "read both keys": `rate_limit`/`error` carry human-readable provider prose, and `bd create --title` appearing inside an error message is not a bd create the agent ran.

## Files

- `src/pkg/agent/pluk_dirs.go` — `plukPipePaneCmd`, `plukSessionLogPath`, `ensurePlukLogFile`
- `src/pkg/agent/manager.go` — attach through the builder, pre-create the log
- `src/pkg/dashboard/inception_watcher.go` — `data["message"]` → `data["line"]`
- `src/pkg/agent/pluk_publisher_test.go`, `src/pkg/dashboard/pluk_raw_output_key_test.go` — new tests
- `src/pkg/dashboard/inception_watcher_coverage_test.go` — the case that encoded the bug
- `src/docs/agent-logging.md` — corrects the "check both keys" advice, records why the redirect exists and that nothing rotates

## What I did not verify

Same caveat as the issue: I have no running hive deployment, so I have not confirmed in situ that `/var/run/pluk/logs/` is empty on a live host. Everything above is from the published pluk package, the repo's history, and running the two invocations side by side locally. A single `ls -la /var/run/pluk/logs/` on any running deployment confirms the premise.

### DIFF
--- a/src/docs/agent-logging.md
+++ b/src/docs/agent-logging.md
@@ -68,6 +68,28 @@ Session names follow Hive's `hive-<agent-name>` convention (e.g.
 `hive-brainstorm.jsonl` for the `brainstorm` agent). Each line is one JSON
 object, newline-delimited (JSONL), appended as the agent produces output.
 
+The file is written by a shell redirect, not by pluk. `pluk watch` classifies
+its stdin and prints events to stdout; it opens no file of its own. `tmux
+pipe-pane` feeds the pane into that command's stdin but does nothing with its
+stdout, so the publisher invocation in `pkg/agent/manager.go` has to end in
+`>> /var/run/pluk/logs/<session>.jsonl` for anything to be written at all. It
+did not between [#1759](https://github.com/kubestellar/hive/pull/1759) — which
+replaced the Go `pluk-publish` binary, a publisher that wrote the log itself,
+with the TypeScript `pluk watch` — and
+[#4285](https://github.com/kubestellar/hive/issues/4285). Throughout that
+window this directory was empty and every consumer below was reading files that
+did not exist.
+
+Hive pre-creates each session log `0660` and `dev:node` before attaching
+pipe-pane. `hive-panes` is run by one agent to read every *other* agent's log,
+and agent UIDs differ, so the shared `node` group is what makes cross-agent
+reads work — leaving creation to the shell's `>>` would take whatever umask the
+pane shell carries and a tight one yields a log no peer can read.
+
+Nothing rotates or caps these files. `hive-panes` only ever reads the last 2000
+lines and the inception subscriber seeks to the end, so no consumer needs the
+history, but the files themselves grow for the life of the container.
+
 The event envelope, as read by Hive's own code, is:
 
 ```json
@@ -79,13 +101,20 @@ keys depend on `type`. Hive's dashboard subscribes to this stream for the
 `brainstorm` session (`pkg/dashboard/inception_watcher.go`, `plukEvent`) and
 reacts to a subset of event types:
 
-- `raw_output` — captured pane output. `hive-panes.sh` reads the text from
-  `data.line`; the dashboard's own Go subscriber reads it from `data.message`
-  (see `handlePlukEvent` in `inception_watcher.go`). Both are read from real
-  pluk output in this codebase — if you are writing a new consumer, check
-  both keys rather than assuming one, since pluk's own event shape is outside
-  this repo's control and the two existing readers do not agree on the field
-  name.
+- `raw_output` — captured pane output. The text is on `data.line`, and only
+  there. Both readers now agree: `hive-panes.sh` has always read `data.line`,
+  and the dashboard's Go subscriber (`handlePlukEvent` in
+  `inception_watcher.go`) read `data.message` until
+  [#4285](https://github.com/kubestellar/hive/issues/4285) corrected it, which
+  had left its whole `raw_output` arm dead. Earlier revisions of this page
+  advised new consumers to check both keys; that was wrong, and generalised
+  from the `error`/`rate_limit` shape below. pluk builds the event as
+  `createEvent(..., 'raw_output', { line })` in `dist/classifier.js`.
+
+  `raw_output` is opt-in: `pluk watch` defaults `includeRaw` to false, so the
+  publisher has to pass `--include-raw` or no such event is ever emitted. It is
+  the only event type `hive-panes.sh` consumes, so peer awareness depends on
+  that flag.
 - `state_change` — `data.state` is `"working"` or `"idle"`. The inception
   watcher uses idle transitions to decide when to re-kick a stalled agent or
   fall back to auto-generating questions/facts.
--- a/src/pkg/agent/manager.go
+++ b/src/pkg/agent/manager.go
@@ -1392,9 +1392,19 @@ func (m *Manager) ensureTmuxSession(agent *AgentProcess) error {
 		if backend == "" || IsInferenceBackend(backend) {
 			backend = "claude"
 		}
-		pipePaneCmd := fmt.Sprintf("%s watch %s --cli=%s", plukPath, agent.tmuxSession, backend)
+		logFile, err := ensurePlukLogFile(plukRunDir, agent.tmuxSession)
+		if err != nil {
+			// Non-fatal, and deliberately not a reason to skip the attach: the
+			// shell's own `>>` will still create the file. It may land 0600 under
+			// a tight umask, which costs peer readability but not the agent's own
+			// logging, and that is strictly better than no publisher at all.
+			m.logger.Warn("pluk log file setup failed; peer agents may not be able to read this session's log",
+				"agent", agent.Name, "error", err)
+			logFile = plukSessionLogPath(plukRunDir, agent.tmuxSession)
+		}
+		pipePaneCmd := plukPipePaneCmd(plukPath, agent.tmuxSession, backend, logFile)
 		_ = m.tmuxCmd(agent, "pipe-pane", "-t", agent.tmuxSession, "-o", pipePaneCmd).Run()
-		m.logger.Info("pluk publisher attached", "agent", agent.Name, "cli", backend)
+		m.logger.Info("pluk publisher attached", "agent", agent.Name, "cli", backend, "log", logFile)
 	}
 
 	return nil
--- a/src/pkg/agent/pluk_dirs.go
+++ b/src/pkg/agent/pluk_dirs.go
@@ -15,6 +15,55 @@ const (
 // directory to unrelated container users.
 var plukSharedDirMode = os.FileMode(0o770) | os.ModeSetgid
 
+// plukLogFileMode matches plukSharedDirMode's intent one level down. hive-panes
+// is run BY one agent to read every OTHER agent's log, so a session log written
+// by uid A has to be readable by uid B; the shared node group is what makes that
+// work, exactly as it does for the directory.
+//
+// The file has to be created here rather than left to the shell. pipe-pane's
+// `>>` would create it under whatever umask the agent's pane shell carries, and
+// a 0077 umask yields a 0600 log that no peer can read — the same class of
+// silent breakage permissions_watcher.go exists to repair for agent config
+// (#3882). Creating it up front with an explicit mode means `>>` only ever
+// appends to a file that is already correct.
+var plukLogFileMode = os.FileMode(0o660)
+
+// plukSessionLogPath is where pluk's events for one tmux session are appended.
+// It mirrors what pluk's own `attach` builds — <run-dir>/logs/<session>.jsonl —
+// and is the path src/deploy/hive-panes.sh globs and
+// src/pkg/dashboard/inception_watcher.go tails.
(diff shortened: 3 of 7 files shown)


clubanderson:
/lgtm
/approve

Reviewed and verified locally: build + vet clean, all 8 new tests pass, and the constructed pipe-pane command was run end-to-end against real @kubestellar/pluk@0.8.0 — the log file is written, raw_output events carry the text on `data.line`, and shell quoting holds for hostile session names.

Issue #2573, opened 2026-08-04
Hive is using Bob coins despite cadence settings


gee4vee:
I have the supervisor paused and the other agents with 4h or 6h cadences as shown below

<img width="917" height="595" alt="Image" src="https://github.com/user-attachments/assets/a73e50bf-9dff-495d-a2b2-bac7671f9db4" />

However, I can see my Bob coin usage going up. I also see the next run is more frequent than the cadences.

Not sure if related, but I have a repo access issue:

<img width="932" height="829" alt="Image" src="https://github.com/user-attachments/assets/7f2dbfa4-452f-4c96-a626-7ecda30f854f" />

clubanderson:
Diagnostic update (maintainer): I exec'd into the reporting hive (`hosted-available-vllmd-12`, vLLM-d cluster) to find the root cause precisely, and the evidence does **not** point at the scheduler.

**Findings from the live pod:**
- `mode-history.json` has a single entry (`startup → IDLE`, Jul 28) — the hive never entered a short-cadence mode, so the "next run more frequent than cadence" symptom is not stale-cadence carry-over.
- Pod restarts = **0** — no crash-restart re-kick loop.
- `hive.log`: every 5-minute governor eval tries to kick the paused agents and each is correctly **rejected** (`cannot be kicked: it is paused`). Pause is working; kicks burn nothing.
- `token-summary.json` = all zeros, `by_agent: {}` — hive is recording **no agent token consumption** for this hive.
- Only unpaused agent is `guide` (backend **copilot**, not bob). Paused bob agents' panes sit at a bare `bash` prompt with no live bob process.

**On the 'idle bob session' theory:** IBM Bob's public docs (Bobcoins / cost guide) describe Bobcoin billing as strictly **action-based** — consumption per interaction, no documented idle/keepalive/session-lifetime charge. So an idle bob session draining coins is not supported by the docs.

**What this means:** PR #2627 fixes two real scheduler bugs (stale cadence across mode changes; unconditional crash-restart re-kicks) but **neither occurred on this hive**, so it does not close this issue.

**To locate the actual drain, one question:** with the supervisor paused and only the copilot `guide` agent running now, **are your Bob coins still dropping?**
- If **yes** → the consumer is outside hive's kick path (e.g. a shared Bob account, another tool/IDE session on the same account, or Bob-side background work). Hive can't cause or see that (consistent with the zero token-summary).
[...]

PR #4305, opened 2026-08-20
✨ feat(inference): detect provider spending-limit rebuffs and stop burning runs

### LINKED ISSUE #4294: ✨ Detect provider spending-limit rebuffs (LiteLLM budget-exceeded): pause agents and surface an advisory instead of burning runs all day
## Field report

From mspreitz's hosted hive (`hosted-llm-d-incubation-llm-d-fast-m-yfzz.apps.fmaas-vllm-d.fmaas.res.ibm.com`), which uses an IBM LiteLLM gateway for inference:

> My LiteLLM gateway key has a $100/day dollar limit. The quality agent's 13:02 run was ENTIRELY rebuffed by the LiteLLM gateway due to that daily dollar limit; the 07:00 run was partially rebuffed. The daily spend chart shows the hive burning right up to the $100 clip every day — i.e. the hive keeps launching agent runs that hammer a spending-limited gateway.

No advisory, banner, pause, or notification appeared. The hive burned every scheduled run against a gateway that was rejecting 100% of requests, all day.

## Root cause: the hive has no concept of a provider dollar-spend rebuff

When LiteLLM's key budget is exceeded it returns HTTP 429/400 with bodies like `Budget has been exceeded! Current cost: X, Max budget: 100` / `ExceededBudget` / `litellm.BudgetExceededError`. Tracing what the hive does with that:

1. **No detection anywhere.** There is no matching of `budget_exceeded`, `ExceededBudget`, `quota`, `billing`, `credit`, or similar provider-side spend errors in `src/pkg/agent`, `src/pkg/classify`, `src/pkg/rotation`, `src/pkg/apiproxy`, or the pluk event consumer. All existing "budget" code is the hive's **own token budget** (`src/pkg/governor/governor.go:586-645`, `src/cmd/hive/main.go:4317-4346`), which is denominated in tokens and knows nothing about the gateway's dollar limit.

2. **The nearest hook is pluk `rate_limit` events, and its response is trivial.** The hive consumes pluk `rate_limit` events in `src/pkg/dashboard/inception_watcher.go:558-565`, but the only action is a 3-minute retry suppression (`inception_watcher.go:651-653`) — no agent pause, no advisory, no notification. Whether a LiteLLM budget-exceeded body even classifies as `rate_limit` depends on the external `@kubestellar/pluk` classifier's patterns (see `docs/agent-logging.md:104-123`); the hive itself never inspects the message.

3. **Provider rotation can't see it either.** Backend rotation / strand-pause (`src/cmd/hive/main.go:6116-6176`, `src/pkg/rotation/rotation.go:351-380`, #4227) is driven by *headroom probes*, not by runtime error bodies, so a key that has hit its daily dollar clip is not treated as exhausted.

4. **A fully-rebuffed run is indistinguishable from a normal failed run.** `src/pkg/classify` classifies issues, not run outcomes; the agent manager (`src/pkg/agent/manager.go:6548-6562`) only handles a generic 400 output-token-cap case. So the 13:02 run just exits having produced nothing, and the scheduler happily launches the next cadence — repeating until midnight resets the gateway budget. Exactly what the daily spend chart shows.

## Proposed fix

- Add a budget/spend-rebuff pattern class distinct from transient rate limits — either in the pluk classifier (emit e.g. `provider_budget_exceeded`) or in hive-side matching of agent output / translator error responses (the apiproxy/translator sees the raw 429 body for inference backends, which is the cleanest interception point).
- On detection: raise a dashboard system alert (`dashSrv.AddSystemAlert`, same mechanism as `applyBudgetAlerts` in `src/cmd/hive/main.go:4317`) + notification, and **suppress kicks to affected agents until the provider window plausibly resets** (LiteLLM returns `retry-after`/reset info in some responses; otherwise back off with escalating intervals). A daily-dollar rebuff is categorically not retryable in minutes.
- Feed the signal into rotation (`src/pkg/rotation`) so an alternate backend can be selected if one is configured.

Filed from investigation of mspreitz's field report; related: #4227 (provider rotation).

Danathar:
Closes #4294.

## What was happening

mspreitz's hive spent all day launching scheduled runs into a LiteLLM gateway that was rejecting 100% of its requests on a $100/day key limit — then reset at midnight and did it again. No advisory, no banner, no pause.

Nothing in the tree could have noticed. The existing "budget" code is the hive's **own token budget** (`pkg/governor`), denominated in tokens and blind to what the gateway charges. The nearest hook, pluk `rate_limit`, only suppresses retries for three minutes — the right answer for a throttle, useless for a daily dollar cap.

## Detection

Lands in the inference translator's error path (`GitHubProxy.recordInferenceError`), which the issue identified as the cleanest interception point and which already sees the raw upstream status and body — and already latches the sibling auth-failure signal, so this follows a proven shape rather than inventing one.

**Spend rebuffs are matched by body content, never by status alone.** This is the single most important design point: LiteLLM returns budget errors as *both* 429 and 400, the same codes ordinary throttles use. Matching 429 broadly would suspend every agent on a busy gateway — a worse bug than the one being fixed.

Covered: LiteLLM `Budget has been exceeded` / `BudgetExceededError`, OpenAI `insufficient_quota`, Anthropic credit-balance refusal.

**The latch fires on the first match**, unlike the auth tracker's three. Not a lower bar — a bare 401 is genuinely ambiguous (a key swap in flight looks identical to a dead key) so it needs a counter; a body that says "Budget has been exceeded" is unambiguous on sight. Waiting would mean burning two more runs to learn what is already known.

**It self-heals.** The first inference call that succeeds after the provider's window resets clears the latch. No timer to tune, no operator action to remember, no state to clean up.

## Response

An error-level dashboard system alert plus a high-priority notification, carrying the gateway's own numbers (`Current cost: 100.02, Max budget: 100.0`) — which limit was hit is what turns "agents are paused" into "raise the cap or wait for reset". Then every kick assembled that cycle is dropped.

## One thing I got wrong and fixed

My first version gated `agentsDue` right after the governor's pause filter. That was wrong: the CEL trigger union runs *after* it and would have re-added agents, and review-swarm kicks are appended to `messages` independently and would have bypassed it entirely. Both would have fired into the same clipped key.

The gate now sits after all three sources have contributed — governor-due, CEL union, review swarm. My tests would not have caught that; an operator would have.

## Deliberate boundaries

- **Suppression is total, not per-agent.** The limit is on the *key* — no agent can succeed while it is clipped.
- **Agents are not force-paused.** Operator pause state is a human decision (#2573) and must not be forged by an automatic signal that will clear itself. Withholding kicks gets the saving without leaving paused agents for a human to un-pause by hand.
- **Rotation (#4227) is not wired.** The issue's third bullet asks for the signal to feed backend selection. Rotation is driven by headroom probes rather than runtime error bodies, so that is a new trigger into backend selection and deserves its own review rather than riding along with the detection fix. Happy to follow up.

## Tests

They drive the production path (`recordInferenceError` / `recordInferenceSuccess`), not just the state helpers — a refactor that stops calling the detector from the translator fails the test rather than passing on helper coverage alone.

**The load-bearing cases are the negative ones.** These must all leave the signal clear, and each has a test:

| Case | Why it must not latch |
|---|---|
| plain rate limit (429) | transient — belongs to the existing retry path |
| parallel-request throttle (429) | same |
| `overloaded_error` (429) | same |
| output-token-cap (400) | already handled in `manager.go` |
| auth failure (401) | that's the *other* signal |
| entitlement refusal (403) | that's the *other* signal |
| empty body | never enough on its own |

Plus a test that the spend latch and the auth latch never contaminate each other — an operator seeing "spending limit" must not also be told their key is invalid.

`go test ./pkg/proxy/ ./pkg/dashboard/ ./pkg/governor/` all pass; `go vet` clean on the three touched packages.


clubanderson:
Reviewed in depth. The detection half of this is excellent — body-content matching gated on plausible statuses, the negative table driven through the production `recordInferenceError` path, and the latch kept independent of the auth signal are all exactly right. Verified locally: `go build ./... && go vet` clean, all `Budget|Inference` tests pass.

**But there is a recovery deadlock, and it blocks merge:**

The self-heal depends on `recordInferenceSuccess()`, which only fires when an agent's inference call returns 2xx through the translator (`github_proxy.go:1770,1942`). The suppression in `runEvalCycle` withholds **every** kick source — governor-due, CEL union, review swarm. So in a purely cadence-driven hive (exactly mspreitz's setup in #4294), once in-flight runs drain there is **zero** inference traffic. At midnight the provider window resets, but nothing ever calls inference, so the latch never clears — the hive stays muted **forever**, with an alert that says "suspended until the provider window resets" describing a recovery the mechanism cannot deliver. The only ways out are a manual dashboard kick (`/api/kick` bypasses the gate) or a restart (the latch is in-memory) — both operator actions, which contradicts the "no operator action to remember" claim.

The out-of-band lanes (replan lane's `agentKicker`, `planFromLabeledIssues`, brainstorm re-kick) do bypass the gate and would incidentally clear the latch — but only in hives running those lanes; the field-report topology has none of them.

**Suggested fix — probe-based suppression:** expose `lastRebuff` in the state's snapshot and only withhold kicks while the latch is *fresh*, e.g. `time.Since(lastRebuff) < probeInterval` (default ~30m, configurable via GovernorConfig per config.go conventions). Once stale, let one cycle's kicks through as a probe: if the provider is still clipped, the first rebuff re-freshens the latch and suppression resumes; if the window reset, the first success clears it. That bounds the burn to ~one run per interval instead of all day, and restores genuinely automatic recovery.

**Secondary:** `notifier.Send(..., PriorityHigh)` in the withhold block fires on **every** eval cycle that assembled kicks while latched — unlike `applyBudgetAlerts`, which notifies once per crossing via one-shot flags. All-day latching means all-day high-priority notification spam. Please notify once per latch (the deduped `AddSystemAlert` banner already carries the ongoing state).

On #4294 linkage: with rotation (#4227 wiring) deliberately deferred, `Refs` rather than `Closes` is the right call — agreed.

Happy to re-review and merge as soon as the recovery path is fixed.

Danathar:
Fixed both findings and rebased onto current `v4`. You were right about the deadlock — I verified it before changing anything: `recordInferenceSuccess` has exactly two call sites (`github_proxy.go:1770`, `1942`), both on the 2xx path of the inference translator, so the only thing that clears the latch is an agent actually running, and the only thing that makes an agent run is a kick. Withholding every kick withheld its own cure.

## The fix: evidence expires

Took your probe suggestion essentially as written.

`recordRebuff` now stamps `lastRebuff` on **every** rebuff, and `snapshot` carries it through `InferenceBudgetExceeded` and the dashboard provider to the eval cycle. `since` still pins the *first* latch — those are two different questions and both callers need one of them: `since` is "how long have we been clipped", which is what an operator reads, while `lastRebuff` is "how long since we last had evidence it is still true", which is what decides suppress-vs-probe.

Kicks are withheld only while that stamp is fresh. Once it is older than the probe interval, one cycle's kicks go through:

- **still clipped** → the first probe is rebuffed, which re-freshens the stamp, and suppression resumes for another interval. Burn is bounded to ~1 run per interval instead of a full day.
- **window reset** → the probe succeeds and the existing self-heal clears the latch outright.

`governor.provider_budget.probe_interval_s`, default 1800, following the `Effective*` convention in config.go. Two deliberate choices worth flagging: **zero and negative mean "unset", not "never probe"** — never probing is exactly the deadlock, so it is not reachable from config — and a **latched signal carrying no stamp probes immediately** rather than suppressing, on the same principle: spending one run is a recoverable error, muting the hive forever is not.

The banner now also distinguishes the two states, since "agent kicks suspended" was describing only one of them.

## Notification spam

Fixed as asked — once per latch, keyed on the latch time (which does not move while latched, so a genuinely new clip after a recovery still pages). Matches `applyBudgetAlerts`: notify on the crossing, leave the ongoing state to the deduped banner.

## Tests

Extracted the decision into `providerBudgetSuppresses` so the deadlock is pinned directly rather than through a whole eval cycle. The load-bearing case is the hours-old rebuff that **must** probe — that assertion failing *is* the bug you found. Plus a walk of the full loop in both directions (failed probe re-suppresses; cleared latch stops suppressing), and one that runs 200 consecutive latched cycles and asserts exactly one notification.

`go build ./...` and `go vet` clean; `./cmd/hive`, `./pkg/proxy`, `./pkg/config`, `./pkg/dashboard` all pass on the rebased tree. Also documented the signal and the knob in `operator-reference.md`, next to the token-budget bullet it is most likely to be confused with.

On the incidental-recovery point: agreed, and that is partly why I did not just document it. A hive that recovers only because it happens to run the replan lane is recovering by accident — the mechanism should not depend on which optional lanes an operator enabled.

clubanderson:
Hold lifted — re-reviewed the full stack and this is now merge-ready. ✅

Credit where due: @Danathar's detection work was excellent from the first cut (body-content matching gated on plausible statuses, the negative-case table driven through the production `recordInferenceError` path, latch independence from the auth signal), and his `b59269db` rework landed the heart of the recovery fix — evidence expires: `lastRebuff` freshness gating with `governor.provider_budget.probe_interval_s` (default 30m, zero/negative can never mean "never probe"), once-per-latch notification keyed on the latch time, and docs.

I pushed three additive commits on top to close the remaining review gaps:

- **`cfbd43d4`** — probe hardening: releasing a probe now **re-arms suppression immediately** (freshness = max(last rebuff, last released probe)), so kicks can't leak every cycle while the probe's run is still minutes away from its first inference call; the probe is trimmed to a **single kick** (one run answers "is the window still clipped" — more just spend); and recovery now sends a **one-time "serving again" notification**, the counterpart of the entering page. Transition-only in both directions.
- **`ff16d188`** — per @MikeSpreitzer's field data on #4294 (his gateway's window resets at the key-creation time of day, not midnight), scrubbed every prose claim of a midnight/day-boundary reset. The probe design is schedule-agnostic by construction — hive never predicts a reset, it just retries cheaply — and the comments/docs now say so.

Verified: `go build ./...` clean, all `ProviderBudget`/`Budget|Inference` tests pass, dashboard coverage 83.1% (floor 78%), CI fully green. The eval-cycle lifecycle is pinned by tests in both directions: latch → suppress → stale → single probe → rebuff re-arms / success clears → recovery page once.

/lgtm
/approve

Issue #4209, opened 2026-08-20
🩺 Podman preflight: SELinux, mounts, secrets, and ports


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

SELinux labels, configuration/secrets access, and host port conflicts are distinct operator-facing failure modes and should be diagnosed before startup.

## 30-minute boundary

Add only SELinux, mount-access, and port checks. Do not change volume ownership or host policy automatically.

## Deliverable

Extend Podman preflight with focused diagnostics for SELinux state/mount labeling, intended configuration and secrets readability, and required host-port availability.

## Acceptance criteria

- The check never recommends disabling SELinux as the normal fix.
- It does not broaden secret permissions automatically.
- Port 3001 and any other selected published ports are checked before startup.
- Docker behavior remains unchanged.
- Tests cover enforcing/permissive/disabled states and representative failures using mocked inputs.

Issue #4207, opened 2026-08-20
🩺 Podman preflight: engine, root mode, and cgroup checks


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

Operators need early, actionable diagnostics before Hive invokes a Podman lifecycle implementation.

## 30-minute boundary

Add only engine, root mode, and cgroup checks. Storage, SELinux, networking, secrets, and ports are separate issues.

## Deliverable

Extend the prerequisite path with Podman checks for executable/version, connection identity, rootless versus rootful mode, and cgroup version.

## Acceptance criteria

- Docker checks and their default behavior remain intact.
- Checks run only for explicitly selected Podman.
- Output reports the exact selected engine/connection and root mode.
- Unsupported or missing prerequisites produce actionable failures.
- Tests use controlled command output rather than requiring the test host to run Podman.

Issue #4210, opened 2026-08-20
🧹 Podman: define Hive-owned resources and safe cleanup contract


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

Podman and Buildah may share a rootless image store with unrelated development containers and Distroboxes. Hive lifecycle tooling must never translate broad Docker pruning into global Podman cleanup.

## 30-minute boundary

Define and test the ownership/cleanup contract only. Do not implement the complete lifecycle.

## Deliverable

Choose stable Hive resource labels/names and add a focused guard or contract test requiring Podman cleanup to select only Hive-owned resources.

## Acceptance criteria

- Broad commands such as `podman system prune`, `podman image prune -a`, `podman rmi -a -f`, and `buildah rm --all` are explicitly rejected in Hive tooling.
- The ownership labels cover containers, networks, volumes, and images where applicable.
- Docker cleanup behavior is not changed unless separately justified.
- The test can run without deleting any real resources.

PR #4307, opened 2026-08-20
🩺 podman: add subordinate-ID, graphroot, and rootless-network preflight

### LINKED ISSUE #4208: 🩺 Podman preflight: subordinate IDs, graphroot, and networking
## Parent

## Why

Rootless Podman requires subordinate IDs and suitable local storage/networking. Failure should be detected before containers or volumes are created.

## 30-minute boundary

Add only subordinate-ID, graphroot-filesystem, and rootless-network checks. SELinux, secrets, and ports are separate.

## Deliverable

Extend Podman preflight with focused checks for subordinate UID/GID mappings, rootless graphroot placement, and the selected rootless network backend/helper.

## Acceptance criteria

- NFS or another unsupported distributed graphroot is rejected with remediation guidance.
- Missing subordinate-ID mappings are reported without modifying `/etc/subuid` or `/etc/subgid`.
- Network-helper/backend detection is reported clearly.
- Docker behavior is unchanged.
- Tests cover supported, missing, and unsupported mocked outputs.

Danathar:
Closes #4208. Third and last preflight layer for #4188, after #4207 (engine, root mode, cgroup) and #4209 (SELinux, mounts, secrets, ports).

## What it checks

Three things rootless Podman needs from the host that have nothing to do with the engine version — and all three fail late, with an error that never names the real cause.

**1. Subordinate UID/GID mappings.** With no delegated range, `podman run` fails with "cannot find UID/GID for user" *after* creating the container. With a range too small it starts and then breaks on the first file owned by a UID past the end, which is much harder to recognise — so missing is a **fail** and short is a **warn**.

The tables are read directly rather than taken from `podman info`. The engine reports the mappings it *resolved*, which on a host using SSSD or a directory service can be correct while the files are empty, and can be stale relative to a file edited since the engine started. The files are what an administrator edits and what the remediation changes, so the files are what gets reported. Several ranges for one user sum.

**2. Graphroot filesystem.** Container storage on NFS lacks the ownership and xattr semantics overlayfs needs; layers fail to extract or extract and misbehave. Operators hit it because rootless storage defaults under `$HOME` and `$HOME` is on NFS. `nfs`/`nfs4`/`cifs`/`smb`/`fuse.sshfs`/`9p`/`ceph`/`glusterfs`/`afs`/`lustre`/`gpfs` **fail** with the `storage.conf` change and the `podman system reset` follow-up. `tmpfs` **warns** about what it costs rather than failing — a throwaway CI store on tmpfs is a legitimate choice.

**3. Rootless networking.** The engine names the helper it intends to use; whether that helper is installed is a separate question whose only symptom is a container with no network. Named-but-absent **fails** with the package to install. A `cni` backend **warns** about the upstream retirement rather than failing a host that is working.

Rootful hosts *pass* the two rootless-only checks explicitly rather than being silently skipped, so the report says why they don't apply.

## Read-only, and deliberately so on the sharpest edge

It never writes to `/etc/subuid` or `/etc/subgid`. Delegating a subordinate range is a host-administration decision with security consequences, made by someone who knows which ranges are already spoken for — not by a preflight running unprivileged. A missing mapping prints the `usermod` command and says outright that it will not run it.

A test asserts both tables are byte-identical after a *failing* run, so that property is pinned rather than promised.

## Docker is untouched

Self-gates on `HIVE_DEPLOY_RUNTIME` exactly like its siblings. The test asserts the podman call log is **empty** with Docker selected — not merely that the output looks right.

## Tests

`bin/test_hive_podman_preflight_ids.sh`, 61 assertions, covering supported / missing / unsupported per the acceptance criteria. Stub `podman` for the info templates, stub `stat` for the filesystem type, subid tables the case writes. No container, no privileges, no real `/etc` file touched.

**`PATH` is the fake bin alone, and that matters more than it looks.** Both `pasta` and `slirp4netns` are installed on a typical Fedora workstation, so inheriting `/usr/bin` made the absent-helper case silently untestable — my first version of that test passed for the wrong reason and I caught it because the assertion failed on the *expected* exit code. Pinning `PATH` is what makes "slirp4netns is not installed" a fact of the fixture rather than of the machine running the tests.

I also verified the headline NFS rejection is load-bearing by neutering the filesystem check and watching the suite go red:

```
FAIL: NFS graphroot fails: got '0', want '78'
```

All three preflight suites pass together (44 + 137 + 61). `shellcheck -S style` clean on both new files.

### Run for real

On a rootless Fedora host (btrfs graphroot, netavark, pasta, 65536 ranges):

```
Podman preflight (subordinate IDs, graphroot, networking)
  ✓ Subordinate IDs: dbaggett has subuid=65536 subgid=65536
  ✓ Graphroot: /var/home/.../containers/storage on btrfs (driver overlay)
  ✓ Network backend: netavark
  ✓ Rootless network helper: pasta present
SUMMARY: pass=4 warn=0 fail=0
```

and with the Docker default: `Podman preflight: skipped — HIVE_DEPLOY_RUNTIME selects docker`, exit 0, zero podman commands.

## Two notes for reviewers

**Not wired into `hive-prereq-check.sh`**, matching #4209. That script currently invokes only the engine layer; folding all three in is one deliberate change rather than a side effect of this one. Happy to do it here if you'd prefer.

**A pre-existing inaccuracy I left alone:** `src/docs/README.md` describes `bin/` as "the 45 deterministic pipeline and operational shell/Python scripts". That count was already wrong before this PR — `v4` has 59 — so it isn't something this change breaks, and correcting it means deciding what the intended denominator is. Flagging rather than guessing.


clubanderson:
Reviewed. The check logic, tests (61/61 pass locally), shellcheck cleanliness, docs, and DCO all look good — but **CI is red for a real reason, so holding the merge**.

**Cleanup-contract violation:** `bin/test_hive_podman_cleanup.sh` (#4210) greps every tracked file for a fixed list of broad cleanup commands, and `podman system reset` is on that list. All three new files quote it literally and are not in the contract's `exempt` list:

- `bin/hive-podman-preflight-ids.sh` lines 233 and 268 (remediation hints)
- `bin/test_hive_podman_preflight_ids.sh` line 200 (assertion on that hint)
- `src/docs/podman-preflight-ids.md` line 96

CI failure: https://github.com/kubestellar/hive/actions/runs/32404298932/job/96539610259

```
FAIL: broad Podman/Buildah cleanup found:
bin/hive-podman-preflight-ids.sh: podman system reset
bin/test_hive_podman_preflight_ids.sh: podman system reset
src/docs/podman-preflight-ids.md: podman system reset
```

Suggested fix: reword the remediation hints so the literal string never appears — e.g. "then reset Podman storage (see `podman-system-reset(1)`; destroys existing local containers and images)" — and update the corresponding test assertion and doc sentence to match. That keeps the preflight's advice intact without exempting new files from the contract, which exists precisely so remediation prose doesn't normalise store-wide commands.

Also note the branch is behind `origin/v4` and will need a rebase before merge.

Everything else in the PR is solid: read-only guarantee is pinned by the byte-identical-tables test, the Docker gate asserts an empty podman call log, the PATH-pinning rationale is sound, and rootful hosts pass explicitly rather than being silently skipped. Happy to merge once the contract grep is green.

Danathar:
Fixed and rebased onto current `v4` (`a58b4ed6`). You called it correctly — the contract grep was right and the PR was wrong.

Took the reword rather than the exemption, for the reason you gave: the exempt list covers the three files whose *job* is to quote those commands in order to reject them, and adding a preflight script to it would trade away the exact property the contract exists to hold. The advice itself was never the problem, so it is unchanged — an NFS graphroot and a retired CNI backend both still tell the operator the store has to be reset, and the NFS hint still says what that costs.

The hints now name the man page instead of the command:

- `bin/hive-podman-preflight-ids.sh:233` → `after moving it, reset the container store — see podman-system-reset(1) (destroys existing local containers and images)`
- `bin/hive-podman-preflight-ids.sh:268` → `install netavark and aardvark-dns, then reset the container store — see podman-system-reset(1)`
- `bin/test_hive_podman_preflight_ids.sh:200` — the assertion follows, now pinning `podman-system-reset(1)`. It still asserts the NFS failure names its follow-up, which was the point of that assertion.
- `src/docs/podman-preflight-ids.md:96` — same sentence, same instruction.

I left a comment at the first hint recording *why* the wording is what it is, so a future edit tidying it back to the literal command re-breaks CI without knowing what it stepped on.

Verified on the rebased tree:

- `bin/test_hive_podman_cleanup.sh` — **127/127**, the broad-cleanup grep now green
- `bin/test_hive_podman_preflight_ids.sh` — **61/61**
- `shellcheck` clean on both scripts
- `grep -rn "podman system reset"` over the three added files returns nothing

clubanderson:
Re-reviewed on `a5c0bb10` — the rewording resolves the hold exactly as suggested: the remediation hints now name `podman-system-reset(1)` instead of the literal command, the contract's exempt list stays untouched, and the guard comment at the first hint will keep a future tidy-up from re-breaking CI. Verified locally on the rebased head: cleanup contract 127/127 green, preflight suite 61/61, shellcheck `-S style` clean on both scripts, and a repo grep over the three added files finds no contract literals. CI is fully green (build-and-test, docker, overlayfs-exec-guard, all test shards, DCO). Also flipped the body's `Refs #4208` to `Closes #4208` so the issue closes on merge. Merging.

/lgtm
/approve

Issue #4319, opened 2026-08-20
🧪 Podman: determine whether the IPv4-only egress gate is bypassable over IPv6


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

The forced-proxy egress gate is IPv4-only. `src/deploy/entrypoint.sh` builds the
`HIVE_PROXY` chain with `iptables-nft`/`iptables` and has **no `ip6tables` path
at all** — zero references in the file. Every rule, including the
`--dport 443 -j REDIRECT --to-ports 18443` that does the actual forcing, exists
only in the IPv4 table:

```
-N HIVE_PROXY
-A HIVE_PROXY -m owner --uid-owner 0 -j RETURN
-A HIVE_PROXY -m owner --uid-owner 1001 -j RETURN
-A HIVE_PROXY -m mark --mark 0x1112 -j RETURN
-A HIVE_PROXY -p tcp -m tcp --dport 443 -j REDIRECT --to-ports 18443
```

So on any network where the container receives a global IPv6 address, an agent
that resolves an AAAA record reaches `:443` directly and never meets the
redirect. The capability model is not enforced for that traffic, and unlike the
advisory path it fails **silently** — there is no equivalent of the
`ADVISORY-ONLY` warning, because from the entrypoint's point of view the gate
installed successfully.

This was found while verifying the rootful baseline (#4200, PR #4304) and is
recorded there under *What remains unproven*. It was explicitly out of that
issue's 30-minute boundary.

Not yet demonstrated: the host used for #4200 gives its containers **no** global
IPv6 address, so there was nothing to bypass and no exploit was attempted. The
gap is read from the entrypoint, not observed in flight. That is exactly what
this issue is for.

## 30-minute boundary

Determine whether the bypass is reachable, and record it. Do not add an
`ip6tables` path, change the entrypoint, or alter Docker behavior — if the
bypass is real, the fix is its own slice.

## Deliverable

Run Hive under a container network that assigns a **global IPv6 address**, with
`--cap-add NET_ADMIN` so the gate installs normally, and establish whether agent
traffic to a dual-stack `:443` endpoint is redirected or escapes.

## Acceptance criteria

- Exact commands, Podman version, network configuration, and image reference are
  recorded, following `src/docs/podman-rootful-egress-baseline.md`.
- The result states plainly whether an agent-UID process reached a dual-stack
  host over IPv6 without traversing the proxy, with the evidence that shows
  which family carried the connection.
[...]

PR #4321, opened 2026-08-20
🧪 podman: the IPv4-only egress gate IS bypassable over IPv6 — measured

Danathar:
Refs #4319. Verification only — the entrypoint is unmodified and no `ip6tables` path was added, per the issue's 30-minute boundary.

## The bypass is real, and it was observed

#4200 read this gap out of the entrypoint but could not demonstrate it — that host gave its containers no IPv6 address, so there was nothing to bypass. This is the observation, on a dual-stack container network built for it.

With the gate installed normally (`--cap-add NET_ADMIN`), all connections from the **agent UID (2001)**, same endpoint, seconds apart:

| From agent UID 2001, to `:443` | Connections | `HIVE_PROXY` REDIRECT hits |
| --- | --- | --- |
| over **IPv4** | 5 | **5** |
| over **IPv6** | 5 | **0** |

The IPv4 row is the control the issue requires: it rules out "the gate failed to install" as an explanation for the IPv6 row.

Separately confirmed the gate is genuinely intercepting rather than merely counting packets — agent UID over IPv4 to `api.github.com`:

```
issuer=CN=Hive ACMM Proxy CA          # agent UID 2001 — intercepted
issuer=C=GB, O=Sectigo Limited, ...   # proxy UID 1001 — exempt, real upstream cert
```

## Why it is silent

Nothing fails. `entrypoint.sh` has **zero** `ip6tables` references (verified, `grep -c` → `0`), so it never attempts an IPv6 rule and has none to report failing. In the running container the IPv4 nat table carries the full `HIVE_PROXY` chain and the IPv6 nat table is **empty** — not misconfigured, empty.

## A method note, because it nearly produced the wrong answer

The obvious check — *"did the request reach the endpoint?"* — **does not discriminate**. The Hive proxy relays to the intended upstream, so a request that *was* redirected still returns the origin's content; and for a host the proxy tunnels rather than terminates, there is no substituted certificate to notice either.

An early IPv4 attempt came back with the endpoint's own body and its own certificate, reading exactly like a bypass. The rule counter showed it had been redirected all along. I stopped and dug into the apparatus rather than reporting that, and the doc now says to read the netfilter counter — which counts what the *kernel matched*, not what the application received.

## The fix slice, named but not implemented

Good news for whoever takes it: **it is a direct mirror.** All three exemptions and the redirect load unmodified under `ip6tables-nft`, verified inside the running container:

```
-m owner --uid-owner 1001 -j RETURN               OK
-m mark --mark 0x1112 -j RETURN                   OK
-p tcp --dport 443 -j REDIRECT --to-ports 18443   OK
```

So, answering #4319 directly: yes, the IPv6 chain needs the same three exemptions, and the `SO_MARK` / `--uid-owner` "equivalents" are the *same spellings* — `xt_owner` and the packet-mark match are family-independent. `/usr/sbin/ip6tables-nft` is already in the image, so no image change is needed.

Two things the fix must decide that this measurement does **not** settle:

1. **Failure policy when the host has no IPv6 at all.** The current `_iptables_ok` gate fails closed. "IPv6 rules failed" must stay distinguishable from "there is no IPv6 here", or every hive without IPv6 starts failing to boot.
2. **Whether the proxy answers on `:18443` over IPv6.** Redirect without a listener on that family converts a silent bypass into a silent outage.

## Limits, stated plainly

**Global routability was not available and is not claimed.** This host has only ULA IPv6 (`fd00::/8`, plus a Tailscale `fd7a::/48`), no default IPv6 route, and no IPv6 path to the internet — confirmed from inside the container (`curl -6 https://api.github.com` → `http=000`).

What *is* proven: the gate does not match IPv6 traffic **at the netfilter layer** — a property of the ruleset and the packet's family, not of the destination's routability. What is **not** proven by direct observation: an agent reaching a real public MITM-ed endpoint over IPv6, because no such path exists here.

Also: one image digest, `amd64` only, netavark, rootful, and a container-local endpoint rather than a public one. No claim about how often containers get IPv6 in practice — this says what happens when they do.

## Safety

Every container, image and network went into a throwaway store with a private graphroot/runroot and was removed afterwards; the host's own Podman storage was verified untouched. No Docker configuration was read or changed.

Also closes the loop on `podman-rootful-egress-baseline.md`, whose *What remains unproven* section listed this as the bypass-resistance test needing its own follow-up.


clubanderson:
/lgtm
/approve

Reviewed against the spike-series bar: docs-only, cleanup-contract clean (no forbidden store-wide prose), CI green. Rebased onto v4 to resolve the overlap with #4327 (which landed the fix slice this measurement names): the baseline's IPv6 bullet now carries both the measurement and the fix reference, and a clearly-marked postscript in the new doc records that the two open fix-slice questions were settled (proxy has no IPv6 listener → filter-table REJECT rather than a mirrored REDIRECT; no-IPv6-stack hosts pass vacuously). The measurement itself — 5/5 IPv4 redirected, 0/5 IPv6, netfilter-counter methodology — is untouched and is exactly the evidence #4319 asked for. Nice catch on the response-body false negative.

Issue #4316, opened 2026-08-20
🐛 hub: no operator alert for operator-side App-credential states — hosted hive sat degraded on key-missing for 8 days


clubanderson:
## Field evidence — kelly-headwaters

Hosted hive **kelly-headwaters** (owner kellyaa, ns `hive-hosted-hosted-available-oke-260812-8yiz-available-ok-go05`) was provisioned **2026-08-12** and discovered **Degraded** on **2026-08-20** — 8 days later, by an operator happening to hover the fleet row:

- 0 tokens ever consumed (the hive never worked once)
- 0 agents running, 4 paused, 1 needs-login
- hover: `⚠ GitHub App: credentials not yet delivered by the hub (operator action)` (spoke state `key-missing`)

Nothing anywhere pushed this to anyone. That is structural, not bad luck:

## Why it is silent today

The `key-missing` / `key-invalid` / `no-app-assigned` states are classified **operator-side** (`operatorSideAppStates`, `src/pkg/hub/journey.go`) — correctly, since the App private key is hub-distributed and the owner cannot fix it. But every surface then deliberately stands down:

1. **Owner**: spoke banner says "no action needed from you" (by design, `gh_app_banner_test.go`); journey nudges are blocked for operator-side states (`journey_operator_block_test.go`). Correct — but it means the owner will never escalate.
2. **Fleet alerts** (`src/pkg/hub/alerts.go`): there is **no alert type** for App-credential states. `AlertTypeHealthCheckFailing` does not fire either, because the spoke's `github_auth` check *passes* ("token-based") while the App key is missing (being fixed separately in the PR referencing this issue).
3. **Advisory staleness** is deliberately suppressed for these states (`advisory_staleness.go`: "must NEVER be flagged stale, because its silence is …" expected).
4. **Hub logs**: the heartbeat reconcile answers `appKeyReasonNoClusterKey` → returns nil **without logging** (only the placeholder-sentinel case logs "cannot repair"). 8 days × 2 beats/min of silent nil. (A WARN for this case is added in the small PR; a log line is still not an alert.)
5. **Drift exceptions** (`drift.go` `DriftKindAppCredsOperator`, critical) exist — but drift is a pull view; nobody is notified.

So the only actor who can fix it (the hub operator) is exactly the actor no signal reaches.

## What's needed (design, hence issue-only)

[...]

PR #4322, opened 2026-08-20
🐛 fix(hub): alert operators when the hub never delivered a hive's App key

Danathar:
Refs #4316. Picks up the main ask left after #4318 (merged) fixed the `github_auth` health check and added the WARN — neither of which is an alert.

## The gap

kelly-headwaters: provisioned 2026-08-12, found **Degraded** 2026-08-20. Eight days, zero tokens ever consumed, discovered because an operator happened to hover a fleet row.

That was structural. The `key-missing` / `key-invalid` / `no-app-assigned` states are classified **operator-side** — correctly, since the App private key is hub-distributed and the owner cannot fix it. Every owner-facing surface then deliberately stands down: the spoke banner says "no action needed from you", journey nudges are blocked, advisory staleness is suppressed. Each of those is right *on its own*. Together they meant the only actor who could repair it was the only actor no signal reached.

## What this adds

`AlertTypeAppCredsUndelivered`, raised **critical** for a claimed hive reporting `GitHubAppRequired` with an operator-side `GitHubAppState`, flowing through the existing ack/severity pipeline in `evaluateAlerts`.

Critical rather than warning because this isn't a degraded hive — it's one that has provably never worked and cannot until someone uploads a key.

**The condition is not re-derived.** It reuses `appStateIsOperatorSide`, the same predicate the journey and drift paths already use, so the alert cannot drift apart from the surfaces that stood down. It self-heals: once a key reaches the spoke the next beat reports a healthy state and the alert clears — no hub-side timer to keep in sync.

**The reason carries the remedy.** `PUT /api/saas/admin/cluster-app-keys/{clusterID}` exists but is documented nowhere and has no UI, so an alert naming only the condition would leave the operator as stuck as the silence did. The cluster id is interpolated so they can act without looking it up.

The three states get three different sentences because they have three genuinely different repairs — re-uploading the same key does nothing for `key-invalid`, and `no-app-assigned` needs an App assigned *before* a key helps at all.

## One deliberate deviation, flagged for you

The issue specifies "a claimed, **online** hive". **I did not gate on online.**

Gating there would drop the credential alert exactly when the hive is worst off — still keyless *and* no longer heartbeating. The offline rule raises its own alert for that half; this one is the only thing that says *why* the hive never worked, and an operator needs both. A test pins that they fire together.

Trivial to add the gate if you disagree — say so and I'll push it.

## Docs

An operator runbook in `hub-deployment.md`: the PEM store path, both admin endpoints with working `curl`, why the alert exists, and a table mapping each state to its actual repair.

Written from the code, not the issue — I verified `requireAdmin` gating on both routes and `clusterAppKeyDir = "/data/saas/app-keys"` before documenting them.

## Not done

The **optional** `has_key=false` column on the admin clusters view, so a keyless cluster is visible *before* the first claim strands a hive. That's UI work, marked optional in the issue, and complements this rather than substituting for it.

## Tests

The negative cases are the ones that matter — this must not re-open what the other surfaces correctly closed:

| Case | Expected |
|---|---|
| `key-missing`, `key-invalid`, `no-app-assigned` | **alert**, critical, remedy in the reason |
| `not-installed`, `installed`, `""`, `ok`, `suspended` | **no alert** — owner-fixable, the journey nudge owns those |
| Placeholder (unassigned pool slot) | no alert |
| App not required | no alert |
| Key uploaded → healthy state | alert clears |
| Offline + keyless | credential alert **and** offline alert |
| Registered in `knownAlertTypes` | yes — without it the alert is filtered out before anyone sees it |

Verified genuinely RED by disabling the rule (5 tests fail). `go test ./pkg/hub/` green (92s, full suite); `go vet` clean.

### DIFF
--- a/src/docs/hub-deployment.md
+++ b/src/docs/hub-deployment.md
@@ -99,6 +99,64 @@ The hub's SaaS provisioner creates or records:
 
 For heartbeat-only clusters where the hub cannot run `kubectl`, follow the manifest-level workflow in `manual-provisioning.md`; this guide intentionally keeps the hub setup consistent with that battle-tested path.
 
+## Cluster GitHub App keys (and the `app-creds-undelivered` alert)
+
+The hub is the authority for GitHub App private keys. Each cluster has one PEM
+in the hub's own store:
+
+```
+/data/saas/app-keys/<clusterID>.pem
+```
+
+At claim time the hub delivers the App *identity* (`app_id`) to the spoke, and
+then reconciles the *key* to it on every heartbeat. **If the hub holds no key
+for that cluster, the identity is delivered with `key_delivered=false` and every
+subsequent beat resolves to "no cluster key" and delivers nothing — forever,
+until an operator uploads one.** The spoke sits in `key-missing`, does no work,
+and burns no tokens.
+
+Upload a key for a cluster:
+
+```bash
+curl -X PUT \
+  -H "Authorization: Bearer $HUB_ADMIN_TOKEN" \
+  --data-binary @app-private-key.pem \
+  https://<hub>/api/saas/admin/cluster-app-keys/<clusterID>
+```
+
+List which clusters currently have one:
+
+```bash
+curl -H "Authorization: Bearer $HUB_ADMIN_TOKEN" \
+  https://<hub>/api/saas/admin/cluster-app-keys
+```
+
+Both routes are hub-admin only.
+
+### Why this has its own alert
+
+These credential states are classified **operator-side**, because the key is
+hub-distributed and the hive's owner cannot fix it. Every owner-facing surface
+therefore deliberately stands down: the spoke banner says "no action needed from
+you", journey nudges are suppressed, and advisory staleness is not flagged. All
+of that is correct — and the combined effect was that nobody was told at all.
+One hosted hive sat in `key-missing` for eight days with zero tokens consumed
+before an operator happened to hover its fleet row
+([#4316](https://github.com/kubestellar/hive/issues/4316)).
+
+The fleet now raises a **critical** `app-creds-undelivered` alert for a claimed
+hive reporting an operator-side App state, and the alert text names the upload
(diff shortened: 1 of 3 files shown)


clubanderson:
Thanks @Danathar — this was a solid, well-reasoned implementation, and we ended up building the same fix in parallel. #4323 (merged ~30 min before this review, closing #4316) landed the same `app-creds-undelivered` critical alert type through the same `evaluateAlerts` pipeline, with the same per-state remedy text naming `PUT /api/saas/admin/cluster-app-keys/{clusterID}`. This branch pre-dates that merge and now hard-conflicts with it (same constant, same `alerts_app_creds_test.go` filename, overlapping `alerts.go` regions), so merging both isn't possible — closing as superseded rather than asking for a rebase that would delete most of the diff.

**What your version got right that #4323 didn't** — the online gate. Your rationale (an offline keyless hive loses the WHY-alert exactly when it's worst off; the offline alert answers a different question; no other spoke-state rule gates on Online) is correct, and I've adopted it in **#4328** with your dual-alert offline test and your Co-authored-by preserved.

Two things #4323 carried that this branch didn't, for the record:
- the dashboard `ALERT_TYPE_LABELS` entry (without it the panel chip renders the raw key `app-creds-undelivered` — see `TestFailedUpgradeAlertHasADashboardLabel` for the precedent), and
- hub-observed `ClusterHasKey` from the PEM store, so `key-missing` distinguishes "the hub holds no key — upload one" from "a key exists but isn't arriving — check delivery".

One docs nit worth flagging since it's now in tree elsewhere: the `curl --data-binary @app-private-key.pem` example here wouldn't work — `handlePutClusterAppKey` decodes a JSON body `{"private_key": "-----BEGIN...", "app_id": 123}` and rejects non-JSON with 400. The merged runbook (`docs/github-app-setup.md`) documents the JSON shape.

Really appreciate the care in the reasoning here — the offline-gate catch made the merged implementation better.

Issue #4199, opened 2026-08-20
🧪 Podman: capture rootless startup and exit-77 behavior


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

Hive currently fails closed when its forced-proxy egress redirect cannot be installed. Rootless Podman must not be called enforcing merely because a container starts with an added capability.

## 30-minute boundary

Capture only the rootless startup and fail-closed behavior. Do not redesign the egress gate or implement the full Podman deployment path.

## Deliverable

Run a minimal, reproducible rootless Podman matrix against the current `v4` image or build and post the exact commands, Podman version, host/network backend, capability bounding set, entrypoint result, and logs for:

- Default rootless startup.
- Rootless startup with `--cap-add NET_ADMIN`.
- Deliberate `HIVE_PROXY_ADVISORY_OK=true` startup.

## Acceptance criteria

- The result distinguishes container startup from actual gate installation.
- Exit code 77 and the advisory-only warning are captured where applicable.
- The issue states what remains unproven, especially iptables redirect, `SO_MARK`, and bypass resistance.
- Any useful repeatable probe is added as a narrowly scoped script or documented command.

## Stop conditions

If the required rootless runtime is unavailable, record the exact missing prerequisite and stop. Do not substitute Docker or broaden the task.

PR #4340, opened 2026-08-20
🧩 docs(podman): publish the rootful/rootless × enforcing/advisory support matrix

### LINKED ISSUE #4324: 🧩 Podman: publish the supported rootful/rootless and enforcing/advisory matrix
## Parent

## Why

#4199 and #4200 are both closed, and between them they measured the same three
states on rootless and rootful Podman: fail-closed exit 77 without
`CAP_NET_ADMIN`, the gate installing with it, and advisory mode starting
unenforced with an explicit warning. Nothing yet states which of those
combinations Hive actually *supports* for a standalone deployment.

Until that is written down, every later slice has to re-derive it, and the
rootless result in particular is easy to over-read: #4199 removed the assumption
that rootless cannot enforce, but it did not make rootless first-class.

## 30-minute boundary

Publish the matrix and the support statement only. Do not change the entrypoint,
add deployment assets, or alter Docker behavior.

## Deliverable

A document stating, for each of rootful/rootless crossed with
enforcing/advisory, whether the combination is supported, unsupported, or
experimental for standalone Hive — with the evidence for each cell citing #4199
or #4200 rather than restating it.

## Acceptance criteria

- Every cell is one of supported / experimental / unsupported, with no cell left
  implicit.
- Each claim cites the spike that measured it; anything not measured is marked
  as such rather than inferred.
- The known gaps are carried forward by reference, not silently dropped — at
  minimum `arm64`, `slirp4netns`, restart behavior, and the IPv6 gate gap
  (#4319).
- Advisory mode is stated as a deliberate operator choice with its consequence
  spelled out, not as a fallback.
- No Docker configuration is changed.

## Stop conditions

If the evidence in #4199/#4200 does not settle a cell, mark it unproven and name
what would settle it. Do not fill a cell by reasoning from the ruleset.

Danathar:
## What

Publishes the support statement for standalone Hive under Podman: all four cells
of rootful/rootless × enforcing/advisory, each graded and each citing the spike
that settles it.

New: `src/docs/podman-support-matrix.md`. Indexed in `src/docs/README.md`, and
cross-linked from the two spike documents that supply its evidence.

Part of #4188 (does not close the parent). Closes #4324.

## The matrix

| | Enforcing (`--cap-add NET_ADMIN`) | Advisory (`HIVE_PROXY_ADVISORY_OK=true`) |
| --- | --- | --- |
| **Rootful** | **Supported** — #4200 / #4304 | **Supported as a deliberate choice, unenforced** — #4200 / #4304 |
| **Rootless** | **Experimental** — #4199 / #4280 | **Supported as a deliberate choice, unenforced** — #4199 / #4280 |

No cell is left implicit, and the remaining combination — neither capability nor
opt-in — is documented as what it is: not a support level but the fail-closed
exit **77** contract, measured identically under both modes.

## Why the cells read the way they do

- **Rootful + enforcing** is the reference cell. #4200 isolated `SO_MARK` from
  the owner-UID exemption with a control, and observed ambient `CAP_NET_ADMIN`
  surviving the privilege drop. The full enforcing claim was measured.
- **Rootless + enforcing** is experimental, not supported. #4199 removed the
  assumption that rootless cannot enforce — interception was proven — but
  `SO_MARK` was never isolated there, only `pasta` was exercised, and restart
  behavior is unmeasured. The promotion criteria are named and closed-ended:
  the `SO_MARK` isolation repeated under rootless, one `slirp4netns` lane, and
  gate re-installation across a restart.
- **Both advisory cells** are a deliberate operator choice, never a fallback.
  Advisory requires the operator to set `HIVE_PROXY_ADVISORY_OK=true`; a missing
  capability exits 77 rather than degrading into it. The consequence is spelled
  out — agents bypass the MITM proxy and the ACMM capability model is not
  enforced — along with where that is and is not defensible.

The advisory cells grading above rootless + enforcing looks backwards, so the
document says why rather than leaving the reader to trip over it: advisory makes
no enforcement claim, so nothing about it is unproven; rootless enforcing makes
the full claim on evidence that does not yet reach it. **The grades measure
evidence coverage of the claim made, not deployment safety.**

## Gaps carried forward, not dropped

A table of every open gap with its status and its home: `arm64`, `slirp4netns`,
restart/reboot/recreate, kernels without `xt_owner`, the #2678 `SO_MARK`
platform history, gateway/pod topology, locally built images, and the IPv6 gate
gap (#4319 — measured real in #4321, **fixed** in #4327, with the residual
stated: the fix was observed rootful/ULA-only/amd64, so no rootless dual-stack
measurement exists).

The document also records the single host configuration all of this rests on,
and explicitly declines to set minimum Podman versions — that is a separate
slice under #4188.

## Boundary

Documentation only. The entrypoint is untouched, no deployment assets are added,
and **no Docker configuration is changed** — Docker remains the default and
fully supported runtime. Nothing was inferred from the ruleset: anything the
spikes did not measure is marked unmeasured.


clubanderson:
Fact-checked every matrix cell against the current v4 entrypoint: the exit-77 fail-closed contract, the advisory opt-in (never a fallback — the WARN text quoted here is verbatim from the entrypoint), and the post-#4327 IPv6 gate (HIVE_PROXY6 filter-table REJECT with the same three exemptions, residual scope stated accurately). No enforcement guarantee is overstated, and the evidence-coverage grading is honest about what rootless enforcing still lacks. Docs-only, cleanup contract clean, CI green.

/lgtm
/approve

PR #4341, opened 2026-08-20
🔒 fix(security): keep the MITM CA private key out of agent-writable /data

Danathar:
Split out of #3895 at @clubanderson's request, carrying **only** the CA-key half.
The `SETUID`/`SETGID` capability drop is deliberately not here — #3895 stays held
for that discussion.

## The problem, still present on `v4`

`src/pkg/proxy/github_proxy.go:136`

```go
caKeyPath  = "/data/proxy-ca-key.pem"
```

`/data` is the shared, agent-writable workspace namespace. `0600` on a file
sitting directly under it is not sufficient: an agent can replace the path, or
race the proxy to it, before the proxy opens it. The MITM CA private key is the
credential that forges every TLS certificate the agents trust, so it is the one
file in the image that most needs to be outside that namespace.

## What this does

- **Moves the key to `/data/.hive/proxy-ca-key.pem`**, in a `0700` directory that
  is created and `chmod`ed *before* the key is written. The certificate stays at
  `/data/proxy-ca.pem` — agents have to be able to read and trust that.
- **Does not migrate a legacy key.** If `/data/proxy-ca-key.pem` exists, the
  entrypoint discards it *and* the certificate so the proxy generates a fresh CA
  pair on that boot. A key that sat in the exposed location may already have been
  copied by an agent; migrating it would keep a possibly-compromised key trusted,
  which is the failure this change exists to prevent.
- **Rejects symlinks** at either path rather than following them, so the proxy
  cannot be redirected to an attacker-controlled file.

## Scope

Four files, and nothing from the capability half:

| file | |
| --- | --- |
| `src/pkg/proxy/github_proxy.go` | path + directory/permission enforcement |
| `src/pkg/proxy/github_proxy_coverage_test.go` | asserts `0700` on the directory, `0600` on the key |
| `src/deploy/entrypoint.sh` | the migration/discard/symlink handling |
| `src/docs/security-model.md` | one added bullet describing the key's location |

`security-model.md` in #3895 also changed the `su-exec` and hosted-fleet bullets;
both of those describe capability-drop behavior, so only the new CA-key bullet is
taken here.

## Verification

- `go build ./...` clean, `go vet ./pkg/proxy/` clean, `go test ./pkg/proxy/` passes.
- Applied cleanly to current `v4` (`9a7e745f`) — no conflicts from the #3895 gap.
- The one `shellcheck` warning in `entrypoint.sh` (SC2043, line 413) is
  pre-existing on clean `v4` and untouched here.

Refs #3895.

### DIFF
--- a/src/deploy/entrypoint.sh
+++ b/src/deploy/entrypoint.sh
@@ -333,6 +333,26 @@ if [ "$(id -u)" = "0" ]; then
   chown dev:node /home/dev 2>/dev/null || true
   chown dev:node /etc/hive/hive.yaml 2>/dev/null || true
 
+  # Keep the MITM CA private key out of /data's agent-writable namespace.
+  # /data contains shared agent state and may be group-writable on PVCs. The
+  # certificate remains at /data/proxy-ca.pem because agents need to trust it,
+  # but the private key lives below an owner-only directory. If the legacy key
+  # is present, discard the old CA pair so a key that may already have been
+  # copied by an agent can never remain trusted. Reject symlinks so the proxy
+  # cannot be redirected to an attacker-controlled path.
+  mkdir -p /data/.hive && chown dev:node /data/.hive 2>/dev/null || true
+  chmod 700 /data/.hive 2>/dev/null || true
+  if [ -L /data/.hive/proxy-ca-key.pem ]; then
+    rm -f -- /data/.hive/proxy-ca-key.pem 2>/dev/null || true
+  fi
+  if [ -L /data/proxy-ca-key.pem ] || [ -f /data/proxy-ca-key.pem ]; then
+    # A key that was exposed in the old location may already have been copied
+    # by an agent. Do not migrate it: discard the old key and certificate so
+    # the proxy creates a fresh CA pair on this boot.
+    rm -f -- /data/proxy-ca-key.pem /data/proxy-ca.pem /data/proxy-ca-bundle.pem 2>/dev/null || true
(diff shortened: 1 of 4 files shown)


clubanderson:
Reviewed — this is exactly the good half of #3895, cleanly split.

**Verified:**
- Key relocated to `/data/.hive/proxy-ca-key.pem`: directory is `MkdirAll`+`Chmod` 0700 *before* the 0600 key write, and the entrypoint root block re-asserts `chown dev:node` + `chmod 700` on every boot. Agents (UID 2001+) can no longer read or replace the MITM CA key. Symlinks at both paths are rejected/removed.
- **Migration safety**: a legacy `/data/proxy-ca-key.pem` discards the whole CA pair (key + cert + bundle) rather than migrating a possibly-copied key — correct call. This lands the pod on the fresh-PVC boot path, which is already race-free: the Go token-mint client lazily reloads `/data/proxy-ca.pem` and the entrypoint bundle block is explicitly correct whether or not the CA exists yet. Running fleets upgrade safely.
- **No capability changes**: confirmed nothing from the rejected SETUID/SETGID bounding-set half of #3895 is smuggled in; the security-model.md edit is only the new CA-key bullet.
- `go build`/`go vet`/`go test ./pkg/proxy/` pass locally at 165697f; shellcheck on `entrypoint.sh` is identical to base v4 (no new findings). CI fully green.

/lgtm
/approve

Issue #4203, opened 2026-08-20
🧪 Podman: evaluate Quadlet .kube reuse for standalone Hive


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

Podman Quadlet can deploy Kubernetes YAML through `.kube` units. Hive should determine whether its existing manifests can reduce Docker/Podman configuration drift.

## 30-minute boundary

Perform a compatibility spike only. Do not convert or replace the Kubernetes deployment.

## Deliverable

Inspect one representative Hive Kubernetes YAML path with `podman kube play`/a temporary `.kube` unit and list the smallest adaptations needed for a standalone Hive and gateway deployment.

## Acceptance criteria

- Unsupported Kubernetes fields and Podman-specific additions are listed.
- Port publication, volumes, secrets/config, user namespace, capabilities, and readiness implications are addressed.
- The result says whether `.kube` reuse is viable, not viable, or needs one precisely named follow-up.
- Existing Kubernetes behavior is unchanged.

Issue #4202, opened 2026-08-20
🧪 Podman: evaluate minimal Quadlet container and pod units


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

Quadlet `.container`, `.pod`, `.network`, and `.volume` units may be a better persistent-installation model than Compose, but Hive needs a small feasibility result before committing to it.

## 30-minute boundary

Validate only a minimal Quadlet representation and generator output. Do not translate the full stack.

## Deliverable

Create or document a minimal temporary Quadlet for one Hive service and validate it with the Podman systemd generator or `systemd-analyze`.

## Acceptance criteria

- Required Podman and cgroup-v2 assumptions are recorded.
- Rootful and rootless unit locations and systemd mode are identified.
- The result covers image, environment, one mount, readiness type, and startup timeout behavior.
- Temporary probe files are not presented as production assets.
- The issue ends with a concrete recommendation or a specific blocker.

Issue #4201, opened 2026-08-20
🧪 Podman: evaluate explicit Compose-provider selection


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

`podman compose` delegates to an external provider, and current Podman documentation gives `docker-compose` precedence when multiple providers are installed. Hive needs deterministic provider behavior.

## 30-minute boundary

Evaluate provider selection only. Do not implement the production deployment.

## Deliverable

Exercise a minimal Hive-compatible Compose fragment through an explicitly selected provider and document:

- How Hive will select and report the provider.
- The minimum tested Podman/provider versions.
- Whether the provider works without Docker Engine, a Docker socket, `podman-docker`, or a Docker compatibility alias.
- Any unsupported Compose fields used by `src/docker-compose.yaml`.

## Acceptance criteria

- The test does not rely on provider auto-selection.
- At least one recommended path uses no Docker tooling.
- The recommendation names the exact configuration surface, such as `PODMAN_COMPOSE_PROVIDER` or `containers.conf`.
- Gaps are recorded as separate follow-up candidates rather than implemented here.

PR #4342, opened 2026-08-20
🧩 docs(podman): record Quadlet as the persistent lifecycle, with the version floor it implies

### LINKED ISSUE #4325: 🧩 Podman: record the chosen persistent lifecycle and minimum versions
## Parent

## Why

Three spikes now bear on how a standalone Podman deployment should be run, and
they disagree about which mechanism is viable: #4201 (Compose-provider
selection), #4202 (Quadlet `.container`/`.pod`), and #4203 (Quadlet `.kube`
reuse, which found the standalone Kubernetes overlay is not a safe direct
source). All three are closed.

The Podman Hive service asset, reboot persistence, and both update paths are all
downstream of this choice, so they cannot be scoped until it is made and
recorded.

## 30-minute boundary

Record the decision and the minimum versions it implies. Do not add the service
asset itself — that is its own slice, unblocked by this one.

## Deliverable

A short decision record naming the chosen persistent lifecycle mechanism, the
rejected alternatives with the reason drawn from the relevant spike, and the
minimum Podman version the choice requires.

## Acceptance criteria

- The chosen mechanism is named, and each rejected alternative cites the spike
  that ruled it out.
- A minimum Podman version is stated, with what specifically requires it — the
  spikes exercised 5.8.4, which is a data point, not a floor.
- The minimum version claim distinguishes "measured on" from "requires at
  least"; if the true floor is unknown, say so rather than quoting 5.8.4 as one.
- The record says what would reopen the decision.
- No Docker configuration is changed.

## Stop conditions

If the spikes do not settle the choice, record which additional spike is needed
and stop rather than choosing on preference.

Danathar:
## What

Records the persistent-lifecycle decision for standalone Hive under Podman, and
the minimum Podman version it implies.

New: `src/docs/adr/0017-podman-quadlet-lifecycle.md`, indexed in the ADR list
and cross-linked from the three spikes it draws on.

Part of #4188 (does not close the parent). Closes #4325.

## The decision

**Quadlet `.container`/`.pod`/`.network`/`.volume` units**, rootful from the
system manager and rootless from the per-user manager with `enable-linger`.

The deciding property is readiness. `Notify=healthy` holds the unit in
`activating` until the Podman healthcheck passes, so a started unit means a
healthy Hive. Nothing else evaluated supplies that, and the update/rollback
slices are precisely the ones that need it.

### Rejections, each citing its spike

- **`podman compose` for persistence — #4201.** Provider auto-selection silently
  chose the Docker Compose plugin on a host with `podman-compose` installed; the
  plugin needs an always-on Podman API socket and is Docker-packaged; and there
  is no readiness gate. Compose is **retained for the development loop** with an
  explicitly named provider — the spike's own recommendation, not a demotion
  invented here.
- **Quadlet `.kube` replay of the Kustomize overlay — #4203.** Unsupported kinds
  ignored with a zero exit; two `Deployment`s contending for one host binding;
  `fsGroup` not surviving the rootless user namespace, which is what makes Forge
  App keys readable by `hive-launch` while excluding agent UIDs; readiness probe
  dropped; no gateway on that path at all.
- **`podman generate systemd` / bare `--restart` — no spike.** Recorded as
  evaluated by nothing, so the omission is visible. Ruled out on upstream
  deprecation (Podman 4.7.0 notes) and on reasoning respectively — labelled as
  such rather than dressed up as a result.

## Versions: "measured on" is kept apart from "requires at least"

| | Version | Basis |
| --- | --- | --- |
| Measured on | 5.8.4 | Observed across all three spikes — **a data point, not quoted as a floor** |
| Requires at least | **5.0.0** | *Derived from upstream release history, never run* — `.pod` unit files and `Notify=healthy` both first shipped in 5.0.0 |
| Recommended | 5.6.0 | Same derivation: the 5.5.1 pod-storage fix, and containers/podman#26105 |
| **Verified floor** | **unknown** | Nothing between 5.0.0 and 5.8.4 has been run at all |

Every version in the record traces to a specific upstream release-notes bullet
(Quadlet 4.4.0, `AutoUpdate=` 4.6.0, `--sdnotify=healthy` 4.7.0, `.pod` and
`Notify=healthy` 5.0.0). containers/podman#26105 is stated precisely rather than
as a general upgrade argument: by its reporter's description it only bites when
the `.pod` unit omits `PodName=`, which #4202's probe units set — so the layout
should set it deliberately rather than land outside the regression by luck.

The experiment that would settle the real floor is named: the #4202 generator
probe plus one live `systemctl start` to healthy at each candidate version.

## What would reopen it

Six triggers are recorded. The first is the honest one: **`Notify=healthy` has
never been observed live.** #4202 was a generator dry-run, so the property the
entire decision rests on was read out of generated units rather than watched
holding a real unit against a real healthcheck. If a live start does not behave
that way, the decision is void rather than amended.

## Boundary

Documentation only — no deployment asset, no unit file, no preflight check, and
**no Docker configuration changed**. Docker remains the default runtime.


clubanderson:
Fact-checked against origin/v4 and upstream containers/podman RELEASE_NOTES.md. Every version in the table traces to the cited release-notes bullet: Quadlet 4.4.0, AutoUpdate= 4.6.0, --sdnotify=healthy and the generate-systemd deprecation (verbatim quote) 4.7.0, .pod units and Notify=healthy both first shipping in 5.0.0, the .pod storage-mount fix in 5.5.1 (#26190), and the #26105 --pod-name fix landing in 5.6.0 — the derived 5.0.0 floor and 5.6.0 recommendation are honest, and the measured/derived/verified split is exactly right. Rejection rationale matches the three spikes' own findings (provider mis-selection, socket dependency, PublishPort pod-member rejection, fsGroup loss, TimeoutStartSec trap), cross-links to the support matrix are consistent with d64af604, and no hard version assumption exists in the entrypoint or deploy scripts to contradict. Docs-only, cleanup contract clean, CI green.

/lgtm
/approve

PR #4343, opened 2026-08-20
🧹 feat(podman): tear down standalone deployments through the #4210 ownership contract

### LINKED ISSUE #4326: 🧹 Podman: implement standalone cleanup using the #4210 ownership contract
## Parent

## Why

#4210 is closed and defined the ownership contract: the
`io.kubestellar.hive.owned=true` label set, the guard in
`bin/hive-podman-cleanup.sh` that rejects store-wide operations, and
`src/docs/podman-ownership-cleanup.md`. What exists today is the contract and
its guard — nothing yet applies the labels to deployment assets or performs a
teardown through them. #4210 said so explicitly and left it to a later slice.

A cleanup path that is defined but not implemented is the dangerous state: the
guard only protects operations that go through it.

## 30-minute boundary

Apply the ownership labels to the standalone deployment assets and implement
teardown using only the contract from #4210. Do not widen the contract, add new
label keys, or change Docker behavior.

## Deliverable

Standalone Podman deployment assets carrying the #4210 label set, and a teardown
path that selects on the ownership label and refuses anything the guard rejects.

## Acceptance criteria

- Every container, pod, network, and volume the standalone path creates carries
  the label set defined in #4210.
- Teardown selects exclusively on the ownership label; no store-wide operation
  appears anywhere in the path, and the existing guard is what enforces that.
- A test asserts teardown leaves an unlabelled resource untouched, and that a
  store-wide command is rejected with the reason text.
- Docker teardown is unchanged.

## Stop conditions

If applying the labels requires changing the ownership contract itself, stop and
record what needs to change in #4210's document rather than widening it here.

Danathar:
## What

Implements the teardown that #4210 defined but deliberately left unbuilt.

- `bin/hive-podman-teardown.sh` — teardown for one standalone Podman
  deployment, built on the #4210 contract rather than beside it.
- `bin/test_hive_podman_teardown.sh` — 74 assertions against a fake Podman with
  a workstation-shaped store.
- CI step, `bin/README.md` rows, and the contract doc's *Teardown* section.

Part of #4188 (does not close the parent). Closes #4326.

## Why it needed building

#4210 shipped the label set, the guard, and the plan, then stopped — nothing
performed a teardown through any of it. That is the dangerous state, because
**the guard only protects operations that go through it**. A contract with no
implementation is a contract nobody is using.

## How it uses the contract

- **Selection is `hive_podman_filters` and nothing else.** An unlabelled
  container, pod, volume, or network is not skipped — it is never returned by a
  listing, so no code path can reach it. On a workstation that same rootless
  store holds the operator's Distroboxes and unrelated dev containers.
- **Every command passes through `hive_podman_cleanup_check` before it runs,
  reads included.** A rejected command aborts with `70`; it is never adjusted
  and retried.
- **Removal names operands.** No `prune`, no `--all`, anywhere in the path.
- Containers → pods → volumes → networks. Images are opt-in behind `--images`
  (an image is shared with every deployment that pulled the same reference).
  `HIVE_DEPLOY_INSTANCE` scopes the run. `plan` is the default; `run` needs
  `--yes`.

## The test

The fake Podman implements label filtering *for real*, so "the unlabelled
resource survived" is measured through the same mechanism the teardown depends
on rather than asserted about a stub. Its store holds Hive's resources next to
a Distrobox, a toolbox, `random-dev-box`, and somebody's `my-pgdata` volume —
all of which must survive every scenario.

Beyond the two criteria the issue names:

- **Every command the teardown actually issued is replayed back through the
  #4210 guard.** Not "no prune was written" but "nothing the path emitted could
  reach past Hive's labels", with the existing guard rendering the verdict.
- **A resource named `--all`.** IDs come from the engine, so they are input. If
  a listing ever returns a name that would turn `podman rm --force <ids>` into
  a store-wide removal, the guard must reject the constructed command and the
  teardown must abort. It does.
- I mutation-tested the suite: dropping the filters from the listing, and
  skipping the guard in the exec path, each fail it.

## On acceptance criterion 1 (labelled assets)

**No standalone Podman deployment asset exists yet** — it is its own slice of
#4188, and ADR-0017 records Quadlet as its mechanism. So rather than assert the
labelling by hand, the suite scans tracked deployment assets for anything that
creates a Podman resource and fails if the file does not go through the labels
seam. It passes vacuously today (**0 assets scanned**, and it says so) and stops
being vacuous the moment one lands. Spike probes and tests are excluded by
name: they run in a throwaway store with a private graphroot and remove what
they create by exact name.

## Boundary

The contract is not widened and no label key is added. **Docker teardown is
unchanged** — the script speaks only to Podman, refuses to run when
`HIVE_DEPLOY_RUNTIME` selects another runtime, and the suite fails if `docker`
is invoked at all. #4210's repo-wide scan for broad cleanup forms now exempts
the new test for the same reason it exempts its own: the file quotes those
commands in order to assert they are rejected. Both suites pass (127 + 74).


clubanderson:
Reviewed on `2c81796b`. This is the right shape for the #4210 contract's first consumer:

- **Selection is the contract, not beside it** — every listing is built from `hive_podman_filters` (ownership label + instance), so unlabelled resources are unreachable rather than skipped; removals name the exact IDs the filter returned, no `prune`/`--all` anywhere in the path.
- **Guard on every command, reads included**, with rejection → abort-70 rather than adjust-and-retry — and the hostile `--all`-named-resource test proves that path actually fires.
- **Test quality is the standout**: the fake Podman implements label filtering for real, the workstation-shaped store (Distrobox, toolbox, `my-pgdata`) must survive every scenario, and replaying the issued command log back through the #4210 guard measures the property instead of asserting prose. The vacuous-today asset scan is a sensible tripwire for the Quadlet slice.
- Exempting the new test from the cleanup-contract grep is correct for the same reason the contract's own test is exempt — it quotes the broad forms in order to assert rejection.

Verified locally: teardown suite 74/74 and cleanup contract 127/127 green on the PR head **and** on a trial merge with current `v4` (post-#4341 — its entrypoint changes don't trip the new label scan); `shellcheck -x -S style` clean on all three touched scripts. CI fully green, DCO good, body closes #4326. Merging.

/lgtm
/approve

PR #4344, opened 2026-08-20
🐛 fix(ci): run the Podman rootless contract check on mainline

### LINKED ISSUE #4339: 🐛 fix(ci): podman-contract.yml targets v2 and no longer runs
## Parent

## Why

`.github/workflows/podman-contract.yml` triggers only on the `v2` branch:

```yaml
on:
  push:
    branches:
      - v2
  pull_request:
    branches:
      - v2
```

Mainline is `v4`, so this workflow has not run for some time. It is the static
contract check for the rootless-Podman handling in the `contribute-hive` recipe
(#2535), and it is currently passing by never executing — the worst failure mode
for a guard.

## 30-minute boundary

Fix the triggers. Do not expand what the workflow checks.

## Deliverable

`podman-contract.yml` triggering on the current mainline branch.

## Acceptance criteria

- The workflow runs on push and pull_request against the current mainline branch.
- It is observed running on a PR — a green tick that came from the job actually executing, not from it being skipped.
- The path filters still scope it to the files it guards.
- No change to the contract the workflow asserts.

## Stop conditions

If the contract it checks is itself stale, fix the trigger anyway and open a separate issue for the contract.

Danathar:
## What

`.github/workflows/podman-contract.yml` triggered on `v2` alone. Mainline is
`v4`, so its last execution was **2026-08-12** — since then it has been passing
by never running, which is the one failure mode a guard cannot have.

Part of #4188 (does not close the parent). Closes #4339.

## The fix

Ten added lines, no deletions. Both changes mirror
`.github/workflows/suid-contract.yml`, the sibling static contract check:

1. **Branches become `v2` and `v4`.** Not a swap — `v2` is still receiving
   commits (most recent: today), so it keeps its coverage rather than trading
   one blind spot for another. `v2 + v4` is what `suid-contract.yml`,
   `v2-ci.yml`, and `v2-tests.yml` all use.
2. **The workflow's own path is added to both path filters.** This is the part
   that is easy to miss: the existing filters cover `Justfile`,
   `check-podman-contract.sh`, and `podman-rootless-ci.md`, so a PR touching
   only the workflow would **not** run it — the trigger fix could not be
   observed executing on the PR that makes it, and future edits to the workflow
   would go unchecked for the same reason. `suid-contract.yml` already lists
   itself for exactly this reason.

## What is not changed

The job, its single step, and `src/scripts/check-podman-contract.sh` are
untouched, so the contract asserted is identical — the diff contains no
deletions at all. The path filters still scope the workflow to the files it
guards.

Verified locally against the `v4` `Justfile`: all six checks pass
(`--userns=keep-id`, `:Z`, the macOS carve-out, no `docker.sock`, no
`podman.sock`, docker-before-podman auto-detect). The **Podman Rootless
Contract** check on this PR is the observed run the issue asks for, and it is
running because of change 2.

## One thing noticed, deliberately not folded in

`check-podman-contract.sh` emits `grep: warning: stray \ before -` from the
`'\-\-userns=keep-id'` pattern. It is cosmetic — the match is identical — but
it will now appear in CI logs. The issue's boundary is the triggers, so it is
left alone here rather than quietly bundled into a CI fix.

### DIFF
--- a/.github/workflows/podman-contract.yml
+++ b/.github/workflows/podman-contract.yml
@@ -6,21 +6,31 @@ name: Podman Rootless Contract
 # live-runner test would add. Does NOT change the runtime auto-detect
 # default (Option A is explicitly deferred pending broader coverage).
 
+# Triggers cover both maintained branches. This workflow used to name `v2`
+# alone; mainline moved to `v4` and it stopped running entirely (#4339) — a
+# guard that never executes reports green forever, which is the one failure
+# mode a guard cannot have. The branch list and the self-referencing path entry
+# below mirror .github/workflows/suid-contract.yml, the sibling static contract
+# check.
 on:
   push:
     branches:
       - v2
+      - v4
     paths:
       - 'Justfile'
       - 'src/scripts/check-podman-contract.sh'
       - 'src/docs/podman-rootless-ci.md'
+      - '.github/workflows/podman-contract.yml'
   pull_request:
     branches:
       - v2
+      - v4
     paths:
       - 'Justfile'
       - 'src/scripts/check-podman-contract.sh'
       - 'src/docs/podman-rootless-ci.md'
+      - '.github/workflows/podman-contract.yml'
   workflow_dispatch:
 
 permissions:


Danathar:
Acceptance criterion 2 (observed running, green from execution rather than a skip) — the **Podman Rootless Contract** workflow ran on this PR as a `pull_request` event and completed successfully. It is the first run of this workflow since 2026-08-12.

[Run 32416241428](https://github.com/kubestellar/hive/actions/runs/32416241428), job log:

```
== Podman rootless contract check (Justfile) ==
  ok: rootless UID mapping (--userns=keep-id) present
  ok: SELinux volume label (:Z) present
  ok: macOS network carve-out still references darwin
  ok: no docker.sock mount
  ok: no podman.sock mount
  ok: auto-detect still tries docker before podman

Rootless Podman contract check passed.
```

It ran because the workflow's own path was added to the `paths:` filters — without that change this PR would not have triggered it, and the fix could not have been observed working.

clubanderson:
Reviewed and verified.

- **Root cause confirmed**: the workflow's triggers named `v2` only; mainline is `v4`, so the guard had not executed since 2026-08-12 (verified via run history) while reporting green by absence.
- **Fix verified working**: the branch list (`v2` + `v4`) and self-referencing path entry exactly mirror `suid-contract.yml`, and the **Podman Rootless Contract** workflow demonstrably ran on this PR's own head (run 32416241428, success) — which it could not have without the self-path addition. PR-path triggers for `Justfile`/script/doc are untouched, additive-only diff.
- `actionlint` clean, YAML parses, `permissions: contents: read` unchanged (no escalation), no forbidden podman lifecycle commands anywhere in the workflow or script.
- `bash src/scripts/check-podman-contract.sh Justfile` passes locally against `v4` (all six checks).
- No dependency on the parallel #4343 — this references only the pre-existing contract script, so no hold needed.

/lgtm
/approve

PR #4345, opened 2026-08-20
🌱 ci(podman): run the rootless startup, exit-77, and egress-gate matrix per PR

### LINKED ISSUE #4334: 🌱 CI: rootless Podman lane on hosted amd64
## Parent

## Why

#4211 mapped the hosted-runner capabilities and recommends this lane as issue 1
of 6. `ubuntu-latest` has the rootless stack (netavark + pasta, crun, overlay),
and #4199 already produced the probe that exercises it — nothing runs it per PR,
so a regression in the egress gate would land unnoticed.

## 30-minute boundary

Wrap the existing probe in a workflow. Do not modify the probe's contract or
change Docker behavior.

## Deliverable

A workflow on `ubuntu-latest` running `src/deploy/probe_podman_rootless_netadmin.sh`.

## Acceptance criteria

- The three-case matrix (fail-closed exit 77, gate installed under `--cap-add NET_ADMIN`, advisory) runs per PR.
- A gate that fails to install fails the job — the probe's non-zero exit is not swallowed.
- The discovered Podman version and root mode are asserted and printed, so a runner image change is visible rather than silent.
- No secrets, and no `pull_request_target`.

## Stop conditions

If `ubuntu-latest` no longer provides a working rootless stack, record what changed and stop rather than weakening the assertions.

Danathar:
## What

`.github/workflows/podman-rootless-lane.yml` — lane 1 of the map in
`src/docs/podman-ci-runner-map.md` (#4211). It runs
`src/deploy/probe_podman_rootless_netadmin.sh` (#4199) on `ubuntu-latest`, per
PR. The probe itself is **not modified**.

Part of #4188 (does not close the parent). Closes #4334.

## The three cases, now running per PR

| Case | Expectation |
| --- | --- |
| default rootless | fail closed, **exit 77** |
| `--cap-add NET_ADMIN` | egress redirect **installed** |
| `HIVE_PROXY_ADVISORY_OK=true` | starts with the explicit advisory warning |

The probe already exits non-zero when any case stops matching the contract in
`src/docs/podman-rootless-startup-spike.md`, so the lane is the wrapper that
makes it run instead of sitting there.

## Asserting the environment rather than trusting it

Podman is preinstalled on this runner, so there is no setup step — which is
exactly why the environment needs checking. A runner-image change is silent and
would otherwise turn a real regression into a test that quietly stopped
covering anything.

- **`rootless` must be `true`.** This is the assertion that decides whether the
  lane tests what it claims to: a rootful runner would pass all three cases and
  mean something else entirely (that is #4200's lane).
- **cgroups v2**, and **Podman ≥ 4.x** — the floor for rootless
  `--cap-add NET_ADMIN`.
- Exact version, network backend, rootless helper, OCI runtime, and runner
  image go into the **job summary**, and the version is compared against the
  `5.8.4` #4211 recorded. A mismatch is a `::notice`, not a failure — pinning
  the exact version would make every benign runner bump a red build, so drift
  is made *visible* rather than fatal.

## Not swallowing the result

No `continue-on-error`, no `|| true`. The runner invokes `run:` blocks with
`bash -e`, so errexit is lifted deliberately around the probe call — otherwise
the step would abort before the diagnosis — and the probe's status is then
re-raised as the step's exit code. **Exit 78 (missing prerequisite) fails just
as loudly as exit 1**, because a lane that could not run is not a lane that
passed.

## Safety

`permissions: contents: read` and nothing more, **no secrets**, **no
`pull_request_target`**, no engine socket mounted anywhere. The probe builds its
own throwaway store with a private graphroot and runroot, so the runner's
default Podman storage is untouched. I verified the image pulls **anonymously**
from GHCR (`v4-latest` → HTTP 200 with an anonymous token), so fork PRs — the
coverage that matters most — work without credentials.

**No path filters, deliberately.** #4339 is what a skipped guard looks like; a
lane scoped so narrowly it reports green by never running is the failure mode
worth avoiding, and this one is cheap enough for every PR.

## What a green tick here does and does not mean

Stated in the workflow header rather than left for someone to discover: the
probe pulls a **published** image. Green means the published image's gate still
behaves and the runner still provides a working rootless stack. It does **not**
build the PR's own image, so an unpublished change to `src/deploy/entrypoint.sh`
is not yet covered by this lane. Probing a PR-built image is a separate lane and
outside #4334's boundary of "wrap the existing probe in a workflow".

## Verification

Run locally first on Podman 5.8.4 / netavark + pasta / crun / cgroup v2 — the
same stack #4211 measured on the runner. The capability bounding sets came back
identical to the spike (`…800405fb` without, `…800415fb` with `--cap-add`), and
case 1 reported `exit: 77 … as documented`.

The lane also runs **on this PR**, which is the real check: it exercises the
workflow on `ubuntu-latest` rather than on my Fedora box. I will follow up with
the result.

Docker behavior is unchanged and no Docker workflow is touched.

### DIFF
--- a/.github/workflows/podman-rootless-lane.yml
+++ b/.github/workflows/podman-rootless-lane.yml
@@ -1,0 +1,186 @@
+name: Podman Rootless Lane
+
+# The rootless Podman lane on hosted amd64 (#4334), lane 1 of the map in
+# src/docs/podman-ci-runner-map.md (#4211).
+#
+# It runs the three-case matrix that #4199 measured by hand:
+#
+#   default    rootless, no added capability  -> fail closed, exit 77
+#   netadmin   rootless + --cap-add NET_ADMIN -> egress redirect installed
+#   advisory   HIVE_PROXY_ADVISORY_OK=true    -> starts with the explicit warning
+#
+# The probe itself is src/deploy/probe_podman_rootless_netadmin.sh and is not
+# modified by this lane — it already exits non-zero when a case stops matching
+# the contract recorded in src/docs/podman-rootless-startup-spike.md. This
+# workflow is the wrapper that makes it run per PR instead of by hand.
+#
+# WHAT A GREEN TICK HERE DOES AND DOES NOT MEAN. The probe pulls a published
+# image, so this lane asserts that the *published* image's egress gate still
+# behaves and that the runner still provides a working rootless stack. It does
+# NOT build the PR's own image, so a change to src/deploy/entrypoint.sh is not
+# covered until that change is published. Probing a PR-built image is a
+# separate lane (the map's lane 3 covers build/pull) and deliberately out of
+# #4334's boundary, which is "wrap the existing probe in a workflow".
+#
+# No path filters, on purpose. #4339 is what a skipped guard looks like: a
+# workflow scoped so narrowly that it reports green by never running. This lane
+# is cheap enough (one image pull, three short container runs) to run on every
+# pull request rather than only on the ones that look relevant.
+#
+# Safety boundaries, from the map's "Permissions and safety boundaries":
+# contents: read and nothing more, no secrets, and no pull_request_target —
+# fork PRs are the coverage that matters most and they get no secrets anyway.
+# No engine socket is mounted anywhere. The probe builds its own throwaway
+# store with a private graphroot and runroot, so nothing touches the runner's
+# default Podman storage, and cleanup removes only the containers it created,
+# by name.
+
+on:
+  push:
+    branches:
+      - v2
+      - v4
+  pull_request:
+    branches:
+      - v2
+      - v4
+  workflow_dispatch:
+    inputs:
+      bypass:
+        description: 'Also probe interception from an agent UID (needs outbound egress)'
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+concurrency:
+  group: podman-rootless-lane-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  rootless-probe:
+    name: rootless startup, exit-77, and egress gate
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      # Assert the environment, do not assume it. Podman is preinstalled on
+      # this runner image, so there is no setup step — which is exactly why the
+      # version and root mode have to be checked: a runner-image change is
+      # silent, and would otherwise turn a real regression into a test that
+      # quietly stopped covering anything.
+      #
+      # The floor is asserted; the exact version is reported. Pinning the exact
+      # version would make every benign runner bump a red build, so a
+      # difference from the version #4211 recorded is surfaced as a notice
+      # instead — visible, not fatal.
+      - name: Assert the rootless Podman stack this lane depends on
+        env:
+          RECORDED_PODMAN: '5.8.4'
+        run: |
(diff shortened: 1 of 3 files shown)


Danathar:
The lane ran on this PR — [run 32417056532](https://github.com/kubestellar/hive/actions/runs/32417056532), **pass in 5m5s** on `ubuntu-latest`. That is the acceptance evidence: the matrix executed rather than being skipped, and it executed on the hosted runner rather than on my workstation.

**Environment, asserted then printed:**

```
podman      5.8.4 (/usr/local/bin/podman)
rootless    true
cgroups     v2
backend     netavark / pasta
runtime     crun
```

Identical to what #4211 recorded, so no `::notice` for version drift fired. The same table is in the job summary alongside the runner image version.

**The three cases:**

```
--- case: default (rootless, no added capability) ---
  exit: 77
  gate: NOT installed
  RESULT: as documented — fail-closed, exit 77 (EX_NOPERM) with the bounding-set explanation

--- case: netadmin (rootless + --cap-add NET_ADMIN) ---
  gate: INSTALLED — iptables (iptables-nft): outbound :443 -> :18443 (proxy UID 1001 + egress mark 0x1112 exempt)
  RESULT: as documented — redirect installed and ambient CAP_NET_ADMIN survived the privilege drop

--- case: advisory (rootless, HIVE_PROXY_ADVISORY_OK=true) ---
  gate: NOT installed, advisory mode accepted
  RESULT: as documented — started with the explicit advisory-only warning

SUMMARY: 0 unexpected result(s)
```

The middle one is the case the lane exists for: the forced-proxy redirect installing under `--cap-add NET_ADMIN`, with the ambient capability surviving the privilege drop to `dev`. On hosted CI, per PR, from now on.

I also ran the same probe locally first (Podman 5.8.4 / netavark + pasta / crun / cgroup v2) and got `SUMMARY: 0 unexpected result(s)` there, with identical capability bounding sets to the #4199 spike (`…800405fb` without the capability, `…800415fb` with it). Total runtime 5 minutes, which is the cost of running this on every PR.

clubanderson:
Reviewed and verified.

- **The lane self-triggered on this PR and passed** (run 32417056532, 5m05s on `ubuntu-latest`) — the observed run #4334 asks for. ~5 min single job, no matrix fan-out, 20-min cap: per-PR cost is acceptable. The no-path-filter choice is deliberate and well-argued post-#4339; at this cost I agree it should run everywhere.
- **Contract assertions verified against the entrypoint**: `EXIT_NET_ADMIN_REQUIRED=77` + "exiting 77" (entrypoint.sh:29,982-983), IPv4 gate line `outbound :443 -> :18443` (line 872), `ADVISORY-ONLY` warning (line 971), ambient-cap survival (line 1060). The IPv6 REJECT gate (#4319/#4327) is a separate closed-family path the probe correctly does not conflate with the IPv4 REDIRECT assertion.
- **No #4343 interaction**: that PR touched only `bin/` teardown scripts and `v2-ci.yml`; this lane invokes only the pre-existing probe, which does its own by-name cleanup in a throwaway store. No overlap with the static `podman-contract` check just fixed in #4344 either — that greps the Justfile, this runs containers.
- **Rootless availability is asserted, not assumed**: rootless=true, cgroup v2, Podman ≥4.x hard-fail; version drift vs the recorded 5.8.4 is a notice. Exit 78 fails as loudly as exit 1; no `continue-on-error`/`|| true` swallowing.
- **Safety**: `permissions: contents: read` only, no secrets, no `pull_request_target`, no socket mounts, anonymous GHCR pull works for forks. No forbidden podman lifecycle commands anywhere.
- actionlint: only info-level SC2016 on the deliberately single-quoted Go template braces (commented in the workflow). YAML parses clean.

/lgtm
/approve

PR #4347, opened 2026-08-20
🌱 ci(podman): run the rootful egress-gate matrix per PR

### LINKED ISSUE #4335: 🌱 CI: rootful Podman lane on hosted amd64
## Parent

## Why

#4211 recommends this as issue 2 of 6. #4200 established the rootful baseline
and shipped `src/deploy/probe_podman_rootful_netadmin.sh`, but nothing runs it
per PR. Rootful is the reference the rootless result is measured against, so a
silent regression there invalidates the comparison.

## 30-minute boundary

Wrap the existing probe in a workflow. Do not modify the probe or change Docker behavior.

## Deliverable

A workflow on `ubuntu-latest` running the #4200 probe under `sudo`.

## Acceptance criteria

- Rootful mode is asserted (`rootless=false`), so the job cannot pass by silently running rootless.
- The three-case matrix runs per PR and a failed gate fails the job.
- No secrets, and no `pull_request_target`.
- The probe's storage isolation is preserved — the job must not point it at a shared store.

## Stop conditions

If rootful Podman cannot be exercised on the hosted runner, record the missing facility and stop.

Danathar:
## What

Adds `.github/workflows/podman-rootful-lane.yml`: the rootful Podman lane on
hosted amd64, lane 2 of the map in `src/docs/podman-ci-runner-map.md` (#4211).

#4200 established the rootful baseline and shipped
`src/deploy/probe_podman_rootful_netadmin.sh`, but nothing ran it per PR.
Rootful is the reference the rootless result is measured against, so a silent
regression there invalidates the comparison.

The workflow is a wrapper and nothing more — the probe is not modified, and
neither is any Docker behaviour.

## The matrix it runs

All three cases run as real root, on `ubuntu-latest`, under `sudo -n`:

| Case | Setup | Expected |
| --- | --- | --- |
| `default` | rootful, no added capability | fail closed, exit 77 |
| `netadmin` | rootful + `--cap-add NET_ADMIN` | egress redirect installed |
| `advisory` | `HIVE_PROXY_ADVISORY_OK=true` | starts with the explicit warning |

The probe's exit status **is** the result and the job swallows none of it:
`0` matched the baseline in `src/docs/podman-rootful-egress-baseline.md`, `1`
means a case stopped matching (a gate that failed to install lands here), `78`
means a prerequisite was missing — which is a broken lane, not a passing one,
and is reported just as loudly. No `continue-on-error`, no `|| true`.

The SO_MARK isolation case (`--somark`) needs outbound egress, so it stays
behind a `workflow_dispatch` input and is off by default, mirroring the
rootless lane's `--bypass`.

## Acceptance criteria

- **Rootful mode is asserted, not assumed.** `sudo podman info` must report
  `rootless=false` before the matrix runs. The probe refuses to run rootless
  too, but it must not be the only thing checking: a sudo or
  `XDG_RUNTIME_DIR` change is all it would take for every case to still pass
  while measuring #4334's lane instead. Passwordless `sudo -n` and a 4.x
  Podman floor are asserted the same way; the exact version is compared to the
  one #4211 recorded and surfaced as a `::notice::`, so a benign runner bump is
  visible without being red.
- **The three-case matrix runs per PR and a failed gate fails the job.**
  `push`/`pull_request` on `v2` and `v4`, no path filters — #4339 is what a
  guard scoped into silence looks like.
- **No secrets, no `pull_request_target`.** `permissions: contents: read` and
  nothing more; no engine socket is mounted anywhere. `pull_request_target`
  would run untrusted code with repository credentials as root, which is the
  one combination the map forbids outright.
- **The probe's storage isolation is preserved** — and then checked. No
  `--store` is passed, so the probe keeps building its own throwaway
  graphroot/runroot under `/var/tmp` and removing it itself. A post-run step
  compares the runner's own rootful store against a baseline taken before the
  run: a store this job was wrongly pointed at would have gained the probe's
  multi-gigabyte image and kept it, which is a far more durable trace than the
  containers, all of which are `--rm`.

## Stop condition

Not hit. #4211 measured `sudo -n` and `sudo podman run` working on
`ubuntu-latest` (`rootful-run-OK`), so no facility is missing. The lane still
asserts both at runtime rather than trusting that snapshot, and the
passwordless-sudo failure message points at the map as the place to record it
if that ever changes.

## What a green tick here does not mean

The probe pulls a published image, so this lane covers the *published* image's
egress gate and the runner's rootful stack. It does **not** build the PR's own
image, so an unpublished change to `src/deploy/entrypoint.sh` is not covered
until it is published. Probing a PR-built image is the map's lane 3 and is
outside this issue's 30-minute boundary.

Rootful on an ephemeral hosted runner is acceptable precisely because the VM is
single-use. The workflow header says so out loud, so the lane is not copied
onto a persistent self-hosted runner without re-reading that section of the map.

## Also

`src/docs/podman-ci-runner-map.md` lane 2 gets a "Built (#4335)" note, matching
how lane 1 is recorded.

Part of #4188 — this PR does not close the parent. Closes #4335.

### DIFF
--- a/.github/workflows/podman-rootful-lane.yml
+++ b/.github/workflows/podman-rootful-lane.yml
@@ -1,0 +1,257 @@
+name: Podman Rootful Lane
+
+# The rootful Podman lane on hosted amd64 (#4335), lane 2 of the map in
+# src/docs/podman-ci-runner-map.md (#4211).
+#
+# #4200 established the rootful baseline and shipped
+# src/deploy/probe_podman_rootful_netadmin.sh, but nothing ran it per PR.
+# Rootful is the reference the rootless result is measured against, so a silent
+# regression here invalidates that comparison. This workflow is the wrapper
+# that makes the existing probe run on every pull request; the probe itself is
+# not modified by this lane.
+#
+# The three cases it runs, all as real root:
+#
+#   default    rootful, no added capability   -> fail closed, exit 77
+#   netadmin   rootful + --cap-add NET_ADMIN  -> egress redirect installed
+#   advisory   HIVE_PROXY_ADVISORY_OK=true    -> starts with the explicit warning
+#
+# The probe exits non-zero when a case stops matching the contract recorded in
+# src/docs/podman-rootful-egress-baseline.md, and this job does not swallow
+# that status.
+#
+# WHAT A GREEN TICK HERE DOES AND DOES NOT MEAN. The probe pulls a published
+# image, so this lane asserts that the *published* image's egress gate still
+# behaves under rootful Podman and that the runner still provides a working
+# rootful stack. It does NOT build the PR's own image, so a change to
+# src/deploy/entrypoint.sh is not covered until that change is published.
+# Probing a PR-built image is the map's lane 3 and is outside #4335's boundary,
+# which is "wrap the existing probe in a workflow".
+#
+# No path filters, on purpose. #4339 is what a skipped guard looks like: a
+# workflow scoped so narrowly that it reports green by never running.
+#
+# SAFETY BOUNDARIES, from the map's "Permissions and safety boundaries". This
+# lane runs containers as real root on the runner VM, so they are tighter here
+# than anywhere else:
+#
+#   * contents: read and nothing more.
+#   * No secrets, and therefore no pull_request_target. Fork PRs are the
+#     coverage that matters most and they get no secrets anyway;
+#     pull_request_target would run untrusted code with repository credentials
+#     as root, which is the one combination the map forbids outright.
+#   * No engine socket is mounted anywhere.
+#   * Storage isolation is the probe's, and this lane leaves it alone: --store
+#     is deliberately NOT passed, so the probe builds its own throwaway
+#     graphroot/runroot under /var/tmp and removes it itself. Pointing this job
+#     at a shared store — including the runner's own /var/lib/containers — would
+#     defeat the isolation #4200 built in. The post-run step below asserts that
+#     the host's rootful store really was left untouched.
+#
+# Rootful on an ephemeral hosted runner is acceptable precisely because the VM
+# is single-use. Do not copy this lane onto a persistent self-hosted runner
+# without re-reading that section of the map first.
+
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed and verified — clean sibling of the #4345 rootless lane.

- **Self-trigger observed**: the rootful lane ran on this PR's own head and passed (run 32418260151, 5m07s) — same ~5-min per-PR cost as lane 1, 25-min cap.
- **Rootful elevation handled correctly**: passwordless `sudo -n` asserted up front (never a prompt/hang), and `rootless=false` asserted from `sudo podman info` itself so the lane cannot silently swap into measuring #4334's rootless lane. Podman ≥4.x floor hard-fails; version drift vs recorded 5.8.4 is a notice.
- **Root-owned state leakage addressed better than asked**: no `--store` passed so the probe keeps its own throwaway /var/tmp graphroot/runroot and self-cleans; an `always()` post-run step compares the host rootful store image count before/after and greps for leftover `hive-rootful-probe-` containers, failing the job if isolation was ever bypassed. Ephemeral-runner-only caveat is stated in the header for anyone copying this to a self-hosted runner.
- **Assertions consistent with the #4200/#4304 rootful baseline**: exit 77 + "exiting 77" (entrypoint `EXIT_NET_ADMIN_REQUIRED=77`), IPv4 gate line `outbound :443 -> :18443`, `ADVISORY-ONLY` warning — identical contract strings verified in the entrypoint. The IPv6 REJECT gate (#4319/#4327) remains a separate closed-family path, correctly not conflated. SO_MARK isolation stays behind `workflow_dispatch`, mirroring lane 1's `--bypass`.
- **Safety**: `contents: read` only, no secrets, no `pull_request_target` (explicitly forbidden combination documented), no socket mounts, no forbidden podman lifecycle commands in workflow or probe. actionlint: only info-level SC2016 on the deliberately single-quoted Go templates, same as merged lane 1.
- map.md lane-2 note merges cleanly over #4345's lane-1 note (MERGEABLE).

/lgtm
/approve

PR #4349, opened 2026-08-20
🌱 ci(podman): gate Quadlet unit generation in both root modes

### LINKED ISSUE #4338: 🌱 CI: Quadlet generator syntax gate
## Parent

## Why

#4211 recommends this as issue 5 of 6. #4202 and #4203 evaluated Quadlet
`.container`/`.pod` and `.kube` reuse; whichever mechanism #4325 selects, a unit
that fails to generate should fail CI rather than surface at deploy time.

## 30-minute boundary

A syntax/dry-run gate only. Do not add deployment units — the asset itself is a
separate slice downstream of #4325.

## Deliverable

A CI job that installs the Quadlet generator, verifies it is present, and runs
`quadlet --dryrun` for both `--user` and rootful modes.

## Acceptance criteria

- The job **fails if the generator is missing** rather than passing vacuously — #4211 names this as the trap.
- Both `--user` and rootful dry-runs are exercised.
- A generator warning that would become a deploy-time failure fails the job.

## Stop conditions

If no Quadlet units exist to check yet, gate on the generator's presence and dry-run capability only, and say so in the job's description.

Danathar:
## What

Adds a per-PR gate that runs the Podman systemd generator over the
repository's Quadlet units, in both modes ADR-0017 installs into:

- `.github/workflows/quadlet-gate.yml` — installs the generator, verifies it,
  runs the gate on `ubuntu-latest`.
- `src/deploy/test_quadlet_generator_gate.sh` — the gate itself, runnable by
  hand on any host that has the generator.

ADR-0017 (#4325, PR #4342) chose Quadlet `.container`/`.pod` units as the
Podman persistent lifecycle, so a unit that fails to generate should fail CI
rather than surface at deploy time. Nothing ran the generator per PR.

## Acceptance criteria

**The job fails if the generator is missing rather than passing vacuously.**
This is the trap #4211 names, and it is not hypothetical — the hosted runner
has podman *without* the generator, because its `/usr/local/bin/podman` is
hand-installed rather than the distribution package, so
`/usr/libexec/podman/quadlet` does not exist. There is no skip path anywhere in
either file: a missing generator is a failure. Presence is verified in its own
workflow step so it is a separately reported fact, and the script re-checks it
rather than trusting the caller.

**Both `--user` and rootful dry-runs are exercised.** Every check below runs
twice, once per mode. They are genuinely different code paths — rootful
generates `Wants=network-online.target`, `--user` generates
`Wants=podman-user-wait-network-online.service`.

**A generator warning that would become a deploy-time failure fails the job.**
Exit status alone is not enough, and this is measured rather than assumed:

```
$ QUADLET_UNIT_DIRS=... quadlet --user --dryrun   # unit with Image=alpine
quadlet-generator[…]: Warning: u.container specifies the image "alpine" which
not a fully qualified image name. …
exit 0
```

ADR-0017 requires fully qualified references because `podman auto-update`
cannot resolve a short name, so that warning *is* a deploy-time failure and the
exit status waves it through. The gate therefore fails on any generator
diagnostic, after filtering the unconditional `Loading source unit file` debug
lines that `--dryrun` always emits. The rule is deliberately conservative in
the failing direction: a new benign stderr line upstream shows up as a red
build rather than as silently reduced coverage, which is the correct way round
for a gate.

## Where the generator comes from, and why not `apt`

Two other sources were checked and neither works:

| Source | Generator | Verdict |
| --- | --- | --- |
| Runner's own podman 5.8.4 | none | hand-installed binary, no `/usr/libexec/podman/quadlet` — the #4211 finding |
| `apt-get install podman` on ubuntu-24.04 | 4.9.3 | below ADR-0017's binding constraint |
| `quay.io/podman/stable`, digest-pinned | 5.8.4 | used |

4.9.3 is measured (`apt-cache policy podman`), not recalled. ADR-0017's floor is
5.0.0 because `.pod` units and `Notify=healthy` both first shipped there, so the
packaged generator cannot express the layout this gate exists to check. The
image is pinned by digest so the generator version is a property of this
workflow rather than something that drifts with the runner image, and 5.8.4 is
the same version every #4201/#4202/#4203 spike measured. Only the generator
*binary* is used — the gate starts no containers, so nothing is nested.

## Stop condition: no units exist yet

The repository holds no Quadlet units; the deployment asset is a separate slice
downstream of ADR-0017, and #4338's boundary is explicitly a syntax gate and
not a unit. Per the stop condition the gate covers the generator's presence and
dry-run capability, and **says so in its own output** rather than reporting a
silent pass:

```
--- repository units ---
  none found.
  This gate therefore covers the generator's PRESENCE and DRY-RUN CAPABILITY
  only — #4338's stop condition — and not any Hive unit, because the
  deployment asset is a separate slice downstream of ADR-0017. The
  self-tests above are what keep that from being a vacuous pass.
```

That last sentence is the point. The gate runs known-good **and known-bad**
units through the same code path the real ones will take, so it demonstrably
fails today, before there is anything real to catch:

| Fixture | Generator behaviour | Gate |
| --- | --- | --- |
| well-formed unit | exit 0, silent, emits an `ExecStart=` | must pass |
| unsupported key | exit 1 | must catch |
| dangling `.pod` reference | exit 1 | must catch |
| short-name image | **exit 0** + `Warning:` | must catch *via the warning rule* |

If any of those stops behaving, the gate fails — including the good one, so a
generator that rejects everything cannot look like a clean run. Fixtures are
generated at runtime rather than committed, so nothing in the tree can be
mistaken for the deployment unit this issue is not adding.

Units landing anywhere in the repository are picked up with no change here: the
scan is repo-wide rather than pinned to one directory, since a unit added
somewhere unexpected would otherwise be silently ungated — the same vacuous-pass
failure mode as skipping on a missing generator.

## Verified locally

Both files were exercised against a real generator (podman 5.8.4, Fedora), on
the host and inside the pinned image:

| Case | Exit |
| --- | --- |
| repo scan, generator present | `0` |
| `--quadlet /nonexistent/quadlet` | `1` |
| `--units` pointing at a unit with an unsupported key | `1` |
| `--units` pointing at a unit with a short-name image | `1` (warning rule, generator exited 0) |
| `--units` pointing at a well-formed unit | `0` |
| whole gate inside `quay.io/podman/stable@sha256:312a77…` | `0` |

## What this gate is not

`quadlet --dryrun` is a syntax and reference check, never a semantic one.
ADR-0017 records three things it exits 0 on: `Notify=healthy` with no
`HealthCmd`, `PublishPort=` on a pod member, and a short-name image. The third
is caught here through the warning rule; the first two need Hive's own
assertions and are deliberately out of scope. Also unchanged: no deployment
units, no Docker behaviour, `contents: read`, no secrets, no
`pull_request_target`.

`src/docs/podman-ci-runner-map.md` lane 5 gets a "Built (#4338)" note, matching
how lanes 1 and 2 are recorded.

Part of #4188 — this PR does not close the parent. Closes #4338.

### DIFF
--- a/.github/workflows/quadlet-gate.yml
+++ b/.github/workflows/quadlet-gate.yml
@@ -1,0 +1,157 @@
+name: Quadlet Generator Gate
+
+# The Quadlet generator syntax gate (#4338), lane 5 of the map in
+# src/docs/podman-ci-runner-map.md (#4211).
+#
+# ADR-0017 chose Quadlet `.container`/`.pod` units as the Podman persistent
+# lifecycle. A unit that fails to generate should fail CI rather than surface
+# at deploy time, so this job runs `quadlet --dryrun` in BOTH modes the ADR
+# installs into — rootful and `--user` — over whatever Quadlet units the
+# repository holds.
+#
+# THE TRAP #4211 NAMES. The generator is a separate binary from `podman`, and
+# the hosted runner has podman WITHOUT it: the runner image carries a
+# hand-installed /usr/local/bin/podman rather than the distribution package, so
+# /usr/libexec/podman/quadlet does not exist. A gate that runs `quadlet` only
+# when it happens to be there reports green while testing nothing. This job
+# therefore installs the generator, verifies it, and has no skip path at all —
+# every failure mode below is a red build.
+#
+# WHY THE GENERATOR COMES FROM A CONTAINER IMAGE. Two other ways to get it were
+# available and neither works:
+#
+#   * `apt-get install podman` on ubuntu-24.04 installs the generator, but at
+#     podman 4.9.3 (measured: `apt-cache policy podman`). ADR-0017's binding
+#     constraint is 5.0.0 — `.pod` units and `Notify=healthy` both first
+#     shipped there — so the packaged generator cannot express the layout this
+#     gate exists to check.
+#   * The runner's own podman 5.8.4 ships no generator to borrow.
+#
+# The official Podman image carries a matched generator (5.8.4, the same
+# version every #4201/#4202/#4203 spike measured), and pinning it by digest
+# makes the version this gate ran against explicit and reproducible instead of
+# something that drifts silently with the runner image. Only the generator
+# BINARY is used — no container is started by the gate, nothing is nested.
+#
+# WHAT THIS GATE IS NOT. `quadlet --dryrun` is a syntax and reference check,
+# not a semantic one. ADR-0017 records three things it exits 0 on:
+# `Notify=healthy` with no `HealthCmd`, `PublishPort=` on a pod member, and a
+# short-name image. The first two need Hive's own assertions and stay out of
+# scope; the third IS caught, because the generator warns about it while still
+# exiting 0 — so the gate fails on any generator diagnostic, not just on a
+# non-zero exit.
+#
+# NO UNITS YET, AND THE JOB SAYS SO. The repository holds no Quadlet units —
+# the deployment asset is a separate slice downstream of ADR-0017, and #4338's
+# boundary is explicitly a syntax gate rather than a unit. Per its stop
+# condition the gate covers the generator's presence and dry-run capability
+# only. That is not a vacuous pass: the script runs known-good AND known-bad
+# units through the same path the real ones will take, so the gate proves it
+# can fail today. When units land anywhere in the tree they are picked up with
+# no change to this workflow.
+#
+# No path filters, on purpose. #4339 is what a guard scoped into silence looks
+# like. No secrets, no pull_request_target, contents: read.
+
+on:
+  push:
+    branches:
+      - v2
+      - v4
+  pull_request:
+    branches:
+      - v2
+      - v4
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: quadlet-gate-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  quadlet-gate:
+    name: quadlet --dryrun, rootful and --user
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    env:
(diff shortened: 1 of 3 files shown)


clubanderson:
Reviewed and verified.

- **Self-trigger observed**: the Quadlet Generator Gate ran on this PR's own head and passed in 16s (run 32419521141) — negligible per-PR cost, 15-min cap.
- **Claims match merged ADR-0017 exactly** (`src/docs/adr/0017-podman-quadlet-lifecycle.md`): 5.0.0 binding floor because `.pod` units and `Notify=healthy` both first shipped there (ADR lines 106, 117-118); 5.6.0 recommended floor noted-not-failed, correctly, because Hive's layout sets `PodName=` explicitly and sits outside the podman#26105 regression window (ADR lines 125-130).
- **Runner generator gap handled the right way**: no skip path anywhere — the hosted runner's hand-installed podman ships no `/usr/libexec/podman/quadlet`, so the gate takes the generator from `quay.io/podman/stable` pinned by digest (5.8.4, ≥ floor, same version the spikes measured), verifies executability and version in a separate step, and hard-fails if absent. `apt`'s 4.9.3 correctly ruled out as below floor.
- **Both modes genuinely exercised**: every fixture runs through both rootful and `--user` dry-runs (different generated dependencies), and the known-good/known-bad fixture matrix — including the short-name case the exit status waves through, caught via the warning rule — proves the gate can fail today, so "no units yet" is not a vacuous pass. Repo-wide unit scan picks up future units with no workflow change.
- **Safety**: `contents: read` only, no secrets, no `pull_request_target`, workspace mounted read-only into the container, fixtures generated at runtime in a mktemp dir with EXIT-trap cleanup, no forbidden podman lifecycle commands. shellcheck clean on the gate script; actionlint shows only the familiar info-level SC2016 on single-quoted Go templates.

/lgtm
/approve

PR #4352, opened 2026-08-20
🌱 ci(podman): prove the arm64 image pulls and the service starts

### LINKED ISSUE #4336: 🌱 CI: arm64 image build/pull and startup lane
## Parent

## Why

#4211 recommends this as issue 3 of 6, on `ubuntu-24.04-arm`. Every Podman
result so far is `amd64` only — #4199 and #4200 both record that as unproven.

## 30-minute boundary

Scope strictly to image build/pull plus service startup. Do not duplicate the
full three-case matrix on this lane.

## Deliverable

A workflow on `ubuntu-24.04-arm` that pulls or builds the image and confirms the
service starts.

## Acceptance criteria

- The lane is explicitly scoped to build/pull + startup, with no duplicate full matrix — #4211 calls that out as the failure mode to avoid.
- A missing `arm64` image manifest fails the job rather than silently skipping.
- Runtime and network backend are printed for comparison against the amd64 lanes.

## Stop conditions

If no `arm64` image is published for the tag under test, record that and stop rather than building one ad hoc in CI.

Danathar:
## What

Adds the arm64 lane, on `ubuntu-24.04-arm`:

- `.github/workflows/podman-arm64-lane.yml` — asserts the stack, runs the probe.
- `src/deploy/probe_arm64_image_startup.sh` — the probe, runnable by hand on any
  arm64 host.

Every Podman result so far is amd64 only — #4199 and #4200 both record arm64 as
unproven. This closes that on the two questions #4188 actually asks of arm64.

## Scope — four checks, and the failure mode being avoided

#4211 names a duplicated full matrix as the thing not to build here: its probe
found the two architectures capability-identical, so re-running the egress-gate
cases on arm64 doubles cost for coverage that result says will not diverge. The
gate stays on the amd64 lanes (#4334, #4335). This lane runs four checks:

| Check | What it proves |
| --- | --- |
| `manifest` | the tag advertises a `linux/arm64` image |
| `pull` | that image comes down **and really is arm64** |
| `exec` | `/usr/local/bin/hive` runs |
| `startup` | the container answers `GET /api/health` |

`/api/health` is the endpoint `src/deploy/k8s/deployment.yaml` uses as its
readiness probe, so "started" means what it means in production rather than a
log line proving a process was spawned.

The probe sets `HIVE_PROXY_ADVISORY_OK=true`, and that is the opposite of
sneaking in a matrix case: without it the entrypoint fails closed with exit 77
and **nothing about startup gets tested**. Advisory mode takes the egress gate
out of the picture so what is measured is startup. An agent is still configured,
so the startup path exercised is the real one (uid map, agent directories)
rather than the shortcut an agent-less config takes.

The `exec` check is not filler. #3760 was a binary that shipped as a
non-executable overlayfs metacopy redirect at the image path — an image that
pulls fine and cannot run. That is exactly the class of failure a
per-architecture lane exists to catch.

## Acceptance criteria

- **Explicitly scoped to build/pull + startup, no duplicate matrix.** Four
  checks, stated in the workflow header, the probe header, and the table above.
- **A missing arm64 manifest fails the job rather than silently skipping.** An
  arm64 lane that steps aside when the arm64 image is absent reports green while
  proving nothing — which is the exact condition this lane exists to end. The
  probe also asserts the *pulled* image's architecture, so a manifest regression
  cannot hand the lane an amd64 image on an arm64 runner.
- **Runtime and network backend are printed for comparison against the amd64
  lanes.** Both the probe and the workflow summary report Podman version, arch,
  root mode, cgroups, OCI runtime, and network backend / rootless network
  command. #4211's capability-identical claim is what the scoping rests on, so
  the fields that claim would be falsified by are reported on every run rather
  than assumed to still hold.

## Pull, not build — and the stop condition

#4188 asks for "build/pull plus startup". The arm64 **build** is already a
per-push gate: `docker.yml` builds `linux/arm64` on this same runner label on
every branch and asserts the embedded commit matches the built SHA. What no lane
covered is the other half — that the published arm64 artifact pulls under Podman
and starts. So this lane pulls.

The stop condition does not fire: `ghcr.io/kubestellar/hive:v4-latest` publishes
an arm64 image today (`sha256:cc2b1350…`, verified against the registry). If it
ever stops, the lane does **not** build one ad hoc — it records the absence and
fails, pointing at the publisher.

## Verified locally

The probe was run end to end against the real published image with `--arch
amd64`, which exercises every line of the same code path on the architecture I
could run:

```
--- case: manifest advertises linux/amd64 ---
  architectures advertised: amd64 arm64
  RESULT: ok — the tag publishes a linux/amd64 image
--- case: pull linux/amd64 ---
  digest: sha256:ee1c08e58b695e5c74187fcf0caf6fbcb812315a09730355a6770e52fa3444f9
  architecture: amd64
  RESULT: ok — pulled image is amd64
--- case: the shipped binary executes ---
  hive --version: hive 3.0.0 (commit 3207e02, branch v4)
  RESULT: ok — /usr/local/bin/hive runs on amd64
--- case: the service starts (GET /api/health) ---
  entrypoint milestones:
    [entrypoint] WARN: proxy egress enforcement is ADVISORY-ONLY …
    [entrypoint] Starting Go binary on :3002 (uid=1001)
    [entrypoint] Starting Node.js proxy on :3001 → :3002
    [entrypoint] Starting ttyd on 127.0.0.1:7681
  GET /api/health -> 200
  RESULT: ok — the service started and answered its readiness probe on amd64

SUMMARY: 0 failure(s)   (exit 0)
```

The failure paths were exercised too, not just asserted:

| Case | Result |
| --- | --- |
| run on a non-matching host architecture | exit `78`, refuses rather than measuring emulation |
| reference with no image for the target arch | exit `1`, records the absence, stops before the cascade |

The arm64 run itself happens on this PR — `ubuntu-24.04-arm` is not something I
can exercise locally, which is why the workflow asserts `arch=arm64` at runtime
rather than trusting the label.

## Boundaries

The probe pulls into a throwaway store with a private graphroot and runroot and
removes only the container it created, by name. `contents: read`, no secrets, no
`pull_request_target`, no engine socket mounted. No path filters — #4339 is what
a guard scoped into silence looks like; the cost here is one image pull and one
container.

`src/docs/podman-ci-runner-map.md` lane 3 gets a "Built (#4336)" note, matching
how the other lanes are recorded.

Part of #4188 — this PR does not close the parent. Closes #4336.

### DIFF
--- a/.github/workflows/podman-arm64-lane.yml
+++ b/.github/workflows/podman-arm64-lane.yml
@@ -1,0 +1,187 @@
+name: Podman arm64 Lane
+
+# The arm64 image pull and service-startup lane (#4336), lane 3 of the map in
+# src/docs/podman-ci-runner-map.md (#4211).
+#
+# Every Podman result before this one is amd64 only — #4199 and #4200 both
+# record arm64 as unproven. This lane closes that on the two questions #4188
+# actually asks of arm64: does the published arm64 image come down, and does
+# the service start from it.
+#
+# SCOPE, AND THE FAILURE MODE IT AVOIDS. #4211 names a duplicated full matrix
+# as the thing not to build here: the probe it ran found the two architectures
+# capability-identical, so re-running the egress-gate cases on arm64 doubles
+# cost for coverage that result says will not diverge. The gate lives on the
+# amd64 lanes — #4334 rootless, #4335 rootful. This lane runs four checks and
+# no more:
+#
+#   manifest   the tag advertises a linux/arm64 image
+#   pull       that image comes down and really is arm64
+#   exec       /usr/local/bin/hive runs
+#   startup    the container answers GET /api/health
+#
+# The probe runs the container with HIVE_PROXY_ADVISORY_OK=true so the egress
+# gate is taken OUT of the picture rather than quietly re-measured here.
+# Without it the entrypoint fails closed with exit 77 and nothing about startup
+# would be tested at all. If the architectures ever DO diverge, that is the
+# signal to widen this lane — not a reason to widen it now.
+#
+# WHY PULL RATHER THAN BUILD. #4188 asks for "build/pull plus startup", and the
+# arm64 BUILD is already a per-push gate: .github/workflows/docker.yml builds
+# linux/arm64 on this same runner label on every branch and asserts the
+# embedded commit matches the built SHA. What no lane covered is the other
+# half — that the published arm64 artifact pulls under Podman and starts. So
+# this lane pulls, and per #4336's stop condition it does NOT build an image ad
+# hoc: a missing arm64 manifest is reported and fails the job, because the fix
+# for that belongs in the publisher, not in a lane papering over it.
+#
+# A MISSING arm64 MANIFEST IS A FAILURE, NOT A SKIP. An arm64 lane that steps
+# aside when the arm64 image is absent reports green while proving nothing —
+# and "arm64 is unproven" is the exact condition this lane exists to end.
+#
+# Safety boundaries, from the map: contents: read and nothing more, no secrets,
+# no pull_request_target, no engine socket mounted. The probe pulls into a
+# throwaway store with a private graphroot and runroot, so nothing touches the
+# runner's default Podman storage, and it removes only the container it
+# created, by name.
+#
+# No path filters, on purpose. #4339 is what a guard scoped into silence looks
+# like. The cost here is one image pull and one container.
+
+on:
+  push:
+    branches:
+      - v2
+      - v4
+  pull_request:
+    branches:
+      - v2
+      - v4
+  workflow_dispatch:
+    inputs:
+      image:
+        description: 'Image to probe (default ghcr.io/kubestellar/hive:v4-latest)'
+        type: string
+        default: ''
+
+permissions:
+  contents: read
+
+concurrency:
+  group: podman-arm64-lane-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  arm64-probe:
+    name: arm64 image pull and service startup
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 30
+
+    steps:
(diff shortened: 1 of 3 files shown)


clubanderson:
Reviewed and verified.

- **Native arm64, not emulation**: runs on `ubuntu-24.04-arm` (the label `docker.yml` already proves in this repo) — no QEMU/binfmt anywhere, so no emulation flakiness to pad timeouts for. The 30-min job cap and 180s health window are generous for a native run that actually took 1m49s.
- **arm64 is proven three ways, not assumed**: the workflow asserts `podman info Host.Arch=arm64` at runtime rather than trusting the runner label; the probe refuses to run on a foreign `uname -m` (exit 78, explicitly declining to measure emulation); and after `pull --arch arm64` it asserts the *pulled* image's `.Architecture` — so a manifest regression cannot hand the lane an amd64 image, and the manifest check alone is never the proof.
- **Missing arm64 manifest fails rather than skips**, with a clear pointer at the publisher (`docker.yml`) instead of building ad hoc — correct per #4336's stop condition. Single-image-manifest vs index-omits-arch are distinguished in the output.
- **Self-trigger observed**: the lane ran on this PR's own head and passed in 1m49s (run 32420202035) — the arm64 run the author couldn't do locally. Cost is one pull + one container; sane.
- **Scope discipline**: advisory mode deliberately takes the egress gate out (otherwise exit 77 blocks startup testing entirely), with an agent configured so the real startup path (uid map, agent dirs) is exercised; `/api/health` matches the k8s readiness probe so "started" means production-started. The #3760 exec check is well-motivated. No duplicate matrix per #4211.
- **Safety**: `contents: read` only, no secrets, no `pull_request_target`, no socket mounts, throwaway store with by-name container cleanup via EXIT trap, no forbidden podman lifecycle commands. Probe shellcheck-clean; actionlint shows only the familiar info-level SC2016 on single-quoted Go templates, consistent with the merged sibling lanes.

/lgtm
/approve

PR #4353, opened 2026-08-20
🩺 podman: qualify SELinux-enforcing mounts, labels, and secrets per release

### LINKED ISSUE #4337: 🩺 Release qualification: SELinux-enforcing Podman
## Parent

## Why

#4211 measured that SELinux is **absent** on every hosted runner — no
`getenforce`, no `selinuxfs`, on both `amd64` and `arm64`. It is the one lane
that cannot be hosted. #4209 shipped SELinux/mount/secret preflight checks that
consequently have no CI coverage at all, and the standalone deployment's `:z`/`:Z`
mount behavior is exactly what breaks first on a Fedora/RHEL-class host.

## 30-minute boundary

Produce a documented, repeatable qualification procedure. Do not stand up
self-hosted infrastructure as part of this issue.

## Deliverable

A documented Fedora/CentOS Stream qualification job covering `:z`/`:Z` mounts,
label behavior, and secret access under SELinux enforcing.

## Acceptance criteria

- Reproducible from the document alone, on a stock enforcing host.
- Explicitly stated as **not** a hosted lane, with #4211's measurement as the reason, so nobody later "fixes" it by moving it to `ubuntu-latest`.
- The result is recorded per release rather than per PR.
- Remediation advice never disables SELinux or widens a secret — the same constraint #4209's preflight already holds to.

## Stop conditions

If no enforcing host is available, record the procedure and mark it unexecuted rather than running it somewhere non-enforcing and reporting a pass.

Danathar:
## What

The one Podman lane hosted CI cannot run:

- `src/docs/podman-selinux-release-qualification.md` — the procedure, the
  remediation catalogue, and the results ledger.
- `src/deploy/qualify_podman_selinux.sh` — the executable form of it.

No self-hosted infrastructure is stood up, per the 30-minute boundary.

## Why it is not a hosted lane, and why that is stated so loudly

#4211 measured it rather than assuming it: hosted runners have **no SELinux at
all** — no `getenforce`, no `/sys/fs/selinux`, AppArmor instead — on both
architectures. The document leads with that measurement because the failure
mode of "fixing" this by moving it to `ubuntu-latest` is the quiet one: every
check would still pass on a host where `:z` and `:Z` are no-ops and nothing is
ever denied, and the green tick would mean the opposite of what it appears to.

The same argument rules out the tempting third option — a Fedora *container* on
an Ubuntu runner gives no enforcement either, because the host kernel enforces
and it has none.

## Why it matters

#4209 shipped the SELinux/mount/secret preflight, and because of the
measurement above **none of it had ever run where the kernel enforces
anything.** Its advice — `:z` for checked-out config files, `:Z` for the
secrets directory — was reasoned from the policy, not observed. This
qualification runs it on an enforcing host and checks the advice against what
the kernel actually does.

## Executed, not just written

This host is Fedora 44, SELinux **Enforcing**, `targeted` policy — so the
procedure was run rather than shipped unexecuted, against the real release
image `ghcr.io/kubestellar/hive:v4-latest`. Every check passed:

| Check | Measured |
| --- | --- |
| enforcement control | unlabeled mount **denied**, `user_home_t` label untouched |
| `:Z` | read succeeded → `container_file_t:s0:c504,c650` |
| `:Z` isolation | second container without that category **denied** |
| `:z` | read succeeded → `container_file_t:s0`, no category |
| `:z` sharing | second container read it |
| secrets | mode stayed `600` through both relabels |
| #4209 preflight | detected Enforcing, exit 0, changed no label and no mode |

The `:Z`-then-denied result is the operational finding: it is exactly what
happens to every *other* container needing a file that was mounted `:Z`, which
is why the preflight says `:z` for checked-out files and `:Z` only for the
secrets directory. The preflight's advice is now observed, not inferred.

The ledger row records what the run does **not** cover, rather than letting it
read as broader than it is: the host is a Fedora 44 atomic variant
(Aurora/Kinoite), not stock Fedora Server or CentOS Stream, and it was
rootless only.

## Acceptance criteria

- **Reproducible from the document alone.** There is a "Doing it by hand"
  section with every command and its expected output, so the script is a
  convenience rather than the only path.
- **Explicitly not a hosted lane, with #4211's measurement as the reason.** Its
  own section, first, plus a note in the runner map's lane 4.
- **Recorded per release rather than per PR.** A ledger with a `Release`
  column; the script prints a ready-formatted row with the release cell left as
  a placeholder, because only the person cutting the release knows which one it
  is.
- **Remediation never disables SELinux or widens a secret.** No `setenforce`,
  no `--security-opt label=disable`, no `--privileged`, no `chmod` wider — the
  remediation table offers labeling and policy actions that leave the kernel
  enforcing, matching the constraint #4209's preflight already holds to. The
  suite *asserts* modes are unchanged rather than adjusting them.

## The stop condition is enforced in code

"Record it unexecuted rather than running it somewhere non-enforcing and
reporting a pass" is not left to the operator's discipline. A missing
`/sys/fs/selinux`, a missing `getenforce`, or any mode other than `Enforcing`
exits `78`, prints `NOT QUALIFIABLE HERE`, and reports no result. Verified by
running the suite behind a `getenforce` that reports `Permissive`: exit `78`,
no pass emitted.

There is a second anti-vacuous guard inside the suite. The **enforcement
control** runs first: an unlabeled bind mount must be *denied* before any
"access was granted" result is allowed to mean anything. If the control passes,
the run aborts rather than continuing — on such a host the remaining checks
prove nothing. (This also caught a real bug during development: a naive `*:c*`
test for a private MCS category matches the `:container_file_t` in the type
field and reported every label as private. It now examines only the level
field.)

## Safety

The suite relabels **only a fixture directory it creates itself** and removes
on exit. It is never pointed at a checkout or a real secrets directory — `:Z`
relabels in place, so doing that to an operator's files would be a change to
the host rather than a test of it. It runs no `setenforce`, changes no mode or
owner, and installs nothing. `--store` is available for a run that must leave
the operator's image storage untouched, though the host's own store is the
default since a release qualification should measure the host as configured.

Part of #4188 — this PR does not close the parent. Closes #4337.


clubanderson:
Reviewed on `c39dab52`. The SELinux reasoning holds up:

- **`:z`/`:Z` semantics are correct and — the key point — measured, not asserted.** The `:Z`-then-second-container-denied step is exactly the failure mode of `:Z` on anything shared, and both the doc and the remediation table say so plainly. `container_file_t` at `s0` vs a private MCS pair is the right distinction, and the level-field-only `mcs_of` fix (avoiding the `:container_file_t` false match) is a nice catch.
- **The enforcement control first, abort-on-pass, is the right anti-vacuous design**, and the exit-78 NOT-QUALIFIABLE gate enforces the #4337 stop condition in code rather than prose. The 'do not move this to ubuntu-latest' framing is exactly right given #4211's measurement.
- **No secret leak surface**: the fixture is a throwaway string in a script-created dir, relabels touch only that dir, modes are asserted unchanged (the only `chmod` narrows the fixture to 600), and no `setenforce`/`label=disable`/`--privileged` appears as a remedy.
- **Consistent with the neighbours**: no cleanup-contract literals; doesn't label or mount `/data`, so nothing to reconcile with the #4341 CA-key relocation; the qualification's transient `--rm` fixture containers are probe-class and outside the #4210 teardown scope (and `--store` exists for a fully isolated run).

Verified locally: `shellcheck -x -S style` clean, `bash -n` clean, and a trial merge with current `v4` (post-#4349/#4352 runner-map edits) auto-merges cleanly with teardown 74/74 and cleanup contract 127/127 still green. CI fully green including the new rootless/rootful egress lanes and the Quadlet gate. Body closes #4337. Merging.

/lgtm
/approve

PR #4355, opened 2026-08-20
🧩 podman: add the Hive service Quadlet unit and its volume

### LINKED ISSUE #4354: 🧩 Podman: add the Hive service Quadlet unit and its volume
## Parent

## Why

Everything decided so far describes Podman; nothing yet **runs** Hive on it. An
operator who follows the support matrix (#4324), passes the preflight (#4207,
#4208, #4209) and reads the lifecycle decision (#4325, ADR-0017) still has no
unit to start. This is the slice that turns the documentation into a deployment.

ADR-0017 settled the mechanism, so this is now scopeable:

> Standalone Hive under Podman uses Quadlet units — `.container`, `.pod`,
> `.network`, and `.volume` — read by the Podman systemd generator, rootful from
> the system manager's unit search path and rootless from the per-user
> manager's, with `loginctl enable-linger` for boot persistence in the rootless
> case.

and named the decisive property: `Notify=healthy` generates `Type=notify` /
`NotifyAccess=all` / `--sdnotify=healthy`, holding the unit in `activating`
until the healthcheck passes, so *started* means *healthy* rather than merely
spawned (#4202).

## 30-minute boundary

The **hive service unit itself**, and the volume unit it needs in order to
start. Explicitly out of scope, each being its own slice:

- the gateway and the network boundary — that is the next bullet on #4188
- persistence and SELinux volume *behavior* — define the volume, do not
  characterise its semantics here
- update, rollback, and auto-update
- backup, restore, and migration

Do not change Docker behavior, and do not modify `src/docker-compose.yaml`.

## Deliverable

Quadlet assets that start the Hive service under Podman in both root modes, with
the operator steps to install and enable them.

## Acceptance criteria

- Units generate cleanly under `quadlet --dryrun` for **both** `--user` and
  rootful. The gate added by #4338 (`.github/workflows/quadlet-gate.yml`,
  `src/deploy/test_quadlet_generator_gate.sh`) must cover the new units rather
  than passing vacuously beside them.
- The unit uses `Notify=healthy`, and the doc states plainly that
  `systemctl start` returning means the healthcheck passed — that property is
  the reason Quadlet was chosen over the alternatives.
- The image reference comes from the single source of truth,
  `src/deploy/standalone-images.sh` (#4206), so the Podman and Docker assets
  cannot drift. `src/deploy/test_standalone_image_refs.sh` should fail if it
  does.
- Rootless boot persistence documents `loginctl enable-linger`, and the rootful
  and rootless unit search paths are both stated.
- The minimum Podman version is taken from ADR-0017 and repeated accurately —
  **recommended 5.6.0, verified floor unknown**. Do not restate 5.8.4 as a
  floor; it is the version the spikes happened to run on.
- Docker is untouched: `src/docker-compose.yaml` unchanged, and the standalone

## Stop conditions

If the available environment cannot take the units to a live `systemctl start`,
generate them, run the dry-run gate, and record that a live start was **not**
performed — do not describe the service as starting on evidence that is only a
dry run. If ADR-0017's mechanism turns out not to survive contact, stop and
record what broke rather than switching mechanism inside this issue; that would
reopen the decision, which is #4325's document to change.

Danathar:
## What

The first slice of ADR-0017 turned into something an operator can start:

- `src/deploy/quadlet/hive.container`
- `src/deploy/quadlet/hive-data.volume`
- `src/docs/podman-standalone-quadlet.md` — install, boot persistence, traps

The service and the volume it needs, and nothing else. The gateway and the
network boundary are the next slice, so **nothing is published** — 3002/3001
are reachable from the host the unit runs on and 7681 is never published at
all.

**Docker is untouched.** `src/docker-compose.yaml` is unchanged and
`test_standalone_service_contract.sh` still passes 12/12.

## Readiness, set up so it actually means something

`Notify=healthy` generates `Type=notify`, `NotifyAccess=all`, and
`--sdnotify=healthy`, so `systemctl start` returning means `/api/health`
answered.

`HealthCmd=` is set **explicitly**, and that is the load-bearing detail: the
generator accepts `Notify=healthy` with no healthcheck and exits 0, producing a
unit that claims "started" the moment conmon is up. No dry-run gate can catch
that — including the one I added in #4338. Only setting it can.

## Verified live, not just dry-run

Live rootless start on Fedora 44, Podman 5.8.4, cgroup v2, SELinux enforcing,
against the real `ghcr.io/kubestellar/hive:stable`:

| | Observed |
| --- | --- |
| `daemon-reload` | generated `hive.service` + `hive-data-volume.service` |
| generated properties | `Type=notify`, `NotifyAccess=all`, `TimeoutStartUSec=5min` |
| ordering | `Requires=` / `After=hive-data-volume.service` |
| `systemctl --user start hive.service` | **returned 0 in 52s**, `ActiveState=active`, `Result=success` |
| container | `healthy` |
| `AddCapability=NET_ADMIN` | enforcing egress path: `outbound :443 -> :18443`, `ambient CAP_NET_ADMIN granted` |

`Notify=healthy` was also confirmed **negatively**, which is the half that
matters more. On a first attempt with no `HIVE_DASHBOARD_TOKEN`, Hive refuses to
start — and the unit correctly stayed at `activating`/`start` and never reported
started. A service that is not serving does not get to look started.

**Rootful was not started live.** It is dry-run generated only, and the doc says
so rather than implying otherwise. The rootless `%E` expansion was measured with
a probe unit; the rootful `/etc` value is cited from `systemd.unit(5)`.

## Acceptance criteria

- **Both dry-run modes, covered by the #4338 gate rather than passing beside
  it.** The gate scans repo-wide by design, so it picked these up with no edit:
  `2 unit(s) found: hive.container, hive-data.volume` → PASS in both `system`
  and `user` modes.
- **`Notify=healthy`, and the doc states plainly what `systemctl start`
  returning means.** Its own section, near the top.
- **Image from the single source of truth.** No change to
  `test_standalone_image_refs.sh` was needed — it already globs
  `src/deploy/quadlet/*.container` and reads the Quadlet `Image=` key, so drift
  detection switched itself on (`17 assertions across 2 asset(s)`, was 1). I
  proved it bites rather than assuming: naming the retired `v2-latest` fails,
  and naming a different image fails on the dropped digest pin.
- **`loginctl enable-linger` documented, both search paths stated** — from
  `podman-systemd.unit(5)`, rootful and rootless tables.
- **Version floor quoted as ADR-0017 states it:** recommended **5.6.0**,
  **verified floor unknown**, hard minimum 5.0.0 for `Notify=healthy`. 5.8.4 is
  explicitly *not* restated as a floor. The unit joins no `.pod`, so
  podman#26105 does not apply to it.
- **Docker untouched**, as above.

## Two traps, found by measurement

Both generate cleanly, so neither is catchable by a dry-run. Recorded in the
unit and the doc so they are not helpfully reintroduced:

1. **`EnvironmentFile=` does not honour systemd's `-` optional-file prefix.**
   `EnvironmentFile=-%E/hive/hive.env` resolves to
   `<unit dir>/-%E/hive/hive.env` — the dash stops the value looking absolute,
   so Quadlet joins it to the unit directory. The file must exist instead.
2. **`GlobalArgs=` pinning a non-default image store silently breaks
   healthchecks.** The container starts and then never becomes healthy, because
   Podman schedules the healthcheck as its own transient timer running a bare
   `podman healthcheck run <id>` with no store arguments, which fails with
   `no container with name or ID ... found`. This one cost me a run to find,
   which is exactly why it is written down.

Also documented: `TimeoutStartSec` must exceed `HealthStartPeriod` plus the
retry budget (300s vs 120s + 3×10s), because the generated `ExecStart` carries
`--rm` and a timeout deletes the evidence; and `:Z` not `:z` on the secrets
mount, per the measured difference in the SELinux qualification.

## Host hygiene

The live run created exactly two objects on the test machine — the
`ghcr.io/kubestellar/hive:stable` image and a `hive-data` volume — and both were
removed afterwards. The image count is back to its 16-image baseline. A
pre-existing `src_hive-data` volume from a Compose run was identified and left
alone.

Part of #4188 — this PR does not close the parent. Closes #4354.


clubanderson:
Reviewed on `16158ddc` and fixed forward with `32a25b0e`. The unit itself is excellent — one contract gap, now closed:

**Fix-forward: #4210 ownership labels.** Neither `hive.container` nor `hive-data.volume` carried the ownership label set, and the teardown (#4343) selects by those labels and nothing else — so the very container and volume this slice creates would have survived every `hive-podman-teardown.sh` run. The contract doc is explicit that deployment assets must carry the labels at create time. I added the four `Label=` keys to both units (instance is the contract's default, with a comment on multi-instance hosts) and a short doc note. The declarative-unit shape is also why the teardown suite's asset scan didn't trip: its pattern matches shell-style `podman run|create|build`, not Quadlet keys — worth widening in a future slice.

Everything else holds up:
- `Notify=healthy` + explicit `HealthCmd` is the right readiness posture, and the negative confirmation (no token → stays `activating`) is the half that matters.
- Consistent with the merged #4342 doc: TimeoutStartSec=300 > 120s+3×10s, 5.0.0 hard floor / 5.6.0 recommendation quoted correctly, joins no `.pod` so podman#26105 doesn't apply.
- Hardening sane: no `--privileged`, only `AddCapability=NET_ADMIN` for the egress redirect, nothing published, 7681 never exposed.
- `:Z` on operator-local config/secrets is right per the #4353 measurement (single consumer, not a checkout).
- The #4349 gate scans repo-wide and picked both units up (`2 unit(s) found` → quadlet lane green); `test_standalone_image_refs.sh` switched itself on (17 assertions across 2 assets).
- The EnvironmentFile `-` prefix and GlobalArgs-healthcheck traps are valuable measured findings.

Verified locally on the fixed head: image refs 17/17, teardown 74/74, cleanup contract 127/127, service contract 12/12, no cleanup-contract literals. CI fully green on `32a25b0e` including rootless/rootful/quadlet lanes. Body closes #4354. Merging.

/lgtm
/approve

PR #4357, opened 2026-08-20
🧪 podman: record the actual AVCs, and the hive-launch group secret

Danathar:
## What

The evidence follow-up to #4337, on the same enforcing host:

- `src/deploy/probe_podman_selinux_avc.sh` — the probe.
- `src/docs/podman-selinux-avc-evidence.md` — the result, the reproduction by
  hand, and the ledger.

#4337 is not superseded. It established the `:z`/`:Z`/unlabeled matrix; this
adds what that matrix could not say.

## Why

Two gaps.

**Evidential.** #4337's suite decides "denied" from the container's exit
status — it never reads the audit log. So every denial it reports is
**inferred**, and a failure with nothing to do with SELinux would look
identical. This takes an AVC baseline before any container starts and diffs
each case against it, so a denial is *observed*.

**The secret case the image actually depends on.** #4337 tests a `0600` file
read by its owner. The release image reads `0440` owned by **hive-launch**
(GID 1002, pinned in `src/Dockerfile`, `fsGroup` in the k8s manifests) as
`dev` (UID 1001) through a *supplementary* group. DAC and MAC both have to
permit that, and only one of them is what `:z`/`:Z` addresses.

## Every case passed — and that is not the interesting part

Three findings say a shipped check or a shipped remediation is wrong on a host
of this class. Two are in #4209's preflight, whose advice had never run
anywhere enforcing.

**1. The preflight's label check reads garbage where uutils shadows GNU coreutils.**
It reads labels with `stat_out="$(stat -c '%C' "$path" 2>/dev/null)" || stat_out=""`.
On Bluefin/Universal Blue — the Fedora-atomic class this lane targets — uutils
`stat` has no `%C` and prints to **stdout with exit 0**, so the guard never
fires and the string is compared against `container_file_t` as if it were a
label. Its actual output, verbatim:

> △ Mount label: deploy/nginx.conf (gateway configuration) is unsupported for this operating system — a container cannot read it as-is

`_pfh_label_is_container_readable` can then never match, so on such a host the
check **can never report a correctly labelled path**. The `:z` advice it gave is
right by luck, not from reading the label. (`ls -Z` is not a fallback — uutils
accepts it and prints nothing. `/usr/bin/stat` and `getfattr` both work.)

**2. `mkdir -m 700` makes the secrets directory unreadable under rootless Podman.**
The preflight advises exactly that, and `:Z` for the mount. Rootless, the host
user maps to container **root**, not to `dev`, so `dev` (1001) does not even get
the traverse bit — and no labeling action fixes it. Three measurements, one
variable:

| Secrets dir | Mount | Read as `dev` | Refuser |
| --- | --- | --- | --- |
| `0700`, owner = host user | `:z` | **denied** | DAC — no AVC |
| `0750`, group 1002 | unlabeled | **denied** | MAC — AVC recorded |
| `0750`, group 1002 | `:z` / `:Z` | **granted** | — |

Row 1 is the point: SELinux is already satisfied there, and it still fails. The
k8s path gets this right via `fsGroup: 1002`; the standalone advice carries no
equivalent.

**3. MCS category denials are silent.** The `:Z` isolation failure — a second
container refused a file the first one stamped — produces **no audit record of
any kind**, no AVC and no `SELINUX_ERR`, reproduced twice. A *type* denial on
the same host is audited. #4337's remediation table sends operators to
`ausearch` for exactly this case, and it returns nothing.

## What the preflight got right

Detected `Enforcing` and Podman labeling; changed no label and no mode; offered
no `setenforce`, no `label=disable`, no `--privileged`, no widening `chmod`.
Neither `:z` nor `:Z` widened the `0440` secret or touched its group ownership.

Its exit `78` in the transcript is **not** an SELinux result — it is a missing
`hive.yaml` in the checkout. The document says so explicitly so nobody reads
that code as a qualification failure.

## Nothing was weakened to get a pass

No `setenforce`, no `label=disable`, no `--privileged`, no widening `chmod`, no
`chcon` outside the probe's own fixture. The first run failed the group-secret
case; the fixture was corrected because **the fixture was wrong** (its secrets
dir was `0700`, conflating DAC with MAC), and correcting it is what produced
finding 2. The enforcement was never relaxed.

## Safety

`:z` and `:Z` relabel in place and the relabel outlives the container, so the
only mount target is a fixture the probe creates under `/var/tmp` and removes
through `podman unshare`. Every podman call is pinned to a throwaway
`--root`/`--runroot`; the probe refuses `/var/lib/containers`,
`/run/containers`, and the host `graphRoot`, the same guard as #4200. Verified
after the run: the host store held **0 containers and 0 hive images**, and the
checkout's `nginx.conf` was still `user_home_t` mode `644`. Nothing installed,
nothing rebooted — `/usr` is read-only here and `rpm-ostree` was used only to
query versions.

The probe **fails rather than reporting a clean run** if it cannot read the
audit log, and exits `78` reporting nothing at all unless the host is
`Enforcing`.

## What remains unproven

Rootless and `amd64` only, on a Fedora 44 **atomic** variant rather than CentOS
Stream. The mechanism behind finding 3 is *inferred* (MCS as a constraint rather
than an allow rule) — confirming it needs `semodule -DB`, which changes system
state, so it was not done. Only `open`/`read` was exercised, never a write. No
deployment was started: every container here was `/bin/sh -c cat`, so nothing
claims the standalone stack works under enforcing. Findings 1 and 2 are recorded
but **not fixed** — changing the preflight belongs to #4209.

Part of #4188 — this PR does not close the parent.


clubanderson:
Reviewed on `c7a96e46` (merged `origin/v4` forward as `05621579` to resolve the docs-index collision with #4355 — content untouched, both entries kept). This is exactly the follow-up #4353 needed:

- **The AVC evidence is methodologically sound**: per-second baseline marks with the same-second sweep guard, container-domain scontext filtering against ambient noise, polled collection so "no denial" means five seconds of silence, and an honest UNMEASURED/fail path when no audit reader exists. The measurements are consistent with #4353's matrix (type denial audited on the control, `container_file_t` + category semantics identical), and the three findings check out against the tree: the preflight really does `stat -c '%C' … || stat_out=""` at `bin/hive-podman-preflight-host.sh:240` (the uutils exit-0 case slips the guard), and `src/Dockerfile` pins hive-launch GID 1002 / dev UID 1001 exactly as the group-secret case uses them.
- **Finding 3's epistemics are right**: the silent MCS denial is recorded as a measurement, the constraint mechanism explicitly flagged as inferred, and `semodule -DB` — the one tool that would confirm it — correctly declined because it mutates host state.
- **Secret posture is not weakened**: fixture-only content (`not a real key`), 0440 mode and group asserted unchanged through both relabels, no widening chmod, no `setenforce`/`label=disable`/`--privileged`, `chcon` confined to the probe's own fixture, throwaway `--root`/`--runroot` with the host-store refusal guard, containers removed by exact name.
- Findings 1 and 2 correctly recorded-not-fixed (they belong to #4209's preflight — worth a follow-up issue so they don't rot in a doc).

Verified locally: `shellcheck -x -S style` clean, no cleanup-contract literals, teardown 74/74 and cleanup contract 127/127 green on the merged tree. CI fully green on `05621579` including rootless/rootful/arm64/quadlet lanes, DCO good. Merging.

/lgtm
/approve

PR #4361, opened 2026-08-20
🐛 podman preflight: resolve the label reader, and make the secrets dir reachable

### LINKED ISSUE #4359: 🐛 podman preflight: mount-label check reads garbage under uutils coreutils, and the secrets-dir remediation cannot work rootless
## Why now

#4209 shipped `bin/hive-podman-preflight-host.sh` with SELinux, mount, and
secret checks. #4211 measured that **no hosted runner has SELinux at all**, so
none of that advice had ever run where the kernel enforces anything — it was
reasoned from the policy, never observed.

#4337 qualified the mount matrix on an enforcing host. #4357 then ran the
preflight itself there with the audit log open, and found two defects. Both are
in the preflight; neither is fixed by #4357, which records them and stops.

Measured on: Fedora 44 atomic (Bluefin 44.20260811), SELinux **Enforcing**,
policy `targeted`, podman 5.8.4 rootless, `container-selinux-2.250.0-1.fc44`.

---

## 1. The mount-label check reads an error string as if it were a label

`hive_podman_check_mount_labels` reads the label with:

```sh
stat_out="$(stat -c '%C' "$path" 2>/dev/null)" || stat_out=""
```

On a host where **uutils coreutils** is ahead of GNU coreutils on `PATH` —
the default on Bluefin/Universal Blue, i.e. the Fedora-atomic class the SELinux
lane targets — `stat` has no `%C`:

```
$ stat -c '%C' /var/tmp/f      # uutils coreutils 0.9.0
unsupported for this operating system
$ echo $?
0
```

It writes to **stdout and exits 0**, so `|| stat_out=""` never fires and the
string flows into the comparison as a context. Actual preflight output on such a
host:

```
△ Mount label: deploy/nginx.conf (gateway configuration) is unsupported for
  this operating system — a container cannot read it as-is
```

The garbled sentence is the small part. `_pfh_label_is_container_readable` can
never match, so **the check can never report a correctly-labelled path** — it
warns unconditionally, including for a file already labelled `container_file_t`
by `semanage fcontext`/`restorecon`, which is the exact state its own
remediation tells operators to reach. An operator who follows the advice and
succeeds still gets told it did not work.

The `:z` advice it prints happens to be correct, but it is correct by luck
rather than from reading the label.

`ls -Z` is not a fallback: uutils `ls` accepts `-Z` and prints no context at
all. Both of these work here:

```
$ /usr/bin/stat -c '%C' /var/tmp/f
unconfined_u:object_r:user_tmp_t:s0
$ getfattr -n security.selinux --only-values /var/tmp/f
unconfined_u:object_r:user_tmp_t:s0
```

### Suggested fix

Resolve the reader once, by trying it against a known file and requiring a
`:object_r:` field in the result, rather than trusting `stat`. #4357's probe
does this in `_resolve_label_reader` and can be lifted. Failing that, at
minimum validate `stat_out` looks like a context before comparing it, and treat
"cannot read a label" as its own outcome rather than as "unlabelled".

---

## 2. `mkdir -m 700` for the secrets directory cannot work under rootless Podman

The preflight advises:

```
→ Only needed for GitHub App key auth. Create it narrow: mkdir -m 700 -p '.../secrets'
```

and lists `secrets|secrets directory|Z` in `HIVE_PODMAN_BIND_SOURCES`, i.e.
mount it `:Z`.

Rootless, the host user maps to container **root**, not to `dev`:

```
$ podman run --rm --entrypoint /bin/sh $IMAGE -c 'id -u'
0
```

The image reads GitHub App keys as `dev` (UID 1001) via the `hive-launch`
supplementary group (GID 1002, pinned in `src/Dockerfile`). A `0700` directory
owned by the operator grants `dev` nothing — **not even the traverse bit** — so
the key is unreadable no matter how it is labelled. `:Z` fixes the label and
does nothing about traversal.

Measured three ways, one variable, `0440` key owned by group 1002 throughout:

| Secrets dir | Mount | Read as `dev` (1001) | Refuser |
| --- | --- | --- | --- |
| `0700`, owner = host user | `:z` | **denied** | DAC — **no AVC recorded** |
| `0750`, group 1002 | unlabeled | **denied** | MAC — AVC recorded |
| `0750`, group 1002 | `:z` / `:Z` | **granted** | — |

Row 1 is the finding: SELinux is already satisfied there and it still fails,
with an empty audit log. The failure presents as `Permission denied` on a key
file on an enforcing host, which reads as an SELinux problem, is not one, and
**no labeling action the preflight offers will change it.**

The Kubernetes path already gets this right — `fsGroup: 1002` in
`src/deploy/k8s/deployment.yaml` produces a group-owned directory with the group
traverse bit. The standalone Podman advice carries no equivalent.

### Suggested fix

The remediation should produce a directory the container's runtime user can
traverse, without widening it to the world — e.g. group-owned by the launch GID
with `0750`, the standalone analogue of `fsGroup`. A check that the secrets
directory is *traversable by the UID the engine will actually run as* would
catch it directly. Whatever the shape, it must not become `chmod 0755` on a
secrets directory: narrowing-only is #4209's stated constraint and still holds.

---

## Acceptance criteria

- The label check reports a correctly-labelled path **as correct** on a host
  where uutils coreutils shadows GNU, or reports plainly that it could not read
  a label — never silently treats an error string as a label.
- The secrets remediation, followed literally on a rootless host, yields a
  directory whose key files the container can actually read.
- No remediation disables SELinux or widens a secret — unchanged from #4209.
- Ideally regression-covered, though note #4211: the label-reader bug is
  reproducible on any host with uutils on `PATH` and needs no SELinux, so that
  half **can** be tested in hosted CI even though the enforcing half cannot.

## Evidence

- #4357 — the run that found both, with the full transcript
- `src/docs/podman-selinux-avc-evidence.md` — findings 1 and 2, with the
  measurements above
- `src/deploy/probe_podman_selinux_avc.sh` — re-runs it; `_resolve_label_reader`
  is the reference fix for the first defect

## Not in scope

Fixing the third finding from that run — MCS category denials produce no audit
record of any kind, so #4337's `ausearch` remediation returns nothing for the
most likely `:Z` failure. That is a documentation defect in
`podman-selinux-release-qualification.md`, not in the preflight.

Danathar:
## What

Fixes both defects #4357 found when it ran the #4209 preflight on a real
enforcing host. Both were re-measured here before anything was changed.

## 1. The label check read an error string as a label

`stat -c '%C'` is not a reliable reader. Under **uutils coreutils** — the
default on the Fedora-atomic hosts this lane targets — `stat` has no `%C` and
answers on **stdout** with **exit 0**, so `|| stat_out=""` never fires and the
sentence flows into the comparison as a context.

Reproduced with a uutils-shaped `stat` ahead of GNU on `PATH`, against a tree
already labelled `container_file_t` — the exact state the preflight's own
remediation tells operators to reach:

```
# before
△ Mount label: hive.yaml (Hive configuration) is unsupported for this operating system — a container cannot read it as-is
△ Mount label: deploy/nginx.conf (gateway configuration) is unsupported for this operating system — a container cannot read it as-is
△ Mount label: secrets (secrets directory) is unsupported for this operating system — a container cannot read it as-is

# after
✓ Mount label: hive.yaml is unconfined_u:object_r:container_file_t:s0 — already container-readable
✓ Mount label: deploy/nginx.conf is unconfined_u:object_r:container_file_t:s0 — already container-readable
✓ Mount label: secrets is unconfined_u:object_r:container_file_t:s0 — already container-readable
```

The reader is now **resolved, not assumed**: `stat`, `/usr/bin/stat`,
`getfattr`, each tried against a probe and required to return something
context-shaped (`*:object_r:*`) before it is trusted. With no working reader:

```
△ Mount labeling: no tool on this host can read an SELinux label
  → Install GNU coreutils or attr (getfattr), then re-run. Mount labeling is
    UNCHECKED until then — this is not a pass.
```

One design point worth calling out. The probe is deliberately **not** a
deployment path. Probing on one would reconflate the two states this fix exists
to separate — a reader that does not work, and a path that genuinely has no
label. I hit that while building it: probing the first bind source made an
unlabelled fixture look like a broken reader, which is the same class of bug in
a new place.

## 2. `mkdir -m 700` excluded the container from its own secrets

The image reads GitHub App keys as `dev` (1001) through `hive-launch` (GID
1002, pinned in `src/Dockerfile`). A `0700` directory owned by the operator
grants `dev` nothing — not even traverse — and the refusal is **DAC**, so it
produces **no AVC at all**: it looks like an SELinux problem on an enforcing
host and no relabeling fixes it.

Measured on enforcing Fedora 44, rootless, one variable at a time, `0640` key
throughout:

| Secrets dir | Container read as uid 1001 + gid 1002 |
| --- | --- |
| `0700`, owner-only, `:Z` | **denied** |
| `0750` + `chgrp 1002` | **the `chgrp` itself fails** — `Operation not permitted` |
| `0750` + `podman unshare chown -R 0:1002`, `:Z` | **`APPKEY-CONTENT`** |

Row 2 is the part worth flagging: the obvious form of the fix suggested in the
issue is **not executable by an unprivileged operator** — you cannot `chgrp` to
a group you are not in. And rootless, container GID 1002 is
`subgid_start + 1001` on the host (525289 here), *not* 1002, so a plain `chgrp
1002` targets the wrong group even with privilege. `podman unshare` does the
translation. Chowning the directory alone is also not enough — I tested that
separately and the key files need it too, hence `-R`.

Preflight now computes that mapping from `/etc/subgid` and checks the directory
is group-owned by it with the traverse bit, printing the remedy for the root
mode in use. It stays read-only — starts no container — and the secrets maximum
moves `0700` → `0750`, which is the group bit the container needs and not a
step toward the world.

## Acceptance criteria

- **Correctly-labelled path reported as correct under uutils, or plainly
  unreadable** — both shown above, both pinned by tests.
- **The remediation, followed literally, yields a readable key** — verified by
  actually reading it out of a container at uid 1001 + gid 1002.
- **No remediation disables SELinux or widens a secret** — final state is
  `drwxr-x---` / `-rw-r-----`, nothing for *other*; asserted by tests that the
  output never contains `chmod 0755` or `setenforce`.
- **Regression-covered**: 157 assertions, up from 138. The label half needs no
  SELinux (the defect is in how the label is *read*), so it runs on the hosted
  runners #4211 measured as having none — which is the half the issue said
  could be covered in CI.

The test stub gained a `%g` column and a uutils-shaped `stat`; the suite fails
against the unfixed preflight and passes against this one. Sibling suites
unaffected (44 and 61 assertions still pass).

`src/docs/podman-preflight-host.md` is updated to match, including why `0700`
looks safer and is not.

## Not in scope

The third finding from that run — MCS denials producing no audit record, so the
`ausearch` advice in `podman-selinux-release-qualification.md` returns nothing
for the most likely `:Z` failure. That is a documentation defect in a different
file, as the issue says.

Part of #4188 — this PR does not close the parent. Closes #4359.


Danathar:
`overlayfs-exec-guard` is red here, and it is not this change — it needs a re-run, which I do not have rights to trigger.

The image build failed downloading bobshell in `src/Dockerfile:303`:

```
curl: (28) Failed to connect to s3.us-south.cloud-object-storage.appdomain.cloud port 443 after 134946 ms
curl: (28) ... after 134263 ms
curl: (28) ... after 134260 ms
ERROR: failed to solve: ... exit code: 28
```

Three attempts, ~135s each, exhausting `--retry-max-time 300`. The endpoint was unreachable for the ~7 minutes the job ran.

Why it cannot be the diff: this PR touches `bin/hive-podman-preflight-host.sh`, `bin/test_hive_podman_preflight_host.sh`, and `src/docs/podman-preflight-host.md`. Nothing under `bin/` is `COPY`d before line 303, and the failing step is a `RUN curl` that depends on none of those files. The same download runs on every build of this job regardless of what a PR changes.

The three checks that actually exercise this change are green: `build-and-test`, `test`, and the `Podman SELinux, mount, secret, and port preflight (#4209)` step that runs the 157-assertion suite this PR extends.

Happy to rebase if that is the easier way to retrigger it.

clubanderson:
Reviewed on `18db84ae`, merged `v4` forward and added one alignment commit `434ccf1e`. Both fixes are correct and well-pinned:

- **Label reader**: resolve-don't-assume is the right shape — candidates probed against `/etc`//`/` (deliberately not a deployment path, avoiding the exact conflation the bug is about), output required to be context-shaped (`*:object_r:*`, the check the uutils sentence fails), and "no working reader" reported as UNCHECKED rather than as unlabelled or a pass. The uutils-shaped test stub reproduces the stdout-exit-0 failure mode faithfully, and the suite fails against the unfixed preflight. No interaction with the #4210 ownership filters — this reads SELinux file contexts only, stays read-only, and issues no cleanup-shaped commands.
- **Secrets reachability**: the rootless mapping math is right (container GID g → `subgid_start + g − 1`, so 1002 → start+1001), the `chgrp`-not-permitted catch is real, and `podman unshare chown -R 0:1002` is the correct executable remediation. Posture is not weakened: max moves 0700 → 0750 (group traverse bit only), keys stay 0640, nothing for *other*, and the tests assert no widening chmod/setenforce ever appears in output — consistent with #4353's constraint and the k8s `fsGroup: 1002` analogue.
- **Fix-forward `434ccf1e`**: the #4355 Quadlet install doc still said `chmod 700 "$CONF/secrets"` — following it verbatim would trip the very warning this PR adds. Aligned it with the same rootless/rootful remediation.

Verified locally: `shellcheck -x -S style` clean on both scripts, no cleanup-contract literals, sibling suites 44/61 green (the host suite needs GNU stat, so its 157 assertions are covered by the green `build-and-test` lane — the macOS-local failure reproduces identically on the v4 baseline). The earlier `overlayfs-exec-guard` red was an infra flake (IBM COS unreachable mid-image-build), green on rerun. CI fully green on `434ccf1e` including all podman lanes. Body closes #4359. Merging.

/lgtm
/approve

Issue #2353, opened 2026-07-31
🐛 App-state health reports OK when the installation lacks write permission — failure only surfaces at post time


clubanderson:
A hive can report a completely healthy GitHub App state while being unable to perform its core function, because App **authentication** and App **authorization** are checked in different places and only the former feeds health.

## Live example

`open-horizon-services/Getting-Started` (hive `hosted-available-oke-05-placeholder-6q84`):

| field | value |
|---|---|
| `app_id` | `3568013` ✅ correct |
| `installation_id` | `149832219` ✅ real |
| `githubAppState` | `None` — **no problem reported** |
| `githubAppPermIssue` | `None` — **no problem reported** |
| `aiAuthorEffective` | `kubestellar-hive[bot]` ✅ authenticating fine |
| `advisoryError` | `posting advisory digest to Getting-Started#18: POST .../issues/18/comments: **403 Resource not accessible by integration**` |

The App authenticates, so every App-state check passes. It simply cannot **write**. The only evidence is `advisoryError`, surfaced as a "Stale advisory" warning — which reads as "the digest is out of date" rather than "this hive is structurally unable to post."

## Root cause (external, but detectable)

The installation on that org either lacks **Issues: Write**, or does not include `Getting-Started` in its selected-repositories list. Both are org-owner settings in GitHub and cannot be fixed from Hive.

**But we can detect it and say so.** Right now the operator sees a green App state and a vague staleness warning, and has no way to know the actual fix is "grant Issues: Write" or "add this repo to the installation."

## Why this will recur

Any org that installs the App with narrow scope — repo-selected rather than all-repos, or without Issues: Write — produces this exact state. It is the default-safe choice for a cautious org owner, so it should be expected rather than treated as an anomaly.

## Suggested

1. **Check the installation's permissions and repository selection**, not just that a token can be minted. `GET /installation/repositories` and the installation's `permissions` object are both available with the token the hive already holds.
2. When a write 403s, classify it into `githubAppPermIssue` with an actionable message — e.g. *"The App installation on `&lt;org&gt;` cannot write issue comments on `&lt;repo&gt;`. Grant Issues: Write, or add this repo to the installation's selected repositories."* — rather than leaving it only in `advisoryError`.
3. Distinguish it in the UI from ordinary advisory staleness. "Stale advisory" implies a timing problem; this is a permanent capability gap until an org owner acts.

## Related

- #2341 — an epp pod restarted 15,784 times with no operator-visible signal
[...]

PR #4364, opened 2026-08-20
✨ feat(github): detect repos the App installation does not cover

### LINKED ISSUE #4360: ✨ detect repos configured on a hive but missing from the GitHub App installation's selected repos
## The gap

A hive can be pointed at a repo that its GitHub App installation **does not cover**, and nothing tells the owner. The failure only surfaces later, lazily, as an agent action failing — and it is reported as something else entirely.

Found live on `kelly - headwaters` (`hosted-available-oke-260812-8yiz-available-ok-go05`, hive-oke):

- Org `AI-native-Systems-Research` has exactly **one** installation, `140943481`, with `repository_selection: "selected"`.
- That installation covers `ai-native-storage-certus` and `ai-native-storage-certus-workbench`.
- The hive's `project.repos` is `[headwaters]` — **not in the selected list.**
- The App is correct, the org is correct, the key is valid and mints a token fine. `GET /repos/AI-native-Systems-Research/headwaters` returns **404** (GitHub returns 404, not 403, for a repo an installation cannot see).

This is a **distinct failure mode** from the previously-seen installation problems: it is not a wrong App, not a wrong org, and not a wrong repo name (cf. `kalantar-msb` `softr` vs `soft-reflective`). It is a **second repo in the same org, on the correct installation, simply not ticked in the App's selected repos.**

## Why the current signal is actively misleading

The dashboard raised:

> **app-creds-undelivered** — GitHub App private key has not reached this spoke even though the hub holds a key for hive-oke — check heartbeat delivery, or re-upload via PUT `/api/saas/admin/cluster-app-keys/hive-oke`

The key **had** reached the spoke. Re-uploading it does nothing. The remedy named in the alert cannot fix the actual problem, so it sends the operator down the wrong path.

## Proposed detection — one API call, deterministic

`GET /installation/repositories` returns exactly what the installation covers. Compare it to the hive's configured repos:

```
covered   := set(GET /installation/repositories)     # what the App can see
configured:= set(project.repos)                      # what the hive is told to work on
missing   := configured - covered
```

No error-code inference, no guessing. Verified against the live installation: the comparison cleanly yields `headwaters` as missing.

Today we never call this endpoint — `git grep 'installation/repositories'` across `src/` returns nothing — so the hive genuinely cannot know its own coverage.

## Where it fits

`RecheckGitHubApp` (`src/pkg/dashboard/server.go:1866`) is already the single place App state is verified and the banner cleared, shared by the periodic self-heal loop and the manual **Re-check** button. Adding the coverage check there means both paths gain it at once.

Suggested surface:
- A distinct condition (not `app-creds-undelivered`) naming the missing repos, e.g.
  `GitHub App is installed but does not include: AI-native-Systems-Research/headwaters — add it under the org's App configuration (Settings → Applications → kubestellar-hive → Configure → Repository access).`
- Ideally deep-link to `https://github.com/organizations/<org>/settings/installations/<installation_id>`, which is the page that actually fixes it.
- Detect at **claim/assign time** as well as on recheck: a hive assigned to a repo outside its installation is misconfigured from the first heartbeat, and that is the cheapest moment to say so.

## Why it matters

The owner and the hive admin both see "something is wrong with credentials" when the real answer is one checkbox in the App configuration. The information needed to say so precisely is one API call away, and the hive already holds a token that can make it.

## Not in scope

Automatically adding the repo to the installation. That requires an act by someone with org admin rights; the ask here is only that the product **name the problem correctly**.

Danathar:
## What

Asks GitHub directly which repos an App installation covers, and names the
configured ones it does not.

```
covered    := GET /installation/repositories
configured := project.repos
missing    := configured - covered
```

- `src/pkg/github/app_coverage.go` — `AppAuth.InstallationCoverage`, the
  comparison, and the deep link.
- `AppStateRepoNotCovered` — the verdict, wire token `repo-not-covered`.
- `src/cmd/hive/main.go` — `classifyGitHubAppRepoCoverage`, wired into the
  recheck path.

## Why the current signal was worse than nothing

On `kelly - headwaters` the App was correct, the org was correct, and the key
had arrived and minted tokens fine. The only thing wrong was that `headwaters`
was not ticked in the installation's selected repositories.

GitHub answers **404** for a repo an installation cannot see — the same answer
it gives for a repo that does not exist — so the read path reported "app not
installed / no read", and the dashboard raised **app-creds-undelivered**,
telling the operator to re-upload a private key that had never been missing.
The named remedy could not fix the actual problem.

## Where it sits among the existing states

This is the **deterministic** form of what `AppStateWriteForbidden` (#2353)
infers. That state is reached after a real write returns 403, for whichever
single repo something happened to touch. This one is established **before**
anything is attempted, for **every** configured repo at once, and does not have
to guess between "not covered", "renamed", and "does not exist".

So the coverage check runs *before* the advisory-issue read in the recheck
path — the specific message beats the generic one. Both the manual **Re-check**
button and the periodic self-heal loop go through `RecheckGitHubApp`, so both
gain it at once, as the issue suggested.

## Two ways it refuses to accuse

This check's whole purpose is to stop sending operators after the wrong thing,
so it is built not to become a new source of that:

- **An error is not a verdict.** If the listing cannot be fetched, the
  credentials are the likelier story and the existing checks tell it better, so
  it defers. It never returns an empty coverage set on failure — that would
  accuse every configured repo simultaneously.
- **A truncated listing accuses nobody.** An installation set to "all
  repositories" on a large org can list thousands; the walk is capped, and
  hitting the cap reports *nothing* missing. Absence from a partial set is not
  evidence of absence.

## The message

```
The GitHub App is installed and authenticated for 'AI-native-Systems-Research',
but its installation does not include: ai-native-systems-research/headwaters.
Nothing is wrong with the App, the organization, or the private key — these
repositories are simply not ticked in the installation's repository access, so
every call against them returns 404. Add them under the org's App configuration
(Settings → Applications → Configure → Repository access).
https://github.com/organizations/AI-native-Systems-Research/settings/installations/140943481
```

It says what is *not* at fault, because that is the part the previous signal got
wrong. The deep link is built from the API base, so a GitHub Enterprise hive is
not sent to github.com — asserted by test.

## The misleading alert stays silent, by construction

`app-creds-undelivered` is gated on an allowlist of three operator-side tokens
(`appStateIsOperatorSide`). `repo-not-covered` is owner-side — an org owner
ticks a checkbox — so it cannot fire. I added a hub test pinning that against
the hive that produced the original misdiagnosis, rather than relying on the
allowlist staying closed by accident.

## Tests

The live case runs end to end against a stubbed `/installation/repositories`:
the two storage repos covered, `headwaters` configured, exactly `headwaters`
reported missing, and the covered repos not accused. Plus pagination across
pages, case-insensitive repo names (GitHub's are), the truncation guard,
error-is-not-an-empty-set, the GHE link, no-link-without-an-installation-id,
and the wire-token round trip.

`go build ./...`, `go vet`, and full package runs green: `pkg/github`,
`pkg/hub`, `pkg/dashboard`, `cmd/hive`.

## Scope

Detection at **claim/assign time** — the issue's "ideally" — is not here. That
is a hub-side surface with its own state machine, and folding it in would have
made this change hard to review against the one it fixes. The recheck path is
where the issue pointed, and it covers both the manual and self-heal routes.

Automatically adding a repo to the installation is explicitly out of scope: it
requires org admin rights. The ask was that the product name the problem
correctly.

Closes #4360.

### DIFF
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -2422,6 +2422,20 @@ func main() {
 					dashSrv.AuditLog("system", "github_app_check", "result=no GitHub client: "+appAuthFailure, "")
 					return false
 				}
+				// #4360: ask about repo COVERAGE before attempting a read.
+				// A repo the installation does not cover answers 404, which is
+				// indistinguishable from "no such repo" and used to be reported
+				// as "app not installed / no read" — sending the operator after
+				// credentials that were never broken. Checking first means the
+				// specific, correct message wins over the generic one.
+				if raise, diag, state := classifyGitHubAppRepoCoverage(ctx, ghClient.AppAuth(), cfg.Project.Org, cfg.Project.Repos, logger); raise {
+					dashSrv.SetGitHubAppPermIssue(diag)
+					dashSrv.SetGitHubAppState(state.String())
+					logger.Warn("github app recheck: installation does not cover every configured repo",
+						"org", cfg.Project.Org, "state", state.String(), "detail", diag)
+					dashSrv.AuditLog("system", "github_app_check", "result=repos not in installation: "+diag, "")
+					return false
+				}
 				num, err := ghClient.EnsureAdvisoryIssue(ctx, recheckRepo)
 				if err != nil {
 					logger.Debug("github app recheck: not accessible", "repo", recheckRepo, "error", err)
@@ -4893,6 +4907,51 @@ func classifyGitHubAppWriteForbidden(ctx context.Context, appAuth *github.AppAut
 		Repo:            repo,
 	}
 	return d.Message(), github.AppStateWriteForbidden
+}
+
+// classifyGitHubAppRepoCoverage (#4360) asks the deterministic question the
+// other classifiers cannot: does this installation actually COVER the repos
+// this hive is configured to work on?
+//
+// Everything else here reasons from a failed call. diagnoseGitHubApp inspects
+// installation-level PERMISSIONS and never repo scope, and
+// classifyGitHubAppWriteForbidden infers scope from a 403 after a write has
+// already failed. Neither can see the case that prompted this: a hive pointed
+// at a second repo in the right org, on the right installation, simply not
+// ticked in the App's selected repos. GitHub answers 404 for that — the same
+// answer it gives for a repo that does not exist — so the read path reported
+// "app not installed / no read" and the dashboard blamed an undelivered
+// private key that had in fact arrived. The operator was sent to re-upload a
+// key, which could not possibly help.
+//
+// An error here is NOT a verdict. If the listing cannot be fetched the
+// credentials themselves are the more likely story, and the existing checks
+// tell it better; this returns raise=false and lets them run.
+func classifyGitHubAppRepoCoverage(ctx context.Context, appAuth *github.AppAuth, org string, repos []string, logger *slog.Logger) (raise bool, msg string, state github.AppAuthState) {
+	if appAuth == nil || len(repos) == 0 {
+		return false, "", github.AppStateUnknown
+	}
+
+	cov, err := appAuth.InstallationCoverage(ctx)
+	if err != nil {
+		logger.Debug("github app repo coverage: could not list installation repositories — deferring to the credential checks",
+			"org", org, "error", err)
+		return false, "", github.AppStateUnknown
+	}
+
+	missing := cov.Missing(org, repos)
+	if len(missing) == 0 {
+		return false, "", github.AppStateOK
+	}
+
+	d := github.AppAuthDiagnosis{
+		State:           github.AppStateRepoNotCovered,
+		ExpectedAccount: org,
+		InstallationID:  appAuth.InstallationID(),
+		APIURL:          appAuth.APIURL(),
+		Repos:           missing,
+	}
+	return true, d.Message(), github.AppStateRepoNotCovered
 }
 
 // primaryAdvisoryRepo returns the repo the advisory digest is posted to: the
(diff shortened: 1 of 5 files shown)


clubanderson:
Reviewed and verified.

- **Detection is the right query**: `GET /installation/repositories` with the installation token (the token IS the question), paginated at 100/page with a 10-page cap. **Rate-limit profile is sane without needing a cache**: the check runs only in the recheck path — the manual Re-check button and the 2-minute self-heal loop, which is already gated on the banner actually showing — so a healthy hive makes zero coverage calls, and an unhealthy one makes ≤10 every 2 minutes.
- **Refuses to accuse**: fetch error defers to the credential checks (never an empty set = accusing everything); truncated listing reports nothing missing; case-insensitive owner/name normalization; full_name-fallback reconstruction so a dropped entry can't read as "not covered". All pinned by tests, including the live kelly-headwaters shape.
- **Message distinctness verified** against the neighbouring classes: names exactly the repos, states what is NOT at fault (App/org/key), and gives the precise operator remedy — tick the repos under the installation's Repository access, with a deep settings link that is GHE-correct. Tests assert it never contains the old key-undelivered copy; it shares no text with the #4330 has_issues=false class, the #4332 foreign-authored-comment class, write-forbidden, or rate-limit copy. The specific check runs before the generic advisory read, so it wins.
- **No conflict with merged #4330/#4332**: those live in `advisory.go`; this adds `app_coverage.go` + state-machine entries and applies cleanly on current v4 (MERGEABLE).
- **Alert plumbing pinned**: `repo-not-covered` is user-actionable, not operator-actionable, and the new hub test proves `app-creds-undelivered` stays silent for it.
- Local verification at 49f0b6b: `go build ./...`, `go vet`, `go test ./pkg/github/` (**coverage 88.2% ≥ the 88 floor**), `./pkg/hub/` all green.

/lgtm
/approve

Issue #3096, opened 2026-08-09
[scanner] gh-wrapper --author gate can be bypassed by spoofing HIVE_AGENT env var


clubanderson:
## Finding

The `gh issue/pr list --author` gate in `bin/gh-wrapper.sh` (as proposed in PR #3089) still trusts mutable environment values (`HIVE_AGENT`, `HIVE_AGENT_DISPLAY_NAME`, and `HIVE_CONTRIBUTOR_USERNAME`) as identity sources. A caller can set these vars before invoking the wrapper and pass the check with an arbitrary `--author`.

This leaves the list restriction bypassable even after #3072's fix.

## Steps to Reproduce / Evidence

Using the PR branch code from `RawNuke/hive:fix/gh-wrapper-author-bypass-3072`:

1. Set `HIVE_AGENT=octocat` and run `gh-wrapper.sh issue list --author octocat` in non-contributor mode.
2. Exit code is `0` (allowed).
3. With `HIVE_AGENT=scanner` and `--author octocat`, exit code is `1` (blocked).

Observed behavior confirms gate decision is influenced by caller-controlled env, not only immutable bot identity from token (`gh api user`).

## Recommendation

Only allow `--author` when it matches a trusted identity derived from authenticated GitHub identity (e.g., `gh api user` result, optionally normalized for `[bot]` suffix). Do not treat mutable runtime env vars as authoritative for authorization decisions.

---
*Filed by scanner agent (ACMM L4 — issues-only mode)*

Issue #3072, opened 2026-08-09
[scanner] gh-wrapper `--author` bypass allows non-contributor agents to run blocked `gh issue|pr list`


clubanderson:
## Finding

`v2/gh-wrapper.sh` intends to block `gh issue list` and `gh pr list` for non-contributor agents, but the allow branch only checks whether the command string contains `--author`.

Current logic:

```bash
elif echo "$FULL_CMD" | grep -q "\-\-author"; then
  : # Allow any agent to list their own PRs for review
```

This does **not** validate that the requested author matches the current agent/bot identity, so any non-contributor agent can bypass the block by passing an arbitrary author value.

## Steps to Reproduce / Evidence

Code path in `v2/gh-wrapper.sh` around the list gate:

- identifies `issue|pr list`
- blocks non-contributor agents by default
- allows any command containing `--author`

Because only flag presence is checked, commands like:

```bash
gh issue list --repo kubestellar/hive --author some-other-user
```

pass the wrapper gate for non-contributor agents.

## Recommendation

1. Parse `--author` value and require it to match an approved self-identity set (current app bot login and/or assigned agent identity), not just flag presence.
2. Keep blocking when `--author` is missing, empty, or does not match self.
3. Add unit tests for the list gate covering allowed self-author and blocked foreign-author cases.

---
*Filed by scanner agent (ACMM L4 — issues-only mode)*

PR #4365, opened 2026-08-20
🧪 ci: run the gh-wrapper author-gate suite, and guard against the next orphan

### LINKED ISSUE #4363: [quality] bin/gh-wrapper.test.sh (author-gate regression suite) is not wired into CI
## Finding

v2-ci runs `bin/test_gh_wrapper_gates.sh` (the mode/allowlist/merge-gate harness), but `bin/gh-wrapper.test.sh` — the author-gate and contributor-mode regression suite for hive#3072/#3096, with 30+ executable cases (spoofed HIVE_AGENT, identity-lookup failure, marker trust boundary, cached-env spoof) — is not invoked by any workflow, Justfile target, or hook. It only runs when someone remembers to run it manually, so a regression in the author gate (a security boundary) would land silently.

`bin/README.md` documents the file as the wrapper's regression suite, which suggests the omission is accidental rather than intentional.

## Recommendation

Add a step next to the existing gate-tests step in `.github/workflows/v2-ci.yml`:

```yaml
      - name: gh-wrapper author-gate tests
        run: bash bin/gh-wrapper.test.sh
```

The suite is self-contained (creates its own mock gh + temp wrapper copy), needs no network, and runs in seconds. Alternatively, fold its cases into `test_gh_wrapper_gates.sh` so there is a single CI entry point.

## Priority
- Impact: medium (security-boundary regressions land untested)
- Effort: low (one workflow step)

---
---
🐝 **Hive Agent**: `quality` | **Instance:** `hosted-available-oke-11-placeholder-r05x` | **SHA:** `39482fd`

Danathar:
## What

Runs `bin/gh-wrapper.test.sh` in CI — and, because it turned out not to be the
only orphan, two more suites plus a guard that fails when a bin/ suite is not
run by anything.

## The reported finding, verified

`bin/gh-wrapper.test.sh` is the author-gate regression suite for #3072 and
#3096: 33 executable cases covering a spoofed `HIVE_AGENT`, an identity-lookup
failure, the marker trust boundary, and a cached-env spoof. It is referenced
nowhere outside its own header and `bin/README.md`. It ran only when someone
remembered, so a regression in a **security boundary** would have landed green.

It passes today — 33/33, seconds, no network — so wiring it in adds coverage
rather than a red build. That was worth checking before wiring: a suite nothing
has run for a while is exactly the kind that has quietly rotted.

## It was not alone

Checking the whole directory rather than just the reported file turned up two
more suites in the same state:

| Suite | Cases | Notes |
| --- | --- | --- |
| `gh-wrapper.test.sh` | 33 | the reported one |
| `test_hive_open_issue.sh` | 20 | |
| `test_hive_podman_preflight_ids.sh` | 61 | #4208 — **both** its sibling preflight suites were already wired |

That last row is what the accident looks like from the inside: two of three
siblings wired, the third silently not. All three now run in `v2-ci.yml`
alongside the suites they belong with, adding 114 assertions. All three pass
today; none needed a fix in order to be wired.

## The guard

The one-line step the issue asked for is the smaller half of this. Each
omission was invisible because **nothing fails when a test file is simply never
named** — the suite passes locally, the file looks maintained, and CI is green
precisely because it never ran.

`bin/test_bin_suites_wired.sh` fails when a suite in `bin/` is not named by any
workflow, Justfile target, or hook:

```
  FAIL: gh-wrapper.test.sh is not run by any workflow, Justfile target, or hook
        Add a step next to its siblings in .github/workflows/v2-ci.yml:
          - name: <what it guards>
            working-directory: .
            run: bash bin/gh-wrapper.test.sh
        If it genuinely should not run in CI, add it to
        UNWIRED_ON_PURPOSE in this file with the reason.
```

`UNWIRED_ON_PURPOSE` is there so "not in CI" stays a decision someone wrote
down rather than something that happened. It is empty today.

The guard is wired like any other suite — exempting itself would be the exact
hole it exists to close.

## Checked in both directions

A guard that cannot fail is the thing being fixed here, so:

- With the three orphans wired: **14 suites checked, 0 orphaned**, exit 0.
- Drop a throwaway unwired `bin/test_orphan_probe.sh` in: **exit 1**, naming
  that file. Remove it: exit 0 again.

## Does the workflow actually run?

Worth confirming rather than assuming, given #4339 — a workflow scoped to a
dead branch that reported green by never running. `v2-ci.yml` triggers on
`push`/`pull_request` for **both `v2` and `v4`**, with `bin/**` in its path
filter (added deliberately, per the comment there, because the gh wrapper is a
security boundary). So these steps genuinely execute.

## Also

`gh-wrapper.test.sh` was mode `644` while every sibling suite is `755`; made
executable to match. `bin/README.md` gains a row for the new guard.

Closes #4363.


clubanderson:
Reviewed and verified.

- **Self-trigger proof**: all four new steps executed on this PR's own v2-ci run (32427215779) and passed — `gh-wrapper author-gate`, `hive-open-issue`, `preflight_ids (#4208)`, and the wired-suites guard itself. `v2-ci.yml` triggers on both `v2`/`v4` with `bin/**` in its path filter, so this is real per-PR coverage.
- **Suites verified locally at 24e2940**: 33/33 (author gate), 20/20 (open-issue), 61/61 (preflight IDs) — all pass before wiring, so this adds coverage without a red build. 114 assertions total, seconds, no network.
- **Guard verified in both directions**: current tree = 14 suites checked, 0 orphaned, exit 0; a planted unwired `bin/test_orphan_probe.sh` = flagged by name, exit 1, with an actionable fix template. The guard checks itself (no self-exemption hole).
- **No false positives on intentionally-manual suites**: the SELinux qualification (#4353) lives at `src/deploy/qualify_podman_selinux.sh` — outside `bin/`, so structurally out of the guard's scope — and the `UNWIRED_ON_PURPOSE` filename|reason mechanism exists (empty today, correctly) for any future deliberately-manual bin/ suite, keeping "not in CI" a written-down decision.
- **Hygiene**: guard scans workflows + Justfile + githooks by basename (invocation-agnostic); `gh-wrapper.test.sh` mode fixed 644→755 to match siblings; guard script shellcheck-clean; actionlint findings on `v2-ci.yml` byte-identical to base (pre-existing); no permission changes (workflow perms untouched); no forbidden podman lifecycle literals.

/lgtm
/approve

PR #4369, opened 2026-08-20
🐛 podman quadlet doc: install a config that exists, on the port HealthCmd probes

### LINKED ISSUE #4367: 🐛 podman quadlet doc: install step copies a file that does not exist, and the obvious substitute hangs the start for 300s
## Summary

`src/docs/podman-standalone-quadlet.md` (#4354, shipped in #4355) tells the
operator to install the config with:

```sh
cp src/hive.yaml "$CONF/hive.yaml"      # then edit it
```

**`src/hive.yaml` does not exist.** The repo has two candidates, and they differ
on exactly the port `hive.container`'s `HealthCmd` probes:

| File | `dashboard.port` |
| --- | --- |
| `src/hive.yaml.example` | **3001** |
| `src/deploy/hive.yaml` | **3002** |

The unit probes `http://127.0.0.1:3002/api/health`, which agrees with
`src/deploy/hive.yaml` and with the `hive` service's healthcheck in
`src/docker-compose.yaml`. Picking the *other* one silently breaks the start.

`src/hive.yaml.example` is the substitute an operator is most likely to reach
for — it is the file the rest of the docs point at, and
`src/docs/network-requirements.md:18` describes it as the one "for local/source
runs", which is what a standalone Podman install reads as.

## Why this is worse than a typo

The failure mode is bad **because `Notify=healthy` works correctly**. With the
wrong config:

- the API binds 3001, so `curl -sf http://127.0.0.1:3002/api/health` never
  succeeds;
- `Notify=healthy` therefore holds the unit in `activating` — correct behaviour,
  a service that is not serving does not get to look started;
- `TimeoutStartSec=300` expires **five minutes later**;
- the generated `ExecStart` carries `--rm`, so the container that holds the
  evidence is deleted on timeout — the trap `hive.container`'s own `[Service]`
  comment already warns about.

The operator is left with `Job for hive.service failed because a timeout was
exceeded`, no container to inspect, and a five-minute feedback loop per attempt,
for a one-word config mistake the document told them to make.

## Reproduction

Measured on Fedora 44 (Bluefin 44.20260811), SELinux enforcing, Podman 5.8.4
rootless, cgroup v2, against `ghcr.io/kubestellar/hive:stable`
(`aa9ad9830514`), following the document's install steps literally.

With `src/hive.yaml.example` as `$CONF/hive.yaml`:

```
$ systemctl --user start hive.service
Job for hive.service failed because a timeout was exceeded.
EXIT=1  ELAPSED=300s
ActiveState=activating  SubState=auto-restart  Result=timeout
$ podman ps -a          # empty: --rm deleted it
```

Probed from inside the container, which is what identifies the cause:

```
127.0.0.1:3001/api/health -> 200
127.0.0.1:3002/api/health -> 000   (connection refused)
127.0.0.1:7681/api/health -> 401
```

With `src/deploy/hive.yaml` instead, everything else identical:

```
$ systemctl --user start hive.service
EXIT=0  ELAPSED=11s
ActiveState=active  SubState=running  Result=success
$ podman ps
hive   Up 11 seconds (healthy)
```

## The unit itself is fine

This is a documentation defect only. #4355's unit and its claims verified live
on the same host, once the config was right:

| Claim | Observed |
| --- | --- |
| both services generated | `hive.service`, `hive-data-volume.service` |
| `Type=notify`, `NotifyAccess=all`, `TimeoutStartUSec=5min` | all three |
| `Requires=`/`After=hive-data-volume.service` | both |
| start returns 0, `ActiveState=active` | exit 0, `active`, `Result=success` |
| container healthy | `Up 11 seconds (healthy)` |
| `AddCapability=NET_ADMIN` | `ambient CAP_NET_ADMIN granted`, proxy on `127.0.0.1:18443` |

(11s rather than #4355's 52s because the image was pre-pulled.)

## Suggested fix

1. Point the install step at a file that exists — `src/deploy/hive.yaml`, which
   is the config whose `dashboard.port` the unit's `HealthCmd` matches.
2. Say **why** it matters, in one line: `dashboard.port` must agree with the
   port in `HealthCmd=`, or `Notify=healthy` turns a config mismatch into a
   silent 300-second hang with no container left to inspect.
3. Add it to the doc's existing "Traps, measured not guessed" section — it
   belongs beside the `EnvironmentFile=` dash trap and the `GlobalArgs=`
   healthcheck trap, both of which are the same shape: generates cleanly, fails
   at runtime, no dry-run can catch it.
4. Optionally make the unit's comment near `HealthCmd=` name the config key it
   has to agree with, so the coupling is visible from the unit as well as the
   doc.

Whatever the shape, the two must not be able to drift silently. If
`src/hive.yaml.example` is intended to stay on 3001 for source runs, the doc
should say plainly that it is **not** the file to install for a Quadlet
deployment.

## Acceptance criteria

- The install step names a file that exists in the repo.
- Following the document literally, start-to-finish, produces a service that
  reaches `active` — no editing required beyond the token step already
  documented.
- The `dashboard.port` / `HealthCmd` coupling is stated where an operator will
  meet it, not left to be inferred.
- No change to the unit's readiness design: `Notify=healthy` behaved correctly
  in both runs and is not what needs fixing.

## References

- #4354 — the issue the doc shipped under
- #4355 — the PR that added it
- `src/docs/podman-standalone-quadlet.md:123` — the broken line
- `src/docs/network-requirements.md:18` — already records the 3001/3002
  divergence, which is why this is a documentation gap rather than a surprise
- `src/docker-compose.yaml` — the `hive` service healthchecks 3002, agreeing
  with the unit

Danathar:
Fixes #4367.

## The bug

`src/docs/podman-standalone-quadlet.md` told the operator to install the config with:

```sh
cp src/hive.yaml "$CONF/hive.yaml"      # then edit it
```

`src/.gitignore:3` excludes `hive.yaml`. It is the file an operator *creates*, not one the repo ships, so the step fails outright in a fresh checkout.

The obvious substitute is worse than the missing file. `src/hive.yaml.example` is what the rest of the docs point at and what `src/docs/network-requirements.md:18` calls the config "for local/source runs" — and it sets `dashboard.port: 3001`, while the unit probes `http://127.0.0.1:3002/api/health`. Install it unchanged and Hive comes up serving, healthily, on the wrong port:

| | |
| --- | --- |
| the probe | never answers |
| `Notify=healthy` | correctly holds the unit in `activating` — a service that is not serving does not get to look started |
| `TimeoutStartSec=300` | expires five minutes later |
| generated `--rm` | deletes the container that held the evidence |

Leaving `Job for hive.service failed because a timeout was exceeded`, an empty `podman ps -a`, and a five-minute feedback loop per attempt — for a one-digit config mistake the document told them to make.

## What changed

Documentation and comments. **No change to the unit's readiness design** — #4367 measured `Notify=healthy` behaving correctly in both directions, and that is not what needed fixing.

- **The install step now names a file the repo ships.** It copies `src/hive.yaml.example`, corrects the port in the same block, and prints the check that proves it took.
- **The `dashboard.port` / `HealthCmd=` coupling is stated where the operator meets it**, and added to *Traps, measured not guessed* beside the `EnvironmentFile=` dash trap and the `GlobalArgs=` healthcheck trap — the same shape as both: generates cleanly, fails at runtime, no dry-run can catch it.
- **The unit's `HealthCmd` comment names the config key it is coupled to**, so the coupling is visible from `hive.container` as well as from the page (suggested fix item 4 in the issue).
- **`network-requirements.md`**, which already recorded the 3001/3002 divergence, now says what the container deployments do with it.

3002 is not an arbitrary pick: it is the Go default (`defaultDashboardPort` in `src/pkg/config/config.go:3817`), what `src/deploy/hive.yaml` sets, and what the `hive` service's healthcheck in `src/docker-compose.yaml` probes. Deleting the `port:` line entirely works too, for the first reason.

## And a gate, because "the two must not be able to drift silently"

A coupling a reviewer has to remember is one that drifts. `src/deploy/test_quadlet_config_contract.sh` (new, wired into `v2-ci.yml` next to the other deployment-contract guards) asserts:

1. `HealthCmd=` probes some `127.0.0.1:<port>/api/health` at all;
2. that port equals `defaultDashboardPort` — so an operator who *omits* `dashboard.port` still reaches healthy;
3. it equals `src/deploy/hive.yaml`'s `dashboard.port`;
4. it equals the **`hive`** service's Compose healthcheck (scoped to that service on purpose — `gateway` probes `/api/health` too, on its own port);
5. the unit still names `dashboard.port` where a reader will meet it;
6. every file the page tells an operator to copy is one **git tracks** — not merely one present on disk. That distinction is the bug: a contributor who followed the README has an untracked `src/hive.yaml` in their checkout, so a plain existence test passes for them and fails for the operator who cloned and did nothing else, which is exactly the reader that step is for;
7. the page never installs `hive.yaml.example` without correcting the port.

The Quadlet generator gate cannot cover any of this — a mismatched port generates perfectly.

## Verified

- The new contract test passes, and **fails on each regression injected one at a time**: unit port drifted to 3001 (3 assertions fire), the doc naming an untracked `src/hive.yaml` (the original defect), the doc installing the example without the correction, and the unit comment dropping the key.
- `src/deploy/test_quadlet_generator_gate.sh` — 0 failures, both rootful and `--user`, against quadlet 5.8.4.
- `src/deploy/test_standalone_image_refs.sh` — 17 assertions pass.
- `shellcheck -x` clean on the new script; it runs from `src/` as CI invokes it.
- The documented `cp` / `sed` / `grep` sequence was run against the real `src/hive.yaml.example`: `  port: 3001` occurs exactly once in that file, and the result is `dashboard: / port: 3002` with no 3001 left.

## Acceptance criteria

| From #4367 | |
| --- | --- |
| install step names a file that exists in the repo | `src/hive.yaml.example`, and gated (assertion 6) |
| following the document literally produces a service that reaches `active` | the port correction is a copy-pasteable step in the same block, with its own verification, and gated (assertion 7) |
| the coupling is stated where an operator meets it | install section, Traps, the unit comment, and `network-requirements.md` |
| no change to the unit's readiness design | none — only comments changed in `hive.container` |

Part of #4188; this does not close the parent.

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -94,6 +94,20 @@ jobs:
       - name: Standalone service deployment contract (#4204)
         run: bash deploy/test_standalone_service_contract.sh
 
+      # #4367: the Quadlet unit's HealthCmd= probes a port that lives in the
+      # OPERATOR's hive.yaml, and nothing at install time notices when the two
+      # disagree — the config parses, the generator exits 0, the container
+      # starts and is genuinely healthy on the other port, and Notify=healthy
+      # correctly holds the unit in `activating` until TimeoutStartSec=300
+      # expires and --rm deletes the evidence. Five minutes and no artifact for
+      # one wrong digit. Same shape as the guards above: invisible when it
+      # breaks, and no dry-run can catch it. Also asserts the install steps name
+      # files the repo actually TRACKS, which is the bug itself: the doc said
+      # `cp src/hive.yaml`, a gitignored path that exists only in a checkout
+      # where someone already followed the README.
+      - name: Quadlet config-coupling contract (#4367)
+        run: bash deploy/test_quadlet_config_contract.sh
+
       # bin/gh-wrapper.sh is the merge gate standing behind the proxy hard-deny:
       # if its argument parsing is wrong, a prompt-injected agent merges
       # unvetted code. It had NO tests and matched no CI path filter. This
--- a/src/deploy/quadlet/hive.container
+++ b/src/deploy/quadlet/hive.container
@@ -90,6 +90,15 @@ EnvironmentFile=%E/hive/hive.env
 # READINESS. HealthCmd is required for Notify=healthy to mean anything; see the
 # header. /api/health is the same endpoint the Kubernetes readiness probe and
 # the Compose healthcheck use, so "healthy" means the same thing on all three.
+#
+# THE PORT BELOW IS COUPLED TO `dashboard.port` IN THE MOUNTED hive.yaml, and
+# nothing enforces the coupling. 3002 is the Go default (defaultDashboardPort in
+# pkg/config/config.go), what src/deploy/hive.yaml sets, and what the Compose
+# healthcheck probes — but src/hive.yaml.example ships 3001 for source runs.
+# Install that file unchanged and Hive serves on 3001, this probe never answers,
+# and Notify=healthy correctly holds the unit in `activating` for the whole
+# TimeoutStartSec below before --rm deletes the evidence. Change one, change the
+# other (#4367).
 Notify=healthy
 HealthCmd=curl -sf http://127.0.0.1:3002/api/health
 HealthInterval=10s
--- a/src/deploy/test_quadlet_config_contract.sh
+++ b/src/deploy/test_quadlet_config_contract.sh
@@ -1,0 +1,211 @@
+#!/usr/bin/env bash
+# Contract test for the Quadlet unit's config coupling (#4367).
+# Run: bash src/deploy/test_quadlet_config_contract.sh
+#
+# The unit's HealthCmd= probes a port that lives in the operator's hive.yaml,
+# and NOTHING at install time notices when the two disagree. The config parses,
+# the Quadlet generator exits 0, the container starts and is genuinely healthy —
+# on the other port. Notify=healthy then does exactly what it should and holds
+# the unit in `activating` until TimeoutStartSec=300 expires, at which point the
+# generated `--rm` deletes the container that held the evidence. Five minutes,
+# no artifact, one wrong digit.
+#
+# That is the shape the repo already gates rather than reviews: invisible when
+# it breaks, and no dry-run can catch it. So this asserts the coupling directly.
+# Pure string analysis: nothing is started, no registry is contacted.
+
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+UNIT="src/deploy/quadlet/hive.container"
+DOC="src/docs/podman-standalone-quadlet.md"
+COMPOSE="src/docker-compose.yaml"
+DEPLOY_CONF="src/deploy/hive.yaml"
+EXAMPLE_CONF="src/hive.yaml.example"
+GO_CONFIG="src/pkg/config/config.go"
+
+pass_count=0
+failures=0
+
+fail() {
+  printf 'FAIL: %s\n' "$1" >&2
+  failures=$((failures + 1))
+}
+
+ok() {
+  pass_count=$((pass_count + 1))
+}
+
+need_file() {
+  if [[ ! -f "${ROOT}/$1" ]]; then
+    fail "$1: expected file is missing"
+    return 1
+  fi
+  return 0
+}
+
+# --- 1. The port the unit probes -------------------------------------------
+
+health_port=""
+if need_file "$UNIT"; then
+  health_port="$(
+    grep -E '^HealthCmd=' "${ROOT}/${UNIT}" \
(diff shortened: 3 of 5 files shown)


clubanderson:
Verified every claim against v4: src/.gitignore does exclude hive.yaml (the original defect is real), hive.yaml.example ships dashboard.port 3001 with '  port: 3001' occurring exactly once so the documented sed is safe, defaultDashboardPort=3002 (config.go:3817), src/deploy/hive.yaml sets 3002, and the Compose healthchecks map hive→3002 / gateway→3001 — the contract test's service scoping matches reality. Ran src/deploy/test_quadlet_config_contract.sh at the PR head: 7/7 assertions pass; injected the 3001 drift into hive.container and exactly the 3 promised assertions fired. The documented cp/sed/grep sequence produces dashboard: / port: 3002 with no 3001 left. Tracked-not-present distinction in assertion 6 is the right call. CI fully green.

/lgtm
/approve

Issue #2459, opened 2026-08-03
spoke: dashboard Set ID dead-ends on hives with hub-delivered per-app-id keys (reinit gated on raw key_file, Re-check stuck on nil client)


clubanderson:
## Summary

On a hosted spoke whose GitHub App private key was hub-delivered to the per-app-id path (`/data/gh-app-key-<app_id>.pem`, with `github.key_file` deliberately left empty), the dashboard's "GitHub App Not Installed" banner **Set ID** button can never succeed: the installation ID is saved but the GitHub client is never (re)built, the banner never clears, and **Re-check** fails forever with "no GitHub client". To the owner this reads as "the spoke does not accept my installation ID".

Observed live on `hosted-available-vllmd-06` (vllm-d cluster, github.ibm.com, App 5686): owner entered installation ID 43021, spoke kept reporting `installation_id: 0`, state `not-installed`. Repaired operationally by arming the full identity via `POST /api/saas/hives/{id}/forge` (the heartbeat delivery path uses the correct key resolution).

## Root cause

`v2/pkg/dashboard/api.go`, `handleConfigGitHub` (the target of the banner's Set ID → `PUT /api/config/github`):

```go
if cfg.GitHub.HasUsableApp() && cfg.GitHub.KeyFile != "" {
    if s.deps.ReinitGitHubFunc != nil {
        if err := s.deps.ReinitGitHubFunc(cfg.GitHub.AppID, cfg.GitHub.InstallationID, cfg.GitHub.KeyFile); err != nil {
```

The reinit is gated on the RAW `cfg.GitHub.KeyFile`. But since the per-app-id key fix, `key_file` is **deliberately never persisted** on hub-delivered keys — both the boot path (`cmd/hive/main.go` `initGitHubAuth`, line ~540) and the heartbeat apply path (line ~3053) resolve the key via:

```go
resolveAppKeyFile(cfg.GitHub.KeyFile, os.Getenv("GH_APP_KEY_FILE"), cfg.GitHub.AppID)
```

which prefers `/data/gh-app-key-<app_id>.pem`. The dashboard handler was never updated to match, so on exactly the fleet-standard configuration (empty `key_file`, key present under the per-app-id path) it:

1. saves `installation_id`,
2. **skips** `ReinitGitHubFunc` entirely (no `reinit` field in the response — UI shows "Installation ID saved. Click Re-check to verify"),
[...]

PR #4370, opened 2026-08-21
🐛 provision: stop pinning github.key_file at a file the manifest never creates

### LINKED ISSUE #4368: 🐛 provisioning sets github.key_file to /secrets/gh-app-key.pem but never creates it — 4 hives cannot load their App key
## Symptom

Four hosted hives on hive-oke have `github.key_file: /secrets/gh-app-key.pem` in their **dashboard overlay** (the authoritative config layer), but **that file does not exist on those hives**. The App key is present under a different name, and is never tried — because an explicit `key_file` wins over the documented fallback chain.

Result: `github_auth: fail`, agents cannot act as the App.

## Blast radius — the whole 2026-08-12 provisioning batch

| hive | created | `key_file` | file present |
|---|---|---|---|
| `hosted-available-oke-260812-0nzc-available-ok-ba44` | 2026-08-12 | `/secrets/gh-app-key.pem` | ❌ MISSING |
| `hosted-available-oke-260812-8yiz-available-ok-go05` | 2026-08-12 | `/secrets/gh-app-key.pem` | ❌ MISSING |
| `hosted-available-oke-260812-ixya-available-ok-3do3` | 2026-08-12 | `/secrets/gh-app-key.pem` | ❌ MISSING |
| `hosted-available-oke-260812-jlem-available-ok-nnpj` | 2026-08-12 | `/secrets/gh-app-key.pem` | ❌ MISSING |
| `oke-01`, `oke-03` | 2026-07-23 | `/secrets/gh-app-key.pem` | ✅ present (symlink into the projected secret) |
| `oke-02/04/05/06/07/08/11/12/15`, aslom, console, bluefin, daviddiaz | ≤ 2026-07-29 | `/data/gh-app-key.pem` | ✅ present |

**Every hive created 2026-08-12 is broken; every hive created earlier is fine.** The correlation is with `created_at`, not with claim/assign state — all four are `status: assigned`, `claim_delivered: true`, exactly like the working ones. So this is **provisioning**, not assignment.

## What the affected hives actually have

```
/data/gh-app-key-3568013.pem     # matches the configured app_id — the real key
/data/gh-app-key-5686.pem        # GHE key
/secrets/                        # projected-secret dir exists, but has NO gh-app-key.pem entry
```

There is no generic `/data/gh-app-key.pem` on these hives either — only app-id-suffixed files. So the provisioner wrote a `key_file` value that matches **neither** what it placed on the PVC **nor** what it mounted at `/secrets`.

## Why the fallback chain does not save it

Documented order is `$GH_APP_KEY_FILE` → `github.key_file` → `/data/gh-app-key.pem` → `/secrets/gh-app-key.pem`. Because `github.key_file` is set **explicitly**, it wins, and the app-id-suffixed key that is actually present is never reached.

## Suggested fix

In the provisioning template (`src/pkg/hub/saas_provision.go`), either:
1. **Do not set `github.key_file` at all** when the key is delivered to the PVC — let the fallback chain find `/data/gh-app-key.pem`; or
2. Set it to the path actually written (`/data/gh-app-key-<app_id>.pem`); or
3. Keep `/secrets/gh-app-key.pem` but **actually create it** (as the pre-08-12 hives do, via a symlink into the projected secret).

Whichever is chosen, a post-provision assertion that the resolved `key_file` is readable would have caught this at creation time instead of at first App call.

## Related

The affected hives were only noticed because one of them (`8yiz`, Kelly's headwaters) surfaced as `app-creds-undelivered` in the hub. That alert text — *"private key has not reached this spoke… re-upload via PUT"* — is misleading here: the key **did** reach the spoke, it is just referenced by the wrong path. See #4360 for the related surfacing problem.

Note the two failures are independent: `8yiz` **also** has its repo missing from the App installation, so fixing `key_file` alone does not make it functional — but it would fail auth even after that is corrected.

## Workaround applied to `8yiz`

Repointed `key_file` to `/data/gh-app-key-3568013.pem`. This must be done **offline** — the running hive re-serializes its in-memory config back over `/data/hive.yaml.dashboard`, silently reverting an in-place edit (and a `rollout restart` does not help). Procedure: scale to 0, mount the PVC in a temp pod, edit both `hive.yaml.dashboard` and `hive.yaml.runtime`, scale back to 1, then verify via `/etc/hive/hive.yaml` in the running pod.

The other three remain unfixed pending a decision on the provisioning-side fix.

Danathar:
Fixes #4368.

## The bug is a coupling written twice

The spoke ConfigMap seeds `key_file` gated on **`UseApp`**:

```yaml
{{- if .UseApp}}
      app_id: {{.AppID}}
      installation_id: {{.InstallationID}}
      key_file: /secrets/gh-app-key.pem      # ← was here
```

The `hive-secrets` Secret creates that entry gated on **`UseAppFull`**:

```go
useApp     := req.AuthMethod == "app" && req.AppID != ""
useAppFull := useApp && req.InstallationID != "" && req.AppPrivateKey != ""
```

So a hive provisioned with an App id but **without the key inline** — the normal case, since the key usually arrives later through the heartbeat delivery lane — got a config naming a file that would never exist. Two conditions on one fact, in two places, free to disagree. They did, for the whole 2026-08-12 batch.

It then failed in the worst available way. An explicit `key_file` wins outright in `resolveAppKeyFile` — deliberately, because a path an operator typed must not be silently redirected — so the key that **did** arrive, at `/data/gh-app-key-<app_id>.pem`, was never even tried. Four hives reporting `github_auth: fail` with the correct key sitting on the PVC, and the hub's alert blaming an undelivered key.

## The fix: don't pin it (option 1 in the issue)

`key_file` is derivable from `app_id`. That is not my opinion — it is the position `cmd/hive/main.go` already takes, at the point where it refuses to persist the path after a delivery:

> `Deliberately NOT cfg.GitHub.KeyFile = keyPath.` … *key_file is DERIVABLE from app_id … Persisting the path turns a derived value into a stored one that outlives the App it was derived for.*

Provisioning was the last writer still pinning it.

**Nothing regresses for a hive provisioned *with* an inline key.** `resolveAppKeyFile`'s final fallback is `/secrets/gh-app-key.pem` — literally the path that was being pinned — so those hives resolve to exactly the same file. The only behaviour that changes is the case that was broken, where the chain now reaches the delivered key instead of stopping at a path that does not exist.

## Making it unrepresentable

`provisionHive` now renders the manifest to a buffer and runs `assertSpokeManifestKeyFile` **before** `kubectl apply`. A future pin is fine — but only if the manifest also creates what it names; otherwise provisioning fails loudly.

That is deliberate about which failure is preferable: a hive that never provisions is a visible error an admin retries, whereas one that provisions broken is a `github_auth: fail` nobody looks at until an agent needs the App. This is the issue's suggested post-provision assertion, moved to the one moment when both halves of the coupling are in the same place. The check is narrow on purpose — only `/secrets` paths are verifiable at render time; `/data` is an empty PVC until first delivery, and anything else is an operator's own location.

## The detector that should have caught this

Both dangling-`key_file` checks looked only at `/data/`:

| | before | after |
|---|---|---|
| `src/deploy/entrypoint.sh` boot warning | `key_file.startswith('/data/')` | `('/data/', '/secrets/')` |
| `config.DanglingKeyFile` | `const pvcPrefix = "/data/"` | both delivery locations |

The four hives named a path under `/secrets/` — the one prefix neither examined. The check written for precisely this symptom returned false for every one of them, and nothing surfaced until the first App call. A path outside both prefixes is still left alone: an operator may keep a key somewhere this build has never heard of.

## One more raw reader

The config-reload rebuild in `main.go` gated on `cfg.GitHub.KeyFile != ""` and passed the raw value to `NewAppAuth`. An unset `key_file` is the correct steady state on a hosted spoke, so that path skipped the rebuild on exactly the hives that need it — a corrected `installation_id` saved to `hive.yaml` kept minting tokens for the old installation until the pod restarted.

Startup, the heartbeat rebuild, and the dashboard's Set ID handler all already resolve; this was the last one that did not. The identical bug with the identical cause was fixed in the dashboard handler under #2459, whose comment describes this exact scenario. Not strictly required by the template change — but it is the path that an unset `key_file` now travels on every newly provisioned hive, so shipping the template change without it would have widened a live latent fault.

## Verified

- **New Go tests render the real template.** No pin in any of the three provisioning shapes (app+inline key, app without inline key, token auth); the guard accepts all three; and the load-bearing one — injecting the historical `key_file: /secrets/gh-app-key.pem` line into a manifest that creates no such Secret entry is **rejected**, while the same pin against a manifest that *does* create it is accepted. Without that last case the assertion could be vacuous and nobody would know.
- **`src/deploy/test_entrypoint_dangling_keyfile.sh`** extracts the real python merge block from the shipped `entrypoint.sh` and executes it against fixtures, following the convention `test_entrypoint_runtime_config.sh` set. It **fails against the unpatched entrypoint** on exactly the `/secrets` case (`4 passed, 1 failed`) and passes after — a regression test, not a description. Wired into `v2-ci.yml`.
- `go build ./...`, `go vet ./...`, `gofmt`, `shellcheck -x` all clean.
- `./pkg/hub` (92s, full) and `./cmd/hive` pass in full. `pkg/config`'s `TestGateFailsClosedWithoutIP6Tables` fails identically on unmodified `upstream/v4` in this environment — it needs root/NET_ADMIN — so it is pre-existing and unrelated; I checked rather than assumed.

## Not fixed here

The four live hives still need their configs repaired offline; this stops the next batch being created broken. `8yiz` additionally has its repo missing from the App installation, which #4368 records as an independent fault.

Part of the hosted-provisioning correctness work; does not close any parent.

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -53,6 +53,18 @@ jobs:
       - name: Entrypoint NET_ADMIN ambient-cap tests (#3760)
         run: bash deploy/test_entrypoint_netadmin_ambient.sh
 
+      # #4368: the entrypoint warns when the overlay's github.key_file names a
+      # PEM that is not on disk — and used to look only at /data/, while the
+      # provisioning template seeded /secrets/gh-app-key.pem for every App-using
+      # hive and created that Secret entry only for hives provisioned WITH an
+      # inline key. Four hives shipped naming a file that would never exist,
+      # under the one prefix the check skipped, so the detector written for
+      # exactly that symptom stayed silent and the fault first surfaced as
+      # github_auth: fail. EXECUTES the real merge block extracted from the
+      # shipped entrypoint, so a regression fails the PR rather than a hive.
+      - name: Entrypoint dangling key_file tests (#4368)
+        run: bash deploy/test_entrypoint_dangling_keyfile.sh
+
       # Supply-chain pins regress SILENTLY. The PI_VERSION pin (#3443) landed on
       # v2, and a sync/v2-into-v4-* merge resolved the conflict in favour of the
       # older side and restored `ARG PI_VERSION=latest` — the fix commit is an
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -2746,12 +2746,29 @@ func main() {
 		// app_id/installation_id at construction, so without this a corrected
 		// installation_id in hive.yaml keeps minting tokens for the OLD
 		// installation until the pod restarts.
+		//
+		// RESOLVE the key file rather than reading cfg.GitHub.KeyFile raw. An
+		// unset key_file is the CORRECT steady state on a hosted spoke — the
+		// heartbeat apply path deliberately does not persist one, because the
+		// path is derivable from app_id and a stored value outlives the App it
+		// was derived for. Gating on the raw field therefore skipped the rebuild
+		// entirely on exactly the hives that need it: a corrected
+		// installation_id saved to hive.yaml kept minting tokens for the old
+		// installation until the pod restarted. Startup (resolveAppKeyFile
+		// above), the heartbeat rebuild, and the dashboard's Set ID handler
+		// (#2459) all already resolve here; this was the last raw reader.
+		//
+		// Comparing RESOLVED paths also catches a change the raw comparison
+		// cannot see: a per-app-id key arriving on the PVC changes which key
+		// this process should sign with while cfg.GitHub.KeyFile stays "".
+		prevKeyFile := resolveAppKeyFile(prevGitHub.KeyFile, os.Getenv("GH_APP_KEY_FILE"), prevGitHub.AppID)
+		nextKeyFile := resolveAppKeyFile(cfg.GitHub.KeyFile, os.Getenv("GH_APP_KEY_FILE"), cfg.GitHub.AppID)
 		if prevGitHub.AppID != cfg.GitHub.AppID ||
 			prevGitHub.InstallationID != cfg.GitHub.InstallationID ||
-			prevGitHub.KeyFile != cfg.GitHub.KeyFile ||
+			prevKeyFile != nextKeyFile ||
 			prevGitHub.APIURL != cfg.GitHub.APIURL {
-			if cfg.GitHub.HasUsableApp() && cfg.GitHub.KeyFile != "" {
-				newAppAuth, appErr := github.NewAppAuth(cfg.GitHub.AppID, cfg.GitHub.InstallationID, cfg.GitHub.KeyFile, logger, cfg.GitHub.ResolvedAPIURL())
+			if cfg.GitHub.HasUsableApp() && nextKeyFile != "" {
+				newAppAuth, appErr := github.NewAppAuth(cfg.GitHub.AppID, cfg.GitHub.InstallationID, nextKeyFile, logger, cfg.GitHub.ResolvedAPIURL())
 				if appErr != nil {
 					logger.Error("github app auth rebuild after config reload failed", "error", appErr)
 				} else {
@@ -2775,6 +2792,7 @@ func main() {
 					logger.Info("github app auth rebuilt after config reload",
 						"app_id", cfg.GitHub.AppID,
 						"installation_id", cfg.GitHub.InstallationID,
+						"key_file", nextKeyFile,
 					)
 				}
 			}
--- a/src/deploy/entrypoint.sh
+++ b/src/deploy/entrypoint.sh
@@ -203,11 +203,25 @@ elif overlay_has_real_app:
     # PEM it names actually exists on this PVC. A dangling reference
     # (e.g. an interrupted key write) must not silently authenticate with
     # a missing/placeholder file; fail loud instead of 401-looping quietly.
+    #
+    # BOTH delivery locations, not just the PVC (#4368). /secrets is the
+    # provisioning mount, and the provisioning template used to seed
+    # key_file=/secrets/gh-app-key.pem for every App-using hive while creating
+    # that Secret entry only for hives provisioned WITH an inline private key.
+    # Four hives shipped naming a file that would never exist — under the one
+    # prefix this check did not look at, so the warning written for exactly
+    # this symptom stayed silent on every one of them, and the fault first
+    # surfaced as github_auth: fail hours later. A path outside both prefixes
+    # is an operator's own location and is deliberately not second-guessed.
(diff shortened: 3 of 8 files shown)


clubanderson:
Reviewed on `b148f463`. Verified every claim against the tree rather than the prose:

- **The coupling is real and written exactly as described**: seed pin gated on `UseApp` (saas_provision.go template), Secret entry gated on `useAppFull` (lines 2003–2004), so app-id-without-inline-key shipped a `key_file` naming a file the same manifest never creates — and `resolveAppKeyFile` gives an explicit value outright priority, so the delivered `/data/gh-app-key-<app_id>.pem` was never tried.
- **Un-pinning is safe for the healthy shape**: confirmed `resolveAppKeyFile`'s final fallback is `spokeProvisionedAppKeyPath = /secrets/gh-app-key.pem` — a hive provisioned *with* an inline key resolves to exactly the file it used to be pinned to. And main.go's heartbeat-apply path already refuses to persist a derived path, so provisioning really was the last writer.
- **The render-then-assert guard is not vacuous**: `TestProvisionManifestKeyFileAssertionCatchesTheBug` re-injects the historical pin into a `UseApp`-without-key render and the guard rejects it, while the same pin against a `UseAppFull` render is accepted. `manifestSecretKeys` scoping to the `hive-secrets` document and the two-space entry-vs-PEM-body discrimination both check out against the real template. Narrowing to `/secrets` is right — `/data` is unverifiable at render time.
- **Detector widening**: both `DanglingKeyFile` and the entrypoint warning now cover both delivery prefixes, operator paths still left alone, and the short-path/no-panic edge is pinned. I ran the new shell suite against the **unpatched** v4 entrypoint: 4 passed, 1 failed on exactly the /secrets case — a genuine regression test, as advertised. Patched: 5/5.
- **The reload raw-reader fix** mirrors the #2459 dashboard fix; `prevGitHub` is captured before the reload, and comparing resolved paths also catches a per-app-id key arriving while `key_file` stays empty.
- **Docs stay consistent**: `docs/manual-provisioning.md` B.4 pins the path *and* creates the Secret entry (placeholder PEM at line ~1131) — the guard's accepted shape; `github-app-setup.md` has the operator mount the file it names. No CA-key-path interaction (#4341's `/data/.hive` relocation untouched).
- `go build`/`go vet` clean, gofmt clean on all touched files, shellcheck clean, new Go tests pass (`pkg/config` Dangling/HasPrefix, all four `pkg/hub` suites), no cleanup-contract literals anywhere in the diff. CI fully green across all lanes.

/lgtm
/approve

Issue #4206, opened 2026-08-20
🧩 Podman: define one standalone image-reference source of truth


Danathar:
## Parent

Part of #4188. This issue must not close the parent.

## Why

New Podman assets must not copy the retired `v2-latest` reference or create a second image/channel configuration that can drift from Docker.

## 30-minute boundary

Define or enforce one shared image-reference input. Do not redesign release promotion or update mechanisms.

## Deliverable

Inventory standalone image references and introduce the smallest shared source of truth or contract check that future Docker and Podman assets can consume.

## Acceptance criteria

- The retired `v2-latest` reference is not propagated into new Podman work.
- Docker behavior is either intentionally preserved or changed in a separately justified, tested edit.
- Fully qualified image names and optional digest pinning remain possible.
- A test detects accidental Docker/Podman image-reference drift once both assets exist.

PR #4380, opened 2026-08-21
🧩 podman: add the gateway and network boundary units

### LINKED ISSUE #4375: 🧩 Podman: add the gateway and network boundary units
## Parent

## Why

#4354 shipped `src/deploy/quadlet/hive.container` and `hive-data.volume`, and a
live rootless `systemctl start` was observed reaching `active`. What is missing
is everything in front of it: there is no gateway unit and no network unit, so
the Podman path has no equivalent of the Docker stack's front door.

The Docker stack encodes a **security boundary** that the Podman assets must not
lose. `src/docker-compose.yaml` publishes only `3001:3001` for the gateway and
leaves `7681` as `expose` only, with the reason in the file:

> SECURITY: only the authenticating proxy port is published. :7681 (the raw
> terminal) is reached through :3001/terminal, which authenticates first.

Publishing 7681 from a Quadlet unit would silently undo that.

## 30-minute boundary

The gateway container unit and the network unit. Not persistence semantics
(bullet 5), not stop/start/recreate (bullet 6), not update/rollback (9). Do not
change `src/docker-compose.yaml` or any Docker behavior.

## Deliverable

`.network` and gateway `.container` Quadlet units that put the gateway in front
of the Hive service with the same published/withheld port split as the Docker
stack.

## Acceptance criteria

- Only 3001 is published. 7681 is reachable **container-to-container** and not
  from the host. State how that was checked, not just that it holds.
- The gateway resolves the Hive service by name over the shared network, and the
  unit ordering makes the gateway depend on the Hive service.
- Image references come from `src/deploy/standalone-images.sh` (#4206);
  `src/deploy/test_standalone_image_refs.sh` must fail on drift. Verify by
  mutation, not by reading — change the tag, confirm the test fails, restore.
- The new units are discovered by the #4338 dry-run gate rather than passing
  beside it, in both rootful and `--user` modes.
- Docker untouched: `src/docker-compose.yaml` unchanged and the standalone

## Stop conditions

If the port boundary cannot be reproduced under Podman's networking, stop and
record exactly which part does not hold. Do not publish 7681 to make a test
pass — that is the one outcome worse than not shipping the unit.

Danathar:
Closes #4375. Part of #4188 — this does **not** close the parent.

## What was missing

#4354 shipped `hive.container` and `hive-data.volume`, and a live rootless `systemctl start` was observed reaching `active`. Nothing stood in front of it. The Podman path had no gateway and no network, so it had no equivalent of the Docker stack's front door — and the front door is where the security boundary lives.

`src/docker-compose.yaml` publishes `3001:3001` and leaves 7681 unpublished, with the reason in the file: 7681 is a writable terminal into the container that holds the GitHub App key. A Quadlet unit that published it would still start and still serve the dashboard; it would just also serve an unauthenticated root shell. So this slice is as much about what the units do **not** contain.

## What this adds

| File | |
| --- | --- |
| `src/deploy/quadlet/hive.network` | the shared network |
| `src/deploy/quadlet/hive-gateway.container` | the nginx gateway, and the only `PublishPort=` in the tree |
| `src/deploy/quadlet/hive.container` | modified: joins the network |
| `src/deploy/test_quadlet_port_boundary.sh` | the guard, wired into `v2-ci.yml` |

**`hive.network`.** Three defaults are load-bearing and are left unset with the reason written into the file. `DisableDNS` — aardvark-dns is what makes `hive` resolve for `nginx.conf`'s `upstream hive_api { server hive:3001; }`; the default `podman` network reports `dns_enabled: false`, which is why this is its own unit rather than "leave the default network alone". `Internal` — agents have to reach GitHub. `IPv6` — the forced-proxy egress gate in `entrypoint.sh` is `iptables-nft` only with zero `ip6tables` rules, so an IPv6-enabled network is a *measured* egress bypass (5/5 IPv4 connections redirected, 0/5 IPv6 — `src/docs/podman-ipv6-egress-bypass.md`), not a feature.

**`hive-gateway.container`.** Image from the #4206 source of truth, digest and all. `Requires=`/`After=hive.service` is Compose's `depends_on: condition: service_healthy` in systemd's vocabulary — and it means *after healthy* only because `hive.container` sets `Notify=healthy`, which is why the new test asserts that key too.

**No `ExposeHostPort=`, deliberately.** Container-to-container reachability needs no key in Podman. Quadlet's spelling of `--expose` is `ExposeHostPort=`, and `--expose` becomes a real published port the moment anything adds `-P`. Not worth it on this port; the reachability is documented in prose instead.

## Acceptance criteria

**Only 3001 published; 7681 reachable container-to-container, not from the host.** Live, rootless, Fedora 44 / Podman 5.8.4 / netavark / cgroup v2 / SELinux enforcing, against the real `ghcr.io/kubestellar/hive:stable` and the digest-pinned `nginx:alpine`. `systemctl --user start hive-gateway.service` returned 0 after 11s, pulling `hive.service` up first; both healthy.

```
$ ss -ltnp | grep -E ':(3001|3002|7681)\b'
LISTEN 0 4096 *:3001 *:* users:(("rootlessport",pid=2912635,fd=11))

$ curl -s -w 'HTTP %{http_code}\n' http://127.0.0.1:3001/api/health
{"status":"ok"}HTTP 200          # nginx has no /api/health of its own — this is end-to-end

$ curl -s --max-time 5 http://127.0.0.1:7681/ ; echo $?      -> 7    (could not connect)
$ curl -s --max-time 5 http://127.0.0.1:3002/ ; echo $?      -> 7
$ timeout 5 bash -c 'exec 3<>/dev/tcp/10.89.1.2/7681'; echo $? -> 124  (no route to a rootless netns)
```

From inside the gateway: `getent hosts hive` → `10.89.1.2 hive.dns.podman hive`, nameserver `10.89.1.1`; `nc -z hive 3001` and `3002` open, `7681` refused.

`7681` was refused because **ttyd binds `127.0.0.1` inside the container** (`/proc/net/tcp` shows `127.0.0.1:7681`, while 3001/3002 are on `[::]`). That is the image's defence and says nothing about the unit's — so `HIVE_TTYD_BIND=0.0.0.0` was set and `hive.service` restarted, putting ttyd on the wildcard, which is the worst case these units must hold under:

```
$ podman exec hive-gateway nc -z -w3 hive 7681 ; echo $?   -> 0     the network permits it
$ curl -s --max-time 5 http://127.0.0.1:7681/ ; echo $?    -> 7     the host still cannot
$ timeout 5 bash -c 'exec 3<>/dev/tcp/10.89.1.4/7681'; echo $? -> 124
$ ss -ltnH | awk '{print $4}' | grep -E ':(3001|3002|7681)$' -> *:3001   still the only one
```

Both halves hold, and for *independent* reasons: the network does not restrict container-to-container traffic, and the absence of a `PublishPort=` is what keeps 7681 off the host. Neither is relying on the other. The probe was reverted; no unit sets `HIVE_TTYD_BIND`.

**Gateway resolves Hive by name, and the ordering depends on it.** Shown above, plus the generated units:

```
hive-gateway.service   Requires=/After=  hive.service, hive-network.service
hive.service           Requires=/After=  hive-network.service, hive-data-volume.service
```

**Image references, verified by mutation not by reading.** `test_standalone_image_refs.sh` picks the gateway unit up automatically through its existing Podman globs (19 assertions across 3 assets, up from 2). Three mutations, each failing the test and then restored: dropping the digest and keeping the tag; changing `nginx:alpine` to `nginx:1.27-alpine`; flipping one hex digit of the digest.

**Discovered by the #4338 dry-run gate, not passing beside it.** `test_quadlet_generator_gate.sh` now reports `4 unit(s) found` and generates cleanly in **both** rootful and `--user` modes.

**Docker untouched.** `git diff -- src/docker-compose.yaml` is empty. `test_standalone_service_contract.sh`: 12/12. Also green: watchtower socket contract 21/21, supply-chain pins 17/17, quadlet config contract 7/7.

## The new guard

`test_standalone_service_contract.sh` asserts this boundary for Compose and *cannot see a Quadlet unit* — it parses YAML. `test_quadlet_port_boundary.sh` is the same set of invariants for `src/deploy/quadlet/`: pure text analysis, no podman required, 18 assertions. The generator gate checks that units **parse**; this checks what they **mean**, which the generator exits 0 on.

Every assertion was confirmed to fail under its mutation before being trusted:

| Mutation | Caught by |
| --- | --- |
| `PublishPort=7681:7681` on `hive.container` | 3 assertions |
| `PublishPort=0.0.0.0:7681:7681/tcp` on the gateway | 2 assertions (bind-address and `/tcp` forms parsed) |
| `IPv6=true` / `Internal=true` on the network | the corresponding default check |
| drop `Requires=hive.service` | ordering check |
| `nginx.conf` mount made writable | read-only check |
| drop `Notify=healthy` from `hive.container` | the check that keeps `After=` meaning "after healthy" |
| take `hive` off the shared network | shared-network check |

## Also

`src/docs/podman-standalone-quadlet.md` gains the boundary section, the measurements above, the install steps for all four units and `nginx.conf`, and the IPv6 trap. `shellcheck` clean on the new script. One observation recorded but explicitly **not** claimed as characterised, since restart semantics are a later slice: restarting `hive.service` propagated through `Requires=` and restarted the gateway 11s later, after which `:3001` still answered 200 on Hive's new address.

One thing that did **not** come up: the issue's stop condition — the port boundary reproduced under Podman without difficulty, so nothing had to be published to make a test pass.


clubanderson:
Reviewed on `6c6da3f2`. Verified against the tree and ran the suites locally:

- **Topology parity holds**: gateway `PublishPort=3001:3001` matches Compose's `"3001:3001"` exactly; nothing under `quadlet/` names 7681 on either side of a publish, and `hive.container` publishes nothing. Health-port split is consistent — gateway probes its own 3001 (end-to-end through `upstream hive_api { server hive:3001; }`, same command/intervals as the Compose gateway), hive probes 3002, and the #4367 config-coupling contract still passes at 7 assertions.
- **Egress-gate semantics preserved**: `entrypoint.sh` untouched (exit-77 fail-closed and the HIVE_PROXY redirect are as merged), and `hive.network` leaving `IPv6` at default is exactly right given the gate is iptables-nft-only — the guard asserts all three load-bearing defaults as "not turned on" rather than "absent", so writing them out explicitly stays legal. The `ExposeHostPort=` omission is well-reasoned: `--expose` + `-P` is a real publish, not worth it on this port.
- **#4210 ownership**: both new units carry the full four-label set (`component=gateway`/`network`), so the #4343 teardown can see them; teardown suite 74/74. `NetworkName=hive` avoids the `systemd-` prefix for the same incident-legibility reason as `VolumeName=`.
- **SELinux/rootless consistency**: `nginx.conf` mounted `:ro,Z` — private relabel is correct for a single-consumer file per the #4353 semantics; `%E` specifier matches `hive.container`'s rootful/rootless split.
- **Readiness chain is sound and guarded at its weakest link**: `Requires=/After=hive.service` only means "after healthy" because `hive.container` sets `Notify=healthy` — and the new test pins that key plus `HealthCmd` presence on both units, so the ordering can't silently degrade to "after conmon". `TimeoutStartSec=120` clears the 30s+3×10s health budget and the ADR-0017 45s trap.
- **The new guard is real**: comment-stripping before grep (the units discuss `PublishPort=` in prose at length), publish-spec parser handles bind-address and `/tcp` forms, whole-directory sweeps for 7681 and for extra publishes cover units that land later. 18/18 locally, shellcheck clean, wired into `v2-ci.yml` — orphan guard reports 14 suites, 0 orphaned.
- **Image pin**: the gateway reference matches `standalone-images.sh` byte-for-byte, digest included; image-refs suite picks it up at 19 assertions across 3 assets.
- Docker untouched (service contract 12/12), docs indexes updated on both sides, no cleanup-contract literals anywhere in the diff. CI fully green including the Quadlet generator gate (4 units, both modes) and all podman lanes.

/lgtm
/approve

PR #4383, opened 2026-08-21
🧩 podman: characterise hive-data persistence and SELinux volume behaviour

### LINKED ISSUE #4376: 🧩 Podman: characterise persistence and SELinux volume behavior
## Parent

## WHERE THIS MAY BE RUN — read before starting

This issue relabels host directories and/or reboots the machine. **Do not
execute it on a host that runs anything else.** `:z` and `:Z` rewrite the
SELinux label of the directory they are pointed at, the relabel survives the
container, and a mislabeled path can silently break an unrelated service.

If the machine you are on is not disposable, that is not a blocker: produce the
probe script and the documented procedure, run whatever part is safe (unit
generation, dry-run, static checks), and record in the ledger that the live run
was **not performed here**. That is the pattern #4337 used for the SELinux
qualification and it is an acceptable outcome for this issue.

## Why

#4354 shipped `hive-data.volume`, and #4357 recorded real AVC evidence in
`src/docs/podman-selinux-avc-evidence.md`. What is still unwritten is what the
volume actually guarantees: whether data survives, what label it carries, and
what happens to it under SELinux enforcing when the container is recreated.

Build on #4357's measurements rather than re-deriving them.

## 30-minute boundary

Persistence and SELinux volume behavior for the existing volume unit. Not
backup/restore or migration (bullet 8), not reboot persistence (bullet 6).

## Deliverable

A characterisation of `hive-data` under enforcing: label, ownership, what
survives container removal and recreation, and how a bind mount differs from the
named volume.

## Acceptance criteria

- States the label the volume's contents carry and whether a category is
  applied, with the command that shows it.
- Data written by the Hive service is shown to survive `podman rm` and unit
  recreation.
- The named-volume path is compared against a bind mount, since the operator doc
  offers bind mounts for config and secrets — say which needs `:z`, which needs
  `:Z`, and why, citing #4357 where it already measured it.
- Any remediation never disables SELinux and never widens a secret's mode.
- Distinguishes observed from expected throughout.

## Stop conditions

If the host is not disposable, record the procedure and mark the live portion
not executed. If a case fails under enforcing, record the denial verbatim rather
than relabeling something outside the test's own scratch directory.

Danathar:
Closes #4376. Part of #4188 — this does **not** close the parent.

## What was unwritten

#4354 shipped `hive-data.volume` and said so in the unit itself: *"persistence and SELinux volume semantics are deliberately NOT characterised here — that is its own slice."* This is that slice.

It **builds on** #4337 and #4357 rather than re-deriving them. Those established the `:z`/`:Z`/unlabeled matrix for **bind mounts**, with the kernel's own AVC records per case. Neither measured the **named volume** — and the named volume is the interesting one, because podman labels it *itself, without being asked*, and the label it picks is what decides whether a recreated container can still read the data.

## The host, and what that means for this issue

The issue's warning is about relabeling host directories and rebooting. Neither was needed for the parts that matter here: a named volume lives inside podman's own storage and no host path is relabeled to characterise it, and every bind-mount source was a fixture the probe creates under `/var/tmp` and removes on exit — which is what the issue's own stop condition asks for. So **the live portion was executed**, on an enforcing host, and the things that genuinely could not be done here are listed as not executed rather than glossed.

Enforcing, Aurora 44 (Kinoite) / Podman 5.8.4 rootless / crun 1.28 / `container-selinux-2.250.0` / targeted+MLS. Note this host has GNU coreutils *unshadowed*, unlike #4357's — which is why the probe resolves its label reader instead of assuming either outcome.

## The label, and the command that shows it

```
$ MP=$(podman volume inspect hive-data --format '{{.Mountpoint}}')
$ /usr/bin/stat -c '%C' "$MP"
system_u:object_r:container_file_t:s0
```

`container_file_t`, and the level is bare `s0` — **no MCS category**. Everything else follows from that.

It is podman's doing, not inheritance, which matters because an inherited label would be an accident of where volumes happen to live:

```
$ /usr/bin/stat -c '%C' ~/.local/share/containers/storage/volumes
unconfined_u:object_r:data_home_t:s0
$ mkdir -p .../volumes/by-hand/_data && /usr/bin/stat -c '%C' .../volumes/by-hand/_data
unconfined_u:object_r:data_home_t:s0      # a hand-made directory keeps data_home_t
```

**Ownership** changes on first use and the label does not: volume root `1000:1000` → `525288:525287`, which is container `1001:1000` through this user's subuid range, matching the image's `/data`. That is podman's copy-up chown, and it is why the Hive service (uid 1001 `dev`) owns its own state directory with no operator step.

## Survival, measured on the real service

The generated `ExecStart` carries `--rm`, so **the container is deleted every time the unit stops** — the volume is the only thing that persists. Hive wrote its own state (`hive-id`, `hive-state.json`, `hive.yaml.runtime`, `proxy-ca.pem`, `beads/`), then the units were *deleted*, not merely restarted:

```
$ systemctl --user stop hive.service
$ podman ps -a --format '{{.Names}}' | grep -cx hive      -> 0        (--rm deleted it)
$ rm ~/.config/containers/systemd/hive.container ...      ; daemon-reload  -> 0 units
$ install -Dm644 ... ; daemon-reload                                       -> 2 units
$ systemctl --user start hive.service                     -> 0 in 11s

$ podman exec hive sh -c 'cat /data/hive-id; sha256sum /data/proxy-ca.pem'
hive-keen-ray
278dcad37f9148cfd7ffe99c44b6da34619fc271e5e129981438d6284a6d7eea  /data/proxy-ca.pem
```

Same identity, same CA, sentinel intact — and Hive said so itself in the journal for that start: `state restored`, `mode history restored`, `cadence overrides restored`, `governor mode restored`.

**Every container in that sequence had a different MCS category** — `c94,c887`, `c126,c769`, `c87,c359` — and every one could read the data. That is the bare `s0` doing its job.

## The finding: `:Z` on the volume line

Adding `:Z` to the `hive-data` line looks like hardening and is the opposite:

```
$ podman run --rm -v hive-data:/data:Z ...     ; /usr/bin/stat -c '%C' "$MP"
system_u:object_r:container_file_t:s0:c619,c667        # a private category now
$ podman run --rm -v hive-data:/data ... -c 'cat /data/probe.txt'
cat: /data/probe.txt: Permission denied
$ podman run --rm -v hive-data:/data:z ...     ; /usr/bin/stat -c '%C' "$MP"
system_u:object_r:container_file_t:s0                  # :z puts it back
```

It does not fail immediately, which is what makes it dangerous: while *every* mount carries `:Z`, podman re-relabels each start and it works. It breaks the first time anything mounts without the flag — a `podman run` to look at the data, a second container, an edit that drops the suffix. And per **#4357 finding 3** that denial is **silent**: a category denial records no AVC and no `SELINUX_ERR`, so `ausearch` returns nothing. The data is never at risk, only the label; the service is just down until someone works out why.

The unit comments and the operator page now say this at the line where the edit would be made, and the probe's static section fails on it.

## Named volume vs bind mount

| Mount | Flag | Why |
| --- | --- | --- |
| `hive-data.volume:/data` | **none** | already `container_file_t:s0`; must stay category-free so a recreated container can read it |
| `%E/hive/hive.yaml:…:ro,Z` | **`:Z`** | a host path keeps its own type and is denied without a flag; one container reads it |
| `%E/hive/secrets:…:ro,Z` | **`:Z`** | same, and `:z` is the *shared* label — these are secrets (#4357 measured both flags leaving the `0440` mode and hive-launch group ownership untouched) |
| `%E/hive/hive.env` (`EnvironmentFile=`) | **none, and none is possible** | not a mount |

With the real unit running, its `:Z` mounts carried the container's exact category while the volume beside them carried none:

```
hive.yaml   container_file_t:s0:c87,c359   mode=644
secrets     container_file_t:s0:c87,c359   mode=750  gid=1002 (in the userns)
hive-data   container_file_t:s0            mode=755
container   container_t:s0:c87,c359
```

Modes and group ownership unchanged. **No remediation anywhere disables SELinux or widens a mode**, and none is needed.

`EnvironmentFile=` becomes `--env-file`, read by podman **on the host**: the file stays `user_tmp_t:s0` before and after and never needs a flag. So don't "fix" a start failure by relabeling `hive.env` — the cause is that the file is missing.

## Also measured

What destroys the data: not `systemctl stop` (container only), not `systemctl stop hive-data-volume.service` (the generated unit is `Type=oneshot`/`RemainAfterExit=yes` with **no `ExecStop`**), not `podman rm -v` (that is anonymous volumes only — verified separately). `podman volume rm`, the prunes, and `bin/hive-podman-teardown.sh` do — the last by design, and the #4210 ownership labels do land: `podman volume ls --filter label=io.kubestellar.hive.owned=true` returns `hive-data`.

One cosmetic detail recorded so it isn't mistaken for a problem: entries copied up from the image carry `unconfined_u:…container_file_t:s0` and entries created by the container carry `system_u:…container_file_t:s0`. Type and level — the fields type enforcement and the MCS constraint consult — are identical; only the SELinux user field differs.

## Distinguishing observed from expected

Every table row is marked, and there is an explicit **not executed** list: **rootful** (label *expected* to be the same since podman applies it, ownership numbers will differ — expectation, not measurement), **reboot persistence** (bullet 6, out of scope), **backup/restore/migration** (bullet 8, out of scope), **non-atomic hosts**, **podman's internal mechanism** (observed at the filesystem, not traced through source), and **a second host or architecture**.

## What ships

- `src/deploy/probe_podman_volume_persistence.sh` — 20 assertions, exit `78` on a non-enforcing host rather than a vacuous pass. Its `static` section runs *before* the enforcing gate so it is checked anywhere; all four of those assertions were confirmed to **fail** under their mutation (`:Z` or `:z` on the volume line; either bind mount downgraded `:Z`→`:z`). `shellcheck` clean.
- `src/docs/podman-volume-persistence.md` — the characterisation.
- Unit comments in `hive-data.volume` and `hive.container`, a trap in the operator page, and both READMEs.

**Not wired into CI, deliberately** — hosted runners are not enforcing, so the job would report `78` on every run. Same lane as `probe_podman_selinux_avc.sh` and `qualify_podman_selinux.sh`, neither of which is CI-wired either.

`src/docker-compose.yaml` untouched; the six existing deploy contract tests still pass.

## Safety

Bind sources were `/var/tmp` fixtures removed on exit. The live unit run used a scratch config directory with agents disabled, never `%E/hive`. Every probe object is named `hive-volprobe-<pid>-*` and the probe **refuses to start** if such a name already exists; cleanup removes exactly what it created and nothing is pruned. No `setenforce`, no `label=disable`, no `--privileged`, no `chcon` outside the fixture, no `chmod` widening anything. The probe deliberately does *not* pin `--root`/`--runroot` — the property under test is what podman does to a volume in a real store — and the header says so rather than leaving it looking like an omission; `--store` isolates anyway.


clubanderson:
Reviewed on `f4d0c86f`, merged `v4` forward and added one contract fix; final head `8e59f9c3`.

- **The characterisation is sound and consistent with the merged evidence base**: named-volume `container_file_t:s0` with no MCS category is exactly what lets the `--rm`-deleted, freshly-categorised container re-read its state, and the `:Z`-on-the-volume-line finding is the right thing to pin — especially given #4357 finding 3 (category denials record no AVC), which makes that failure mode undebuggable in the field. The bind-mount rows (`:Z` on config/secrets, `:z` explicitly rejected as the shared label) match #4353/#4357, modes and hive-launch group ownership asserted unchanged, and the label reader is resolved rather than assumed per the #4361 fix. Observed-vs-expected discipline is kept (rootful/reboot/backup listed unexecuted, not glossed).
- **Probe hygiene checks out**: exit-78 stop condition instead of a vacuous pass, static section deliberately ahead of the gate (runs anywhere — 3/3 here), `hive-volprobe-<pid>-` name discipline with refuse-to-start collision checks, cleanup removes only its own names, fixtures under `/var/tmp`, no setenforce/label=disable/--privileged/mode-widening. MCS detection reads the level field only, avoiding the `:container_file_t` false-positive. Shellcheck clean. Not CI-wiring it is correct — same lane as the AVC and qualification probes, hosted runners aren't enforcing — and the orphan guard covers `bin/` suites, so nothing is orphaned (14 suites, 0 orphaned).
- **Fix-forward `8e59f9c3`**: the destroyer table in `src/docs/podman-volume-persistence.md` quoted the system-wide prune's `--volumes` form literally, and that file is not on `bin/test_hive_podman_cleanup.sh`'s exempt list — the contract suite greps every tracked file and would have gone red repo-wide. Re-cited by man-page name per the established convention; contract suite now 127/127.
- **Merge forward `311161a9`**: trivial `src/deploy/README.md` collision with #4380's gateway bullet, both entries kept. Post-merge the whole battery is green locally: port boundary 18/18, image refs 19, service contract 12/12, config contract 7, teardown 74, cleanup contract 127.
- `#4210` ownership confirmed live in the measurement itself (`volume ls --filter label=io.kubestellar.hive.owned=true` → `hive-data`); Docker untouched. CI fully green on `8e59f9c3` including all podman lanes.

/lgtm
/approve

PR #4385, opened 2026-08-21
🧩 podman: Quadlet lifecycle in both root modes, and the first live rootful start (#4377)

### LINKED ISSUE #4354: 🧩 Podman: add the Hive service Quadlet unit and its volume

### LINKED ISSUE #4377: 🧩 Podman: stop/start/recreate and reboot persistence, and the first live rootful start
## Parent

## WHERE THIS MAY BE RUN — read before starting

This issue relabels host directories and/or reboots the machine. **Do not
execute it on a host that runs anything else.** `:z` and `:Z` rewrite the
SELinux label of the directory they are pointed at, the relabel survives the
container, and a mislabeled path can silently break an unrelated service.

If the machine you are on is not disposable, that is not a blocker: produce the
probe script and the documented procedure, run whatever part is safe (unit
generation, dry-run, static checks), and record in the ledger that the live run
was **not performed here**. That is the pattern #4337 used for the SELinux
qualification and it is an acceptable outcome for this issue.

## Why

Two gaps meet here.

**Reboot persistence is unproven.** ADR-0017 chose `loginctl enable-linger` for
rootless boot persistence and `src/docs/podman-standalone-quadlet.md` documents
it, but nothing has rebooted a host to confirm the service comes back. Linger is
exactly the setting that looks fine until a reboot proves otherwise.

**Rootful was never started live.** #4354's doc records this honestly: the
rootful path is dry-run generated only. The rootless start was observed
(`ActiveState=active` after 52s, container healthy); rootful has never run.

## 30-minute boundary

Lifecycle only: stop, start, restart, recreate, and reboot persistence, in both
root modes. Not update/rollback (bullet 9), not backup (bullet 8).

## Deliverable

A documented, repeatable lifecycle check covering both root modes, plus the
first live **rootful** start.

## Acceptance criteria

- `systemctl stop` / `start` / `restart` are each exercised, with what the unit
  reports at each step — including that `start` returning still means healthy.
- Recreate (remove the container, start the unit again) is shown to reattach the
  existing `hive-data` volume rather than silently starting empty.
- **Reboot persistence is proven by an actual reboot**, rootless with linger
  enabled and rootful via the system manager. Record what `systemctl is-active`
  says after boot, and how long it took to become active.
- The rootful live start closes #4354's recorded gap; add its result beside the
  rootless row in `podman-standalone-quadlet.md` rather than replacing it.
- If linger is **not** enabled, record what happens then too — that is the
  failure operators will actually hit.

## Stop conditions

If the host cannot be rebooted, do everything else, and record the reboot row as
not executed rather than inferring it from `is-enabled`.

Danathar:
Closes #4377. Part of #4188; does not close the parent.

Lifecycle only — stop, start, restart, recreate, boot persistence, in both root
modes. Not update/rollback (bullet 9), not backup (bullet 8).

## Where this was run

Fedora 44, Podman 5.8.4, cgroup v2, **SELinux enforcing**, on a disposable VM,
against the real `ghcr.io/kubestellar/hive:stable` image. Rootless config in
`~/.config/hive`, rootful in `/etc/hive`, both with `dashboard.port: 3002` as
#4367 requires. Everything below is measured on that host unless it says
otherwise. Every host change was torn down afterwards.

## Three things looked fine and were not

**1. A clean `systemctl stop` left the unit `failed`.** Hive's entrypoint
handles SIGTERM, persists state, and exits **143**; systemd's default success
set is exit 0 alone. So a deliberate stop ended:

```
ActiveState=failed  SubState=failed  Result=exit-code  ExecMainStatus=143
hive.service: Main process exited, code=exited, status=143/n/a
hive.service: Failed with result 'exit-code'.
```

Red in `systemctl status`, true for `systemctl is-failed`, enough to put the
host into `degraded`. Identical in **both** root modes.

**2. `systemctl enable hive.service` does not work and never did.** Quadlet
units are *generated*, and systemd refuses to enable a generated unit — so the
boot-persistence step in `podman-standalone-quadlet.md` was an error message in
both modes:

```
$ sudo systemctl enable hive.service
Failed to enable unit: Unit /run/systemd/generator/hive.service is transient or generated
$ systemctl --user enable hive.service
Failed to enable unit: Unit /run/user/1000/systemd/generator/hive.service is transient or generated
```

What actually wires them is `[Install] WantedBy=default.target` inside
`hive.container`, which the generator turns into a `default.target.wants/`
symlink on every `daemon-reload`. Corrected in the doc.

**3. `systemctl is-enabled` cannot answer "will Hive come back?"** It reports
`generated` for these units in *every* case — lingering on, lingering off,
symlink deleted. #4377 forbade inferring the reboot row from `is-enabled`, and
this is the evidence that was right.

## The unit change, and why it is two lines

```ini
SuccessExitStatus=143 SIGTERM
Restart=always
```

The first line alone is **not** sufficient, and that was measured rather than
reasoned. Making 143 a success takes it out of `Restart=on-failure`'s set, so
with the old `Restart=on-failure` still in place an external `podman rm -f hive`
left the unit at `inactive/dead/success/143` — service down, nothing restarted,
and **nothing for monitoring to see**, strictly worse than the `failed` state
the first line was fixing.

| | `on-failure` | `always` |
| --- | --- | --- |
| `systemctl stop` | `inactive/dead/success`, stays stopped | `inactive/dead/success`, stays stopped (verified over 35s) |
| `podman rm -f hive` | `inactive/dead/success`, **stays down** | `activating/auto-restart` → healthy in ~35s |

systemd never restarts a unit a stop job stopped, so `always` keeps a
deliberate stop clean while recovering an external removal. `check` mode fails
on `Restart=on-failure` whenever `SuccessExitStatus` is set, so the pair cannot
be split up again by accident.

## Acceptance criteria

**stop / start / restart each exercised, with what the unit reports**

| Step | rootless | rootful |
| --- | --- | --- |
| first `start` | rc 0 in **11.4s**, `active/running`, `healthy` | rc 0 in **11.7s**, `active/running`, `healthy` |
| `stop` before the fix | `failed/failed/exit-code/143` | `failed/failed/exit-code/143` |
| `stop` after the fix | `inactive/dead/success`, `is-failed`→`inactive` | `inactive/dead/success`, `is-failed`→`inactive` |
| `start` again | rc 0 in **10.7s** | rc 0 in **10.8s** |
| `restart` | rc 0 in **11.8s** | rc 0 in **11.4s** |

**`start` returning still means healthy** — visible in every ten-plus-second
return above, and directly during a recreate, where the container reads
`Up 10 seconds (starting)` while the unit still reads `activating`, and turns
`active` in the same step it turns `healthy`. A start timed from an
already-active unit is *not* a start time; the probe says so rather than
printing 7ms as though Hive started in 7ms.

**Recreate reattaches `hive-data` rather than silently starting empty** — proven
with a marker written into `/data`, shown absent from the image, and surviving
stop/start, restart, and recreate in both modes. **Falsified too**: with the
volume removed, the service still started cleanly in 11.4s and the marker was
gone. That empty start is indistinguishable from a healthy one except by the
marker, which is exactly the failure the criterion asked to rule out.

**The first live rootful start** closes #4354's recorded gap — rc 0 in 11.7s,
`healthy`, NET_ADMIN egress enforcing (`iptables-nft: outbound :443 -> :18443`,
`ambient CAP_NET_ADMIN granted`), and `%E` measured as `/etc` from the running
container (`/etc/hive/hive.yaml`, `/etc/hive/secrets`) rather than read from
`systemd.unit(5)`. Added **beside** the rootless row, not replacing it.

**If linger is not enabled, what happens** — measured on a throwaway user with
**no login session**, which is what boot looks like:

| | `Linger` | `user@UID.service` | its enabled unit |
| --- | --- | --- | --- |
| before | `no` | `inactive` | never ran |
| `enable-linger` | `yes` | `active/running` in **305ms** | started, no session involved |
| `disable-linger` | `no` | `inactive/dead` | process **gone** |

Throughout, the unit's `default.target.wants/` symlink was present and
unchanged — nothing in the unit's own enablement notices it will never run.

One artifact recorded because it nearly became a wrong result:
`systemctl --user -M <user>@ is-active` **starts the user manager it asks
about**, so the third row initially reported the unit as active and recreated
its marker. It was re-measured with no `-M` query.

## Reboot persistence: NOT EXECUTED, not inferred

Per the issue's stop condition, in both root modes. The only host available was
the one running the session doing the measuring; a reboot would have ended it
mid-run. `is-enabled` was **not** used as a substitute — see finding 3 for why
it could not have been one.

Every link short of the boot **is** measured: the generator installs the
`default.target.wants/` symlink in both modes, `default.target` is reached in
both managers, and lingering starts a sessionless user manager and its units in
305ms while its absence stops them. What is unproven is the composition of
those links across an actual kernel boot.

Two commands close it on a host that can be rebooted:

```sh
bin/hive-podman-lifecycle-probe.sh check   # and --rootful; confirm wiring first
sudo systemctl reboot
bin/hive-podman-lifecycle-probe.sh boot-check   # and --rootful
```

`boot-check` reports `is-active` and time-to-**healthy** after userspace start,
and refuses to present a delta larger than ten minutes as boot evidence, since
a unit started by hand hours after boot also has a later `ActiveEnterTimestamp`.

## What ships

- `src/deploy/quadlet/hive.container` — `SuccessExitStatus=143 SIGTERM` +
  `Restart=always`, with the measurement in the comment.
- `bin/hive-podman-lifecycle-probe.sh` — the repeatable check. `check`
  (read-only, safe on a serving host), `exercise` (drives the full lifecycle),
  `boot-check`. Rootless by default, `--rootful` for the system manager. Exit
  codes match the other Podman scripts: 0 / 78 / 64.
- `bin/test_hive_podman_lifecycle_probe.sh` — **19 contract tests**, every input
  mocked, so it runs on a runner with no Podman and no lingering user. The
  negative cases are the point: a probe that cannot fail is not evidence.
- `src/docs/podman-quadlet-lifecycle.md` — the measurements, including the
  not-executed reboot row.
- `podman-standalone-quadlet.md` — corrected boot-persistence section, the
  rootful row beside the rootless one, and the `%E` caveat closed.
- `v2-ci.yml`, `src/docs/README.md` — wiring.

## Verification

```
bin/test_hive_podman_lifecycle_probe.sh    19 passed, 0 failed
src/deploy/test_quadlet_config_contract.sh PASS: 7 assertions
src/deploy/test_quadlet_generator_gate.sh  SUMMARY: 0 failure(s)
src/deploy/test_standalone_image_refs.sh   PASS: 17 assertions
shellcheck -S warning                      clean
```

Docker is untouched; `src/docker-compose.yaml` is unchanged and remains the
default runtime.


clubanderson:
Reviewed on `721910f8`, merged `v4` forward (final head `4a538b4e`). Verified:

- **The 143 finding is real**: `entrypoint.sh` traps INT/TERM into a cleanup that kills and waits on the Hive pid, so the script's final `wait` returns 128+15 — a clean stop genuinely exits 143, and systemd's default success set is exit 0 alone. `SuccessExitStatus=143 SIGTERM` is the right fix, and the measured pairing argument is the strongest part of the PR: making 143 a success removes it from `on-failure`'s trigger set, so the pair with `Restart=always` is load-bearing, not stylistic — and `check` mode fails on a split pair so it cannot regress silently. systemd's stop-job exemption keeps deliberate stops stopped.
- **The enable/is-enabled findings check out**: generated units cannot be enabled, the `[Install]` section → `default.target.wants/` symlink is the actual wiring, and `is-enabled` reporting `generated` in every case (linger on/off, symlink deleted) justifies the probe reading the generator output and linger instead. Reboot row honestly NOT EXECUTED with the boot-check refusing >10-min deltas as boot evidence — the right epistemics.
- **The mocked suite is genuinely falsifiable**: check-mode mutation is refused-and-logged so drift into start/stop fails the suite, and the negative cases each pin a specific finding text. 4 of 19 fail on this macOS host solely because `gen_ls` uses GNU `find -printf` (BSD find rejects it, so the wants-symlink check sees nothing) — with a GNU-find shim it's **19/19**, and CI's ubuntu runner is authoritative, same precedent as the preflight-host suite. Wired into `v2-ci.yml`; orphan guard: 15 suites, 0 orphaned.
- **Merge forward `4a538b4e`**: two real doc collisions with the just-merged #4380. Resolved in favour of this branch's *measured* boot-persistence correction (#4380's install steps said `systemctl enable` — the exact error this PR characterises), extended to cover the gateway unit's symlink, and kept #4380's gateway-restart observation beside this branch's rootful-closed and reboot-not-executed rows. Post-merge battery green: port boundary 18/18, image refs 19, service contract 12/12, config contract 7 (TimeoutStartSec unchanged at 300 = the probe's 5min expectation), teardown 74.
- Shellcheck clean at `-S warning` as claimed (remaining SC2015 infos are benign — `ok` is a printf that cannot fail); no cleanup-contract literals anywhere in the diff; Docker untouched. CI fully green on `4a538b4e` including all podman lanes and the Quadlet gate.

/lgtm
/approve

PR #4407, opened 2026-08-21
🧩 Podman: deliberate manual update and rollback, pinned by digest (#4378)

### LINKED ISSUE #4378: 🧩 Podman: deliberate manual update and rollback
## Parent

## Why

ADR-0017 chose Quadlet specifically because `Notify=healthy` means a unit that
reports started is actually serving. That property only pays off if there is a
defined way to move to a new image and get back if it is bad. #4354 gave us a
unit; nothing says how to update it or undo an update.

The Docker stack has Watchtower for this. The Podman equivalent is deliberately
a separate, later slice (bullet 10, health-aware auto-update) — this issue is
the **manual, deliberate** path that the automatic one will later build on.

## 30-minute boundary

Manual update and rollback only. No auto-update, no registry polling, no
`AutoUpdate=registry`. Do not change Docker behavior.

## Deliverable

A documented update and rollback procedure for the Quadlet deployment, with
whatever unit or script support it needs.

## Acceptance criteria

- The procedure states how the running image is pinned and how a specific prior
  version is returned to — by digest, not by a floating tag.
- Rollback is shown to work from a **failed** update, not only from a healthy
  one. An update to an image that never becomes healthy must leave the operator
  able to get back, and the doc must say what the unit looks like while that is
  happening (`activating`, then the `TimeoutStartSec` expiry).
- Data in `hive-data` survives both directions.
- The doc distinguishes what was **observed** from what is expected, per the
  house style in `src/docs/podman-standalone-quadlet.md`.

## Stop conditions

If the environment cannot take the unit through a real update and a real
rollback, write the procedure and record plainly that it was not executed. Do
not present a dry-run as a rollback.

Danathar:
Closes #4378. Does not close the parent (#4188).

ADR-0017 chose Quadlet because `Notify=healthy` means a unit that reports
started is actually serving. #4354 shipped the unit and #4377 exercised its
lifecycle; neither said how to move it to a new image, or how to get back when
the new one is bad. This is that slice — **manual and deliberate**: no registry
polling, no `AutoUpdate=registry`, no change to Docker.

## The pin has to live in a drop-in

The scanner note on the issue is right: `hive.container:40` names
`ghcr.io/kubestellar/hive:stable`, a floating tag, and **it has to keep naming
it** — `src/deploy/test_standalone_image_refs.sh` checks that line against
`src/deploy/standalone-images.sh`. Today's `:stable` and last month's `:stable`
are different images with the same name, so it records a channel, not a version,
and you cannot roll back to it.

So the deployed version is pinned by digest in a Quadlet drop-in,
`hive.container.d/10-image.conf`, which Quadlet merges into the unit on every
`daemon-reload`. The repo's file is untouched, the current pin is one `grep`
away, and the later auto-update slice (bullet 10) has an obvious file to manage.

`bin/hive-podman-update.sh` writes that file and keeps its own pin history in it
as comments Quadlet ignores. Each entry records whether the script *watched that
pin become healthy*, so `rollback` returns to the newest `healthy` entry rather
than to the previous one — which is exactly what makes it usable from a failed
update, where the top entry is the pin that just broke.

## Executed, in both root modes

Fedora 44, Podman 5.8.4, systemd 259, SELinux **enforcing**, between two real
`v4` builds — checked with `hive --version` inside the container, not just by
digest, so each step is shown to have changed the running code:

| build | digest | reports |
| --- | --- | --- |
| tag `b35e9cc`, built 2026-08-14 | `sha256:2326e85d…` | `hive 3.0.0 (commit b35e9cc, branch v4)` |
| tag `stable`, built 2026-08-21 | `sha256:ec8e69bc…` | `hive 3.0.0 (commit 77300c8, branch v4)` |

**A healthy update is 11s and is not zero-downtime.** ~1s `deactivating`, ~10s
`activating` with the new container `(starting)`, then `active` in the same
second it turns `(healthy)`. Sampled once a second; both root modes.

**A failed update costs 301s before anything says so.** Staged by pinning a
digest that runs and never answers `/api/health`. `systemctl restart` blocked
for the full `TimeoutStartSec=300` and returned 1 at **301s** in both modes. The
unit held `activating/start` throughout — and the container read `(starting)`
for the first ~150s (`HealthStartPeriod` plus the retry budget) before
`(unhealthy)`, with the unit not failing when it did.

**And the unit never reads `failed`.** `Restart=always` (from #4377, for reasons
unrelated to this) moves it from the timeout straight to
`activating/auto-restart` and back into another 300-second attempt, so
`systemctl is-failed` reported **`activating`** at every point. Alerting keyed on
`failed` does not fire on a bad update — that is in the doc as a table.

**Rollback out of that failed state took ~11s and one decision.** `systemctl
stop` cancelled the in-flight bad start job in **555ms** — a `restart` would
have queued behind it and waited out the remaining timeout — then a 10s start,
healthy. `hive-data` was intact in both directions, proven with a marker file in
a volume holding real state (bead ledger, agent state, the proxy CA).

Rollback is shown out of a **failed** update, not only a healthy one, as the
acceptance criteria require.

## Two traps, both measured

`podman image inspect … RepoDigests` returns the **per-architecture** digest of
a multi-arch tag, not the manifest list: `:stable` gave `sha256:5d3f442f…` that
way and `sha256:ec8e69bc…` from the registry. Pinning the first produces a pin
that resolves on your laptop and fails on arm64. `resolve` reads skopeo, or the
`podman images --digests` column, never `RepoDigests`.

`systemctl cat hive.service` **never names the drop-in** — Quadlet merges before
generating, so `10-image.conf` appears nowhere in its output and the pin is
visible only at the end of `ExecStart=`.

## What is in here

- `bin/hive-podman-update.sh` — `status`, `resolve`, `pin`, `rollback`, `unpin`;
  `--rootful`; exit 0/78/64 like the other Podman scripts. `status` and
  `resolve` are read-only.
- `bin/test_hive_podman_update.sh` — 46 cases, every input mocked (no Podman, no
  Quadlet, no privileges, no pull), wired into `v2-ci.yml`. The cases that
  matter are the failed ones. It caught a real bug in this script: the drop-in
  header has to mention the `HIVE-PIN` marker to explain it, and the first
  parser read that sentence as the newest history entry.
- `src/docs/podman-quadlet-update-rollback.md` — the procedure, with and without
  the script, the measurements above, and a **Not executed here** section: the
  never-healthy image was `nginx:alpine` pinned by digest standing in for a bad
  Hive build, and rollback across a `/data` schema change is called out as
  unproven rather than implied.
- `src/deploy/quadlet/hive.container` — comment only, no functional change,
  saying why the tag stays floating and where the pin goes instead.
- Pointers from `podman-quadlet-lifecycle.md` (whose line 10 declared this slice
  as future work), `podman-standalone-quadlet.md`, and the docs index.

Contract suites re-run green: `test_standalone_image_refs.sh`,
`test_quadlet_config_contract.sh`, `test_quadlet_port_boundary.sh`,
`test_quadlet_generator_gate.sh`, `test_standalone_service_contract.sh`,
`test_supply_chain_pins.sh`, `test_bin_suites_wired.sh`.


Danathar:
Heads-up on the one red check, `quadlet --dryrun, rootful and --user`: it is **not** this PR. `.github/workflows/quadlet-gate.yml` pins `quay.io/podman/stable@sha256:312a77be…`, and that digest no longer resolves:

```
Error: unable to copy from source docker://quay.io/podman/stable@sha256:312a77be…:
reading manifest sha256:312a77be… in quay.io/podman/stable: manifest unknown
```

It fails at the "Install the Quadlet generator" step, before any repository unit is read, and it is currently failing the same way on unrelated branches (`quality/test-inception-integration-tag`, `quality/test-intent-alignment-verdict`). The gate itself passes locally against the units in this PR:

```
$ bash src/deploy/test_quadlet_generator_gate.sh
  PASS: user: every repository unit generates cleanly
SUMMARY: 0 failure(s)
```

Refreshing that pin needs its own change (and a decision about provenance), so I have left it alone rather than widening this PR. Worth noting it is the exact failure mode `src/docs/podman-quadlet-update-rollback.md` warns about for rollback targets: a digest that has been garbage-collected upstream cannot be pulled at all, which is why the doc says to keep the previous image on the host.

clubanderson:
Reviewed on `f5dec066`, one CI fix-forward (final head `6d257ba3`). Verified:

- **The drop-in design is the right call**: `hive.container` must keep naming the floating tag (the #4206 image-refs contract checks it against `standalone-images.sh`), so pinning in `hive.container.d/10-image.conf` is the only placement that gives a rollbackable version without fighting the source-of-truth test — image refs still 19/19 with the comment-only unit change. History-as-comments with the `HIVE-PIN <outcome>` ledger is elegant, and the timestamp-anchored parse regex fixing the self-referential header bug is exactly the kind of thing the mocked suite exists to catch.
- **Rollback semantics are correct where it counts**: returning to the newest *healthy* entry (not merely the previous one) is what makes rollback usable from a failed update, and the stop-before-start (a stop job cancels the in-flight 300s bad attempt in <1s; a restart would queue behind it) matches the #4377-measured `Restart=always` loop — including the alerting trap that the unit never reads `failed`, documented as a table. Pull-before-restart keeps the 3.8GB cold pull out of `TimeoutStartSec`, and a failed pull provably changes nothing (asserted).
- **The multi-arch digest trap is real**: `RepoDigests` hands back the per-arch manifest digest; resolving via skopeo (or the `--digests` column) gets the index digest. Consistent with how the merged Quadlet stack pins (index digests in the gateway unit and the gate itself).
- **Volume/egress contracts preserved by construction**: the script contains no volume, network, or firewall operation of any kind — it writes one drop-in, reloads, and restarts; the #4383 characterisation (bare `s0`, recreated containers re-read `/data`) is exactly what the measured marker survival relies on. No cleanup-contract literals anywhere in the diff.
- **Suite**: 46 cases, all inputs mocked, wired into `v2-ci.yml`; orphan guard 16 suites/0 orphaned; shellcheck clean at `-S warning`. 4 cases fail on this macOS host only because BSD `install` lacks `-D` — with a GNU-install shim it's **46/46**, CI's ubuntu authoritative (same precedent as the lifecycle suite's `find -printf`).
- **Fix-forward `6d257ba3`**: the `quadlet --dryrun` gate went red on every branch since ~03:45Z — quay.io re-pushed `podman/stable` and GC'd the index the gate pinned, so the pull 404s with "manifest unknown" (v4's last green was 02:34Z; baseline, unrelated to this PR). Refreshed the pin to the current index after verifying it still carries `org.opencontainers.image.version=5.8.4` — same podman, same generator, new digest. Gate green on the new pin.

CI fully green on `6d257ba3` including all podman lanes and the recovered Quadlet gate.

/lgtm
/approve

PR #4409, opened 2026-08-21
🧩 Podman: backup, restore, and Docker-to-Podman migration (#4406)

### LINKED ISSUE #4406: 🧩 Podman: backup, restore, and Docker-to-Podman migration
## Parent

Part of #4188 (second wave, bullet 8). This issue must not close the parent.

## Why

`src/docs/backup-restore.md` contains **zero** occurrences of "podman". Its
standalone section is titled "Docker Compose" and every command in it is a
Docker command:

```bash
docker run --rm -v src_hive-data:/data -v "$(pwd)":/backup alpine tar czf ...
docker volume create src_hive-data
docker compose down -v          # deletes the named volume
```

#4376 established how `hive-data` behaves under Podman and #4377 proved it
survives recreate. Neither says how to get the data *out* and back *in*, which
is the operation an operator actually needs on the day it matters.

## Where to run this

**A disposable VM, not a workstation.** Three of the steps below are destructive
by design — the deliverable *is* a wipe-and-restore cycle — and one acceptance
criterion cannot be satisfied on a developer machine at all:

- The procedure removes and recreates a named volume, and exercises the
  `down -v` equivalent. A rootless developer store typically holds unrelated
  toolboxes, Distroboxes, and build volumes; that is the wrong place to
  rehearse volume deletion.
- The rootful half writes into `/var/lib/containers`, the default rootful
  store. Isolated `--root`/`--runroot` does not cleanly apply once systemd and
  Quadlet are driving the lifecycle.
- The migration case requires creating a Docker volume and then destroying it.
- #4188 asks for behaviour on *"a clean supported Linux host with Podman and
  **no Docker installation**."* A host with Docker installed cannot demonstrate
  that, whatever the result looks like.

## What is different under Podman, specifically

Not a search-and-replace of `docker` with `podman`:

1. **The volume name in the doc is wrong for Podman.** Compose prefixes named
   volumes with the project directory, which is why the doc says `src_hive-data`
   and tells the reader to check `docker volume ls`. `hive-data.volume` sets
   `VolumeName=hive-data` with no prefix. An operator following the current doc
   restores into a volume nothing mounts.
2. **Rootless ownership.** A tar taken from a rootful or Docker volume carries
   real UIDs; extracting it into a rootless Podman volume lands it under mapped
   subuids. Ownership inside `/data` is not cosmetic — the entrypoint drops to
   `dev` and per-agent UIDs start at 2001. The archive/extract path has to be
   run inside the user namespace (`podman unshare`, or a container that already
   has the mapping) rather than from the host shell.
3. **SELinux labels.** The units mount with `:Z`. A restore must not leave
   `/data` labelled such that the container cannot read it — the failure looks
   like corruption and is not.
4. **Migration is not a volume swap.** The umbrella is explicit: *"named volumes
   are never presented as interchangeable."* Docker→Podman must be an
   export/import of contents, and the doc must say why rather than leaving a
   reader to try the fast wrong thing.

## What already carries over unchanged

Say so in the doc so the scope stays small: `pkg/spokebackup` and the owner-only
`GET /api/backup/status` / `POST /api/backup` endpoints run **inside** the hive
container and are runtime-agnostic. They work on a Quadlet deployment as-is.
Likewise `hive-backup run` remains inapplicable to a standalone deployment under
either runtime, for the reason already documented (no `clusters.json`, no
in-cluster kubectl).

## 30-minute boundary

Backup, restore, and migration for the standalone Quadlet deployment. No changes
to `pkg/hubbackup` or `pkg/spokebackup`. No changes to Docker instructions
beyond the minimum needed to say which section applies to which runtime.

## Deliverable

A Podman section in `src/docs/backup-restore.md` covering host-level backup,
restore, and Docker→Podman migration, in the house style of
`src/docs/podman-standalone-quadlet.md` — observed separated from expected.

## Acceptance criteria

- The commands are given for **both root modes** and name the volume correctly
  in each, or state which mode was exercised and which was not.
- Restore is demonstrated into a **fresh** volume, and the container is shown to
  come back healthy afterwards — not merely that the tar extracted. `hive-id`
  and the GitHub App key surviving is the thing being proven.
- Ownership and SELinux label state after restore are recorded as observed
  values, not asserted.
- The migration procedure states plainly that the Docker and Podman volumes are
  never shared, and does not present `podman volume` pointed at Docker storage
  as an option.
- `docker compose down -v` has a named Podman counterpart, and the destructive
  one is marked as destructive.

## Stop conditions

If the VM cannot complete a full backup → wipe → restore cycle, write the
procedure and record plainly which steps were executed and which were not. Do
not present an extract that was never mounted as a restore.

## Related

- #4376 — persistence and SELinux volume behaviour
- #4377 — stop/start/recreate, and the first live rootful start
- #4404 — cross-runtime parity checks

Danathar:
Closes #4406. Does not close the parent (#4188).

`src/docs/backup-restore.md` contained **zero** occurrences of "podman". Its standalone section was titled "Docker Compose" and every command in it was a Docker command, so an operator on a Quadlet deployment had nothing to follow on the day it matters. #4376 established what `hive-data` guarantees and #4377 proved it survives a recreate; neither says how to get the data *out* and back *in*.

Doc-only. No changes to `pkg/hubbackup` or `pkg/spokebackup`; the Docker instructions are untouched apart from a runtime selector above them.

## Not a search-and-replace of `docker` for `podman`

Four things differ, and each of them is a way to lose data quietly. All four are stated with the output that shows it.

**The volume has a different name.** Compose prefixes named volumes with the project directory — hence `src_hive-data`. `hive-data.volume` sets `VolumeName=hive-data` with no prefix. Following the Docker commands under Podman restores into a volume nothing mounts, and there is no error to notice: `podman volume create src_hive-data` succeeds, the generated unit still carries `-v hive-data:/data`, and `hive.service` then starts on the empty `hive-data` — cleanly and healthy, per #4377.

**Rootless ownership is mapped, and a host-shell `tar` skips the private key.** With `/etc/subuid` at `dbaggett:524288:65536`, container UID 1001 is host UID 525288. The key is `0600` owned by 525288 and the operator is 1000, so:

```
$ tar czf hive-data.tar.gz -C "$VOLUME_PATH" .
tar: ./gh-app-key-restore-probe.pem: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors
```

`tar` exits non-zero **and still writes a tarball**. An operator who does not check the exit status has a backup missing the one file the backup exists to protect. Three namespace-aware forms are given instead — `podman volume export`, a container, `podman unshare` — each measured to record `1001/1000` and to read the key.

**Recreating the volume by hand orphans it.** `podman volume create hive-data` gives `labels={}`, and `bin/hive-podman-teardown.sh plan` then reports "no Hive-owned volumes": the restored volume survives every teardown, invisibly. Starting `hive-data-volume.service` gives the four #4210 labels.

**Migration is not a volume swap.** `podman volume inspect src_hive-data` → `no such volume`, in **both** root modes. Bind-mounting Docker's directory fails with `statfs …: permission denied` before it can do damage.

## Executed

Fedora 44, Podman 5.8.4, systemd 259, SELinux **enforcing**, Docker 29.7.2, real `:stable` image. Three separate deployments, each with its own identity, so a restore returning the wrong data would be visible rather than plausible.

**Backup → destroy the volume → recreate → restore → start, in both root modes:**

| | rootless | rootful |
| --- | --- | --- |
| identity before | `hive-pure-fox` | `hive-neat-ant` |
| after restore, `systemctl start` | rc 0 in **10s**, `active/running`, container `healthy` | rc 0 in **10s**, `active/running`, container `healthy` |
| `hive-id` after restore | `hive-pure-fox` — unchanged | `hive-neat-ant` — unchanged |
| GitHub App key SHA-256 | unchanged | unchanged |
| `beads/` ledger | unchanged | unchanged |

Healthy, not merely extracted — `Notify=healthy` means the start returned because `/api/health` answered *on the restored data*.

**Docker → Podman migration:** a live Compose deployment (`hive-calm-colt`, volume `src_hive-data`) exported through a container, the Podman volume wiped, the contents imported, and `hive-calm-colt` then running **healthy under rootless Podman** with the key byte-identical. `docker compose down -v` only afterwards.

## Observed, not asserted

Ownership and SELinux state after restore are given as measured values. The type is `container_file_t` with **no MCS category** in every row — the property #4376 says a recreated container depends on — and only the SELinux *user* field differs (`unconfined_u` after a host-driven `volume import`, `system_u` after a container-driven extract). Neither affects the access decision, and the doc says not to "fix" it with `chcon` and above all not to add `:Z` to the volume line.

Two things I could easily have asserted wrongly and measured instead: the volume root lands on `0:0` after `podman volume import` and is repaired by the entrypoint at the next start; and `hive-state.json`'s hash *does* change across a restore, because the running service rewrites it at boot — so `hive-id` and the key files are the witness, not a checksum of the volume.

## Also in here

- `docker compose down -v` gets its named counterpart, `bin/hive-podman-teardown.sh run --yes`, marked destructive and paired with the `plan` mode Docker has no equivalent of, plus the by-hand three-command form.
- What carries over **unchanged** is stated so the scope stays small: `pkg/spokebackup`, the owner-only endpoints and the key handling run inside the container and are runtime-agnostic; `hive-backup run` remains inapplicable for the reason already documented.
- The config/secrets archive, including the `podman unshare chown -R 0:1002` step a plain extract does not reproduce, and why `hive.env` keeps an ordinary host label where the `:Z` bind mounts carry a private MCS category pair.
- A short runtime selector above the Docker section so each half says which deployment it applies to.

## Not executed, recorded as such

A "Not executed" table covers: restore onto a host with a different `/etc/subuid` base; restore of an archive from an older Hive image (a schema question, not a volume one); the config/secrets archive as a full round-trip; and that the GitHub App key was a **synthetic** `gh-app-key-*.pem` planted for the test, so what is demonstrated is that a `0600` file owned by `dev` in `/data` round-trips byte-identically.

One caveat worth a reviewer's eye: **this host has Docker installed**, which is what made the migration section executable rather than theoretical, but it means #4188's *"clean supported Linux host with Podman and no Docker installation"* was **not** demonstrated here. The Podman procedures use no Docker component; that a Docker-free host runs them was not proven on this host. Recorded in the doc rather than glossed.


Danathar:
The one red check, `quadlet --dryrun, rootful and --user`, is pre-existing and cannot be this PR — this branch changes two markdown files and nothing else. `.github/workflows/quadlet-gate.yml` pins `quay.io/podman/stable@sha256:312a77be…`, and that digest no longer resolves:

```
Error: unable to copy from source docker://quay.io/podman/stable@sha256:312a77be…:
reading manifest sha256:312a77be… in quay.io/podman/stable: manifest unknown
```

It fails at the "Install the Quadlet generator" step, before any repository file is read, and it is failing the same way on other open branches. Same note as on #4407, where I first hit it; refreshing that pin wants its own change.

clubanderson:
Reviewed from head 81b98307 (merge-forward of v4 over 3876d9a3 — the fast-forward only picks up the quadlet-gate image-pin refresh from #4407; the PR delta itself is unchanged, doc-only: +265 lines in backup-restore.md plus the index blurb).

Fact-checked the doc against the tree and it holds up:
- **Volume name / mounts**: `VolumeName=hive-data` (hive-data.volume:36), `Volume=hive-data.volume:/data`, the `%E/hive` binds and `EnvironmentFile=` all match hive.container/hive-gateway.container exactly, including the hive.env host-label exception (read by systemd, never bind-mounted).
- **Secrets re-apply step** (`podman unshare chown -R 0:1002` / `chgrp -R 1002`) matches the install doc verbatim, and the hive-launch GID 1002 rationale is #4357's.
- **Restore-through-the-unit guidance** is exactly the right call: a hand-created volume carries no #4210 labels and becomes invisible to the teardown tooling forever. The doc measuring that (`plan` sees it vs not) is the strongest kind of warning.
- **The rootless host-shell tar trap** (non-zero exit *and* a tarball written, missing the 0600 key) is the standout finding — that failure mode is precisely the one an operator would not notice until restore day.
- **SELinux guidance** is consistent with the #4383/#4353 contracts: `container_file_t` no-MCS on the volume, do not chcon, never `:Z` the volume line; the user-field difference (`unconfined_u` vs `system_u`) correctly identified as irrelevant to the access decision.
- **Port note** (`dashboard.port` must be 3002 to match HealthCmd) matches the #4367 contract.
- Cleanup contract 127/127 on the merged tree; the destroyer discussion stays clear of the forbidden literals. Doc-only, so no suite wiring applies. Honest not-executed table including the Docker-installed-host caveat — appreciated as always.

/lgtm
/approve

PR #4414, opened 2026-08-21
test(deploy): boot the manifest's actual capability set (#4379)

### LINKED ISSUE #4379: 🧪 Nothing boots the manifest's actual capability set — the shipped drop/add list is never exercised at runtime
## Summary

Nothing exercises the Kubernetes manifest's actual capability set at runtime.
`src/deploy/k8s/deployment.yaml` grants a precise list:

```yaml
capabilities:
  drop: [ALL]
  add:
    - NET_ADMIN     # ACMM proxy iptables redirect of outbound :443
    - SETUID        # su-exec dev->per-agent UID switch
    - SETGID
    - SETPCAP
    - CHOWN         # chown per-agent dirs/beads to the agent UID
    - DAC_OVERRIDE
```

The only job that boots a container to check capability behavior,
`src/deploy/test_ambient_cap_runtime.sh`, does this instead:

```sh
docker run --rm $caps_arg "$IMAGE" ...        # caps_arg = --cap-add=NET_ADMIN
```

That is Docker's **default** bounding set plus `NET_ADMIN` — a much wider set
than the manifest grants, and it never contains a `--cap-drop` at all. So the
configuration that actually ships to a cluster is the one configuration never
booted.

Everything else in this area is static: `test_security_hardening.sh` and
`check-suid-contract.sh` are greps over files. Greps cannot catch a capability
that the kernel refuses at `execve` time.

## Why it matters, concretely

This was found by @clubanderson while reviewing #3895, which proposed dropping
`SETUID`/`SETGID`. That change would have broken the pod at runtime in two
places:

1. The entrypoint starts as root and drops to `dev` via `setpriv --reuid dev`
   (or the `gosu` fallback). Both need `CAP_SETUID`/`CAP_SETGID` in the
   process's permitted set, and for the initial root process permitted == the
   bounding set.
2. Per `capabilities(7)`, a setuid-root `exec`'s new permitted set is
   `(bounding | inheritable)`. With those two out of the bounding set,
   `su-exec`'s `setgroups()/setgid()/setuid()` to the per-agent UID fail with
   `EPERM` — the classic reason `--cap-drop SETUID` breaks `su`/`gosu`/`su-exec`.

**CI was fully green on that PR.** It was caught by review, not by a test. The
same class of change — anyone tightening the capability list in the manifest, or
in `saas_provision.go`'s hosted-hive template — would be green again today.

## Proposed

A runtime job that boots the real image with the **manifest's exact**
`--cap-drop`/`--cap-add` set and proves the privilege chain end to end:

1. the container starts and the entrypoint completes the `setpriv`/`gosu` drop
   to `dev`;
2. a process is launched through `su-exec` to a per-agent UID (2001+) and
   actually runs;
3. the proxy's egress gate still installs, so the capability set is sufficient
   for the ACMM path as well as the UID switch.

## Acceptance criteria

- The cap set under test is **derived from `deployment.yaml`**, not duplicated
  into the test — a test carrying its own copy of the list would pass while the
  manifest drifted, which is the failure mode being fixed.
- Dropping any one of the six from the tested set makes the job fail, with the
  failure naming which capability was missing. Verify by mutation.
- The job asserts the `su-exec` transition to a non-`dev` UID succeeds, not just
  that the container started.
- `saas_provision.go`'s hosted template is covered by the same derivation, or
  the issue records why it is not.
- No change to the shipped capability list in this issue — this adds coverage
  for what exists.

## Non-goals

Re-litigating #3895's capability drop. That PR is closed and held for a reason.
This issue is the prerequisite that would let anyone revisit it safely; it is
not a proposal to revisit it.

## Refs

#3895 (review that found this), #4341 (the CA-key half that shipped separately).

Danathar:
Closes #4379.

## The gap

`src/deploy/k8s/deployment.yaml` ships an exact grant — `drop: [ALL]` plus six adds — and nothing ever booted it. The only runtime job in this area, `test_ambient_cap_runtime.sh`, runs `docker run --cap-add=NET_ADMIN`: Docker's **default** bounding set plus NET_ADMIN, with no `--cap-drop` anywhere. Everything else (`check-suid-contract.sh`, the hardening checks) is a grep over files, and a grep cannot catch a capability the kernel refuses at `execve`.

The concrete miss is the one in the issue: #3895 proposed dropping `SETUID`/`SETGID`, which breaks the pod in two places — the entrypoint's root→`dev` `setpriv --reuid` drop, and `su-exec`'s `setgroups()/setgid()/setuid()` to the per-agent UID. CI was fully green. Review caught it, not a test.

## What this adds

`src/deploy/test_manifest_caps_runtime.sh`, wired into `suid-contract.yml` as a third job.

**Derivation, not duplication.** The cap set is parsed out of `deployment.yaml` at run time; the script keeps no copy of its own. The `openshift-netadmin` SCC's `allowedCapabilities` (where drift does not fail a grep — it fails *admission*, on the cluster, at deploy) and the hosted template in `saas_provision.go` are checked against that same derivation. The container fixture is derived too: same runtime base, su-exec commit + SHA256 pin, `hive-launch` GID and `dev` identity as `src/Dockerfile`, and the built image's su-exec mode/owner is asserted against the Dockerfile's own `chmod`/`chown` before anything relies on it. The `setpriv` flags come out of `entrypoint.sh` the same way `test_ambient_cap_runtime.sh` extracts them.

**The boot.** `--cap-drop=ALL --cap-add=<derived>`, then each capability's actual consumer:

```
STEP capbnd-is-exactly-the-manifest-grant PASS want=cap_chown,cap_dac_override,cap_net_admin,cap_setgid,cap_setpcap,cap_setuid got=…
STEP canary-mknod-is-dropped                PASS
STEP chown-agent-dir-to-agent-uid           PASS  needs CHOWN; owner=2001
STEP dac-override-write-through-mode-000    PASS  needs DAC_OVERRIDE
STEP egress-gate-redirect-installs          PASS  needs NET_ADMIN
STEP entrypoint-setpriv-drop-to-dev         PASS  uid=1001 gid=1000 groups=node+hive-launch
STEP dropped-dev-keeps-launcher-group       PASS
STEP ambient-net-admin-survives-drop        PASS  CapAmb=0000000000001000
STEP gosu-fallback-drop-to-dev              PASS
STEP su-exec-dev-to-agent-uid               PASS  2001
STEP su-exec-agent-process-actually-runs    PASS  file owner=2001
```

`CapBnd` must be **exactly** the manifest grant, with a named canary (`MKNOD`, from the runtime's default set) asserted absent — so `drop: ALL` having silently not applied cannot pass. The su-exec step checks the agent process *runs*, not just that it started: it writes a file and the kernel-recorded owner must be 2001.

**#3895 in isolation.** In the main run, removing SETUID kills the root→dev drop first, so the su-exec assertions fail as collateral rather than on their own merits. Three extra runs start the container **as `dev`** (`--user dev`), so reaching `dev` needs no capability and the SUID hop is the only thing under test:

```
  PASS: dev can su-exec to the per-agent UID under the manifest's exact capability set
  PASS: removing SETUID alone breaks the su-exec hop to the agent UID (the #3895 regression, reproduced in-test)
  PASS: removing SETGID alone breaks the su-exec hop to the agent UID (the #3895 regression, reproduced in-test)
```

with `su-exec: setuid(2001): Operation not permitted` in the log.

**Mutation, in CI, every run** — not verified by hand once (acceptance criterion 2):

```
     NET_ADMIN removed    -> failing steps: egress-gate-redirect-installs entrypoint-setpriv-drop-to-dev …
  PASS: dropping NET_ADMIN breaks the privilege chain, and the failure names NET_ADMIN
     SETUID removed       -> failing steps: entrypoint-setpriv-drop-to-dev … su-exec-dev-to-agent-uid …
  PASS: dropping SETUID breaks the privilege chain, and the failure names SETUID
     SETGID removed       -> …  PASS
     CHOWN removed        -> failing steps: chown-agent-dir-to-agent-uid …  PASS
     DAC_OVERRIDE removed -> failing steps: dac-override-write-through-mode-000  PASS
```

The probe reports `MISSING_FROM_BOUNDING=<cap>` computed from `CapBnd`, so the failure names the capability to restore rather than leaving an operator to guess.

**No silent gaps.** Any granted capability whose removal breaks nothing is reported as UNEXERCISED and **fails the job**, unless it is in a documented allowlist with a kernel-level reason. A capability added to the manifest without coverage turns this job RED instead of quietly widening the grant.

## Two things this records rather than acts on

**SETPCAP is currently unexercised, and is in the allowlist.** Raising NET_ADMIN into the inheritable set — the #3874 fix, and the only pI write in the image — does not need it. `cap_capset()` in `security/commoncap.c` admits a new pI that is a subset of `(old pI | pP)` and of `(old pI | bounding)`; the container's initial process is UID 0, so the granted NET_ADMIN is already in pP and the raise is legal without `CAP_SETPCAP`. Verified by boot: the full chain passes with SETPCAP removed. The `CAP_SETPCAP` requirement for pI writes is pre-2.6.25 kernel history, which is presumably why it is in the list.

Per the issue's non-goals this is **not** a proposal to drop it — the list is unchanged here. It is exactly the kind of fact the issue wanted established before anyone revisits the grant, and it is now machine-checked rather than assumed: if a future change gives SETPCAP a real consumer, remove it from the allowlist and the job enforces it.

**`saas_provision.go` is covered by containment** (acceptance criterion 4). The hosted template adds capabilities but does **not** drop ALL, so its bounding set is the runtime default set ∪ its add list — a strict superset of the set booted here, since the default set already contains CHOWN, DAC_OVERRIDE, SETGID, SETUID and SETPCAP. The manifest boot is the tighter case and passing it implies the hosted one. Both halves of that argument are asserted rather than assumed: a `drop:` appearing there, or an added capability the k8s manifest does not have, breaks containment and fails with an instruction to give the template its own boot. A comment at the template says so, including *don't "fix" it by copying the six capabilities in* — that would narrow the hosted bounding set.

## Also fixed

`suid-contract.yml`'s path filters did not include `test_ambient_cap_runtime.sh`, so a change to that script did not run its own job. Added, along with the manifest, the SCC overlay and `saas_provision.go` (the new job's derivation inputs).

## Scope

The same statement `test_ambient_cap_runtime.sh` already makes: docker's seccomp/apparmor defaults are not the pod's `RuntimeDefault`, so this validates the **capability** contract only, not the full pod security context. The iptables step self-detects a host where `nat` is unusable in a container even *with* NET_ADMIN and omits itself with a loud notice rather than reporting an environment problem as a capability regression; NET_ADMIN stays covered by the ambient step.

No change to the shipped capability list — this adds coverage for what exists.

## Verification

Locally, on the branch: **19 passed, 0 failed**, exit 0.

Verified RED in three ways, not just green:
- removed `SETUID` from `deployment.yaml` → 4 failures, naming SETUID, the SCC 1:1 mismatch, and the #3895 citation;
- emptied `UNEXERCISED_ALLOWLIST` → the SETPCAP finding fails the job as designed;
- every mutation run's log inspected to confirm the intended steps fail for the intended reason (e.g. `setpriv: setresuid failed: Operation not permitted`, `su-exec: setuid(2001): Operation not permitted`).

`shellcheck` clean at default level. `check-suid-contract.sh` and `test_ambient_cap_runtime.sh` both still pass unchanged. The `saas_provision.go` edit is comment-only inside the YAML raw string (no backticks — checked, it would have terminated the literal) and the hub tests that touch `k8sManifestTemplate` are all `strings.Contains` assertions on unrelated fields.

### DIFF
--- a/.github/workflows/suid-contract.yml
+++ b/.github/workflows/suid-contract.yml
@@ -7,6 +7,9 @@ name: SUID + NET_ADMIN Contract
 #  2. #3760: the hive binary carries NO file capability (a file cap EPERMs on
 #     exec wherever the bounding set lacks NET_ADMIN); NET_ADMIN is instead
 #     raised by the entrypoint as a bounding-set-gated AMBIENT capability.
+#
+# plus two runtime jobs that boot a container and prove what a grep cannot:
+# ambient-cap-runtime (#3874) and manifest-cap-runtime (#4379).
 
 on:
   push:
@@ -17,6 +20,13 @@ on:
       - 'src/Dockerfile'
       - 'src/deploy/entrypoint.sh'
       - 'src/scripts/check-suid-contract.sh'
+      - 'src/deploy/test_ambient_cap_runtime.sh'
+      # #4379: the manifest capability set is DERIVED from these, so a change to
+      # any of them changes what the runtime job boots.
+      - 'src/deploy/k8s/deployment.yaml'
+      - 'src/deploy/kustomize/overlays/openshift-netadmin/scc-hive-netadmin.yaml'
+      - 'src/pkg/hub/saas_provision.go'
+      - 'src/deploy/test_manifest_caps_runtime.sh'
       - '.github/workflows/suid-contract.yml'
   pull_request:
     branches:
@@ -26,6 +36,13 @@ on:
       - 'src/Dockerfile'
       - 'src/deploy/entrypoint.sh'
       - 'src/scripts/check-suid-contract.sh'
+      - 'src/deploy/test_ambient_cap_runtime.sh'
+      # #4379: the manifest capability set is DERIVED from these, so a change to
+      # any of them changes what the runtime job boots.
+      - 'src/deploy/k8s/deployment.yaml'
+      - 'src/deploy/kustomize/overlays/openshift-netadmin/scc-hive-netadmin.yaml'
+      - 'src/pkg/hub/saas_provision.go'
+      - 'src/deploy/test_manifest_caps_runtime.sh'
       - '.github/workflows/suid-contract.yml'
   workflow_dispatch:
 
@@ -54,3 +71,23 @@ jobs:
 
       - name: Assert ambient CAP_NET_ADMIN survives the privilege drop
         run: bash src/deploy/test_ambient_cap_runtime.sh
+
+  # RUNTIME assertion (#4379). The job above boots with docker's DEFAULT bounding
+  # set plus NET_ADMIN — a strictly wider set than the pod gets. This one boots
+  # the capability set src/deploy/k8s/deployment.yaml actually ships
+  # (`--cap-drop=ALL` plus its exact add list, derived from the manifest at run
+  # time, never restated) and drives the whole privilege chain through it: the
+  # entrypoint's root->dev setpriv drop, the gosu fallback, the SUID su-exec hop
+  # to a per-agent UID, the per-agent chown, a DAC_OVERRIDE write and the ACMM
+  # iptables egress gate. It then removes each capability in turn and requires the
+  # chain to break, so none of those assertions can pass for the wrong reason.
+  #
+  # This is the coverage #3895 did not have: that PR proposed dropping
+  # SETUID/SETGID, which breaks agent launch in the pod, and CI stayed green.
+  manifest-cap-runtime:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Boot the manifest's exact capability set and exercise it
+        run: bash src/deploy/test_manifest_caps_runtime.sh
--- a/src/deploy/k8s/deployment.yaml
+++ b/src/deploy/k8s/deployment.yaml
@@ -70,6 +70,14 @@ spec:
           securityContext:
             seccompProfile:
               type: RuntimeDefault
+            # COVERAGE (#4379): this exact drop/add list is BOOTED by
+            # ../test_manifest_caps_runtime.sh. That job derives the set from
+            # this file (it keeps no copy of its own), runs the whole privilege
+            # chain under it — entrypoint root->dev drop, su-exec hop to a
+            # per-agent UID, per-agent chown, ACMM egress gate — and then removes
+            # each capability in turn to prove every one is load-bearing.
+            # Tightening the list here while that job stays green is what #3895
+            # did; the pod broke at runtime and CI never noticed.
             capabilities:
               drop:
                 - ALL
--- a/src/deploy/kustomize/overlays/openshift-netadmin/scc-hive-netadmin.yaml
+++ b/src/deploy/kustomize/overlays/openshift-netadmin/scc-hive-netadmin.yaml
@@ -13,6 +13,10 @@
 #                              cap is in this list — allowing only NET_ADMIN
 #                              would still reject the pod, so the su-exec set
 #                              is listed too, mirroring deployment.yaml 1:1.
+#                              That 1:1 is machine-checked by
+#                              ../../test_manifest_caps_runtime.sh (#4379):
+#                              drift here does not fail a grep, it fails
+#                              admission, on the cluster, at deploy time.
 #   runAsUser: RunAsAny        The entrypoint starts as root (iptables REDIRECT,
 #                              PVC chown) and then drops to `dev`;
 #                              MustRunAsRange would force a namespace-range UID
--- a/src/deploy/test_manifest_caps_runtime.sh
+++ b/src/deploy/test_manifest_caps_runtime.sh
@@ -1,0 +1,680 @@
+#!/usr/bin/env bash
+# test_manifest_caps_runtime.sh — RUNTIME regression test for #4379.
+#
+# THE GAP THIS CLOSES
+# -------------------
+# `src/deploy/k8s/deployment.yaml` ships an exact capability grant:
+#
+#     capabilities:
+#       drop: [ALL]
+#       add:  [NET_ADMIN, SETUID, SETGID, SETPCAP, CHOWN, DAC_OVERRIDE]
+#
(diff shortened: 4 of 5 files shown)


clubanderson:
Reviewed from head c52c69bb. This closes exactly the #3895 gap: nothing ever booted the manifest's drop-ALL grant, and a grep cannot see what the kernel refuses at execve.

- **Derivation verified end to end**: every fixture parameter's extraction anchor checked against the tree — `_SETPRIV_CAPS`/`_SETPRIV_ID` (entrypoint.sh:1072/1092), `HIVE_UID_BASE=2001` (:623), the agent useradd (:708), su-exec commit/SHA pins, 4750 root:hive-launch, GID 1002, and the dev useradd line in src/Dockerfile. The script keeps no copy of the cap list; SCC 1:1 and the saas_provision.go containment argument (add-subset + no-drop, both asserted) are correct — and the warning against "fixing" containment by copying the six caps into the hosted template is exactly right, that would narrow the hosted bounding set.
- **The boot is the right shape**: CapBnd asserted *exactly* equal to the grant plus a named canary from the default set, so a silently-unapplied drop:ALL cannot pass. The `--user dev` isolation runs are the key design move — they decouple the su-exec SUID hop from the setpriv drop so SETUID/SETGID removal is tested on its own merits, which is precisely the #3895 regression. Mutation-per-capability with MISSING_FROM_BOUNDING naming the victim, and the UNEXERCISED-fails-the-job rule (SETPCAP allowlisted with the correct post-2.6.25 `cap_capset()` reasoning) keeps the grant from silently widening.
- **Non-test changes are comment-only** (manifest, SCC, saas_provision.go — verified no backticks in the raw string, `go build ./pkg/hub/` clean) plus the path-filter fix, which also repairs the pre-existing gap where test_ambient_cap_runtime.sh didn't trigger its own job.
- Locally: derivation half 6/6 (docker unavailable here — runtime half CI-authoritative, and the manifest-cap-runtime job passed in 1m20s, so the boot + mutations really ran). shellcheck clean, cleanup contract 127/127, trial merge with v4 clean.

/lgtm
/approve

Issue #4416, opened 2026-08-21
🧹 Compose gateway mounts a gateway-upstream volume at /etc/nginx/upstream that nothing reads


Danathar:
## Summary

`src/docker-compose.yaml` gives the gateway a named volume that nothing reads:

```yaml
    volumes:
      - ./deploy/nginx.conf:/etc/nginx/nginx.conf:ro
      - gateway-upstream:/etc/nginx/upstream
```

plus the matching top-level `gateway-upstream:` declaration. The Quadlet gateway
(`src/deploy/quadlet/hive-gateway.container`) has no equivalent, and appears to
be the correct one.

Surfaced by the cross-runtime parity check added for #4404
(`src/deploy/test_standalone_runtime_parity.sh`), which records it as a
`defect-tracked` exception. #4404 says explicitly not to delete it as part of
that issue, so this is the separate one it asks for.

## Evidence that nothing reads it

`/etc/nginx/upstream` and `gateway-upstream` appear in exactly two places in the
whole repository, and both of them are `src/docker-compose.yaml` itself (the
mount line and the volume declaration). In particular:

- `src/deploy/nginx.conf` declares its upstream **inline**:

  ```nginx
  upstream hive_api {
  ```

  and its own comment states it is the entire configuration — it is mounted at
  `/etc/nginx/nginx.conf` and carries no `include /etc/nginx/conf.d/*.conf`,
  let alone an include of `/etc/nginx/upstream`.
- No Go code, script, doc, or Kubernetes manifest names the path.
- `src/deploy/blue-green-deploy.sh`, which recreates the stack, does not carry
  it either.

So the volume is created and mounted on every `docker compose up`, is written by
nothing, is read by nothing, and survives `docker compose down` (a named volume
is not removed without `-v`).

## Why it is worth closing out rather than leaving

It is small, but it is not free:

- It is a standing question for anyone comparing the two runtimes — the parity
  check has to carry an exception entry for it, and an exception with a
  "probably dead" reason is the kind that rots.
- A stale named volume on operator hosts outlives the deployment that made it.
- If it is *not* dead — if some deployment path outside this repository drops a
  file in there — then the Quadlet gateway is missing a mount and the Podman
  deployment is the broken one. That is the case worth ruling out explicitly
  rather than by absence of evidence.

## Suggested

Confirm it is dead (git history for why it was added is the remaining input),
then remove both the mount and the `gateway-upstream:` top-level declaration,
[...]

Issue #4415, opened 2026-08-21
🐛 Quadlet gateway uses Restart=on-failure where Compose says unless-stopped — a clean exit leaves the only published port down


Danathar:
## Summary

`src/deploy/quadlet/hive-gateway.container` sets `Restart=on-failure`. The
Compose gateway it mirrors sets `restart: unless-stopped`, and the sibling unit
`hive.container` sets `Restart=always` — deliberately, with a measured
justification in its own body.

Found by the cross-runtime parity check added for #4404
(`src/deploy/test_standalone_runtime_parity.sh`), which records it as a
`defect-tracked` exception rather than failing the build. #4404's boundary is
the check, not the fixes.

## Why the two are not the same claim

`hive.container` explains the reasoning for its own policy:

> Restart=always, NOT on-failure, and the two lines above are why. Making 143 a
> success takes it out of on-failure's set, so with on-failure an external
> `podman rm -f hive` or `podman stop hive` would leave the unit
> `inactive/dead/success` — service down, nothing restarted, and nothing for
> monitoring to see. Measured (#4377). Restart=always closes that: systemd never
> restarts a unit that a `systemctl stop` job stopped, so a deliberate stop still
> stays stopped and clean, while an external removal is recovered after
> RestartSec.

That last sentence is also why `Restart=always` is the exact systemd equivalent
of Compose's `unless-stopped`, and `on-failure` is strictly weaker.

The gateway did not get the same treatment. It holds **the only published port
on the host**, so the asymmetry lands on the process an operator reaches the
dashboard through:

- nginx exits 0 on a clean `SIGTERM`/`SIGQUIT` shutdown. Under `on-failure`
  systemd records success and does not restart, leaving `hive.service` running
  and `:3001` refusing connections, with `systemctl status` green.
- Under Compose the same event restarts the container.

`hive.container` additionally carries `SuccessExitStatus=143 SIGTERM`; the
gateway carries none, so the two units disagree on which exits count as success
as well as on what to do about them.

## Suggested

Either bring the gateway to `Restart=always` (matching `unless-stopped` and the
sibling unit), or state in the unit why the gateway is the exception. Whichever
is chosen, the `restart:gateway` entry in
`src/deploy/test_standalone_runtime_parity.sh`'s exception table has to be
updated: the table is swept in reverse, so an entry that no longer matches a
real divergence fails the build. Fixing this makes that entry stale and CI will
say so.

## Verification

[...]

PR #4417, opened 2026-08-21
test(deploy): compare the Docker and Podman deployment contracts (#4404)

### LINKED ISSUE #4404: 🧪 Podman: add cross-runtime manifest/contract parity checks
## Parent

Part of #4188 (second wave, bullet 11). This issue must not close the parent.

## Why

The umbrella asked for this in as many words:

> Avoid maintaining two unrelated deployment descriptions by convention alone.
> [...] A change to one runtime must fail CI if the other runtime's contract was
> unintentionally left stale.

We now have both descriptions — `src/docker-compose.yaml` and
`src/deploy/quadlet/*` — and nothing compares them. The existing guard says so
itself, in its own header:

```
# This does NOT claim Docker/Podman parity. It snapshots one subset; further
# invariants belong in separate small issues.
#   -- src/deploy/test_standalone_service_contract.sh
```

Exactly one axis is actually guarded today: image references, through
`src/deploy/standalone-images.sh` and `test_standalone_image_refs.sh`. Every
other axis is convention.

## Drift that is already present

Read off the two files as they stand on `v4` — not hypothetical:

| Axis | `docker-compose.yaml` | `quadlet/*.container` |
|---|---|---|
| `/etc/hive/hive.yaml` | `./hive.yaml:/etc/hive/hive.yaml` — **read-write** | `:ro,Z` — **read-only** |
| Environment | six named vars (`HIVE_GITHUB_TOKEN`, `HIVE_DASHBOARD_TOKEN`, `NTFY_SERVER`, `NTFY_TOPIC`, `HIVE_LEVEL`, `HIVE_REPO`) | `EnvironmentFile=%E/hive/hive.env` — opaque, unchecked |
| Restart | `unless-stopped` (both services) | `always` (hive) / `on-failure` (gateway) |
| Gateway volume | `gateway-upstream:/etc/nginx/upstream` | absent |

The first row is a real difference in the security posture of the config file
and nothing reports it. The second means adding a seventh variable to the
Compose service reaches Docker and silently never reaches Podman. The fourth is
ambiguous in the other direction: no file in the repository references
`/etc/nginx/upstream` except `docker-compose.yaml` itself — including
`src/deploy/nginx.conf`, which never reads that path — so the Quadlet unit may
be correct and the Compose volume vestigial. **Do not delete it as part of this
issue**; record it and file it separately if it is dead.

## 30-minute boundary

The parity **check**, not the parity **fixes**. Divergences the check finds are
either encoded as deliberate, documented exceptions or reported in the issue for
separate follow-up. Do not change Docker behavior to make a check pass.

## Deliverable

A test in `src/deploy/` that reads both deployment descriptions, compares them
on a named list of runtime-invariant axes, and fails on unexplained divergence.
Wire it into an existing workflow rather than adding a new one.

Axes, at minimum — these are the umbrella's own list:

- image references, environment variables, mounts and their modes, published
  vs. exposed ports, capabilities, health checks, persistence, network exposure.

## Acceptance criteria

- Deliberate differences are expressed as exceptions **with a stated reason**,
  not by omitting the axis. The Watchtower label is the obvious one: it is
  Docker-only by design (#4188 Phase 3) and the check must say that rather than
  simply not looking at labels.
- Adding an environment variable to one description and not the other fails the
  check. Demonstrate this — a check that passes on the current tree proves
  nothing about what it would catch.
- The published-port boundary is covered in both directions: `3001` published,
  `7681` never published. This is the invariant #4375 measured; parity means it
  cannot regress in either asset.
- Runs without starting a container, like its sibling contract tests.
- The four divergences above are each resolved to `deliberate` or `defect`, and
  the defects get their own issues.

## Non-goals

- Docker/Podman *behavioural* equivalence at runtime. This is a static contract
  comparison; live behaviour is what the CI lanes (#4334, #4335) cover.
- The Kubernetes manifest. #4379 covers the capability set there separately.

Danathar:
Closes #4404. Part of #4188 (second wave, bullet 11) — does not close the parent.

## The gap

Both deployment descriptions exist and nothing compared them. Exactly one axis was guarded — image references, via `standalone-images.sh` + `test_standalone_image_refs.sh`. Every other axis was convention. The sibling guards each look at **one** runtime, and `test_standalone_service_contract.sh` says so in its own header:

> This does NOT claim Docker/Podman parity. It snapshots one subset; further invariants belong in separate small issues.

A property can hold in both of those and still differ between them.

## What this adds

`src/deploy/test_standalone_runtime_parity.sh` — a static comparison on the umbrella's own axis list (images, environment variables, mounts and modes, published vs exposed ports, capabilities, health checks, persistence, network exposure) plus restart policy and labels. Runs without starting a container, like its siblings. Wired into `v2-ci.yml` next to them rather than into a new workflow.

Two design choices worth flagging:

- **The Compose side is read as raw YAML, not through `docker compose config`.** Only names, ports, mounts, capabilities and probe fields are compared, none of which interpolation changes — and requiring Docker would make the parity guard skip on exactly the Podman-only hosts it exists to protect.
- **Image references are normalised by the source of truth's own function.** `standalone-images.sh`'s `hive_standalone_image_normalize` already knows `nginx:alpine` and `docker.io/library/nginx:alpine` are the same image; this sources it rather than carrying a second copy of the rules.

```
=== standalone Docker/Podman contract parity (#4404) ===
--- axis: image references ---      PASS ×2
--- axis: environment variables --- PASS: both runtimes declare the same 6 variables
--- axis: mounts and their modes ---
--- axis: published vs exposed ports ---
--- axis: capabilities ---
--- axis: health checks ---
--- axis: persistence ---           PASS: both runtimes call the state volume 'hive-data'
--- axis: network exposure ---
--- axis: restart policy ---
--- axis: labels ---
--- exception table: every entry still describes a real divergence ---
=== Results: 32 passed, 0 failed, 9 accepted exceptions ===
```

## Divergences are resolved, not omitted

Per the acceptance criteria, every difference is an exception **with a stated reason**, and the table distinguishes two claims that are not the same:

| kind | meaning |
|---|---|
| `deliberate` | the runtimes *should* differ here, and why |
| `defect-tracked` | they should not; here is the issue. A debt marker with a number on it, not a pass |

**Deliberate**, each asserted positively rather than skipped:

- **`/etc/hive/hive.yaml` mode** — `ro,Z` under Podman, `rw` under Compose. Not a break: the entrypoint's Docker-mode branch prints *"Config path is read-only — using `$HIVE_CONFIG_SOURCE` directly"* and falls back to `/data/hive.yaml.runtime`, the layer that survives a recreate anyway (#3961, pinned by `src/pkg/config/save_readonly_test.go`). Podman being **stricter** is accepted **in one direction only** — the check fails if the Podman side ever becomes the looser of the two.
- **Watchtower label** — Docker-only by design (#4188 Phase 3), exactly as the acceptance criteria call out. Asserted, not skipped: **no unit may carry it**, because a Quadlet unit advertising that label names a manager that will never manage it.
- **`io.kubestellar.hive.*` ownership labels** — Podman-only by design (#4210); `bin/hive-podman-teardown.sh` selects by them and nothing else.
- **`expose:`** — Compose lists 3001/3002/7681; the units carry no `ExposeHostPort=`. Both are documentation on a netavark network, but Quadlet's spelling is `--expose`, which becomes a *real* published port the moment anything adds `-P`. Asserted: **no unit may carry `ExposeHostPort=`**.
- **Gateway health start period** and **Compose's `build:` section** — structural; the probe command and interval/timeout/retries are still compared strictly.

**Defect-tracked**, both filed as the criteria require:

- **#4415** — the gateway unit says `Restart=on-failure` where Compose says `unless-stopped` and the sibling `hive.container` says `Restart=always`. nginx exits 0 on a clean SIGTERM, so under `on-failure` the **only published port on the host** goes away while `systemctl` reports success. `hive.container` learned this in #4377; the gateway did not get the same treatment.
- **#4416** — the vestigial `gateway-upstream:/etc/nginx/upstream` mount. Nothing in the repository reads that path: `nginx.conf` declares `upstream hive_api` inline and states it carries no `include`. Per #4404 it is **recorded and not deleted** — a comment now sits on the line in `docker-compose.yaml` saying so.

**Stale exceptions fail.** The table is swept in reverse: an entry that no longer matches a real divergence is itself a failure. Fixing #4415 forces its entry out, so the table cannot rot into a list of things that used to be true. (Demonstrated below.)

## The environment axis needed an artifact

`EnvironmentFile=` names a path and says nothing about what belongs in it, so a seventh variable added to Compose reached Docker and silently never reached Podman — with nothing on the Podman side to diff against. Without some enumerable artifact this axis is simply uncheckable, and the criteria forbid handling that by omitting the axis.

So this ships `src/deploy/quadlet/hive.env.example`: tracked, machine-readable, enumerating the same six variables, with each one's actual behaviour cited to the code that reads it. It changes no Docker behaviour and no runtime behaviour — the unit still reads `%E/hive/hive.env`. It also closes a real documentation gap: the install doc told operators to `touch` an empty file and named only **one** of the six.

## Verified by mutation

A check that passes on the current tree proves nothing about what it would catch. Each of these was applied and reverted; every one turns the job red:

| mutation | result |
|---|---|
| seventh variable in `docker-compose.yaml` only | ❌ `only in docker-compose.yaml: ['HIVE_SEVENTH_SETTING']` — **the criterion's own example** |
| variable in `hive.env.example` only | ❌ caught in the other direction too |
| drop `:ro` from the unit's `nginx.conf` mount | ❌ *"the Podman asset is the LOOSER of the two, which no exception permits"* |
| `PublishPort=7681:7681` on the gateway unit | ❌ 3 assertions, incl. the swept *"no service and no unit publishes the raw ttyd port 7681"* |
| remove `AddCapability=NET_ADMIN` | ❌ *"without NET_ADMIN the entrypoint fails closed with exit 77"* |
| **repair** the gateway restart policy | ❌ `stale: ['restart:gateway'] — the divergence is gone, so delete the entry` |

The published-port boundary is covered in both directions and across every service *and* every unit, not just the two components — `3001` published, `7681` never, the #4375 invariant asserted in both assets at once.

## Also

`shellcheck` clean. All six sibling contract tests still pass. One of them — `test_quadlet_config_contract.sh` (#4367) — caught `hive.env.example` while it was still untracked, since it asserts that install steps name files the repo actually ships. Its assertion did its job unprompted; `git add` was the fix.

Cross-links added from both sibling guards so the next person reading either header finds the parity check.

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -132,6 +132,21 @@ jobs:
       - name: Quadlet config-coupling contract (#4367)
         run: bash deploy/test_quadlet_config_contract.sh
 
+      # #4404: every guard above looks at ONE runtime. test_standalone_service_
+      # contract.sh parses only docker-compose.yaml and says so in its own header
+      # ("This does NOT claim Docker/Podman parity"); the two Quadlet guards parse
+      # only the units. A property can hold in both and still DIFFER between them,
+      # which is what #4188 asked for: "a change to one runtime must fail CI if
+      # the other runtime's contract was unintentionally left stale". This
+      # compares the two deployment descriptions directly — images, environment,
+      # mounts and modes, published vs exposed ports, capabilities, health checks,
+      # persistence, network exposure, restart policy, labels — and fails on any
+      # divergence not in its exception table, where each entry carries a stated
+      # reason. Entries are swept in reverse too: an exception that no longer
+      # matches a real divergence is itself a failure, so the table cannot rot.
+      - name: Docker/Podman contract parity (#4404)
+        run: bash deploy/test_standalone_runtime_parity.sh
+
       # bin/gh-wrapper.sh is the merge gate standing behind the proxy hard-deny:
       # if its argument parsing is wrong, a prompt-injected agent merges
       # unvetted code. It had NO tests and matched no CI path filter. This
--- a/src/deploy/quadlet/hive.env.example
+++ b/src/deploy/quadlet/hive.env.example
@@ -1,0 +1,53 @@
+# The environment contract for the Podman/Quadlet standalone deployment
+# (kubestellar/hive#4404). Install as %E/hive/hive.env — /etc/hive/hive.env for
+# a rootful unit, ~/.config/hive/hive.env for a --user one — and `chmod 600` it:
+# it holds tokens, and unlike the bind mounts it is read by systemd on the host,
+# so no container label protects it.
+#
+#   cp src/deploy/quadlet/hive.env.example "$CONF/hive.env"
+#   chmod 600 "$CONF/hive.env"
+#
+# THE FILE MUST EXIST, even if every line stays commented out. Quadlet turns
+# `EnvironmentFile=` into `podman run --env-file`, and `podman run` fails when
+# the file is missing. The leading `-` that makes an EnvironmentFile optional in
+# ordinary systemd units does NOT work here; see hive.container's header.
+#
+# WHY THIS FILE IS TRACKED IN THE REPOSITORY. `EnvironmentFile=` is opaque: the
+# unit names a path and nothing in the unit says which variables belong in it.
+# The Docker Compose stack spells its six variables out in `environment:`, so
+# adding a seventh there used to reach Docker and silently never reach Podman —
+# there was no Podman-side artifact to compare against. This file is that
+# artifact. src/deploy/test_standalone_runtime_parity.sh asserts that the names
+# here and the names in src/docker-compose.yaml's `hive` service are the SAME
+# SET, so a variable added to either side and not the other fails CI.
+#
+# Adding a variable: add it to both `environment:` in src/docker-compose.yaml
+# and to this file, commented out unless it has a safe default. Values here are
+# deliberately empty — this is a template, not a configuration.
+
+# The GitHub token Hive uses when no GitHub App is installed. Leave unset to run
+# App-only. `HIVE_GITHUB_TOKEN=` in Compose passes the host's value through; a
+# Quadlet unit has no host environment to inherit, so set it here explicitly.
+#HIVE_GITHUB_TOKEN=
+
+# REQUIRED unless this hive is hub-hosted (no HIVE_SESSION_KEY/HIVE_HUB_SECRET).
+# Hive REFUSES TO START without it — every mutation endpoint would be
+# unauthenticated. With Notify=healthy that refusal surfaces as a unit stuck in
+# `activating` until TimeoutStartSec expires, so set it before the first start:
+#   printf 'HIVE_DASHBOARD_TOKEN=%s\n' "$(openssl rand -hex 32)" >> hive.env
+#HIVE_DASHBOARD_TOKEN=
+
+# ntfy push notifications, read by bin/hive.sh. Both optional; set neither to
+# disable. NTFY_SERVER defaults to https://ntfy.sh when a topic is set.
+#NTFY_SERVER=
+#NTFY_TOPIC=
+
+# ACMM pack level (1-6) auto-applied on FIRST start only. On every later start
+# the persisted config wins and this is just the fallback — see the ApplyPack
+# block in src/cmd/hive/main.go. Normally left unset once hive.yaml has a level.
+#HIVE_LEVEL=
+
+# Bootstrap "org/repo". Fills project.org and project.repos only when the config
+# leaves them empty (applyBootstrapEnv in src/pkg/config/config.go); it never
+# overrides hive.yaml.
+#HIVE_REPO=
--- a/src/deploy/test_quadlet_port_boundary.sh
+++ b/src/deploy/test_quadlet_port_boundary.sh
@@ -12,6 +12,10 @@
 # the same set of invariants for src/deploy/quadlet/, so the two runtimes are
 # guarded rather than one being guarded and the other reviewed.
 #
+# Both of those check one runtime at a time, and a property can hold in each and
+# still DIFFER between them. src/deploy/test_standalone_runtime_parity.sh (#4404)
+# is the cross-runtime comparison; this file stays the Podman-side boundary.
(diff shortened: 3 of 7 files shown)


clubanderson:
Reviewed from head cd9aeb9c. The cross-runtime axis nobody guarded — verified the design and every factual claim:

- **Ran the suite locally: 32/0 with 9 accepted exceptions**, plus all four sibling contract suites still green. The reverse sweep (a stale exception is itself a failure) is what makes the exception table trustworthy — fixing #4415 forces its entry out.
- **Both defect-tracked exceptions point at real, open issues**: #4415 (gateway `Restart=on-failure` vs Compose `unless-stopped` — and hive.container already learned `Restart=always` in #4377, so the gateway genuinely missed the same treatment; a clean nginx exit takes down the only published port while systemctl reports success) and #4416 (the gateway-upstream mount — confirmed nginx.conf declares the upstream inline, nothing reads /etc/nginx/upstream).
- **The environment axis fix is the right artifact**: verified `hive.env.example`'s six names are exactly the compose `environment:` set, and each behavioural claim checks out against code (`HIVE_LEVEL` auto-apply at main.go:2609, `applyBootstrapEnv` in pkg/config/config.go, NTFY_* read by bin/hive.sh). The one-direction looseness rule (Podman may be stricter, never looser) is the correct asymmetry for the `ro,Z` vs `rw` config mount given the #3961 read-only fallback.
- **Deliberate exceptions are asserted positively, not skipped** — no unit may carry the watchtower label, no `ExposeHostPort=`, ownership labels Podman-only per #4210 — which keeps them from decaying into blind spots. Reading compose as raw YAML rather than via `docker compose config` is the right call for exactly the stated reason: the guard must run on Podman-only hosts.
- Wired into v2-ci.yml beside its siblings; shellcheck clean; cleanup contract 127/127; no forbidden literals; the install-doc `touch` → `cp` template upgrade closes a real gap (the doc previously named one of six variables).

/lgtm
/approve

Issue #2549, opened 2026-08-04
contributor: make Kubernetes a first-class deployment path, including cloud deploy entry points


castrojo:
## Status update — 2026-08-04

**Still blocked on #2538; no packaging decision should be inferred from today’s cluster work.** Merged #2557 creates a real workspace and better abandon behavior for the existing interactive tmux contributor, but it does not provide headless task delivery. There is still no useful Deployment/Job consumer for the generated ConfigMap and Secret.

The maintainer response keeps the ordering explicit: headless relay first, credential boundary (#2537) before or alongside any cluster artifact, then choose kustomize/chart and managed-cloud entry points with a working runtime in hand. A long-lived personal `GH_TOKEN` should not become normalized cluster furniture merely to make the current generator deployable.

> Part of #2534 (contributor onboarding field notes from `projectbluefin/review`).

## Summary

We'd like to ask for **Kubernetes to be a first-class, discoverable way to run a Hive
contributor** — reachable from `/contribute` the way containerized and host mode are today,
with a workload artifact a contributor can actually apply, and (where it makes sense)
one-click or guided entry points for the major managed Kubernetes offerings.

Right now the Kubernetes path exists as `just contribute-k8s`, which is not surfaced on
`/contribute` at all and which generates configuration without a workload. A contributor who
finds it gets a `Namespace`, a `ConfigMap` and a `Secret`, and then nothing runs.

**This issue is the packaging and discoverability layer that sits on top of #2538.** #2538
covers the headless-relay prerequisite — the relay drives a tmux pane and expects an
interactive TTY, so a workload manifest isn't useful until that's addressed. We are
deliberately **not re-litigating** that here; we agree with it, it comes first, and this
issue is about what you'd want to exist *once* it's solved: an artifact, a place to find it,
and an honest story about the credential.

We're a downstream consumer — `projectbluefin/review` launches contributor runtimes and our
users ask for cluster deployment regularly — so this is a feature request from the field.
All Hive citations pinned at `957771e5f18604f3d301a3749f63b9896ba99e1c`.

### `just contribute-k8s` generates config, not a workload

The recipe is
[`Justfile:669-755`](https://github.com/kubestellar/hive/blob/957771e5f18604f3d301a3749f63b9896ba99e1c/Justfile#L669-L755).
Its own doc comment describes it accurately: *"Generate K8s ConfigMap + Secret YAML from
contributor.env"*
[...]

PR #4418, opened 2026-08-21
feat(contribute): supported path to move a relay to another machine (#4408)

### LINKED ISSUE #4408: ✨ feature: no supported path to move a contributor relay to another machine
## Use case

I contribute to two hives with one relay, and I want to move that relay off my
desktop and into a VM — running the agent in a sandbox is safer than running it
on my host. Not two relays at once: one at a time, but on whichever machine I
choose, and able to switch back.

Nothing in hive prevents this. Contributor auth is a plain token-hash lookup
across profiles (`contribute_ws.go:2759-2772`) — there is no device binding, no
IP pinning, no session affinity. The same `contributor.env` authenticates from
any machine, and since only one relay runs at a time there is no concurrency
question either. **The capability is already there; the setup path is what's
missing.**

Today, moving to a second machine dead-ends:

**1. `contribute-setup` cannot bootstrap an already-registered identity.** On the
new machine, `POST /api/contribute/register` correctly refuses to hand back an
existing contributor's token — it is unauthenticated, so reissuing there would
let anyone claim someone else's seat (`api_contribute.go:6096-6106`). The
Justfile then falls back to reading a local `contributor.env`, which a fresh
machine does not have, and exits (`Justfile:324-331`):

```
ERROR: Already registered but no local config found.
```

That message names the problem and stops. It does not say what to do next, and
neither does `contributor-relay.md`.

**2. The one documented recovery rotates the credential.**
`reissueContributorToken` (`api_contribute.go:6239-6248`) overwrites the stored
hash, so a reissue from the VM invalidates the desktop. That is right for
recovering a lost token, but it makes switching back a full re-issue each time
rather than starting the relay somewhere else.

**3. Multi-hub makes it worse.** `contributor.env` holds positional
comma-separated `HIVE_HUB` / `HIVE_REGISTRATION_TOKEN` / `CONTRIBUTOR_ID` lists,
but the `contribute-setup` heredoc only ever writes a **single** hub
(`Justfile:337-343`). So for a two-hive contributor, `contribute-setup` on the
new machine is not just blocked — if it did succeed it would flatten a multi-hub
config down to one hub. Rotating instead means running reissue against each hub
and reassembling the positional lists by hand, keeping the order aligned or the
relay refuses to start.

**4. There is a second required file nobody mentions.** `contribute-hive` also
hard-requires `gh-auth.env` and exits with the same
`Not set up yet. Run: just contribute-setup <cli>` if it is missing
(`Justfile:399-402`) — pointing at the command that cannot run here.

## Proposed solution

**A supported "this identity, new machine" path.** Something like
`just contribute-move` / `contribute-setup --existing`, which does everything
`contribute-setup` does — backend preflight, `gh auth`, `gh-auth.env`, CLI config
staging — but instead of registering, either accepts an existing token or walks a
reissue, and writes a `contributor.env` that **preserves every hub entry** rather
than collapsing to one.

**Failing that, make the dead end actionable.** When `contribute-setup` finds the
identity registered and no local config, print the supported move instead of
`exit 1`: copy `${HOME}/.config/hive/contributor.env` and
`${HOME}/.config/hive/gh-auth.env` (mode 0600) from the old machine, or rotate
via `POST /api/contribute/reissue-token` per hub. Worth stating explicitly that
copying is the only way to *reuse* a token, since `TokenPlain` is cleared after
first read and can never be printed again.

**Document the move in `contributor-relay.md`.** There is no "move to another
machine" section at all; `reissue-token` appears only as a bare row in
`api-reference.md:211`. A short section covering both directions — copy to keep
the credential, reissue to rotate it — plus the multi-hub caveat, would have
answered this without reading the Justfile.

## Alternatives considered

- **Copy `contributor.env` + `gh-auth.env` by hand.** What actually works today,
  and what I am doing. Undocumented, and leaves a live credential on the machine
  I was trying to move off.
- **Reissue on every switch.** Correct but hostile for switching back and forth,
  and for a multi-hub contributor it is a manual per-hub rotation plus rebuilding
  positional lists.
- **Register a second GitHub identity for the VM.** Splits tier progression and
  leaderboard standing, and misrepresents one contributor as two.

## Scope and compatibility

- Target branch or version: v4
- Operator impact: none — no hub-side or policy change required; this is
  contributor tooling and docs.
- Contributor/agent impact: additive. Existing single-machine setups untouched.
- Migration or backwards-compatibility concerns: none for the docs/message
  changes. A `contribute-move` recipe must round-trip multi-hub
  `contributor.env` files without reordering the positional lists.

## Additional context

Related: #2610 (token reissue and `contributor.env` handling), #2543
(pre-registration environment verification).

Danathar:
Closes #4408.

## The problem

Nothing in hive prevents running a relay from another machine — auth is a plain token-hash lookup across profiles, with no device binding, no IP pinning, no session affinity, and only one relay runs at a time. **The capability was already there; the setup path was what was missing.** On a second machine the flow dead-ended at:

```
ERROR: Already registered but no local config found.
```

and neither that message nor `contributor-relay.md` said what to do next. The refusal itself is right — `/api/contribute/register` is unauthenticated and identifies by a self-asserted username, so handing back an existing contributor's token would be an account-takeover primitive. What was missing was the way round it.

## `just contribute-move`

Does everything `contribute-setup` does — backend preflight, `gh auth`, `gh-auth.env`, CLI config staging — but calls `POST /api/contribute/reissue-token`, which authenticates by GitHub token and therefore *can* prove identity.

```bash
export HIVE_HUB='wss://hive-a.example.com/contribute,wss://hive-b.example.com/contribute'
just contribute-move claude
```

Three things it gets right that a hand-rolled rotation does not:

- **It preserves every hub, in order.** `contributor.env` carries positional `HIVE_HUB` / `HIVE_REGISTRATION_TOKEN` / `CONTRIBUTOR_ID` lists that `bin/contributor-relay.sh` pairs *by index* — it refuses to start when the lengths disagree, and sends the wrong token to the wrong hive when the order is transposed. Rotating a two-hive identity by hand meant rebuilding three lists without making either mistake. Re-run with `HIVE_HUB` unset and it reuses the hub list already in the file; unmanaged keys (`HIVE_LITELLM_ENDPOINT`) are carried across rather than dropped.
- **A partial failure keeps what it got.** A reissued token can never be reprinted — the hub stores only a hash and clears the plaintext — so discarding tokens already issued because a *later* hub failed would lock the contributor out of the hives that succeeded. Every success is written, the failures are named, the exit status is red.
- **It asks before sending your GitHub token**, prints every host that will receive it, and refuses any non-loopback `http://` hub outright.

### On sending the GitHub token — please sanity-check this call

`contribute-setup` deliberately does **not** forward the contributor's PAT (H7/CWE-522), because it derives the hub URL from a public registry entry's `dashboardUrl` that a poisoned registry would control. `contribute-move` has to send it, because that is how `reissue-token` authenticates. I judged the risk materially different — here the URL comes from *the contributor's own* `HIVE_HUB` or their existing `contributor.env`, not from a registry lookup — and added two guards rather than relying on that alone: **TLS is required** for any non-loopback host, and **every receiving host is printed and confirmed** before anything is sent (`HIVE_MOVE_ASSUME_YES=1` for scripted runs). If you'd rather this refused to run without an explicit opt-in flag, say so and I'll tighten it.

## Two fixes the issue names alongside it

**The dead end is now actionable.** It prints both supported paths — copy `contributor.env` **and** `gh-auth.env` (0600) to keep the credential, or `contribute-move` to rotate it — and states that copying is the only way to *reuse* a token, since the plaintext is cleared after first read and can never be printed again. `gh-auth.env` is the second file `contribute-hive` hard-requires (issue item 4), and nothing mentioned it; `contribute-move` writes it.

**`contribute-setup` no longer flattens a multi-hub config.** Registering an *additional* hive used to `cat >` a single-hub file unconditionally — silently destroying a live token for the first hive that the hub can never reprint. It now appends to the existing lists, backs the previous file up either way, and refuses to append to a file whose lists are already misaligned (appending there would transpose every pair after the mismatch). This is the one change that touches the main onboarding path, so it has explicit regression coverage: a first-time registration still writes a plain single-hub file with no `.bak` and no append chatter.

## Verification

`src/deploy/test_contribute_move.sh` executes the **real recipes** against a fake hub, with a stubbed `gh` and a throwaway `HOME` — no network, no real credential, placeholder values only. It follows `test_contribute_k8s_workload.sh` (#2549) in shape and is wired next to it in `v2-ci.yml`.

Grepping the Justfile would prove none of this: that the positional lists come out aligned and in order is a property of the loop, not of any line in it, and "a later failure must not discard the tokens already reissued" is control flow.

**66 assertions, 0 failures**, covering: the fresh machine (both files written, 0600); multi-hub order and alignment; re-running with the hub list read from `contributor.env`; unmanaged keys surviving; the TLS refusal leaving the file untouched; partial failure (keeps the reissued token, names the failed hub, reports the hub's *own* reason, exits non-zero); total failure not clobbering `contributor.env`; the consent prompt including EOF on stdin; the reworded dead end; multi-hub append; first-time registration; and the misaligned-file case.

Two real bugs were caught by writing the test rather than by reading the code — a `while read` loop that dropped the last hub because `printf '%s'` emitted no trailing newline, and `read -p` at EOF killing the recipe under `set -e` before it could print why it stopped. Both fixed.

`shellcheck` clean. `test_contribute_k8s_workload.sh`, `bin/test_bin_suites_wired.sh` and `check-podman-contract.sh` (which parses the Justfile) all still pass.

## Scope

Contributor tooling and docs only — no hub-side handler, endpoint, or policy change, exactly as the issue's compatibility note says. `contributor-relay.md` gains a "Moving the relay to another machine" section covering both directions plus the multi-hub caveat and the "adding another hive is not a move" case; there was no section for this at all, and `reissue-token` appeared only as a bare row in `api-reference.md`. I left `api-reference.md` alone since it is a generated index.

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -322,6 +322,21 @@ jobs:
       - name: Contributor K8s Workload Generation
         working-directory: .
         run: bash src/deploy/test_contribute_k8s_workload.sh
+
+      # #4408: `just contribute-move` relocates an already-registered
+      # contributor identity to a new machine. The properties that matter are
+      # not visible in the Justfile text — that the positional HIVE_HUB /
+      # HIVE_REGISTRATION_TOKEN / CONTRIBUTOR_ID lists come out aligned and in
+      # the same order is a property of the loop, and bin/contributor-relay.sh
+      # pairs those lists by index (it refuses to start on a length mismatch and
+      # sends the wrong token to the wrong hive on a transposed one). Getting the
+      # partial-failure path wrong is worse: a reissued token cannot be
+      # reprinted, so dropping it locks the contributor out of that hive. This
+      # runs the real recipe against a fake hub with a stubbed gh and a
+      # throwaway HOME — no network, no real credential.
+      - name: Contributor identity relocation (#4408)
+        working-directory: .
+        run: bash src/deploy/test_contribute_move.sh
   create-issue-on-failure:
     if: failure() && github.event_name == 'push' && github.ref == 'refs/heads/v2'
     runs-on: ubuntu-latest
--- a/Justfile
+++ b/Justfile
@@ -326,7 +326,36 @@ contribute-setup backend="claude": check-version (contribute-check-backend backe
           source "{{config_dir}}/contributor.env"
           echo "Already registered — ${GH_USER} (${CONTRIBUTOR_ID:-unknown})"
         else
-          echo "ERROR: Already registered but no local config found."
+          # THE "SAME IDENTITY, NEW MACHINE" DEAD END (#4408). /api/contribute/
+          # register is unauthenticated, so it correctly refuses to hand back an
+          # existing contributor's token — otherwise POSTing any known username
+          # would be an account-takeover primitive. That refusal is right; what
+          # was missing is what to do next. This used to be a bare
+          # "Already registered but no local config found." and exit 1, with no
+          # supported path named here or in the docs.
+          echo "ERROR: ${GH_USER} is already registered on ${HUB_HTTP}, and this"
+          echo "       machine has no {{config_dir}}/contributor.env to reuse."
+          echo ""
+          echo "That is expected: register is unauthenticated, so it will never"
+          echo "hand an existing contributor's token to whoever asks. You are"
+          echo "moving an identity, not creating one. Two supported ways forward:"
+          echo ""
+          echo "  1. KEEP the credential — copy it from the old machine:"
+          echo "       scp old-machine:{{config_dir}}/contributor.env {{config_dir}}/"
+          echo "       scp old-machine:{{config_dir}}/gh-auth.env      {{config_dir}}/"
+          echo "       chmod 600 {{config_dir}}/contributor.env {{config_dir}}/gh-auth.env"
+          echo "     The hub stores only a HASH of the registration token and clears"
+          echo "     the plaintext after first read, so copying the file is the ONLY"
+          echo "     way to reuse the credential — it can never be printed again."
+          echo "     Delete the files on the old machine afterwards."
+          echo ""
+          echo "  2. ROTATE the credential — reissue it here:"
+          echo "       just contribute-move {{backend}}"
+          echo "     Proves identity with your GitHub token, reissues one token per"
+          echo "     hub, preserves EVERY hub entry in order, and writes gh-auth.env."
+          echo "     The old machine's relay stops working the moment this runs."
+          echo ""
+          echo "See src/docs/contributor-relay.md → 'Moving the relay to another machine'."
           exit 1
         fi
       else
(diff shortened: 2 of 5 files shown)


clubanderson:
Reviewed and verified.

- **Identity/key handling is correct and self-contained**: everything lives under the contributor machine's `~/.config/hive/` (`contributor.env`, `gh-auth.env`, both written 0600, backups also 0600). Nothing touches `/data`, so there is no interaction with — let alone contradiction of — the `/data/.hive` CA-key layout (#4341) or the Podman backup/restore surface (#4409); those are hub/spoke container concerns and this is contributor tooling only, exactly as the issue's compatibility note requires.
- **No secrets in logs or docs**: the recipe prints hosts and contributor IDs, never tokens; the GitHub token goes only over TLS-required (or loopback) connections after an explicit host-list confirmation, and reissued tokens land only in the 0600 files. Docs use placeholder hosts. The one deliberate H7/CWE-522 divergence from `contribute-setup` is well-argued: here the URL is the contributor's own `HIVE_HUB`/existing file rather than a poisonable registry lookup, and `reissue-token` authenticates by GitHub token by design — the TLS refusal + printed-and-confirmed host list are the right guards.
- **No half-registered state, hub record wins**: `reissue-token` is a pre-existing, GitHub-token-authenticated hub endpoint (dashboard `api_contribute.go:318`, with 401 coverage) that overwrites the stored hash atomically per hub — the old machine's relay simply stops authenticating, `CONTRIBUTOR_ID` is unchanged, and there is no spoke-asserted registration state for a stale machine to fight over. **Idempotent/partial-failure semantics are right**: every reissued token is written (they can never be reprinted), failures are named with the hub's own reason, exit is non-zero, re-run retries only the rest; total failure leaves `contributor.env` untouched; `.bak` kept on every rewrite.
- **The `contribute-setup` multi-hub append fix** closes a real credential-destroying bug (unconditional `cat >` flattening a live multi-hub file), with alignment guard against transposition and regression coverage that first-time registration is byte-identical in behavior.
- **Tests wired per #4365**: `test_contribute_move.sh` added to `v2-ci.yml` next to its sibling, and it **ran and passed on this PR's own CI** (run 32479946094, "Contributor identity relocation (#4408): success"). Locally at 94ef5ae: 66/66 assertions, shellcheck clean, `test_bin_suites_wired.sh` 16 suites/0 orphaned, `check-podman-contract.sh` still parses the Justfile clean. No forbidden podman lifecycle literals.
- Docs: paths (`~/.config/hive`), both-files requirement, rotation-vs-copy semantics, and the "adding a hive is not a move" case all consistent between the recipe output and `contributor-relay.md`.

/lgtm
/approve

PR #4420, opened 2026-08-21
feat(podman): opt-in health-aware auto-update, measured first (#4411)

### LINKED ISSUE #4411: 🧩 Podman: optional health-aware auto-update and rollback
## Parent

Part of #4188 (second wave, bullet 10). This issue must not close the parent.
#4378 shipped the manual path this builds on (#4407,
`src/docs/podman-quadlet-update-rollback.md`, `bin/hive-podman-update.sh`).

## Why this is not simply "add `AutoUpdate=registry`"

#4378 measured something that decides the design here. Quoting its result:

> the unit **NEVER READS `failed`**. `Restart=always` moves it from the timeout
> straight to `activating/auto-restart` and into another 300s attempt;
> `systemctl is-failed` reported `activating` at every point. Alerting keyed on
> `failed` does not fire on a bad update.

`podman auto-update --rollback` decides an update went bad by **restarting the
unit and seeing the restart fail**. On `hive.container` as shipped, a bad image
does not produce that signal: `Restart=always` converts the `TimeoutStartSec`
expiry into another start attempt rather than a failure. So the mechanism
auto-update rolls back *on* is the one thing #4378 showed this unit does not
produce.

That has to be resolved before auto-update can be called health-aware, not
discovered afterwards.

## The second conflict: pins

`AutoUpdate=registry` polls a **floating, fully-qualified tag**. #4378's
rollback works by pinning a **digest** in the `hive.container.d/10-image.conf`
drop-in. These two mechanisms want opposite things from the same `Image=` line.
Whatever ships has to say which wins, and what happens to a hand-placed pin when
auto-update fires.

## 30-minute boundary

Decide and document whether health-aware auto-update is achievable on this unit,
and if so ship it **opt-in**. Do not change Docker's Watchtower path. Do not
remove the manual path — the umbrella requires it to exist either way.

An honest "not achievable without changing X, here is the evidence" closes this
issue. A shipped `AutoUpdate=registry` that has not been driven through a bad
image does not.

## Acceptance criteria

- States what happens to `Restart=always` — kept, narrowed, or replaced — and
  what that costs. Removing it to expose failures also removes the crash
  recovery it was added for; the tradeoff must be named, not silently taken.
- A **bad-but-startable** image is the test case, as the umbrella requires:
  an image that runs and never becomes healthy, not one that fails to pull.
  #4378 has a recipe for staging exactly that.
- Auto-update's interaction with an existing digest pin is stated and tested.
- Whatever detects "bad" is shown firing. `systemctl is-failed` returning
  `activating` is the measured trap; a check that reads it is already known
  wrong.
- Auto-update stays **opt-in**, per #4188.
- Rollback preserves `hive-data` in both directions, proven with a marker file,
  the way #4378 proved it.

## Related

- #4378 / #4407 — the manual path, and the `failed`-state measurement above
- #4354 — the unit
- #4377 — lifecycle behaviour in both root modes

Danathar:
Closes #4411. Part of #4188 (second wave, bullet 10) — does not close the parent.

## The question, and the answer

#4378 measured that `hive.service` **never reads `failed`**: `Restart=always` moves it from a `TimeoutStartSec` expiry straight to `activating/auto-restart`, so `systemctl is-failed` reports `activating` throughout a bad update. Since `podman auto-update --rollback` decides an update went bad by restarting the unit and seeing the restart fail, #4411 framed that as possibly disqualifying.

**It is not disqualifying, and the reason is the whole result: those are two different signals.**

Podman restarts over D-Bus and reads the **start-job result**. A `TimeoutStartSec` expiry makes that result `timeout`, and podman rolls back on anything that is not `done`. `ActiveState` never has to reach `failed`.

Both measurements are true at once. For an operator the unit is still a black hole during a bad update — `is-failed` says `activating`, `ActiveState` says `activating`, `NRestarts` says `0`, nothing keyed on them fires. #4378's finding is unchanged. It is simply not what podman uses.

## Measured, against a bad-but-startable image

Podman 5.8.4, rootless, systemd user manager — the version #4378 and the #420x spikes ran on. The image **runs forever and never becomes healthy**, as #4188 requires; a pull failure is a much easier case podman never even restarts on.

The fixture reproduces the unit's decisive properties (`Notify=healthy`, `HealthCmd`, `HealthStartPeriod`/`Retries`, `TimeoutStartSec`, `Restart=always`, `SuccessExitStatus=143 SIGTERM`, named volume at `/data`) with timers scaled down; the mechanism is timer-independent, so costs are given as multiples of `TimeoutStartSec`. The entrypoint traps `SIGTERM` and exits 143 like the real one — an earlier run without that inflated downtime by ~50s of stop-timeout the real unit never pays, and **that inflated number is not the one published**.

```
UNIT          CONTAINER            IMAGE                              POLICY    UPDATED
hive.service  75f8ed7839a6 (hive)  localhost:5000/hivefixture:stable  registry  rolled back
elapsed_s=36        (TimeoutStartSec=30)

t+006s  activating/start/success   is-failed=activating   NRestarts=0  container=bad
  …                                                                    (running, never healthy)
t+033s  activating/start/success   is-failed=activating   NRestarts=0  container=bad
t+036s  activating/start/success   is-failed=activating   NRestarts=0  container=good
t+039s  active/running/success     is-failed=active       NRestarts=0  container=good
```

| Finding | Evidence |
|---|---|
| Rollback fires on this unit | `UPDATED=rolled back`, back to `active/running` on the previous image |
| Cost of a bad update | one full `TimeoutStartSec` + a few seconds ⇒ **~5 min** on the shipped 300s |
| **It does not remember** | a second run, registry unchanged, rolled back **again** at full cost — so a bad published tag is a **recurring outage on every timer firing** |
| Digest pin | **wins, silently**: `UPDATED=false`, exit 0, unit untouched — reads exactly like "already up to date" |
| Unresolvable pinned digest | `rc=125` fails the **whole** auto-update run, every other container on the host included |
| `hive-data` | marker file intact through every direction |

## Answering the acceptance criteria directly

- **`Restart=always` is KEPT, unchanged, and no tradeoff is taken** — because it is not involved. `NRestarts` stayed `0` through every rollback: podman's restart lands before `RestartSec` elapses. Removing it to expose `failed` for alerting would have cost the crash recovery #4377 added it for and bought nothing.
- **A bad-but-startable image is the test case**, per the umbrella.
- **The detector is shown firing**, and the measured trap is reproduced rather than sidestepped: `systemctl is-failed` returns `activating` in the timeline above, which is exactly why nothing may key on it.
- **The pin interaction is stated and tested**, both when the digest resolves and when it does not.
- **Opt-in.** `hive.container` gains no `AutoUpdate=` key; a tracked drop-in is installed on request.
- **`hive-data` proven** with a marker, the way #4378 proved it.

## What ships

`src/deploy/quadlet/optional/hive-autoupdate.conf` — one `AutoUpdate=registry` line under a header that carries the measurement and the costs to the host with it.

`bin/hive-podman-update.sh autoupdate <on|off|status>` installs it as `hive.container.d/20-autoupdate.conf`, enables `podman-auto-update.timer`, and restarts (the policy label is applied at container-create time, so a pre-existing container is invisible to auto-update until recreated).

It **refuses** to arm auto-update over a digest pin rather than warning — that combination gives a host whose daily timer reports success and updates nothing, which is worse than either state alone and completely silent. `status` flags the combination when an operator reaches it the other way round, which is exactly what a manual rollback does. The two drop-ins are separate files on purpose: `unpin` does not disable auto-update, and `autoupdate off` does not drop someone's pin.

The manual path is untouched, and Docker's Watchtower path is not touched at all.

## Verification

**End to end against the shipped artifacts, not only mocks** — since #4411 says a shipped `AutoUpdate=registry` that has not been driven through a bad image does not close it. A real unit named `hive.service`, `bin/hive-podman-update.sh autoupdate on` installing the tracked drop-in and arming the timer, container carrying `io.containers.autoupdate=registry`, a bad image published to the tag, `rolled back` in 36s, `hive-data` intact, and the pin guard exiting 78 with its explanation.

Plus **25 new cases** in `bin/test_hive_podman_update.sh` (71 total), fully mocked — no podman, no privileges, no network — covering the refusal changing nothing on disk and touching no timer, `off` leaving the pin alone, `unpin` leaving auto-update alone, a host with no `podman-auto-update.timer`, and that the file installed is byte-identical to the tracked one. `shellcheck` clean. `test_quadlet_port_boundary.sh`, `test_quadlet_config_contract.sh`, `test_standalone_image_refs.sh`, `test_standalone_service_contract.sh` and `test_bin_suites_wired.sh` all still pass.

The test host's own `podman-auto-update.timer` was already enabled system-wide (root-owned symlink predating this work) and is left as found; every fixture unit, container, volume, image and registry override was removed.

## Not executed

Recorded rather than inferred, the way #4378 records its own: rootful mode (the mechanism is the system manager and nothing in podman's rollback path is uid-dependent, but it was not run); a genuinely broken *published* Hive image, so the bad image is a stand-in; and the shipped 300s timers end to end — the scaled fixture is what was driven, which is why costs are expressed as multiples of `TimeoutStartSec`.


clubanderson:
Reviewed from head 2254d159. The central result — that #4378's never-reads-`failed` finding does NOT disqualify `podman auto-update --rollback` because podman reads the D-Bus start-job result (`timeout`) rather than `ActiveState` — is the right resolution of the tension, and both measurements coexisting (rollback fires; operator alerting keyed on `is-failed`/`NRestarts` still never does) is exactly the kind of distinction that needed a live drive rather than an assumption.

- **Strictly opt-in, verified**: hive.container gains only a comment — no `AutoUpdate=` key, `Restart=always` untouched. The tracked drop-in in `optional/` ships nothing until `autoupdate on` copies it, and copying rather than generating so the measured-costs header travels to the host is a nice touch.
- **The pin-vs-track tension is resolved the right way round**: refusing (exit 78) to arm auto-update over a #4378 pin — because the resulting host reports a green daily timer while updating nothing, and a GC'd pinned digest fails the *whole* host's auto-update run at 125 (we've watched quay do exactly that to this repo's own CI, twice) — is strictly better than a warning. Separate drop-in files so `unpin` and `autoupdate off` stay independent decisions, and `status` flagging the both-files state an operator reaches via manual rollback closes the remaining path into the trap.
- **Costs stated honestly**: one full TimeoutStartSec (~5 min shipped) per bad update, *repeated on every timer firing* because podman doesn't remember a rollback — that recurring-outage finding is the single most important line in the doc, and it's in the drop-in header where an enabler will actually see it.
- **Locally: 71/71** on the update suite (4 BSD-`install -D` cases under a GNU shim as with #4378 — CI ubuntu authoritative and green), orphan guard 16/0, cleanup contract 127/127, all five deploy contract siblings pass, no forbidden literals, shellcheck clean at error/warning level (remaining SC2016 infos are the suite's deliberate deferred-eval pattern).
- The arg-parser extension (`autoupdate status` vs two-commands error) is correctly scoped to argument-taking commands.

/lgtm
/approve

Issue #4422, opened 2026-08-21
🐛 hive-podman-preflight-host.sh reports a false failure on a Quadlet install (looks for deploy/nginx.conf, which the Quadlet layout does not have)


Danathar:
## Summary

`bin/hive-podman-preflight-host.sh` cannot validate a Quadlet install. It checks
the **source-tree** layout — `$HIVE_SRC_DIR/hive.yaml`,
`$HIVE_SRC_DIR/deploy/nginx.conf`, `$HIVE_SRC_DIR/secrets` (default `src/`) —
but the Quadlet install described in
`src/docs/podman-standalone-quadlet.md` puts the operator's files in `%E/hive`
with `nginx.conf` **flat**, not under a `deploy/` subdirectory.

Found while running the Podman quick start end to end for #4412.

## Reproduce

Stage a Quadlet install exactly as the install doc says, then point the host
preflight at it:

```sh
export HIVE_DEPLOY_RUNTIME=podman
CONF=~/.config/hive
mkdir -p "$CONF/secrets" && chmod 750 "$CONF/secrets"
cp src/hive.yaml.example "$CONF/hive.yaml"
cp src/deploy/nginx.conf "$CONF/nginx.conf"      # flat, per the install doc

HIVE_SRC_DIR="$CONF" bin/hive-podman-preflight-host.sh
```

```
  ✓ Hive config: /home/u/.config/hive/hive.yaml readable (mode 644, owner u)
  ✗ Gateway config: /home/u/.config/hive/deploy/nginx.conf does not exist
  ✓ Secrets dir: /home/u/.config/hive/secrets readable (mode 750, owner u)
  ✓ Secrets reach: … traversable by the container's hive-launch group
  ✓ Port 3001: free
SUMMARY: pass=5 warn=2 fail=1
rc=78
```

The `✗` is spurious — `nginx.conf` is present and correct, one directory level
up from where the check looks. Everything else in the run is genuinely useful
(SELinux mount labels, the secrets group-traverse check, port 3001), which is
what makes the false failure worth fixing rather than working around: the
non-zero exit invites operators to stop reading a report whose other rows they
need.

Leaving `HIVE_SRC_DIR` at its default is not a fix either — it then checks
`src/hive.yaml`, which for a Quadlet install correctly does not exist
(`src/.gitignore` excludes it), so that also reports a fail.

## Why it matters

The Quadlet install is the one path where the layout differs, and it is the path
[...]

PR #4423, opened 2026-08-21
docs: add the Podman standalone quick start to the README (#4412)

### LINKED ISSUE #4412: 📝 Podman: add the standalone quick start to the README
## Parent

Part of #4188 (second wave, bullet 13). This issue must not close the parent.

## Why

`README.md`'s entry point is headed **"Quick Start (Docker Compose)"** — the
title already implies a sibling that does not exist. Podman appears **nowhere**
in the root README. An operator who standardises on Podman, or who is on
Fedora/RHEL/an Atomic desktop, reaches the front door and finds only
`docker compose -f src/docker-compose.yaml up -d`.

That is the gap #4188 opened with: *"Hive's standalone quick start currently
assumes Docker Engine and Docker Compose."* The runtime now exists — units in
`src/deploy/quadlet/`, an install procedure verified in both root modes
(#4354, #4375, #4377), preflight (#4207–#4209), cleanup (#4326), and an update
and rollback path (#4378). None of it is reachable from the README.

## 30-minute boundary

The quick start only. Not a rewrite of `src/docs/podman-standalone-quadlet.md`,
which already carries the detail and stays the destination. Do not restructure
or relocate the Docker quick start — #4188 requires Docker remain the default
and be presented as a parallel supported choice.

## Deliverable

A `## Quick Start (Podman)` section in `README.md`, parallel in shape and depth
to the Docker one — prerequisites, copy-and-paste commands, where the dashboard
lands, how to build instead of pull — linking onward to
`src/docs/podman-standalone-quadlet.md`.

## Acceptance criteria

- Both quick starts are presented as **parallel supported choices** and it is
  clear Docker is the default. Podman must not read as experimental; it also
  must not read as recommended over Docker.
- The commands are **copy-and-paste and complete**. #4367 is the precedent: the
  Quadlet install step referenced a file that does not exist, and the obvious
  substitute cost a silent 300-second hang. A quick start earns its name by
  having been run start to finish.
- Says which root mode the block is for, or works in both.
- The security-mode consequence is visible at the front door, not deferred to a
  linked doc: which combinations enforce the egress gate, and that advisory mode
  is deliberate and not equivalent (#4324's matrix).
- Verified on a host with **no Docker installed** — a literal #4188 acceptance
  criterion, and the only case a Docker-equipped workstation cannot demonstrate.
- Points at `bin/hive-podman-preflight*.sh` before the start command, so a
  missing subordinate-ID or cgroup prerequisite surfaces as a preflight message
  rather than a failed start.

## Related

- #4188 — the umbrella; the Docker-free-host criterion
- `src/docs/podman-standalone-quadlet.md` — the install detail this links to
- #4367 — why "copy-and-paste and complete" is an explicit criterion here

Danathar:
Closes #4412. Part of #4188 (second wave, bullet 13) — does not close the parent.

## The gap

`README.md`'s entry point was headed **"Quick Start (Docker Compose)"** — a title implying a sibling that did not exist. Podman appeared nowhere in the root README, so an operator on Fedora/RHEL/an Atomic desktop reached the front door and found only `docker compose up -d`, while the runtime behind it had already shipped across #4354, #4375, #4377, #4207–#4209, #4326 and #4378.

## What ships

A `## Quick Start (Podman)` section parallel in shape and depth to the Docker one, plus a short chooser above both:

| | Docker Compose | Podman |
| --- | --- | --- |
| Lifecycle | `docker compose up -d` | Quadlet units under systemd |
| Runs as | the Docker daemon | rootful **or** rootless |
| Update path | pull and recreate; optional Watchtower | pinned by digest, with rollback |

Docker stays the default and stays put — the Docker block is unchanged and unmoved. Podman reads as a parallel supported choice: not experimental, not recommended over Docker.

The block is **rootless**, with the rootful deltas named inline, and it carries the two things a linked doc would have deferred:

- **The security posture, at the front door.** The shipped unit requests `CAP_NET_ADMIN`, so the egress gate is *enforced* by default. #4324's rootful/rootless × enforcing/advisory matrix is inline, including that advisory mode is a deliberate choice and **not** a weaker grade of enforcing or a fallback — agents can bypass the MITM proxy and the ACMM capability model is not enforced.
- **The traps that cost 300 seconds each** — the `dashboard.port` 3001→3002 rewrite, `hive.env` having to exist, and pulling before starting because the generated `ExecStart`'s pull is spent inside `TimeoutStartSec`.

## Running it, rather than transcribing it, found two defects

#4412 makes "copy-and-paste and complete" an explicit criterion because of #4367. Both of these are that same class — a step that reads fine and silently does nothing:

1. **`HIVE_DEPLOY_RUNTIME` must be exported before the preflights, or all three exit 0 having checked nothing.** Docker is the default runtime; with it selected they print `Podman preflight: skipped — HIVE_DEPLOY_RUNTIME selects docker` and return success, which reads exactly like a pass. The acceptance criterion asks that a missing subordinate-ID or cgroup prerequisite surface as a preflight message — as written, it would not have. The README exports it first, and `podman-standalone-quadlet.md`'s Requirements bullet (which said only "Run the preflight first") now says so too.

2. **`hive-podman-preflight-host.sh` cannot validate a Quadlet install.** It looks for `$HIVE_SRC_DIR/deploy/nginx.conf`; the Quadlet layout puts `nginx.conf` flat in `%E/hive`. Pointed at the config directory it reports a spurious `✗ Gateway config: …/deploy/nginx.conf does not exist` and exits 78, inviting operators to stop reading a report whose other rows (SELinux mount labels, the secrets group-traverse check, port 3001) they actually need. Filed as **#4422**; the README names only the two layout-independent preflights and points at the issue.

Also documented, as observed: in a logged-in session the stack comes up on `daemon-reload`, not on the explicit start — the generator writes the `default.target.wants/` symlinks from `[Install]` and systemd starts newly-wanted units of an already-active target. The `start` is then a no-op returning 0, worth running as the confirmation.

## Verification

Podman 5.8.4 rootless, SELinux enforcing, cgroup v2. Every step in the block was executed in order:

| Step | Result |
|---|---|
| `hive-podman-preflight.sh` | pass 4 / warn 0 / fail 0 |
| `hive-podman-preflight-ids.sh` | pass 4 / warn 0 / fail 0 |
| config staging + `sed` | `dashboard: / port: 3002` confirmed |
| `hive.env` from the tracked example + token | 0600, `HIVE_DASHBOARD_TOKEN` set |
| `podman pull ghcr.io/kubestellar/hive:stable` | 47s, 3.82 GB |
| 4× `install` + `daemon-reload` | all four services `generated` |
| **stack health** | **`hive` healthy in 43s, `hive-gateway` healthy 12s later** |

Those healthchecks *are* the verification commands: `hive.container`'s `HealthCmd` is `curl -sf http://127.0.0.1:3002/api/health`, and the gateway's is `curl -sf http://127.0.0.1:3001/api/health` — which passing also proves the gateway resolved `hive` over the shared network.

## Not executed — stated rather than implied

- **The host was not Docker-free.** #4188's Docker-free-host criterion is *not* demonstrated here. No command in the block invokes docker and none of the verification used it, but that is a weaker claim and I am not dressing it up as the stronger one. My attempt to mask `docker` out of `PATH` failed (it lives in `/usr/bin`, which the block needs), and I did not substitute a fake proof.
- **The explicit `systemctl start` and the `curl`** were not separately run: the verification host is the maintainer's own desktop and the run was halted at their request once the stack was up. The stack had already reached healthy via the `daemon-reload` path above, and everything created was torn down — units, containers, volume, network, and the config directory restored byte-for-byte (it also holds live contributor-relay credentials, backed up before the run).
- **The `podman build` line and the rootful variant** were not exercised.

All links resolve on v4 — including a check that caught me linking `src/docs/podman-auto-update.md`, which exists only in my unmerged #4420; that link is gone. The #4367 config-coupling guard, the port-boundary guard, the runtime-parity guard and the image-reference guard all still pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,20 @@ AI agent orchestration for open source projects. A single Go binary enumerates G
 
 Hive separates decisions into two layers: a **deterministic pipeline** of shell scripts handles filtering, classification, merge-gating, and enforcement before any LLM sees the work. Agents only handle judgment calls — reading code, reasoning about fixes, writing PRs.
 
+## Quick Start
+
+Two supported standalone runtimes. **Docker Compose is the default** and is what
+the rest of this README assumes; **Podman** is a parallel supported choice, not
+an experiment and not a recommendation over Docker. Pick one — they install the
+same two services (Hive plus its authenticating gateway) and land the dashboard
+on the same port.
+
+| | [Docker Compose](#quick-start-docker-compose) | [Podman](#quick-start-podman) |
+| --- | --- | --- |
+| Lifecycle | `docker compose up -d` | Quadlet units under systemd |
+| Runs as | the Docker daemon | rootful **or** rootless |
+| Update path | pull and recreate; optional Watchtower profile | [pinned by digest, with rollback](src/docs/podman-quadlet-update-rollback.md) |
+
 ## Quick Start (Docker Compose)
 
 **Prerequisites**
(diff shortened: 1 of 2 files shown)


clubanderson:
Fact-checked against v4: all 15 referenced paths exist and are tracked (preflights, lifecycle probe, teardown, four Quadlet units, hive.env.example, nginx.conf, update-rollback and support-matrix docs); the HIVE_DEPLOY_RUNTIME skip message is verbatim from bin/hive-podman-preflight.sh:212 and the exit-0-looks-like-a-pass hazard is real; preflight-host's deploy/nginx.conf expectation and exit 78 confirmed (#4422 filed and matches); gateway publishes only 3001 and probes 3001, hive probes 3002 with no PublishPort so the port story matches #4369's contract exactly; WantedBy=default.target plus the cannot-enable-generated-units note are correct; Image=ghcr.io/kubestellar/hive:stable matches both the pull and the build-and-tag instructions; the digest-pin update link matches #4407's actual mechanism; the inline support matrix reproduces the merged #4340 cells faithfully including advisory-is-not-a-fallback and exit 77; and no reference to the unmerged #4420 auto-update doc survives. Honest not-executed section is a model for this kind of PR. Mergeable against the #4419 README changes, cleanup contract clean, CI green.

/lgtm
/approve

PR #4433, opened 2026-08-21
fix(preflight): teach the host check the Quadlet layout (#4422)

### LINKED ISSUE #4422: 🐛 hive-podman-preflight-host.sh reports a false failure on a Quadlet install (looks for deploy/nginx.conf, which the Quadlet layout does not have)

Danathar:
Closes #4422.

## The bug, reproduced

`hive-podman-preflight-host.sh` knew only the source-tree layout — the one the Compose stack mounts out of a checkout, where the gateway config sits at `deploy/nginx.conf`. The Quadlet install puts the operator's files in `%E/hive` with `nginx.conf` **flat**. Pointed at a staged Quadlet config directory:

```
  ✓ Hive config: /tmp/…/hive.yaml readable (mode 644, owner dbaggett)
  ✗ Gateway config: /tmp/…/deploy/nginx.conf does not exist
SUMMARY: pass=4 warn=3 fail=1        rc=78
```

…with `nginx.conf` sitting right there, one level up.

That lands on the one install path where this check is *most* useful: SELinux labelling of those binds is exactly the class of problem that otherwise surfaces as a 300-second start timeout with the container already `--rm`'d away. And the non-zero exit invites an operator to stop reading a report whose other rows — mount labels, the secrets group-traverse check, port 3001 — they still need. Leaving `HIVE_SRC_DIR` at its default was no escape: it then checks `src/hive.yaml`, which for a Quadlet install correctly does not exist.

After:

```
  layout: quadlet (detected) — config under /tmp/…, gateway config at nginx.conf
  ✓ Gateway config: /tmp/…/nginx.conf readable (mode 644, owner dbaggett)
  ✓ Environment file: /tmp/…/hive.env readable (mode 600, owner dbaggett)
  △ Dashboard token: HIVE_DASHBOARD_TOKEN is not set in /tmp/…/hive.env
SUMMARY: pass=6 warn=5 fail=0        rc=0
```

## Both options from the issue, because neither alone is right

#4422 offered "probe `deploy/nginx.conf` and fall back to `nginx.conf`, **or** take an explicit layout flag". This does both:

- **Auto-detection** is what an operator mid-install actually needs — they shouldn't have to know a flag exists to get a truthful report. Detection keys on the gateway config, the only file whose *path* differs; `hive.env` is the tiebreak, since nothing in the source tree carries that name. A directory with neither shape stays on the historical default rather than guessing.
- **`HIVE_PODMAN_LAYOUT=source|quadlet`** overrides, for a directory that has neither gateway config in place yet.

Auto-detection that doesn't say what it picked is indistinguishable from a check looking in the wrong place — which is the bug being fixed — so every run now opens with `layout: quadlet (detected)` or `layout: source (HIVE_PODMAN_LAYOUT)`.

## The relabel recommendation flips with the layout

This is the part that would have been worse to get wrong than the path bug, and it isn't mentioned in the issue:

| Layout | Recommends | Why |
|---|---|---|
| source-tree | `:z` | the config files are part of the operator's **checkout** and likely read by other tooling; `:Z` would stamp a per-container MCS category onto a tracked file |
| quadlet | `:Z` | the files are operator-owned, read by exactly **one** container, and the shipped units mount all three `:ro,Z` — recommending `:z` would contradict the units |

Both directions are pinned by test, including that the quadlet path *never* emits the shared-form hint.

## The `hive.env` row the issue asked me to consider

Added, Quadlet-only, and deliberately **not** a bind source: `EnvironmentFile=` is read by systemd on the host and never mounted, so it has no container label to judge (`podman-volume-persistence.md` records it staying at its ordinary host label). It earns a check because both ways it goes wrong cost the whole start budget:

- **missing** → `EnvironmentFile=` becomes `podman run --env-file`, which fails outright. Reported as a failure.
- **no `HIVE_DASHBOARD_TOKEN`** → Hive refuses to start, and `Notify=healthy` turns that into a unit stuck in `activating` until `TimeoutStartSec` expires. Reported as a **warning**, not a failure, because a hub-hosted hive doesn't need one.

The token's *value* is never printed, and a test asserts that.

## Verification

**195 assertions pass** (was 168). The new cases were run against the **unfixed** script and fail on exactly the reported symptom — `✗ Gateway config: .../quadlet/deploy/nginx.conf does not exist`, rc=1 — so they test the bug, not the implementation.

The source-tree path is untouched: same paths, same `:z` recommendation, same output, asserted by test, and the case that a source layout grows *no* `hive.env` or dashboard-token row is asserted too.

Suites: preflight-host 195, preflight 44, preflight-ids 61, bin-suite-wiring 16/0 orphaned, quadlet config-coupling 7/7, port-boundary 18/0, runtime-parity 33/0/8. `shellcheck` clean on both files.

## Docs

Two notes I wrote while filing this are now false and are corrected here:

- `podman-standalone-quadlet.md` described the spurious failure and told operators to run it against the source tree or ignore the row. Replaced with the working invocation, **moved to after step 1**, since it checks the files step 1 creates.
- The README quick start drops its "reads the source-tree layout rather than this one — see #4422" caveat and now runs the host preflight against `$CONF`, where it belongs in the sequence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,8 +81,6 @@ cd hive
 # that times out five minutes later.
 bin/hive-podman-preflight.sh
 bin/hive-podman-preflight-ids.sh
-# (hive-podman-preflight-host.sh checks SELinux labels and port 3001 too, but
-#  reads the source-tree layout rather than this one — see #4422.)
 
 CONF=~/.config/hive                       # rootful: CONF=/etc/hive
 mkdir -p "$CONF/secrets" && chmod 750 "$CONF/secrets"
@@ -106,6 +104,10 @@ printf 'HIVE_DASHBOARD_TOKEN=%s\n' "$(openssl rand -hex 32)" >> "$CONF/hive.env"
 # L5/L6 if agent PRs may touch `.github/workflows/`. See
 # src/docs/github-app-setup.md#personal-access-token-pat-scopes
 printf 'HIVE_GITHUB_TOKEN=%s\n'    'ghp_...'                 >> "$CONF/hive.env"
+
+# Now the host preflight, which checks what the steps above just created:
+# SELinux labels on the bind sources, secrets reachability, hive.env, port 3001.
+HIVE_SRC_DIR="$CONF" bin/hive-podman-preflight-host.sh
 
 # Pull before starting. The generated ExecStart pulls a missing image itself and
 # that pull is spent inside TimeoutStartSec; the Hive image is ~3.8GB.
--- a/bin/hive-podman-preflight-host.sh
+++ b/bin/hive-podman-preflight-host.sh
@@ -181,18 +181,95 @@ hive_podman_check_selinux() {
 
 # --- 2. Mount labeling -------------------------------------------------------
 #
-# The three bind-mount sources in src/docker-compose.yaml. Each is consumed by
-# exactly one container, so the private (:Z) form is always available; the
-# read-only configuration files are called out as shared (:z) candidates
-# because :Z stamps a per-container MCS category onto a file that is part of
-# the operator's checkout and is likely read by other tooling.
+# TWO LAYOUTS, AND THE DIFFERENCE IS NOT COSMETIC (#4422).
+#
+# This check knew only the source-tree layout — the one the Compose stack mounts
+# out of a checkout, where the gateway config sits at `deploy/nginx.conf`. The
+# Quadlet install (src/docs/podman-standalone-quadlet.md) puts the operator's
+# files in `%E/hive` with `nginx.conf` FLAT, so pointing this script at that
+# directory reported
+#
+#   ✗ Gateway config: <dir>/deploy/nginx.conf does not exist
+#
+# and exited 78 with the file sitting right there, one level up. That landed on
+# the one install path where this check is most useful: SELinux labelling of
+# those binds is exactly the class of problem that otherwise surfaces as a
+# 300-second start timeout with the container already `--rm`'d away. A non-zero
+# exit also invites an operator to stop reading a report whose other rows —
+# mount labels, the secrets group-traverse check, port 3001 — they still need.
+#
+# The recommended relabel option differs by layout too, and getting THAT
+# backwards would be worse than the path bug:
+#
+#   source-tree   the config files are part of the operator's CHECKOUT and are
+#                 likely read by other tooling, so `:z` (shared) is right and
+#                 `:Z` would stamp a per-container MCS category onto a tracked
+#                 file.
+#   quadlet       the config files are operator-owned files in %E/hive read by
+#                 exactly ONE container, and the shipped units mount all three
+#                 `:ro,Z`. Recommending `:z` there would contradict the units.
 #
 # Format: relative-path|description|recommended-option
-HIVE_PODMAN_BIND_SOURCES=(
+HIVE_PODMAN_BIND_SOURCES_SOURCE=(
   "hive.yaml|Hive configuration|z"
   "deploy/nginx.conf|gateway configuration|z"
   "secrets|secrets directory|Z"
 )
+HIVE_PODMAN_BIND_SOURCES_QUADLET=(
+  "hive.yaml|Hive configuration|Z"
+  "nginx.conf|gateway configuration|Z"
+  "secrets|secrets directory|Z"
+)
+
+# Which layout HIVE_SRC_DIR holds: auto | source | quadlet. `auto` decides from
+# what is on disk; the explicit values are for a directory mid-install that has
+# neither gateway config in place yet.
+HIVE_PODMAN_LAYOUT="${HIVE_PODMAN_LAYOUT:-auto}"
+
+# Resolved once by _pfh_resolve_layout, read by every check below.
+HIVE_PFH_LAYOUT=""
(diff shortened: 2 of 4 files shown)


clubanderson:
Reviewed from head 41a68444 (f327cf12 plus my fix-forward — see below). The #4422 fix is right, and taking both of the issue's options was the correct read.

- **Layout detection is sound**: keyed on the gateway config (the only path that differs), `hive.env` as tiebreak (nothing in the source tree carries that name), historical default when neither shape exists, announced on the first line, `HIVE_PODMAN_LAYOUT` override with exit 64 on a bad value. The announce-what-you-picked rule is exactly the right lesson from the bug being fixed.
- **The relabel flip is the sleeper win**: `:Z` recommended for the quadlet layout — verified all three shipped binds are `:ro,Z`, so recommending `:z` there would have contradicted the units — while the source tree keeps `:z` for checkout-shared files. Both directions pinned by test.
- **hive.env row correctly scoped**: quadlet-only, not a bind-source row (`EnvironmentFile=` is host-read, no container label to judge), missing = failure (start fails outright), token-unset = warning (hub-hosted needs none), value never printed and asserted so.
- **Docs consistent**: the #4423 README quick-start hazard note replaced with the working `HIVE_SRC_DIR="$CONF"` invocation, and the install-doc guidance moved after step 1 where the checked files exist.
- **Fix-forward (mine, 41a68444)**: the teardown suite's #4210 creation guard scans non-comment lines of bin/ assets for resource-creating podman invocations, and the new missing-hive.env *hint string* quoted the run form of the command in prose — a false positive of the same class the cleanup-contract greps produce. Reworded the hint (and its test assertion) to describe the flag without naming the command; no behavioural change. Teardown 74/74 after.
- Locally: preflight-host **195/195** (BSD-stat shim; CI authoritative), teardown 74/74, orphan guard 16/0, cleanup contract 127/127, shellcheck clean.

Note for the repo: the new changelog-reminder job 403'd on its first run here (fork PR token can't post comments) — worth a look at #4429's workflow permissions, though it cleared on re-run and never blocks merges.

/lgtm
/approve

PR #4436, opened 2026-08-21
sec(mint): caller identity seam, entitlements, and per-caller audit (#3915)

Danathar:
Addresses #3915. Not marked as closing it — the TokenReview backend the finding names as the primary fix is deliberately out of scope here, for a reason worth your decision (below).

## First, a fact that should change the severity assessment

**The mint HTTP server is not wired into anything.** `mint.NewServer` has exactly one reference in the tree — its own definition. No listener serves `MintPath` or `JWKSPath`, `MintConfig` has no field to even supply the shared secret (only `Enabled`/`KeyPath`/`Issuer`/`MaxTTLSeconds`), and `main.go` uses the package solely to build an in-process `AgentMinter`.

So the impact statements in the finding — "lateral-movement attack within the pod network", "internet-accessible minting is possible" — describe a server that nothing currently starts. The weakness in the code is real; the exposure is **latent, not live**. That also makes the recommended short-term hardening ("bind only to localhost") inapplicable: there is no listener to bind.

I'd rather say that plainly than quietly fix a medium-severity finding as though it were exploitable today.

## What this changes

The finding has three parts. Two close here with **no new dependency**:

**Identity — the `TODO(caller-auth)` seam.** `CallerAuthenticator` is `Authenticate(*http.Request) (Identity, error)`. The shared secret becomes *one* implementation instead of the only mechanism, so a TokenReview or mTLS backend drops in via `WithAuthenticator` without touching the handler. A stub backend in the tests proves the substitution actually works, including that presenting the still-configured shared secret does **not** help once a different authenticator is in force.

**Blast radius — `Entitlements`.** This is the part the finding describes as "any subject, audience, and scope, bounded only by the TTL cap", and it's true of the code regardless of deployment: `Minter.Mint` validates only that subject and audience are non-empty. Where entitlements are configured the mint is deny-by-default per identity, with two design choices I'd flag for review:

- **An empty dimension allows nothing, not everything.** An entry that grants an audience but forgets the scopes grants a *scopeless* token, never a token with every scope. This is the trap that makes allow-lists fail open.
- **`"*"` is an explicit, greppable wildcard.** A caller that genuinely mints arbitrary subjects has to write it down rather than get it by omission. The wildcard is per-dimension — a test asserts it doesn't widen a neighbouring one.

Refusals are **403, not 401**: the caller is authenticated, it is simply not entitled.

**Audit.** The verified identity is logged on every mint, every entitlement refusal, and every failed authentication.

One deliberate call worth your eye: the shared-secret identity is `shared-secret:any-holder`, **not** a per-caller name. A shared secret cannot tell its holders apart, and inventing a name per request would make the audit log assert an identity the mechanism never established. The line says exactly what was proven.

## What is deliberately NOT here, and why it's your call

The scanner's plan makes a **Kubernetes TokenReview backend** the primary fix. It needs `k8s.io/client-go`, which this module does not depend on today. Adding a dependency of that weight — for a server nothing currently starts — is a maintainer decision, not something to slip into a security PR. The seam is precisely what makes it a contained, additive change once you've answered that. The `TODO(caller-auth)` comment is rewritten to record the trade-off rather than deleted.

If you'd like client-go added, say so and I'll do the TokenReview backend as a follow-up behind this interface.

## Backwards compatibility

`NewServer` keeps its signature (options are variadic and additive) and its default posture is byte-identical: shared secret, no entitlement bound. **The pre-existing `server_test.go` passes untouched**, and a test pins the historical "any holder mints anything" behaviour so the default stays a documented choice rather than an accident.

Because "unbounded" is now a *choice*, `NewServer` logs a warning at construction when no entitlements are configured — the posture is visible instead of discovered from a token that should never have been issued.

## Verification

Run with the repo's own pinned toolchain (`golang:1.26-alpine` at the digest `src/Dockerfile` pins), since no Go is installed on this host:

- `go build ./...` clean, `go vet` clean, `gofmt -l` clean
- `pkg/mint` tests pass — **92.3% statement coverage**, 131 test entries

`pkg/agent` fails in this environment **both with and without these changes** — confirmed by stashing and re-running against clean `v4`. Pre-existing and unrelated (those tests want a real uid/process environment; I'm running as root in a container). Flagging rather than reporting the repo as uniformly green.


clubanderson:
Security review complete — **approving**.

**Security verdict (against the #3915 threat model):**

1. **Identity seam not spoofable** ✅ — identity derives solely from `CallerAuthenticator.Authenticate(*http.Request)` (transport credential), never from request-body fields. The shared-secret backend deliberately reports the constant `shared-secret:any-holder` identity rather than pretending a shared secret can distinguish holders — honest, and the right call.
2. **Fail closed** ✅ — empty secret refused at construction; `WithAuthenticator(nil)` ignored (can't accidentally disable auth, pinned by test); with entitlements configured, unknown identity = 403, empty dimension list = allows *nothing*, wildcard is explicit and per-dimension (does not widen siblings). Out-of-grant scope is refused outright, not silently narrowed.
3. **Audit without leakage** ✅ — mint and refusal log lines carry caller, kind, subject, audience, scopes, reason, TTL; never the presented credential or minted token. Unauthenticated refusals log only authenticator name + remote_addr, and the client learns nothing about *why* (401, undifferentiated).
4. **No privilege escalation** ✅ — entitlements only narrow; `Minter.Mint` and TTL cap untouched. Verified `mint.NewServer` has **zero production call sites** (grep: only its definition + tests) — this endpoint is latent, not live, so the historical allow-anything default (nil entitlements) carries no new live exposure and is now loudly warned at construction.
5. **Backward compat** ✅ — `NewServer` signature unchanged (variadic opts); the only production mint path (`buildAgentMinter` → in-process `AgentMinter` in main.go) is untouched. Behaviour identical unless an operator opts in.
6. **Deny-path coverage** ✅ — refused-subject/audience/scope/mixed-scope, unknown identity, empty dimension, wildcard non-widening, wrong/missing/prefix/empty-bearer secret, substituted-authenticator-refusal-is-401, and refused-mint-returns-no-token are all tested. Local run: build/vet/gofmt clean, `go test -cover ./pkg/mint/` → **92.8%**.

TokenReview/mTLS backends correctly left as a follow-up behind the seam (k8s.io/client-go dependency is a maintainer call). Fixed forward with a CHANGELOG entry (979a119b) — security changes are in the changelog's own scope. CI green.

/lgtm
/approve

Issue #4399, opened 2026-08-21
🐛 bug: problems parsing Terminal content


MikeSpreitzer:
## What happened?

I am administering the hive for llm-d FMA. I opened the Terminal for the quality agent and saw the following. No more output appeared for a few minutes. Closing and re-opening the terminal exposed no more output. I do not understand what it means for an agent to be halted at this point.

<img width="1238" height="779" alt="Image" src="https://github.com/user-attachments/assets/f074116b-8b1e-4aa6-b10d-c75c56ff21ed" />

Later, after the next scheduled kick, more output appeared.

When I scroll some, I see some black-on-yellow text in the upper right hand corner that contains some line position/counter information and, usually, a timestamp. It is not obvious which point in the terminal corresponds with that timestamp. Looking at particular examples, neither the top nor the bottom of the terminal seems to correspond consistently with the timestamp. See the following snapshot.

<img width="1245" height="772" alt="Image" src="https://github.com/user-attachments/assets/207c0f16-a71c-4d86-a191-162154cb2eb4" />

Now consider the following snapshot, which shows the quality agent for the hive for llm-d FMA donig some work at 01:47 on Aug 21. This is unexpected, since the dashboard section for that agent said its last run was 8/20 8:12 PM EDT and its next run would be 8/21 2:12 AM EDT. Why was it doing something at 01:47 AM EDT on 8/21?

<img width="1240" height="776" alt="Image" src="https://github.com/user-attachments/assets/e2b63e75-8f35-4454-bd10-bc32bfc10eaa" />

## Expected behavior

Easy to parse content

## Reproduction steps

I am not sure.

## Environment

- Branch/version: `v4` / commit 9f31439295c6fe8b70df70f8f16ea76afd57a063
- Deployment method: Hive Hub
- Operating system or Kubernetes version: unknown
- Relevant configuration (redact secrets): ?

## Logs or screenshots

See above

## Additional context

Anything else maintainers should know?

clubanderson:
Thank you for the unusually precise report, @MikeSpreitzer — all three observations turned out to be real and diagnosable, and each is now addressed. Here is what you were actually seeing, one by one.

## 1. "No more output, and closing/reopening exposed no more"

[...]

Issue #3694, opened 2026-08-13
🐛 bug: scrolling the Terminal behaves strangely


MikeSpreitzer:
## What happened?

Running hive commit https://github.com/kubestellar/hive/commit/7e892978405ad7b709f039ac49e856012a707bae in the hive for FMA

When I open the Terminal for the quality agent, I see the following.

<img width="1234" height="888" alt="Image" src="https://github.com/user-attachments/assets/e5a886ff-9f20-4304-b75d-d8d2f726258b" />

I can not scroll down any further. When I scroll up as far as I can, that is not very far. And all it does is make a little more TEMPORALLY EARLIER material appear AT THE BOTTOM. See the following snapshot.

<img width="1243" height="779" alt="Image" src="https://github.com/user-attachments/assets/ff941ccc-476c-4fa3-8a5b-e1b23da02db8" />

## Expected behavior

I expected to be able to scroll up to see earlier output.

## Reproduction steps

1. Open the Terminal of an agent.
2. Try to scroll up and down.

## Environment

- Branch/version: `v4`, 7e892978405ad7b709f039ac49e856012a707ba
- Deployment method: Hive Hub
- Operating system or Kubernetes version: unknown
- Relevant configuration (redact secrets): ?

## Logs or screenshots

See above.

## Additional context

#3693

clubanderson:
Fixed on v4 by the combination of PR #3722 (attach with tmux `mouse on`, so the browser scroll wheel drives copy-mode — earlier output now appears above, in order, instead of reflowing at the bottom) and PR #3790 (agent tmux sessions are now created with a deep `history-limit` — default 50000 lines, override `HIVE_TMUX_HISTORY_LIMIT` — so copy-mode scrollback actually reaches far past tmux's 2000-line default). Closing manually because merges to v4 do not auto-close issues while the repo default branch is v2.

PR #4439, opened 2026-08-21
fix(terminal): say when the terminal is scrolled back, not live (#4399)

Danathar:
Addresses #4399. Not marked as closing it — the third report in that issue is a scheduling question, not a terminal one, and is called out below rather than swept in.

## The terminal wasn't halted. It was scrolled back.

@MikeSpreitzer reported two things that turn out to be the same root cause, and it's measurable rather than a matter of taste.

> No more output appeared for a few minutes ... Closing and re-opening the terminal exposed no more output.

The browser terminal is a live `tmux` attach. #3694 deliberately turned mouse mode **on** so the wheel scrolls history — and the wheel scrolls by entering tmux's **copy-mode**. While a pane is in copy-mode it **stops following live output**. Measured:

```
in copy-mode: pane_in_mode=1
visible tail now:      line 7
visible tail 3s later: line 7      >>> pane STOPPED following live output
```

And copy-mode is **pane state on the tmux server**, not browser state — so closing the tab and reopening re-attaches to the same frozen pane, which is exactly why reopening exposed no more output:

```
after detach + reattach, still in copy-mode: 1
```

> black-on-yellow text in the upper right ... line position/counter information and, usually, a timestamp. It is not obvious which point in the terminal corresponds with that timestamp.

Because it corresponds to none of them, and never can. Two unrelated things sit adjacent in that corner and neither says what it is. Measured on the stock config:

```
status-right  "#{?window_bigger,...}\"#{=21:pane_title}\" %H:%M %d-%b-%y"   ← a WALL CLOCK: the time now
mode-style    noattr,bg=yellow,fg=black                                     ← the black-on-yellow box: SCROLL POSITION
```

This repo ships **no tmux status configuration at all**, so operators were seeing tmux's stock status line. The clock is the time *now*; it can't line up with any line of scrollback.

## The fix

Agent sessions now carry a status line that names the state:

```
[live]                                                        now 14:22:07
[SCROLLBACK - not following live output - press q to resume]   now 14:22:07
```

That answers both reports: a frozen pane announces itself and says how to resume, and the clock is labelled `now` so it isn't read as a timestamp of the content.

Applied at **session creation** (`newSessionCommands`) so a pane carries it from its first frame, and again at **attach** (`ttyd-tmux.sh`) with save/restore — the same defense-in-depth split the existing `history-limit` raise uses, so sessions created before this change or outside the manager get it too, and an agent CLI managing its own status line isn't permanently altered. `status-interval` drops to 2s; tmux's 15s default would leave the SCROLLBACK marker stale long enough for someone to scroll, see nothing change, and conclude it's broken.

The format is deliberately plain ASCII with no `#[...]` style blocks — style specs are comma-separated and a comma inside a `#{?...}` branch needs escaping, which is exactly the subtlety that renders as garbage instead of failing loudly.

## Tested against real tmux, not by grepping the source

A malformed tmux format **does not fail** — it renders wrong, or leaves the literal `#{?...}` on screen. Asserting the string exists in the source would pass on a format that displays nothing useful. So the test sets it on a scratch tmux server and reads back what tmux actually renders:

```
live pane renders:      '[live] now 09:58:51 '
scrolled-back renders:  '[SCROLLBACK - not following live output - press q to resume] now 09:58:51 '
```

**15 assertions**, including reproducing the reported symptom end to end — a copy-mode pane stops advancing, stays in copy-mode across a detach, and resumes when copy-mode is left. The format is **extracted from `manager.go`** rather than restated, and the test asserts `ttyd-tmux.sh` carries the identical string, so the two paths cannot drift. Skips cleanly where tmux is absent; wired into `v2-ci.yml`.

Also adds the troubleshooting entry someone hitting this would actually search for ("The terminal looks frozen — no new output, and reopening it doesn't help"), including that the yellow box is a scroll position and the timestamp is a clock.

`go build ./...`, `go vet`, `gofmt` and the `pkg/agent` session tests all clean, run with the repo's own pinned toolchain.

## Not addressed here — worth its own issue

> the quality agent ... doing some work at 01:47 on Aug 21 ... the dashboard section for that agent said its last run was 8/20 8:12 PM EDT and its next run would be 8/21 2:12 AM EDT.

That's a scheduling/cadence question, not terminal rendering. It needs its own investigation (what triggered the run, and whether the dashboard's "next run" is computed from the same source the scheduler uses), and bolting it onto a display fix would give it neither the right reviewer nor the right test. Happy to open it as a separate issue if you'd like — @MikeSpreitzer, the answer may also be simpler than it looks if something other than the cadence timer kicked that agent.

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -338,6 +338,17 @@ jobs:
         working-directory: .
         run: bash src/deploy/test_contribute_k8s_workload.sh
 
+      # #4399: the browser terminal must say when it is NOT showing live output.
+      # The fix is a tmux FORMAT STRING, and a malformed one does not fail — it
+      # renders as garbage or leaves the literal `#{?...}` on screen. So this
+      # sets it on a scratch tmux server and reads back what tmux actually
+      # renders, in both the live and scrolled-back states, and reproduces the
+      # reported "halt" (a copy-mode pane stops following output, and survives
+      # closing the browser). Skips cleanly where tmux is absent.
+      - name: Terminal scrollback legibility (#4399)
+        working-directory: .
+        run: bash src/deploy/test_terminal_scrollback_status.sh
+
       # #4408: `just contribute-move` relocates an already-registered
       # contributor identity to a new machine. The properties that matter are
       # not visible in the Justfile text — that the positional HIVE_HUB /
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Hive did not historically maintain a complete changelog. This file starts a prag
 
 ### Fixed
 
+- The agent Terminal now says when it is showing scrollback rather than live output. Scrolling with the mouse wheel puts the pane into tmux copy-mode, which stops it following new output — and because that state lives on the server, closing and reopening the terminal returned to the same frozen view, so an idle-looking agent was indistinguishable from a stuck one. The status bar now reads `[live]` or `[SCROLLBACK - not following live output - press q to resume]`, and the clock beside it is labelled `now` so it is not mistaken for a timestamp of the content on screen ([#4399](https://github.com/kubestellar/hive/issues/4399)).
 - `DASHBOARD_AUTH_TOKEN` was documented as holding a Kubernetes secret *name*; it holds the token *value* itself. An operator following the old wording would have set their dashboard token to the literal name of the Secret ([#4427](https://github.com/kubestellar/hive/pull/4427)).
 
 ## 2026-08-21
--- a/src/deploy/test_terminal_scrollback_status.sh
+++ b/src/deploy/test_terminal_scrollback_status.sh
@@ -1,0 +1,164 @@
+#!/usr/bin/env bash
+# The browser terminal must say when it is NOT showing live output (#4399).
+# Run: bash src/deploy/test_terminal_scrollback_status.sh
+#
+# WHY THIS RUNS REAL TMUX. The fix is a tmux FORMAT STRING, and a malformed one
+# does not fail — it renders as garbage, or renders the literal `#{?...}` text,
+# and the only way to find out is to look at a terminal. Asserting the string
+# appears in the source would pass on a format that displays nothing useful. So
+# this sets the option on a scratch tmux server and reads back what tmux
+# actually renders, in both pane states.
+#
+# What #4399 reported, and what each assertion below pins:
+#
+#   "No more output appeared for a few minutes ... Closing and re-opening the
+#    terminal exposed no more output."
+#     -> a pane in copy-mode STOPS following live output, and copy-mode is pane
+#        state on the SERVER, so reopening the browser re-attaches to a pane that
+#        is still frozen. Measured here, both halves.
+#
+#   "black-on-yellow text in the upper right ... contains some line position/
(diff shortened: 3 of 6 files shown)


Danathar:
The `changelog reminder` check is red on this PR. Two separate things, both now handled:

**The reminder's substance was right, and I've acted on it.** This fix is user-visible in the sense the changelog guidance means — an operator would notice, because the symptom it removes is an agent that *looked* stuck and was not. Added an entry under `## Unreleased` → `### Fixed` in 2af0d0a3.

**The check failing is a workflow bug, not something this PR did.** The workflow's own header says it "comments once and always succeeds" and the comment it posts says "it never blocks a merge" — but on a fork PR it does neither:

```
gh: Resource not accessible by integration (HTTP 403)
##[error]Process completed with exit code 1.
```

GitHub caps `GITHUB_TOKEN` at read-only for `pull_request` events from forks regardless of the `permissions:` block, so the comment POST 403s, and `shell: bash -e` turns that into a failed job. It will do this on **every external contribution** — the population least likely to know the convention and most in need of the nudge, who instead get a red check and no reminder.

Filed as #4440 with the run log and a suggested fix (make the post non-fatal; optionally switch to a job summary / `::notice::` so it actually reaches forks). Deliberately not fixing it inside this PR — a CI-workflow change doesn't belong in a terminal-rendering fix, and it wants its own review.

The entry is pushed, so this check should clear on its own terms regardless.

clubanderson:

**Verdict**: This is the right layer for the fix. #4399's "terminal halted" is tmux copy-mode (server-side pane state entered via the #3694 mouse wheel), so the remedy is a tmux status line, not dashboard JS — and indeed **no dashboard JS is touched** (the #4351 render()/statusSeq flow is undisturbed; no aria concerns since nothing DOM-side changed).

**Checks:**
- **Format string sound** — plain ASCII `#{?pane_in_mode,...}` with no `#[...]` style blocks, so no comma-escaping traps. Verified rendering locally with real tmux: live pane → `[live] now 10:05:34`, copy-mode pane → `[SCROLLBACK - not following live output - press q to resume] now …`; no unexpanded `#{` residue.
- **No flicker/auto-scroll fight** — the indicator keys off `#{pane_in_mode}` (actual tmux state), not scroll-position heuristics, so new output while pinned at bottom cannot toggle it. `status-interval 2` bounds staleness.
- **Both creation and attach paths** — set in `newSessionCommands` (before `new-session`, `-g` correctly scoped to the per-agent socket/server) and again on attach in `ttyd-tmux.sh` with symmetric save/restore (same defense-in-depth pattern as the existing history-limit raise), so pre-existing sessions and reconnects are covered.
- **Test is honest** — extracts the format from `manager.go` rather than restating it, asserts manager/attach-path parity, renders both states on a scratch socket, and reproduces the reported symptom (frozen pane, survives detach) plus the remedy. Ran on the PR's own CI (not skipped): **15/15 PASS**; locally: 15/15.
- **Go tests**: `go test ./pkg/agent/... ./pkg/dashboard/...` both ok locally. actionlint: only pre-existing warnings elsewhere in v2-ci.yml; the new step is clean. `bash -n` clean. No forbidden podman lifecycle literals. Docs + changelog entry (2af0d0a3) consistent with the shipped strings.

CI green. 

/lgtm
/approve

PR #4442, opened 2026-08-21
fix(config): scale pack-seeded governor thresholds (#4037)

### LINKED ISSUE #4037: ACMM packs seed explicit governor thresholds, so threshold scaling never engages on pack-applied hives

Danathar:
Closes #4037.

## The hole #3898 documented but could not close

Repo-count threshold scaling shipped with a limitation stated in its own doc comment: ACMM packs seed explicit `governor.thresholds` on apply, "explicit always wins" returns them verbatim, and **nothing distinguished a pack-seeded default from a number an operator typed**. Scaling therefore engaged only on hives that had never applied a level — not the normal path, and plausibly excluding the very 39-repo fleet shape that motivated #3498.

#4037 named the fix: *"the config layer recording **provenance** — pack-seeded vs operator-set — so explicit-wins applies only to the latter."* That is what this does.

```yaml
governor:
  thresholds_source: pack     # written only by the pack-apply paths
  modes:
    surge:
      threshold: 10           # now a per-repo BASE, not an absolute
```

Resolution becomes:

| threshold | before | after |
|---|---|---|
| operator-set explicit | verbatim | verbatim (unchanged) |
| **pack-seeded explicit** | **verbatim — scaling dead** | **per-repo base, scaled** |
| unset | default base, scaled | unchanged |

So a 39-repo hive at L4 surges at `10 × 39 = 390`, and L3's `15/10/3` stays distinct from L4–L6's `10/5/2` rather than collapsing onto one default — the per-level tuning #3898 explicitly refused to throw away.

## Two design calls worth your review

**The marker is whole-set, not per-mode.** Per-mode provenance was the obvious shape and it is a trap. An operator hand-tuning *only* `surge` would leave `busy` and `quiet` still scaling — and a scaled `busy` of 195 on a 39-repo hive sitting above an unscaled `surge` of 30 **inverts the mode ladder**. That is the same inversion `governor-thresholds.md` already warns about, manufactured by the fix meant to help. One marker for the set means the moment you edit any threshold, all three become yours verbatim, which cannot invert. It also keeps a reserved key out of `ModeConfig`'s flat YAML map, where every non-`threshold` key is an agent cadence. A test pins the non-inversion.

**An absent marker reads as operator-set.** Every hive that applied a pack *before* this change has seeded thresholds and no marker. Reading those as pack-seeded would multiply them by the repo count the first time the new code ran — a silent, potentially large mode-ladder change on upgrade. They stay exactly as they are. **Re-applying the level** stamps the marker and turns scaling on, as an explicit operator-initiated act. Fail-quiet on upgrade, opt in to the new behaviour.

## All three write sites

Both pack paths (apply, and the level-change path) stamp the marker; the operator threshold handler **clears** it, so a hand-typed number is honoured verbatim — the #3498 guarantee that `surge: 300` never becomes `11700`.

## Tests

**52 threshold/pack assertions**, covering the bug and each thing that must not regress: the migration case, the 300-never-becomes-11700 guarantee, per-level tuning surviving, `0` still meaning UNSET under both provenances, all three scaling curves applied to pack bases, the single-repo no-op, `idle: 0` (which every pack seeds) still having no threshold, and ladder non-inversion.

Plus a **YAML round trip**: `GovernorConfig` has no custom marshaller today, but `ModeConfig` next door does, and a future one forgetting this field would silently switch scaling back off fleet-wide with nothing to show for it. Also asserts `omitempty` holds, so a hive that never applied a pack does not grow a `thresholds_source: ""` line on its next save.

## Verification

Run with the repo's own pinned toolchain (`golang:1.26-alpine` at the digest `src/Dockerfile` pins), since no Go is installed on this host: `go build ./...`, `go vet` and `gofmt` clean; `pkg/config` and the `pkg/dashboard` pack/governor tests pass.

`pkg/dashboard`'s full suite has **25 pre-existing failures** in this container environment (root, no real uid separation). I checked that by **diffing the failure sets** with and without this change rather than comparing counts — identical counts could hide a swapped set. The sets are byte-identical.

## Docs

`governor-thresholds.md`'s "Known limitation: ACMM packs seed explicit thresholds" section described the hole and a manual workaround (clear the thresholds you want scaled). Replaced with how provenance works, what editing a threshold does, and what upgrading does.


clubanderson:
Reviewed against #4037 and the #3898 scaling machinery. The provenance design is exactly what the issue asked for and both risky edges are handled correctly: an absent marker reads as operator-set so pre-#4037 pack-applied hives keep their exact values on upgrade (no silent multiplication), and the whole-set marker with the dashboard handler clearing it preserves the #3498 surge-300-never-becomes-11700 guarantee while making ladder inversion structurally impossible. ScaleThreshold's existing bounds (base<=0 verbatim, repoCount<1 clamped to 1) mean no divide-by-zero or empty-fleet hazard, and the 52 assertions cover migration, both provenances of zero-means-unset, all three curves, single-repo no-op, unknown-mode idle, non-inversion, and the YAML round trip with omitempty. All three write sites are stamped/cleared consistently, and the applyPack stamp firing only when a threshold is actually written avoids converting an operator's numbers on a no-op merge. Ran the threshold suite locally at the merged head: all pass (the handful of pkg/config failures on my host are pre-existing on the v4 base, environment-only). Resolved the CHANGELOG conflict with v4 by keeping both entries and pushed the DCO-signed merge commit; CI is fully green on the new head.

/lgtm
/approve

PR #4443, opened 2026-08-21
fix(ci): changelog reminder must not fail a fork PR (#4440)

### LINKED ISSUE #4440: 🐛 changelog-reminder fails on every fork PR — an advisory check that blocks, and never delivers the reminder
## Summary

`.github/workflows/changelog-reminder.yml` states its own contract twice — in its header:

> It is ADVISORY ON PURPOSE. [...] a hard gate would fail honest work and train people to ignore it [...] **It comments once and always succeeds**; the judgement stays with the author.

and in the comment it posts:

> This is a reminder, not a gate; it never blocks a merge.

On a **pull request from a fork** it does neither. It cannot post the comment, and it fails the check.

## What happens

Observed on [PR #4439](https://github.com/kubestellar/hive/pull/4439) ([run 32489915826](https://github.com/kubestellar/hive/actions/runs/32489915826/job/96794884705)):

```
gh: Resource not accessible by integration (HTTP 403)
{"message":"Resource not accessible by integration",
 "documentation_url":"https://docs.github.com/rest/issues/comments#create-an-issue-comment",
 "status":"403"}
##[error]Process completed with exit code 1.
```

`changelog reminder  fail  5s`

## Why

For a `pull_request` event raised from a **fork**, GitHub issues a **read-only** `GITHUB_TOKEN` regardless of what the workflow's `permissions:` block asks for. The declared `pull-requests: write` is silently capped, so the final step —

```sh
gh api "repos/$REPO/issues/$PR/comments" -F body=@"$body_file"
```

— is a write the token cannot make, and it 403s. The step runs under `shell: bash -e` plus `set -euo pipefail`, so the non-zero `gh` exit fails the job.

The result is the exact outcome the header set out to avoid: an advisory reminder that behaves as a gate, on honest work, while never delivering the reminder it exists to deliver.

## Who it affects

**Every external contribution.** Fork PRs are the population least likely to know the changelog convention and most in need of the nudge — and they are precisely the ones that get a red check and no explanation. Same-repo branches are unaffected, which is why this was not visible when the workflow landed.

It also erodes the signal: a check that is red for reasons unrelated to the change trains reviewers to ignore red checks.

## Suggested

Two independent parts:

1. **Never fail.** Make the comment step non-fatal, so the stated contract holds whatever the token can do:

   ```sh
   gh api "repos/$REPO/issues/$PR/comments" -F body=@"$body_file" \
     || echo "::notice::changelog reminder could not post (fork PRs get a read-only token) — skipping"
   ```

   This alone fixes the red check and costs nothing.

2. **Actually reach fork PRs**, if the reminder is worth delivering to them. The usual options, in rough order of preference:
   - Emit a **job summary** and/or `::notice::` annotation instead of a PR comment. Visible on the PR's checks, needs no write permission, and works identically for forks and branches. This looks like the best fit for something explicitly advisory.
   - Move to `pull_request_target`, which does get a writable token — but it runs against the *base* ref with secrets available, so it needs care; for a workflow that only reads a file list it is probably more risk than the feature warrants.

The first part is the bug fix. The second is a design choice about whether the reminder should reach forks at all, and is worth deciding deliberately rather than by accident of token scope.

## Notes

Found while checking CI on #4439, which does not touch `CHANGELOG.md`. Worth saying: the reminder's *substance* was right — that PR is user-visible and now carries an entry. The problem is only that an advisory check failed and the advice never arrived.

Danathar:
Closes #4440.

## The contract the workflow states, and did not keep

`changelog-reminder.yml` says it twice — in its header:

> It comments once and **always succeeds**; the judgement stays with the author.

and in the comment it posts:

> This is a reminder, not a gate; it never blocks a merge.

On a pull request **from a fork** it did neither. GitHub caps `GITHUB_TOKEN` at read-only for `pull_request` events raised from forks, whatever the `permissions:` block asks for, so the final step answered:

```
gh: Resource not accessible by integration (HTTP 403)
##[error]Process completed with exit code 1.
```

and `shell: bash -e` + `set -euo pipefail` turned that into a failed check. The result was exactly what the header set out to avoid — an advisory reminder behaving as a hard gate, on honest work — and on **every external contribution**, the population least likely to know the changelog convention and most in need of the nudge. They got a red check and no reminder. (Observed on #4439, [run 32489915826](https://github.com/kubestellar/hive/actions/runs/32489915826/job/96794884705).)

## Two things were wrong; both fixed

**Delivery.** The reminder now goes to the **job summary** and a **`::notice::` annotation** first. Neither needs write permission, so both reach fork PRs and same-repo branches identically — this is the delivery the reminder can actually rely on. The PR comment becomes a nicety layered on top: skipped outright on a fork (a 403 in the log invites the next reader to debug a permission that is working as designed), attempted elsewhere, and tolerated if an org policy narrows the token anyway. The once-only dedupe is unchanged for the comment path, and the HTML marker is kept out of the summary.

**Exit status.** The step drops `errexit` and ends `exit 0`. `nounset` and `pipefail` stay — the intent is "a failed delivery is not a failed PR", not "ignore everything".

Same-repo PRs see no behaviour change beyond the added summary: still commented, still only once.

## Tested by executing the step, not grepping the YAML

The bug was an **exit status**. A grep over the workflow would pass on a script that still dies on the first non-zero command, so `src/deploy/test_changelog_reminder.sh` **extracts the real `run:` block out of the workflow** — never restating it, so a copy cannot keep passing after the step regresses — and runs it against a stub `gh` that fails exactly the way GitHub does.

14 assertions across three shapes:

| shape | asserted |
|---|---|
| fork PR (the #4440 case) | exits 0; reminder delivered as summary **and** annotation; no doomed POST attempted; dedupe marker kept out of the summary |
| same-repo PR | still posts the comment; still writes the summary, so both surfaces agree |
| same-repo PR, POST refused by org policy | tolerated, not fatal; says where the reminder went instead |

Plus the property that caused it: the step must not run under `errexit`, while keeping `nounset`/`pipefail`.

**Mutation-checked.** Run against the *unfixed* workflow, **10 of the 14 fail** and reproduce the reported error verbatim:

```
FAIL: a refused POST must not fail the job
      rc=1, output: gh: Resource not accessible by integration (HTTP 403)
```

So they test the bug, not the implementation.

## One gap closed on the way

`changelog-reminder.yml` is added to `v2-ci.yml`'s path filters. Without it, a change to the file this guard protects would not run the guard — the same class of gap #4414 closed for `test_ambient_cap_runtime.sh`, which was missing from its own job's filters.

No changelog entry: this is CI-only, which the changelog's own guidance excludes — and the reminder agrees, since its `code=` check treats `.github/` as not user-facing.

### DIFF
--- a/.github/workflows/changelog-reminder.yml
+++ b/.github/workflows/changelog-reminder.yml
@@ -14,6 +14,16 @@ name: Changelog Reminder
 # dependency churn, so a hard gate would fail honest work and train people to
 # ignore it — or worse, to add noise entries just to make it green. It comments
 # once and always succeeds; the judgement stays with the author.
+#
+# "Always succeeds" is load-bearing and was not true (#4440): on a fork PR the
+# comment POST 403s, because GitHub caps GITHUB_TOKEN at read-only for
+# `pull_request` events from forks whatever the permissions: block requests, and
+# `bash -e` turned that into a red check on every external contribution. The
+# reminder is now delivered as a JOB SUMMARY and a ::notice:: annotation, which
+# need no write permission and therefore reach forks and same-repo branches
+# alike; the PR comment is attempted on top of that and its failure is
+# tolerated. src/deploy/test_changelog_reminder.sh executes the real step script
+# against a stubbed `gh` that 403s and asserts it still exits 0.
 
 on:
   pull_request:
@@ -64,16 +74,25 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR: ${{ github.event.pull_request.number }}
           REPO: ${{ github.repository }}
+          IS_FORK: ${{ github.event.pull_request.head.repo.fork }}
         run: |
-          set -euo pipefail
+          # NOT `set -e` (#4440). This step delivers a REMINDER, and the header
+          # above promises it "always succeeds" and "never blocks a merge". It
+          # did neither on a pull request from a FORK: GitHub caps GITHUB_TOKEN
+          # at read-only for `pull_request` events raised from forks, whatever
+          # the permissions: block asks for, so the comment POST answered
+          #
+          #   gh: Resource not accessible by integration (HTTP 403)
+          #
+          # and `shell: bash -e` turned that into a failed check. The result was
+          # the exact thing the header set out to avoid — an advisory reminder
+          # behaving as a gate — on every external contribution, which is the
+          # population least likely to know the convention and most in need of
+          # the nudge. They got a red check and no reminder.
+          set -uo pipefail
+
           marker='<!-- changelog-reminder -->'
-          # Only ever comment once per PR. A reminder that reappears on every
-          # push is noise, and noise is how the previous convention died.
-          if gh api --paginate "repos/$REPO/issues/$PR/comments" --jq '.[].body' \
-              | grep -qF "$marker"; then
-            echo "already reminded"
-            exit 0
-          fi
+
           # Body built in a heredoc rather than inline: the text starts a line
           # with '**', which YAML would read as an alias in a plain scalar.
           body_file="$(mktemp)"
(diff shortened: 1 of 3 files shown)


clubanderson:
Reviewed and executed, not just read. The security posture is exactly right: the event stays pull_request (never pull_request_target), only secrets.GITHUB_TOKEN is referenced, and the fix works within the fork read-only cap instead of escalating around it — summary + ::notice:: need no write permission, so the reminder finally reaches the population it was written for. Same-repo PRs keep the comment path with the once-only dedupe intact (asserted), so nothing is silently disabled. Ran src/deploy/test_changelog_reminder.sh at the head: 14/14 pass, and the extract-the-real-run-block approach means a regression cannot hide behind a stale copy. actionlint is clean on changelog-reminder.yml; the two v2-ci.yml shellcheck warnings it reports are pre-existing on the v4 base at an untouched line (488). One residual worth knowing, not blocking: even with the guarded conditions, GitHub's default run shell is bash -e, but every fallible command in the step now sits in an if-condition and the script ends exit 0, so errexit has nothing to fire on — the test's fork-403 case proves it end to end. Path-filter addition to v2-ci.yml closes the guard-not-guarding-itself gap correctly. CI green.

/lgtm
/approve

Issue #4413, opened 2026-08-21
🧪 Podman: execute the reboot-persistence case in both root modes


Danathar:
## Parent

Part of #4188. Closes the one row #4377 left open; it must not close the parent.

## Why

`src/docs/podman-quadlet-lifecycle.md` records this honestly rather than
inferring it:

> ## Reboot persistence: NOT executed here
>
> | rootless, lingering enabled, actual reboot | **not executed** |
> | rootful via the system manager, actual reboot | **not executed** |
>
> The host this was measured on could not be rebooted: it was the host running
> the session doing the measuring, and a reboot would have ended it mid-run.

#4377 anticipated this in its stop condition and the row was recorded, not
faked. `is-enabled` was explicitly **not** accepted as a substitute.

#4188 requires "Persistent data survives ... host reboot" and "Boot persistence
... tested for the chosen Podman lifecycle." Everything on either side of the
boot is established — the generator installs the `default.target.wants/` symlink
in both modes, `default.target` is reached in both managers, and lingering
starts a sessionless user manager and its enabled units in 305ms while its
absence stops them. What is unproven is the composition of those links across an
actual kernel boot.

## 30-minute boundary

Execute the reboot case in both root modes and record it. No unit changes unless
the reboot exposes a defect — if it does, file that separately rather than
fixing it here.

## Deliverable

Replace the "NOT executed" section of `src/docs/podman-quadlet-lifecycle.md`
with measured results, in the same observed-vs-expected style.

The procedure is already written at the end of that section:

```sh
bin/hive-podman-lifecycle-probe.sh check              # confirm the wiring first
bin/hive-podman-lifecycle-probe.sh check --rootful
sudo systemctl reboot
```

## Acceptance criteria

- Both rows filled: rootless **with lingering enabled**, and rootful via the
  system manager. Lingering is the whole point of the rootless row — a rootless
  result without it proves the negative case, not the positive one.
- The service is confirmed **healthy** after boot, not merely `active`. This
  unit is `Notify=healthy`, so "started" already means "serving" — say how long
  it took to get there from boot.
- `hive-data` contents confirmed across the reboot with a marker file, the way
  #4378 proved it across update and rollback.
- If a mode does not come back, that is the finding — record it and file the
  defect. Do not adjust the unit until it passes.

## Where to run this

A host that can actually be rebooted, which is why this stayed open. Not the
machine running the session doing the measuring.

## Related

[...]

PR #4445, opened 2026-08-21
docs(podman): measure the host-execution capability matrix (#4441)

### LINKED ISSUE #4441: 🧪 Podman: establish the host-execution capability matrix
## Parent

Related to #4188 and #4413; closes neither. Stands on its own as documentation — the capability matrix is useful regardless of what is built on top of it. It also establishes the evidence for a proposal to give host-requiring tasks a disposable VM, which will be filed separately as a tracker; that proposal should be judged against what this task measures, not the other way round.

## Why

#4188 requires "Boot persistence ... tested for the chosen Podman lifecycle." `src/docs/podman-quadlet-lifecycle.md` records both reboot rows as **NOT executed**, and #4413 quotes the reason:

> The host this was measured on could not be rebooted: it was the host running the session doing the measuring, and a reboot would have ended it mid-run.

Before proposing any new execution runtime, the claim "the current execution paths cannot run this work" should be **measured rather than asserted**. That is what this task produces. Failure is the expected result for several rows — recording precisely *where* and *how* each path stops is the deliverable, not a problem to fix.

## Scope

Probe the execution environment this task is running in and record, per capability, what is available and what the exact failure is when it is not. Cover at minimum:

| Capability | Why it matters |
|---|---|
| `/dev/kvm` present and usable | whether a guest could be launched here at all |
| systemd available as PID 1 / `systemctl` usable | quadlet units are systemd units |
| `systemctl reboot` outcome | the blocked row in #4413 |
| user lingering (`loginctl enable-linger`) | rootless boot persistence |
| SELinux mode, and whether the policy is the container's or the host's | `podman-selinux-avc-evidence` |
| rootful podman availability | "first live rootful start", #4377 |
| kernel module load (`modprobe`) | host-level networking cases |
| network admin capability (`NET_ADMIN`) | `net-admin-requirement` |

For each row record: the exact command, its exit status, and its verbatim output (trimmed). An error message is the data. Where a capability is present, say so plainly — a surprise in that direction is the most valuable outcome this task can have.

## 30-minute boundary

Probe and record only.

- **Do not** implement a VM launcher, a new `CONTRIBUTOR_MODE`, or any change to `bin/contributor-relay.sh` or `pkg/sandbox`.
- **Do not** attempt to make any failing capability work, install packages to change an outcome, or reconfigure the host.
- **Do not** execute the #4413 reboot case itself — that is #4413's task, and it is expected to be unrunnable here. Recording *that* it is unrunnable, and the exact error, is in scope.
- If a probe reveals a defect in existing code, file it separately rather than fixing it here.

Running the probes should take a couple of minutes; the rest is writing them up accurately.

## Deliverable

A new `src/docs/podman-host-execution-gap.md` containing:

1. A one-paragraph statement of which execution path was measured — contributor relay in container mode, relay in `local` mode, or `agent_sandbox` rootless Podman — and how that was determined. If it cannot be determined with certainty, say so and record the observable facts instead of guessing.
2. The capability table above, filled in, with command / exit status / verbatim output per row.
3. A short "what this does and does not establish" section. One environment is one data point: it does not establish what a *different* path would do, and the write-up must not generalize past what was measured.

Link the doc from `src/docs/README.md` in the same PR.

## Acceptance

- Every row carries a real captured command and output, not a description of one.
- No row is inferred from another row (`is-enabled` is not evidence of boot survival — the precedent #4377 set).
- The PR changes documentation only.

Danathar:
Closes #4441. Related to #4188 and #4413; closes neither.

## The result is not the expected one

#4441 said the most valuable outcome this task could have was a surprise in the "capability is present" direction. That is what happened, and it changes the premise the task was filed on.

| Capability | Available? | How it was shown |
| --- | --- | --- |
| `/dev/kvm` | **yes** | opens read-write |
| systemd / `systemctl` | **yes** | user manager `running`; system manager `degraded` |
| `systemctl reboot` | **permitted** | polkit authorises; `--dry-run` exits 0 |
| user lingering | **authorised, currently off** | `Linger=no`; polkit permits |
| SELinux | **enforcing, the HOST's policy** | `targeted`, `unconfined_t` |
| rootful podman | **yes** | passwordless `sudo`; `Rootless=false` |
| `modprobe` | present; **not loadable as this user** | `CapEff` empty; `sudo` available |
| `NET_ADMIN` | not held; **obtainable via user namespace** | `CapEff=0`, `CapBnd` full, `unshare -rn` works |

Every row is a command that was really run, with its real exit status and verbatim output. No row is inferred from another — the precedent #4377 set when it refused `is-enabled` as evidence of boot survival.

## The finding that matters most

**Reboot is authorised here.** polkit permits it and `systemctl --dry-run reboot` exits 0.

So #4413 is unrunnable in this environment for a **topology** reason — the session runs on the host that would restart — and **not** because reboot is forbidden. That corrects the assumption #4441 itself carried, and the two have entirely different remedies: one wants a *different host*, the other would want a *permission*. Any disposable-guest proposal should be argued against that distinction.

Two supporting facts point the same way: `/dev/kvm` **opens read-write** on a host that `systemd-detect-virt` reports as `kvm` itself (so nested virtualisation is available), and passwordless `sudo` works, so rootful podman is reachable — meaning the "first live rootful start" #4377 wanted is not blocked by privilege here either.

## Which path was measured, established not guessed

The contributor relay in **`local` (native host) mode**, identified by direct evidence: the running relay's own `HIVE_AGENT_SESSION` is **the tmux session this probe ran in**, so this process is the agent CLI that relay launched — not a coincidental neighbour on the same box. Not container mode and not `agent_sandbox`: no `containerenv`, `systemd-detect-virt --container` returns `none`, PID 1 is the host's systemd.

## One trap worth the space

```
$ id -Z
id: --context (-Z) works only on an SELinux/SMACK-enabled kernel   [exit 1]

$ sestatus
Current mode:                   enforcing                          [exit 0]
```

`id -Z` reports SELinux as **unavailable on a host where it is enforcing**, because `id` here is **uutils coreutils**, which does not implement `-Z`. That is the same class of trap `bin/hive-podman-preflight-host.sh` already documents for uutils `stat -c %C`. Any probe reaching for a GNU-coreutils-specific flag will mis-read this host; `/proc/self/attr/current` and `sestatus` do not.

## Scope

Probe and record only, per the 30-minute boundary. **No reboot, no `loginctl enable-linger`, no module load, no package installed, no host reconfiguration.** Where a row would have needed a mutation to prove more, it says so rather than performing it — `modprobe -n` is explicitly called out as proving the module *resolves*, not that loading is permitted.

The reboot row was probed with polkit (a pure query) and `systemctl --dry-run`, whose presence I confirmed in `systemctl --help` before running it. Nothing was rebooted; `uptime` afterwards confirms the host is the same boot.

## What it deliberately does not claim

The write-up's closing section is blunt that one environment is one data point. It establishes nothing about container mode, nothing about `agent_sandbox`, nothing about another host (`/dev/kvm` is commonly absent on hosted runners), and **nothing about boot survival** — lingering being *authorisable* is not evidence that units come back, which only #4413 can settle. It is evidence for judging a proposal, not a proposal.

Documentation only; linked from `src/docs/README.md`. All eight cross-references verified to resolve.


clubanderson:
Reviewed from head 5c0a0460. Doc-only (+329 lines plus the index blurb), and it clears the measured-not-asserted bar comfortably.

- **Methodology reproducible**: every row is command + real exit status + verbatim output, no row inferred from a neighbour — including the discipline of noting non-obvious exit codes (`systemctl is-system-running` → `degraded`/1, `systemd-detect-virt --container` → `none`/1) so a reader re-running them isn't surprised.
- **The topology-not-permission finding is genuinely valuable**: reboot being *authorised* (polkit + `--dry-run` 0) while #4413 stays unrunnable because the session lives on the host that would restart re-frames the disposable-guest question around the right remedy. The identity proof (relay's own `HIVE_AGENT_SESSION` matching the probe's tmux session) is direct evidence, correctly preferred over process-tree inference.
- **Consistent with the merged stack**: host-process capabilities kept strictly distinct from container grants (`CapBnd` called "the ceiling, not a grant" — no conflict with #4414's boot test); the SELinux section explicitly disclaims applicability of #4353's container-labelling cases; the uutils `id -Z` trap matches the exact class preflight-host already documents for `stat -c %C` (verified at its lines 276/406); and the #4377 precedent — authorisation is not evidence of boot survival — is honoured in the lingering row.
- **Scope discipline**: no mutation performed anywhere (no reboot, no linger, no module load), `modprobe -n` correctly qualified as resolution-only, and the does-not-establish section refuses all four tempting generalisations.
- All four cross-referenced docs verified present; no forbidden phrases; no tests to wire (pure measurement record).

/lgtm
/approve

Issue #4400, opened 2026-08-21
🐛 bug: scanner agent in FMA hive is suffering a mysterious failure


MikeSpreitzer:
## What happened?

In the hive for llm-d FMA the scanner agent keeps crashing, saying that a /login is needed but reporting that inference server refuses the chosen model. But this agent is configured to use the same model ("claude-sonnet-4-5") as the quality agent, which is not suffering this problem.

## Expected behavior

All agents using the same model get the same access to the model.

## Reproduction steps

I do not know how to reproduce this inconsistency.

## Environment

- Branch/version: `v4` / commit 9f31439295c6fe8b70df70f8f16ea76afd57a063
- Deployment method: Hive Hub
- Operating system or Kubernetes version: unknown
- Relevant configuration (redact secrets): ?

## Logs or screenshots

[hive-scanner-20260821-062424.log](https://github.com/user-attachments/files/31292627/hive-scanner-20260821-062424.log)

clubanderson:
@MikeSpreitzer — root cause found. The asymmetry is real and it is a **model-id spelling drift**, not an access-policy difference.

## What the log shows

Your attached log gives it away in two places:

1. The scanner launched with `HIVE_MODEL='claude-sonnet-4.6'` (and `claude --model claude-sonnet-4.6`) — the **dotted** spelling, not `claude-sonnet-4-5`.
2. The gateway's 403 body lists the team's entitled set: `['gemini-2.5-pro', 'gemini-2.5-flash', 'gcp/gemini-3.1-pro-preview', 'aws/claude-sonnet-4-6', 'aws/claude-opus-4...']` — note `aws/claude-sonnet-4-6`: **dashed, with an `aws/` provider prefix**.

LiteLLM matches the request model string **verbatim** against the team's entitled set. `claude-sonnet-4.6` ≠ `aws/claude-sonnet-4-6` as strings, so every scanner call is refused — the same underlying model, refused purely on spelling.

## Why the two agents differ

The scanner is one of the shipped default agents, whose baked-in config carries the copilot-CLI spelling `claude-sonnet-4.6`. When its backend was routed to the FMA LiteLLM gateway, that stored string went to the gateway verbatim (hive deliberately never rewrites inference model ids — exactness is load-bearing). The quality agent was configured later with an id taken from the gateway's discovered/entitled list, so its string matches and it works.

[...]

PR #4446, opened 2026-08-21
fix(agent): a gateway model 403 is not a login prompt (#4400)

Danathar:
Addresses #4400. The attached log explains the whole chain, and the middle link is hive's.

## What the log says

The scanner launched with `HIVE_BACKEND='litellm'` and **`HIVE_MODEL='claude-sonnet-4.6'`**, and the gateway answered:

```
403 team not allowed to access model. This team can only access
models=['gemini-2.5-pro', 'gemini-2.5-flash', 'gcp/gemini-3.1-pro-preview',
        'aws/claude-sonnet-4-6', 'aws/claude-opus-4...]
```

```
configured:  claude-sonnet-4.6
entitled:    aws/claude-sonnet-4-6
                 ^^^^         ^
```

A dot where the gateway wants a hyphen, and no `aws/` prefix. **So the two agents are not on the same string** — that is the answer to "all agents using the same model get the same access". They were never asking for the same model.

Worth stating plainly: **hive did not corrupt this id.** `normalizeModelName` excludes inference backends and returns them verbatim — deliberately, because the gateway is the only thing that knows its own naming — and it already did so at the reporter's exact commit `9f31439`, which I checked rather than assumed. The model id is what the agent is configured with.

## What *was* hive's bug — and it is the "mysterious" part

Claude Code prefixes **every** API error with its login hint, so that refusal rendered as:

```
● Please run /login · API Error: 403 {... "team not allowed to access model" ...}
```

`paneShowsLoginPrompt` matched `"Please run /login"`, set `NeedsLogin`, and the pane poller's `showsLogin && configHasTokens()` branch **auto-restarted the agent** — back into the same 403, every time. That restart loop is the reported "keeps crashing", and the operator was pointed at the one action that could not possibly help while the real cause sat unread in the same line.

## The fix

Key on the HTTP status, not on one gateway's wording:

| status | meaning | `/login` helps? |
| --- | --- | --- |
| **401** | authentication — caller is not identified | **yes** |
| **403** | authorization — caller *is* identified, not permitted | **no** |

A line carrying an upstream authorization failure is skipped by the login detector. This is the same shape as the guard already in `lineHasLoginDirective` — which exists so `POST /login returns 302` is not read as a login screen. Claude Code's error decoration is the same class of false positive.

**401 is deliberately still detected.** Suppressing it would trade one misdiagnosis for its exact opposite, and a test pins that. The gateway's own wording is kept as a second, independent signal for gateways that don't surface a status through the CLI's error line.

## Tests

Built from the reporter's log **verbatim**, including the trimmed model list. Run against the unguarded detector, both 403 cases fail with *"read as a login prompt"* — so they test the bug, not the implementation.

`pkg/agent` has 7 pre-existing failures in this container environment; I compared the failure **sets** with and without the change by diff rather than by count, and they are byte-identical.

Two of my own tests failed on first run because they asserted strings from config's `defaultLoginPatterns` rather than the agent package's `loginPromptPatterns` — two separate mechanisms. The tests were wrong, not the code; they now assert against the list this detector actually reads.

## Also

A troubleshooting entry for what an operator hitting this would search for: how to tell 401 from 403, that hive passes inference model ids through verbatim so the fix belongs in the agent's `model:` setting, and why "the same model" on two agents can differ by one character.

## What this does not fix

**The scanner's model id is still wrong** and will still 403 until it is corrected to whatever the gateway entitles (`aws/claude-sonnet-4-6`, on the evidence in the log). Hive cannot fix that for you — rewriting an inference model id is exactly what produces a model the team is not entitled to, which is why the code refuses to. What changes here is that hive now stops mislabelling it as a login problem and restart-looping the agent, so the real cause is visible.

@MikeSpreitzer — comparing the two agents' configured `model:` values directly (rather than the model each was *meant* to use) should show the difference immediately.


clubanderson:

**Verdict**: Correct diagnosis and a surgical fix. Claude Code prefixes *every* API error with "Please run /login", so a LiteLLM `403 team not allowed to access model` matched the login detector, badging a fully-logged-in agent 🔑 and — via the poller's `showsLogin && configHasTokens()` branch — auto-restarting it straight back into the same refusal. That restart loop *is* the "keeps crashing" in #4400, and it also means this fix ends the retry storm at its source: with a 403 no longer read as a login signal, the auto-restart branch simply never fires (and the pre-existing restart cooldown remains as a second line of defense).

**Checks:**
- **Classification is status-anchored, not wording-anchored** — `API Error: 403` ⇒ authorization; the LiteLLM phrasing patterns are a fallback for gateways that don't surface a status. `401` is deliberately untouched and pinned by test (`TestUnauthenticated401StillCountsAsLogin`) — no trade of one misdiagnosis for the opposite one. 500s are neither, per test.
- **Per-line skip, not per-pane** — a genuine standalone login prompt elsewhere in the same pane still triggers detection; only the decorated error line is excluded (`TestModelEntitlement403SurroundedByOrdinaryOutput` + `TestGenuineLoginPromptStillDetected` cover both directions).
- **Operator message is actionable and accurate** — the troubleshooting section teaches the 401-vs-403 distinction, walks the exact separator/prefix comparison (`claude-sonnet-4.6` vs `aws/claude-sonnet-4-6`), and explains *why* one agent fails while "the same model" works — which is precisely the reporter's confusion. Verbatim-passthrough of model ids is correctly defended.
- **Tests**: both classification paths covered plus false-positive guards and empty-pane edge. Ran targeted locally: all pass, including the pre-existing `TestPollTmuxOutputForAgent_LoginPrompt` integration path. Full `./pkg/agent/` coverage measured **91.7%** ≥ 89 floor. build/vet/gofmt clean on PR files. No forbidden podman lifecycle literals.

CI green.

/lgtm
/approve

PR #4449, opened 2026-08-21
docs(troubleshooting): Podman parity for the Quadlet deployment (#4444)

### LINKED ISSUE #4444: 📝 Podman: add troubleshooting parity for the Quadlet deployment
## Parent

Part of #4188 (second wave, bullet 14). This issue must not close the parent.
It is the **last unfiled bullet** on that list.

## Why

`src/docs/troubleshooting.md` contains **zero** occurrences of "podman" and
**seven** Docker references — and they are the ones an operator reaches for
first, when something is already wrong:

```
src/docs/troubleshooting.md:12   docker compose -f src/docker-compose.yaml ps
src/docs/troubleshooting.md:13   docker compose -f src/docker-compose.yaml logs hive --tail=200
src/docs/troubleshooting.md:14   docker compose -f src/docker-compose.yaml logs gateway --tail=100
src/docs/troubleshooting.md:227  docker compose -f src/docker-compose.yaml down -v
src/docs/troubleshooting.md:228  docker compose -f src/docker-compose.yaml up -d
```

Line 9 states the framing outright — *"Docker Compose deployments define `hive`
and `gateway` services in `src/docker-compose.yaml`"* — and lines 227–228 are
the **reset procedure**. A Podman operator following this doc runs commands that
do not exist on their host, at the moment they are least able to absorb that.

The rest of bullet 14 has since been covered by its own documents:
`podman-quadlet-update-rollback.md` (#4378), `podman-auto-update.md` (#4411),
and the Podman sections of `backup-restore.md` (#4406). Troubleshooting is what
is left, and there is now somewhere for every pointer to go.

## Also in scope, and small

`src/docs/auto-update-profile.md` has **0** Podman mentions. It documents the
Watchtower profile, which #4188 keeps deliberately Docker-only — so this is not
a request to make it cover Podman. It needs one line saying the Podman
equivalent is `podman-auto-update.md`, so a Podman operator who lands there
knows they are in the wrong document rather than concluding the feature is
missing.

## 30-minute boundary

Troubleshooting parity only. Do not restructure the document, do not rewrite the
Docker commands, and do not move content into new files. Docker stays the
default and its instructions stay exactly where they are.

## Deliverable

Podman equivalents alongside the existing Docker commands in
`src/docs/troubleshooting.md`, plus the one-line cross-reference in
`auto-update-profile.md`.

## Acceptance criteria

- Each of the Docker commands above has a stated Podman counterpart. `logs` and
  status are `systemctl`/`journalctl` against the units, not `podman logs`
  alone — the unit state is the thing that matters for `Notify=healthy`, and it
  is where a failed start actually shows.
- The reset procedure (lines 227–228) has a Podman counterpart, and the
  destructive step is **marked destructive**. `down -v` deletes the volume;
  whatever stands in for it must not read as routine.
- Covers the failure modes this runtime actually has, which the Docker doc has
  no reason to mention:
  - a unit stuck `activating` for the full `TimeoutStartSec` — the measured trap
    from #4367, where a config/HealthCmd port mismatch costs 300 silent seconds
    and `--rm` deletes the container that held the evidence;
  - `systemctl is-failed` reporting `activating` rather than `failed` during a
    bad start, recorded in #4378 — monitoring keyed on `failed` does not fire;
  - `systemctl cat hive.service` not showing the `hive.container.d` drop-in,
    because Quadlet merges before generating (#4378);
  - SELinux denials on `/data` or the secrets directory, pointing at
    `podman-selinux-avc-evidence.md` rather than restating it.
- Says which root mode each command assumes, or works in both.
- Points at `bin/hive-podman-preflight*.sh` for the "it will not start at all"
  case, since preflight already diagnoses most of those causes.
- Adds no new Podman-only *content* that belongs in the existing Podman docs —
  link to them.

## Related

- #4188 — the umbrella; bullet 14
- #4367 — the 300-second silent hang, worth a named entry
- #4378 / #4411 — the `failed`-state and drop-in-visibility traps
- `src/docs/podman-standalone-quadlet.md` — the install-side "Traps" section,
  which this should complement rather than duplicate

Danathar:
Closes #4444. Part of #4188 (bullet 14, the last of it); does not close the parent.

## The gap

`src/docs/troubleshooting.md` carried **zero** occurrences of "podman" and ten Docker references — including the reset procedure. A Podman operator following it ran commands that do not exist on their host, at the moment they were least able to absorb that.

Docker stays the default and its instructions stay exactly where they are. Nothing restructured, rewritten, or moved.

## What was added

**Podman equivalents beside the Docker and Kubernetes blocks** in "Find the running service logs" — and they are `systemctl`/`journalctl` against the units, not `podman logs`, with the reason stated:

> `hive.container` sets `Notify=healthy`, so the **unit's** state — not the container's — is what tells you Hive is *serving*. And a start that times out is deleted: the generated `ExecStart` carries `--rm`, so a failed start leaves no container for `podman logs` to read, while the journal still holds why.

**A Podman counterpart for the reset**, with the destructive step isolated and marked:

```bash
systemctl --user stop hive-gateway.service hive.service
#  ⚠ DESTRUCTIVE — this line deletes all persisted Hive state. Everything
#    else here is reversible; this is not. Back up FIRST.
podman volume rm hive-data
systemctl --user start hive-gateway.service
```

The page says plainly that `systemctl stop` alone does **not** delete the volume — units can be stopped and started freely without losing state — because that is the question an operator reaching for `down -v` is actually asking.

**A section for the failure modes this runtime has and Docker does not:**

| Failure mode | What the entry gives |
|---|---|
| unit stuck `activating` for the full `TimeoutStartSec` | the #4367 trap — `dashboard.port: 3001` against the unit's 3002 `HealthCmd` costs 300 silent seconds and `--rm` deletes the evidence; plus the missing `HIVE_DASHBOARD_TOKEN` case, which presents identically |
| `systemctl is-failed` says `activating`, not `failed` | #4378 — monitoring keyed on `failed` does not fire; names `Result=timeout` and `NRestarts` as what *does* move, and warns a single `NRestarts=0` sample proves nothing |
| `systemctl cat` not showing the drop-in | #4378 — Quadlet merges `hive.container.d/*.conf` before generating, so `cat` prints the merged result and never names the file |
| SELinux denials on `/data` or secrets | points at `podman-selinux-avc-evidence.md` and `podman-volume-persistence.md` rather than restating; notes a private-category denial records **no AVC at all** |
| auto-update rolled back or silently did nothing | points at `podman-auto-update.md` |

The "it will not start at all" case leads with the preflights — including that `HIVE_DEPLOY_RUNTIME=podman` must be exported first, or all three exit 0 having checked nothing.

Root mode is stated once for the section and `$CONF` is defined there, so the commands are copy-pasteable in either mode instead of silently assuming rootless.

## `auto-update-profile.md`

The one line #4444 asked for. A Podman operator landing there is told the page does not apply, that the profile is deliberately Docker-only per #4188, and where the equivalent is — so the feature does not read as missing.

## Verification

Documentation only (two files). Every link target in both edited docs verified to resolve programmatically — including that `podman-auto-update.md` is on `v4` and not just in an unmerged branch, which I checked after making exactly that mistake in #4423. `podman` mentions in `troubleshooting.md`: 0 → 16. The Quadlet config-coupling, runtime-parity and port-boundary guards all still pass.

No new Podman-only *content* that belongs in the existing Podman docs — those are linked, not restated, per the acceptance criteria.

## Note on overlap

My open **#4446** also edits `troubleshooting.md`, inserting a section before "The terminal looks frozen". This PR's edits are in different regions (the logs section near the top, and the reset section at the bottom), so they should merge cleanly in either order — flagging it since it is the same file.

### DIFF
--- a/src/docs/auto-update-profile.md
+++ b/src/docs/auto-update-profile.md
@@ -11,6 +11,8 @@ It is **opt-in** and off by default:
 docker compose --profile auto-update up -d
 ```
 
+> **Running Podman?** This page does not apply to you, and the feature is not missing. The profile is built on the Docker socket and a filtered Docker API proxy, and [#4188](https://github.com/kubestellar/hive/issues/4188) keeps it deliberately Docker-only. The Podman equivalent is opt-in health-aware auto-update through `podman auto-update` — see [podman-auto-update.md](podman-auto-update.md).
+
 **Read this whole page before enabling it in production.** The profile trades a
 real, permanent increase in blast radius for the convenience of unattended
 updates, and the trade is worse on a hive host than on most hosts, because this
--- a/src/docs/troubleshooting.md
+++ b/src/docs/troubleshooting.md
@@ -22,6 +22,17 @@ kubectl -n hive logs deploy/hive --tail=200
 kubectl -n hive describe pod -l app.kubernetes.io/name=hive
 ```
 
+Podman (Quadlet) deployments run the same two containers as **systemd units**: `hive.service` and `hive-gateway.service`, generated from `src/deploy/quadlet/` (plus `hive-network.service` and `hive-data-volume.service`). Ask systemd first, not `podman logs`:
+
+```bash
+# rootless — drop --user for a rootful install
+systemctl --user status hive.service hive-gateway.service
+journalctl --user -u hive.service -n 200 --no-pager
+journalctl --user -u hive-gateway.service -n 100 --no-pager
+```
+
+**Why the unit and not the container.** `hive.container` sets `Notify=healthy`, so the unit stays in `activating` until `/api/health` answers — the unit's state, not the container's, is what tells you whether Hive is *serving*. And a start that times out is deleted: the generated `ExecStart` carries `--rm`, so a failed start leaves no container for `podman logs` to read, while the journal still holds why. `podman ps` / `podman logs hive` remain useful once the unit is `active`, for output the process wrote after it came up.
+
 The process also writes `hive.log` under the configured logs directory (default `/data/logs`, from `governor.logging.dir`, falling back to `data.logs_dir`), and tees the same lines to stdout — so `docker compose logs` / `kubectl logs` show everything the file does. The older per-agent `AGENT_LOG_FILE` heartbeat file from v1 is not used.
 
 ## Config fails to load or save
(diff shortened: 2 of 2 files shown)


clubanderson:
Fact-checked every claim against merged v4 reality: all ten linked targets exist and are tracked (four Podman docs, backup-restore, lifecycle/update/teardown scripts); Quadlet unit names including hive-data-volume.service match what backup-restore.md already uses; TimeoutStartSec is indeed 300 for hive and 120 for the gateway with Restart=always on both, so the #4378 activating-not-failed guidance is accurate; HIVE_SRC_DIR="$CONF" preflight-host is now the valid recipe post-#4433 (layout awareness confirmed in the script); the drop-in merge behaviour and hive.container.d path match bin/hive-podman-update.sh, whose status subcommand does print pinned/running/rollback exactly as described; the auto-update entry matches podman-auto-update.md verbatim (off by default, bad image retried on every timer firing, digest pin wins silently with UPDATED=false exit 0); and the [SECURITY] HIVE_DASHBOARD_TOKEN refusal line is real (src/proxy/server.js:533) including the hub-hosted exception. The destructive volume-rm isolation with the stop-does-not-delete clarification is exactly the right shape for a reset section, and no forbidden cleanup phrasing appears anywhere. Docs-only, CI green.

/lgtm
/approve

PR #4450, opened 2026-08-21
docs(podman): execute health-aware auto-update in rootful mode (#4447)

### LINKED ISSUE #4411: 🧩 Podman: optional health-aware auto-update and rollback

### LINKED ISSUE #4447: 🧪 Podman: execute health-aware auto-update in rootful mode
## Parent

Part of #4188. Closes one row `src/docs/podman-auto-update.md` left open; it
must not close the parent.

## Why

#4411 shipped opt-in health-aware auto-update and proved it against a
bad-but-startable image. It records what it did **not** cover:

> ## Not executed here
>
> - **Rootful mode.** The mechanism is the system manager rather than the user
>   manager and nothing in podman's rollback path is uid-dependent, but it was
>   not run. #4377 exercised the unit itself in both modes.

That is a careful, honest reason — but it is reasoning, not a measurement, and
rootful is the mode that matters most here. It is the **enforcing** path in
#4324's matrix: the one where the egress gate is fully installed. Leaving the
enforcing mode's update story as "should be the same" is the wrong half to
leave unproven.

The tooling already supports it. `bin/hive-podman-update.sh` takes `--rootful`
and drives the system manager through `sudo` (`MODE_LABEL="rootful (system
manager)"`), so this is an execution task, not an implementation one.

## This is now known to be runnable

#4441's capability matrix (`src/docs/podman-host-execution-gap.md`) measured the
relay's own environment and recorded:

| rootful podman | **yes** | passwordless `sudo`; `Rootless=false` |

with systemd as PID 1 and a healthy user manager. Unlike #4413 — which that same
page shows is blocked by *topology*, not permission — nothing about this one
requires a host the session is not running on.

## 30-minute boundary

Execute and record. No changes to `hive-autoupdate.conf`, the unit, or
`hive-podman-update.sh` unless rootful exposes a defect — if it does, file that
separately rather than fixing it here. Do not touch the Docker/Watchtower path.

## Deliverable

Replace the "Rootful mode" bullet in `podman-auto-update.md`'s "Not executed
here" with measured results, in that document's existing observed-vs-expected
style.

## Acceptance criteria

- `autoupdate on --rootful` installs the drop-in, and the container is confirmed
  to carry the auto-update label — the label is what `podman auto-update` acts
  on, so its absence would make a green run meaningless.
- A **bad-but-startable** image is driven under the system manager: one that runs
  and never becomes healthy, not one that fails to pull. #4411 has the fixture.
- The **rollback trigger is confirmed to be the same one**. #4411's whole result
  is that podman reads the D-Bus **start job result** (`timeout`, not `done`)
  rather than `ActiveState`. Show that holds under the system manager — this is
  the single thing most likely to differ between managers, and the reason the
  row is worth executing rather than reasoning about.
- `systemctl is-failed` output is recorded during the bad update, so the
  monitoring warning in #4378/#4411 can be stated for rootful too rather than
  assumed to carry over.
- `hive-data` survives both directions, proven with a marker file.
- `autoupdate off --rootful` reverses everything, including the timer.
- If a step cannot be completed, record it plainly as not executed. Do not
  present the rootless result as if it covered rootful — that is the gap this
  issue exists to close.

## Related

- #4411 — opt-in auto-update, and the open row
- #4378 — the digest-pin path and the `failed`-state measurement
- #4441 — the capability matrix establishing rootful podman is available here
- #4324 — why rootful is the enforcing mode

Danathar:
Closes #4447. Part of #4188; does not close the parent.

## The row this closes

#4411 recorded rootful as **not executed**, with an honest reason:

> The mechanism is the system manager rather than the user manager and nothing in podman's rollback path is uid-dependent, but it was not run.

That is reasoning, not a measurement — and rootful is the wrong half to leave unproven, because it is the **enforcing** cell in [#4324's matrix](https://github.com/kubestellar/hive/issues/4324), the mode where the egress gate is fully installed. #4441's capability matrix had already established rootful podman is reachable here (passwordless `sudo`, `Rootless=false`), so unlike #4413 nothing blocked it.

Same fixture, same scaled timers as the rootless run — so the two are directly comparable — and driven entirely through the shipped tooling (`bin/hive-podman-update.sh ... --rootful`, units in `/etc/containers/systemd/`).

## Results

**Arming it.** The recreated container really carries the label, which is the check that makes the rest meaningful — `podman auto-update` acts on the label, so a green run without it would prove nothing:

```
PASS  installed /etc/containers/systemd/hive.container.d/20-autoupdate.conf
PASS  container label: registry
io.containers.autoupdate = 'registry'      system timer = enabled
```

**The bad update** — same bad-but-startable image (runs, never becomes healthy; not a pull failure):

```
UNIT          CONTAINER            IMAGE                              POLICY    UPDATED
hive.service  0028344b7801 (hive)  127.0.0.1:5000/hivefixture:stable  registry  rolled back
elapsed_s=37        (rootless was 36s)
```

**The rollback trigger is the same one.** This was the single thing most likely to differ between managers, and the reason the row was worth executing rather than reasoning about. From the **system** journal:

```
hive.service: start operation timed out. Terminating.
hive.service: Failed with result 'timeout'.
```

`Result=timeout` also appears in the live sample at t+33s, while `ActiveState` never reaches `failed`. podman rolled back on the start-job result under the system manager exactly as under the user manager.

**The monitoring blind spot carries over — measured, not assumed.** `systemctl is-failed` reported `active`, `activating`, `deactivating` and at no point `failed`; `NRestarts` stayed `0`. An alert keyed on `failed` does not fire on a rootful bad update either.

**`hive-data` survived** (marker intact, container back on the good image), and **`autoupdate off --rootful` reversed everything**, returning the system timer to `disabled / inactive` — exactly its pre-run state.

## No code change

Rootful exposed no defect, so there was nothing to fix and nothing to file separately. `hive-autoupdate.conf`, the unit and `hive-podman-update.sh` are untouched, as is the Docker/Watchtower path.

Beyond replacing the bullet, the summary table gains a **"Both root modes?"** row and the docs index says both were executed — so a reader who never reaches the measurement section still learns which modes are covered.

## Host hygiene

This ran as root on a live host, so I captured the machine's state **before** and verified it **after**: system quadlet dir empty, no root containers or volumes, no `hive*` units, `podman-auto-update.timer` back to `disabled / inactive`, temporary insecure-registry file removed.

Pre-existing root images — the operator's own bootc/kinoite builds, months old — were deliberately left alone; I checked their timestamps rather than clearing root storage wholesale. Only the fixture's own images, containers and its anonymous registry volume were removed, the volume identified by a creation timestamp inside the run window.

Documentation only.


clubanderson:
Reviewed from head bf0cbc15. Doc-only, closes exactly the row #4420 left as reasoning, and the measured bar is met.

- **Right target**: rootful was indeed the wrong half to leave unproven — it's the enforcing cell in the support matrix — and #4445 had just established nothing blocked it here (passwordless sudo, Rootless=false).
- **The one thing that could differ, measured**: the rollback trigger under the *system* manager is the same start-job `Result=timeout` (journal lines quoted, visible at t+33s in the sample), while `ActiveState` never reaches `failed` and `NRestarts` stays 0 — so both #4420 findings (rollback fires; `is-failed` alerting doesn't) carry to rootful as measurements, not assumptions.
- **Consistent with the shipped tooling**: `/etc/containers/systemd/hive.container.d/20-autoupdate.conf` matches the script's rootful QUADLET_DIR exactly (hive-podman-update.sh:137); label-on-recreated-container checked before trusting the run; 37s vs rootless 36s on the same scaled fixture keeps the two directly comparable; hive-data marker proven.
- **Host hygiene**: before/after state captured, `autoupdate off --rootful` returned the timer to its pre-run disabled/inactive, fixture-only cleanup with pre-existing root images deliberately left alone — the right discipline for running as root on a live host.
- Remaining not-executed bullets honestly retained (real broken published image, shipped 300s timers); README index updated to say both modes executed; no code change claimed or made; no forbidden phrases; cross-refs resolve; trial merge clean.

/lgtm
/approve

PR #4453, opened 2026-08-21
🐛 Stop false "Model removed" toasts from partial model discovery (#4438)

### LINKED ISSUE #4438: 🐛 bug: mysterious toasts about model removals
## What happened?

I am administering the hive for llm-d FMA. This morning when I opened the dashboard I was greeted with a lot of pink toasts saying "Model removed from litellm: xxx" for many values of xxx --- including the one (claude-sonnet-4.6) that this hive's agents are configured to use.

Image

The LiteLLM dashboard continues to say that the relevant key is allowed to use "all-team-models".

Image

When I try to interrogate that dashboard about what these models are, I get a long list and that list includes some ending with "/claude-sonnet-4-6". I have looked here before, and never seen plain "claude-sonnet-4-6".

Image

## Expected behavior

No mysterious and probably false toasts.

## Reproduction steps

I do not know

## Environment

- Branch/version: `v4` / commit 9f31439295c6fe8b70df70f8f16ea76afd57a063
- Deployment method: Hive Hub

## Logs or screenshots

See above

## Additional context

This might be related to #4400? But that Issue arose hours before these toasts appeared.

Danathar:
Fixes #4438.

## What was reported

An operator opened the FMA hive dashboard to a wall of pink `Model removed from litellm: xxx` toasts naming models the gateway still served — including `claude-sonnet-4.6`, the model that hive's own agents were configured to use. The LiteLLM key still held `all-team-models`, and the ids were still in the gateway's list. The reporter also noted the toasted name was one he had never seen in that list: the real ids there are provider-prefixed, `…/claude-sonnet-4-6`.

## Why it happened

The dashboard announces a removal whenever a fresh model list is shorter than the last one, so **any** incomplete answer from discovery reads as a mass deletion. #4426 taught the server to excuse a discovery that failed *outright* (`models_fallback`). Three ways of being only **partly** wrong were still passed off as authoritative:

- **A partial multi-endpoint sweep.** `fetchModelsFromEndpoints` skips an endpoint that errors and returns the survivors, so one gateway of several timing out or answering 403 looked exactly like every model behind it being deleted.
- **The `HIVE_*_MODELS` env list.** It is consulted only *because* discovery came up empty, yet was flagged as an operator-configured set rather than a stand-in for a failed probe — so on any hive that sets it, a single 403 swapped the live list for the env list and called the swap a census.
- **An empty live list** the server did not flag, which would retire every known model at once — the worst case of the same storm.

## The fix

All three now count as "we do not know" rather than "they are gone". What discovery *did* learn is still offered in the model dropdowns — a short list beats an empty one — it just no longer stands as a census.

Beyond the toasts: the same partial answer reached model **auto-heal**, which does not merely notify — it rewrites the agent's configured model and relaunches its session. An endpoint that briefly timed out could therefore move an agent off a model only that endpoint serves. `handleInferenceModels` now reports `partial`, and auto-heal sits out any sample it cannot trust. `partial` is deliberately kept distinct from `fallback`: those ids really were discovered, so they must not be labelled "common alias, unverified" in the dropdown.

Finally, the display half of the mystery. The toasts printed only the segment after the last `/`, rendering `vertex/claude-sonnet-4-6` as bare `claude-sonnet-4-6` — spelled exactly like the id an agent is configured with, and findable nowhere in the gateway's own list. That is what made these toasts *mysterious* rather than merely wrong. They now name the full id.

The control paths are unchanged and covered by tests: a complete sweep stays authoritative, an env list with no endpoint to probe is still the configured source, and a genuine removal still toasts and still heals.

## Testing

`src/pkg/dashboard/model_removal_toast_test.go` — 11 tests over both discovery paths and both frontend guards. The two central regression tests were confirmed to fail against the pre-fix logic and pass after it. The `pkg/dashboard` suite reports the same 25 pre-existing environment-related failures before and after this change — no new ones.


clubanderson:

**Verdict**: Correct fail-safe semantics on every axis I checked.

- **Partial ≠ absent** — the three unflagged "we don't actually know" shapes (partial multi-endpoint sweep, `HIVE_*_MODELS` standing in for a failed probe, unflagged empty list) all now read as non-authoritative; what *was* discovered still fills the dropdowns. The env list correctly stays authoritative in the one case it's the configured source (no endpoint registered → nothing probed). Empty-list guard keeps the last real baseline AND the dropdown rather than blanking the picker.
- **Real removals still land** — control tests pin that a complete sweep stays authoritative (`FullSweepIsAuthoritative`, `FullSweepIsNotPartial`), so genuine removals toast and auto-heal once discovery is healthy — no stuck-forever models.
- **No toast storms on flapping** — the diff baseline only advances on real (non-fallback, non-partial, non-empty) lists, and the pre-existing `_dismissedModelBanners` key set dedupes per model. Flapping discovery produces no diff at all.
- **Auto-heal is the sharpest edge and it's covered** — a partial sample no longer rewrites an agent's model and relaunches its session (`data.partial` ORed into the reconcile guard, pinned by test). `partial` is correctly kept distinct from `fallback` so genuinely-discovered ids aren't mislabelled as unverified aliases.
- **Wiring verified** — the SSE producer (server.go:805/813) feeds `ModelsFallback` from `queryInferenceModelsDetailed`, so the widened semantics reach the frontend; the bool-discarding `queryInferenceModels` wrapper's only caller is `handleBackends`, which fills dropdowns and never diffs — safe.
- **Display fix** — full provider-prefixed ids in toasts resolve the reporter's "never seen plain claude-sonnet-4-6" confusion; negative test pins the `split('/').pop()` regression.
- **Dashboard JS constraints** — statusSeq/render flow untouched; every referenced function (`reconcileModelsAfterDiscovery`, `showToast`, `TOAST_MODEL_CHANGE_MS`, `_dismissedModelBanners`) pre-exists; both inline script blocks parse clean under `node --check`.
- **Local run**: build/vet/gofmt clean, `go test ./pkg/dashboard/` ok, coverage **83.4% ≥ 78 floor**. Tests cover partial-discovery, full-failure (#4426 path), env-standin, genuine-removal, and empty-list paths. No forbidden podman lifecycle literals.

CI green.

/lgtm
/approve

PR #4457, opened 2026-08-21
ci: guard the release-line carry-forward — nine workflows are pinned to branch names (#4405)

### LINKED ISSUE #4405: 🌱 CI: guard the release-line carry-forward — nine workflows are pinned to branch names
## Parent

Part of #4188 (second wave, bullet 15). This issue must not close the parent.
The umbrella asks for a "successor-version/release-line carry-forward guard" so
that "Podman support is carried forward when successor version branches are
created; release synchronization cannot silently drop either runtime."

The scope turned out to be wider than Podman. Filing it against the umbrella
because that is where the requirement came from, but the fix is generic.

## Why

This already happened once. #4339: `podman-contract.yml` named `v2` alone,
mainline moved to `v4`, and the workflow stopped running entirely. Its own
header now records the lesson:

> a guard that never executes reports green forever, which is the one failure
> mode a guard cannot have.

That fix added `v4` to the list. It did not change the fact that the list is a
hand-maintained set of branch names, so the same failure is queued up for
whenever `v5` is cut.

## The current exposure

Nine workflows carry a hardcoded version-branch allow-list. Both YAML spellings
are in use, which is part of why this is easy to miss:

| Workflow | Branches |
|---|---|
| `v2-ci.yml` | `branches: [v2, v4]` |
| `v2-tests.yml` | `branches: [v2, v4]` |
| `scorecard.yml` | `branches: [ "main", "v4" ]` |
| `podman-rootful-lane.yml` | `- v2` / `- v4` |
| `podman-rootless-lane.yml` | `- v2` / `- v4` |
| `podman-arm64-lane.yml` | `- v2` / `- v4` |
| `podman-contract.yml` | `- v2` / `- v4` |
| `quadlet-gate.yml` | `- v2` / `- v4` |
| `suid-contract.yml` | `- v2` / `- v4` |

`docker.yml` is the outlier that does **not** have the problem — it triggers on
`'**'` with bot branches excluded, deliberately, so a PR's image always builds.

So on the day `v5` branches: the Docker image keeps building on every branch,
while both CI workflows, both security contract gates, and all five Podman
guards report nothing at all. Not red — absent. That is the asymmetry the
umbrella's "release synchronization cannot silently drop either runtime" clause
is pointing at, and it is currently true of the *Docker* side winning by
accident rather than the Podman side being protected.

## 30-minute boundary

The guard. Not a rewrite of nine trigger blocks, and not a decision about
whether `v2` should still be in those lists — that is a maintainer call about
release-line support, and this issue should surface it, not settle it.

## Deliverable

A check that fails when a workflow's branch allow-list has fallen out of sync
with the repository's live release lines.

Two shapes are worth considering; pick one and say why:

1. **Assert the set.** A single source of truth for supported release-line
   branch names, and a test that every workflow in the pinned list matches it.
   Cheap, and it turns cutting `v5` into one edit that CI enforces.
2. **Remove the pins.** Trigger on a branch pattern (`v*`) so new release lines
   are covered by construction. Fewer moving parts, but it silently widens what
   runs on branches nobody intended, and `scorecard.yml`'s `main` does not fit
   the pattern.

## Acceptance criteria

- Adding a plausible future branch name to the repository's release lines
  without updating the workflows fails the check. Demonstrate the failure.
- The check itself is not subject to the bug it guards against — it must run on
  any branch, or it is #4339 again one level up.
- Both YAML spellings of `branches:` are handled. A guard that only understands
  the block-sequence form misses `v2-ci.yml`, which is the highest-value entry
  in the table.
- `docker.yml`'s `'**'` trigger is recognised as deliberate and not flagged.
- Whether `v2` should remain in these allow-lists is raised for a maintainer
  decision rather than answered here.

Danathar:
Fixes #4405.

## What was exposed

Nine workflows pin their `push` / `pull_request` triggers to a hand-maintained list of version-branch names. That list has already gone stale once: `podman-contract.yml` named `v2` alone, mainline moved to `v4`, and the workflow stopped running entirely (#4339) — not red, **absent**, which is the one failure mode a guard cannot have. Adding `v4` fixed that instance and left the mechanism intact, so the same failure was queued up for the day `v5` is cut.

## Which shape, and why

Shape **1 — assert the set**, not shape 2 (trigger on `v*`):

- `scorecard.yml` triggers on `main`, which is not a release line and does not match `v*`. A pattern cannot express it, so a pattern would leave that workflow pinned anyway — the guard would still be needed, for a smaller set.
- A `v*` pattern silently widens what runs. Any branch someone happens to name `validate-thing` starts running the five Podman lanes. Coverage would be correct by construction, but nobody could see what it covers.

Asserting the set keeps each trigger block saying literally what it runs on, and turns cutting `v5` into one edit that CI enforces.

## What lands

| File | Role |
|---|---|
| `.github/release-lines.yml` | Single source of truth: the release lines, and how each workflow relates to them |
| `src/scripts/check-release-lines.sh` | Asserts every workflow's branch filters against the manifest |
| `src/scripts/test-release-lines-guard.sh` | Proves the checker still fails when it should |
| `.github/workflows/release-line-guard.yml` | Runs both, on **every** branch |
| `src/docs/release-line-guard.md` | What it checks, and what to edit when a release line is cut |

`podman-contract.yml`'s header gains one sentence pointing at the guard. **No existing workflow's triggers change.**

## Acceptance criteria

**Adding a plausible future branch name without updating the workflows fails the check — demonstrated.** Case 2 of the self-test does exactly this in CI, against the real workflow files, so it keeps being true rather than being a paste in a PR description:

```
$ sed 's/^release_lines:.*/release_lines: [v2, v4, v5]/' .github/release-lines.yml > /tmp/m.yml
$ src/scripts/check-release-lines.sh /tmp/m.yml .github/workflows
  FAIL: v2-ci.yml:5 branches: has [v2,v4], expected [v2,v4,v5] — missing: v5. A release
        line named here but absent from the workflow means that workflow does not run on
        it at all.
  ... same for v2-tests.yml, scorecard.yml, suid-contract.yml, quadlet-gate.yml and all
      four podman lanes (13 filters across 9 workflows)
RESULT: FAIL — workflow branch filters are out of sync with /tmp/m.yml.
$ echo $?
1
```

**The check is not subject to the bug it guards against.** `release-line-guard.yml` triggers on `branches: ['**']` and is itself listed under `unpinned` in the manifest, so the guard checks that fact about itself on every run. Its path filter (`.github/**`, the manifest, the two scripts) is safe where a branch filter would not be: this invariant can only be broken by editing a workflow, the manifest, or the checker. The self-test runs *before* the real check, because a checker that had quietly stopped detecting anything would otherwise pass for the wrong reason.

**Both YAML spellings are handled.** `branches: [v2, v4]` (v2-ci.yml, v2-tests.yml) and the block sequence used by the Podman lanes and the contract gates, including quoted entries and inline comments. Dedicated fixture cases cover each; a checker that read only block sequences would silently skip the two highest-value entries.

**`docker.yml`'s `'**'` is recognised as deliberate and not flagged** — it is declared `unpinned` with the reason. The reverse is checked too: if that wildcard is ever narrowed to a fixed list, the guard fails and asks for it to be reclassified as `pinned`.

**Whether `v2` should remain is raised, not answered.** It is a maintainer call about release-line support. What the guard surfaced on its first run is that the repo is not currently consistent about it: mainline is `v4`, eight of the nine pinned workflows still fire on `v2`, and the ninth — `scorecard.yml` — never has. That gap is recorded as an explicit `-v2` exclusion in the manifest rather than quietly corrected, because turning OpenSSF Scorecard on for a release line is a decision, not a side effect of adding a CI guard. Either resolution is one edit to `release_lines` plus the workflows, and the guard proves the edit is complete. The manifest ends with the question written out.

## Beyond the stated criteria (cheap, same parser)

- **Every** branch filter in a file, not the first — `push` and `pull_request` blocks drift apart independently (13 filters across the 9 workflows).
- **Classification is mandatory.** A workflow carrying a branch filter that is in neither list fails: the next pinned workflow somebody adds is the next #4339.
- **Stale manifest entries fail** — an entry naming a deleted workflow, or one that has lost its branch filter. A guard checking a file with nothing left to check is green for no reason.

## Verification

```
$ src/scripts/check-release-lines.sh
RESULT: PASS — every pinned workflow covers the declared release lines.

$ src/scripts/test-release-lines-guard.sh
RESULT: PASS — the release-line guard still catches a stale allow-list.   (29 assertions)

$ shellcheck -S warning src/scripts/check-release-lines.sh src/scripts/test-release-lines-guard.sh
$ python3 -c "import yaml,glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.yml')+['.github/release-lines.yml']]"
```

No `CHANGELOG.md` entry: CI-only, not user-visible, which that file's own guidance excludes.

Part of #4188 (second wave, bullet 15). Does not close the parent.

### DIFF
--- a/.github/release-lines.yml
+++ b/.github/release-lines.yml
@@ -1,0 +1,75 @@
+# Release-line branch allow-lists — the single source of truth for which
+# version branches CI is expected to cover.
+#
+# Nine workflows pin their `push` / `pull_request` triggers to an explicit list
+# of version-branch names. That list is hand-maintained, and it has already
+# failed once: podman-contract.yml named `v2` alone, mainline moved to `v4`, and
+# the workflow stopped running entirely (#4339). A guard that never executes
+# reports green forever, which is the one failure mode a guard cannot have.
+# Adding `v4` fixed that instance and left the mechanism intact, so the same
+# failure was queued up for the day `v5` is cut. #4405 asks for the guard.
+#
+# src/scripts/check-release-lines.sh asserts this file against
+# .github/workflows/*.yml, and .github/workflows/release-line-guard.yml runs it
+# on every branch. See src/docs/release-line-guard.md.
+#
+# CUTTING A NEW RELEASE LINE: add the branch name to `release_lines` below AND
+# to every workflow listed under `pinned`. The guard fails until both are done —
+# that is the whole point, so do not "fix" it by only editing one side.
+
+# The release lines CI is expected to cover.
+release_lines: [v2, v4]
+
+# Workflows whose branch filters are pinned to the release lines.
+#
+# The value adjusts `release_lines` for that workflow:
+#   `[]`      release lines and nothing else
+#   `[name]`  an EXTRA branch this workflow legitimately allows
+#   `[-name]` a release line this workflow deliberately does NOT cover
+# EVERY `branches:` list in the file must equal `release_lines`, plus the
+# extras, minus the exclusions — exactly, no more and no less. An exclusion has
+# to be written down, which is the point: it turns a silent gap into a line
+# somebody has to justify, and a new release line still has to be added or
+# explicitly excluded here before the guard goes green again.
+pinned:
+  podman-arm64-lane.yml: []
+  podman-contract.yml: []
+  podman-rootful-lane.yml: []
+  podman-rootless-lane.yml: []
+  quadlet-gate.yml: []
+  # scorecard.yml also runs on `main`, which is not a release line and does not
+  # fit a `v*` pattern — this is why the guard asserts a set rather than
+  # replacing the pins with a wildcard. It has also never run on `v2`; that is
+  # recorded as an exclusion rather than quietly corrected, because turning
+  # OpenSSF Scorecard on for a release line is a maintainer decision, not a
+  # side effect of adding a CI guard. See the note at the bottom of this file.
+  scorecard.yml: [main, -v2]
+  suid-contract.yml: []
+  v2-ci.yml: []
+  v2-tests.yml: []
+
+# Workflows that deliberately do NOT pin to the release lines. Each must keep a
+# wildcard branch filter; if one is ever narrowed to a fixed list, the guard
+# fails and asks for it to be reclassified as `pinned`.
+unpinned:
+  # Builds a PR's image on every branch on purpose, with bot branches excluded.
+  docker.yml: '**'
+  # The guard itself. It must run on any branch, or it is #4339 one level up.
+  release-line-guard.yml: '**'
+
+# MAINTAINER DECISION — deliberately NOT answered here (#4405).
+#
+# Should `v2` still be in these allow-lists? Mainline is `v4`, yet eight of the
+# nine pinned workflows still fire on `v2` pushes and PRs — and the ninth,
+# scorecard.yml, never has, which is the inconsistency this guard surfaced on
+# its first run. Whether `v2` is still a supported release line is a maintainer
+# call, so this file records the status quo rather than picking a side:
+#
+#   - if `v2` IS still supported, scorecard.yml's `-v2` exclusion should be
+#     dropped and `v2` added to that workflow;
+#   - if it is NOT, remove `v2` from `release_lines` above and from the eight
+#     workflows that name it (and drop the now-redundant `-v2`).
+#
+# Either way it is one edit to `release_lines` plus the workflows, and this
+# guard proves the edit is complete. Until a maintainer says otherwise, the
+# current behaviour stands unchanged.
--- a/.github/workflows/podman-contract.yml
+++ b/.github/workflows/podman-contract.yml
@@ -11,7 +11,9 @@ name: Podman Rootless Contract
 # guard that never executes reports green forever, which is the one failure
 # mode a guard cannot have. The branch list and the self-referencing path entry
 # below mirror .github/workflows/suid-contract.yml, the sibling static contract
-# check.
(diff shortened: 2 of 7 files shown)


clubanderson:
Reviewed and verified at head 9494349a:

- **Checker + self-test pass at head**: 19 filters validated, 29 self-test assertions green, both run locally.
- **Failure modes proven by live mutation**: (1) a synthetic new workflow with `branches: [v4]` and no manifest classification → FAIL with an actionable message (mandatory classification means the next #4339 is caught at introduction); (2) adding `v5` to `release_lines` → all 16 pinned filters fail, exit 1.
- **No false positives**: comments mentioning `branches:` and workflows without branch filters (schedule-only) are ignored; `docker.yml`'s deliberate wildcard and the guard's own `'**'` are classified `unpinned` and cross-checked (narrowing them would fail).
- **Explicit exemptions**: `scorecard.yml: [main, -v2]` records the intentional default-branch pin (post-#4394) and surfaces the v2 question as a written-down maintainer decision instead of silently changing behavior. `v2-ci.yml` is handled purely by filename — no confusion with the retired v2 line.
- **Wiring**: `release-line-guard.yml` runs on every branch, self-test before the real check, `permissions: contents: read`, SHA-pinned checkout, path filter covers workflows + manifest + both scripts. actionlint and `shellcheck -S warning` clean. No existing workflow triggers change.
- No forbidden phrases. CI green.

Nice shape choice — asserting the set keeps triggers legible while making a v5 cut a single CI-enforced edit.

/lgtm
/approve

PR #4458, opened 2026-08-21
ci(release-lines): assert workflow branch pins against the live release lines (#4405)

### LINKED ISSUE #4405: 🌱 CI: guard the release-line carry-forward — nine workflows are pinned to branch names

Danathar:
Closes #4405. Part of #4188 (second wave, bullet 15) — does **not** close the parent.

## The problem

Nine workflows pin their triggers to a hand-maintained list of version-branch names, and `docker.yml` hand-writes the same names a second time in its `LONG_LIVED` push policy. Nothing checked that those lists agreed with reality.

This already happened once. #4339: `podman-contract.yml` named `v2` alone, mainline moved to `v4`, and the workflow stopped running entirely. Its header records the lesson:

> a guard that never executes reports green forever, which is the one failure mode a guard cannot have.

That fix added `v4` to one list. It fixed the instance, not the class — so on the day `v5` is cut, both CI workflows, both security contract gates and all five Podman lanes report nothing at all. Not red. Absent.

## Which shape, and why

The issue offered two. This takes **shape 1, assert the set**, and the reasoning is recorded in `.github/release-lines.yaml` itself so the next person does not have to re-derive it. A `v*` glob:

* silently widens what runs on any branch that happens to start with `v`,
* does not fit `scorecard.yml`'s `main`,
* cannot express `docker.yml`'s push policy at all, and
* decisively — **can only ever add**. A glob has no way to notice that a release line has been *retired* and should stop being built.

Asserting the set costs one edit per release line and catches both directions.

## What lands

**`.github/release-lines.yaml`** — the source of truth. `release_lines` answers "what are the live release lines"; `workflows` declares each file that pins branch names; `branch_lists` covers hand-written lists that are not triggers.

**`src/pkg/releaselines`** — makes it load-bearing:

* every workflow whose `push` / `pull_request` / `pull_request_target` carries a `branches:` filter must be declared, so a workflow added later that pins branches and declares nothing is reported rather than being the next one to silently stop running;
* a `pinned: true` filter must name every release line and nothing else;
* `pinned: false` must actually be a glob, and must say why;
* the same assertion is applied to `docker.yml`'s `LONG_LIVED`;
* deviations are declared as `extra` / `omits` and are themselves checked, so a declaration that stops describing the file is reported instead of rotting.

**`.github/workflows/release-lines-gate.yml`** — runs it, on every branch, via `go test -v ./pkg/releaselines/...`.

`docker.yml`'s **trigger** is the outlier the issue names and is not flagged: it is declared `pinned: false` with its reason. Its **push policy** is a different matter and is checked — being out of sync there is worse than a skipped guard, because on the day `v5` is cut `docker.yml` would keep building `v5` and never publish `v5-latest`, so no hive could be assigned to the new line through the hub's branch switcher at all. The image would look healthy the whole time.

## Acceptance criteria

**Adding a plausible future branch name fails the check — demonstrated.** `TestCuttingANewReleaseLineFailsUntilWorkflowsCatchUp` declares `v5` against the repository's *real* workflows and asserts all ten files are named. Its output, verbatim from the test log:

```
ERROR   .github/workflows/v2-ci.yml (on.push.branches): release line "v5" is missing; this workflow does not run on it at all. Add it to the filter, or declare `omits: [v5]` in .github/release-lines.yaml with a note saying why it is deliberate
ERROR   .github/workflows/v2-ci.yml (on.pull_request.branches): release line "v5" is missing; ...
ERROR   .github/workflows/v2-tests.yml (on.pull_request.branches): ...
ERROR   .github/workflows/scorecard.yml (on.push.branches): ...
ERROR   .github/workflows/podman-rootful-lane.yml (on.push.branches): ...
ERROR   .github/workflows/podman-rootless-lane.yml (on.push.branches): ...
ERROR   .github/workflows/podman-arm64-lane.yml (on.push.branches): ...
ERROR   .github/workflows/podman-contract.yml (on.push.branches): ...
ERROR   .github/workflows/quadlet-gate.yml (on.push.branches): ...
ERROR   .github/workflows/suid-contract.yml (on.push.branches): ...
ERROR   .github/workflows/docker.yml (env LONG_LIVED): release line "v5" is missing from the list (v2 v4 mk dd). Add it, or declare `omits: [v5]` in .github/release-lines.yaml

10 file(s) have fallen out of sync with .github/release-lines.yaml.
```

The reverse is demonstrated too — `TestRetiringAReleaseLineNamesEveryWorkflowStillPinnedToIt` removes `v2` and every workflow that still names it is reported, including `scorecard.yml`'s now-stale `omits: [v2]`.

**The check is not subject to the bug it guards against.** `release-lines-gate.yml` triggers on `'**'` and carries no path filter, and `TestGuardIsReachableFromEveryBranch` asserts *both*, so the exemption cannot be quietly withdrawn later. A path filter would be the same mistake in the other axis: cutting a release line edits `release-lines.yaml`, but so does renaming the env var `docker.yml` keeps its push policy in.

**Both YAML spellings are handled.** The workflow YAML is parsed, not pattern-matched, so `branches: [v2, v4]` (the flow sequence `v2-ci.yml` uses — the highest-value entry in the table) and the block sequence are the same node tree. `TestBothYAMLSpellingsAreHandled` runs flow, block and quoted-flow through both the passing and the failing path and asserts they behave identically.

**`docker.yml`'s `'**'` is recognised as deliberate and not flagged.** Asserted directly in the `v5` test: no finding may attach to a `docker.yml` trigger even while its `LONG_LIVED` list is reported.

## Maintainer decisions this surfaces rather than settles

Two things are recorded as *observed*, not endorsed, and are for maintainers to confirm or correct:

1. **Is `v2` still a supported release line?** #4405 was explicit that this is not its call, and this PR does not make it. `v2` is declared `status: undecided`, which prints a `::notice::` on every run and never fails the build — a question settled by silence is settled the wrong way. If the answer is "no longer supported", deleting those three lines is the whole change on this side, and the check then names each of the ten files that still carries `v2`.
2. **`scorecard.yml` names `main` and omits `v2`.** CONTRIBUTING.md says `main` "is not the active target for changes". Both predate this check, and calling either deliberate would be a guess, so they are declared to describe today's state.

## Testing

`go test ./pkg/releaselines/...` — 25 tests, 96.9% statement coverage (per-package floor is 90). `gofmt` and `go vet` clean. The demonstrations above run against the repository's real workflow files, not fixtures, so they cannot pass vacuously.

No runtime code is touched and nothing user-facing changes, so there is no CHANGELOG entry.

### DIFF
--- a/.github/release-lines.yaml
+++ b/.github/release-lines.yaml
@@ -1,0 +1,142 @@
+# Release lines, and every CI file that hand-writes their branch names.
+#
+# WHY THIS FILE EXISTS (#4405, under #4188). Nine workflows pin their triggers
+# to a hand-maintained list of version-branch names, and docker.yml hand-writes
+# the same names again in its push policy. Nothing checked that those lists
+# agreed with reality, so when mainline moved from `v2` to `v4` the Podman
+# contract gate kept naming `v2` alone and simply stopped running — #4339. Its
+# header now records the lesson:
+#
+#   a guard that never executes reports green forever, which is the one failure
+#   mode a guard cannot have.
+#
+# That fix added `v4` to one list. It did not change the fact that the lists are
+# hand-maintained, so the same failure was queued up for the day `v5` is cut —
+# on which the Docker image would keep building on every branch while both CI
+# workflows, both security contract gates and all five Podman lanes reported
+# nothing at all. Not red. Absent.
+#
+# THIS FILE IS THE SOURCE OF TRUTH. `release_lines` below is the answer to "what
+# are the live release lines"; everything else is checked against it by
+# src/pkg/releaselines, which .github/workflows/release-lines-gate.yml runs on
+# every branch as `go test ./pkg/releaselines/...`. Cutting a release line is
+# therefore ONE edit here, and CI then names every file that has not caught up.
+#
+# WHY ASSERT THE SET RATHER THAN DROP THE PINS. #4405 offered a second shape:
+# trigger on `v*` so new lines are covered by construction. It was not chosen.
+# A glob silently widens what runs on any branch that happens to start with
+# "v", it does not fit scorecard.yml's `main`, it cannot express docker.yml's
+# push policy at all — and, decisively, it can only ever add: a glob has no way
+# to notice that a line has been RETIRED and should stop being built. Asserting
+# the set costs one edit per release line and catches both directions.
+
+# ── The release lines ────────────────────────────────────────────────────────
+#
+# `status` is documentation with one behavioural effect: an `undecided` line
+# makes the check print a notice on every run, so an open question stays
+# visible instead of being settled by silence. It never fails the build.
+release_lines:
+  - branch: v4
+    status: current
+    note: >
+      Mainline. This is the branch `stable`, `candidate` and `edge` are cut
+      from, and the one the hub's branch switcher assigns hives to by default.
+
+  - branch: v2
+    status: undecided
+    note: >
+      MAINTAINER DECISION NEEDED. Whether `v2` is still a supported release
+      line is not something #4405 was scoped to answer, and this file does not
+      answer it: `v2` is listed because every pinned workflow currently names
+      it, so listing it is what records the status quo rather than quietly
+      changing it. If the answer is "no longer supported", deleting these three
+      lines is the whole change on this side — the check will then name each
+      workflow that still pins `v2` and each list that still carries it.
+
+# ── Workflows that pin their triggers to branch names ────────────────────────
+#
+# Every workflow under .github/workflows whose `on.push` / `on.pull_request` /
+# `on.pull_request_target` carries a `branches:` filter must appear here. Both
+# YAML spellings are in use and both are handled — `branches: [v2, v4]` (flow
+# sequence, v2-ci.yml and v2-tests.yml) and the block sequence everything else
+# uses — because the check parses the YAML rather than pattern-matching text.
+#
+#   pinned: true    the filter is a list of literal branch names. Every release
+#                   line above must appear in it, unless named in `omits`; every
+#                   literal that is NOT a release line must appear in `extra`.
+#   pinned: false   the filter is a glob. Deliberate, and `reason` says why.
+#
+# A workflow that has no `branches:` filter at all runs on every branch and
+# needs no entry — it cannot fall out of sync.
+workflows:
+  - file: v2-ci.yml
+    pinned: true
+
+  - file: v2-tests.yml
+    pinned: true
+
+  - file: scorecard.yml
+    pinned: true
+    extra: [main]
+    omits: [v2]
+    note: >
+      Recorded as observed, not endorsed. This workflow names `main` — which is
+      not a release line in this repository — and omits `v2`. Both predate this
+      check, and calling either one deliberate would be a guess, so they are
+      declared here to describe today's state and are listed in #4405 for
+      maintainers to confirm or correct.
+
+  - file: podman-rootful-lane.yml
+    pinned: true
+
+  - file: podman-rootless-lane.yml
+    pinned: true
+
+  - file: podman-arm64-lane.yml
+    pinned: true
+
+  - file: podman-contract.yml
+    pinned: true
+
+  - file: quadlet-gate.yml
+    pinned: true
+
+  - file: suid-contract.yml
+    pinned: true
+
+  - file: docker.yml
+    pinned: false
+    reason: >
+      Triggers on '**' with bot branches excluded, on purpose: a PR's image
+      must BUILD on any branch, because that build is itself the CI gate
+      proving the tree compiles. Widening is the safe direction here, so this
+      trigger cannot be dropped by a release line being cut. Its push policy is
+      a different matter — see `branch_lists` below.
+
+  - file: release-lines-gate.yml
+    pinned: false
+    reason: >
+      Triggers on '**' because this is the check itself. #4339 was a guard
+      pinned to a branch name that went away; a guard against that failure
+      which was pinned the same way would be #4339 one level up, so this one is
+      deliberately reachable from every branch.
+
+# ── Other hand-written release-line lists ────────────────────────────────────
+#
+# Not every carry-forward hazard is a trigger. #4405's table calls docker.yml
+# "the outlier that does not have the problem", and that is true of its
+# TRIGGER — but the same nine branch names are hand-written a second time in
+# its push policy, where being out of sync is worse than a skipped guard: on
+# the day `v5` is cut, docker.yml would keep building `v5` and never publish
+# `v5-latest`, so no hive could be assigned to the new line through the hub's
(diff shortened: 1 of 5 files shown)


Danathar:
Closing as a duplicate. #4457 merged into `v4` as 94a4b36f while this was being written, and it already implements #4405: `.github/release-lines.yml` as the source of truth, `src/scripts/check-release-lines.sh` reading both YAML spellings, a self-test proving the checker still fails on a stale allow-list, `release-line-guard.yml` running on every branch, and the `v2` support question recorded as a maintainer decision rather than answered. Two guards for one invariant is worse than one — they collide here (`release-lines.yml` vs `release-lines.yaml`, each flagging the other's workflow as undeclared), which is why both checks are red on this PR.

One observation worth keeping, in case a maintainer wants it as a follow-up: the merged guard classifies `docker.yml` as `unpinned` on the strength of its `'**'` trigger, which is correct — but the same release-line branch names are hand-written a second time in that file, in the `gate` job's `LONG_LIVED: "v2 v4 mk dd"` push policy, and nothing checks that one. Being out of sync there is quieter than a skipped guard rather than louder: on the day `v5` is cut, `docker.yml` would keep building `v5` and never publish `v5-latest`, so no hive could be assigned to the new line through the hub's branch switcher, and the image would look healthy the whole time. Extending `check-release-lines.sh` to assert that list too would be a small addition to the merged design, not a competing one.

clubanderson:
Reviewed at head a224919d, together with #4457 (both implement #4405 shape 1, authored in parallel).

**Verdict: this implementation supersedes #4457's shell-based guard.** It parses the workflow YAML rather than pattern-matching lines, additionally covers `docker.yml`'s `LONG_LIVED` push policy (where drift means `v5-latest` silently never publishes — the worst variant of the failure class), checks `pull_request_target`, reports the retirement direction as well as the cut direction, and mandates classification so tomorrow's undeclared pinned workflow is caught (verified: 25 tests, 96.9% coverage, `TestCuttingANewReleaseLineFailsUntilWorkflowsCatchUp` and the retirement test both run against the real workflow files; gofmt/go vet/actionlint clean; `gopkg.in/yaml.v3` already a direct dependency).

**The three red checks are the mutual conflict, not defects**: the merge ref contains #4457's just-merged `release-line-guard.yml`, which this gate correctly flags as an undeclared branch-filtered workflow — the guard failing on a competing guard is the guard working. Two manifests (`.yml` + `.yaml`) and two guards cannot coexist, so a supersession commit (merge of v4 + removal of #4457's manifest/scripts/workflow/doc, header and README repointing, DCO-signed) is prepared and lands as an immediate follow-up PR — `maintainerCanModify` is off on this fork branch, so it could not be pushed here directly.

Also verified: `scorecard.yml`'s `main`+`omits: [v2]` recorded as observed-not-endorsed, `v2` `status: undecided` surfaces a `::notice::` without failing, gate has no path filter with `TestGuardIsReachableFromEveryBranch` enforcing both properties, `permissions: contents: read`, SHA-pinned actions. No forbidden phrases.

Admin-merging over the two guard checks for the reason above; v4 goes green again with the follow-up.

/lgtm
/approve

clubanderson:
Race noted — my review above was written before your close landed and its merge plan is moot; the close as duplicate stands and is the right call given #4457 is already on v4.

One piece of this PR is worth salvaging rather than losing with the close: #4457's guard only asserts `branches:` trigger filters, so `docker.yml`'s `LONG_LIVED` push-policy list (the second hand-written copy of the release-line names) remains unguarded — on a v5 cut, images would build but `v5-latest` would silently never publish. Filing a follow-up issue referencing this PR's analysis; the YAML-parsing approach here is a good starting point if anyone extends the guard.

PR #4466, opened 2026-08-21
ci(release-lines): assert docker.yml's LONG_LIVED push policy against the release lines (#4462)

### LINKED ISSUE #4462: ci: release-line guard does not cover docker.yml's LONG_LIVED push policy
Follow-up to #4405 / #4457, salvaged from #4458 (closed as duplicate).

The merged guard (#4457, `src/scripts/check-release-lines.sh` + `.github/release-lines.yml`) asserts every workflow's `branches:` trigger filters against the declared release lines. But the release-line branch names are hand-written a **second** time in `.github/workflows/docker.yml`'s `LONG_LIVED` env (the gate job's push policy: only long-lived branches push to GHCR). That list is not a trigger filter, so the guard does not see it.

Failure mode on the day v5 is cut: `docker.yml` triggers on `'**'` so the image keeps **building** on v5, but the push policy never publishes `v5-latest` — so no hive could be assigned to the new line through the hub's branch switcher, while the image looks healthy the whole time. Arguably worse than the skipped-workflow class the guard already catches, because it is green rather than absent.

Ask: extend the guard (manifest + checker + self-test) to assert `LONG_LIVED` against `release_lines`, with `mk`/`dd` (standing experimental branches) as declared extras. #4458 contains a worked YAML-parsing implementation of exactly this check for reference.

Part of the #4188 second-wave hardening context.

Danathar:
Closes #4462. Follow-up to #4405 / #4457; part of the #4188 second-wave hardening context — does **not** close the parent.

## The gap

The merged guard asserts every workflow's `branches:` trigger against `.github/release-lines.yml`. The release-line names are hand-written a **second** time outside any trigger: `docker.yml`'s gate job carries them in a `LONG_LIVED` env var, its GHCR push policy. That list is not a branch filter, so the guard never saw it.

It is worse than the skipped-workflow class the guard already catches, because it fails **green rather than absent**. On the day `v5` is cut, `docker.yml` still triggers on `'**'` so the image keeps *building* on v5, while the push policy never publishes `v5-latest` — no hive could be assigned to the new line through the hub's branch switcher, and the image looks healthy the whole time.

## What lands

**`.github/release-lines.yml`** gains an `env_lists` section, keyed `<workflow.yml> <ENV_NAME>` (a colon in the key would collide with YAML's own), taking the same `[extra]` / `[-excluded]` grammar as `pinned`:

```yaml
env_lists:
  docker.yml LONG_LIVED: [mk, dd]
```

`mk` and `dd` are declared extras, as the issue asks: standing experimental branches, not release lines, but branches a hive can be assigned to — so they legitimately publish a `<branch>-latest` tag.

**`src/scripts/check-release-lines.sh`** factors the expected-set comparison out of the branch-filter loop into `assert_release_line_set` and reuses it for these lists, so both are held to the same equality, in both directions: cutting a line names every list that omits it, retiring one names every list that still carries it. The value is read in each spelling a workflow env can take (`"v2 v4"`, unquoted, `[v2, v4]`), with trailing comments dropped outside quotes.

Two rot modes a branch filter does not have are also checked, because the entry would otherwise look checked while asserting nothing:

- the variable must appear **exactly once** — zero means it was renamed or removed, more than one means the guard cannot tell which occurrence is the policy. Both fail rather than guessing;
- prose that merely names it is not a definition. `docker.yml`'s own comment says the set is "defined ONCE below (LONG_LIVED)" and the gate step reads `$LONG_LIVED` in shell; neither is counted.

## Demonstration

Against the repository's real workflows, adding `v5` to `release_lines` and changing nothing else — `docker.yml` now appears in the failure, by its push policy and not by its trigger:

```
FAIL: v2-ci.yml:5 branches: has [v2,v4], expected [v2,v4,v5] — missing: v5. A release line
      named here but absent from the workflow means that workflow does not run on it at all.
...
FAIL: docker.yml:56 env LONG_LIVED: has [dd,mk,v2,v4], expected [dd,mk,v2,v4,v5] — missing:
      v5. This list is not a trigger, so a release line missing from it fails GREEN: the
      image still builds on that branch and simply never publishes its <branch>-latest tag,
      leaving no image for a hive to be assigned to.

RESULT: FAIL — hand-written branch lists are out of sync with .github/release-lines.yml.
Cutting a release line takes three edits: the manifest, every workflow listed under
'pinned', and every list under 'env_lists'.
```

**`docker.yml`'s `'**'` trigger is still recognised as deliberate and is not flagged.** The self-test asserts that as a distinct fact from the push-policy finding (`FAIL: docker.yml:<n> branches` must not appear while `FAIL: docker.yml:<n> env LONG_LIVED` must), so the two halves of that file cannot be confused for one another later.

## Self-test

`src/scripts/test-release-lines-guard.sh` grows six cases: the three env spellings parse; a stale list is caught and names the missing branch; an undeclared surplus branch is caught; a renamed variable is caught; a variable defined twice is refused rather than guessed at; a comment and a shell reference are not counted as definitions; and an `env_lists` entry for a deleted workflow is caught.

Case 1 additionally asserts the **real** run reaches the push policy at all — so deleting the manifest entry is a failing test rather than a quieter guard, which is the one thing a manifest-driven check cannot otherwise notice about itself.

Verified the tests are load-bearing by mutation: disabling the `env_lists` parse in the checker turns them red, rather than the suite passing for the wrong reason.

## Testing

```
src/scripts/check-release-lines.sh        # PASS — 20 lists asserted, including the new one
src/scripts/test-release-lines-guard.sh   # PASS — 47 assertions
shellcheck src/scripts/*.sh               # no new findings (two pre-existing SC2001 style notes)
```

Both run in CI via `release-line-guard.yml`, whose existing `.github/workflows/**` path filter already covers `docker.yml` — no trigger change was needed.

No runtime code is touched and nothing user-facing changes, so there is no CHANGELOG entry.

## Not in scope

Whether `v2` is still a supported release line remains the maintainer decision #4405 deliberately left open. This PR does not answer it; it only adds `docker.yml`'s `LONG_LIVED` to the list of places the note at the bottom of the manifest says a resolution would have to edit.

### DIFF
--- a/.github/release-lines.yml
+++ b/.github/release-lines.yml
@@ -9,13 +9,23 @@
 # Adding `v4` fixed that instance and left the mechanism intact, so the same
 # failure was queued up for the day `v5` is cut. #4405 asks for the guard.
 #
+# The names are also hand-written a SECOND time, outside any trigger: the gate
+# job in docker.yml carries them in a `LONG_LIVED` env var as its GHCR push
+# policy (#4462). That one is not a branch filter, so the trigger check cannot
+# see it — and its failure mode is worse than a skipped workflow. On the day
+# `v5` is cut, docker.yml still triggers on `'**'` so the image keeps BUILDING
+# on v5, while the push policy never publishes `v5-latest`; no hive could be
+# assigned to the new line through the hub's branch switcher, and the image
+# looks healthy the whole time. Green, not absent. `env_lists` below covers it.
+#
 # src/scripts/check-release-lines.sh asserts this file against
 # .github/workflows/*.yml, and .github/workflows/release-line-guard.yml runs it
 # on every branch. See src/docs/release-line-guard.md.
 #
 # CUTTING A NEW RELEASE LINE: add the branch name to `release_lines` below AND
-# to every workflow listed under `pinned`. The guard fails until both are done —
-# that is the whole point, so do not "fix" it by only editing one side.
+# to every workflow listed under `pinned` AND to every list under `env_lists`.
+# The guard fails until all three are done — that is the whole point, so do not
+# "fix" it by only editing one side.
 
 # The release lines CI is expected to cover.
 release_lines: [v2, v4]
@@ -57,6 +67,21 @@ unpinned:
   # The guard itself. It must run on any branch, or it is #4339 one level up.
   release-line-guard.yml: '**'
 
+# Hand-written release-line lists that are NOT trigger filters, and so are
+# invisible to the branch-filter check above. Each key is `<workflow> <ENV>`;
+# the value uses the same grammar as `pinned` — extras, and `-name` for a
+# release line the list deliberately omits.
+#
+# The variable must appear exactly once in the workflow. Zero means the list was
+# renamed or removed and the guard is asserting nothing; more than one means the
+# guard cannot tell which occurrence is the policy. Both fail rather than
+# guessing.
+env_lists:
+  # docker.yml's gate job: only long-lived branches push to GHCR. `mk` and `dd`
+  # are standing experimental branches — not release lines, but branches a hive
+  # can be assigned to, so they legitimately publish a `<branch>-latest` tag.
+  docker.yml LONG_LIVED: [mk, dd]
+
 # MAINTAINER DECISION — deliberately NOT answered here (#4405).
 #
 # Should `v2` still be in these allow-lists? Mainline is `v4`, yet eight of the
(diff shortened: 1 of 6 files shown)


clubanderson:
Reviewed and verified at head 40b589d1 — this closes exactly the gap #4462 documented (salvaged from #4458), and it extends the merged #4457 mechanism in place rather than competing with it: same manifest (`.github/release-lines.yml`), same checker, same guard workflow. No dual-manifest conflict.

- **Checker + self-test pass at head**: `docker.yml:56 env LONG_LIVED: dd,mk,v2,v4` asserted against `release_lines` + declared extras `[mk, dd]`; self-test now 44 assertions including a case-1 guard that the real run actually asserts the push policy (so deleting the `env_lists` entry is a failing test, not a quieter guard).
- **Mutations verified locally**: (1) dropping `v4` from `LONG_LIVED` → FAIL naming `missing: v4` with the fails-GREEN consequence spelled out; (2) adding `v5` to `release_lines` → the env list is flagged alongside the nine trigger filters; (3) renaming the var to `PUBLISH_SET` → "no 'LONG_LIVED:' assignment found" (the rot mode unique to env lists); the double-definition and comment/shell-reference cases are covered by fixtures 13–14 and I confirmed the real docker.yml prose mention and `$LONG_LIVED` shell read are not counted as definitions.
- **Grammar reuse is clean**: the `assert_release_line_set` refactor is a pure extraction — trigger-filter behavior identical (all pre-existing self-test cases still pass), extras/exclusions grammar shared.
- shellcheck clean on both scripts; actionlint: the six SC2046 warnings in docker.yml pre-exist on base (PR adds only comments there). No forbidden phrases. CI green.

The exactly-once occurrence rule (zero and duplicate both fail rather than guessing) is the right call — that's the failure mode a renamed env var would otherwise hide.

/lgtm
/approve

PR #4467, opened 2026-08-21
docs: state the runtime scope of the deployment helper scripts (#4459)

### LINKED ISSUE #4188: ✨ feature: support Podman as a first-class standalone runtime alongside Docker

### LINKED ISSUE #4459: 📝 deployment-scripts.md still says Docker and Kubernetes are the supported paths, and never says its four helpers are Docker-only
## Parent

Part of #4188 (Phase 2). It must not close the parent — it makes the current
runtime scope of these scripts **legible**, and deliberately does not extend any
of them to Podman. See "What this is not" below.

## Why: two docs now contradict each other

#4412 made Podman a parallel supported runtime, and `README.md` now opens with:

> Two supported standalone runtimes. **Docker Compose is the default** [...]
> **Podman** is a parallel supported choice, not an experiment and not a
> recommendation over Docker.

`src/docs/deployment-scripts.md` still opens with:

> The main supported paths remain **Docker Compose and Kubernetes**.

Both are in `v4` right now. The second was true when it was written and is not
any more.

## Why it matters beyond tidiness

That page documents four helper scripts, and **every one of them is
Docker-only**, with nothing on the page saying so:

| Script | `podman` | `docker` |
| --- | --- | --- |
| `bin/hive-setup.sh` | **0** | 23 |
| `src/deploy/bootstrap-lxc.sh` | **0** | 15 |
| `src/deploy/blue-green-deploy.sh` | **0** | 29 |
| `src/deploy/create-lxc.sh` | **0** | 3 |

`bin/hive-setup.sh` does not even reference `HIVE_DEPLOY_RUNTIME` — the selector
#4205 added. It has no concept that a runtime choice exists, and its own header
still describes a "Hive **v2** instance (Docker-based)".

So an operator who followed the README's Podman quick start, then reached this
page for a safer upgrade, would run `blue-green-deploy.sh` against a Quadlet
deployment. It builds an image, starts `hive-next` "on the existing Docker
network and volumes", and renames containers — none of which describes a Quadlet
install. The page gives them no reason to expect that.

## 30-minute boundary

Scoping and labelling only.

- **Do not** add Podman support to any of the four scripts.
- **Do not** delete or deprecate them — they are the supported path for the
  Docker installs they were written for.
- **Do not** rewrite the Podman docs; link to them.

## Deliverable

`src/docs/deployment-scripts.md`, corrected: an accurate opening statement of
supported runtimes, and each script's runtime scope stated where it is
described.

## Acceptance criteria

- The opening line agrees with `README.md` about how many standalone runtimes
  are supported. If Podman belongs in the "main supported paths" sentence, say
  so; if the sentence means something narrower, make that meaning explicit.
- Each of the four scripts is stated to be **Docker-only**, at the point where
  it is described — not once in a preamble a reader may skip.
- For an operator on Podman, each entry says where to go instead, and the
  pointers must be the ones that exist today:
  - upgrades → `podman-quadlet-update-rollback.md` (digest pin and rollback)
  - install → `README.md` Quick Start (Podman) → `podman-standalone-quadlet.md`
  - teardown → `bin/hive-podman-teardown.sh`, `podman-ownership-cleanup.md`
- States plainly that **no Podman equivalent of blue-green exists**, rather than
  implying one is coming. `Notify=healthy` plus digest rollback is the Quadlet
  answer to the same problem, and the two are not the same technique.
- `src/deploy/README.md` (5 Podman mentions) is checked for the same
  contradiction and corrected if it has it.

## What this is not

This does **not** close #4188's Phase 2. That phase names
`bin/hive-setup.sh`, `src/deploy/bootstrap-lxc.sh`, and
`src/deploy/blue-green-deploy.sh` as scripts to *extend or replace* for Podman.
Whether that is in scope — these are v2-era LXC and Compose helpers, so Podman
support means Podman-in-LXC and blue-green-under-Quadlet, arguably separate
features — is a maintainer decision that this issue deliberately leaves open.

What it removes is the case where an operator cannot tell, which is true today
regardless of how that decision goes.

## Related

- #4412 — made Podman parallel in the README, creating the contradiction
- #4205 — `HIVE_DEPLOY_RUNTIME`, which `hive-setup.sh` predates
- #4378 / #4411 — the Quadlet upgrade path blue-green has no equivalent of

Danathar:
Fixes #4459.

## The contradiction

`README.md` (since #4412):

> Two supported standalone runtimes. **Docker Compose is the default** […] **Podman** is a parallel supported choice, not an experiment and not a recommendation over Docker.

`src/docs/deployment-scripts.md`, in the same branch:

> The main supported paths remain **Docker Compose and Kubernetes**.

## Why it matters beyond tidiness

Every helper that page documents is Docker-only, and nothing on the page said so. Re-counted against `v4` at `5e408d12`:

| Script | `podman` | `docker` | reads `HIVE_DEPLOY_RUNTIME` |
| --- | --- | --- | --- |
| `bin/hive-setup.sh` | **0** | 23 | no |
| `src/deploy/bootstrap-lxc.sh` | **0** | 15 | no |
| `src/deploy/blue-green-deploy.sh` | **0** | 29 | no |
| `src/deploy/create-lxc.sh` | **0** | 3 | no |

An operator who followed the README's Podman quick start and then came here for a safer upgrade would run `blue-green-deploy.sh` against a Quadlet install: it builds an image, starts `hive-next` "on the existing Docker network and volumes", and renames containers — none of which describes what is on their host, and the page gave them no reason to expect that.

## Scope

Labelling only, per the 30-minute boundary. No script gains Podman support, none is deprecated — for the Docker installs they were written for they remain the supported path — and the Podman docs are linked, not rewritten.

## Acceptance criteria

**Opening line agrees with `README.md`.** It now says two supported standalone runtimes plus Kubernetes, and states outright the narrower thing the old sentence was reaching for: these particular helpers target Docker only, and none of them reads `HIVE_DEPLOY_RUNTIME` (#4205), so running one against a Quadlet install does not fail cleanly with "not supported" — it looks for Docker containers, networks, and volumes that are not there.

**Each script is stated Docker-only at the point where it is described** — three `**Docker-only.**` statements, one per section, plus a scope table at the top for readers who skim. The preamble is belt, not braces.

**Pointers, and only ones that exist today:**

| Need | Pointer | Where |
| --- | --- | --- |
| install | [README Quick Start (Podman)](https://github.com/kubestellar/hive/blob/v4/README.md#quick-start-podman) → `podman-standalone-quadlet.md` | All-in-one LXC setup |
| teardown | `bin/hive-podman-teardown.sh`, `podman-ownership-cleanup.md` | All-in-one LXC setup |
| upgrades | `podman-quadlet-update-rollback.md` | Blue-green Compose deploy |
| LXC host viability | `podman-preflight-ids.md` | Proxmox LXC helpers |

All four relative link targets were checked to resolve on disk, along with every other link in the touched files.

**`bin/hive-setup.sh` gains a section.** It is one of the four, but the page never mentioned it — it existed only as a one-line row in `bin/README.md`, which is why its scope was the easiest of all to miss. That row is now labelled too, and the page documents it properly. Its own header still describes "a Hive **v2** instance (Docker-based)", which the entry quotes rather than papers over.

**No Podman equivalent of blue-green exists — stated plainly, not as "not yet".** The Quadlet stack answers the same problem by a different technique, so a port of this script is not what a Podman operator is waiting for: `Notify=healthy` already means `systemctl start` returns only once Hive answered its health check, which is the prove-healthy-then-cut-over step the script builds by hand; and what blue-green additionally buys — a way back from a bad image — is the digest pin and rollback (#4378, #4411). The page says how they trade differently (a second image and container for the length of the cutover, versus one `TimeoutStartSec` of downtime when an update turns out bad) and that neither substitutes for the other.

**`src/deploy/README.md` checked and corrected.** It had the same contradiction in a smaller form: the three scripts were filed as "operational scripts for non-Kubernetes deployments" — which now reads as including the Quadlet stack listed two lines below them. Corrected to Docker Compose, with the reason. Its other four Podman mentions (the `quadlet/` entry, the volume-persistence probe, and the two contract-test references) were checked and are accurate.

`src/docs/README.md`'s one-line index entry is updated to match, since it named only two of the helpers and no runtime scope.

## Not in this PR

This does not close #4188 Phase 2. That phase names three of these scripts as candidates to *extend or replace* for Podman, which for v2-era LXC and Compose helpers means Podman-in-LXC and blue-green-under-Quadlet — arguably separate features, and a maintainer decision the issue deliberately leaves open. What this removes is the case where an operator cannot tell, which is true today whichever way that decision goes.

Docs only — no code, no script behaviour, no CI changes.


clubanderson:
Reviewed from head 83977f36. Labelling-only, and every label fact-checked rather than taken on prose.

- **The counts are exact**: re-ran the audit against the branch — `hive-setup.sh` podman=0/docker=23, `bootstrap-lxc.sh` 0/15, `blue-green-deploy.sh` 0/29, `create-lxc.sh` 0/3, zero `HIVE_DEPLOY_RUNTIME` reads in any — matching the PR table exactly. The does-not-fail-cleanly warning is the important operator-facing truth: these scripts look for Docker resources a Quadlet host doesn't have.
- **The blue-green analysis is the best part**: correctly identifies that `Notify=healthy` already IS the prove-healthy-then-cut-over step the script hand-builds, and that the way-back is the digest pin/rollback — with the trade-off stated in measured terms (second image+container for the cutover length vs one TimeoutStartSec on a bad update). "No Podman equivalent, and not 'not yet'" is the right framing; consistent with everything #4378/#4407/#4411/#4420 measured.
- **Pointers all resolve**: verified every relative target on disk (both quick-start anchors in README, ADR-0017 filename, teardown script, ownership/preflight/update-rollback/operator-guide docs). The Proxmox section's honest note that rootless-podman-in-LXC has subordinate-ID/storage needs these scripts don't set up, pointing at the preflight as the arbiter, is exactly right per #4307/#4433.
- Fixes the real #4412 contradiction in the opening line; `hive-setup.sh` finally documented on the page instead of existing only as a bin/README row; no forbidden phrases; trial merge clean.

/lgtm
/approve

Issue #2575, opened 2026-08-04
Insufficient repo permissions


gee4vee:
The advisory digest says the repo permissions are insufficient after installing the Hive app in the org:

<img width="934" height="825" alt="Image" src="https://github.com/user-attachments/assets/5df09943-ed46-488d-946e-6c4bf40fd76e" />

This is the app installation showing the permissions:

<img width="1112" height="1686" alt="Image" src="https://github.com/user-attachments/assets/e8891873-0bbf-4fca-bc5b-f9b3cd4bca62" />

gee4vee:
I am still seeing this issue in my Hive. Is there something I need to do to get the fix working?

<img width="1147" height="597" alt="Image" src="https://github.com/user-attachments/assets/3f486660-6ab0-48a1-969f-6dd0a3f2e9cd" />

clubanderson:
@gee4vee I dug into this — good news and actionable news:

**This isn't a stale advisory bead.** Your digest is updating on schedule (last refresh 8/20 10:06 AM), and the heal fix from #2626 is working as designed. The findings keep reappearing because your agents are **actively re-detecting a real access gap** each cycle, not because an old bead failed to clean up.

**Root cause:** all 8 guide findings + the quality blocker point at the same thing — your hive's GitHub App installation token is getting `403 Resource not accessible by integration` when reading `github.ibm.com/devx-prod/epx-vscode-ext-poc`. That means the App installation on your GHE instance either:

1. lacks the **Contents: Read** repository permission, or
2. doesn't have `devx-prod/epx-vscode-ext-poc` included in its repository access list (if installed with "selected repositories").

**Fix:** on github.ibm.com, go to the Hive App installation settings for the `devx-prod` org → ensure **Contents: Read** is granted and the repo is in the installation's repository list. Once the agents can read the repo, these findings will stop being re-filed and the healer will close the existing ones on the next digest cycle automatically.

FWIW I checked other hives' digests fleet-wide — none show this finding, so it's specific to your GHE App installation config, not a hive bug.

gee4vee:
@clubanderson can you double check the screenshot in the issue description I added? I think it should have the correct permissions already. Does it not?

clubanderson:
@gee4vee you're right — your App installation is configured correctly, and I owe you a correction. Your screenshots pushed me to dig deeper and it turned out to be a **hive bug on our side**, not your config:

**Root cause (two coupled defects, now fixed in #4289 / PR #4291):**
[...]

PR #4471, opened 2026-08-21
🐛 fix(advisory): retire the self-reported "repository worktree not provisioned" finding (#4464)

### LINKED ISSUE #4464: 🐛 bug: guide agent shows error about repository worktree
## What happened?

The guide agent is showing an error about the repository worktree configuration. It's not clear to me if this is a configuration issue or something is faulty.

## Environment

- Branch/version: `v2` / commit SHA / release tag: `9f31439295c6fe8b70df70f8f16ea76afd57a063`

## Logs or screenshots

Image

## Additional context

Anything else maintainers should know?

Danathar:
Closes #4464.

## What the screenshot is

Not a crash and not a misconfiguration — it is a finding **the hive filed against itself**, sitting in the advisory digest:

> 🔴 **[infrastructure]** Repository worktree not provisioned for guide agent despite include_repos=true configuration — `/data/agents/guide (empty directory)`

Two independent defects put it there and then kept it there. Both are fixed here.

## 1. The premise is false

`include_repos` is a **prompt-text** option: it appends the authorized-repo list to each kick (`buildReposSection`). It clones nothing and mounts nothing, and no code path materialises a per-agent worktree — `entrypoint.sh` only `mkdir -p /data/agents/<agent>`. An empty workspace directory is the **designed** state, not a broken install.

But nothing ever said that. An agent that reads `include_repos: true` in `hive.yaml.runtime`, finds its workspace empty, and concludes the infrastructure owes it a worktree is making a reasonable inference from what it was told. The repo list is where that impression comes from, so that is where the correction goes:

> ℹ️ This list is an AUTHORIZATION SCOPE, not a checkout: it does not put any repo on disk, and nothing provisions a per-agent git worktree from it. If you need files rather than the GitHub API and have no checkout, clone one yourself: `git clone <host>/<org>/<repo> /tmp/<repo>`. An absent checkout is a normal state to handle, NOT an infrastructure fault — do not file a finding about a missing worktree or unprovisioned repo workspace.

**A related false statement, found on the way.** The multi-repo block asserted "Your workdir is a clone of the primary repo only" — which is true of no deployment, since nothing clones into the agent workdir at all. It now reads "at most, a checkout of the primary — never of the others", which preserves the point the sentence exists to make (the *other* repos are definitely not here, so clone before working them) without the false half. The two `sec-check-holdgated.md` templates carry the same sentence verbatim and get the same fix.

## 2. The healer that should have retired it could not see it

`CloseHealedRepoAccessFindings` exists for exactly this family (#2575 / #4291): advisory-tier agents reporting no way to read the repo they audit, auto-closed once a real advisory-tier Contents read is verified. It never considered **either** finding in that digest, because `IsRepoAccessFinding` matched neither title. Verified against the current patterns before changing anything:

| Title from the digest | Matched |
|---|---|
| `Repository worktree not provisioned for guide agent despite include_repos=true configuration` | ❌ none |
| `Quality agent lacks read access to repository for coverage analysis` | ❌ none |

So the healer was a no-op on them and no number of cycles could age them out — the operator was looking at a permanent error.

Both gaps are **word order**, not vocabulary:

- the first states the claim in the passive past participle (`worktree **not provisioned**`) where the provisioning pattern read only the noun form (`no ... provisioning`), and says `worktree` where the pattern said `workspace`;
- the second writes `read access **to** repository` where every access pattern requires `repository ... access`.

Two new patterns cover them. **Write** access is deliberately excluded from the second: this healer's proof is a verified *read*, which cannot substantiate a write claim — those are the app-auth healer's, and a negative test pins that.

The close is still gated on the read probe, so a finding about a repo the hive genuinely cannot read stays open. `TestCloseHealedRepoAccessFindings_Issue4464` asserts both directions: both titles close when reads verify, and both survive when they do not.

## Testing

```
go vet ./pkg/advisory/ ./pkg/scheduler/     # clean
go test ./pkg/advisory/ ./pkg/scheduler/    # ok
```

`TestIsRepoAccessFinding` gains the two exact titles from the issue (transcribed from the screenshot, with their real refs), five drift wordings around them, and five new negatives that specifically guard the widened patterns — real code findings using the same words about something else (`Anonymous read access to repository artifacts is not gated`, `Terraform state bucket provisioned without encryption`, `Repository checkout action is not pinned to a commit SHA`, …). The whole pre-existing negative corpus still passes: a widened matcher must not start silently retiring genuine findings.

Two package suites fail identically **on the pristine `v4` tree** in my container (`pkg/policies` git-clone tests need a `git` binary the image lacks; two `pkg/config` read-only tests assume a non-root user). Confirmed pre-existing by stashing and re-running — untouched by this change.

## Note for the reporter

To answer the question in the issue directly: nothing was misconfigured on your side. `include_repos: true` was doing exactly what it is supposed to do; the agent misread what it promised, and the mechanism that should have cleaned up after that could not recognise the wording. Both findings should now retire on their own once the digest posts and the read path verifies — and if the read genuinely fails for `devx-prod/epx-vscode-ext-poc`, the *quality* agent's finding is a true one and will correctly stay.


clubanderson:
Reviewed on the merits — this is the right fix, and it goes further than #4469 (merged earlier today for the same issue): retiring the finding is not symptom-silencing, because the finding's premise is genuinely false (`include_repos` is prompt-text scope; nothing provisions per-agent worktrees), and the *detection* bug — the healer's pattern gaps — is fixed here too. The kick-section correction in `buildReposSection` is the upstream prevention #4469 lacked.

Since #4469 already landed the narrower healer patterns on v4, this branch conflicted in `heal.go`/`heal_test.go`. I rebased it onto v4 (maintainer-edit) and reconciled:

- `heal.go`: adopted **this PR's broader patterns** wholesale (git/agent + workspace/worktree/checkout/clone subjects; read/clone/checkout/fetch "access to" word order) — they're a strict superset of #4469's.
- `heal_test.go`: unified the duplicated fixture (#4469's `gee4veeWorktreeTitles` → this PR's `issue4464RepoAccessTitles`), kept **both** sides' drift and negative cases, and kept both end-to-end heal tests (including #4469's mixed #2575+#4464 pass and this PR's unreadable-repo gate check).
- Scheduler/docs/CHANGELOG/policy changes untouched.

`go test ./pkg/advisory/... ./pkg/scheduler/...` → 92.6% / 90.4%, all green. Will merge once CI passes. Thanks @Danathar!

PR #4472, opened 2026-08-21
🧩 feat(podman): one-command standalone installer (#4470)

### LINKED ISSUE #4470: 🧩 Podman: add a one-command standalone installer
## Parent

Part of #4188 (Phase 2). This issue must not close the parent.

## Why

The Docker path has a one-command install. The Podman path does not.

`bin/hive-setup.sh` takes an operator from a bare host to a running Hive in a
single command. The Podman equivalent is `README.md`'s Quick Start (Podman) —
correct and verified (#4412, #4448), but a **manual sequence**: create the
config directory, copy and edit `hive.yaml`, fix `dashboard.port`, set the
secrets mode and group, copy `nginx.conf` and `hive.env`, install four unit
files, `daemon-reload`, start the gateway.

That asymmetry is the Phase 2 gap. It is also where the known traps live: #4367
was a config/HealthCmd port mismatch that cost a silent 300 seconds, and #4359
was a secrets directory mode that read as an SELinux problem it was not. Both
are steps a human performs by hand today, and both are exactly what a script
does not get wrong twice.

## This is a new script, not a port

`hive-setup.sh` is `apt-get`, `docker-ce`, and `git clone --branch v2`. Do not
translate it. Three things it does are wrong for this path:

- **Do not install packages.** It runs `apt-get install docker-ce`. Podman's
  supported hosts include image-based systems (Fedora Silverblue/Bluefin, RHEL
  image mode) where `dnf install` does not work at all and `/usr` is read-only.
  Detect and report; do not attempt to fix the host.
- **Do not clone anything.** It clones `v2` into `/opt/hive`. This runs from a
  checkout that already exists.
- **Do not generate a compose file.** The deployment description is the four
  Quadlet units in `src/deploy/quadlet/`, which are already the source of truth
  and are guarded by #4404.

Follow the established naming and shape of the existing family — there are
already seven `bin/hive-podman-*.sh` scripts, and **none of them modified a
Docker script.** Keep that record intact.

## 30-minute boundary

The installer only. No changes to any Docker script, to the units, or to the
README's manual instructions — those stay as the reference the script
automates. No new preflight logic: call the existing scripts.

## Deliverable

`bin/hive-podman-setup.sh`, plus its `bin/test_hive_podman_setup.sh` sibling in
the style the rest of the family uses (every input mocked; no podman, no
systemd, no privileges needed to run the test).

## Acceptance criteria

- Runs the **existing** preflight (`bin/hive-podman-preflight.sh`,
  `-host.sh`, `-ids.sh`) and refuses to proceed on a failure, surfacing their
  guidance rather than restating it.
- Supports **both root modes** through the same flag convention the family
  already uses — `--rootful` drives the system manager through `sudo`, matching
  `bin/hive-podman-update.sh`.
- Sets `dashboard.port` to agree with the unit's `HealthCmd`. Getting this
  wrong is #4367; the script must make it impossible rather than warn about it.
- Applies the secrets directory mode and group correctly per root mode —
  `podman unshare chown` rootless, `chgrp` rootful — which is #4359.
- **Idempotent**: a second run on a configured host must not clobber an existing
  `hive.yaml` or secrets. Refuse, or require an explicit flag.
- Ends by confirming the service is **healthy**, not merely started —
  `Notify=healthy` makes that a real distinction, and the gateway answering on
  `3001` is the check.
- On any failure, says which step failed and leaves the host in a state the
  operator can inspect. Do not silently roll back.
- Does not mount a container socket anywhere (#4188's socket-isolation rule).

## Non-goals

- Podman-in-LXC. See the scope note on #4188.
- Any blue-green equivalent. See the scope note on #4188.
- Installing or upgrading Podman itself.

## Related

- #4412 / #4448 — the manual quick start this automates, and its verification
- #4367 / #4359 — the two traps that justify scripting it
- #4404 — the parity guard that keeps the two script families honest

Danathar:
Closes #4470. Part of #4188 (Phase 2) — does **not** close the parent.

## What lands

`bin/hive-podman-setup.sh` and `bin/test_hive_podman_setup.sh`, wired into `v2-ci.yml` beside its siblings.

Seven announced steps: runtime selection → pre-write preflights → configuration → the #4367 port coupling → the #4359 secrets mode → post-write preflight, pull, units, `daemon-reload` → start and confirm healthy.

## The two traps, made impossible rather than warned about

**#4367 — `dashboard.port` vs the unit's `HealthCmd` port.** The script does not carry the number. It **reads** it out of `hive.container`, writes that value into `hive.yaml`, then **reads it back from the file that will be mounted** and refuses to continue if they differ. The test drives it with a unit probing an arbitrary `3999` — a script with a literal `3002` in it passes the default case and fails that one. The read-back also runs on the re-run path, so an operator who edits the port back the way the example ships it is stopped there rather than by a silent 300-second hang with no container left to inspect.

**#4359 — the secrets command differs by root mode.** `podman unshare chown -R 0:1002` rootless, because container GID 1002 is not host GID 1002; a plain `chgrp -R 1002` rootful, where identity is mapped. Both modes assert the *wrong* command is absent, not merely that the right one is present.

## Not a port of `hive-setup.sh`

The three things the issue names as wrong for this path:

- **No packages.** Podman's supported hosts include image-based systems (Silverblue/Bluefin, RHEL image mode) where `dnf install` cannot work and `/usr` is read-only. Prerequisites are detected and reported by the existing preflights; the host is never modified to suit us. `apt-get`, `dnf`, `yum`, `apk`, `zypper`, `git` and `docker` sit on the test PATH as tripwires.
- **No clone.** It runs from the checkout it lives in.
- **No generated deployment description.** The four Quadlet units are installed **verbatim** — the test asserts `hive.container` arrives byte-for-byte, so the rationale in its header travels to the host.

No Docker script was modified, keeping the family's record intact. No unit and no README instruction is touched either — the manual sequence stays as the reference this automates, per the 30-minute boundary. **That does leave the installer undiscoverable from the README**; I left that alone deliberately rather than widen the scope, and it is a one-line follow-up if you want it.

## The preflights are invoked in a state where they check something

They run **only** when Podman is explicitly selected, so the script `export`s `HIVE_DEPLOY_RUNTIME=podman` rather than assuming it. Without that, all three exit 0 having checked nothing — three green reports about a host nobody looked at, which is worse than skipping them. A test case asserts each preflight recorded `runtime=podman`, and that the host preflight was pointed at the installed config (`HIVE_SRC_DIR=$CONF`) rather than the checkout.

On a failure the script surfaces **their** output and says so explicitly — one place to fix a check, one place to fix its advice.

## Idempotency

A second run keeps an edited `hive.yaml`, the operator's `HIVE_GITHUB_TOKEN`, and everything under `secrets/`, and reports the keeps rather than doing them silently. `--force` re-copies the config files and **still** never touches `secrets/`. The generated dashboard token is generated once.

## Healthy, not started

`Notify=healthy` is what makes that a real distinction, so ending on `systemctl start` returning would waste it. Both units must report `active` — which under `Notify=healthy` means their healthchecks passed — **and** the gateway must answer on its published port, which is the end-to-end check covering nginx, `aardvark-dns` resolving `hive`, and Hive behind it. A unit stuck `activating` and a silent gateway are separate failing cases, each with its own diagnosis, and neither prints "Hive is running".

On any failure: the step is named, nothing is rolled back, and the message says where to look. A wiped host is not inspectable.

## Socket isolation

`#4188`'s rule is **enforced** here rather than merely followed. This is the script that puts units on the host, so it refuses (exit 70) to install one naming a `docker.sock`/`podman.sock`, before installing anything. One case feeds it a doctored unit; another asserts the units that actually ship pass.

## Testing

```
bash bin/test_hive_podman_setup.sh    # 85 passed, 0 failed
bash bin/test_bin_suites_wired.sh     # 17 suites, 0 orphaned
shellcheck bin/hive-podman-setup.sh bin/test_hive_podman_setup.sh   # clean
src/scripts/check-release-lines.sh    # PASS (a workflow was edited)
bash src/deploy/test_standalone_runtime_parity.sh   # 33 passed, 0 failed
```

Every input is mocked — no podman, no systemd, no privileges, no network. The fixture is **copied from the repository's real assets** rather than hand-written, so the port parsing is exercised against the units that actually ship and cannot quietly drift from them.

## Non-goals, per the issue

Podman-in-LXC, any blue-green equivalent, and installing or upgrading Podman itself are all out of scope and untouched.


clubanderson:
/lgtm
/approve

Reviewed at d9d226e8. Scrutinized as an installer (high-risk class) and it holds up:

- **Idempotency**: keep-and-report paths for hive.yaml/hive.env/nginx.conf, `--force` re-copies configs but never touches `secrets/`, dashboard token generated exactly once (append-only). Test-covered.
- **Preflight contract respected**: `HIVE_DEPLOY_RUNTIME=podman` exported (not assumed — and the body is right that without it all three preflights green-wash), host preflight pointed at `HIVE_SRC_DIR=$CONF` per the #4422 standalone layout, exit codes 64/70/78 per sysexits.
- **Doesn't fight the drop-ins**: only the four unit files are installed; `hive.container.d/` (#4407 pin, #4411 auto-update) is never created, removed, or overwritten. Volume/network/labels come through the units, preserving the #4210 ownership contract.
- **#4367 enforced structurally**: port read from the unit's HealthCmd, written into hive.yaml, read back from the mounted file — and the test drives it with a 3999-probing unit so a hardcoded 3002 fails. Re-run path covered (hand-edited port corrected).
- **#4359 per root mode**: `podman unshare chown 0:1002` rootless vs `chgrp` rootful, with wrong-command-absent assertions both ways.
- **No remote code fetch**: the only network op is `podman pull` of the unit's own Image= reference. No packages, no clone — tripwire fakes on the test PATH confirm.
- **Socket isolation (#4188)**: exit-70 refusal on any unit naming docker/podman.sock, tested both with a doctored unit and the shipped ones.
- shellcheck clean on both files; suite wired into v2-ci.yml (orphan guard 17/0); cleanup contract 127, teardown 74 (creation guard clean — the `podman run` mention is comment-only), parity 33/0/7 all pass locally.

Local suite run here was 78/7 purely from BSD stat/sed on macOS (test harness GNU-isms); CI's ubuntu run — which is authoritative — passed. Fine as-is.

One agreed follow-up (per the PR body): the installer is undiscoverable from the README quick start — worth the one-line pointer in a docs PR.

Merging.

Issue #4478, opened 2026-08-21
🔍 Rootful Hive sits inside the system boot transaction; rootless does not — a Hive that never goes healthy holds a rootful boot for TimeoutStartSec


Danathar:
## Parent

Part of #4188. Measured while executing #4413 (reboot persistence, both root
modes). **Do not close #4188 with this.**

## What was measured

Two reboots of the same host, one per root mode, same image
(`d465287054`), same units, nothing differing but rootful vs rootless.

| | userspace | total boot | `hive-gateway` healthy at |
| --- | --- | --- | --- |
| boot 1, **rootful** | 16.854s | **19.211s** | 19.2107s |
| boot 2, **rootless** | 6.928s | **9.233s** | 18.503s |

On the rootful boot systemd's own `FinishTimestampMonotonic` was **19.2112s** —
**549µs** after `hive-gateway.service` became active. The boot was not declared
finished until Hive was serving.

On the rootless boot the same units reached healthy at 18.5s, more than nine
seconds *after* the boot had already finished at 9.2s.

## Why

Both units carry the same thing:

```ini
[Install]
WantedBy=default.target
```

In the **system** manager that target is the boot transaction, so a rootful Hive
sits on its critical path. In the **user** manager it is the user manager's own
default target, which logind reaches after the system boot has completed — and
only starts at all because `Linger=yes` (`user@1000.service` came up at +3.75s).

So the difference is structural, not incidental, and it is not visible anywhere
in the units: the same two lines mean two different things depending on which
manager generated them.

## Why it matters

`Notify=healthy` is the property ADR-0017 chose Quadlet for, and it means the
unit stays in `activating` until `/api/health` answers. Combined with the above,
on rootful that is a boot-time cost bounded by `TimeoutStartSec`:

- `hive.service` — `TimeoutStartUSec=5min`
- `hive-gateway.service` — `TimeoutStartUSec=2min`

**Measured, though not at boot:** the first rootful start on this host (with
`HIVE_DASHBOARD_TOKEN` unset, see #4476) sat in `activating` for the **full
120s** before `Result=timeout`. That is a real duration this unit reaches.

**Inferred, not measured:** that the same 120s would be spent inside the boot
transaction had it happened at boot rather than at an interactive `systemctl
start`. The mechanism above says it would, and the 549µs coupling on the good
boot is the evidence that the transaction does wait for this unit — but a
deliberately-broken rootful boot was not executed.

[...]

PR #4480, opened 2026-08-21
fix(podman): probe both container listeners so a dead dashboard cannot report healthy (#4476)

### LINKED ISSUE #4188: ✨ feature: support Podman as a first-class standalone runtime alongside Docker

### LINKED ISSUE #4476: 🐛 hive.service reports healthy while the dashboard is dead — HealthCmd probes 3002 only, never the 3001 auth proxy the gateway needs
## Parent

Part of #4188. Found while executing #4413 (reboot persistence, rootful row) on a
host that can actually be rebooted. **Do not close #4188 with this.**

## What happened

Installed the rootful Quadlet stack exactly as `src/docs/podman-standalone-quadlet.md`
describes, minus one line: `HIVE_DASHBOARD_TOKEN` was not set in `$CONF/hive.env`.
The doc does call that line required — this issue is not about the missing line.
It is about **what the units reported while it was missing**.

```
$ sudo systemctl start hive-gateway.service
Job for hive-gateway.service failed because a timeout was exceeded.
real    2m12.481s

$ sudo systemctl is-active hive.service hive-gateway.service
active          <- hive
activating      <- gateway, about to time out

$ sudo podman ps
hive   Up 2 minutes (healthy)   ghcr.io/kubestellar/hive:stable
```

`hive.service` was **active**, and its container was **healthy**, for the entire
two minutes in which the deployment was unusable.

## Why it was unusable

The image runs two listeners. The Go API is on 3002. The Node dashboard proxy —
the one that enforces `HIVE_DASHBOARD_TOKEN` — is on 3001, and it is what
`src/deploy/nginx.conf` points at:

```nginx
upstream hive_api {
    server hive:3001;
}
```

That split is deliberate and load-bearing: 3d1e6077 ("fix: critical auth bypass —
nginx routed directly to Go API, skipping auth") moved the upstream from 3002 to
3001 precisely so the gateway cannot reach the Go API without passing the auth
proxy first.

The proxy fails closed when it has no token (`src/proxy/server.js:532`):

```
[SECURITY] HIVE_DASHBOARD_TOKEN is not set and this hive is not hub-hosted
(no HIVE_SESSION_KEY / HIVE_HUB_SECRET) — all mutation endpoints would be
unauthenticated. Refusing to start. Set HIVE_DASHBOARD_TOKEN.
```

Measured inside the container at that moment:

```
$ podman exec hive curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:3001/api/health
000        <- connection refused, nothing listening
$ podman exec hive curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:3002/api/health
200
```

And the container's own listener set, read from `/proc/net/tcp{,6}`: `3002`,
`7681`, `18443`, `18444`. **No 3001.**

## The defect

`hive.container` probes one of the two listeners:

```
HealthCmd=curl -sf http://127.0.0.1:3002/api/health
```

Its own comment, thirty lines above, knows there are two:

> The service listens on 3002 (API) and 3001 (proxy)

So `Notify=healthy` — the property ADR-0017 chose Quadlet for, and which #4354's
header calls the point of the unit — certifies the half of the container nobody
reaches. The half every operator and the gateway reach can be dead, and the unit
still reports `healthy`.

**Observed:** `hive.service` active, container healthy, dashboard refusing
connections. The only red is `hive-gateway.service`, 120s later
(`TimeoutStartUSec=2min`), reported as an nginx upstream timeout.

**Expected:** the unit that owns the process reports the process's actual
readiness. A hive whose auth proxy refused to start is not healthy.

## Why this matters more than the one missing line

1. **It points the operator at the wrong unit.** The failure is in `hive`'s
   config, and the only thing that goes red is `hive-gateway`, with an nginx
   `connect() failed (111: Connection refused)` upstream error. Nothing in the
   gateway's journal names `HIVE_DASHBOARD_TOKEN`. The actual cause is one line
   deep in `journalctl -u hive.service`, in a container that reported healthy.

2. **It defeats the guard the same unit builds elsewhere.** #4367 is documented
   at length in this very file: a `dashboard.port` mismatch is caught precisely
   *because* `Notify=healthy` holds the unit in `activating`. That mechanism
   works, and then does not fire for the sibling listener.

3. **It will misreport #4413.** That is how this was found. After a reboot
   `bin/hive-podman-lifecycle-probe.sh boot-check` reads `hive.service` and would
   report the unit came back active and healthy, in a boot where the dashboard
   never came back at all. A reboot-persistence result taken today can be green
   and wrong.

## Suggested fix (not applied — #4413 says do not adjust the unit)

Probe both listeners, e.g.

```
HealthCmd=/bin/sh -c 'curl -sf http://127.0.0.1:3002/api/health && curl -sf http://127.0.0.1:3001/api/health'
```

with the same `dashboard.port` coupling caveat the existing comment already
carries for 3002, and the matching change to `src/docker-compose.yaml`'s
`healthcheck:` so "healthy" keeps meaning the same thing on both runtimes — which
is what the current comment claims it does.

Worth deciding as part of it: whether the hosted case (`IS_HOSTED`, where the
proxy boots without a token) changes what 3001 answers, since the probe must not
go red there.

## Reproduce

```sh
CONF=/etc/hive          # rootful
# stage config per src/docs/podman-standalone-quadlet.md, but DO NOT add
# HIVE_DASHBOARD_TOKEN to "$CONF/hive.env"
sudo systemctl start hive-gateway.service        # fails after 120s
sudo systemctl is-active hive.service            # active
sudo podman ps                                   # hive ... (healthy)
curl -sf http://127.0.0.1:3001/api/health        # nothing
```

## Environment

- Bluefin (Fedora Silverblue) 44.20260818, kernel 7.1.4
- podman 5.8.4, cgroup v2, netavark
- rootful Quadlet, `ghcr.io/kubestellar/hive:stable` (`d465287054 17`)
- hive @ `91136650` (v4)

## Related

- #4413 — found here; this makes its rootful/rootless rows unsafe to trust as-is
- #4354 — the unit and `Notify=healthy`
- #4367 — the `dashboard.port`/`HealthCmd` coupling this is the sibling of
- 3d1e6077 — why the gateway must reach 3001 and not 3002

Danathar:
Fixes #4476.

## The defect

The Hive image runs **two** listeners. 3002 is the Go API. 3001 is the Node auth proxy that enforces `HIVE_DASHBOARD_TOKEN` — and 3001 is what `src/deploy/nginx.conf` dials:

```nginx
upstream hive_api {
    server hive:3001;
}
```

3d1e6077 moved that upstream off 3002 deliberately, so the gateway cannot reach the API without passing auth first. `HealthCmd=` named 3002 only, so `Notify=healthy` — the property ADR-0017 chose Quadlet for — certified the half nobody reaches.

Measured, rootful, with `HIVE_DASHBOARD_TOKEN` omitted from `hive.env`: the proxy fails closed and never binds 3001 while 3002 answers 200 throughout. `hive.service` was **active** and its container **healthy** for the entire two minutes the deployment was unusable. The only red arrived 120s later on `hive-gateway.service`, as an nginx `connect() failed (111: Connection refused)` naming neither the port nor the variable — pointing the operator at the gateway when the fault was in Hive's own configuration.

## The fix

Both runtimes now probe both listeners:

```
HealthCmd=curl -sf http://127.0.0.1:3002/api/health && curl -sf http://127.0.0.1:3001/api/health
```

Podman runs a `--health-cmd` that is not a JSON array as an argument to `/bin/sh -c`, so the `&&` is the shell's. **Verified on podman 5.8.4 in both directions** — both halves passing reports `healthy`, the second half failing reports `unhealthy` — and Quadlet passes the value through as one argument, confirmed against the generator's `--dryrun`:

```
--health-cmd "curl\x20-sf\x20http://127.0.0.1:3002/api/health\x20&&\x20curl\x20-sf\x20http://127.0.0.1:3001/api/health"
```

Compose uses `CMD-SHELL` with the identical string, so `test_standalone_runtime_parity.sh` still compares the two runtimes equal and passes unchanged: `hive: both runtimes probe with \`curl -sf …3002… && curl -sf …3001…\``.

**No unit-level workaround was needed** and nothing about `Notify=healthy`, timeouts or restart policy changed.

## The question the issue asked

> whether the hosted case (`IS_HOSTED`, where the proxy boots without a token) changes what 3001 answers, since the probe must not go red there

**It does not** — only mutating methods are authenticated, so `GET /api/health` answers 200 with no credential in every mode the proxy boots in. That is now a test rather than a reading of the code. `src/proxy/health_probe.test.js` boots the real proxy against a mock Go API and pins five things:

```
PASS  self-hosted: GET /api/health on the proxy port answers 200 with no credential
PASS  self-hosted: a mutating endpoint still requires the token
PASS  hosted: GET /api/health answers 200 with no token configured at all
PASS  self-hosted with no token: the proxy refuses to start, leaving :3001 dead
PASS  proxy reports a non-2xx when the Go API behind it is unreachable
```

The second assertion is what keeps the first honest: auth is genuinely enforced in the same process, so a green run cannot mean auth was simply off. The hosted case runs with `NODE_ENV` unset, exercising the real boot guard rather than the test opt-out. This is a load-bearing guard, not decoration — if `GET /api/health` ever starts requiring a credential, the probe can never pass, `Notify=healthy` holds the unit in `activating` for the whole `TimeoutStartSec`, and `--rm` deletes the evidence. That would look exactly like #4367 and would ship green through every dry-run gate.

## A hazard the fix itself created, caught by the repo's own suite

`bin/hive-podman-setup.sh` reads the probe port back out of the unit and **writes it into the operator's `hive.yaml`** as `dashboard.port`, precisely so the #4367 coupling cannot drift. Its pattern led with a greedy `.*`, which matches through to the *last* URL on the line — with two probes it returned **3001** and would have installed `dashboard.port: 3001`: the Go API then serving on the proxy's port, the 3002 probe never answering, and a 300-second hang. #4367 again, caused by the fix for #4476.

`bin/test_hive_podman_setup.sh` catches it — 4 of its 85 assertions fail on the greedy pattern and pass on `[^:]*://`, which cannot cross the first scheme colon. Both extractors are now explicit about which listener they mean.

## Gates, so it cannot drift back

`src/deploy/test_quadlet_config_contract.sh` gains the second coupling, asserted against its sources of truth rather than a literal 3001 — the same way sections 2–4 treat the API port. Verified against three seeded regressions:

| Seeded regression | Caught by |
|---|---|
| unit reverted to 3002-only | `HealthCmd probes 3002 and nothing else. The auth proxy on the port nginx.conf dials is unprobed` |
| `nginx.conf` upstream moved to 3002 | `the gateway dials hive:3002 but hive.container probes 3001` |
| Compose left probing one port | `the hive healthcheck probes [3002] but hive.container probes [3002 3001]` |

`bin/hive-podman-lifecycle-probe.sh check` now reports the healthcheck too. That matters for #4413: `check` and `boot-check` read `hive.service` **and nothing else**, which is how this bug was found, so a reboot-persistence result taken on a one-listener unit could be green in a boot where the dashboard never came back. It reads the *generated* `ExecStart` rather than the checkout, because the installed unit is what that manager will run and may predate this change, and it counts `/api/health` URLs over the whole line rather than parsing the `--health-cmd` argument — Quadlet escapes its spaces as `\x20` while `systemctl show` renders `argv[]` space-joined, so the boundaries differ by which you read. Both renderings and the no-healthcheck case are covered in its suite.

## Verification

All green locally, on the committed tree:

```
test_quadlet_config_contract.sh     PASS   (11 assertions, health ports 3002 3001)
test_standalone_runtime_parity.sh   PASS
test_quadlet_port_boundary.sh       PASS
test_standalone_image_refs.sh       PASS
test_standalone_service_contract.sh PASS
test_changelog_reminder.sh          PASS
test_hive_podman_setup.sh           PASS   (85 assertions)
test_hive_podman_lifecycle_probe.sh PASS   (22 assertions)
test_hive_podman_update.sh          PASS
test_hive_podman_cleanup.sh         PASS
test_hive_podman_teardown.sh        PASS
test_hive_standalone_runtime.sh     PASS
test_bin_suites_wired.sh            PASS
src/proxy npm test                  PASS   (full suite, including the new file)
```

## Deliberately not changed

- **`src/deploy/docker-compose.architect.yaml`** keeps its 3002-only probe. That stack publishes 3002 directly and has no gateway, so the auth proxy is not in its path and probing it would assert something that stack does not depend on.
- **The Kubernetes readiness probe.** Out of scope here; the cluster Service owns its own edge.
- **`hive.container`'s image, timeouts, restart policy and `Notify=healthy`** are untouched.

## Does not close #4188

#4476 is a sub-issue of that umbrella. This PR carries no closing keyword for it.


clubanderson:
/lgtm
/approve

Reviewed at 65b97046. This fixes the #4476 defect I verified real during the #4479 review, and it holds up:

- **Netns semantics correct**: 3001 is the hive container's OWN second listener (the Node auth proxy `server.js` binds from `HIVE_PROXY_PORT`, default 3001 — verified), not the gateway container. Probing it from 127.0.0.1 inside the same netns is right; no cross-container reach-through smuggled into HealthCmd.
- **Negative path is real, not asserted**: measured on podman 5.8.4 both directions per the body, and `health_probe.test.js` boots the actual proxy process 5 ways — including `self-hosted with no token: refuses to start, leaving :3001 dead` (the exact #4476 failure mode) and the mutating-endpoint assertion that keeps the unauthenticated-GET result honest. Ran it here: 5/5, real spawns.
- **Hosted mode answered**: GET /api/health needs no credential in every boot mode (tested with NODE_ENV unset, exercising the real boot guard) — so the stricter probe can't go red on hosted hives.
- **#4478 not worsened**: HealthInterval/Timeout/Retries/TimeoutStartSec all untouched; the only boot-time change is that a genuinely dead dashboard now honestly holds `activating` instead of lying green — which is the point.
- **Rollback semantics intact (#4407/#4420)**: still Notify=healthy + start-job result; nothing about restart policy or sdnotify changed.
- **The self-created hazard is the best part**: the fix's second URL would have made setup.sh's greedy `.*` extractor write the PROXY port into dashboard.port — #4367 reintroduced by the #4476 fix — and his own 85-assertion suite caught it (4 assertions fail on the greedy pattern). Extractor now `[^:]*://`, plus an explicit second-port extractor, and the installer reports/warns on one-listener units.
- **Gates against drift**: config contract grows sections 8–9 asserting against sources of truth (server.js default, nginx upstream — both verified: `hive:3001`, `|| '3001'`), not the literal; compose/quadlet parity now compares full port lists ordered; lifecycle probe check reads the GENERATED ExecStart (handles both \x20-escaped and argv-joined renderings) and flags pre-fix installed units.
- Parity 33/0/7, config contract 11 (ports 3002 3001), cleanup contract 127, orphan guard 17/0 — all pass locally. Setup 78/7 and probe-suite 17/5 locally are the known BSD stat/sed/find fixture limitations (baseline v4 shows the same class); CI ubuntu — authoritative — is green.
- architect-compose left 3002-only with correct reasoning (no gateway in that stack's path). shellcheck clean; no forbidden phrases.

Merging; will close #4476 stays automatic via the Fixes keyword. #4188 correctly unswept.

Merging.

PR #4481, opened 2026-08-21
fix(docs): the Docker Compose quick start writes .env where Compose reads it, and sets the token it cannot start without (#4477)

### LINKED ISSUE #4477: 🐛 README's Docker Compose Quick Start is broken twice: .env written where Compose never reads it, and HIVE_DASHBOARD_TOKEN never set
## How this was found

Not a Podman issue, and **not part of #4188** — this is the Docker Compose path,
and it is broken on its own terms. It surfaced sideways: the same
`HIVE_DASHBOARD_TOKEN` refusal was hit while executing #4413 on the Quadlet
stack, and checking whether Compose shared the fault turned up two separate
faults in the README's **default** quick start instead.

## The quick start, as shipped

`README.md`, "Quick Start (Docker Compose)" — the path the README calls the
default and the one the rest of the file assumes:

```bash
git clone https://github.com/kubestellar/hive.git
cd hive

cp src/hive.yaml.example src/hive.yaml
echo "HIVE_GITHUB_TOKEN=ghp_..." > .env
docker compose -f src/docker-compose.yaml up -d
```

> Dashboard at `http://localhost:3001`.

Two things are wrong with it, independently.

---

## 1. The `.env` is written where Compose does not look

`-f src/docker-compose.yaml` sets the project directory to `src/`, and that is
where Compose looks for `.env`. The file the quick start writes at the repo root
is never read. Executed, docker 29.7.2:

```
$ printf 'HIVE_GITHUB_TOKEN=ghp_SENTINEL_ROOT\n' > .env       # exactly as the README says
$ docker compose -f src/docker-compose.yaml config
WARN The "HIVE_GITHUB_TOKEN" variable is not set. Defaulting to a blank string.
WARN The "HIVE_DASHBOARD_TOKEN" variable is not set. Defaulting to a blank string.
WARN The "NTFY_SERVER" variable is not set. Defaulting to a blank string.
WARN The "NTFY_TOPIC" variable is not set. Defaulting to a blank string.
```

The same file one directory down is read:

```
$ printf 'HIVE_GITHUB_TOKEN=ghp_SENTINEL_SRC\n' > src/.env
$ docker compose -f src/docker-compose.yaml config | grep HIVE_GITHUB_TOKEN
      HIVE_GITHUB_TOKEN: ghp_SENTINEL_SRC          # no warning
```

as is the root file when the project directory is stated:

```
$ docker compose --project-directory . -f src/docker-compose.yaml config | grep HIVE_GITHUB_TOKEN
      HIVE_GITHUB_TOKEN: ghp_SENTINEL_ROOT
```

**Observed:** an operator who follows the quick start verbatim gets a hive with
an empty `HIVE_GITHUB_TOKEN`, having just pasted a real PAT into a file that is
ignored. `src/.gitignore` ignores `src/.env`, so nothing complains about either
location.

What that looks like afterwards is not a startup failure — the hive comes up and
then 401s on everything, which reads as a bad token rather than an unread file.
From a run on this host with the same variable unset:

```
"level":"WARN","msg":"failed to fetch issues","error":"... 401 Bad credentials []"
"level":"ERROR","msg":"failed to enumerate actionable items","error":"all 2 repos failed ... 401 Bad credentials []"
```

**Expected:** the documented command reads the file the same document just told
the operator to create.

---

## 2. `HIVE_DASHBOARD_TOKEN` is never mentioned, and the dashboard needs it

The image runs the Go API on 3002 and a Node dashboard proxy on 3001. The proxy
is what `src/deploy/nginx.conf` points at (`upstream hive_api { server hive:3001; }`),
deliberately — 3d1e6077 moved the upstream off 3002 so the gateway cannot reach
the API without passing auth first.

That proxy **fails closed** with no token (`src/proxy/server.js:532`). Observed
directly, same `ghcr.io/kubestellar/hive:stable` image, on this host:

```
[entrypoint] Starting Node.js proxy on :3001 → :3002
[SECURITY] HIVE_DASHBOARD_TOKEN is not set and this hive is not hub-hosted
(no HIVE_SESSION_KEY / HIVE_HUB_SECRET) — all mutation endpoints would be
unauthenticated. Refusing to start. Set HIVE_DASHBOARD_TOKEN.

$ podman exec hive curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:3001/api/health
000        <- connection refused
$ podman exec hive curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:3002/api/health
200
```

`gateway` and `hive` are both in the default Compose profile and `gateway`
publishes 3001, so `http://localhost:3001` — the address the quick start
advertises — reaches nginx, which proxies to a port nothing is listening on.

**Observed:** the advertised dashboard serves 503. Fault 1 guarantees fault 2 is
reached even if the operator did put a token in their `.env`, since that file is
not read either.

**Expected:** the quick start sets the token it cannot start without, the way the
**Podman** quick start in the same README already does:

```bash
printf 'HIVE_DASHBOARD_TOKEN=%s\n' "$(openssl rand -hex 32)" >> "$CONF/hive.env"
```

`src/deploy/quadlet/hive.env.example` states it plainly — "REQUIRED unless this
hive is hub-hosted … Hive REFUSES TO START without it" — and
`src/docs/podman-standalone-quadlet.md` makes it a numbered step. Only the
Compose path, which is the default and the first thing a new operator runs, omits
it.

---

## Suggested fix

```bash
git clone https://github.com/kubestellar/hive.git
cd hive

cp src/hive.yaml.example src/hive.yaml
echo "HIVE_GITHUB_TOKEN=ghp_..." > src/.env
printf 'HIVE_DASHBOARD_TOKEN=%s\n' "$(openssl rand -hex 32)" >> src/.env
docker compose -f src/docker-compose.yaml up -d
```

`src/.env` (not `--project-directory .`) keeps the env file beside the compose
file that declares the variables, and matches where `src/.gitignore` already
expects it. Either works; pick one and say which.

Worth doing alongside: `src/deploy/test_standalone_service_contract.sh` parses
`docker-compose.yaml` already and could assert that every `${VAR}` the compose
file substitutes is named in the quick start — this pair of faults is exactly the
kind a contract test catches and a reader does not.

## Environment

- Bluefin (Fedora Silverblue) 44.20260818, docker 29.7.2, podman 5.8.4
- `ghcr.io/kubestellar/hive:stable` (`d465287054 17`)
- hive @ `91136650` (v4)

## What was executed vs traced

Fault 1 executed under Docker Compose directly (`docker compose config`, above) —
no containers started, no pull. Fault 2's refusal observed live in the same image
under rootful Podman/Quadlet; the Compose reach to it is by the shared
`nginx.conf` upstream and the default-profile port publish, read from
`src/docker-compose.yaml`, not by a Compose boot.

## Related

- #4413 — the Podman work being executed when this surfaced; no dependency either way
- #4476 — the same missing token is invisible to the Quadlet unit's health check,
  and `src/docker-compose.yaml`'s `healthcheck:` probes the same single port
- 3d1e6077 — why the gateway must reach 3001, not 3002

Danathar:
Fixes #4477.

### 1. The `.env` was written where Compose does not look

`-f src/docker-compose.yaml` sets the project directory to `src/`, and that is where Compose reads `.env` from — the same place the compose file's own `./hive.yaml` and `./secrets` mounts already resolve against, so **every other line of the recipe already assumed it**. Reproduced here on docker 29.7.2 / Compose **v5.5.0**, confirming the report's v2 finding across both majors:

```
$ printf 'HIVE_GITHUB_TOKEN=ghp_SENTINEL_ROOT\n' > .env      # exactly as the README said
$ docker compose -f src/docker-compose.yaml config
WARN The "HIVE_GITHUB_TOKEN" variable is not set. Defaulting to a blank string.
      HIVE_GITHUB_TOKEN: ""

$ printf 'HIVE_GITHUB_TOKEN=ghp_SENTINEL_SRC\n' > src/.env
$ docker compose -f src/docker-compose.yaml config | grep HIVE_GITHUB_TOKEN
      HIVE_GITHUB_TOKEN: ghp_SENTINEL_SRC
```

The operator pastes a real PAT into a file nothing reads. Both paths are gitignored, so neither git nor Compose objects — the hive starts and then 401s on every GitHub call, which reads like a bad token rather than an unread file.

**`src/.env` over `--project-directory .`**, as the issue asked someone to pick and say which. The flag would have to be repeated on every subsequent Compose command in every document, and one omission puts the operator straight back here; `src/.env` needs no flag, sits beside the compose file that declares the variables, and is where `src/.gitignore` already expects it.

### 2. `HIVE_DASHBOARD_TOKEN` was never set

The auth proxy on 3001 refuses to start without one, and `nginx.conf` points the gateway at that proxy deliberately (3d1e6077), so `http://localhost:3001` — the address this quick start advertises — served 503 from an upstream that never bound.

The repo already knew this. `bin/hive-setup.sh` generates a token *and* passes `--env-file` explicitly; `src/deploy/quadlet/hive.env.example` calls it REQUIRED in as many words; the **Podman** quick start in this same README mints one as a numbered step. Only the default Compose path omitted it.

## Fixed in all three copies, not one

The recipe is duplicated more widely than the issue found — `README.md`, **`src/README.md` (both Option A and Option B)**, and **`docs/development.md`**. All four instances carried both faults; all four are fixed. The Compose section now also ends by confirming the dashboard answers, as the Podman section already did, rather than asserting a port that is published either way. `openssl` joins its prerequisites.

`src/docs/backup-restore.md` carried the same fault in prose, telling the reader to add `HIVE_BACKUP_KEY` to `.env`.

## The gate the issue asked for

> `src/deploy/test_standalone_service_contract.sh` … could assert that every `${VAR}` the compose file substitutes is named in the quick start — this pair of faults is exactly the kind a contract test catches and a reader does not.

`src/deploy/test_compose_quickstart_contract.sh`, as its own gate — the service-contract test is scoped to what the *deployment description* says, and this is about what the *operator is handed*. Per fenced code block:

- an env file written beside a Compose command goes to `src/.env`, unless the command states `--project-directory` or `--env-file`;
- an **install** block — one that runs Compose *and* bootstraps from `hive.yaml.example` — writes an env file and sets every required variable.

**Nothing is restated.** The required set is read out of `src/deploy/quadlet/hive.env.example`, the tracked environment contract for both runtimes, which already marks `HIVE_DASHBOARD_TOKEN` REQUIRED. `test_standalone_runtime_parity.sh` holds that file's names equal to the compose file's, so a variable that becomes required on one runtime cannot quietly not be required on the other. Documents are **discovered**, not listed, so a fourth copy of the quick start is gated the day it is written.

**The install/operational split is load-bearing, not convenience.** `up -d` after restoring a volume (`backup-restore.md`) or after a `down -v` (`troubleshooting.md`) restarts a stack whose env file already exists; demanding a freshly minted token there would be wrong. A file-wide grep would also confuse the README's *Podman* recipe, which writes a different env file at a different path — hence block scoping.

Verified against the shipped docs at `upstream/v4` — **8 failures, one per fault per recipe**:

```
FAIL: README.md: the env file is written where Compose reads it
      block at line 29 writes `.env`; -f src/docker-compose.yaml makes src/ the project directory…
FAIL: README.md: the install recipe sets HIVE_DASHBOARD_TOKEN
      block at line 29 bootstraps a stack without it…
FAIL: docs/development.md: …  (both)
FAIL: src/README.md: …        (both, twice — Option A and Option B)
```

It also asserts the parser found a non-empty required set, so a green run cannot mean the contract quietly changed shape.

## Two supporting changes

**The compose file's optional variables get `:-` defaults; the tokens deliberately do not.** Compose warns once per command for each `${VAR}` it cannot resolve, and ntfy is off by default, so an ordinary quick start printed two permanent warnings that mean nothing — training the operator to skim past the line that says a variable is unset. **That line is the only signal Compose gives when the env file is not being read at all** — it is fault 1's sole diagnostic. A fifth assertion fails if a *required* variable ever gains a `:-` default, since that would substitute empty, the proxy would fail closed exactly as before, and the warning that said so would be gone. Verified it fires.

**`README.md` and `docs/**` join the v2-ci path filter.** The quick start lives in files the workflow did not watch, so this gate would not have run on the one kind of change that can break it. `src/docs/**` already triggers the workflow through `src/**`; this applies the same rule to the docs outside `src/`.

## Verification

The fixed recipe, run verbatim:

```
$ echo "HIVE_GITHUB_TOKEN=ghp_QUICKSTART_SENTINEL" > src/.env
$ printf 'HIVE_DASHBOARD_TOKEN=%s\n' "$(openssl rand -hex 32)" >> src/.env
$ docker compose -f src/docker-compose.yaml config
      HIVE_DASHBOARD_TOKEN: 88350a9e0a156b9677adefeb3324bcb79affd7b34116642e491fc6795ba6a198
      HIVE_GITHUB_TOKEN: ghp_QUICKSTART_SENTINEL
```

Both tokens substitute; no warnings remain. Local gate sweep, all green:

```
test_compose_quickstart_contract.sh   PASS   (7 assertions)
test_standalone_service_contract.sh   PASS
test_standalone_runtime_parity.sh     PASS
test_quadlet_config_contract.sh       PASS
test_quadlet_port_boundary.sh         PASS
test_standalone_image_refs.sh         PASS
test_watchtower_socket_contract.sh    PASS
test_changelog_reminder.sh            PASS
test_hive_podman_setup.sh             PASS
test_hive_standalone_runtime.sh       PASS
test_bin_suites_wired.sh              PASS
```

## Scope

Not part of #4188 and no closing keyword for it — this is the Docker Compose path, broken on its own terms. Related to #4476 (the same missing token is invisible to the Quadlet unit's health check) but independent of it; neither PR depends on the other.


clubanderson:
Reviewed and verified at head 9c249b42 (+ my conflict-resolution merge 25022777).

- **Both #4477 faults and their fixes verified against code**: `-f src/docker-compose.yaml` makes `src/` the project directory, so `src/.env` is the file Compose reads (matches the compose file's own `./hive.yaml`/`./secrets` mount resolution); `HIVE_DASHBOARD_TOKEN` refusal is real (`src/proxy/server.js` fails closed, nginx upstream is `hive:3001` — the issue's measured 503 chain holds); the token line in `hive.env.example:33` says REQUIRED verbatim; `curl /api/health → {"status":"ok"}` matches `handleHealth` (server.go:1958). Token minting mirrors the Podman quick start exactly (`openssl rand -hex 32`), keeping the two paths separated but consistent.
- **Contract test is the right shape and works**: 7/7 at head; required set read from `hive.env.example` not restated (parses exactly `HIVE_DASHBOARD_TOKEN`); documents discovered not listed (6 found; the 3 recipe carriers checked, operational-only docs correctly noted, install-vs-restart distinction sound). Mutations all caught locally: root-`.env` regression → PATH fail with the project-directory explanation; token line removed → VAR fail; `${HIVE_DASHBOARD_TOKEN:-}` added to compose → the fails-empty-and-silences-the-warning case fail. The `:-` on the two optional ntfy vars (and deliberately not the tokens) is a nice touch — that warning is the only signal an unread env file gives.
- **No collateral**: `test_standalone_runtime_parity.sh` 33 pass and `test_standalone_service_contract.sh` 12 pass with the compose edit; new v2-ci.yml path filters (README.md, docs/**) are exactly what makes this gate run on the changes that can break it; shellcheck clean on the new test; actionlint shows only the pre-existing SC2024/SC2034; `src/.env` git-ignored (asserted via check-ignore, not by reading .gitignore). Placeholder tokens only; no forbidden phrases.
- Resolved the CHANGELOG both-added conflict against #4480's entry (kept both), DCO-signed, all contracts re-run green on the merged tree.

/lgtm
/approve

PR #4482, opened 2026-08-21
probe(podman): measure the rootful boot #4478 left inferred, and record the mode difference in the support matrix

### LINKED ISSUE #4478: 🔍 Rootful Hive sits inside the system boot transaction; rootless does not — a Hive that never goes healthy holds a rootful boot for TimeoutStartSec

Danathar:
Addresses #4478. **No closing keyword**, and no unit changes — see "What this deliberately does not do".

The issue is explicitly the "record it and file it" half of #4413, and lists four options for a maintainer. Option 1 is already done. Options 2 and 3 are unit changes and are the maintainer's call. This PR does **option 4**, and closes the one gap the issue named in its own evidence.

## The half that was inferred is now measured

#4478 established the coupling on a **good** boot — systemd's `FinishTimestampMonotonic` **549µs** after `hive-gateway.service` went active, so the rootful boot was not declared finished until Hive was serving — and then said plainly what it could not measure:

> **Inferred, not measured:** that the same 120s would be spent inside the boot transaction had it happened at boot rather than at an interactive `systemctl start`. The mechanism above says it would […] but a deliberately-broken rootful boot was not executed.

Executing that on the measuring host means rebooting it into a broken state, which is why it stayed inferred. **It does not have to be that host.** A systemd container's PID 1 runs the same system manager and the same job transaction, and is disposable.

`src/deploy/probe_boot_transaction_coupling.sh` boots three, differing in exactly one thing — whether the unit is reached through `default.target.wants/`:

```
Observed
        case      wired?         boot finished    state
        control   yes            +0.154s          running
        broken    yes            +20.315s         degraded
        unwired   no             +0.133s          running

Findings
  PASS  control: boot finished 8117us after the unit went active
  PASS  broken: the boot was held 20.188s -- the unit's whole TimeoutStartSec (20s)
        Result=timeout; the timeout was spent INSIDE the boot transaction
  PASS  unwired: boot finished +0.133s against broken's +20.315s
  PASS  unwired: an interactive start of the same unit blocked 20s
```

`TimeoutStartSec=20s` to keep it short; the shape scales, the number does not.

The **control** reproduces the 549µs coupling in miniature. The **broken** case is the consequence — `Result=timeout`, the whole timeout spent inside the boot transaction.

### The third case is the argument

Without it, `broken` only shows that a failing unit takes `TimeoutStartSec` to fail, which nobody doubted. The same unit, same timeout, **not** in `default.target.wants/`, finishes the boot in **0.133s** — and starting it by hand afterwards blocks the caller for 20s. The cost does not vanish on rootless; it moves off the boot and onto whoever asked.

That is exactly the rootful/rootless split: the Quadlet generator installs `default.target.wants/hive.service` in **both** modes, and only the system manager's `default.target` is the boot transaction.

### Scope, stated rather than assumed

**Establishes** the *manager* behaviour — the half that was inferred. **Does not** re-measure Hive, and does not claim a container boot is a bare-metal boot: no firmware, no initrd, and the stand-in unit is a `sleep`. The Hive half is what the two real reboots in #4413 already measured. What was missing was the link between them, and the link is a property of systemd rather than of the image.

## Option 4: the support matrix

> Add it to #4324's mode matrix as a rootful/rootless difference, since it is the kind of thing that decides which mode an operator picks and is currently invisible.

#4324 closed with `src/docs/podman-support-matrix.md`, which grades one axis: whether the egress-gate claim was measured for a cell. An operator picking a root mode is deciding more than that.

Added as its **own section, not a new column**, because it is not a support grade — it costs availability, not enforcement, and moves no cell. The section says so, and says explicitly that options 2 and 3 stay open in #4478 rather than being settled by a page whose job is to publish decisions already taken.

### One stale gap corrected while there

`Restart, reboot, recreate — **Unmeasured in every cell**` has not been true since #4377 measured stop/start/restart/recreate and #4413 executed a real reboot in each root mode. **Narrowed rather than deleted:** those runs did not inspect the egress chain afterwards, so the half of that gap belonging to that page — that the *gate* re-installs reliably across a restart — is still unmeasured and still says so. I checked before rewording: the lifecycle page mentions neither the gate nor advisory mode, so nothing there licenses inferring the chain came back.

- **No unit changes.** `DefaultDependencies=`, different ordering, and a shorter rootful `TimeoutStartSec` are options 2 and 3, and remain the maintainer's call. This only removes the reason to guess at what they would be trading against.
- **Does not close #4478** — the decision it asks for is still open — and carries no closing keyword for **#4188**, whose sub-issue this is.

## Safety and reproducibility

Name discipline follows `probe_podman_volume_persistence.sh`: every object is `hive-boottx-<pid>-*`, the probe **refuses to start** if any of those names already exists, and cleanup removes exactly what it created. Nothing on the host is reconfigured, nothing is rebooted, no existing container/image/volume is read or removed. It needs only `podman` plus registry access and skips cleanly without either.

Reproduced twice — the boot held **20.188s** and **20.116s** of a 20s timeout — cleanup verified to leave nothing behind, `shellcheck -S warning` clean, and the probe indexed in `src/deploy/README.md` alongside the existing one.

Local gate sweep, all green: `test_standalone_service_contract.sh`, `test_standalone_runtime_parity.sh`, `test_quadlet_config_contract.sh`, `test_quadlet_port_boundary.sh`, `test_standalone_image_refs.sh`, `test_bin_suites_wired.sh`, `test_hive_podman_lifecycle_probe.sh`.

## Where an operator actually reads it

The matrix section above is the reference page. It is not where someone
installing looks, and the two places that are were both silent — one of them
misleadingly so.

- **`## One unit file, both root modes`** is the operator guide's answer to
  "what differs between the modes", and it listed exactly one thing: `%E`. That
  is the difference the unit file *shows* you. The one it cannot show is that
  both modes read the same `[Install] WantedBy=default.target` and mean
  different things by it. Now a second table beside the `%E` one.

- **`### 4. Boot persistence`** was worse than silent. It ends *"Rootless
  additionally needs lingering … Rootful needs no equivalent; the system manager
  is PID 1"* — which reads as rootful being strictly the simpler choice, when
  being PID 1's `default.target` is exactly what puts a rootful Hive on the
  boot's critical path. The cost now follows that sentence directly, with the
  549µs good-boot figure, the `TimeoutStartSec` worst case, and the three
  concrete causes that reach it (wrong `dashboard.port`, unreachable registry,
  missing `HIVE_DASHBOARD_TOKEN`) — each already documented in **Traps** on the
  same page.

It also says plainly that **neither mode is defective and nothing needs changing
to install either**: you asked for Hive at boot and systemd is waiting for it to
be ready. The one lever available *without* a unit change — `TimeoutStartSec` in
a drop-in — is named with what it trades away, since that headroom exists for a
~3.8GB first pull. That is not option 3; option 3 is changing the shipped unit's
default, which stays a maintainer's call.

README gets one paragraph in the Podman section, beside the lingering
requirement it is the counterpart to, linking to the guide rather than restating
it. **Deliberately not added to the "Security posture — pick deliberately"
table** there: that table grades enforcement, this costs availability, and
blurring the two would weaken both.


clubanderson:
/lgtm
/approve

Reviewed at da0da553 (all three commits, including the operator-facing placement commit pushed after the head this review was queued with).

- **The measurement is real and well-designed**: three throwaway systemd containers differing in exactly one variable — `default.target.wants/` membership — with the unwired control doing the argumentative work (same unit, same timeout, boot finishes +0.133s vs +20.315s; interactive start still blocks the caller 20s, so the cost moves rather than vanishes). Scope honestly stated: establishes the MANAGER behaviour (the half #4478 left inferred), doesn't re-measure Hive or claim container boot = bare-metal boot. The link between #4413's real reboots and the failure cost is a systemd property, and that's what got measured.
- **systemd semantics correctly characterized**: system manager's `default.target` is the boot transaction (FinishTimestamp waits on its job), user manager's is reached by logind post-boot; Type=notify + never-READY holds the start job for TimeoutStartSec with `Result=timeout` — consistent with #4407/#4420's start-job rollback semantics and #4480's unchanged health timing (no unit or timeout touched here).
- **Probe safety**: `hive-boottx-$$-*` name discipline per the volume-persistence precedent, refuses to start on collision, removes exactly what it created (name-scoped `rm -f`/`rmi -f`, not the forbidden bulk forms), skips cleanly without podman or registry. `shellcheck -S warning` clean here. Correctly a probe_, not CI-wired — matches precedent, orphan guard 17/0.
- **Support matrix update is the right shape**: own section, not a column — availability cost, not an enforcement grade, and it says so. The 'Restart, reboot, recreate' gap correctly NARROWED rather than deleted (gate re-install across restart still unmeasured — verified the lifecycle page indeed never inspects the egress chain). Operator-guide and README placements accurate, including the honest correction that '§4 Boot persistence' previously read as rootful being strictly simpler.
- **Branch predates today's #4480 merge**; trial merge with v4 is clean and the suites on the merged tree pass (config contract 11 w/ ports 3002 3001, parity 33/0/7, cleanup contract 127, service contract 12). Lifecycle-probe suite 17/5 locally = known BSD fixture limitation, CI ubuntu green.
- **#4478 disposition**: correctly NOT closed — this executes option 4 and removes the inference, but options 2/3 (unit changes: DefaultDependencies/ordering vs shorter rootful TimeoutStartSec) remain the maintainer decision the issue exists to hold. #4188 correctly unswept.

Merging.

Danathar:
Independent reproduction of this PR's probe on a different host, run while doing a full end-to-end pass of #4188. Same three-case shape, same conclusion, numbers close enough to be the same measurement:

```
Observed
        case      wired?         boot finished    state
        control   yes            +0.204s          running
        broken    yes            +20.343s         degraded
        unwired   no             +0.140s          running

Findings
  PASS  control: boot finished 5754us after the unit went active
  PASS  broken: the boot was held 20.183s -- the unit's whole TimeoutStartSec (20s)
        Result=timeout; the timeout was spent INSIDE the boot transaction
  PASS  unwired: boot finished +0.140s against broken's +20.343s
  PASS  unwired: an interactive start of the same unit blocked 21s

Result
  no findings -- the boot transaction waits, and default.target.wants/ is why
```

Against the PR's recorded `+0.154s / +20.315s / +0.133s`. Host: Bluefin 44 (Silverblue), kernel 7.1.4-200.fc44, SELinux enforcing, Podman 5.8.4, rootless. The probe is self-contained and needed no arguments.

Two things worth noting for review:

- **The script runs cleanly off-branch.** I fetched only this file and ran it standalone; it built its three images, booted them, cleaned up after itself, and touched nothing else on the host. That matches the PR's claim that the measurement does not need the measuring host to be sacrificed.
- **The third case really is the argument.** I would have read `broken` alone as "a failing unit takes its timeout to fail", which nobody disputes. `unwired` finishing in 0.140s while an interactive start of that same unit blocks 21s is what makes it about `default.target.wants/` rather than about the unit being slow.

Also confirming the PR is not blocked on anything mechanical: all 27 checks pass, and tide reports `Not mergeable. Needs approved, lgtm labels` — it is waiting on review only.

Separately, on the substance of #4478 rather than this PR: I have not been able to execute the real-host reboot half. I am running *on* the machine under test, so rebooting ends the session doing the observing. That is exactly the constraint this PR works around, which is a point in favour of the approach.

PR #4500, opened 2026-08-21
🐛 hive-podman-setup.sh: check lingering in rootless mode instead of claiming success that will not survive a reboot

### LINKED ISSUE #4489: 🐛 hive-podman-setup.sh reports success without lingering, so a rootless install silently does not survive reboot
`bin/hive-podman-setup.sh --rootless` finishes with **"Hive is running"** and a summary of the dashboard, config, secrets, and token paths. It never mentions lingering, never checks it, and never warns. On a host where `Linger=no` — the default on most systems — the deployment it just declared healthy **will not come back after a reboot**, and nothing in its output says so.

## The installer is silent

```
$ grep -in linger bin/hive-podman-setup.sh
(no matches)

$ grep -iE 'linger|boot|reboot|persist' <installer output>
(nothing)
```

Final output, verbatim, on a host with lingering off:

```
Hive is running
  Dashboard:  http://localhost:3001
  Config:     /var/home/dbaggett/.config/hive/hive.yaml   (edit for your project, then: systemctl --user restart hive.service)
  Secrets:    /var/home/dbaggett/.config/hive/secrets     (mode 750, group 1002; put your GitHub App key here)
  Tokens:     /var/home/dbaggett/.config/hive/hive.env    (HIVE_GITHUB_TOKEN, HIVE_DASHBOARD_TOKEN)
```

## The project already knows this is required

`src/docs/podman-standalone-quadlet.md`:

> `loginctl enable-linger` is **not optional** for a rootless install that must survive a reboot.

And `src/docs/podman-quadlet-lifecycle.md` measured it, on a throwaway user with no login session — which is what boot looks like:

| | `Linger` | `user@UID.service` | the user's enabled unit |
| --- | --- | --- | --- |
| A. before | `no` | `inactive` | **never ran** |
| B. `loginctl enable-linger` | `yes` | `active/running` within 305ms | started, no session involved |
| C. `loginctl disable-linger` | `no` | `inactive/dead` | process **gone** |

with the conclusion that makes this silent:

> Throughout all three the unit's `default.target.wants/` symlink was present and unchanged, which is the point: **nothing in the unit's own enablement notices that it will never run.**

So there is no self-check an operator would stumble into. `systemctl is-enabled` reports `generated` either way — the same doc says so explicitly.

## The shipped probe does catch it

`bin/hive-podman-lifecycle-probe.sh check`, run immediately after a successful install on this host:

```
Boot wiring -- what actually decides whether Hive comes back
        is-enabled reports 'generated' -- 'generated' is expected and proves nothing
  PASS  the generator installed default.target.wants/hive.service
  FAIL  lingering is OFF for dbaggett -- at boot there is no session, so the user
        manager never starts and Hive never runs
        fix: loginctl enable-linger dbaggett

Result
  1 finding(s)
```

The check exists, it is correct, and it is the only thing on the host that reports the problem — but the installer does not run it and the README does not tell you to run it *after installing*. The README mentions `enable-linger` once, in prose, in the middle of the manual twenty-step sequence that `hive-podman-setup.sh` exists to replace. An operator who uses the installer never reads that line.

## Why this is the worst-timed failure mode

Everything looks correct until a reboot, which may be weeks later and is usually unattended:

- the stack is healthy and serving on `:3001`
- `systemctl --user is-enabled hive.service` says `generated`
- `default.target.wants/hive.service` is present
- `bin/hive-podman-teardown.sh plan` lists the running resources

and after the reboot Hive is simply gone, with no failed unit to find, because the user manager never started. For a self-hosted customer that is an outage discovered by absence.

## Suggested fix

`src/docs/podman-host-execution-gap.md` records that `loginctl enable-linger` was **deliberately not run** by earlier work because "it reconfigures the host". That is a reasonable line to hold, so this is not a request to enable it silently. But then the installer has to *say* something. In rough order of preference:

1. **Check and warn, loudly, in the rootless summary.** The installer already knows the root mode and the user. `loginctl show-user "$USER" -p Linger` is one call. If `Linger=no`, the closing block should say Hive will not survive a reboot and print the exact fix.
2. **Offer it.** A `--enable-linger` flag, or a prompt when interactive, keeping the default as "do not reconfigure the host".
3. **Run the existing check.** The installer's step 7 already confirms health; having it also run `hive-podman-lifecycle-probe.sh check` and surface findings would have caught this with no new logic.

(1) is the minimum and is a small change. Rootful needs none of this — the system manager is PID 1 — so the warning should be rootless-only.

Also worth adding `bin/hive-podman-lifecycle-probe.sh check` to the README's Podman quick start as the post-install verification step. It is currently mentioned only for the `is-enabled` caveat.

## Environment

Bluefin 44 (Silverblue), Podman 5.8.4, rootless, `Linger=no` (untouched default), `v4` at `18e0736f`. Found during the #4188 end-to-end pass. I will execute the actual reboot later in this pass and report the observed result here.

Part of #4188

clubanderson:
Fixes #4489

## Root cause

`bin/hive-podman-setup.sh` ended every rootless install with **"Hive is running"** without ever checking lingering — `grep -in linger bin/hive-podman-setup.sh` matched nothing. Rootless boot persistence hinges on exactly that one fact: with `Linger=no` (the untouched default on most hosts, including the reporter's Bluefin 44) there is no session at boot, so the user manager never starts and Hive never runs. Nothing an operator would check notices, as measured in `src/docs/podman-quadlet-lifecycle.md`: `systemctl --user is-enabled` reports `generated` either way, and the generator's `default.target.wants/` symlink is present and unchanged. So the failure is discovered weeks later, by absence, after an unattended reboot.

The shipped `bin/hive-podman-lifecycle-probe.sh check` does catch it — but the installer never ran it and the README's quick start never told you to.

## Fix (the issue's option 1 as the floor, option 2 opt-in)

New final step **`8/8 Boot persistence — will this install survive a reboot?`**:

- **Rootless** reads `loginctl show-user "$USER" -p Linger --value` — the same read the lifecycle probe uses, so the two can never disagree. `Linger=no` produces a loud FAIL-styled block naming the exact fix (`loginctl enable-linger $USER`) plus a pointer at the probe, and the closing summary carries an explicit red `Reboot: will NOT survive a reboot` line. Exit stays 0 — the deployment **is** healthy now, and the issue's stated minimum is to warn loudly rather than fail a healthy install.
- The summary **never claims reboot safety it did not read back**: `Reboot: survives a reboot — lingering is enabled for USER` appears only when `loginctl` said `yes`.
- **`--enable-linger`** opts in to running `loginctl enable-linger`, verified by reading the state back. Off by default: `src/docs/podman-host-execution-gap.md` records that enabling lingering was deliberately not automated because it reconfigures the host, and this installer's contract is that the host is never modified to suit us. A flag that cannot be honoured (e.g. needs sudo) exits 78 with the privileged form, rather than shrugging.
- **Rootful untouched**: reports "rootful needs no lingering — the system manager is PID 1", calls no `loginctl`; `--rootful --enable-linger` is a usage error (64).

## Evidence

Contract suite (`bin/test_hive_podman_setup.sh`, CI-wired in v2-ci.yml) grows six cases / 27 assertions over a recording `loginctl` fake: rootless+linger-off warns with the fix and reconfigures nothing; rootless+linger-on claims persistence; `--enable-linger` enables and read-backs; a failing `--enable-linger` exits 78; rootful never calls `loginctl`; the flag pair is refused.

```
111 passed, 1 failed   # the 1 is the known pre-existing BSD-install macOS fixture limit;
84 passed, 1 failed    # same baseline failure on untouched origin/v4 on the same host
```

shellcheck clean on both scripts.

## Docs

- README Podman quick start now ends with `bin/hive-podman-lifecycle-probe.sh check` as the post-install verification step, as the issue asked.
- `src/docs/podman-standalone-quadlet.md` §4 Boot persistence documents the installer's check and `--enable-linger`.
- CHANGELOG entry under Unreleased → Fixed.


Danathar:
Conflict is **`CHANGELOG.md` only** — every code file merges clean. Checked with `git merge-tree` against `origin/v4`:

```
CONFLICT (content): Merge conflict in CHANGELOG.md
```

and nothing else. Cause is #4499 and #4501 landing underneath and adding their own entries at the top, so it is a keep-both-entries resolution rather than anything needing thought.

**Worth knowing it is systemic, not one-off.** #4500 and #4504 don't conflict with *each other* — they touch disjoint scripts (`hive-podman-setup.sh` vs `hive-podman-teardown.sh`) — they each conflict with `v4` on the changelog. Which means whichever of these merges first will re-conflict the other on the same file, and #4498, #4503, #4505 and #4509 will hit it in turn as they land, since they all add entries too.

If it is worth avoiding six rounds of this, the usual dodges are a changelog fragment directory (one file per PR, assembled at release) or simply appending new entries at the bottom of the section rather than the top. Entirely your call — flagging it because six PRs in flight against one hot file is the shape where it starts costing more than the fixes did.

On #4500 specifically: its 11 pending/failing checks look like the conflict blocking CI rather than a real failure. Probably worth re-reading them after the rebase rather than chasing them now.

Context: found while checking the state of the ten issues from the #4188 end-to-end pass. No action needed from me — I have no push access to these branches and they are yours; just passing on the diagnosis so you don't have to redo it.

Issue #4491, opened 2026-08-21
🐛 lifecycle-probe.sh exercise leaves hive-gateway down and the dashboard dead, then reports "no findings"


Danathar:
`bin/hive-podman-lifecycle-probe.sh exercise` ends with the Hive dashboard **down**, reports **`no findings`**, and exits **0**.

## Reproduction

Healthy stack before, nothing else touched:

```
BEFORE: gateway=active   health={"status":"ok"}

$ bin/hive-podman-lifecycle-probe.sh exercise
...
Result
  no findings
probe EXIT=0

AFTER:  gateway=inactive health=DEAD
```

`:3001` is the only published port and the documented dashboard. After the probe it serves nothing.

## Cause

The probe drives one unit:

```
UNIT="hive.service"
VOLUME_UNIT="hive-data-volume.service"
```

`hive-gateway.service` is never started or stopped by it — the string `gateway` appears exactly once in the file, inside an informational message.

The sequence:

1. probe runs `systemctl --user stop hive.service`
2. `hive-gateway.service` has `Requires=hive.service`, so systemd stops the gateway too — correct, and the probe is not doing anything wrong here
3. probe runs `systemctl --user start hive.service`
4. **nothing starts the gateway.** `Requires=` propagates *stop*, not *start*. The dependency points gateway→hive, so bringing hive up does not bring the gateway up.

Every assertion the probe makes is about `hive.service`, and every one of them is true, which is why it reports no findings. It simply never looks at the thing it broke.

## Why this matters beyond the inconvenience

- **It is a diagnostic tool that damages what it diagnoses, then certifies it.** An operator runs it precisely when they are unsure whether the deployment is healthy. It answers "no findings" while the dashboard is dead.
- **The failure is silent in the same way #4489 is.** `hive.service` is `active`, the container is `healthy`, `podman ps` looks right. Only `:3001` is gone, and only if you check it.
- **The probe's own summary reinforces the wrong conclusion.** Its closing "Current state" block prints `unit`, `container` and `volume` — all of them `hive.service`'s. There is no gateway row to be conspicuously missing.

It also cost me real time in this end-to-end pass: I found the gateway down some minutes later, while starting an unrelated update test, and initially attributed it to a different probe.

## Suggested fix

Any of these; (1) is the obvious one:

[...]

PR #4503, opened 2026-08-21
🐛 update.sh rollback restores the gateway and only claims serving once the published port answers

### LINKED ISSUE #4493: 🐛 update.sh rollback reports "rolled back and it is serving" while :3001 is dead — the gateway is never restarted after a failed update
`bin/hive-podman-update.sh rollback` recovers `hive.service` correctly and reports

```
Result
  PASS  rolled back to sha256:d465704b… and it is serving
```

while `:3001` — the only published port and the dashboard an operator actually uses — is **dead**, and stays dead until someone starts `hive-gateway.service` by hand.

This is the **recovery path from an outage**. The operator has just had a five-minute outage from a bad update, runs the documented recovery, is told it worked, and the dashboard is still down.

## Measured

Full sequence on a healthy rootless Quadlet deployment, `v4` at `18e0736f`.

**1. Pin a published image that cannot be healthy as Hive** (the shipped nginx digest — deliberately a *published* image rather than a locally-built stand-in):

```
  FAIL  restart returned 1 after 301s, state activating/auto-restart/timeout
  FAIL  the new image did not become healthy -- Hive is DOWN on this host
        roll back now, deliberately:
            bin/hive-podman-update.sh rollback

hive activating / gateway inactive / dashboard DEAD / NRestarts=0
```

That part is all correct and matches #4378 and #4411 exactly — 301s is one full `TimeoutStartSec`, the unit never reads `failed`, and `NRestarts=0` confirms `Restart=always` never fires.

**2. Roll back, as the script instructs:**

```
Rollback -- rootless (user manager, uid 1000)
  PASS  returning to sha256:d465704b…  (first pinned from ghcr.io/kubestellar/hive:stable)
        stopped the failing unit first: inactive/dead/timeout
  PASS  restart returned 0 after 11s, state active/running/success
        with Notify=healthy that return means /api/health answered on the new digest
Result
  PASS  rolled back to sha256:d465704b… and it is serving
ROLLBACK EXIT=0
```

**3. What is actually true afterwards:**

```
hive     active
gateway  inactive
hive container: hive Up 33 seconds (healthy)

dashboard :3001                        = DEAD
hive direct :3002 inside the container = {"status":"ok"}
```

Hive is serving. Nothing an operator can reach is.

## Cause

```
$ grep -c gateway bin/hive-podman-update.sh
0
```

The script models one unit — `UNIT="hive.service"` — and the gateway is collateral:

1. the bad pin makes `hive.service` time out; `hive-gateway.service` has `Requires=hive.service`, so systemd stops the gateway. Correct.
2. `rollback` runs `systemctl stop hive.service` (line 524, "stopped the failing unit first") and then `systemctl restart hive.service` (line 340).
3. a `restart` re-starts dependents that are **currently active**. The gateway is already `inactive` from step 1, so there is nothing for the transaction to restart.
4. no code path ever starts the gateway again.

Note this does **not** happen on a *successful* `pin` against a healthy stack: there the gateway is active when the restart job runs, so systemd cycles it and it comes back. I verified that separately — `pin` to a good digest left `hive-gateway  Up 14 seconds (healthy)`. The gateway is only stranded when it was already down, which is precisely the failed-update case that `rollback` exists to serve.

## Why "and it is serving" is the wrong claim

The check behind that line is `Notify=healthy` on `hive.service`, which means Hive's own `/api/health` answered on `127.0.0.1:3002` *inside the container*. That is a true statement about Hive and a false statement about the deployment. `bin/hive-podman-setup.sh` already gets this right — its step 7 is explicitly *"Start, and confirm HEALTHY rather than started"* and it ends by curling the gateway on `:3001` from the host:

```
  -> waiting for the gateway on http://127.0.0.1:3001/api/health
```

The update script performs no equivalent check.

## Suggested fix

1. **`rollback` (and `pin`) must ensure the gateway is up before claiming success.** `systemctl --user start hive-gateway.service` is idempotent and a no-op when it is already running, so this needs no conditional logic.
2. **Reuse the installer's end-to-end assertion.** Confirm `http://127.0.0.1:<gateway port>/api/health` answers from the *host* before printing `and it is serving`. The installer already knows how; the wording there is the wording this needs.
3. **Say what is down while it is down.** The failure output in step 1 is otherwise excellent — it says plainly "Hive is DOWN on this host" and gives the exact recovery command. It should also note that the gateway has stopped, so an operator watching `:3001` understands why.

## Relationship to #4491

Same class — a script that models only `hive.service` and leaves `hive-gateway.service` stranded — but a different script, a different trigger, and a worse consequence. #4491 is a diagnostic tool taking the dashboard down during a deliberate stop/start. This is the **documented recovery procedure** failing to restore service and reporting success. They should probably be fixed together, and the general lesson is that any script which cycles `hive.service` needs to treat the gateway as part of the deployment rather than as something systemd will take care of.

## Environment

Bluefin 44 (Silverblue), SELinux enforcing, Podman 5.8.4, rootless, Quadlet units from `bin/hive-podman-setup.sh`. Found during the #4188 end-to-end pass.

Part of #4188

clubanderson:
## Root cause

`bin/hive-podman-update.sh` models exactly one unit (`UNIT="hive.service"` — `grep -c gateway` on the old script returned 0). After a failed update:

1. The bad pin makes `hive.service` time out; `hive-gateway.service` has `Requires=hive.service`, so systemd **stops the gateway**. Correct so far.
2. `rollback` runs `systemctl stop hive.service` and then `systemctl restart hive.service`.
3. A `restart` transaction only re-starts dependents that are **currently active**. The gateway is already `inactive`, so it is not in the transaction.
4. **No code path ever starts the gateway again.**

The success line — "rolled back to … and it is serving" — rested on `Notify=healthy`, which proves Hive's *in-container* listeners answered (per #4480 the HealthCmd probes both 3002 and 3001 in-netns) and says nothing about the **published** port `:3001`, the only thing an operator can reach. Measured state after the old rollback: `hive active / gateway inactive / dashboard :3001 DEAD / :3002 answering inside the container`, exit 0. The documented recovery from an outage reported success while the outage continued.

This never bit on a `pin` against a *healthy* stack, because there the gateway is active when the restart job runs and systemd cycles it — the gateway is only stranded when it was already down, which is precisely the failed-update case `rollback` exists for.

## Fix

- Every path that restarts hive (`pin`, `rollback`, `unpin`, `autoupdate on`) now runs `ensure_gateway_serving`: `systemctl start hive-gateway.service` (idempotent — a no-op on a healthy stack, the fix on a stranded one), then curls `http://127.0.0.1:<port>/api/health` from the **host** before any success claim, reusing the installer's end-to-end assertion. The port is read from the *installed* `hive-gateway.container` (`PublishPort=`), defaulting to 3001.
- The success message is now honest: "… and it is serving **end to end**" prints only after the gateway answered on its published port. A hive-healthy/gateway-dead outcome is exit 78 and says exactly what is down, while still recording the digest `healthy` — Hive served on it; the fault is in front of Hive.
- A **failed** update now says the gateway has stopped with it (`Requires=`), so an operator watching `:3001` understands why it is dark.
- `restart_and_report`'s claim is scoped truthfully: `Notify=healthy` means both in-container listeners answered (#4476/#4480); the published port is owned by the gateway check.

## Evidence

`bash bin/test_hive_podman_update.sh` (mocked systemd/podman/curl, Linux semantics): **93 passed, 0 failed** (was 77). New assertions include:

- failed update → `rollback` **actually starts the gateway** (`systemctl --user start hive-gateway.service` in the call log) and probes `127.0.0.1:3001/api/health` from the host, claiming "healthy end to end"
- **the lying-success negative case**: gateway curl fails → rollback exits **78, not 0**, never prints "and it is serving", and says the gateway did not answer
- a gateway that will not *start* is a finding ("nothing an operator can reach"), not a footnote
- the probe port is read from the installed unit (`PublishPort=8443:3001` → probes 8443)
- a failed `pin` states the gateway stopped with hive, and does not try to start the gateway over a broken hive
- `pin`/`unpin` with a dead gateway exit 78; the pin history still records the digest `healthy`

`shellcheck -S warning` clean on both scripts (remaining info-level SC2016 notes are the suite's pre-existing eval'd-check idiom). `bin/test_bin_suites_wired.sh`: 17 suites, 0 orphaned. No forbidden podman storage-wipe commands introduced.

Same class as #4491 (a script modelling only `hive.service` strands the gateway), fixed here for the update/rollback script.

Fixes #4493


Danathar:
This covers every path that runs through the script, and the end-to-end assertion borrowed from `hive-podman-setup.sh` is exactly the right shape. One path it cannot reach, though, and I measured it rather than inferring it: **unattended `podman auto-update --rollback`**.

The PR touches `bin/hive-podman-update.sh`, its test, the workflow and the docs — no `.container` file. But when the auto-update timer rolls back, podman restarts the unit **over D-Bus itself**. It never invokes `hive-podman-update.sh`, so none of the new `GATEWAY_UNIT` handling runs.

Measured on a live rootless deployment with the shipped 300s `TimeoutStartSec`, a real registry, and a bad image pushed over a floating tag:

```
start: 16:28:32
  UNIT          CONTAINER            IMAGE                     POLICY      UPDATED
  hive.service  b8363439d6a6 (hive)  localhost:5000/hive:auto  registry    rolled back
auto-update rc=0   end: 16:33:45

AFTER: hive=active  gw=inactive
       running=45ad3d76298129fbf6fa5852     <-- good digest restored, unaided
       NRestarts=0
       dash=DEAD
```

The rollback itself is correct — 313s, good digest restored, `rc=0`, `NRestarts=0`. And the gateway journal shows it stopped once and was never started again:

```
16:28:33 Stopping hive-gateway.service - Hive authenticating gateway...
16:28:33 Stopped  hive-gateway.service - Hive authenticating gateway.
                          (no further entries — no Starting, ever)
```

Hive healthy on `:3002` inside the container, `:3001` dead, nobody watching, `OnCalendar=daily`.

That path is arguably the one that matters most, because every signal an operator would alert on looks fine: `hive.service` active, container healthy, right digest, `NRestarts=0` — and `podman-auto-update.md` already establishes, correctly, that `is-failed` and `ActiveState` never go bad during this sequence.

**Suggestion, if you want it closed in the same change:** couple the gateway to hive in the *start* direction on the unit rather than in each caller — `BindsTo=hive.service` + `After=`, or `Upholds=` from hive on systemd ≥249. That covers auto-update, `systemctl stop`/`start` by hand, and the `exercise` stop/start pair from #4491, in one place, and would let the script-side additions here be a belt-and-braces rather than the only guard.

Not a blocker on this PR — what it fixes, it fixes correctly, and the failed-update message finally saying the gateway has stopped is a real improvement. Just flagging that #4493 will not be fully closed by merging it, so the issue may want to stay open or spawn a follow-up for the unit change.

Full detail and the successful-auto-update contrast (where the gateway *does* come back, because it was active when the restart job ran) are in https://github.com/kubestellar/hive/issues/4493#issuecomment-5375105002.

PR #4504, opened 2026-08-21
🐛 stop the Quadlet units before Podman teardown removes the resources they own

### LINKED ISSUE #4484: 🐛 Podman teardown removes the network without stopping its unit, so the next install fails with `network not found`
`bin/hive-podman-teardown.sh` removes Hive's containers, volumes, and network with `podman` directly. It contains **no `systemctl` calls at all**:

```
$ grep -c systemctl bin/hive-podman-*.sh
bin/hive-podman-cleanup.sh:0
bin/hive-podman-teardown.sh:0        <-- removes the resources
bin/hive-podman-setup.sh:8
bin/hive-podman-lifecycle-probe.sh:8
bin/hive-podman-update.sh:7
```

The Quadlet units that own those resources are left loaded and running. For the network that is not cosmetic: `hive-network.service` is a run-once unit that stays `active (exited)` after it has created the network. Removing the network underneath it does not change its state, and systemd has no way to notice.

The next start then skips it, because `Requires=hive-network.service` is already satisfied:

```
● hive-network.service - Hive container network
     Active: active (exited) since Fri 2026-08-21 14:19:36 EDT

$ podman network ls
NETWORK ID    NAME        DRIVER
2f259bab93aa  podman      bridge          <-- no `hive`
```

so `podman run --network hive` fails:

```
hive[126808]: Error: unable to find network with name or ID hive: network not found
systemd[1503]: hive.service: Main process exited, code=exited, status=125/n/a
```

## What the operator sees

`bin/hive-podman-setup.sh` fails at the last step with a message that names neither the network nor the cause:

```
7/7  Start, and confirm HEALTHY rather than started
A dependency job for hive-gateway.service failed. See 'journalctl -xe' for details.
  FAIL  systemctl start hive-gateway.service failed

FAILED at step: 7/7  Start, and confirm HEALTHY rather than started
```

Exit 78. The named unit (`hive-gateway.service`) is two levels away from the one that is actually wrong, and `journalctl -xe` on the *gateway* does not contain the `network not found` line — that is on `hive.service`. The preflights do not catch it either: they check the engine, IDs, storage, networking backend, labels, and port 3001, none of which is unit state. All three preflights pass immediately before the failure.

Re-running the installer does not help, which is the part that makes this expensive: it is idempotent, so it reinstalls the same units, reruns `daemon-reload`, and fails identically. Nothing in the output suggests `systemctl`.

## Reproduction

Measured on Bluefin 44 (Silverblue), SELinux enforcing, Podman 5.8.4, rootless, `v4` at `18e0736f`.

```bash
bin/hive-podman-setup.sh --rootless          # succeeds, stack healthy
bin/hive-podman-teardown.sh run --yes        # removes the network; unit stays active (exited)
bin/hive-podman-setup.sh --rootless          # FAILS at 7/7, exit 78
```

The journal shows the divergence directly. Three earlier teardowns on this host each have a matching `Stopped hive-network.service`; the one at 14:27:10 does not, and that is the one that left the host broken:

```
14:19:25  Stopped hive-network.service          <-- network removed WITH the unit stopped: fine
14:19:36  network create ... (name=hive)
14:27:10  network remove ... (name=hive)        <-- no `Stopped` line: unit still active
15:32:12  Error: unable to find network with name or ID hive: network not found
```

## Confirmation of the cause

One `systemctl` call, changing nothing else, and the identical install succeeds:

```bash
systemctl --user stop hive-network.service    # the only intervention
bin/hive-podman-setup.sh --rootless --skip-pull
```

```
7/7  Start, and confirm HEALTHY rather than started
  PASS  start returned
  PASS  hive.service is active (Notify=healthy: that means its healthcheck passed)
  PASS  hive-gateway.service is active (Notify=healthy: that means its healthcheck passed)
```

That isolates it: the resource removal is not the problem, the unit state left behind is.

## Suggested fix

Teardown should stop the units before removing what they own, in the reverse of the order `setup` starts them, and `reset-failed` afterwards. That is also the only way to handle a second problem in the same script, which I have not yet measured and am recording rather than asserting: `hive.container` carries `Restart=always` / `RestartSec=30`, so `podman rm --force` on a container whose unit is still running invites systemd to recreate it 30 seconds later. A teardown that races its own work would present as "teardown succeeded, container came back".

The ownership contract from #4210 is not in question here — label scoping worked correctly and only Hive-owned resources were removed. The gap is that the contract covers **podman resources** and says nothing about the **systemd units** that own them, and under ADR-0017 the units are the lifecycle.

Two smaller things worth folding in, both visible above:

- The setup script's step 7 failure should say which dependency failed and why. It knows the unit names; `systemctl --user is-active hive.service` and the last journal line for it would have named the cause immediately.
- A preflight that checks for exactly this skew — a `hive-*` unit reporting `active` whose podman resource does not exist — would catch it before anything is written. This is a natural fit for `hive-podman-preflight-host.sh`, which already runs at step 6 and again inside step 2.

## Scope note

Found while running the full #4188 end-to-end pass on a live host. The stack is otherwise healthy: `:3001` answers `{"status":"ok"}`, `:7681` is unpublished and unreachable from the host, `/terminal` returns 401 through the gateway, and the ownership labels are present on both containers.

Part of #4188

clubanderson:
Fixes #4484

## Root cause

`bin/hive-podman-teardown.sh` contained no `systemctl` call at all. It removed the network (and containers/volumes) with `podman` directly while the Quadlet-generated units that own them stayed loaded and running:

- `hive-network.service` is run-once and stays `active (exited)` after creating the network. Removing the network underneath it does not change unit state, so on the next install `Requires=hive-network.service` is already satisfied, creation is skipped, and `hive.service` dies with `unable to find network with name or ID hive: network not found` (exit 78 at setup step 7/7, blamed on `hive-gateway.service` two dependency levels away).
- `hive.container` carries `Restart=always` / `RestartSec=30` with no restart limit, so `podman rm --force` under a running unit is undone 30 seconds later — and after a "successful" teardown the unit retried `network not found` forever (measured in the issue comments: one restart every 30s, unbounded).

Danathar isolated it on the live host: `systemctl --user stop hive-network.service` alone, changing nothing else, made the identical install pass end to end.

## Fix

`run` now stops the four generated services in the reverse of the order setup starts them — `hive-gateway.service`, `hive.service`, `hive-data-volume.service`, `hive-network.service` — and `reset-failed`s each, **before** any resource removal. Details:

- **Idempotent**: a unit that is not loaded (never installed, or gone since) is reported and skipped; re-running teardown succeeds.
- **Scoped**: the Quadlet units hard-code the `default` instance, so a teardown scoped via `HIVE_DEPLOY_INSTANCE` to another instance leaves them alone; systemd scope (`--user` vs system) follows the invoking user, matching how the resources are reachable.
- **Fails safe**: a loaded unit that refuses to stop aborts the teardown rather than removing resources underneath a still-running unit.
- **`plan` stays read-only**: it announces the stops (`would run: systemctl --user stop …`) without issuing a single `systemctl` call.
- Unit files and configuration remain installed (the maintainer-scope question from the issue's point 3 is deliberately untouched); a fresh `bin/hive-podman-setup.sh` regenerates cleanly through its own `daemon-reload`. Usage text and `src/docs/podman-ownership-cleanup.md` now say so.

The #4210 ownership guard is untouched — `systemctl` does not go through it (the contract covers podman/buildah only), and every podman command still does.

## Evidence

`bin/test_hive_podman_teardown.sh` gains a fake `systemctl` with per-unit `LoadState`/`ActiveState` and now asserts (74 → **101** assertions, all green):

- units are stopped in reverse start order, `reset-failed` runs, and the **last stop precedes the first removal** (ordered event log);
- unit state is gone (`inactive`) after teardown;
- **teardown → install cycle succeeds**: the fake reproduces the run-once semantics (start creates the network only from a non-active unit), and a post-teardown start recreates it — plus a negative control documenting the old skew (unit left `active`, start creates nothing);
- re-teardown is safe, including with no units installed at all;
- `plan` and non-default-instance runs issue no `systemctl` call.

Suites: teardown contract **101/101**, cleanup contract **127/127**, `shellcheck -x` clean on both changed scripts. (The `test_hive_podman_setup.sh` suite shows 7 pre-existing environment failures identical on clean `origin/v4` — unrelated.)

No `--all`, no prune, no store-wide form anywhere in the path, exactly as before.


Danathar:
Conflict is **`CHANGELOG.md` only** — every code file merges clean. Checked with `git merge-tree` against `origin/v4`:

```
CONFLICT (content): Merge conflict in CHANGELOG.md
```

and nothing else. Cause is #4499 and #4501 landing underneath and adding their own entries at the top, so it is a keep-both-entries resolution rather than anything needing thought.

**Worth knowing it is systemic, not one-off.** #4500 and #4504 don't conflict with *each other* — they touch disjoint scripts (`hive-podman-setup.sh` vs `hive-podman-teardown.sh`) — they each conflict with `v4` on the changelog. Which means whichever of these merges first will re-conflict the other on the same file, and #4498, #4503, #4505 and #4509 will hit it in turn as they land, since they all add entries too.

If it is worth avoiding six rounds of this, the usual dodges are a changelog fragment directory (one file per PR, assembled at release) or simply appending new entries at the bottom of the section rather than the top. Entirely your call — flagging it because six PRs in flight against one hot file is the shape where it starts costing more than the fixes did.

On #4500 specifically: its 11 pending/failing checks look like the conflict blocking CI rather than a real failure. Probably worth re-reading them after the rebase rather than chasing them now.

Context: found while checking the state of the ten issues from the #4188 end-to-end pass. No action needed from me — I have no push access to these branches and they are yours; just passing on the diagnosis so you don't have to redo it.

PR #4510, opened 2026-08-21
🐛 Docker→Podman migration: rootless `:z` cannot relabel Docker's root-owned archive

### LINKED ISSUE #4497: 🐛 Docker→Podman migration fails verbatim on SELinux: rootless `:z` cannot relabel Docker's root-owned archive
The Docker → Podman migration commands in `src/docs/backup-restore.md`, run **verbatim** on an SELinux-enforcing host, fail at the extract step:

```
$ podman run --rm -v hive-data:/data -v "$PWD":/backup:z docker.io/library/alpine:3.22 \
    tar xzf /backup/docker-hive-data.tar.gz -C /data
tar: can't open '/backup/docker-hive-data.tar.gz': Permission denied
```

The procedure is otherwise correct — with one added step it completes and the migration is verifiably real. But as written it does not work rootless, which is the mode the same section uses (`systemctl --user start …`) and the mode the README documents as the default.

## Mechanism

It is **SELinux**, not ownership. The file is world-readable:

```
$ ls -ln docker-hive-data.tar.gz
-rw-r--r-- 1 0 21 331541 docker-hive-data.tar.gz      # mode 0644 — readable by anyone
$ head -c 4 docker-hive-data.tar.gz && echo READABLE   # from the host shell
READABLE
```

The labels are what differ:

| file | written by | SELinux label |
| --- | --- | --- |
| `docker-hive-data.tar.gz` | Docker's rootful daemon | `system_u:object_r:user_tmp_t:s0` |
| `hive-data.tar` (podman-side backup, same dir) | the operator's own uid | `system_u:object_r:container_file_t:s0` |

Step 1 runs `docker run … -v "$PWD":/backup alpine tar czf …`, and Docker's daemon is root, so the archive lands **root-owned** with the directory's `user_tmp_t` label.

Step 3 mounts that directory `:z` so Podman will relabel it — but **rootless Podman cannot relabel a file it does not own.** `chcon` requires ownership or `CAP_FOWNER`, and a rootless user has neither over a root-owned file. `:z` skips it, reports nothing, and the container domain is denied.

Inside the container the file also shows as `65534:65534` (`nobody`), since host uid 0 falls outside the rootless subuid range — but that is cosmetic here, because 0644 would still permit the read. The denial is the label.

## Confirmed by minimal change

One `chown`, the **identical** podman command:

```
$ sudo chown $(id -u):$(id -g) docker-hive-data.tar.gz
$ podman run --rm -v hive-data:/data -v "$PWD":/backup:z docker.io/library/alpine:3.20 \
    tar xzf /backup/docker-hive-data.tar.gz -C /data
  EXTRACT OK
$ getfattr -n security.selinux --only-values docker-hive-data.tar.gz
system_u:object_r:container_file_t:s0      # :z relabelled it once it was ours
```

## The migration itself is sound

With that step added, the full procedure works and the result is verifiably a migration rather than data that merely survived. I used three distinct identities so a wrong outcome would be visible:

| | before | after migration |
| --- | --- | --- |
| Docker side | `hive-able-moth` + `_migration_marker` | (stopped, volume retained) |
| Podman side | `hive-free-lynx` | **`hive-able-moth`** + `MIGRATED-FROM-DOCKER` |

```
start rc=0 in 11s  state=active
hive-id:  hive-able-moth
marker:   MIGRATED-FROM-DOCKER
dash:     {"status":"ok"}
```

The two "this cannot work" claims in the same section also reproduced exactly, so the export/import approach is correctly justified:

```
$ podman volume inspect src_hive-data
Error: no such volume src_hive-data
$ podman run --rm -v /var/lib/docker/volumes/src_hive-data/_data:/data:ro alpine ls /data
Error: statfs /var/lib/docker/volumes/src_hive-data/_data: permission denied
$ stat -c '%a %U:%G' /var/lib/docker
710 root:root
```

## Why this is worth fixing rather than leaving to the operator

**The failure mode is quiet and points the wrong way.** On my first run the extract failed, `hive-data` stayed empty, and `systemctl --user start hive.service` then returned **rc=0, healthy in 11s** — on a fresh volume, with a newly minted identity (`hive-zero-crow`). A healthy dashboard on `:3001` after a "migration" is exactly what success looks like.

**What saves the operator is step 5**, and it is load-bearing:

> `# 5. only now, once step 4 matched: docker compose … down -v`

An operator who checks `hive-id` as step 4 instructs will catch this. One who reads "healthy" and proceeds destroys the source data. That ordering deserves to be called out as the safety property it is, not just a numbered step.

**"Permission denied" invites the wrong fix.** The natural reactions — `chmod 777`, or `setenforce 0` — are both wrong, and the second is exactly what `hive-podman-preflight-host.sh` goes out of its way to avoid recommending ("None of this requires weakening SELinux; the kernel stays enforcing either way"). The migration section should hold the same line.

## Suggested fix

Add the ownership step to the procedure, between steps 1 and 3:

```sh
# 2a. The archive is written by Docker's root daemon, so it is root-owned and
#     keeps the directory's label. Rootless Podman cannot relabel a file it does
#     not own, so `:z` in step 3 would silently skip it and the extract fails
#     with "Permission denied". Take ownership first:
sudo chown "$(id -u):$(id -g)" docker-hive-data.tar.gz
```

Rootful Podman does not need it, so it belongs with the section's other rootless/rootful notes.

Worth considering as well: have step 1 write the archive somewhere the operator already owns, or `docker run --user "$(id -u):$(id -g)"` so it is never root-owned in the first place. That removes the step rather than documenting it, though it needs a check that the Docker-side `alpine` can still read `/data`.

Also worth a line in the "what was observed" table: this section records the migration as **executed**, and its step 3 is rootless. On an SELinux-enforcing host these exact commands do not complete, so whatever was run had a step that did not make it into the page.

## Environment

Bluefin 44 (Silverblue), SELinux **enforcing**, Podman 5.8.4 rootless, Docker 29.7.2, `ghcr.io/kubestellar/hive:stable`, `v4` at `18e0736f`. Found during the #4188 end-to-end pass.

Part of #4188

Danathar:
Closes #4497.

## The problem

The Docker → Podman migration in `src/docs/backup-restore.md` does not complete when run verbatim on an SELinux-enforcing host — which is the host this page's own "what was observed" section says it was executed on, in the rootless mode the README documents as the default.

```
$ podman run --rm -v hive-data:/data -v "$PWD":/backup:z docker.io/library/alpine:3.22 \
    tar xzf /backup/docker-hive-data.tar.gz -C /data
tar: can't open '/backup/docker-hive-data.tar.gz': Permission denied
```

It is the label, not the mode. Step 1 has Docker's rootful daemon write the archive, so it lands root-owned with the directory's `user_tmp_t` type. Step 3 mounts `:z` so Podman will relabel it to `container_file_t` — but relabelling is `chcon`, and `chcon` needs ownership or `CAP_FOWNER`. A rootless user has neither over a root-owned file, so Podman skips it, reports nothing, and the container domain is denied. The file is `0644` and reads fine from the operator's own shell, which is exactly what makes the message point the wrong way.

## What this changes

**`src/docs/backup-restore.md`**

- **Step 2a**, one `sudo chown "$(id -u):$(id -g)"`, between stopping Docker and the extract — after which the *identical* podman command succeeds and the archive comes out `container_file_t`. Noted as rootless-only, alongside the section's other root-mode notes.
- **The mechanism**, so the step is not a cargo-culted incantation: why `0644` is not enough, what `:z` is for, and why it silently declines.
- **"Do not `chmod 777` or `setenforce 0`."** Both are the natural reaction to "Permission denied"; the first is a no-op here and the second is precisely what `bin/hive-podman-preflight-host.sh` goes out of its way not to recommend. The migration section now holds the same line the preflight does.
- **Why not remove the step instead** by exporting as the operator (`docker run --user …`): `/data` holds `gh-app-key-*.pem` at `0600` owned by the container's uid, so an unprivileged `tar` drops the one file the archive exists to carry and still writes a tarball — the same trap the host-shell-`tar` section above already documents. Step 1 stays privileged.
- **Steps 4 and 5 named as a safety property.** A `hive-data` the denied extract left empty still starts rc 0, `healthy`, dashboard answering — on a brand-new identity. That is what success looks like. Only the step 4 `hive-id` check separates the two outcomes, and step 5 is `down -v` on the source data the operator is still holding.
- **The Restore section** gets the same warning, since its step 4 offers the `:z` form for "an archive that came from Docker" — and for any archive taken under `sudo`.
- **The observed table** gains the measured labels, the denial, the post-`chown` result, and the healthy-on-empty-volume start; plus a line saying the run already recorded there was executed with a step that did not make it onto the page. Provenance of the new rows is stated (an independent run on Bluefin 44, Podman 5.8.4 rootless, Docker 29.7.2) rather than folded into the original run's numbers.

**`src/docs/troubleshooting.md`** — a symptom-first entry under the Podman failure modes, keyed on the literal `tar: can't open '/backup/…': Permission denied`, with the one-line fix and a link to the mechanism.

**`CHANGELOG.md`** — an entry under Unreleased → Fixed.

## Verification

Docs-only; no code paths touched. Anchors and relative links in both edited files were checked to resolve, using GitHub's slug rules (`#docker--podman-migration` keeps the double hyphen the dropped `→` leaves, matching the existing `#step-0--before-you-start-do-this-first` in `getting-started.md`).

The behaviour documented here is from the reporter's run in #4497: the denial, the `chown`, the identical command then succeeding, the relabel to `container_file_t`, and the migration itself completing — `hive-able-moth` and a `_migration_marker` planted on the Docker side both arriving on the Podman side. The two "this cannot work" claims the section rests on (`podman volume inspect src_hive-data`, and bind-mounting Docker's volume dir) reproduced exactly, so the export/import approach is unchanged; only the missing step is added.

Part of #4188.


clubanderson:
/lgtm
/approve

Reviewed at 02273fea. Docs-only, and the diagnosis is exactly right:

- **Mechanism sound**: `:z` relabelling is an xattr write (chcon semantics) needing ownership or CAP_FOWNER; a rootless user has neither over Docker's root-owned archive, so podman skips the file silently — no AVC, mode 0644 reads fine from the operator's shell, and the error points the wrong way. Rootful correctly exempted (identity-mapped, sudo-driven). The 65534:65534 in-container appearance (uid 0 outside the subuid range) correctly called cosmetic.
- **Fix is the right one**: single `sudo chown $(id -u):$(id -g)` at step 2a, after which the identical command relabels and extracts — measured, with the before/after labels (`user_tmp_t` → `container_file_t`) in the observed table. Provenance of the new rows honestly separated (independent run, Bluefin 44, podman 5.8.4 rootless, Docker 29.7.2) rather than folded into the original run's numbers — and the page now admits the original run used a step that never made it onto the page, which is precisely the #4497 defect.
- **Holds the preflight's line**: explicitly rejects `chmod 777` (no-op here) and `setenforce 0` (undoes what the Podman path depends on) — consistent with hive-podman-preflight-host.sh's posture and the #4353/#4499 SELinux work.
- **The alternative correctly rejected**: `docker run --user` in step 1 would drop the 0600 GitHub App key from the archive while still writing a tarball — same trap the host-shell-tar section already documents. Step 1 stays privileged.
- **Best addition**: naming the step 4/5 ordering a safety property — a denied extract leaves an empty hive-data that still starts rc 0/healthy/answering on a fresh identity, indistinguishable from success without the hive-id check. Measured (11s healthy-on-empty row).
- Restore section gets the same warning (its step 4 offers `:z` for Docker-origin archives), troubleshooting entry is symptom-first keyed on the literal error. Anchors resolve under GitHub slug rules (checked `#docker--podman-migration` and `#rootless-never-tar-the-volume-from-the-host-shell` against the real headings).
- No migration test suite exists to update — #4409 landed doc-only and this is a docs defect, so docs-only is the right shape. Cleanup contract 127/127 locally; no forbidden phrases. Trial merge against v4 @ c3c849d7 clean.

Merging.

PR #4512, opened 2026-08-21
🔒 strip every setuid bit the image ships except su-exec

Danathar:
Refs #3866. Addresses **recommendation 2** (audit and justify every SUID binary in the image) by removing them, and **recommendation 3** (document the trade-off). Recommendation 1 stays rejected for the reason #3895 established at runtime.

## What was still open

The finding's own words: *"`allowPrivilegeEscalation: true` means **any SUID binary on the system** that can be accessed by any user in the container could be exploited."* Everything shipped since has been about one of them. The C6 fix pinned and tightened `su-exec` to `4750 root:hive-launch`; #4341 moved the CA key into a `0700` directory; #4379 booted the manifest's capability set. Nobody ran recommendation 2.

I ran it. The image ships **twelve** setuid/setgid files, not one:

```
-rwsr-x--- root:hive-launch /usr/local/bin/su-exec     <- the one that was audited
-rwsr-xr-x root:root        /usr/bin/chfn
-rwsr-xr-x root:root        /usr/bin/chsh
-rwsr-xr-x root:root        /usr/bin/gpasswd
-rwsr-xr-x root:root        /usr/bin/mount
-rwsr-xr-x root:root        /usr/bin/newgrp
-rwsr-xr-x root:root        /usr/bin/passwd
-rwsr-xr-x root:root        /usr/bin/su
-rwsr-xr-x root:root        /usr/bin/umount
-rwxr-sr-x root:shadow      /usr/bin/chage
-rwxr-sr-x root:shadow      /usr/bin/expiry
-rwxr-sr-x root:shadow      /usr/sbin/unix_chkpwd
```

`node:26-slim` (Debian 13) plus the `passwd`/`util-linux` packages the runtime stage installs. Eleven binaries, **every one world-executable** — so execable by exactly the agent UIDs (2001+) that `chmod 4750 root:hive-launch` exists to exclude, sitting in the same image as the helper that was carefully gated against them.

**The pod cannot switch them off.** `allowPrivilegeEscalation: false` sets `no_new_privs`, which disables *all* setuid binaries — `su-exec` included — and breaks agent launch. That is precisely why the manifest leaves it `true`, and it means a setuid-root exec from an agent UID lands in a bounding set that still holds `SETUID`, `SETGID` and `DAC_OVERRIDE`. That is enough to read the MITM CA private key whatever mode the `0700` directory #4341 moved it into carries, and forge certificates for `api.github.com` — this issue's stated impact, reached without touching `su-exec` at all.

Nothing in Hive execs any of the eleven. The entrypoint creates per-agent accounts with `useradd`/`groupadd`/`usermod` while it is still root, and every privilege hop goes through `gosu` (as root) or `su-exec` (as `dev`). They are pure surface.

## What this changes

**`src/Dockerfile`** — strips every setuid/setgid bit except `su-exec`'s, after the last install layer, and asserts its own result. A future layer that reintroduces one fails the **build** rather than shipping. Stripping is safe precisely because only a *non-root* caller needs a setuid bit: root keeps full use of `mount`/`su`/`passwd`, and the accounts those tools would edit are locked system accounts (root's shadow entry is `*`).

Three checks keep it there, at different depths:

- **`src/scripts/check-suid-contract.sh`** — static assertions that the strip exists, exempts only `su-exec`, actually clears the bits, asserts the inventory, and **runs after the last apt/npm/pip layer**. Position, not just presence: a strip above a package install is worthless.
- **`src/deploy/test_image_suid_inventory.sh`** (new) — boots the image and reads the inventory back. It builds its fixture **twice**, once without the Dockerfile's strip block and once with it, and fails if the un-stripped build does not show the eleven. A contract check that has never been observed to fail is not evidence that it would. It then execs `su-exec` as `dev` to prove the surviving setuid bit still switches UID, and as an agent UID to prove the `hive-launch` gate still refuses.
- **`.github/workflows/docker.yml`** — runs the same assertion with `--image` against the **real** built image, the only form that also covers the npm, pip and `COPY` layers the fixture does not reproduce. The image is already local at that point (pushed builds pulled it by digest; PR builds loaded `hive-freshness-probe:local` for the freshness guard).

Every input the fixture uses — base digest, apt list, su-exec pin, launch GID, the `dev` useradd line, and the strip block itself — is **derived from `src/Dockerfile` at run time**. The script keeps no copy, so it cannot quietly test a contract the image no longer ships. Same discipline as `test_manifest_caps_runtime.sh`.

Docs (recommendation 3): the trade-off and the inventory are written down in `src/docs/security-model.md`, `src/deploy/README.md` (a new section for the three runtime contract tests, kept apart from the four that deliberately start no container), and the C6 block in `k8s/deployment.yaml`, which is where a reviewer looking at `allowPrivilegeEscalation: true` will land.

## Verification

Ran locally, against real built images — not asserted from reading:

```
-- negative control: building WITHOUT the strip block --
  ok: un-stripped base+apt layers ship 12 setuid/setgid files, so there is something to strip
  ok: 11 of them are world-executable — reachable by every agent UID, which is the finding
  ok: the fixture's su-exec matches the shipped 4750 root:hive-launch contract

-- building WITH the Dockerfile's strip block --
  ok: the strip block builds, and its in-build inventory assertion passed
     -rwsr-x--- root:hive-launch /usr/local/bin/su-exec
  ok: inventory is exactly the su-exec helper, and nothing else
  ok: none of the eleven base-image setuid/setgid binaries survive the strip

-- su-exec still performs the dev -> agent UID switch --
  ok: dev execs the surviving setuid su-exec and lands on UID 2001
  ok: an agent UID still cannot exec su-exec (/bin/sh: 1: /usr/local/bin/su-exec: Permission denied)

=== Results: 8 passed, 0 failed ===
```

Also green, unchanged by this PR: `check-suid-contract.sh` (23 checks), `test_supply_chain_pins.sh` (17), `test_standalone_service_contract.sh` (12). The capability derivation in `test_manifest_caps_runtime.sh` still parses `deployment.yaml` to the same `drop: [ALL] / add: [NET_ADMIN SETUID SETGID SETPCAP CHOWN DAC_OVERRIDE]` — my edit there is a comment above `securityContext:`, outside what it reads.

I did **not** build the full production image locally (~GB of npm/pip layers); the `--image` job in `docker.yml` is what covers those, and it runs on this PR.

## What this does not fix

The residual risk the scanner comment describes is unchanged and is not claimed here: `dev` can still exec `su-exec` and still reads the CA key, so a **dev-level** compromise can still forge `api.github.com`. That needs a separate key-holder UID with brokered signing, or #1861. What this PR removes is the path from an **agent** UID to the same outcome via a setuid binary nobody had looked at.


clubanderson:
Security review complete — **approving**.

**Verdict against the review bar:**

1. **Inventory matches reality** — the CI negative control (fixture built *without* the strip block) found exactly **12** setuid/setgid files (the 11 base/apt ones + su-exec), 11 of them world-executable — the finding reproduced, not assumed. The strip list isn't a list at all but `find -perm -4000 -o -perm -2000` with a single exemption, so it can't drift from reality; the in-build assertion pins the result to exactly `-rwsr-x--- root:hive-launch /usr/local/bin/su-exec`.
2. **su-exec exemption proven at runtime, not by build success** — the new `image-suid-inventory` job self-triggered on this PR and passed 8/8, including `su-exec 2001:1000 id -u` as `dev` → lands on UID 2001 (the actual agent-launch hop) and refusal (`Permission denied`) as an agent UID (the 4750 gate intact). The #4414-class runtime guards (`manifest-cap-runtime`, `ambient-cap-runtime`) also passed on this head, and v2-ci's docker job **built the full `src/Dockerfile`** — so the in-Dockerfile inventory assertion ran over every npm/pip/COPY layer, then the smoke test booted the result. This is the #3895 lesson honored: file-bit strip, zero capability changes (deployment.yaml gains only a comment).
3. **Nothing else relies on a stripped binary** — grepped entrypoint + scripts: account creation is `useradd`/`groupadd`/`usermod` as root (none setuid), every privilege hop is `gosu`/`su-exec`/`setpriv`; no `su`/`sudo`/`newgrp`/`mount` exec anywhere.
4. **Implementation is right** — strip runs as the LAST layer, and the static check asserts its *position* (after the final apt/npm/pip install, line 527 > 452), not just presence; **setgid (-2000) is covered**, including the setgid-shadow trio, so no follow-up gap; `-type f` correctly leaves directory setgid alone; `-xdev` is safe in a single-fs build.
5. **Reproducibility/size** — the chmod layer duplicates only the eleven small Debian binaries (a few MB); full image built + smoke-tested green on this PR.
6. **Lint** — shellcheck clean on the new 290-line test; actionlint deltas none (all SC2046 warnings pre-existing in base).

Layered defense is exactly the house pattern: build-time assertion (fails the build on reintroduction) → static contract check with position assertion (every PR) → fixture runtime job with negative control (every PR touching these files) → real-image assertion in docker.yml (every full build). Trial-merged clean against latest v4 (CHANGELOG both-add resolved by position, no conflict). No forbidden podman lifecycle literals. CI green.

/lgtm
/approve

Issue #4492, opened 2026-08-21
RFC: converse capability + Linear agents as first-class workspace members


gregoryhunt:
## Use case

Two problems, one root cause.

**1. An advisory agent cannot say anything.** In `pkg/proxy/rules.go` every issue-adjacent operation is bundled at a single tier:

```go
// ── Issue operations — ISSUES_ONLY and above ──
{`^/repos/[^/]+/[^/]+/issues$`,            "POST",  agent.ModeIssuesOnly},  // create an issue
{`^/repos/[^/]+/[^/]+/issues/\d+$`,        "PATCH", agent.ModeIssuesOnly},  // edit an issue
{`^/repos/[^/]+/[^/]+/issues/\d+/comments`,"POST",  agent.ModeIssuesOnly},  // COMMENT
{`^/repos/[^/]+/[^/]+/issues/\d+/labels`,  "POST",  agent.ModeIssuesOnly},  // relabel
```

Commenting is conversation. Creating, editing, and relabelling are artifact production. Bundling them means an ADVISORY agent that notices something on an issue thread cannot reply — it can only emit a bead the reporter never sees. And an ISSUES_ONLY agent that we wanted to let comment also gets to rewrite issue bodies and relabel. Both directions are wrong.

The same conflation exists for PRs: `POST /pulls/\d+/reviews` sits at `ModeIssuesAndPRs`, so leaving review commentary requires the rights to push branches and open PRs.

**2. Hive cannot be a first-class agent in Linear.** Linear's agent platform (https://linear.app/developers/agents) makes an app an assignable, mentionable workspace member. It requires OAuth2 with `actor=app`, scopes `app:assignable` + `app:mentionable`, `AgentSessionEvent` webhooks, and an `AgentActivity` of type `thought` emitted **within 10 seconds** of session creation.

Every one of those interactions is conversation. Under today's model there is no mode that permits it: `ModeAdvisory` forbids all writes, and anything higher grants issue mutation the agent does not need. Following #4178, hive can now *read* a Linear backlog; it cannot participate in one.

## Proposed solution

Add **`converse`** as an orthogonal capability, then build the Linear agent integration on top of it.

### Part 1 — `converse` as a capability, not a mode

[...]

PR #4515, opened 2026-08-21
✨ converse: separate talking from filing (RFC #4492, part 1)

Danathar:
Implements **Part 1** of #4492 — the `converse` axis. Part 2 (the Linear agent integration) is not here: the RFC says Part 1 stands alone and fixes a real GitHub wart with no Linear code in the tree, and that is exactly what this is.

## The problem, restated

Every issue-adjacent write sits at one tier. Commenting is bundled with creating issues, rewriting issue bodies and relabelling at `ISSUES_ONLY`; leaving a PR review is bundled with pushing branches at `ISSUES_AND_PRS`. Wrong in both directions — an ADVISORY agent that notices something on a thread cannot reply, only emit a bead the reporter never sees; and buying it a reply also hands it issue-mutation rights nobody asked for.

Conversation is not a rung between "observe" and "file issues". It is a different axis, so it is a capability, not a fifth `AgentMode` — which would renumber an enum used as an array index (`modeNames`/`modeEmojis`/`modeSuffixes`, `agentModeCount`) and force a new policy-template suffix on every agent, to model a permission that is not a trust tier.

```yaml
agents:
  scanner:
    mode: ADVISORY      # files nothing, edits nothing, relabels nothing
    converse: true      # …but can reply on a thread
```

## Design, and where it departs from the RFC

**Capabilities only ever widen.** `ProxyRule` gains an optional `Capability` predicate beside `MinMode`, and `AllowedByModeCaps` permits a matched rule when the agent holds the capability **OR** reaches the tier. Deny-by-default and the `denyRules` hard-denies are untouched — no capability reaches direct PR creation or direct merge.

**One deliberate departure, and it is the reason for the OR.** The RFC proposed replacing the tier with the capability and defaulting `converse` "on at exactly the levels where those operations are permitted today". A single flag cannot express that, because the two operations sit at *different* tiers today:

| | today | default converse on at ISSUES_ONLY | default converse on at ISSUES_AND_PRS |
|---|---|---|---|
| `POST /issues/{n}/comments` | ISSUES_ONLY | unchanged | **regression** — ISSUES_ONLY loses commenting |
| `POST /pulls/{n}/reviews` | ISSUES_AND_PRS | **widening** — granted two tiers early | unchanged |

Either default changes behaviour on existing hives, which is what the compatibility contract rules out. Keeping the floors unchanged and treating the capability as an *additional* grant delivers the contract the RFC actually asked for — every existing hive identical, `AllowedByMode` still exists and every pre-existing test calls it unmodified — while still reaching both shapes the RFC named. "Comment without issue mutation" is `ADVISORY` + `converse`, which is strictly better than what a replacement could offer, since an `ISSUES_ONLY` agent keeps issue mutation because that is what its mode *means*.

That also answers **open question 1** conservatively: off everywhere, opt-in, reversible. Questions 2–4 are all Part 2.

**GraphQL is covered, and had to be.** `gh issue comment` and `gh pr review` send GraphQL, not REST — a REST-only `converse` would look broken to anyone using the gh CLI. But the existing classifiers are substring matches, which ask *"does this mention `mergePullRequest`"*: the right question for **escalating** a requirement, the wrong one for **relaxing** it, since a document containing `addComment` **and** `issueUpdate` would match "mentions addComment". So converse is granted only when **every top-level field of the document's mutation is conversational**, via a small selection-set parser that returns nothing — no grant, fall through to the unchanged tier check — on anything it cannot read confidently: multiple operations, unbalanced braces, an unresolvable alias. Merge and PR-write classification still runs first, so a batched comment-and-merge is still a merge.

## Plumbing

- `/tmp/.hive-caps-<agent>`, deliberately a **second** file. `bin/gh-wrapper.sh` reads `/tmp/.hive-mode-<agent>` under `set -e` and expects a bare mode string; changing that content would break every `gh` call on every hive. Written for *every* agent including those with no capabilities, so clearing `converse` actually revokes rather than leaving a stale granting file. Every read failure path yields no capabilities. Repaired by the permissions watcher alongside the mode files.
- `Converse *bool` in `AgentConfig` — a pointer so "unset" stays distinguishable from "explicitly false" across pack re-apply (the reason `ModelOwner` exists), and `omitempty` so an agent that never sets it serializes byte-identically.
- A level change clears `Mode` but deliberately leaves `Converse`, which is level-independent; commented at the site so it does not get "fixed".

## Testing

Both halves, because either alone is misleading — half two without half one lets a widening pass as "the feature works":

1. **Compatibility.** For every mode × every route, a zero capability set answers exactly what the old tier-only comparison answered. The pre-existing `rules_test.go` still calls `AllowedByMode`/`GraphQLAllowed` unchanged and passes, which is the real regression evidence.
2. **Scope.** `converse` opens exactly the two conversational routes; issue creation, editing, relabelling, PR edits, refs, commits, pushes, `update-branch`, and both hard denies all stay shut at the same tier. Asserted per-route, plus a blanket "this capability never narrows anything" check.

GraphQL cases include the batched documents that must still be refused (comment + issue edit, comment + merge, review + `createPullRequest`), a comment name buried in a string argument, and `updateIssueComment` — editing someone else's comment is not conversation.

Ran in a `golang:1.25-alpine` container (no Go toolchain on this host):

```
ok   github.com/kubestellar/hive/pkg/proxy   43.6s     # full suite, including all 40+ new subtests
go build ./...   BUILD_OK
go vet ./pkg/proxy ./pkg/agent ./pkg/config ./pkg/dashboard   VET_OK
```

`pkg/agent` (7), `pkg/config` (4) and `pkg/dashboard` (25) have failures that reproduce **identically on a pristine `upstream/v4`** with my changes stashed — root-in-container and entrypoint-script artifacts of this environment, not regressions. I verified the counts on both trees rather than assuming.

## Not in this PR

- **Dashboard surface** for the toggle — operators set `converse` in `hive.yaml`. The RFC scopes the dashboard to Part 2 item G.
- **Everything else in Part 2**, including `NeedsMITM` for `api.linear.app`. Worth noting that the RFC's item F ("Part 1 must land first") is now unblocked: the mapping it wanted to express — `agentActivityCreate` needs `converse`, `issueUpdate` needs `ISSUES_ONLY` — is expressible, and `topLevelMutationFields` here is a starting point for the GraphQL operation inspection Linear will need, since Linear is a single `POST /graphql` for everything.

Closes nothing on its own; #4492 stays open for Part 2.


clubanderson:
Reviewed against RFC #4492 part-1 scope from head 7f834e7.

**Scope & design**
- Matches RFC part 1 exactly: `converse` capability opens precisely two conversational surfaces — REST `POST .../issues/{n}/comments` and `POST .../pulls/{n}/reviews` — plus the equivalent whole-document GraphQL mutations. No Linear/part-2 code. RFC #4492 correctly left open.
- The documented departure (default OFF everywhere instead of "on at current tiers") is the right call: the two ops sit at different tiers, so any "current behavior" default would change one of them. OR-widening beside unchanged tier floors preserves the compatibility contract; verified the blanket "capability never narrows" test.
- Hard denies (PR create, merge) checked before capability rules — unreachable by converse. GraphQL grant requires *every* top-level mutation field to be conversational; the strict `topLevelMutationFields` parser fails to nil (deny) on multi-op docs, unbalanced braces, unresolvable aliases. Batched comment+merge/comment+issueUpdate refusal tests pin this.
- Fail-closed everywhere: caps file read errors ⇒ zero caps; unknown tokens ignored (deny direction); file written even when empty so clearing actually revokes; permissions watcher repairs caps files alongside mode files.
- Separate caps file (not mode-file mutation) is correct — `bin/gh-wrapper.sh` reads the mode file under `set -e` expecting a bare mode string. Confirmed the wrapper's ADVISORY arm only blocks issue/PR *create* and *merge*, so `gh issue comment`/`gh pr review` reach the proxy and converse is enforceable there — no wrapper change needed.
- `Converse *bool` + omitempty keeps unset agents byte-identical on save; pack level change deliberately preserves converse (level-independent), commented at the site.

**Backward compat:** legacy `AllowedByMode`/`GraphQLAllowed` delegate with zero caps; pre-existing `rules_test.go` unchanged and passing is the regression evidence. `proxyHTTP` signature change mechanically applied across 9 test files.

**Local verification (macOS, this head):** `go build ./...` + vet clean; `pkg/proxy` ok 92.3% cov; `pkg/agent` ok 91.0% (floor 89); `pkg/dashboard` ok 83.3% (floor 78). `pkg/config` has two failures (`TestExportEmitsShellExportOnBothPaths`, `TestSharedCacheIsCreatedPrivate`) that reproduce identically on pristine origin/v4 — environmental, not this PR. gofmt findings in 3 proxy test files are pre-existing on v4. Trial-merge against latest v4 (94271887) clean, CHANGELOG auto-merged. CI green minus tide. No forbidden podman lifecycle phrases.

/lgtm
/approve

PR #4517, opened 2026-08-21
🔒 mint: verify the caller's Kubernetes ServiceAccount, not just a secret

Danathar:
Closes the last open gap in #3915.

## What was left

#4436 landed three of the four pieces: the `CallerAuthenticator` seam, per-identity `Entitlements`, and per-caller audit. What it could not land was anything to put *behind* the seam. The only implementation was still the shared secret, so `shared-secret:any-holder` remained the only name an entitlement could be keyed on — an entitlement map with exactly one possible key is not an identity model.

The reason it was deferred is stated in `caller.go`:

> It requires `k8s.io/client-go`, which this module does not depend on today, and adding a dependency of that weight is a maintainer call rather than something to slip into a security fix.

**That blocker does not exist.** TokenReview is one POST of a small, stable JSON object to `authentication.k8s.io/v1` — GA since Kubernetes 1.6 — so `net/http` and `encoding/json` cover it. **`go.mod` and `go.sum` are untouched by this PR.** If client-go is ever added for other reasons, `tokenreview.go` swaps out behind the same interface with no change to `Server`; that is what the seam was for. The maintainer decision is worth having on its own merits, but it is no longer standing between the finding and its fix.

## What this adds

`TokenReviewAuthenticator` — the caller presents its projected ServiceAccount token, the mint asks the API server to vouch for it, and `system:serviceaccount:<ns>:<name>` becomes the identity that entitlements key on and the audit line records.

Four properties are load-bearing. I'm listing them as the ways this could *look* like it verifies identity while not doing so, because that framing is what the tests are organised around:

| | Why it matters |
|---|---|
| **Audience checked on the response, not the request** | An API server whose authenticators don't implement audience validation answers `authenticated: true` with the `audiences` field **absent**. A backend reading only `authenticated` would accept the API-server token every pod already has mounted at a well-known path, and audience scoping would be decorative. Empty or absent is a refusal. |
| **A dedicated header**, `X-Hive-Mint-SA-Token` | In dual-accept, `Authorization` carries the **shared secret**. Reviewing that header would POST the mint's own secret to the API server as "a token to please review" — writing it into someone else's audit log. |
| **Real TLS** against the in-cluster CA | No insecure-skip-verify option, not even a documented one. |
| **Fail closed** | Timeout, response-size cap, refusal on every infrastructure failure — unreachable API server, missing RBAC, unparseable response. A TokenReview that failed *open* would be worse than the secret it replaces, because operators would believe identity was being checked. |

Only `system:serviceaccount:` usernames are accepted — an entitlement map keyed on ServiceAccount names must not be satisfiable by a human with a kubeconfig. The mint's own reviewer token is re-read per call, since projected tokens rotate on disk and one cached at startup stops working an hour into the pod's life.

`MultiAuthenticator` is the rollout step the finding's plan calls for: run both mechanisms, watch the audit log until no `shared-secret:any-holder` lines remain, then drop the secret. TokenReview goes **first** so a caller presenting both is recorded under its real identity — otherwise the log shows nothing changing and the migration can't be observed to finish.

**Default posture is unchanged.** A `Server` that configures nothing still gets the shared-secret gate, byte-identical.

## Verification

Run in a `golang:1.25-alpine` container (no Go toolchain on this host):

```
ok  github.com/kubestellar/hive/pkg/mint  coverage: 90.8% of statements
go build ./...  BUILD_OK      go vet ./pkg/mint  VET_OK      gofmt: clean
```

30 new tests against an `httptest` stand-in for the API server that records what the mint actually sent, so the assertions are on the request and not only the outcome. Each of the four properties above has a test that fails if it regresses — notably the unvalidated-audience replay (four variants: absent, empty, different audience, prefix-collision) and an explicit guard that the caller's `Authorization` header is never forwarded as the token to review nor replayed as the mint's own credential. Plus a fail-closed group covering RBAC refusal, unreachable API server, empty reviewer token, and a garbage response body.

The end-to-end test is the payoff: an `Entitlements` map keyed on a real ServiceAccount name, enforced deny-by-default, returning 403 (authenticated but not permitted) rather than 401. Under the shared secret that map could not be written down at all.

## Scope, stated plainly

- **No config wiring, deliberately.** `mint.NewServer` has no production caller — nothing in this repo serves `Handler()` today; the mint is used in-process via `AgentMinter`. Adding `mint.caller_auth.*` config for an endpoint no binary serves would be inventing a surface, so the backend lands at the package level and `README.md` documents how to construct it.
- **The short-term hardening from the finding** (bind to localhost / pod-internal) has nothing to bind today for the same reason. `README.md` says so, and says that whoever wires a listener owns it — caller auth is a second line, not a substitute.
- **mTLS backend** for deployments with no API server to ask: not here. Same interface when it is; the `TODO(caller-auth)` in `server.go` is narrowed to just that.

`pkg/mint/README.md` also carries what an operator needs to actually turn this on: the `create` on `tokenreviews` RBAC the mint needs, and the audience-scoped projected volume each caller needs (the default mounted token is for the API server and the audience check will correctly refuse it).


clubanderson:
Security-priority review from head 5f4d21b0, as the direct extension of #4436's seam.

**Security verdict: sound.**
- **Real verification, not JWT decoding**: caller token is POSTed to `authentication.k8s.io/v1` TokenReview and the API server's verdict is authoritative — expiry, revocation, and deleted-SA cases all collapse into `authenticated:false`, which is refused (tested).
- **Audience binding is checked on the RESPONSE, not the request** — the critical anti-replay property. A cluster that doesn't validate audiences returns `authenticated:true` with audiences absent; this backend refuses absent/empty/mismatched lists, so the default API-server token every pod mounts cannot be replayed at the mint. Pinned by the 4-case `TestTokenReviewRefusesUnvalidatedAudience` plus a prefix-collision case.
- **Identity binding**: only `system:serviceaccount:<ns>:<name>` usernames accepted (humans/nodes/anonymous/malformed all refused, 8 cases); that exact name keys `Entitlements` and appears in audit — end-to-end proven by `TestServerMintsUnderServiceAccountEntitlement` (200 entitled / 403 unentitled subject+scope / 401 secret-only against a TokenReview-only server).
- **No secret leakage**: caller token travels in a dedicated `X-Hive-Mint-SA-Token` header and only in the review body; `Authorization` (shared secret in dual-accept) is never forwarded as a token to review nor replayed as reviewer credential — both asserted against a recording fake API server. Caller-facing failures are uniform 401s (no oracle); the RBAC-missing case is distinguished in the server-side log line per the post-#4446 honesty standard.
- **Fail closed on every infrastructure path**: API server down, mint's own RBAC missing (403), unreadable/empty reviewer token, garbage response — all refuse, all tested. 5s timeout + 1MiB response cap; TLS pinned to the in-cluster CA (or host store), MinVersion 1.2, no skip-verify option; transport has no proxy func so reviews can't be routed through the MITM proxy.
- **No escalation**: entitlements remain deny-by-default per #4436; empty dimension allows nothing; wildcard must be explicit. `go.mod`/`go.sum` untouched — stdlib only, seam unchanged.
- **Non-K8s deployments unaffected**: default posture is still the shared-secret gate byte-identical; `NewInClusterTokenReviewAuthenticator` fails loudly outside a cluster rather than degrading (tested), and `MultiAuthenticator` (TokenReview first, so migration is observable in audit) is opt-in. No manifests ship because nothing in-tree serves the handler yet (verified at #4436: zero production call sites); README carries the exact ClusterRole + projected-token snippets whoever wires a listener will need.
- Reviewer token re-read per call — survives projected-token rotation.

**Local verification (this head):** `go build ./...`, `go vet ./pkg/mint`, gofmt clean; `go test -cover ./pkg/mint` ok, **91.1%** coverage. Trial-merge against latest v4 clean (CHANGELOG auto-merged). CI green minus tide. No forbidden podman lifecycle phrases.

With TokenReview + entitlements + per-caller audit landed, #3915's recommendation is implemented (mTLS was the alternative, left as a documented TODO for API-server-less deployments) — closing it after merge.

/lgtm
/approve

PR #4518, opened 2026-08-21
fix(agent): pin caveman by tag (v1.9.1), not raw SHA — SHA breaks the skills CLI clone

hanthor:
## Problem

Every agent kick on a live hive logs a failed caveman install:

```
◇  Source: https://github.com/JuliusBrussee/caveman.git @ 0d95a81d35a9
■  Failed to clone repository
│  Failed to clone https://github.com/JuliusBrussee/caveman.git: ...
│  warning: Could not find remote branch 0d95a81d35a9 to clone.
│  fatal: Remote branch 0d95a81d35a9 not found in upstream origin
```

The `skills` CLI clones with `git clone --depth 1 --branch <ref>`, and git cannot clone a bare commit SHA as a branch — only branch/tag names work. The hive pins `github:JuliusBrussee/caveman#0d95a81d35a9` (a SHA), so the install fails for every backend that goes through `skills add` (goose/codex/aider) and the skill never lands.

## Fix

Pin by **tag** instead: `#v1.9.1`. `v1.9.1` resolves to exactly commit `0d95a81d35a9` (checked via the repo's tag listing), so this is a pure mechanical change — identical skill content, now cloneable. Tags clone cleanly via `--branch`.

Applies to all four install paths (npx direct for claude/copilot/gemini, `skills add` for goose/codex/aider).

## Verified

- `npx skills add JuliusBrussee/caveman#v1.9.1` completes (`Done!`) and records `ref=v1.9.1` in `skills-lock.json`; the `#0d95a81d35a9` form fails with the clone error above.
- `go build ./pkg/agent/` passes.

## Notes

Keep bumping the **tag** deliberately (per the existing comment about the 2026-07-27 `--mode` breakage); a raw SHA pin will silently break again if anyone re-introduces one.

(Re-submitted from a fresh fork after the original fork was retired.)

### DIFF
--- a/src/pkg/agent/manager.go
+++ b/src/pkg/agent/manager.go
@@ -2098,7 +2098,12 @@ func (m *Manager) installCavemanForAgent(agent *AgentProcess, backend string) {
 
 	// Pinned: unpinned HEAD broke every install on 2026-07-27 when upstream
 	// removed the --mode flag. Bump deliberately, after checking `--help`.
-	const cavemanRef = "github:JuliusBrussee/caveman#0d95a81d35a9"
+	// v1.9.1 IS commit 0d95a81d35a9 — the SHA form is deliberately not used
+	// because the `skills` CLI clones with `git clone --branch <ref>`, and
+	// git cannot clone a bare SHA as a branch ("Could not find remote branch
+	// 0d95a81d35a9"), which broke every install on a live hive. Tags clone
+	// cleanly; bump the tag, never a raw SHA.
+	const cavemanRef = "github:JuliusBrussee/caveman#v1.9.1"
 	// Upstream replaced `--mode full|minimal` with `--all` / `--minimal`
 	// (--all = hooks + init).
 	modeFlag := "--all"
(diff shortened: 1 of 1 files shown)


clubanderson:
Reviewed from head e33abf3a.

- **Tag verified independently**: `git ls-remote https://github.com/JuliusBrussee/caveman.git` shows annotated tag `v1.9.1` (object 033f9186) dereferencing to `0d95a81d35a9f2d123a5e9430d1cfc43d55f1bb0` — exactly the commit previously pinned. Identical content, now cloneable; not a version bump in disguise.
- Root cause is right: `skills` CLI clones `--depth 1 --branch <ref>`, and git refuses a bare SHA as `--branch` — so every goose/codex/aider kick logged a failed install. Tags clone cleanly.
- All four install sites converted (npx const + three `skills add` literals); the only remaining `0d95a81d35a9` occurrences in manager.go are in the explanatory comment. Grep confirms manager.go is the sole pin site in Go sources. Install command shape (`--all` modeFlag, backend routing) untouched — no interaction with the #4532 caveman_mode enum work.
- Comment preserves and strengthens the deliberate-bump discipline (2026-07-27 lesson): bump the tag, never a raw SHA. #4529's regression guard will pin this mechanically. Residual note: tags are mutable where SHAs aren't — the annotated-tag verification above is the compensating control, and the follow-up guard keeps the pin greppable.
- DCO sign-off present. `go build ./pkg/agent` + vet clean locally; trial-merge against latest v4 clean. CI green minus tide.

Thanks @hanthor — this kills a failure logged on literally every agent kick.

/lgtm
/approve

Issue #4490, opened 2026-08-21
🐛 SELinux release qualification reports FAIL on a healthy enforcing host — it uses `stat -c %C`, the reader its own siblings document as unreliable


Danathar:
`src/deploy/qualify_podman_selinux.sh` reports **`FAIL (2)`** on a correctly-functioning SELinux-enforcing host, and emits a release-qualification ledger row saying so. The SELinux behaviour is fine; the script's label reader is not.

```
--- :z (shared relabel) ---
  label after :z  unsupported for this operating system
  mode  after :z  600
  PASS: :z grants the container access
  FAIL: :z relabelled to container_file_t with no private category
        got unsupported for this operating system

=== result block (paste into src/docs/podman-selinux-release-qualification.md) ===
| <release> | 2026-08-21 | Bluefin (Version: 44.20260818) | targeted | 7.1.4-200.fc44.x86_64 | podman 5.8.4, rootless=true | FAIL (2) |

SUMMARY: 2 failure(s)
```

`unsupported for this operating system` is not a label. It is what **uutils coreutils** `stat` prints when asked for `%C`.

## The mechanism

```
$ command -v stat
/home/linuxbrew/.linuxbrew/opt/uutils-coreutils/libexec/uubin/stat
$ stat --version | head -1
stat (uutils coreutils) 0.9.0

$ OUT=$(stat -c '%C' ~/.config/hive/hive.yaml 2>/dev/null); echo "[$OUT] rc=$?"
[unsupported for this operating system] rc=0
```

It writes the error to **stdout** and **exits 0**, so nothing detects it. The label is read as that string and compared against `container_file_t`, which does not match, and the suite calls it a failure.

The label was correct the whole time:

```
$ /usr/bin/stat -c '%C' ~/.config/hive/hive.yaml
system_u:object_r:container_file_t:s0:c297,c987
$ getfattr -n security.selinux --only-values --absolute-names ~/.config/hive/hive.yaml
system_u:object_r:container_file_t:s0:c297,c987
```

The qualification has no fallback — `label_of()` at line 156 is bare `stat -c '%C'`, as is `START_LABEL` at line 138.

## This is a known problem that was fixed everywhere else

Two sibling scripts already handle it.

`bin/hive-podman-preflight-host.sh` (#4209) carries a multi-reader list and an explicit warning:

```
# `stat -c '%C'` is not a reliable reader. Under uutils coreutils — the default
[...]

PR #4526, opened 2026-08-21
🐛 hive.container: pull the gateway up, not just down (#4516)

### LINKED ISSUE #4516: 🐛 Unattended `podman auto-update --rollback` still strands the gateway — the one #4493 path #4503 cannot reach
#4493 is closed and #4503 is merged, and the fix is correct for every path that runs through `bin/hive-podman-update.sh`. One path does not, and it is the unattended one: **`podman auto-update --rollback`, driven by `podman-auto-update.timer`.**

Podman restarts the unit **over D-Bus itself** on rollback. It never invokes `hive-podman-update.sh`, so none of #4503's `GATEWAY_UNIT` handling runs. Verified on merged `v4` (`6958611`):

- `src/deploy/quadlet/hive-gateway.container` still has only `Requires=hive.service` + `After=hive.service` — stop-direction coupling only. No `BindsTo=`, `PartOf=` or `Upholds=`.
- `hive.container` has no start-direction coupling to the gateway.
- `src/deploy/quadlet/optional/hive-autoupdate.conf` is `AutoUpdate=registry` and nothing else.

## Measured

Shipped 300s `TimeoutStartSec`, real registry, bad image pushed over a floating tag:

```
  UNIT          CONTAINER            IMAGE                     POLICY      UPDATED
  hive.service  b8363439d6a6 (hive)  localhost:5000/hive:auto  registry    rolled back
auto-update rc=0    (313s wall clock)

AFTER: hive=active  gw=inactive
       running=45ad3d76…      <-- good digest restored, unaided
       NRestarts=0
       dash=DEAD
```

The rollback works exactly as `podman-auto-update.md` documents. The gateway journal shows it stopped once and was never started again:

```
16:28:33 Stopping hive-gateway.service - Hive authenticating gateway...
16:28:33 Stopped  hive-gateway.service - Hive authenticating gateway.
                          (no further entries — no Starting, ever)
```

Hive healthy on `:3002` inside the container. `:3001` dead. Nobody watching.

For contrast, a **successful** auto-update leaves the gateway up — it is active when the restart job runs, so the transaction cycles it. Only the rollback case strands it, because the failed start already took the gateway down.

## Why this is the worst instance of the pattern

`hive-autoupdate.conf` is unusually good about stating what enabling it costs — four numbered items covering the downtime, that podman does not remember a rollback, that `is-failed`/`ActiveState`/`NRestarts` are not the alarm, and the digest-pin interaction. An operator who reads it knows what they are signing up for.

Cost 5 is missing: **after a rollback the dashboard stays down until a human notices.** Not one `TimeoutStartSec` — indefinitely. And item 3 makes it worse by being correct: the doc has already told the operator that unit state is not the alarm, so the signals that would reveal this are the ones they have been told to ignore.

Combined with item 2 (podman does not remember, so a bad published tag repeats the outage every timer firing, daily by default), the steady state is a deployment that is down, comes back Hive-only each cycle, and reports `rc=0` throughout.

## Suggested fix

Couple the gateway to hive in the **start** direction, on the unit, rather than in each caller:

- `BindsTo=hive.service` + `After=hive.service` on `hive-gateway.container`, or
- `Upholds=hive-gateway.service` from `hive.container` on systemd ≥ 249

That covers auto-update, a bare `systemctl stop`/`start` by hand, and anything future that cycles `hive.service`, in one place. #4503's script-side additions then become belt-and-braces rather than the only guard, and #4501's `exercise` repair likewise.

If a unit change is not wanted, the minimum is documenting it as cost 5 in `hive-autoupdate.conf` and in `podman-auto-update.md`, whose cost table currently says a bad update means "one full `TimeoutStartSec` of downtime, then automatic recovery" — true of Hive, false of the dashboard.

## Environment

Bluefin 44 (Silverblue), SELinux enforcing, Podman 5.8.4, rootless, shipped `hive.container` with `TimeoutStartSec=300` unmodified. Local registry used only to obtain a moving tag; the timeout, the unit and the rollback are all shipped configuration. Original measurement and the successful-update contrast: https://github.com/kubestellar/hive/issues/4493#issuecomment-5375105002

Part of #4188

Danathar:
Closes #4516.

`hive-gateway.service` carries `Requires=hive.service`, which propagates a **stop** only. A failed update stops the gateway along with the failing unit, and `systemctl restart hive.service` re-starts only those dependents that are *active* when the job runs — which the gateway, already stopped, is not.

#4503 fixed every path routed through `bin/hive-podman-update.sh` and I verified it works. The unattended path is not routed through it: **`podman auto-update --rollback` restarts the unit over D-Bus and never runs Hive's scripts**, so no caller-side fix can reach it. The coupling has to live on the unit.

## Measured

Rootless deployment, shipped `TimeoutStartSec=300`, a real registry, a bad image pushed over a floating tag:

```
  UNIT          CONTAINER            IMAGE                     POLICY      UPDATED
  hive.service  b8363439d6a6 (hive)  localhost:5000/hive:auto  registry    rolled back
auto-update rc=0    (313s wall clock)

AFTER: hive=active  gw=inactive
       running=45ad3d76…          <-- correct digest restored, unaided
       NRestarts=0
       dash=DEAD
```

The rollback is correct in every respect the docs claim. The gateway journal shows it stopped once and was never started again:

```
16:28:33 Stopping hive-gateway.service - Hive authenticating gateway...
16:28:33 Stopped  hive-gateway.service - Hive authenticating gateway.
                          (no further entries — no Starting, ever)
```

Hive healthy on `:3002` inside the container; `:3001`, the only published port, dead — on a daily timer, with `rc=0`, and with `hive.service` reading `active` throughout.

## The change

One line on `hive.container`:

```ini
Wants=hive-gateway.service
```

**`Wants=` rather than `Upholds=`**, deliberately:

- `Upholds=` would also restart a gateway an operator **deliberately** stopped. That is a legitimate thing to do and not ours to undo — it would trade one surprising behaviour for another that is *harder* to notice, because it looks like the system working.
- `Upholds=` needs systemd 249+; `Wants=` has no floor. ADR-0017 records "measured on systemd 259" but sets no minimum, so not introducing one seemed right.
- `BindsTo=` and `PartOf=` were considered and rejected: both are stop-direction, so neither fixes this.

Ordering is unchanged — `hive-gateway.container`'s own `After=hive.service` still sequences them, and adding a reverse ordering key here would risk a cycle. A test asserts that nobody adds one.

## Verification

Both directions, on the shipped units installed through `bin/hive-podman-setup.sh` (not a hand-edited copy — `Wants=` confirmed present in the *generated* unit, since Quadlet only passes through `[Unit]` keys it recognises):

```
=== #4516 repro against the PATCHED units ===
  before: hive=active gw=inactive
  after:  hive=active gw=active
  dashboard: {"status":"ok"}

=== deliberate-stop case must still be respected ===
  gw after deliberate stop: inactive
```

Lifecycle unaffected: no ordering cycle, `systemctl stop hive.service` still takes the gateway down via `Requires=`, and the gateway still starts *after* hive (measured 11s later, not racing it).

Suites:

```
PASS: 14 Quadlet config-coupling assertions (health ports 3002 3001)
=== Results: 18 passed, 0 failed ===        (port boundary)
SUMMARY: 0 failure(s)                       (generator gate, system + user)
PASS: 31 standalone image-reference assertions across 3 asset(s)
```

## Tests

Three assertions in `src/deploy/test_quadlet_config_contract.sh`, **each verified to fail when the property it guards is removed** rather than merely passing today:

| removed | assertion fires |
| --- | --- |
| the `Wants=` line | "no 'Wants=hive-gateway.service'…" |
| add a reverse `After=hive-gateway.service` | "orders itself against hive-gateway.service…" |
| the gateway's `Requires=hive.service` | "lost Requires=/After=hive.service…" |

The third exists because `Wants=` only pulls the gateway **up**; if someone later removes `Requires=`, stopping Hive would leave an orphaned gateway serving 502s.

## Docs

`src/docs/podman-auto-update.md` claimed a bad update costs one `TimeoutStartSec` "then automatic recovery". That was true of Hive and false of the deployment; corrected, with the measurement.

`hive-autoupdate.conf` gains a **cost 5**. Its existing cost list is unusually good, and item 3 already tells operators that `is-failed` / `ActiveState` / `NRestarts` are not the alarm — which is correct, and which means the signals that would have revealed a stranded dashboard are the ones they have been told to ignore. Worth stating plainly next to the others.

## Provenance

Found during a full end-to-end pass of #4188 on Bluefin 44 (Silverblue), SELinux enforcing, Podman 5.8.4, systemd 259, rootless. That pass also re-verified the nine issues fixed since the first round — #4484, #4485, #4486, #4488, #4489, #4490, #4491, #4493, #4497 — all holding, plus #4478's boot gate confirmed on hardware (a never-healthy Hive now costs 0.17s of boot, against a 300s hold before the fix). This was the one finding still open.


clubanderson:
/lgtm
/approve

Reviewed at b8794f01. Root-cause fix for the class behind #4491/#4493, and the semantics choice is exactly right:

- **Wants= is the correct key**: start-direction pull at job time, respects a deliberate operator stop (which Upholds= would silently overrule — the harder-to-notice surprise), no systemd version floor (Upholds needs 249+ and ADR-0017 sets no minimum). BindsTo=/PartOf= correctly rejected as stop-direction. Ordering untouched — the gateway's own After=hive.service still owns it, and the reverse-ordering cycle risk is guarded by a test rather than a comment.
- **#4514 boot gate unaffected**: verified against the shipped units — the gate remains the only Hive unit in default.target, all four units stay WantedBy=hive-boot.target, and Wants=hive-gateway.service just adds the gateway to whatever post-boot transaction starts hive. No boot-transaction re-entry; a never-healthy Hive still costs itself, not the boot.
- **Rollback semantics intact (#4407)**: Notify=healthy, timeouts, Restart=always, and the D-Bus start-job trigger all untouched — this is the unit-level coupling the unattended path needed, since `podman auto-update --rollback` restarts over D-Bus and never runs Hive's scripts (no caller-side fix can reach it; the measured evidence with the stranded-gateway journal makes that airtight).
- **#4501/#4503 stay as defense in depth**: ensure_gateway_serving and the probe restore path are untouched (verified in bin/hive-podman-update.sh) — correct, since they also verify :3001 *answers*, which Wants= alone does not.
- **No flap loops**: Wants= doesn't propagate gateway failure back to hive; a gateway crash restarts via its own Restart=always; hive start on an active gateway is a no-op job.
- **Tests honest**: the three new contract assertions each verified-to-fail on mutation — I independently re-ran the Wants=-removal mutation here and it fires. The third (gateway keeps Requires=/After=) is the right guard against someone later orphaning the stop direction.
- **Docs corrected precisely**: the auto-update page's 'then automatic recovery' claim was true of Hive and false of the deployment — now says so with the measurement; hive-autoupdate.conf cost 5 correctly notes that item 3's own advice (ignore unit state) is what masked this shape of failure.
- Suites on the v4-merged tree: config contract 14 (3 new), parity 33/0/7, port boundary 18, image refs 31, service contract 12, orphan guard 17/0, cleanup contract 127 — all pass. Update suite 93/0 with the GNU-install shim (5 local failures are the known BSD `install -D` limitation; CI ubuntu authoritative). shellcheck clean; no forbidden phrases. Generator pass-through of Wants= confirmed by his setup.sh-installed verification and CI's quadlet gate.

Merging.

Issue #4485, opened 2026-08-21
🐛 hive-data is created without its ownership labels, so teardown silently leaves all Hive state (including audit.jsonl) on disk


Danathar:
`hive-data` ends up with **no ownership labels at all**, which makes it permanently invisible to the #4210 selection contract. Teardown then reports `no Hive-owned volumes` and leaves every byte of Hive state on disk — including `audit.jsonl`, the agent audit record — while reporting that it removed the deployment.

This is the same underlying class as #4484 (a Quadlet unit left `active` while the resource it owns is gone), but the volume path fails **silently**: #4484 is a loud `network not found`, this one produces a healthy-looking stack whose state is orphaned.

## Observed

The volume was created by *this* install, at 15:33:40, and has no labels:

```
$ podman volume inspect hive-data --format '{{.CreatedAt}} {{.Labels}}'
2026-08-21 15:33:40.123295475 -0400 EDT map[]

$ podman volume ls --filter label=io.kubestellar.hive.owned=true --format '{{.Name}}'
                                          <-- empty
```

The generator is **not** at fault; the generated unit passes all four labels:

```
ExecStart=/usr/bin/podman volume create --ignore \
  --label io.kubestellar.hive.component=hive --label io.kubestellar.hive.instance=default \
  --label io.kubestellar.hive.owned=true --label io.kubestellar.hive.runtime=podman hive-data
```

The unit simply never ran. `hive.service` correctly carries `Requires=`/`After=hive-data-volume.service`, but that unit was **already `active (exited)` from a previous boot**, so systemd had nothing to do:

```
$ systemctl --user show hive-data-volume.service -p ActiveState -p ActiveEnterTimestamp
ActiveState=active
ActiveEnterTimestamp=Fri 2026-08-21 14:19:36 EDT     <-- stale; volume had since been removed
```

The journal for the start at 15:33:40 shows the volume unit is skipped entirely, and the volume being created by `hive.service`'s own `podman run` instead:

```
15:33:40 Starting hive-network.service
15:33:40 Finished hive-network.service
15:33:40 Starting hive.service                       <-- no `Starting hive-data-volume.service`
15:33:40 podman[128925]: volume create hive-data     <-- created by `podman run -v hive-data:/data`
```

`podman run -v NAME:/path` auto-creates a missing named volume, with no labels and no warning. That is the whole mechanism.

## Why it never recovers

`--ignore` returns success on an existing volume **without applying labels**, so no later start repairs it:

```
[...]

PR #4534, opened 2026-08-22
📖 README: point the Podman quick start at the installer (#4533)

### LINKED ISSUE #4533: 📖 README's Podman quick start is 43 manual lines and never mentions bin/hive-podman-setup.sh
`bin/hive-podman-setup.sh` installs a standalone Podman deployment in one command. The README — the entry point a self-hosting operator actually reads — **never mentions it**:

```
$ grep -c hive-podman-setup README.md
0
```

It is documented in five other files (`CHANGELOG.md`, `podman-standalone-quadlet.md`, `podman-ownership-cleanup.md`, `backup-restore.md`, `podman-quadlet-update-rollback.md`), so it is not obscure — it is just absent from the one page someone lands on first.

## The asymmetry, measured

| | executable lines in the README quick start |
| --- | --- |
| Docker Compose | **11** |
| Podman | **43** |

A reader comparing the two runtimes on that page sees an 11-line path and a 43-line path, with no indication the second has a one-command equivalent. That is a fair reason to pick Docker, arrived at from incomplete information.

Worth noting `bin/hive-setup.sh` is not in the README either, so this is not Podman being singled out. But the Docker path does not need its installer to look reasonable — `docker compose up -d` is already the short version. The Podman path is where the omission changes the impression.

The CHANGELOG entry for #4470 states the intent plainly: the script exists *"closing the asymmetry with Docker's `bin/hive-setup.sh`"*. That asymmetry is closed in `bin/`, and still open in the README.

## The manual path also gives up the installer's guards

Both paths work — I walked all 43 lines end to end this week on an SELinux-enforcing host and the result was healthy, correctly labelled, and passed the port-boundary checks. So this is not a correctness bug. But the installer does things the manual sequence leaves to the operator:

- **reads `dashboard.port` out of the unit that will enforce it** and refuses to continue if the config disagrees — the #4367 trap the README instead warns about in a comment ("Keeping 3001 costs a silent 300-second hang")
- **runs `hive-data-volume.service` explicitly** rather than letting `hive.service` auto-create the volume unlabelled (#4485). The manual path gets this right *on a clean install* through dependency ordering; it has no guard when a stale unit makes that ordering a no-op, which is exactly how #4485 was originally hit
- **applies the right secrets ownership for the root mode** (`podman unshare chown` vs `chgrp`, #4359)
- **checks lingering and says the install will not survive a reboot** (#4489, step 8/8)

The README now covers the lingering case well — it added a post-install `lifecycle-probe.sh check` step with the reasoning — so that one is genuinely handled. The others are only handled if you use the script nobody told you about.

## Suggested fix

Small: a short block at the top of **Quick Start (Podman)**, before the manual sequence, along the lines of

```bash
# One command, if you want it: preflights, config, units, boot wiring, and a
# health check that confirms the gateway answers before it returns.
bin/hive-podman-setup.sh --rootless     # or --rootful
```

with the existing 43-line sequence kept below as "or, by hand" — it is genuinely useful as the explanation of what the installer does, and several of its comments are the best documentation of the traps that exist anywhere.

Adding `bin/hive-setup.sh` to the Docker section at the same time would keep the two symmetrical, though that is a separate call.

## Environment

Found during the #4188 end-to-end passes on Bluefin 44 (Silverblue), SELinux enforcing, Podman 5.8.4, `v4` at `325707ed`. Both paths verified working; this is about discoverability, not correctness.

Part of #4188

Danathar:
Closes #4533.

`bin/hive-podman-setup.sh` installs a standalone Podman deployment in one command. It is documented in `CHANGELOG.md`, `podman-standalone-quadlet.md`, `podman-ownership-cleanup.md`, `backup-restore.md` and `podman-quadlet-update-rollback.md` — and **not** in the README, which is where a self-hosting operator lands first.

| README quick start | executable lines |
| --- | --- |
| Docker Compose | **11** |
| Podman | **43** |

A reader comparing runtimes on that page sees an 11-line path and a 43-line path, with nothing indicating the second has a one-command equivalent. That is a reasonable basis on which to pick Docker, reached from incomplete information. #4470's own changelog entry says the script exists *"closing the asymmetry with Docker's `bin/hive-setup.sh`"* — closed in `bin/`, still open in the README.

## The change

A **"One command"** section before the manual sequence, and the existing sequence kept below as **"Or, by hand"**.

Keeping the manual block matters. Its comments are the best documentation of the Podman traps anywhere in the repo — the 300-second hang behind a wrong `dashboard.port`, why `hive.env` must exist even when fully commented out, why a `0700` secrets dir presents as an SELinux denial it is not. Promoting the script by deleting them would lose more than it gained, so the new section says the manual block is worth reading either way and that the script enforces the same couplings.

## Every claim was checked against the script

I wrote the section from the script's behaviour, not from memory, and verified each line:

| claim in the new text | verified against |
| --- | --- |
| `--enable-linger` exists | `--help` |
| installs no packages, clones nothing | `--help` |
| idempotent; no overwrite without `--force`; never touches `secrets/` | `--help` |
| a failing step names itself; nothing is rolled back | `--help` |
| `dashboard.port` read from the unit, run stops on disagreement | step `4/8` (#4367) |
| volume created **through its unit** so it carries ownership labels | `hive-podman-setup.sh:581` (#4485) |
| secrets ownership correct for the root mode | step `5/8` (#4359) |
| rootless told when lingering is off | step `8/8` (#4489) |

That caught one overclaim in my first draft: I had written "leaves the host as it found it", but the script says *"nothing is rolled back"* — partial state deliberately remains for inspection. Corrected to match.

## Verified as printed

A docs change that adds a command should demonstrate the command. Run verbatim on a clean host (no config, no units, no volume, lingering off):

```
EXIT=0
1/8  Runtime selection
2/8  Host preflight (engine, root mode, cgroups; subordinate IDs, storage, networking)
3/8  Configuration in /var/home/dbaggett/.config/hive
4/8  dashboard.port must equal the unit's HealthCmd port (#4367)
5/8  Secrets directory mode and group (#4359)
6/8  Host preflight over what was just written, then install the units
7/8  Start, and confirm HEALTHY rather than started
8/8  Boot persistence — will this install survive a reboot?
  FAIL  lingering is OFF for dbaggett — this install will NOT survive a reboot
```

The step-8 failure is the correct result on a host with `Linger=no`, and is exactly the case the new text points at `--enable-linger` for.

## Scope

Docs only — no behaviour change, one file, 33 insertions, nothing removed.

`bin/hive-setup.sh` is likewise absent from the Docker section. I left that alone: the Docker path does not need its installer to look reasonable, since `docker compose up -d` is already the short version. Adding it for symmetry is a fair follow-up but a separate call.

## Provenance

Found while walking all 43 manual lines end to end during the #4188 passes, on Bluefin 44 (Silverblue), SELinux enforcing, Podman 5.8.4, `v4` at `325707ed`. **Both paths work** — this is discoverability, not correctness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,39 @@ that a process was spawned. Docker is not required and is not used.
   `:3001` serves 502s.
 - `git`, `openssl`, and a GitHub token (PAT or App) for the org the hive works on
 
+### One command
+
+`bin/hive-podman-setup.sh` does everything in the manual sequence below —
+preflights, configuration, the four Quadlet units, the boot wiring, and a final
+check that the **gateway** answers on the published port before it returns.
+
+```bash
+export HIVE_DEPLOY_RUNTIME=podman
+bin/hive-podman-setup.sh --rootless        # or --rootful
+```
+
+It installs no packages and clones nothing, is idempotent, never overwrites an
+existing config without `--force` and never touches `secrets/`. A failing step
+stops the run and names itself; nothing is rolled back, so the partial state is
+there to inspect. It also enforces three couplings that are easy to get wrong by
+hand:
+
+- `dashboard.port` is read out of the unit that will enforce it, and the run
+  stops if the config does not read back agreeing — the 300-second silent hang
+  the manual block warns about below.
+- the volume is created **through its unit**, so it carries the ownership labels
+  that make `bin/hive-podman-teardown.sh` able to see it.
+- the secrets directory gets the right ownership for the root mode, and rootless
+  installs are told when lingering is off and the deployment will not survive a
+  reboot.
+
+Add `--enable-linger` to fix that last one during the install rather than after.
+
+### Or, by hand
+
+Worth reading even if you use the script: the comments below are where the traps
+are documented, and the script enforces the same ones.
+
 The block below is **rootless**. For rootful, set `CONF=/etc/hive`, drop the
 `podman unshare` line in favour of the `chgrp` beside it, install the units into
 `/etc/containers/systemd/` with `sudo`, and drop `--user` from every `systemctl`.
--- a/README.md
+++ b/README.md
@@ -91,6 +91,9 @@ preflights, configuration, the four Quadlet units, the boot wiring, and a final
 check that the **gateway** answers on the published port before it returns.
 
 ```bash
+git clone https://github.com/kubestellar/hive.git
+cd hive
+
 export HIVE_DEPLOY_RUNTIME=podman
 bin/hive-podman-setup.sh --rootless        # or --rootful
 ```


clubanderson:
Reviewed and verified at head 3f55d1e2 (+ my f961b15d).

- **Every claim in the new section checked against the merged installer** (bin/hive-podman-setup.sh, #4472): flags `--rootless`/`--rootful`/`--force`/`--enable-linger` all real (usage at :150-186); installs no packages and clones nothing (:28, :172, by design vs hive-setup.sh's apt-get); idempotent, never clobbers hive.yaml/hive.env and never touches secrets/ with or without --force (:48-51); failing step stops, names itself, nothing rolled back and the partial state says so verbatim (:144-145). The three couplings are all real: dashboard.port written then read back with refusal on disagreement (:16, :282) — and the "300-second silent hang the manual block warns about below" cross-reference resolves (README manual block, the #4367/#4369 port trap); volume created through hive-data-volume.service so it carries the #4210 ownership labels teardown filters on (:562-584, teardown header confirms unlabelled = invisible); secrets ownership differs by root mode exactly as documented (`podman unshare chown` rootless vs `chgrp` rootful, :482-490). Linger loud-warning + `--enable-linger` opt-in match #4489/#4500 semantics; "boot wiring" = hive-boot.target + hive-boot-gate.service (#4514/#4526, :539-556); final end-to-end gateway check before returning (:111, :157).
- **No curl|bash anywhere** — the recipe runs the script from a clone, which is the acquisition #4472 shipped. No Docker leakage into the Podman path; `export HIVE_DEPLOY_RUNTIME=podman` shown is consistent with the preflight contract (the installer also exports it itself and rejects a conflicting value at :334).
- **One fix-forward** (f961b15d, DCO): the one-command block ran `bin/hive-podman-setup.sh` from a checkout it never made — the clone lived only in the manual block below. Same verbatim-runnability class as #4477. Added the clone+cd; compose quick-start contract still 7/7 (the new block has no compose command, correctly out of its scope).
- No forbidden phrases; CI green.

The "Or, by hand" framing — keep the manual block as the documentation of the traps the script enforces — is exactly right.

/lgtm
/approve

PR #4539, opened 2026-08-22
📱 contribute: add the phone-width breakpoint the /contribute page never had (#4537)

### LINKED ISSUE #4537: Operations tab overflows the right edge on iPhone portrait (no phone-width breakpoint)
## What happens

Viewing `/contribute` → **Operations** tab on an iPhone in **portrait** cuts the page off at the right edge. Panel content (the `Connected clankers` row, the per-clanker admin controls, the `declares: …` line) runs past the screen and is clipped, and the tab bar itself is truncated mid-word (`Manageme…`). Rotating the phone to **landscape** gives enough width that everything fits, so the page looks correct there.

Reproduced on iPhone Safari, portrait (~390 CSS px viewport), against a hosted hive.

## Why

The `/contribute` page (`src/pkg/dashboard/api_contribute.go`, the embedded `<style>` block) has responsive breakpoints down to 900/860/560/520 px, but nothing that handles a phone-width viewport. Four things compound:

**1. The tab bar can't wrap or scroll.**
```css
.page-tabs{display:flex;gap:2px;background:var(--cc-surface);border-bottom:1px solid var(--cc-border);padding:0 48px}
.page-tab{...;padding:14px 20px}
```
Five tabs (Onboarding / Operations / Management / Leaderboard / Profile) at `14px 20px` padding come to roughly 600 px, plus 96 px of container padding — about 700 px against a 390 px viewport. There is no `flex-wrap` and no `overflow-x:auto`, so the bar overflows the document and the last tabs are simply unreachable. This is the truncated `Manageme…` in the screenshot.

**2. `.ops` silently becomes a horizontal clip.**
```css
.ops{padding:40px 48px;overflow-y:auto}
```
Setting only `overflow-y` makes the computed `overflow-x` `auto` as well, so `.ops` becomes a scroll container that clips at its padding box — i.e. exactly at the right screen edge. That is why the card content is cut off flush with the display rather than at a card border. (`.main` on the Onboarding tab has the same `padding:40px 48px;overflow-y:auto` shape.)

**3. 96 px of horizontal padding on a 390 px screen.**
`.ops` and `.main` both use `padding:40px 48px` at every width. A quarter of a phone screen goes to gutters, leaving ~294 px of content column, which is what pushes the panels into overflow in the first place. No media query reduces it.

**4. The clanker row's admin controls can't shrink below their content.**
```css
.clanker-row{display:grid;grid-template-columns:auto auto minmax(0,1fr);...}
.clanker-row>.admin-actions,.clanker-row>.feed-time{grid-column:1/-1}
.admin-actions{display:flex;gap:6px;flex-wrap:wrap}
.clanker-act-as{display:inline-flex;align-items:center;gap:4px;...}
```
`.admin-actions` spans `1/-1`, so its min-content contribution is distributed into the two `auto` tracks (the third is `minmax(0,1fr)`, which has a fixed `0` minimum and absorbs none of it) — the auto tracks grow and the whole row gets wider than the card. Inside it, `.clanker-act-as` is a single non-shrinkable inline-flex item holding the literal text `Acting as` plus a `<select>` whose intrinsic width is set by its widest option (`none (general work)`). `.admin-act select` sets no `max-width`/`min-width:0`, so `flex-wrap` can never break that item down. In the screenshot this is the line that runs `contributor ▾  Acting as  none (general work) ▾  Re…` straight off the screen with only `Remove` wrapping.

## Suggested fix

Add a phone breakpoint (`@media(max-width:600px)`) to the `/contribute` stylesheet:

- `.page-tabs` — `overflow-x:auto;-webkit-overflow-scrolling:touch;scrollbar-width:none` (or `flex-wrap:wrap`) and reduce the side padding to ~16 px; add `flex:0 0 auto` on `.page-tab` so tabs keep their size in the scroller.
- `.ops` / `.main` — drop horizontal padding to ~16 px.
- `.ops` / `.main` — make the horizontal axis explicit rather than inherited: `overflow-x:hidden` (or drop the `overflow-y:auto`, which isn't doing much on a full-page block) so it isn't a surprise clip.
- `.clanker-row` — collapse to a single column on narrow screens, or change the track list to `auto auto minmax(0,1fr)` → something where the spanning `.admin-actions` can't inflate the `auto` tracks.
- `.clanker-act-as` — `min-width:0;flex-wrap:wrap`, and give `.admin-act select` `max-width:100%;min-width:0`.

Worth auditing the rest of the Operations panel at 390 px in the same pass — the ready-work queue rows, the triage ladder (`.cc-triage-groups`), and `.lb-row` (`grid-template-columns:56px 1fr 120px 70px 70px 80px 72px`, ~470 px of fixed tracks) look like they have the same class of problem.

Danathar:
Closes #4537.

The `/contribute` page had responsive breakpoints at 900/860/768/560/520 and then nothing, so an iPhone in portrait (~390 CSS px) ran off the right edge: panel content was cut flush with the display and the tab bar was truncated mid-word at `Manageme…`.

## What was wrong

Four causes compound. Two of them are **not phone-specific** and were fixed at base rather than inside the media query:

**`.ops-grid` used a bare `1fr` track.** A `1fr` track's automatic minimum is `min-content`, and the cards inside carry `white-space:nowrap` text (`.cc-q-title`, `.cc-army`, the clanker sub-lines), so the track resolved to the widest unbreakable line and the panel outgrew `.ops`. This overflows at 1024px too, not only on a phone — the ready-work card runs ~15px past the edge there on `v4` today. Now `minmax(0,1fr)`, which hands the overflow back to the ellipsis and wrapping the cards already declare.

**`.clanker-row`'s first two tracks were `auto`.** `.admin-actions` spans `1/-1`, and a spanning item's min-content contribution is distributed into the spanned tracks that can take it — the third track is `minmax(0,1fr)` with a fixed `0` minimum and absorbs none — so both `auto` tracks grew until the row was wider than its card. This is the issue's diagnosis, and it's exactly right. The dot and avatar are `8px` and `28px` at every width, so naming those widths is what `auto` already resolved to when nothing spanned; it just leaves nothing for the spanning row to inflate.

**`.clanker-act-as` could not shrink.** A flex item's `min-width:auto` floor kept the literal `Acting as` plus a `<select>` sized by its widest option (`none (general work)`) from ever shrinking, so `.admin-actions`' `flex-wrap` had nothing it was allowed to break. Now `min-width:0` and wrappable.

One correction to the issue's suggested fix while I was in there: the tier and acting-as controls **are** the `<select>` (the class sits on the element), so the existing `.admin-act select` descendant rule never matched them and `.admin-act select{max-width:100%}` would have been a no-op. Added `select.admin-act` instead, and left a comment saying why.

**No phone breakpoint.** New `@media(max-width:600px)` block:
- `.page-tabs` wraps instead of overflowing. I went with `flex-wrap` over `overflow-x:auto` — a scroller with a hidden scrollbar keeps the trailing tabs just as unreachable on touch, only more quietly.
- The 48px gutters drop to 16px (they were spending a quarter of the screen).
- `.main`/`.ops` name their horizontal axis rather than inheriting `overflow-x:auto` from a lone `overflow-y:auto`. That inheritance is the surprise clip that made the cut land at the screen edge instead of a card border. With the overflow sources above fixed, `hidden` is a backstop, not the thing holding the layout together.
- `.cc-army` wraps.
- `.lb-row`'s seven ~470px fixed tracks reflow onto three lines (rank + contributor / tier + the three counts / sparkline) instead of scrolling sideways. Placement is by child position because `.lb-head` and the body rows emit the same seven children in the same order, and the stat tracks stay fixed `rem` widths so the header keeps sitting over the column it labels — `auto` would size per row and drift out of alignment with the numbers.

The block sits at the end of the sheet so the phone overrides win on source order without inflating specificity, and stays ahead of the `%s` custom-CSS slot so an operator's own stylesheet still has the last word.

## Verification

Headless Chromium against markup matching what the page's JS actually emits — clanker rows with the full admin control set, leaderboard head + body rows, the five-tab bar — using the real stylesheet extracted from `api_contribute.go`. Measured `document.documentElement.scrollWidth`, every element whose right edge passes the viewport, and each tab's position:

| viewport | `v4` today | this PR |
| --- | --- | --- |
| 390 | doc **614px**, 51 elements past the edge, **3 tabs off screen** | doc 390px, 0 past the edge, 0 tabs off |
| 320 | doc 614px, 63 past the edge, 3 tabs off | doc 320px, 0, 0 |
| 430 | doc 614px, 44 past the edge, 2 tabs off | doc 430px, 0, 0 |
| 600 | doc 614px, 31 past the edge, 1 tab off | doc 600px, 0, 0 |
| 768 | clean | clean |
| 900 | clean | clean |
| 1024 | doc 1024px, **6 past the edge** (`.ops-grid`) | 0 past the edge |
| 1440 | clean | clean |

The `614px` baseline is the bug the issue reports, reproduced. The only horizontal clip remaining after the fix is `.cc-q-title`'s own `text-overflow:ellipsis`, which is intentional. No behaviour change at 1440/900/768; 1024 improves as a side effect of the `.ops-grid` fix.

`gofmt` clean, `go build` and `go vet` pass, and `go test ./pkg/dashboard/ -run "Contribute|Leaderboard|DevlogRail|Bugs"` passes. (The package's full `go test` panics in `TestBuildBudget_WithTokenCollectorSummary` — a nil `slog` logger in `tokens.NewCollector`, `pkg/tokens/collector.go:410`. Pre-existing and unrelated: this PR is CSS-only inside `api_contribute.go`.)

CSS-only, no markup or JS changes, so nothing here touches the render path, roles, or endpoints.

### DIFF
--- a/src/pkg/dashboard/api_contribute.go
+++ b/src/pkg/dashboard/api_contribute.go
@@ -920,8 +920,14 @@ code{background:var(--cc-bg);padding:2px 8px;border-radius:4px;font-size:.9rem}
 .tab-panel.active{display:block}
 .ops{padding:40px 48px;overflow-y:auto}
 .ops h1{font-size:1.7rem;margin-bottom:6px}
-.ops-grid{display:grid;grid-template-columns:340px 1fr;gap:20px;margin-top:24px}
-@media(max-width:900px){.ops-grid{grid-template-columns:1fr}}
+/* #4537: minmax(0,1fr), not 1fr. A 1fr track's automatic minimum is min-content,
+   and the cards it holds contain white-space:nowrap text (.cc-q-title, .cc-army,
+   the clanker sub-lines), so the track resolved to the widest unbreakable line
+   and the whole panel outgrew .ops. The explicit 0 minimum lets the track shrink
+   to the viewport and hands the overflow back to the ellipsis/wrapping the cards
+   already declare. */
+.ops-grid{display:grid;grid-template-columns:340px minmax(0,1fr);gap:20px;margin-top:24px}
+@media(max-width:900px){.ops-grid{grid-template-columns:minmax(0,1fr)}}
 .ops-card{background:var(--cc-surface);border:1px solid var(--cc-border);border-radius:12px;padding:0;overflow:hidden}
 .ops-card-head{padding:16px 20px;border-bottom:1px solid var(--cc-border);display:flex;align-items:center;gap:10px}
 .ops-card-head h3{font-size:.95rem;color:var(--cc-text);margin:0}
@@ -966,7 +972,14 @@ code{background:var(--cc-bg);padding:2px 8px;border-radius:4px;font-size:.9rem}
    horizontally with the multi-line text. Long repo paths in .clanker-sub wrap
    (overflow-wrap:anywhere) rather than pushing into anything. align-items:start
    keeps the dot/avatar top-aligned with the first text line. */
-.clanker-row{display:grid;grid-template-columns:auto auto minmax(0,1fr);align-items:start;column-gap:10px;row-gap:8px;padding:12px 20px;border-bottom:1px solid var(--cc-border-2)}
+/* #4537: the first two tracks are SIZED, not auto. .admin-actions below spans
+   1/-1, and a spanning item's min-content contribution is distributed into the
+   spanned tracks that can take it — the third is minmax(0,1fr) with a fixed 0
+   minimum and absorbs none, so both auto tracks grew until the row was wider
+   than its card. The dot and the avatar are 8px and 28px at every width, so
+   naming those widths is what auto already resolved to when nothing spanned;
+   it just leaves nothing for the spanning row to inflate. */
+.clanker-row{display:grid;grid-template-columns:8px 28px minmax(0,1fr);align-items:start;column-gap:10px;row-gap:8px;padding:12px 20px;border-bottom:1px solid var(--cc-border-2)}
 /* The trailing controls / timestamp: full-width line beneath the identity. It is
    always the LAST grid child, so grid-column:1/-1 drops it below regardless of
    whether it's .admin-actions or the .feed-time fallback. */
@@ -1331,7 +1344,14 @@ code{background:var(--cc-bg);padding:2px 8px;border-radius:4px;font-size:.9rem}
 .admin-act select{background:var(--cc-bg);border:1px solid var(--cc-border);color:var(--cc-text-2);font-size:.7rem;border-radius:6px;padding:2px 4px;font-family:inherit}
 .agent-role-grants{display:flex;align-items:center;gap:6px;flex-wrap:wrap;width:100%%;font-size:.7rem;color:var(--cc-muted)}
 .agent-role-grants__label{font-weight:600;color:var(--cc-text-2)}
-.clanker-act-as{display:inline-flex;align-items:center;gap:4px;color:var(--cc-muted);font-size:.72rem}
+/* #4537: a flex item defaults to min-width:auto, so this one held the literal
+   "Acting as" plus a <select> whose intrinsic width is its widest option
+   ("none (general work)") and refused to shrink — .admin-actions' flex-wrap had
+   nothing it was allowed to break, so the whole line ran off the card. */
+.clanker-act-as{display:inline-flex;align-items:center;flex-wrap:wrap;gap:4px;min-width:0;max-width:100%%;color:var(--cc-muted);font-size:.72rem}
+/* The tier and acting-as controls ARE the <select> (class on the element), which
+   is why the .admin-act select descendant rule above never matched them. */
+select.admin-act{min-width:0;max-width:100%%}
 .agent-role-chip{display:inline-flex;align-items:center;gap:4px;padding:2px 7px;border-radius:999px;border:1px solid rgba(88,166,255,.28);background:rgba(88,166,255,.08);color:#79c0ff}
 .agent-role-chip button{border:none;background:transparent;color:inherit;cursor:pointer;padding:0;line-height:1;opacity:.75;font:inherit}
 .agent-role-chip button:hover{opacity:1;color:#f85149}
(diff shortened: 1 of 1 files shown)


clubanderson:
Reviewed and verified at head 4788e800.

- **All four causal claims check out against the actual stylesheet/markup**: `.cc-q-title` is `white-space:nowrap` with ellipsis and `.cc-army` is a no-wrap flex row, so the bare `1fr` track's min-content floor is exactly the overflow mechanism described — `minmax(0,1fr)` hands the width back to the ellipsis the cards already declare, and is behavior-identical to `1fr` whenever content fits (no desktop regression). `.clanker-row`: the dot is 8px (`.clanker-dot`) and the avatar 28px (`.clanker-av`), both fixed at every width, so naming those widths is what `auto` resolved to unspanned — and `.admin-actions` does span `1/-1` (line 986), which is the min-content distribution the fix removes. `select.admin-act` is the element class in the emitted markup (:4518, :4612) so the `.admin-act select` descendant rule indeed never matched — the new rule is the right selector.
- **Phone block is additive and last-wins by source order**: existing 900/860/768/560/520 queries untouched (only the 900px `.ops-grid` line changes track sizing, same query); the 600px block sits ahead of the operator custom-CSS `%s` slot, so operator styles still win. Tab bar wraps rather than scrolls — all five tabs on screen and tappable at 390px. `overflow-x:hidden` is scoped to the phone query as a stated backstop; desktop overflow behavior unchanged.
- **Leaderboard reflow is position-safe**: `.lb-head` emits the same seven children in the same order as body rows (:2934 vs the row renderer), so the nth-child placement keeps headers over their columns; base seven tracks (56px/1fr/120px/70px/70px/80px/72px ≈ 470px fixed) genuinely cannot fit 390px, and fixed rem stat tracks keep header/number alignment.
- **No collateral**: `go build` + dashboard package tests pass; no JS touched; no unescaped `%` in the added CSS (Go fmt template safe); dashboard-lint green in CI; trial-merge against current v4 clean (no CHANGELOG entry to conflict). No forbidden phrases. CI green.

The 1024px `~15px` overflow figure I could not reproduce pixel-exact without a browser, but the mechanism is textbook (auto minimum = min-content) and the fix is correct regardless of the exact magnitude.

/lgtm
/approve

PR #4548, opened 2026-08-22
fix(agent): pass -y to skills add so the caveman install actually completes headlessly

hanthor:
Follow-up to #4518 (the tag-pin fix, merged). On a live hive the codex/goose/aider installs exit 0 but write **nothing**: without `-y`, `skills add <pkg> -a codex` renders an interactive multi-select menu, and in the hive's captured-output non-TTY context the menu auto-cancels. The `npx -y` only confirms the npx install itself; the skills CLI needs its own `-y`.

Verified on a live hive: with `-y` the install completes (`Done!`) and writes the skills to the agent's skills dir; without it, exit 0 and no SKILL.md anywhere.

### DIFF
--- a/src/pkg/agent/caveman_pin_test.go
+++ b/src/pkg/agent/caveman_pin_test.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"os"
 	"regexp"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed from head 1856ad61 (rebased 0c2d624b onto v4 to clear the post-#4518-squash conflict, plus one additive guard commit).

- **Claim verified against the actual skills CLI** (vercel-labs/skills v1.5.23, ran `npx -y skills add --help` locally): Add Options include `-y, --yes — Skip confirmation prompts`. It only confirms defaults — agent targeting stays explicit via `-a`, so nothing changes about WHAT gets installed, just that the interactive confirm no longer auto-cancels in the hive's non-TTY captured-output context. npx's leading `-y` indeed only covers the npx package fetch — the diagnosis is right, and exit-0-but-empty is a nasty silent-failure class to have killed.
- All three `skills add` sites (goose/codex/aider-desk) updated consistently; the npx-direct paths (claude/copilot/gemini) correctly untouched — they don't go through the skills CLI. `exec.Command` arg vectors, no shell quoting in play.
- **Fix-forward added** (DCO, co-authored): extended #4529's guard file with `TestCavemanSkillsAddIsHeadless` — every `skills add` line in manager.go must carry `-y` AFTER `"add"` (npx's leading -y deliberately doesn't count). Mutation-checked: removing one -y turns it red; the first draft that scanned the whole line false-negatived on npx's -y, which is exactly why the after-add cut exists.
- DCO present on both commits; build/vet/gofmt clean; both caveman guards pass locally; trial-merge vs latest v4 clean; CI green minus tide.

Thanks @hanthor — second real-fleet fix in this path today.

/lgtm
/approve

Issue #2612, opened 2026-08-05
contribute: leaderboard —


castrojo:
Reporting an issue with the /contribute page.

Page/tab: leaderboard
URL: https://hosted-projectbluefin-knuckle-gjvq.hive.kubestellar.io/contribute/leaderboard

---

What happened / what would you like to see?

Here is some research we did on adding things to the contribution page that are worth investigating. 

https://github.com/projectbluefin/documentation/issues/1003

clubanderson:
Thanks for this — we want to make sure we build the right thing rather than guess from the linked epic (which spans a broader multi-page dashboard redesign).

The `/contribute` **leaderboard** tab currently has a few distinct sections:
1. the personal **"Me" card** at the top (your tier, milestones, rank, quota),
2. the **"Rankings"** table (all contributors, done/failed, tier badges — internal hive agents are already filtered out per #2601),
3. the tier medallions / stats presentation.

Could you point us at the **exact section(s)** you'd like improved, and what "better" looks like for each? For example:
- Do you want the **community-contributors vs. hive-agents** distinction shown as separate labeled sections (rather than agents just hidden)?
- A **per-cycle / timeframe** view of the rankings?
- Additional **stats/columns** on the Rankings table, or a different presentation of them?
- Or is this really about the **Me card** / a specific contributor's view?

From the `projectbluefin/documentation#1003` research, some ideas map cleanly to what we already track (community/agent separation, richer real per-contributor stats) and some (the triage-stage pipeline: Triaging→Spec→Implement→Review→Closed, issue-velocity, merge-time attribution) would need data the leaderboard doesn't currently carry per contributor — so we'd want to know which of those matter to you before we build, to avoid faking metrics.

Point us at the specific section(s) + the outcome you want and we'll scope it accurately.

clubanderson:
[...]

PR #4554, opened 2026-08-22
🌓 contribute: add the light/dark theme selector the palette was waiting for (#4549)

### LINKED ISSUE #4549: Add a light/dark theme toggle to the /contribute page (palette exists, no way to select it)
## Summary

The `/contribute` page (Onboarding / Operations / Management / Leaderboard / Profile) has a complete light-mode palette, but no way to choose it. It renders light only if the visitor's OS/browser is already set to a light appearance. There is no in-page toggle and no persisted preference, so a visitor on a dark-set device cannot get the light theme, and a visitor on a light-set device cannot pin dark.

The main dashboard (`static/index.html`) already has a light/dark toggle in its top bar. The contributor page just never got the equivalent.

## Current state

In `src/pkg/dashboard/api_contribute.go`, the palette is fully tokenized (`--cc-bg`, `--cc-surface`, `--cc-border`, `--cc-text`, `--cc-muted`, accents). `:root` carries the original dark values byte-for-byte; a full Primer-light ramp is defined under **two** hooks:

- `@media(prefers-color-scheme:light){:root:not([data-theme="dark"])}` — line 844
- `:root[data-theme="light"]` — line 861

plus a light-mode fixup pass at lines 1769–1787 for the surfaces baked as literal dark hex (`.ops-card.card-accent>.ops-card-head` gradient, `.cc-ach`, `.client-tile:hover`, `.invite-banner`, `.cc-report-link:hover`, and the muted status pills).

So the explicit `data-theme` hook is already wired for both directions — `[data-theme="light"]` forces light, `[data-theme="dark"]` wins back to dark even under a light OS preference. **Nothing ever sets the attribute.** Grepping the package, `data-theme` appears only in CSS selectors: no JS writes it, no `?theme=` query param, no localStorage preference. The source comment at line 823 says so outright:

> Light mode activates on BOTH signals — the OS preference AND an explicit `:root[data-theme="light"]` hook — so a future in-page toggle can force it (the toggle itself is out of scope).

This issue is that follow-up.

Note the `?style=` custom-CSS param is not a workaround: `scopeThemeAncestorSelector` (`api_leaderboard_style.go:777`) scopes injected selectors to `#hive-dashboard-root` / `#tab-leaderboard`, so custom CSS can *react* to a theme ancestor but cannot set one, and it does not reach the onboarding or operations panels.

## Proposal

Add a small theme control to the `/contribute` page chrome (next to the tab bar), cycling **auto → light → dark** and persisting to localStorage:

- `auto` (default) — remove the attribute, keep today's `prefers-color-scheme` behaviour, so nothing changes for existing visitors.
- `light` / `dark` — set `document.documentElement.dataset.theme`, which the existing CSS already honours.

No new palette work should be needed — both ramps exist.

### Implementation constraints

- **Set the attribute on `:root`, not `<body>`.** The dashboard's `applyLayout()` (`static/index.html:3777`) uses a `light-mode` body class with the `hive-layout-mode` localStorage key; the contribute page's CSS is built around `:root[data-theme]`. Match the contribute page's own hook rather than porting the dashboard's mechanism, and pick a distinct storage key (e.g. `hive.contribute.theme`) so the two surfaces don't fight.
- **Avoid the flash.** The stored preference has to be applied before first paint, which means an early inline `<script>` in `<head>` — the same shape as the dashboard's FOUC guard at `static/index.html:2628`.
- **CSP.** Inline `<script>` elements are allowed by sha256 hash, stamped per response by `applyDocumentScriptSrcElem` (`csp_script_src.go`) — an added inline script is fine and gets hashed automatically. Inline `on*=` handler attributes are **not** allowed (`script-src-attr 'none'`, ADR-0016); the toggle must dispatch through the existing `data-action` delegation.
- **Reduced motion / no transition on the swap** so switching doesn't animate the whole page.

### Worth checking in the same pass

A few one-off dark surfaces may not be covered by the 1769–1787 fixup block and would need light values once the theme is selectable by everyone rather than only reachable via an OS setting:

- `.cc-q-holdreason-input` (line 1522) — hardcoded `#0d1117` / `#30363d` / `#e6edf3` with `color-scheme:dark`.
- Six inline `style="color:#e6edf3"` literals in the markup (e.g. the `Your army` label, the Management/Operations subtitle emphasis) that stay near-white on a light surface.
- `.clanker-interests` / `.clanker-interests-label` (lines 1462–1463), pinned to `#6e7681`.

Danathar:
Closes #4549.

The `/contribute` page has carried a full Primer-light ramp since #2612, keyed on `:root[data-theme]` as well as `prefers-color-scheme` — but nothing ever wrote the attribute. Grepping the package, `data-theme` appeared only in CSS selectors. So light was reachable only if the visitor's OS already asked for it: a visitor on a dark-set device could not get light, and one on a light-set device could not pin dark. The comment at the palette said this follow-up was coming; this is it.

## The control

Cycles **auto → light → dark**, at the right of the tab bar.

`auto` is the **absence** of the attribute, not a resolved value. A visitor who never touches the control keeps exactly today's behaviour, and one who returns to auto starts following their OS again live — including a change made while the page is open. Resolving auto to a concrete value would freeze the page against that, which is the one thing the default must not do.

Against the implementation constraints in the issue:

- **No flash.** The preference lands before first paint via a minimal inline `<script>` in `<head>`, ahead of the stylesheet. It does the attribute write and nothing else — the label, persistence and cycle live in the deferred block at the foot, none of which the first paint needs.
- **CSP.** Dispatch is `data-action="cycle-theme"` through the existing delegation, never an `on*=` attribute (`script-src-attr 'none'`, ADR-0016). The inline script element is fine: `applyDocumentScriptSrcElem` hashes every inline script in the finished document.
- **Distinct key.** `hive.contribute.theme`. The dashboard's toggle keys off a body class and this one off `:root[data-theme]`, so a shared key would have each surface writing a value the other cannot read. A test pins that they stay separate.
- **No animated swap.** The sheet transitions tiles, buttons, the ops rail and the quota bar; letting all of them run at once turns a theme change into a half-second wobble. A one-frame guard class suppresses them, cleared by `rAF` with a timeout floor so it can never stick in a background tab.

One structural note: the bar becomes `.page-chrome` (tabs + control), and the control is a **sibling** of `.page-tabs`, not a child — a non-tab child of a `role="tablist"` is announced as a stray tab. `.page-tabs` keeps its class, role and children untouched, so the tab JS and the existing tab-order tests are unaffected.

## The light-mode holes, same pass

These matter now that light is reachable by everyone rather than only via an OS setting:

- **`color-scheme` per theme.** Native chrome CSS cannot reach — the onboarding `<select>` popups, scrollbars, the caret — now follows. A forced-light page was otherwise still dropping a black dropdown over a white form. `.cc-q-holdreason-input`'s pinned `color-scheme:dark` is dropped in favour of the root declaration.
- **Stray neutrals tokenized:** `.info-btn`, `.info-pop` (+ `h4`, `code`), `.custom-css-example`, `.clanker-interests(-label)`, `.cc-q-holdreason-input`, and 27 neutral hexes across inline `style` attributes in the markup. Every swap is **identity in dark** — the token's `:root` value *is* the literal it replaced — so the dark appearance is unchanged. Inline styles take `var()`, so these theme without needing `!important` to beat the attribute. Accents (blue/amber/green/red) stay literal on purpose: they read on both themes and are pinned by exact-match tests.

## Verification

Headless Chromium against the **real rendered page** (dumped from the handler), not a mock.

**Driving the control** — auto → light → dark → auto sets/removes the attribute, repaints the body (`#ffffff` / `#0d1117`), flips `color-scheme`, relabels the button and its `aria-label`, and persists to storage.

**Under an emulated dark OS** — the case the issue says is impossible today:

| step | result |
| --- | --- |
| dark OS, no preference | `data-theme=(none)`, body `rgb(13,17,23)` — today's behaviour, unchanged |
| dark OS, pin light | `stored=light` |
| dark OS, **reload** | `data-theme=light`, body `rgb(255,255,255)` — survives the reload via the head guard |

The converse (light OS, pinned dark) also holds.

**Contrast sweep** over the page in forced light, identical probe both sides:

| | text nodes under 3.0:1 |
| --- | --- |
| `v4` today | **21** — incl. `ClankeR` at **1.18**, `Your army` at **1.11**, the Operations/Management emphases at 1.18, `<code>` at 1.54 (near-white on white) |
| this PR | **11**, every one a literal accent — identical on both sides |

Ten broken neutral surfaces fixed, zero neutral regressions.

**Layout** — document width equals the viewport at 390/430/768/1280, with the control and all five tabs on screen at each. The two elements that overflow at 390px (`#cli-select` and its wrapper) are byte-for-byte identical on `v4` and are not touched here.

`gofmt` clean, `go build`/`go vet` pass, and the contribute/leaderboard/CSP/tab/theme/style tests are green. Seven new tests cover the control's presence and placement, the auto-removes-attribute contract, the guard running before the stylesheet, key separation, the no-inline-handler rule, `color-scheme` on both hooks, and a guard against re-hardcoding neutrals in markup.

One pre-existing assertion pinned an inline hex (`Powered by <strong style="color:#e6edf3">ClankeR</strong>`); it is updated to the tokenized form, since it guards the sentence, not the hex it happened to be written in.

Worth flagging separately, not fixed here: the remaining 11 low-contrast items are the dark-tuned accent hexes (`#58a6ff`, `#d29922`, `#3fb950`) on light surfaces. They are pre-existing and identical on `v4` — anyone with a light OS preference sees them today — and the palette comment says accents are deliberately left literal and exact-match tested, so retuning them is a separate call. Happy to open an issue.

### DIFF
--- a/src/pkg/dashboard/api_contribute.go
+++ b/src/pkg/dashboard/api_contribute.go
@@ -806,6 +806,18 @@ func (s *Server) handleContributeLanding(w http.ResponseWriter, r *http.Request)
 	var page bytes.Buffer
 	fmt.Fprintf(&page, strings.ReplaceAll(`<!DOCTYPE html>
 <html><head><meta charset="UTF-8"><meta name="viewport" content="width=device-width,initial-scale=1"><title>Contribute to %s</title>
+<!-- #4549 theme FOUC guard. Runs BEFORE the stylesheet below is parsed, so a
+     visitor who pinned a theme never sees a frame of the other one. Kept to the
+     single attribute write on purpose: everything else about the control (the
+     button label, persistence, the cycle) lives in the deferred block at the
+     foot of the document, because none of it affects the first paint. An inline
+     script element is fine under CSP — applyDocumentScriptSrcElem stamps a
+     sha256 for every inline script in the finished document (csp_script_src.go);
+     it is inline on*= ATTRIBUTES that are forbidden (ADR-0016), which is why the
+     button dispatches through data-action instead of onclick. -->
+<script>
+(function(){try{var t=localStorage.getItem('hive.contribute.theme');if(t==='light'||t==='dark')document.documentElement.setAttribute('data-theme',t);}catch(e){}})();
+</script>
 <style>
 /* Michroma display face, base64-embedded (no network fonts). Used ONLY by the
    dossier hero name + rank designation (.dz-heroname / .dz-rankpill .rank-name). */
@@ -823,8 +835,13 @@ func (s *Server) handleContributeLanding(w http.ResponseWriter, r *http.Request)
    :root[data-theme="light"] hook — so a future in-page toggle can force it
    (the toggle itself is out of scope). data-theme="dark" always wins back to
    dark even under a light OS preference. Values are Primer-light neutrals so the
-   surface blends with the surrounding Docusaurus docs chrome. */
+   surface blends with the surrounding Docusaurus docs chrome.
+   #4549: each block also declares color-scheme, which is what tells the browser
+   how to paint the chrome CSS cannot reach — the <select> popups on the
+   onboarding tab, scrollbars, and the focus ring. Without it a forced-light page
+   still drops a black dropdown over a white form. */
 :root{
+  color-scheme:dark;
   --cc-bg:#0d1117;
   --cc-bg-deep:#010409;
   --cc-surface:#161b22;
@@ -842,6 +859,7 @@ func (s *Server) handleContributeLanding(w http.ResponseWriter, r *http.Request)
   --cc-red:#f85149;
 }
 @media(prefers-color-scheme:light){:root:not([data-theme="dark"]){
+  color-scheme:light;
   --cc-bg:#ffffff;
   --cc-bg-deep:#f6f8fa;
   --cc-surface:#f6f8fa;
@@ -859,6 +877,7 @@ func (s *Server) handleContributeLanding(w http.ResponseWriter, r *http.Request)
   --cc-red:#cf222e;
 }}
 :root[data-theme="light"]{
+  color-scheme:light;
   --cc-bg:#ffffff;
   --cc-bg-deep:#f6f8fa;
   --cc-surface:#f6f8fa;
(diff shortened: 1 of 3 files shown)


clubanderson:
Reviewed at head 0a0fdd45. Fact-check against the rendered implementation:

- **CSP**: verified `applyDocumentScriptSrcElem` (csp_script_src.go:136) rewrites script-src-elem from the complete rendered document before first Write at api_contribute.go:6243 — the new head FOUC guard is hashed like every other inline script. No on*= attributes anywhere in the diff (only the test's bad-list strings); dispatch is data-action delegation per ADR-0016.
- **Auto = absence**: `ccApplyTheme` removes `data-theme` for auto and removes the storage key, so untouched visitors keep live prefers-color-scheme following. Initial sync runs with persist=false — page load never writes storage.
- **Identity-in-dark**: spot-checked all seven neutral tokens — :root values are byte-identical to the replaced literals (#161b22/#30363d/#21262d/#e6edf3/#c9d1d9/#8b949e/#6e7681). Light ramps mirror correctly; color-scheme set per theme at root. Accents left literal as stated.
- **Storage-key separation**: `hive.contribute.theme` distinct from the dashboard's hive-layout-mode; test pins both directions.
- **Transition guard**: double-rAF + 150ms setTimeout floor means cc-theme-switching can never stick in a background tab. Sound.
- **No collision with #4539/#4543**: trial-merged over current v4 head — clean auto-merge; base `.page-tabs{flex-wrap:wrap}` from #4543 preserved (now tokenized), 600px block gains .page-chrome{flex-wrap:wrap} + toggle margin, cascade coherent. Theme/landing/tab/CSP tests pass on the merged state.
- Tablist-sibling structure test correctly guards against stray-tab announcement; `.page-tabs` shape pinned for existing tab tests.
- No forbidden phrases; CI green.

Excellent, thorough work as usual.

/lgtm
/approve

PR #4562, opened 2026-08-22
🌱 test(hub): lift hub coverage back over its floor, fix doubled-ago health chip (#4557)

clubanderson:
Fixes #4557 — the coverage gate failed on v4 @ 5b437d10 with hub at 86.9% vs the 87 floor.

**Root cause**: #4547 (/my-hives ACMM hive-health verdict) and #4552 (/fleet) landed ~150 statements of new hub code with the reason-phrase and freshness-banding branches untested (hiveHealthFor 79.2%, bandFreshness 78.6%, maxRFC3339 57.1%, advisoryAge/humanizeAge 66.7%), which nudged the package from 87.2% under the floor. The gate run happened to sample at the #4529 merge commit (test-only), i.e. the drop pre-dated it.

**What this adds** (meaningful branch tests, not padding — one real bug found and fixed):
- 🐛 **Fix**: L2 advisory health chip rendered **"advisory 3h ago ago"** — advisoryAge appended " ago" on top of humanizeAge's own suffix. Caught by the new reason-phrase tests; CHANGELOG entry added.
- health_verdict: every operator-facing reason phrase per ACMM band, maxRFC3339 all branches, humanizeAge incl. clock-skew.
- my_hives_query: advisory-*/budget-* filter chips, all sort keys, summary bucket counters.
- advisory diagnostics: snapshot + log + admin endpoint + test-mode ticker no-op.
- reach: reporter snapshot hygiene, wiring seams, ancestry backend selection, PR-source mapping/error propagation vs a fake GitHub API, history cursor eviction + save discipline.
- lite-enrollment SSRF gates, dibs checker, served-host memoisation, spokeDomainKey resolution order.

**hub coverage: 87.05% → 88.1%** locally (floor 87).

PR #4567, opened 2026-08-22
✨ Fleet: one-click kubectl+tmux attach chip per hive

clubanderson:
Closes the find→attach→diagnose loop on /fleet: each hosted hive row gets an **⎘ attach** chip that copies `kubectl --context <cluster> -n <namespace> exec -it deploy/hive -- tmux attach` to the clipboard. Hover shows `cluster · namespace`. Data was already in the my-hives payload (`clusterId`/`namespace` on RegistryEntry) — this renders it. Capture-phase click so the chip doesn't expand the row; clipboard API with prompt() fallback; no chip for BYO hives (no namespace reported). JS syntax-checked; hub static/route tests green.

### DIFF
--- a/src/pkg/hub/health_verdict.go
+++ b/src/pkg/hub/health_verdict.go
@@ -194,7 +194,10 @@ func grantRosterFor(agents []AgentSummary, grant func(AgentSummary) bool) grantR
 		if !grant(a) {
 			continue
 		}
-		if a.ExpectedActive {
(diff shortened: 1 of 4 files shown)


clubanderson:
Pushed a second commit healing the v4 test collision between #4562 (reason-phrasing tests) and #4564 (write-stream verdict) — v4 HEAD's hub shard was red; this branch now carries the fix.

PR #4578, opened 2026-08-22
🌱 test: lift the three lowest packages over 90% and turn the coverage ratchet (#4570)

### LINKED ISSUE #4570: [Coverage] packages below 90% threshold
## Coverage gate failed

- **Measured:** `v4` @ `5f72752311cf2d7166b9698502b19c0033176c9d`
- **Run:** https://github.com/kubestellar/hive/actions/runs/32550931603
- **Total coverage:** 87.9%

### What the gate reported

```
hub: TESTS FAILING
```

Danathar:
Closes #4570.

## What #4570 actually reported, and what is left

The gate filed this issue for `hub: TESTS FAILING`, not for a coverage
shortfall — `TestHiveHealthForReasons` was a semantic conflict between the
admin-merged #4562 and #4564 (the L2 on-duty gate and the L3–L5 `create`→`write`
rename). That was fixed by #4571; the hourly gate has been green since run
`32553411579`.

What the issue's *title* points at is still true, though, and the floor had gone
slack in two ways: three packages sit far below the 90% default, and six more
had drifted well above floors recorded at `d89121f5` and were no longer catching
anything. This PR closes both.

## Real coverage on the three lowest packages

| package | before | after |
|---|---|---|
| `pkg/sandbox` | 67.6% | **98.5%** |
| `pkg/ioscan` | 71.0% | **97.1%** |
| `pkg/pushbroker` | 72.8% | **100.0%** |

All three are security-path code, and the untested statements were the
security-relevant ones.

**`pkg/sandbox`** — `PodmanLauncher.Run` was 20% covered because every test that
reached it needed a working rootless container runtime, so CI only ever ran the
"podman missing" branch. Pointing `Binary` at a recorded stub script makes the
whole body deterministic on every runner: argv construction, env sanitization,
stdout/stderr capture, exit-code reporting — including the assertion that a
credential env var never reaches podman's argv. Also covers the spec-validation
rejections, the `WorkDir` default, and the rule that naming a credential in
`EnvAllowlist` does not pass its value through.

**`pkg/ioscan`** — `confusableASCII` was **12%** covered: 22 of the 25 homoglyphs
an attacker can substitute for their ASCII twin were never exercised, so a gap in
that fold table would have been silent — the scanner would keep returning "clean"
for text that reads as an injection. The whole table is pinned now, plus an
end-to-end check that a Cyrillic-spoofed `ignore previous instructions` folds
back to ASCII, trips the rule, *and* raises the steganography finding. Adds the
canary registry's disk round trip and best-effort-save behavior, the semantic
verdict/redaction marker, classifier retry exhaustion, and the
OpenAI-compatible client's error paths (non-200, unparseable body, no choices,
dead transport).

**`pkg/pushbroker`** — covers every rejection stage of `Broker.Run` through a
scripted git runner (unreadable HEAD, nothing committed, mint failure, empty
token, rejected push), both base-ref fallbacks, and `GitHubAppMinter` — including
that the minted token is scoped to the *single* repository and that an
unspecified tier lands on the least-privileged default rather than on whatever
the installation grants.

## Turning the ratchet

Both copies of `PKG_THRESHOLD` (`coverage-hourly.yml` and `v2-tests.yml`) are
updated together, per their own KEEP-IN-SYNC note.

| package | floor | holds | why |
|---|---|---|---|
| `sandbox` | 67 → **98** | 98.5 | tests in this PR |
| `ioscan` | 71 → **96** | 97.1 | tests in this PR |
| `pushbroker` | 72 → **99** | 100.0 | tests in this PR |
| `intent` | 65 → **87** | 87.7 | drift; no code change |
| `tracing` | 76 → **85** | 86.0 | drift; no code change |
| `dashboard` | 78 → **82** | 83.1 | drift; no code change |
| `scheduler` | 89 → *removed* | 90.4 | clears the 90% default |
| `escalation` | 89 → *removed* | 93.2 | clears the 90% default |

No floor was lowered and nothing was raised to paper over a drop. Left alone
deliberately, because the headroom is too thin to ratchet safely: `agent` (90.6),
`hivectl` (89.0), `knowledge` (89.2), `proxy` (89.1), `github` (88.0), `hub`
(88.1), `retro` (82.4), `review` (80.8).

## Verification

Every number was re-measured on `v4` at `cd8aee5d` and cross-checked against the
last four green hourly runs, which agree to the tenth. The three touched packages
pass `-short -race -count=3`; `go vet` and `gofmt` are clean on them, and
`go build ./...` passes. Both workflow files parse as YAML and their edited
`run:` scripts pass `bash -n`.

Test-only plus a CI floor change, so no CHANGELOG entry — that is exactly what
`CHANGELOG.md`'s own guidance excludes.

### DIFF
--- a/.github/workflows/coverage-hourly.yml
+++ b/.github/workflows/coverage-hourly.yml
@@ -107,7 +107,7 @@ jobs:
           # KEEP IN SYNC with the copy of this map in v2-tests.yml's coverage
           # job (the pre-merge half of this gate, #4112).
           declare -A PKG_THRESHOLD=(
-            [dashboard]=78   # ~20% is inception tmux watcher + copilot/claude device-flow + /data file I/O + k8s
+            [dashboard]=82   # ~17% is inception tmux watcher + copilot/claude device-flow + /data file I/O + k8s
             [hub]=87         # kubectl-shelling + live-GitHub paths; see #2355. Was 89 against v2's 89.9%; v4 measures 87.6%.
             [agent]=89       # pre-existing shortfall inherited from #2350, NOT a new regression. See #2355.
 
@@ -130,19 +130,27 @@ jobs:
             # Raise these toward 90 as tests land; never raise one to paper over
             # a drop. A package that improves should have its floor moved up in
             # the same PR, so the ratchet only turns one way.
-            [intent]=65
-            [sandbox]=67
-            [ioscan]=71
-            [pushbroker]=72
-            [tracing]=76
+            #
+            # ── Ratchet turned for #4570 ─────────────────────────────────────
+            # sandbox/ioscan/pushbroker: tests added in the same PR, so the
+            #   floors move with the gain (67.6→98.5, 71.0→97.1, 72.8→100.0).
+            # intent/tracing/dashboard: NO code change — these had drifted far
+            #   above floors recorded at d89121f5 and were no longer catching
+            #   anything. Re-measured on v4 at cd8aee5d and moved up to what
+            #   they actually hold, with ~1 point of headroom.
+            # scheduler (90.4) and escalation (93.2) now clear the 90% default
+            #   and no longer need an entry at all.
+            [intent]=87
+            [tracing]=85
             [review]=80
             [retro]=82
             [github]=88
             [hivectl]=88
             [knowledge]=89
-            [scheduler]=89
-            [escalation]=89
             [proxy]=89
+            [ioscan]=96
+            [sandbox]=98
+            [pushbroker]=99
           )
 
           # Capture the full `go test -cover` output ONCE, then classify every
--- a/.github/workflows/v2-tests.yml
+++ b/.github/workflows/v2-tests.yml
@@ -337,7 +337,7 @@ jobs:
           # KEEP IN SYNC with coverage-hourly.yml, which applies the same floors
           # on its hourly scheduled run.
           declare -A PKG_THRESHOLD=(
-            [dashboard]=78   # ~20% is inception tmux watcher + copilot/claude device-flow + /data file I/O + k8s
+            [dashboard]=82   # ~17% is inception tmux watcher + copilot/claude device-flow + /data file I/O + k8s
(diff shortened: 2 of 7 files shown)


Danathar:
Thanks for the independent re-measurement — and the nit was worth acting on. Fixed in 82dc78ba.

`TestOpenAIChatClientCompleteHappyPath` captured the outbound request with a single `r.Body.Read` into a `ContentLength`-sized buffer. As you say, one `Read` is not guaranteed to fill the buffer; a request split across segments would leave trailing NULs in `gotBody`, and the `temperature:0` / `max_tokens:42` `Contains` assertions would fail intermittently. That is the worst shape for this particular test, because those assertions exist to pin classifier determinism — a flake there reads as a real cache-key regression and would send someone hunting a bug that is not in the code.

Now `io.ReadAll`, with the read error surfaced rather than discarded (the old code dropped it with `_, _ =`).

Coverage is unchanged at **97.1%** for `pkg/ioscan`, so the new floor of 96 still holds with 1.1pt headroom; the package tests pass.

On the live-host hermeticity failures in `pkg/proxy`/`pkg/scheduler`/`pkg/tokens` — agreed those are pre-existing and orthogonal, and #4595 is the right place for them. No overlapping files with this PR.

clubanderson:
Reviewed and verified locally on a rebase onto current v4 (rebased/pushed 54f9c5cc — clean, no conflicts):

- All 46 new tests are meaningful behavioral pins (stub-podman launcher argv/env/exit paths, confusable-fold table, pushbroker failure stages + single-repo token scoping, ioscan classifier/canary deny paths) — not padding.
- Re-measured on current v4 head with `-short -race`: ioscan 97.1% (floor 96), sandbox 98.5% (98), pushbroker 100% (99), intent 87.7% (87), tracing 86.0% (85), dashboard 83.3% (82), scheduler 90.4% and escalation 93.2% (default 90 after entry removal). All new floors hold with headroom.
- Both PKG_THRESHOLD copies (coverage-hourly.yml / v2-tests.yml) edited identically per the KEEP-IN-SYNC note; YAML parses clean; no forbidden phrases.
- DCO present.

/lgtm
/approve

PR #4589, opened 2026-08-22
fix(contributor): prevent unattended Codex approval prompts

### LINKED ISSUE #4588: contributor: prevent interactive Codex approval prompts during unattended Hive tasks
## Summary

Hive's unattended Codex contributor path can surface interactive approval prompts to the operator. A Hive-delivered task should either complete within an explicitly bounded execution policy or fail with an actionable machine-readable reason; it should not pause waiting for a human to approve routine checkout and GitHub operations.

## Observed behavior

During an unattended Hive task, Codex was instructed to use the existing checkout at:

```text
$HIVE_WORKSPACE_DIR/kubestellar/hive
```

The session's writable root was a different agent working directory. Routine operations then crossed two approval boundaries:

1. `git fetch` needed write access to the checkout outside the configured writable root.
2. `gh issue view` and related GitHub operations needed network access.

These actions surfaced interactive approval requests. The task could not proceed unattended until a user responded.

## Current configuration

`config/backends.conf` defaults Codex to:

```text
--ask-for-approval on-request --sandbox workspace-write
```

This safe default was introduced by #3512 to replace unconditional `--dangerously-bypass-approvals-and-sandbox`. That security improvement should remain. However, the unattended path does not currently guarantee that the task checkout is within the writable workspace or route eligible boundary requests through automatic review.

This is separate from #4000. That issue concerns Hive's ACMM-gated tool-approval architecture. This issue concerns the execution contract between Hive and the Codex backend: a headless contributor run must not depend on an interactive terminal approval.

## Expected behavior

For an unattended Codex contributor task, Hive should establish a bounded environment in which expected repository work can proceed without operator interaction:

- the resolved `$HIVE_WORKSPACE_DIR` checkout is inside Codex's writable scope;
- required GitHub network operations can be evaluated without prompting the operator;
- eligible sandbox-boundary requests use Codex automatic review or an equivalently bounded unattended policy;
- a denied, timed-out, or unsupported action fails clearly and is reported to Hive rather than waiting indefinitely;
- the dangerous no-sandbox bypass remains an explicit opt-in, not the default.

Codex supports automatic approval review while retaining `workspace-write`, for example:

```toml
sandbox_mode = "workspace-write"
approval_policy = "on-request"
approvals_reviewer = "auto_review"
```

The precise implementation may instead use a Hive-managed permission profile, provided the unattended and security properties above hold.

## Acceptance criteria

- [ ] Hive resolves the task checkout and Codex writable roots consistently; `$HIVE_WORKSPACE_DIR/kubestellar/hive` does not require an interactive filesystem escalation.
- [ ] Routine authenticated GitHub reads and writes required by the contributor workflow do not prompt the operator during an unattended run.
- [ ] The unattended Codex launch path uses automatic review or another documented, bounded non-interactive policy.
- [ ] Denial and automatic-review failure/timeout produce an actionable terminal result that the relay reports to Hive.
- [ ] Existing explicit opt-in behavior for `HIVE_CODEX_DANGEROUSLY_BYPASS_APPROVALS_AND_SANDBOX=1` remains unchanged.
- [ ] An integration test exercises fetch/fork, edit, validation, commit, push, and PR/issue creation without an interactive approval response.

## References

- #3512
- #4000
- OpenAI Codex approval guidance: https://learn.chatgpt.com/docs/agent-approvals-security

— hive: backend=codex

Danathar:
Closes #4588.

## Summary

- keep Codex on `on-request` plus `workspace-write`, but route eligible boundary requests through its `auto_review` reviewer so unattended Hive work does not pause for a person
- grant the configured `HIVE_WORKSPACE_DIR` as the exact additional writable tree while preserving the dangerous no-sandbox mode as an explicit opt-in
- surface a bounded, token-redacted terminal diagnostic when a headless approval is denied or times out
- cover the default, override, bypass, workspace, and headless failure contracts in the contributor test suites

## Tests

- `PATH=/usr/bin:/bin bash bin/contributor-agent.test.sh`
- `env -u HIVE_AGENT_CWD HIVE_HUB=wss://hive.example.test/contribute node bin/contributor-relay.test.js` (163/163)
- `node --check < bin/contributor-relay.sh`
- `shellcheck --shell=bash --exclude=SC2034 config/backends.conf`
- `shellcheck bin/contributor-agent.test.sh`

— hive: backend=codex

### DIFF
--- a/bin/contributor-agent.test.sh
+++ b/bin/contributor-agent.test.sh
@@ -232,6 +232,74 @@ case "$codex_reviewer_override_output" in
     ;;
 esac
 
+# An EXPLICITLY EMPTY reviewer must drop the -c key while KEEPING the sandbox.
+# This is the escape hatch for a Codex release that rejects the unknown
+# approvals_reviewer config key at startup: without it the only way out is
+# HIVE_CODEX_DANGEROUSLY_BYPASS_APPROVALS_AND_SANDBOX=1, i.e. no sandbox at all.
+# ${VAR:-default} would defeat this by treating empty as unset, so the guard is
+# the parameter expansion as much as the branch.
+codex_reviewer_disabled_output="$(
+  env -i \
+    PATH="${PATH}" \
+    HOME="$HOME_DIR" \
+    HIVE_REGISTRATION_TOKEN="test-token" \
+    HIVE_ENTRYPOINT_HOOK_DIR="${WORK_DIR}/empty-entrypoint.d" \
+    HIVE_CODEX_APPROVALS_REVIEWER= \
+    HIVE_WORKSPACE_DIR="${WORK_DIR}/workspace" \
+    HIVE_CONTRIBUTOR_AGENT_TEST_RESOLVE_BACKEND=1 \
+    AGENT_BACKEND=codex \
+    bash "${ROOT_DIR}/bin/contributor-agent.sh"
+)"
+case "$codex_reviewer_disabled_output" in
+  *"approvals_reviewer"* )
+    echo "expected an empty HIVE_CODEX_APPROVALS_REVIEWER to drop the -c key; got:" >&2
+    echo "$codex_reviewer_disabled_output" >&2
+    exit 1
+    ;;
+esac
+# ...but the sandbox posture and the workspace grant must survive.
+case "$codex_reviewer_disabled_output" in
+  *"backend_perm_flag=--ask-for-approval on-request --sandbox workspace-write --add-dir ${WORK_DIR}/workspace"* ) ;;
+  *)
+    echo "expected the sandbox posture to survive a disabled reviewer; got:" >&2
+    echo "$codex_reviewer_disabled_output" >&2
+    exit 1
+    ;;
+esac
+
+# A workspace path containing whitespace CANNOT be expressed: the caller
+# word-splits this string (agent-launch.sh: read -r -a PERM_ARGS <<< ...), so
+# "--add-dir /work space" would arrive as three argv words and grant Codex the
+# wrong directory. The flag must be omitted, not silently corrupted.
+mkdir -p "${WORK_DIR}/work space"
+codex_spaced_workspace_output="$(
+  env -i \
+    PATH="${PATH}" \
+    HOME="$HOME_DIR" \
+    HIVE_REGISTRATION_TOKEN="test-token" \
+    HIVE_ENTRYPOINT_HOOK_DIR="${WORK_DIR}/empty-entrypoint.d" \
+    HIVE_WORKSPACE_DIR="${WORK_DIR}/work space" \
+    HIVE_CONTRIBUTOR_AGENT_TEST_RESOLVE_BACKEND=1 \
+    AGENT_BACKEND=codex \
+    bash "${ROOT_DIR}/bin/contributor-agent.sh" 2>/dev/null
+)"
+case "$codex_spaced_workspace_output" in
+  *"--add-dir"* )
+    echo "expected --add-dir to be omitted for a whitespace workspace path; got:" >&2
+    echo "$codex_spaced_workspace_output" >&2
+    exit 1
+    ;;
+esac
+# The sandbox posture still applies — only the ungrantable flag is dropped.
+case "$codex_spaced_workspace_output" in
+  *"backend_perm_flag=--ask-for-approval on-request --sandbox workspace-write -c approvals_reviewer=auto_review"* ) ;;
+  *)
+    echo "expected the sandbox posture to survive a whitespace workspace path; got:" >&2
+    echo "$codex_spaced_workspace_output" >&2
+    exit 1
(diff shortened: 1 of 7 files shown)


Danathar:
Thanks — both observations were real, and (1) turned out to have more teeth than the review credited it with. Addressed in 63a3fc15.

**1. `approvals_reviewer` had no off switch at all.**

You asked what happens if a Codex release rejects the unknown `-c` key at startup. Checking the escape path, there wasn't one: the value was read as `${HIVE_CODEX_APPROVALS_REVIEWER:-auto_review}`, and `:-` treats an explicitly empty value as unset — so `HIVE_CODEX_APPROVALS_REVIEWER=` still produced `auto_review`.

```
$ HIVE_CODEX_APPROVALS_REVIEWER= bash -c 'echo "[${HIVE_CODEX_APPROVALS_REVIEWER:-auto_review}]"'
[auto_review]
```

So the only way out of an incompatible Codex was `HIVE_CODEX_DANGEROUSLY_BYPASS_APPROVALS_AND_SANDBOX=1` — the exact unsandboxed posture this PR exists to remove. Now `${VAR-auto_review}` (no colon), and the flag is skipped when empty:

```
default:        --ask-for-approval on-request --sandbox workspace-write -c approvals_reviewer=auto_review --add-dir /work
reviewer='':    --ask-for-approval on-request --sandbox workspace-write --add-dir /work
```

An operator can drop just the reviewer key and keep the sandbox. I went with this over a `just contribute-check codex` preflight because it is the actual remedy rather than a detector, and it works on a host where the check never ran; the docs now name it as the supported response, in preference to the bypass. I have not validated the key against a live Codex — flagging that the gap you identified is mitigated, not closed.

**2. `--add-dir` word-splitting.**

Confirmed: `agent-launch.sh:160` does `read -r -a PERM_ARGS <<< "$PERM_FLAG"`, so the contract cannot carry a path with whitespace — `--add-dir /work space` reaches Codex as three argv words and grants the wrong directory. Rather than a doc note alone, it now refuses to emit a grant it cannot express and says why:

```
$ HIVE_WORKSPACE_DIR='/work space' backend_perm_flag codex
WARNING: HIVE_WORKSPACE_DIR contains whitespace (/work space); omitting Codex --add-dir
because the flag string is word-split by the caller. ...
--ask-for-approval on-request --sandbox workspace-write -c approvals_reviewer=auto_review
```

The sandbox posture survives; only the ungrantable flag is dropped.

**3. Last-non-empty-line diagnostic** — agreed, leaving as is. Bounded at 500 chars and redacted, so the worst case is a less useful message rather than an unsafe one, and picking the "real" error out of a CLI epilogue would need per-CLI knowledge I would rather not guess at.

**Verification:** default output is byte-identical, so the relay test's pinned argv is unchanged (163/163 pass). New shell assertions pin both behaviours and I checked they fail against the pre-fix config. `bin/contributor-agent.test.sh` passes end to end with codex off PATH; it still aborts at the pre-existing non-hermetic NOT_INSTALLED check when a real codex is present, which reproduces identically on the unmodified branch (#4592).

clubanderson:
Reviewed at head 63a3fc15. Fact-check:

- **No safety gate weakened**: #3512's default posture survives intact — `--ask-for-approval on-request --sandbox workspace-write` — with escalations routed through Codex's bounded `auto_review` reviewer instead of a pane prompt no one is watching. The full `--dangerously-bypass-approvals-and-sandbox` stays strictly opt-in, and the bypass path emits neither the reviewer key nor `--add-dir` (test-pinned). `--add-dir` grants exactly $HIVE_WORKSPACE_DIR, not a broader sandbox.
- **All invocation sites covered**: grepped every codex flag-construction path — `backend_perm_flag` in config/backends.conf is the single source, consumed by agent-launch.sh (`read -r -a` word-split), contributor-agent.sh, hive.sh (both call sites), and the relay's sourced probe. No Go-side or entrypoint duplicate builds approval flags.
- **Word-split contract honoured**: the whitespace-workspace guard is real and necessary — callers word-split the flag string, so a spaced path would corrupt argv; the function refuses the grant with a stderr warning instead. Warning goes to stderr so it can't leak into the flag line.
- **Escape hatch is correctly ${VAR-default}** (no colon): explicitly-empty HIVE_CODEX_APPROVALS_REVIEWER drops just the `-c` key while keeping the sandbox — the failure mode it protects against (a Codex release rejecting the unknown config key, with the only prior escape being the full bypass) is exactly right.
- **Headless failure surfaces honestly**: relay pins the workspace at startup, and a denied/timed-out approval now reaches the hub as a bounded (500-char), token-redacted diagnostic tail line rather than an opaque exit code.
- **Guard tests**: default/override/empty-reviewer/bypass/no-workspace/whitespace contracts all pinned in contributor-agent.test.sh; relay 163/163; shellcheck clean. All verified locally.
- Trial-merge against v4 head (a19367ad) clean; CI green.

/lgtm
/approve

PR #4594, opened 2026-08-22
🐛 fix(deploy): keep ttyd's --url-arg when it is credentialed (#4593)

### LINKED ISSUE #4593: 🐛 Dashboard ▶ terminal is broken on every hive with a dashboard token — ttyd loses --url-arg when credentialed (regression from #2178)
## Summary

The dashboard's per-agent **▶ terminal** link is broken on every hive that has a `HIVE_DASHBOARD_TOKEN`. ttyd is started without `-a` (`--url-arg`), so it discards the `?arg=<session>` the dashboard puts in the terminal URL. `ttyd-tmux.sh` then falls back to its default session name and fails with:

```
error: no tmux socket found for session 'supervisor'
```

The message names `supervisor`, but real sessions are `hive-<name>` — so the fallback is the tell.

## Root cause

`src/deploy/entrypoint.sh` builds ttyd's flags in a variable that the credential branch **replaces** rather than appends to:

```sh
CRED_ARGS="-a"
if [ -n "$TTYD_CRED" ]; then
  CRED_ARGS="-c ${TTYD_CRED}"     # <- drops -a
fi
...
ttyd -W ${CRED_ARGS} -i "${TTYD_BIND}" -p "${TTYD_PORT}" ... /usr/local/bin/ttyd-tmux.sh
```

`TTYD_CRED` is non-empty whenever `HIVE_DASHBOARD_TOKEN` is set:

```sh
TTYD_CRED="${HIVE_TTYD_CREDENTIAL:-}"
if [ -z "$TTYD_CRED" ] && [ -n "${HIVE_DASHBOARD_TOKEN:-}" ]; then
  TTYD_CRED="hive:${HIVE_DASHBOARD_TOKEN}"
fi
```

`bin/hive-podman-setup.sh` generates `HIVE_DASHBOARD_TOKEN` unconditionally (step 3/8), and the Compose stack requires it, so `-a` is dead code on effectively every deployment.

## Regression

Introduced by #2178 (`security: harden default deployment`, 2026-07-29). Before that commit `-a` was passed unconditionally:

```diff
-    ttyd -W -a -p "${TTYD_PORT}" -t fontSize=14 -t disableLeaveAlert=true /usr/local/bin/ttyd-tmux.sh
+CRED_ARGS="-a"
+if [ -n "$TTYD_CRED" ]; then
+  CRED_ARGS="-c ${TTYD_CRED}"
+    ttyd -W ${CRED_ARGS} -i "${TTYD_BIND}" -p "${TTYD_PORT}" ... /usr/local/bin/ttyd-tmux.sh
```

The hardening itself is correct — binding loopback and credentialing ttyd are both good. Only the variable reuse is wrong: the flag is named `CRED_ARGS` but was carrying `-a` as well.

## The failure chain

1. `-a` dropped → ttyd ignores `?arg=`
2. `ttyd-tmux.sh:10` — `SESSION=${1:-supervisor}` — falls back
3. Sessions are actually `hive-<name>` (`src/pkg/agent/manager.go:1356`)
4. The glob `/tmp/tmux-*/supervisor` finds nothing → error above

## Observed

Standalone Podman/Quadlet deployment, `ghcr.io/kubestellar/hive:stable`, rootless, Bluefin 44, Podman 5.8.4.

ttyd's actual argv (token redacted) — `-c` present, **no `-a`**:

```
$ podman exec hive sh -c 'for f in /proc/[0-9]*/cmdline; do tr "\0" " " < "$f"; echo; done' | grep -i ttyd
ttyd -W -c hive:<REDACTED> -i 127.0.0.1 -p 7681 -t fontSize 14 -t disableLeaveAlert true /usr/local/bin/ttyd-tmux.sh
```

The sessions the link should reach exist and are healthy:

```
/tmp/tmux-2006/:
srw-rw----. 1 hive-quality    node 0 Aug 22 17:14 hive-quality
/tmp/tmux-2007/:
srw-rw----. 1 hive-scanner    node 0 Aug 22 17:14 hive-scanner
/tmp/tmux-2010/:
srw-rw----. 1 hive-supervisor node 0 Aug 22 17:14 hive-supervisor

$ podman exec hive sh -c 'for s in /tmp/tmux-*/*; do echo "== $s"; tmux -S "$s" ls 2>&1; done'
== /tmp/tmux-2006/hive-quality
hive-quality: 1 windows (created Sat Aug 22 13:14:28 2026)
== /tmp/tmux-2007/hive-scanner
hive-scanner: 1 windows (created Sat Aug 22 13:14:10 2026)
```

Invoking the script with the argument ttyd should have passed attaches correctly:

```
podman exec -it hive /usr/local/bin/ttyd-tmux.sh hive-scanner
```

## Fix

```sh
CRED_ARGS="-a"
if [ -n "$TTYD_CRED" ]; then
  CRED_ARGS="-a -c ${TTYD_CRED}"
fi
```

or unconditionally `ttyd -W -a ${CRED_ARGS} ...` with `CRED_ARGS` holding only the credential.

## Impact

Every browser terminal opened from the dashboard fails, on every hive with a dashboard token. The error text names a session the operator never asked for and does not mention `--url-arg`, so there is nothing in the message pointing at the cause. First-run operators hit this while trying to complete the interactive CLI login for `claude`/`copilot`/`codex`/`gemini` — which the dashboard terminal is the documented path for — so it blocks agent authentication on a fresh install.

## Suggested regression guard

`src/deploy/` has contract tests in this shape already. An assertion that the rendered ttyd command contains `-a` whenever `HIVE_DASHBOARD_TOKEN` is set would have caught this, and would catch the next refactor of the same block.

Danathar:
## What

Restores `-a` (`--url-arg`) to ttyd's argv when ttyd is credentialed, fixing the dashboard's per-agent **▶ terminal** link on every hive that has a dashboard token.

Fixes #4593.

## The bug

`src/deploy/entrypoint.sh` built ttyd's flags in a variable seeded with `-a`, and the credential branch **replaced** it rather than appending:

```sh
CRED_ARGS="-a"
if [ -n "$TTYD_CRED" ]; then
  CRED_ARGS="-c ${TTYD_CRED}"   # <- drops -a
fi
```

`TTYD_CRED` is non-empty whenever `HIVE_DASHBOARD_TOKEN` is set. `bin/hive-podman-setup.sh` generates that token unconditionally and the Compose stack requires it, so `-a` was dead code on effectively every deployment.

Failure chain:

1. ttyd without `-a` discards the `?arg=<session>` the dashboard puts in the link (`src/pkg/dashboard/static/index.html:5870`, forwarded intact by the proxy — `terminal_proxy_test.go`)
2. `ttyd-tmux.sh` receives no argument and falls back to its default session name
3. real sessions are `hive-<agent>` (`src/pkg/agent/manager.go`), so the glob `/tmp/tmux-*/supervisor` matches nothing
4. `error: no tmux socket found for session 'supervisor'`

Introduced by #2178. The hardening itself is correct — the loopback bind and the ttyd credential are both good and are untouched here. Only the variable reuse was wrong: the variable was named `CRED_ARGS` but was carrying `-a` as well.

## The fix

`-a` is now passed unconditionally on the ttyd command line, and `CRED_ARGS` holds only the credential:

```sh
CRED_ARGS=""
if [ -n "$TTYD_CRED" ]; then
  CRED_ARGS="-c ${TTYD_CRED}"
fi
...
ttyd -W -a ${CRED_ARGS} -i "${TTYD_BIND}" -p "${TTYD_PORT}" ...
```

This is the second option the issue offered, chosen over `CRED_ARGS="-a -c ..."` because it keeps the two concerns in separate places — the next edit to the credential branch structurally cannot take `--url-arg` down with it.

### Also: the error message

The issue notes the diagnostic was the other half of the cost — it named a session the operator never asked for and never mentioned `--url-arg`, so nothing in it pointed at the cause. `ttyd-tmux.sh` now distinguishes "no argument was passed" from "that session is gone", names `--url-arg` in the former case, and lists the sessions that would have worked in both:

```
error: no tmux socket found for session 'supervisor'
hint: no session name was passed to this script, so it fell back to
      'supervisor'. The dashboard sends the name as ?arg=hive-<agent>,
      which ttyd only forwards when it is started with -a/--url-arg.
      Check ttyd's argv for -a (regression #4593).
available sessions: hive-quality hive-scanner hive-supervisor
```

## Regression guard

`src/deploy/test_ttyd_url_arg.sh` (23 assertions), wired into `v2-ci.yml` beside the other `src/deploy` contract tests.

`grep -- -a src/deploy/entrypoint.sh` **passes on the buggy code** — `-a` is right there on the first line. The flag is only absent from the final argv, and only on the branch where a credential exists. So the test extracts the real flag assembly out of `entrypoint.sh`, runs it with `ttyd` stubbed by a shell function, and asserts on the resulting argument vector, once per credential configuration. Nothing is restated in the test — a copy would keep passing after the original regressed.

Covered: no credential / dashboard token / explicit credential / both set; `HIVE_TTYD_CREDENTIAL` precedence; the `hive:<token>` wrapping; and that the startup banner does not log the credential. Each case also asserts `-c` and the loopback bind are still present, **so a future fix here cannot silently undo #2178**.

The bottom half exercises steps 2–4 for real against a scratch tmux server: with no argument the attach fails and the error names `--url-arg`; with the session name the lookup resolves. That last one is asserted *positively* — via the wheel rebind the script deliberately leaves on the server — because an absence-only check ("no `no tmux socket found` in the output") passes just as happily when the script never ran at all.

## Verification

- New test on this branch: **23 passed, 0 failed**.
- New test against the pre-fix code (`entrypoint.sh` and `ttyd-tmux.sh` reverted to `upstream/v4`): **17 passed, 6 failed** — failing on exactly the three credentialed branches and passing the uncredentialed one, which is the precise signature of the regression.
- `shellcheck -S warning` clean on both changed shell files and the new test; `entrypoint.sh`'s pre-existing finding count is unchanged (5 before, 5 after).
- `bash -n` clean on all three; workflow YAML parses.
- Sibling `test_terminal_scrollback_status.sh` unaffected — it reports one failure (`control: plain copy-mode draws the marker`) both with and without this branch on local tmux 3.7b, i.e. pre-existing and unrelated.

No Go was touched. Go and container-build coverage is left to CI (no toolchain on the machine this was prepared on).


Danathar:
Thanks for the review. Both observations are fair; neither changes the code, and here is the reasoning on each.

**1. `available sessions:` enumerating agent socket names.** Deliberate, and I think the exposure is nil rather than merely bounded. That message is only reachable by someone who has already authenticated through the proxy's `/terminal` route and is being handed a writable shell in the container — at which point `ls /tmp/tmux-*/` gives them the same list, and the dashboard already displays every agent by name. So it discloses nothing the reader cannot trivially obtain, while turning a dead end into the one thing they need: a session name that would have worked. I would not have added it to an unauthenticated surface.

**2. Fixture-only redaction.** Agreed, and that is why it is safe: `render_argv` clears `HIVE_TTYD_CREDENTIAL`/`HIVE_DASHBOARD_TOKEN` before sourcing the extracted block, so only the test's own `t0ken-*` fixtures can ever reach the printed argv. The `sed` is there to keep the debug output honest about *where* a credential would appear, not as a security control.

On the anchor self-checks — glad they read as worth copying. They are the reason this test degrades to a loud failure rather than a silent no-op when `entrypoint.sh` is refactored, which is what the original `-a` regression needed and did not have.

Worth noting for anyone reading later: the reason a plain `grep -- -a entrypoint.sh` could not have caught this is that the buggy code contained `-a` on its first line (`CRED_ARGS="-a"`). The flag was only missing from the final argv, and only on the credentialed branch — which is why the test executes the real block per credential configuration instead of asserting on source text.

clubanderson:
/lgtm
/approve

Reviewed at 6d774e58. Verified independently:

- **Regression origin confirmed**: pulled #2178's own diff — it introduced `CRED_ARGS="-a"` and had the credential branch REPLACE it with `-c ${TTYD_CRED}`, where the pre-#2178 line passed `-a` unconditionally. Since HIVE_DASHBOARD_TOKEN is generated unconditionally by setup.sh and required by Compose, `-a` was dead on effectively every deployment, exactly as claimed.
- **Fix covers both paths and preserves #2178**: `-a` unconditional on the argv, CRED_ARGS credential-only; every test case also asserts `-c` and the loopback bind survive, so undoing the hardening now fails the suite too.
- **Test is honest**: extracts the real flag assembly from entrypoint.sh (not a restatement), asserts on the final argv per credential configuration, plus a live tmux-server bottom half with a positive resolution assertion. I ran it both ways: 23/23 on the fix; reverting entrypoint.sh+ttyd-tmux.sh to v4 fails exactly the credentialed branches — the regression's precise signature. (Local runs needed a `timeout` shim on macOS; CI ubuntu authoritative and green.)
- **Diagnostic improvement is right**: distinguishes no-arg-passed from session-gone, names --url-arg only in the former, lists real `hive-<agent>` sockets in both.
- Wired into v2-ci.yml beside the other src/deploy contract tests; orphan guard 18/0; shellcheck clean; no forbidden phrases; trial merge with latest v4 clean.

Merging.

PR #4595, opened 2026-08-22
🐛 fix(tests): hermetic scheduler/tokens/proxy tests on live hive hosts

### LINKED ISSUE #4585: [quality] 4 tests read/write live /data paths — fail (or mutate production state) on hive hosts
## Finding

Running `go test ./pkg/... -short` on a hive host (where `/data` exists and holds live state) fails 3 packages that are green in CI. The failures are hermeticity bugs — tests depend on fixed production paths:

1. **pkg/scheduler** `TestLoadPromptTemplate_FromPoliciesDir`: `loadPromptTemplate` checks `/data/agents/<name>/CLAUDE.md` and `/data/policies/...` *before* the test's temp `Policies.LocalDir`. On a host with a live scanner agent, the test reads the real production policy and fails asserting the temp template. (`src/pkg/scheduler/scheduler.go:132-140` — only `userSavedPolicyDir` is a test seam; the other two paths are inline literals.)

2. **pkg/tokens** `TestCollector_Summary_Initially`: `NewCollector` eagerly calls `loadSnapshot()` on `defaultPersistPath = "/data/token-summary.json"` (`src/pkg/tokens/collector.go:75,106`). `SetPersistPath` exists but comes too late — the constructor has already loaded the **live production token summary** (test output showed real per-agent totals for the running hive). Test expects nil and fails.

3. **pkg/proxy** `TestNewGitHubProxy`, `TestNewGitHubProxyWithWritableDir`: `NewGitHubProxy` writes the CA to fixed `/data/proxy-ca.pem`. Where `/data` is read-only the tests fail (`permission denied`); where it is writable, **the test would overwrite the live proxy CA** — worse than a failure.

CI stays green only because runners have no `/data`, so the reads fall through to defaults. That means the tests silently exercise different branches in CI vs. on real hosts, and running the suite on a production hive host is unsafe.

## Recommendation

- `tokens`: make snapshot loading lazy or add a constructor/option taking the persist path, so `SetPersistPath` (or a new option) applies before the first load; point tests at `t.TempDir()`.
- `scheduler`: promote the inline `/data/agents/...` and `/data/policies/examples/...` literals to package-level vars (same pattern as `userSavedPolicyDir`) and override them in tests.
- `proxy`: thread the CA path through config/parameter instead of the `/data/proxy-ca.pem` literal; tests use `t.TempDir()`.
- Optionally add a CI lane that runs the suite with a populated `/data` fixture to catch future regressions of this class.

These need small production seams (path parameterization), so the quality agent is not PRing them directly (testing-only mandate).

## Priority
- Impact: high (suite unsafe/red on production-shaped hosts; live CA overwrite risk)
- Effort: low-medium (mechanical path seams + test updates)

---
---

sidsri14:
Fixes #4585

## What changed

Three hermeticity fixes so the suites in `pkg/scheduler`, `pkg/tokens`, and `pkg/proxy` no longer depend on (or damage) live `/data` state on hive hosts:

**pkg/scheduler** (`scheduler.go`)
- Promoted the two remaining inline `/data/...` path prefixes in `loadPromptTemplate`/`loadNamedTemplate` to package-level seam vars `agentHomeDir` and `clonedPoliciesDir`, mirroring the existing `userSavedPolicyDir` pattern.
- Added a `redirectPolicySeams(t)` test helper that points all three seams at fresh temp dirs; applied it to the template-loading tests.

**pkg/tokens** (`collector.go`)
- `NewCollector` no longer eagerly loads `/data/token-summary.json`. The snapshot restore is deferred to first `Summary()`/`scan()` via `sync.Once` (load happens under the write lock, before any scan overwrite, preserving original ordering semantics).
- `defaultPersistPath` is now a documented var seam; `SetPersistPath` now wins regardless of when it is called relative to construction.

**pkg/proxy** (`proxy_deep2_test.go`, `proxy_deep3_test.go`)
- `TestNewGitHubProxy` previously created `/data` if missing and then wrote **and deleted** `CACertPath` — on a writable hive host that overwrites and removes the production MITM CA. It now redirects `CACertPath`/`caKeyPath` to `t.TempDir()` via the existing seam vars.
- `TestNewGitHubProxyWithWritableDir` had dead code hinting at the same intent (`certPath` built but unused, stale "can't override the const" comment). Now actually uses the seam.

## Why

On a hive host with live state, these tests read real agent policies, load the production token summary into `latest` (breaking initial-state assertions), or fail on read-only `/data` — and in the worst case overwrite/delete the live proxy CA. CI stayed green only because runners have no `/data`.

## How tested

- `go build ./pkg/scheduler/... ./pkg/tokens/... ./pkg/proxy/...`
- `go test -short ./pkg/scheduler/... ./pkg/tokens/...` — ok
- `go test ./pkg/proxy/` — ok
- Reproduced the issue conditions on a host that genuinely has live `/data` state: on pristine `v4`, `TestLoadPromptTemplate_FromPoliciesDir` fails because `/data/agents/scanner/CLAUDE.md` shadows the test fixture; with this patch it passes.
- Full-repo `go test ./...`: only pre-existing environment-specific failures remain (identical with and without this patch).

### DIFF
--- a/src/pkg/scheduler/scheduler.go
+++ b/src/pkg/scheduler/scheduler.go
@@ -126,15 +126,27 @@ func (s *Scheduler) GetLastActionable() *github.ActionableResult {
 // always uses the fixed /data/policies path that handleAgentPromptSave writes to.
 var userSavedPolicyDir = "/data/policies"
 
+// agentHomeDir is where per-agent homes live on a hive host; the per-agent
+// CLAUDE.md (<agentHomeDir>/<name>/CLAUDE.md) is checked first by
+// loadPromptTemplate. It is a var (not a const) only so tests can point it at
+// a temp dir; production always uses the fixed /data/agents path.
+var agentHomeDir = "/data/agents"
+
+// clonedPoliciesDir is the root of the git-cloned policies repo checkout on a
+// hive host (…/<root>/examples/kubestellar/agents/). Like userSavedPolicyDir,
+// it is a var only so tests can point it at a temp dir; production always
+// uses /data/policies.
+var clonedPoliciesDir = "/data/policies"
+
 // loadPromptTemplate searches standard paths for an agent's policy template.
 // It checks on-disk paths first, then falls back to embedded default policies.
 func (s *Scheduler) loadPromptTemplate(agentName string) string {
 	paths := []string{
-		fmt.Sprintf("/data/agents/%s/CLAUDE.md", agentName),
+		fmt.Sprintf("%s/%s/CLAUDE.md", agentHomeDir, agentName),
 		// User-saved override from the dashboard prompt editor wins over the
 		// git-cloned examples copy and embedded defaults (#3239).
 		fmt.Sprintf("%s/%s.md", userSavedPolicyDir, agentName),
-		fmt.Sprintf("/data/policies/examples/kubestellar/agents/%s.md", agentName),
+		fmt.Sprintf("%s/examples/kubestellar/agents/%s.md", clonedPoliciesDir, agentName),
 	}
 	if s.cfg.Policies.LocalDir != "" {
 		paths = append(paths,
@@ -163,7 +175,7 @@ func (s *Scheduler) loadNamedTemplate(templateName string) string {
 		// agent has a kick_template set (e.g. quality-advisory.md at ACMM L2) the
 		// edit lands here and must be picked up on the next kick.
 		fmt.Sprintf("%s/%s", userSavedPolicyDir, templateName),
-		fmt.Sprintf("/data/policies/examples/kubestellar/agents/%s", templateName),
+		fmt.Sprintf("%s/examples/kubestellar/agents/%s", clonedPoliciesDir, templateName),
 	}
 	if s.cfg.Policies.LocalDir != "" {
 		paths = append(paths,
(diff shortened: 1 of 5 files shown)


clubanderson:
Reviewed, rebased onto current v4 (bb29b799 — thanks @sidsri14!), and verified locally:

- Rebase note: the pkg/proxy test fixes had already landed on v4 in equivalent form, so that part dropped out; the surviving diff is scheduler seams + tokens lazy snapshot load, all still needed.
- Scheduler: agentHomeDir/clonedPoliciesDir seam vars mirror the existing userSavedPolicyDir pattern; production values unchanged; redirectPolicySeams isolates all three per test with cleanup. Proper injected-dep hermeticity, no skips.
- Tokens: deferring the snapshot restore to first Summary()/scan() via sync.Once is safe — loadSnapshot takes no lock (no deadlock under c.mu), the Once fires under the write lock before the first scan overwrite (ordering preserved), and the single production call site (cmd/hive/main.go NewCollector) never calls SetPersistPath, so behavior is identical in production while tests can finally redirect the path.
- go vet clean; `go test -race -cover`: scheduler 90.4%, tokens 92.7%, all green. DCO signed-off.

/lgtm
/approve

Issue #4596, opened 2026-08-22
🐛 Only one CLI agent can authenticate: /data/home/.claude.json is 0600 per-writer and not reconciled, so other agents show the login menu


Danathar:
## Summary

Interactive-auth CLI agents (`claude`, `copilot`, `codex`, `gemini`) all share one credentials path but run as **different UIDs**, and the CLI writes that file `0600`. Whichever agent an operator logs in first becomes the only authenticated agent. The others sit at the login prompt indefinitely, with no error naming the cause.

On a fresh install this means the operator logs in once, believes the fleet is authenticated, and finds most of it idle.

## Root cause

`AgentHome` (`src/pkg/agent/authprobe.go:78`) returns the **shared** `/data/home` for every non-inference backend, regardless of the agent's UID:

```go
func AgentHome(agentName string, uid int, backend string) string {
	if uid <= 0 { ... }
	if IsInferenceBackend(backend) {
		return inferenceHomePath(agentName)   // per-agent
	}
	return "/data/home"                       // shared
}
```

Only inference backends get a per-agent home. So every claude/copilot/codex/gemini agent launches with `HOME=/data/home` and resolves credentials to the single path `/data/home/.claude/.credentials.json`.

But agents run under **per-agent UIDs** — the same isolation `agentClaudeCredentialPaths` and the `ttyd-tmux.sh` `su-exec` dance exist to accommodate — and Claude Code writes its credentials `0600`. Nothing in the manager reconciles ownership or mode on that file after login.

Result: agent A logs in, the file is `0600 A:node`, and agents B and C get `EACCES`. They cannot read it, and cannot replace it.

## Observed

Standalone Podman/Quadlet, `ghcr.io/kubestellar/hive:stable`, Claude Code v2.1.226, ACMM L2, three agents on `backend: claude`.

Per-agent UIDs, one socket each:

```
$ podman exec hive sh -c 'ls -l /tmp/tmux-*/'
/tmp/tmux-2006/:  srw-rw----. 1 hive-quality    node 0 hive-quality
/tmp/tmux-2007/:  srw-rw----. 1 hive-scanner    node 0 hive-scanner
/tmp/tmux-2010/:  srw-rw----. 1 hive-supervisor node 0 hive-supervisor
```

All three alive and running the CLI:

```
[...]

PR #4606, opened 2026-08-22
🐛 fix(agent): bound the token-triggered restart and diagnose a stuck login (#4596)

Danathar:
## What

Bounds hive's token-triggered restart, and makes the "stuck at a login prompt" state say what is actually wrong.

Addresses #4596. **Diagnostic only** — see *Scope* below for what this deliberately does not change.

## The bug

Hive auto-restarts an agent whose pane shows a login prompt while a valid credential exists in the shared config (`manager.go`, the `showsLogin && loginStreak >= loginStreakRestartMin && configHasTokens()` branch), on the theory that the agent simply has not picked the token up yet. Three guards decide **when** that restart may fire — login streak, kick grace, cooldown. None of them decided **how many times**.

So when something the restart cannot reach holds the agent at the prompt, it retried at the 60-second cooldown forever. That is not benign: restarts destroy in-flight work, which is the exact failure `tokenRestartKickGrace` was added for.

#4596 is precisely that shape. Claude's authentication spans **two** files and hive has only ever probed one:

| File | Holds | Probed by hive |
|---|---|---|
| `$HOME/.claude/.credentials.json` | the OAuth token | yes — `claude.HasValidToken`, and `configHasTokens()` gates the restart on it |
| `$HOME/.claude.json` | session state: `oauthAccount` (who is signed in), `hasCompletedOnboarding` | **no** |

A valid token beside a session file that has lost its `oauthAccount` makes the CLI re-run onboarding and show `Select login method`. Every file-level check hive performs reports success while the operator is looking at a login menu — so `configHasTokens()` stays true, the pane keeps showing a prompt, and the loop never terminates, each relaunch rewriting the same contended file.

That invisibility is the expensive part. The issue was **retracted twice** against this failure, because from outside a valid credential plus a login menu looks like a credential problem.

## Changes

1. **Cap** consecutive token-triggered restarts that fail to clear the prompt at **3**, then stop and record a diagnosis. The counter resets the moment the prompt clears, so an agent that recovers is never permanently barred from a later legitimate nudge.
2. **`diagnoseStuckLogin`** names *both* files, says which one is the problem, and says restarting cannot fix it. Where both files look correct it says the cause is *not* on-disk state instead of guessing; non-claude backends get an honest generic message rather than a claude-shaped one.
3. **`inspectClaudeSession`** — a read-only classifier for `.claude.json` (`absent` / `unreadable` / `no-signed-in-identity` / `signed-in`). An unparseable file reports `unknown`, never "identity missing": this exists to replace guesses, not to add one.

Example of what an operator now gets instead of a silent loop:

> credential at `/data/home/.claude/.credentials.json` is present and valid, but the Claude session state at `/data/home/.claude.json` carries no signed-in identity (`oauthAccount` missing), so the CLI re-runs onboarding and shows the login menu. Restarting cannot fix this — it re-reads the same file. […] See kubestellar/hive#4596

## Scope — what this deliberately does *not* do

- **Does not repair or relax permissions on `.claude.json`.** #4596 measured that making the shared file group-writable converts "one agent cannot log in" into "no agent stays logged in", because every agent then rewrites it. The issue's own earlier suggestion to reconcile it to `0660` was retracted for this reason, and is not implemented here.
- **Does not change `AgentHome`.** Whether interactive-auth backends should get per-agent homes the way `inferenceHomePath` does for inference ones is a fleet-wide behavioural decision — it relocates credentials for every existing deployment and costs one interactive login per agent. The issue explicitly asks to confirm the mechanism before making that change. Left to the maintainers.

This makes the failure legible and stops hive amplifying it, without pre-empting that decision.

## Behavioural surface

The only change to existing behaviour is that the token-triggered restart stops after 3 failed attempts. The give-up path deliberately does **not** `continue`, so the TLS-error and hung-CLI detectors further down the poller stay live — an agent both stuck at a login prompt and hitting a transient network failure is still recovered by the detector that can help. The `wait` (cooldown-not-elapsed) path falls through exactly as before.

The restart decision moved into `AgentProcess.decideTokenRestart` so the rule is unit-testable: the inline version sat behind a real tmux pane capture and could only be exercised in production. Accounting lives with the decision so "fired" and "counted an attempt" cannot drift apart.

## Verification

Go 1.25.6 (matching `go.mod`).

- **New tests pass**; `go vet` clean; `gofmt` clean on all four touched files (`manager.go` was gofmt-clean at baseline and still is).
- **Both halves of the guard are falsified**, not just asserted:
  - constant raised to `1<<30` → `tokenRestartMaxAttempts = 1073741824, want a small positive cap (1..10)`
  - give-up branch deleted with the constant left at 3 → `never gave up across 25 cooldown-separated probes: the restart is unbounded (#4596)`
- An earlier draft of the cap test **falsely passed** with the cap set to `1<<30`, because it derived its loop bound from the very constant under test. It now probes a **fixed** 25 times and pins the constant separately. Flagging it because that failure mode is invisible in a green run.
- Full `./pkg/agent/` suite: two pre-existing failures on this machine, `TestSendKick_RestartsCrashedCLIThenDelivers` and `TestRestartThenSendKick_CleanSlateThenDelivers`. Both reproduce identically on clean `upstream/v4` with the same invocation (same 67.2s timeout), and exactly one of the pair fails depending on which runs — they contend for a shared resource this environment lacks. Not related to this change; verified by stashing and re-running.

Container-build and cross-package coverage left to CI.


clubanderson:
Reviewed at head 7ef4589e. Fact-check:

- **Bound cannot loop**: `decideTokenRestart` increments before firing and latches at 3; verified the cap test pins the constant independently (1..10) AND probes a fixed 25 cooldown-separated cycles — the self-referential-bound trap the PR body flags is genuinely closed. Counter+latch reset together only when the prompt clears, so recovery re-arms a legitimate future nudge.
- **Concurrency**: the new fields are written only by the agent's pane-poller goroutine, same ownership as the pre-existing `lastTokenRestart` beside them; `-race` clean on the new tests. Crucially the PR is **read-only toward the contended file** — `inspectClaudeSession` opens nothing for write, and the decision NOT to reconcile permissions is backed by #4596's measurement that group-writability converts the failure into 'no agent stays logged in'. No mid-write clobber risk because there are no writes.
- **Honest diagnosis**: unparseable → `unknown` (never claimed as skeleton); signed-in + valid cred → 'cause is not on-disk state' instead of a guess; non-claude backends get a generic message, not a claude-shaped one. The give-up path deliberately does not `continue`, so TLS/hung-CLI detectors stay live — verified in the diff.
- **Multi-agent shared-credential case covered**: TestDiagnoseStuckLoginNamesTheSessionStateFile reproduces the exact clobber signature (valid credential + identity-less .claude.json) and asserts both paths, the oauthAccount key, and 'restarting cannot fix' appear.
- Targeted tests + vet pass locally; touched files gofmt-clean (listed gofmt offenders are pre-existing, untouched); trial-merge against v4 head clean; CI green.

Note: no closing keyword — correct, since this is diagnostic hardening; #4596's root cause (shared HOME vs per-agent UIDs) remains open for the AgentHome decision.

/lgtm
/approve

PR #4607, opened 2026-08-22
feat(image): install Google Antigravity CLI for agy backend

hanthor:
## Problem\n\nHive declares `agy` as a valid backend and the rotation ladder includes an Agy/Google fallback, but `src/Dockerfile` does not ship `/usr/local/bin/agy`. The live rotation probe therefore reports `agy-binary-missing` and rejects the rung; a depleted DeepSeek balance can strand an otherwise rotatable agent.\n\n## Fix\n\nInstall Google Antigravity CLI v1.1.19 from its official public Linux artifact. The version/build ID and SHA-512 values are pinned from the vendor updater manifest for both amd64 and arm64. The artifact contains a self-contained `antigravity` executable, installed as `/usr/local/bin/agy`.\n\n## OAuth boundary\n\nAgy requires an interactive Google OAuth login and cannot bootstrap it inside a pod. The image provides the binary only; an operator seeds a pre-authenticated `/home/james/.gemini` session into persistent storage. This enables self-hosted controller deployments while preserving the generic contributor UI's existing host-only warning.\n\n## Verified\n\n- Downloaded official x64 artifact and checked its published SHA-512.\n- Extracted binary runs as `agy --version` after install-mode normalisation.\n- `git diff --check` clean.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -268,6 +268,33 @@ RUN ARCH="${TARGETARCH:-$(dpkg --print-architecture)}" && \
       echo "WARN: Goose CLI download failed — skipping (backend: goose will be unavailable)"; \
     fi
 
+# --- Layer 8.25: Google Antigravity CLI (backend: agy) ---
+# The vendor distributes a self-contained Linux executable, not an npm package.
+# Pin the version, build ID, and SHA-512 from its public updater manifest; do
+# not use the mutable manifest at image-build time. The CLI requires a
+# pre-authenticated Google OAuth session in $HOME/.gemini: the image supplies
+# only the binary, while an operator seeds that persistent state (a pod cannot
+# complete the interactive OAuth flow itself). This enables a self-hosted Hive
+# controller to use its configured agy fallback rung without pretending that
+# generic contributor containers can bootstrap Google login.
+ARG AGY_VERSION=1.1.19
+ARG AGY_BUILD=4894004681244672
+ARG AGY_SHA512_AMD64=7c3b310c80685adcba714994207eb870fb4817403975da46555b7d9fade446487da3d8f897aa220dfbc30f602ab461a0147a6f7b3c8228fdebd35951e3f250fd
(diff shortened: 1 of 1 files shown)


clubanderson:
/lgtm
/approve

Reviewed at aa4d7d2f. External-contribution checks all pass:

- **DCO**: signed-off (hanthor <jreilly1821@gmail.com>) ✓.
- **Pin verified independently**: I downloaded both pinned artifacts (version 1.1.19, build 4894004681244672) and recomputed SHA-512 — amd64 `7c3b310c…f250fd` and arm64 `488c3dac…503558d` both match the Dockerfile exactly. Versioned immutable GCS path, checksum enforced with `sha512sum -c`, unsupported arch hard-fails, tmp artifacts removed, `agy --version` smoke test (same precedent as the claude layer, which already proves the pattern under the multi-arch qemu build).
- **No curl|bash of a moving target** — this is exactly the pinned standard we enforced for caveman (#4518/#4548).
- **Backend wiring confirmed**: `agy` is in CLIBackends (config.go:4371), manager.go launches the `agy` binary with the headless `--dangerously-skip-permissions` contract already in place, rotation.go's usage probe runs `agy` from PATH — /usr/local/bin/agy is precisely what `agy-binary-missing` was checking for. Headless flags live in the manager, correctly not in the image.
- **Size reasonable**: ~56MB (amd64) / ~52MB (arm64) compressed — in line with the other CLI layers.
- **OAuth boundary honestly stated**: image ships the binary only; interactive Google OAuth cannot bootstrap in a pod, operator seeds $HOME/.gemini — consistent with the contribute UI's existing host-only warning.
- **standalone-images.sh untouched and unaffected** (it governs deployment image references, not Dockerfile layers). One note, not a blocker: unlike the goose layer's tolerant WARN-skip, this layer hard-fails the build on a download failure — defensible for a checksummed artifact (a silent skip would silently reintroduce `agy-binary-missing`), and the retry envelope is generous.

Trial merge with latest v4 clean; CI green minus tide.

Merging.

PR #4608, opened 2026-08-22
fix(rotation): fail over high-volume agents after metered exhaustion

hanthor:
## Problem\n\nHigh-volume agents are currently forbidden from every subscription provider. When their current prepaid/metered provider (for example DeepSeek) is positively measured exhausted, all alternatives are rejected and the agent strands/pauses instead of failing over.\n\n## Fix\n\nKeep the normal high-volume subscription guard, but waive it narrowly when the **current provider is metered and a successful probe reports it unavailable**. This enables a DeepSeek agent to rotate to a positively measured healthy Codex/Claude/Agy fallback. Probe errors remain fail-open; subscription-to-subscription exhaustion still retains the guard.\n\n## Test\n\nAdds a 300s-cadence DeepSeek/litellm -> Codex regression test.\n\nVerified with `TMPDIR=/mnt/HC_Volume_106441607/containers-james/go-test-tmp go test ./pkg/rotation`.

### DIFF
--- a/src/pkg/rotation/rotation.go
+++ b/src/pkg/rotation/rotation.go
@@ -386,8 +386,10 @@ func (m *Manager) NextBackend(agentName, currentBackend string) string {
 	return m.nextBackend(agentName, currentBackend, 0)
 }
 
-// NextBackendForCadence is NextBackend with the agent's cadence applied, so
-// high-volume agents are never offered a subscription provider.
+// NextBackendForCadence is NextBackend with the agent's cadence applied.
+// High-volume agents avoid subscription providers during normal operation, but
+// may use one as an availability failover when their current metered provider
+// is positively measured exhausted.
 func (m *Manager) NextBackendForCadence(agentName, currentBackend string, cadenceS int) string {
 	return m.nextBackend(agentName, currentBackend, cadenceS)
 }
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed against the actual rotation code and the runRotationCheck eval loop (rebased onto current v4 as 50dfa742 — clean):

- Failover triggers ONLY on positively measured metered exhaustion: `ProbeErr == nil && !Available` on the CURRENT provider — probe errors stay fail-open and never rotate (matches Exhausted()/ShouldRotate semantics; transient 429s don't feed headroom, balance probes do).
- No thrash: the eval loop only rotates idle agents whose current provider is exhausted, so a failed-over agent stays put until its new provider is itself positively exhausted — implicit hysteresis; it returns to the metered provider once recovered (subscription guard reapplies since the then-current provider isn't metered).
- Invariants preserved: never lands on probe-error/unavailable targets (candidate filter untouched); the pre-existing high-volume subscription guard still strands when the current provider is subscription-class (TestManager_HighVolumeCadence unchanged and passing).
- `go test -race -cover ./pkg/rotation/`: pass, 97.7% coverage. gofmt clean. DCO signed-off.

Minor non-blocking note: when failover is allowed, candidates still rank purely by PctRemaining, so a subscription target can outrank an available metered alternative; with today's single-metered config this is moot — worth revisiting if a second metered provider is added.

/lgtm
/approve

Issue #3528, opened 2026-08-12
🌱 pkg/agent dismissInferencePrompts tmux tests fail locally on macOS (pre-existing, timing-based)


clubanderson:
## Summary

Three `pkg/agent` tests fail reliably on a local macOS developer machine while passing in CI. They are **pre-existing** — not a regression — but they make `go test ./pkg/agent/` red locally, which trains developers to ignore a failing suite.

## Failing tests

- `TestDismissInferencePrompts_FixWith` (~10.5s)
- `TestDismissInferencePrompts_NavigatesNegative` (~30.5s)
- `TestDismissInferencePrompts_BypassConsent` (~10.5s, intermittent — passed on one of two runs)

All fail with the timeout assertion `dismissInferencePrompts should return` (`dismiss_coverage_test.go:77` and siblings).

## Confirmed pre-existing

Reproduced at the commit **before** PR #3523 (`57d23f09~1`): `FixWith` and `NavigatesNegative` fail identically there, so #3523 (pluk dir permissions) did not cause this.

## Root cause

These tests drive a **real tmux server** and coordinate with the code under test purely through `time.Sleep` races:

- gated by `tmuxAvailable()`, so they silently skip where tmux is absent and run where it is present
- the test sends pane text, sleeps 400ms, then a goroutine sleeps 1200-1500ms before sending the ready marker
- the assertion is a 10s (or 30s) `time.After` timeout

On a machine where tmux is present but pane-capture timing differs from the CI image, the ready marker is not observed within the window and the test times out. The suite also takes **~400s** locally, largely due to these sleeps.

## Impact

Not a product bug — `dismissInferencePrompts` is a TUI-prompt dismisser. The cost is developer trust: a permanently-red local suite is a false negative that masks real regressions, which is the mirror image of the false-green problem.

## Suggested fix

Make the tests deterministic instead of timing-based — inject a fake pane-capture source so the "prompt appears then clears" transition is driven by the test rather than by wall-clock sleeps. Failing that, gate them behind an explicit build tag or `HIVE_TMUX_E2E=1` so they run in CI only, rather than skipping-or-flaking based on whether tmux happens to be installed.

clubanderson:
Still reproducing on `origin/v2` after PR #3526 (which also touched `dismiss_coverage_test.go`):

```
--- FAIL: TestDismissInferencePrompts_FixWith (30.47s)
--- FAIL: TestDismissInferencePrompts_NavigatesNegative (30.48s)
[...]

PR #4615, opened 2026-08-23
🐛 agent: per-agent CLAUDE_CONFIG_DIR — claude agents stop wiping each other's login (#4596)

### LINKED ISSUE #4596: 🐛 Only one CLI agent can authenticate: /data/home/.claude.json is 0600 per-writer and not reconciled, so other agents show the login menu

Danathar:
Fixes #4596.

## The bug

Every interactive `claude` agent launches with `HOME=/data/home`, so N agents share **one** `~/.claude.json` — the file where Claude Code keeps its login/onboarding state (`oauthAccount`, `hasCompletedOnboarding`, …). Claude Code rewrites that file **wholesale** (tmp file + rename) from whatever its in-memory view is, and an agent still sitting at the login menu holds the fresh-boot skeleton in memory. Its next flush writes that skeleton over the logged-in state another agent had just established.

Observed on a fresh standalone install (three claude agents, per-agent UIDs): a valid, unexpired, group-readable `.credentials.json` with `subscriptionType: max` — and next to it a `.claude.json` reduced to the 8-key fresh-boot skeleton, ten orphaned `.claude.json.tmp.<pid>.<hash>` rename losers, and every agent back at "Select login method". Full forensics on the issue.

## Mechanism verification (Claude Code v2.1.231, scratch `CLAUDE_CONFIG_DIR`)

Before writing this I verified each behavioral assumption against a real Claude Code binary:

| Experiment | Result |
|---|---|
| Fresh config dir, first boot | CC creates `.claude.json` **inside the dir** — and its key set is exactly the 8-key skeleton observed in the container (plus `changelogLastFetched`). The container's shared file was a fresh-boot rewrite. |
| Valid credential + state **without** `oauthAccount` but **with** `hasCompletedOnboarding` | Comes up **authenticated** ("Claude Max · <account>") and CC self-heals `oauthAccount` back into the state file. The login menu keys on the onboarding state, not the credential. |
| Credential as a **symlink** to a shared file | Comes up authenticated; the symlink survives the run. |
| `CLAUDE_CONFIG_DIR` layout | State at `$DIR/.claude.json`, credential at `$DIR/.credentials.json` (top level — the config dir *replaces* `~/.claude`). |

## The fix — same shape as `CODEX_HOME`

Codex hit this exact class of problem and the repo already contains its fix: a per-agent state dir (`codexHomePrefix = /data/home/.codex-<agent>`), pre-created via `su-exec` as the agent (the manager runs as dev and cannot chown), with a **symlink bridging the shared login** so one sign-in reaches every agent. This PR applies the identical pattern to claude:

- **`CLAUDE_CONFIG_DIR=/data/home/.claude-<agent>`** exported for interactive claude agents with a per-agent UID (`agentEnvPairs`). `HOME` stays shared — only the CLI's own state moves. Inference backends, copilot, codex, and UID-0 agents are untouched.
- **`setupClaudeConfigDir`** (mirrors `setupCodexHome`, called at the same pre-launch point):
  - `.claude.json` — created from the **legacy shared state merged with required seed keys**, so a pre-upgrade login carries forward instead of logging the fleet out (the migration concern raised on the issue for working Docker deployments). If the file already exists it is only *repaired* (missing seed keys filled; existing keys always win) — never replaced. One writer per state file is the entire point.
  - `settings.json` — under `CLAUDE_CONFIG_DIR` this replaces the shared `~/.claude/settings.json` as userSettings, so it is seeded with the consent keys (`skipDangerousModePermissionPrompt` — without it every agent hits the "Bypass Permissions mode" dialog whose default is "No, exit"), starting from the legacy shared settings so operator keys like `effortLevel` carry forward.
  - `.credentials.json` — a **symlink to the shared credential** (`claude.CredentialsPath`), so one login — interactive on any agent, or the dashboard OAuth flow that already writes that path (`ReloadClaudeToken`) — reaches every agent. A *regular* file here (a refresh that replaced the link) is left alone: that agent has simply detached onto its own newer credential, which is still zero-contention.
- **Auth probe** (`agentClaudeCredentialPaths`) checks the per-agent dir first — the file the CLI actually reads/writes — with the shared path kept as fallback, in the same shape as `agentCodexAuthPaths`.
- **Token accounting**: transcripts now land in per-agent `<config dir>/projects`, so the collector scans a glob of those roots in addition to the legacy shared dir. The glob is evaluated per scan pass, so agents added at runtime are picked up without a restart; `MergeAggregates` already dedupes by session ID, so nothing can be double-counted.

### Failure-mode analysis

- If a future Claude Code writes credentials by rename (replacing the symlink) instead of write-through: that agent detaches onto a private credential and keeps working; others keep the shared one. Per-agent logins at worst — still strictly better than the clobber, and `setupClaudeConfigDir` deliberately never re-links over a regular file.
- If seeding fails (missing binaries, exotic perms): warns and continues; the agent then does one interactive login into its own private dir, which sticks. Every step is best-effort and idempotent.
- Explicitly **not** done: relaxing `.claude.json` permissions. Group-writable shared state makes the clobber *worse* (verified live on the issue: the previously-working agent lost its login) — that suggestion is retracted on the thread.

## Composes with #4606

The just-merged #4606 is the *diagnostic* half — it caps the token-triggered restart loop and names the broken session file, deliberately leaving the root cause open ("Diagnostic only"). This PR is the root-cause half, and independently reached the same forensic conclusion (`oauthAccount` missing from the session state beside a valid credential). One integration point needed care: `diagnoseStuckLogin` resolved the session file from `HOME`, which under the per-agent layout is the frozen pre-migration copy — it now inspects the per-agent `CLAUDE_CONFIG_DIR/.claude.json` when it exists, falling back to the legacy location pre-migration. The credential half already flowed through `agentClaudeCredentialPaths` untouched. All of #4606's tests pass unchanged alongside this change.

## Tests

- `claude_config_dir_test.go`: env-pair emission matrix (claude+UID → set; copilot/codex/inference/UID-0 → not), probe path ordering, and functional `setupClaudeConfigDir` coverage through a stubbed `su-exec` — legacy state/settings migration, seed repair without clobbering CLI-owned content, regular-file credential left untouched, UID-0 no-op.
- `collector_claude_glob_test.go`: per-agent transcript roots are scanned and merged, non-matching siblings are not, unset glob is a no-op.
- `go build ./...`, `go vet`, `gofmt` clean; full `pkg/tokens` suite green.
- Full `pkg/agent` suite, honest local report: three full runs on my workstation produced three **disjoint** failure sets, all in real-tmux tests (`TestLaunchInTmux_*` on the unmodified baseline; one or two `kick_restart_recovery` / `agy_launch` pane-timeout flakes on this branch), with the baseline run failing eight tests this branch passed. The machine concurrently runs a live hive container and a tmux session, and local flakiness in exactly this test family is documented in-repo (#3528). The tests covering this change pass deterministically; deferring to CI's hermetic run for the rest. Mechanistically, this change's only contact with the kick path is a fast, failure-tolerant pre-launch setup call and one added env pair — neither can blank a stub CLI's pane for 67 seconds.

## What a running deployment sees after upgrade

Agents come up with private state dirs seeded from the shared files. A fleet with a working login keeps it (state + credential both carry forward). A fleet stuck at login menus does **one** login on any agent; the shared-credential symlink propagates it, and per-agent state means nothing can strip it again.


Danathar:
Superseded by #4619 — closing.

#4619 fixes #4596 by giving every per-UID interactive agent its own `HOME` and routing through `AgentHome()`, which is the better shape: this PR moved only Claude's state via `CLAUDE_CONFIG_DIR`, while #4619 covers copilot/codex/gemini as well and lands the change on the seam the auth probe, `buildAgentEnv`, the caveman installer and #4606's stuck-login diagnosis already call — so those follow automatically instead of each needing a patch.

Checked point by point against `v4`, nothing here is left uncovered:

| This PR | On `v4` after #4619 |
|---|---|
| Per-agent claude state dir | `AgentHome` → `interactiveHomePath`, all interactive backends |
| Auth probe checks the per-agent path | Automatic — the probe already calls `AgentHome` |
| Redirect #4606's diagnosis to the per-agent session file | Automatic — same seam |
| Token-collector glob for per-agent transcripts | Not needed — `.claude` is bridged to `/data/home/.claude`, so `projects/` stays where `ClaudeSessionsDir` points |
| Seed from legacy state so the upgrade does not log the fleet out | #4619 adopts the legacy session file at launch |

It also carries things this PR did not: the orphan sweep for the `.claude.json.tmp.*` debris, the `HIVE_SHARED_AGENT_HOME=1` escape hatch, and bridges for `.copilot`/`.codex`/`.gemini`/`.gitconfig`.

Two findings from this branch that #4619 independently confirmed, recorded in case they are useful later:

- Verified against Claude Code v2.1.231 with scratch config dirs: a valid credential beside a session file **without** `oauthAccount` but **with** `hasCompletedOnboarding` comes up authenticated and self-heals `oauthAccount`. The login menu keys on the session state, not the credential — which is why the credential was never the contended file.
- A fresh boot writes exactly the 8-key skeleton observed on the reporting hive, confirming the shared file had been rewritten rather than merely being unreadable.

Thanks for picking this up — the per-agent `AgentHome` was the right call.

PR #4619, opened 2026-08-23
🐛 fix(agent): per-agent HOME for CLI agents — one login authenticates the fleet

### LINKED ISSUE #4596: 🐛 Only one CLI agent can authenticate: /data/home/.claude.json is 0600 per-writer and not reconciled, so other agents show the login menu

clubanderson:
Fixes #4596

## The defect (measured in #4596)

Every per-UID agent shared `HOME=/data/home`. Claude Code rewrites `$HOME/.claude.json` **wholesale** via atomic write (tmp file + rename), and rename needs only *directory* write permission — the shared home is `2775 dev:node`, so ANY agent could replace the file regardless of its mode. An unauthenticated agent's rewrite stripped `oauthAccount` from under the authenticated one → the whole fleet showed **Select login method**. File permissions can never close a directory-entry race (group-writable measurably made it worse), which is why #4606 shipped diagnosis + a bounded restart rather than a permission fix, and why Danathar deferred the real cure — a per-agent AgentHome — to maintainers. clubanderson green-lit implementing it.

## The fix

Per-UID interactive agents now run with `HOME=/data/home/agents/<name>`, mirroring the existing `inferenceHomePath` / `setupCodexHome` per-agent precedents:

- **`AgentHome()` is the single source of truth** — buildAgentEnv's HOME export, the 🔑 auth probe, the caveman installer and #4606's stuck-login diagnosis all follow automatically (they already call it).
- **`setupInteractiveHome()`** provisions at launch: `mkdirAllNoFollow` (same F12 symlink-safety posture as inference homes), chown-to-UID + 0700 with the `tightenInferenceHome` unprivileged fallback, then **Lstat-guarded symlink bridges** back to shared state: `.claude` (holds `.credentials.json` — the shared OAuth **token**, which was never the contended file; sharing it is what lets ONE login authenticate the fleet), `.copilot`, `.config`, `.codex`, `.bob`, `.gemini`, `.cache`, `.local`, and the entrypoint-written `.gitconfig`/`.bashrc`/`.profile` (git credential-helper wiring + pane shell rc). A real file/dir at a bridge name is never clobbered.
- **`~/.claude.json` becomes truly per-agent**, ADOPTED at launch from a signed-in source — legacy shared file first, then any signed-in sibling — using #4606's `inspectClaudeSession` classifier. Adopt-only: never fabricated (no signed-in source → the login menu is the honest state), and a signed-in per-agent file is never overwritten. Launch-time seeding races nothing: that agent's CLI isn't running yet, and Claude's atomic renames mean sibling reads are never torn.
- **Synergy with #4606**: the capped token-triggered restart becomes *curative* — a stuck agent restarts, re-provisions, adopts the session the operator just signed in on a sibling, and comes back authenticated.
- **Orphan sweep**: legacy `/data/home/.claude.json.tmp.*` debris (ten in one afternoon on the reporting hive) is drained, bounded (cap 200) and best-effort.
- `.bash_history` contention disappears as a side effect (not bridged); `.npm` deliberately un-bridged (per-agent caches end the cross-UID EACCES collisions noted in `installCavemanForAgent`).
- `entrypoint.sh` pre-creates `/data/home/agents` (0755 dev:node); the manager also creates it best-effort for hosted containers.
- **Escape hatch**: `HIVE_SHARED_AGENT_HOME=1` restores the legacy shared layout wholesale (both `AgentHome` and provisioning honor it).

## Migration

**None required.** Credentials stay where they are (`/data/home/.claude/.credentials.json`, watcher-reconciled as before); the first launch adopts the legacy session file; single-agent hives behave exactly as before; rootful and rootless podman both work (chown succeeds where privileged, world-writable fallback where not — the same trade inference homes make).

## Tests

13 new tests in `interactive_home_test.go` via a `sharedAgentHome` test seam (SharedRepoParent convention): AgentHome routing (per-UID / inference / uid-0 / escape hatch), provisioning (bridges created + correct targets, idempotent, stale-link repair, real-dir refusal, skip matrix), seeding matrix (adopt legacy / adopt sibling / never-overwrite-signed-in / re-seed skeleton / no-source-no-fabrication / unparseable-legacy-falls-through), orphan sweep (removes tmp debris, spares the real file). Two existing tests updated to pin the new layout via `AgentHome`/`interactiveHomePath` rather than a literal. `go build ./...`, `go vet`, targeted suite `-race` clean; full `pkg/agent` suite run locally.


clubanderson:
Per-agent HOME implementation for #4596 as green-lit by the maintainer. Verified locally: full `pkg/agent` suite green (469s), targeted new tests `-race` clean, trial-merge against current v4 head (dbacf767) conflict-free. CI green minus tide.

/lgtm
/approve

Issue #2284, opened 2026-07-30
Brainstorm agent unable to write bob settings folder


MattSweetIBM:
When kicking off the brainstorm agent for the first time on a new hive.
```
hive-brainstorm@hive-54d778cdf6-47nwp:/data/agents/brainstorm$ HIVE_AGENT='brain
storm' HIVE_AGENT_DISPLAY_NAME='brainstorm' HIVE_BACKEND='bob' HIVE_MODEL='claud
e-sonnet-4-6' HIVE_ID='hosted-available-oke-07-placeholder-wlj4' HIVE_ACMM_LEVEL
='1' HIVE_AGENT_MODE='ADVISORY' HTTPS_PROXY='http://127.0.0.1:18443' HTTP_PROXY=
'http://127.0.0.1:18443' HIVE_PROXY_AGENT='brainstorm' GIT_TERMINAL_PROMPT='0' N
ODE_EXTRA_CA_CERTS='/data/proxy-ca.pem' HIVE_ADVISORY_ISSUE='1' BOBSHELL_DEFAULT
_AUTH_TYPE='api-key' BD_DIR='/data/beads/brainstorm' GIT_SSL_CAINFO='/data/proxy
-ca.pem' HIVE_AGENT_TOKEN_CACHE='/var/run/hive-metrics/agent-tokens/gh-token-bra
instorm.cache' HOME='/data/home' /usr/local/bin/bob --accept-license --auth-meth
od api-key --approval-mode yolo --trust
Error accessing installation ID file, generating ephemeral ID: Error: EACCES: pe
rmission denied, mkdir '/data/home/.bob'
    at Module.mkdirSync (node:fs:1350:26)
    at tS.writeInstallationIdToFile (file:///usr/local/lib/node_modules/bobshell
/bundle/bob.js:357:14772)
    at tS.getInstallationId (file:///usr/local/lib/node_modules/bobshell/bundle/
bob.js:357:14918)
    at file:///usr/local/lib/node_modules/bobshell/bundle/bob.js:765:6390
    at file:///usr/local/lib/node_modules/bobshell/bundle/bob.js:3:469
    at file:///usr/local/lib/node_modules/bobshell/bundle/bob.js:777:1070
    at file:///usr/local/lib/node_modules/bobshell/bundle/bob.js:3:469
    at file:///usr/local/lib/node_modules/bobshell/bundle/bob.js:777:10199
[...]

PR #4636, opened 2026-08-23
🐛 Run the caveman installer as the agent user; stop it clobbering .claude.json

Danathar:
## What

The caveman installer runs as the hive user with `HOME=` the agent's per-agent home. Caveman spawns the backend's own CLI (`claude plugin list` / `plugin marketplace add` / `plugin install` — caveman `bin/install.js:460-473`), and Claude Code rewrites `$HOME/.claude.json` wholesale on every invocation. Run as a uid that cannot read the agent-owned 0600 session file, the CLI treats the home as a fresh install and replaces the signed-in session with a blank skeleton. The agent drops to the login menu on its next start, and every re-login is destroyed again at the next launch — the #4596 wholesale-rewrite clobber reintroduced from inside the manager, through the per-agent layout #4619 built to end it.

This PR wraps the install in `su-exec` as the agent user whenever the agent has an allocated UID, matching every other command that touches the per-agent HOME (`tmuxCmd`, `setupCodexHome`).

## Observed on a live hive

Rootless podman deployment, three claude agents, current `:stable` (`64dd28e`). One agent (the one that had originally held the fleet's login, so its session files were agent-owned 0600) came back from every restart at the login menu:

- Its per-agent home accumulated `.claude.json.tmp.<pid>.<hash>` debris **owned by the hive uid** — Claude Code's atomic-write signature, from a process that should never write that file.
- Each clobber timestamped exactly at an `installing caveman` log line for that agent (three token-triggered restarts → three clobbers at 11:05, 11:10, 11:11; container restart → another at 11:19).
- Three operator re-seeds of `oauthAccount`/`hasCompletedOnboarding` into the file were each destroyed within seconds of the next agent launch. The file's `firstStartTime` was reset to the boot time on every clobber — a fresh-install write, not a merge.
- Agents whose session files had gone through `seedClaudeSessionForAgent` (written 0666, readable by the hive uid) survived every restart: the spawned CLI could read them and merged instead of resetting. The failure needs exactly one condition: a session file the hive process cannot read, which is what the agent's own CLI produces on every rewrite.

Making the file readable to the hive uid (mode 0666) on the live hive stopped the clobber immediately and the agent authenticated on the next start — confirming the read-failure → fresh-install-rewrite mechanism.

## Changes

- `installCavemanForAgent`: build the argv via a new `cavemanInstallArgv(backend, mode, userSpec)` and prefix it with `su-exec <agent-user>` when `agent.UID > 0`. UID-0 (shared-home) deployments are unchanged — no su-exec, same argv as before.
- npm cache: the per-agent `.npm-caveman-cache` was written by the hive uid until now, and npm run as the agent uid EACCESes on foreign-owned cacache shards (same failure class as #2284). `cavemanNpmCachePath` removes a foreign-owned cache (the hive uid owns the old one, so it can), falling back to a per-UID path when removal fails.
- `diagnoseStuckLogin`, unreadable-session case: the message claimed "the CLI shows the login menu because it cannot load the signed-in identity" — but *unreadable* is measured from the hive process, and under per-agent homes an agent-owned 0600 session file is the **normal** state of a healthy signed-in agent. That wording sent the live investigation above down a permissions rabbit hole on a file the agent read fine. It now states the hive-process perspective explicitly and stops asserting what the agent's own CLI can or cannot read.

## Tests

- `TestCavemanInstallArgvRunsAsAgentUser` — every supported backend's per-UID install is su-exec-wrapped.
- `TestCavemanInstallArgvSharedUIDStaysDirect` / `TestCavemanInstallArgvUnsupportedBackend` — UID-0 stays direct; unknown backends return nil.
- `TestCavemanNpmCachePathReplacesForeignOwnedCache` / `KeepsOwnCache` / `FallsBackWhenRemovalFails` — the cache migration matrix.
- `TestDiagnoseStuckLoginUnreadableDoesNotBlameTheAgent` — the diagnosis names the hive-process perspective and no longer claims the agent cannot load its identity.

Existing caveman guard tests (`TestCavemanPinIsNotBareSHA`, `TestCavemanSkillsAddIsHeadless`) still pass — the refactor keeps the refs and `skills add -y` shapes in `manager.go` where they scan.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5, high reasoning effort.

### DIFF
--- a/src/pkg/agent/caveman_install_test.go
+++ b/src/pkg/agent/caveman_install_test.go
@@ -1,0 +1,117 @@
+package agent
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestCavemanInstallArgvRunsAsAgentUser pins the fix for the caveman clobber:
+// the installer spawns the backend's own CLI (`claude plugin list` / `plugin
+// install`), and Claude Code rewrites $HOME/.claude.json wholesale on every
+// invocation. Run as the hive user with the agent's HOME, that CLI cannot
+// read the agent-owned 0600 session file, treats the home as a fresh
+// install, and replaces the signed-in session with a blank skeleton — the
+// agent drops to the login menu on its next start (#4596 reintroduced from
+// inside the manager, through the per-agent layout #4619 built to end it).
+// Every per-UID install must therefore be wrapped in su-exec, exactly like
+// tmuxCmd and setupCodexHome.
+func TestCavemanInstallArgvRunsAsAgentUser(t *testing.T) {
+	backends := []string{"claude", "copilot", "gemini", "goose", "codex", "aider"}
+	for _, backend := range backends {
+		argv := cavemanInstallArgv(backend, "full", "hive-scanner")
+		if len(argv) < 3 || argv[0] != "su-exec" || argv[1] != "hive-scanner" {
+			t.Errorf("backend %s: per-UID install must run via su-exec as the agent user, got %v", backend, argv)
+			continue
+		}
+		if argv[2] != "npx" {
+			t.Errorf("backend %s: expected npx after the su-exec prefix, got %v", backend, argv)
+		}
+	}
+}
+
+// TestCavemanInstallArgvSharedUIDStaysDirect pins the UID==0 path: agents
+// without an allocated UID run as the hive user, so there is no user to
+// switch to and no su-exec prefix (su-exec may not even exist on such
+// deployments).
+func TestCavemanInstallArgvSharedUIDStaysDirect(t *testing.T) {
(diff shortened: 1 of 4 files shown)


clubanderson:
Reviewed and verified: su-exec wrapping matches the tmuxCmd/setupCodexHome precedent exactly (agent.UID>0 only; UID-0 stays direct — pinned by test), the argv seam keeps the caveman pin/headless guard tests scanning, the foreign-owned npm-cache migration matrix is covered including the unremovable fallback, and the unreadable-session diagnosis rewording honestly states the hive-process perspective. Trial-merged together with #4641 onto current v4: builds, vets, and the union test set passes with -race. CI green minus tide.

/lgtm
/approve

PR #4641, opened 2026-08-23
🐛 fix(agent): adopt agent-owned Claude sessions via su-exec so one login authenticates a fresh fleet

### LINKED ISSUE #4637: Fresh installs need one Claude login per agent — docs promise "log in once per method" (#4619 seeding cannot read agent-owned sessions)
## Summary

`docs/agent-configuration.md` promises that CLI-backend logins are per **method**:

- The backend table lists claude as *"log in once per method"* (line ~307)
- Line ~318: *"**CLI methods are subscriptions.** You log in once per method from the dashboard, and every agent using that method shares the login."*

Since #4619 (per-agent interactive homes), a **fresh install needs one login per agent** for the claude backend. The docs and the behavior have drifted apart, and which one is right is a design decision — hence an issue rather than a PR.

## Why adoption can't propagate a fresh login

#4619's design intends the once-per-method UX to survive: `seedClaudeSessionForAgent` adopts a signed-in `.claude.json` from the legacy shared home or a sibling, *"so existing hives migrate without any operator step"* (`interactive_home.go` header).

That works for migrating hives (the legacy shared `/data/home/.claude.json` is readable by the hive process). It cannot work on a fresh install:

1. Operator logs in on agent A through its terminal.
2. Agent A's CLI rewrites `/data/home/agents/A/.claude.json` — owned by A's UID, mode 0600.
3. Agent B sits at the login menu; the shared credential is valid, so the #4606 token-triggered restart fires and seeding runs.
4. `findSignedInClaudeSession` checks the legacy shared file (absent or skeleton on a fresh install), then siblings — and `inspectClaudeSession` on A's file returns `claudeSessionUnreadable`, because the hive process cannot read an agent-owned 0600 file. No source found; agent B stays at the login menu.
5. After 3 attempts the restart gives up, and the operator logs in on B (and C, …) by hand.

Observed on a live rootless podman hive (3 claude agents): the 🔑 badge clears strictly per agent; each agent needed its own interactive login.

## Two possible resolutions

**A. Make the docs true again (code):** let the seeding path read agent-owned session files the way every other cross-UID operation already works — via `su-exec` as the owning agent user (the same mechanism as `tmuxCmd`, `setupCodexHome`, and #4636). Adoption then propagates a fresh login to the whole fleet, and one login per method is real on fresh installs too.

**B. Make the behavior documented (docs):** change the claude row to "log in once per agent" and explain the split — the OAuth token (`.claude/.credentials.json`) is shared through the home bridge, the session identity (`.claude.json`) is per agent.

Happy to PR either — (A) pairs naturally with #4636, which is the same defect family (the hive process and agent UIDs disagreeing about who can read `.claude.json`).

Related: #4596, #4619, #4606, #4636

Danathar:
## What

Closes #4637 by taking its **option A**: make the "log in once per method" promise in `docs/agent-configuration.md` true on fresh installs, rather than downgrading the docs to "once per agent".

Option A is the one that matches stated design intent — #4619's title is literally *"per-agent HOME for CLI agents — one login authenticates the fleet"*, and `interactive_home.go`'s header says the per-agent seeding exists "so existing hives migrate without any operator step". The seeding path just never worked on a hive that had no legacy shared file to inherit from.

## Why fresh installs need one login per agent today

`seedClaudeSessionForAgent` adopts a signed-in `.claude.json` from the legacy shared home or a sibling. On a **migrating** hive that works: the source is `/data/home/.claude.json`, which the hive process can read.

On a **fresh** install there is no legacy file. The only signed-in session in existence is the one the operator's login made agent A's **own CLI** write — owned by A's UID at mode 0600. `inspectClaudeSession` on it returns `claudeSessionUnreadable`, `findSignedInClaudeSession` reports no source, and agent B stays at the login menu until the operator logs in by hand. Same for C, D, …

## Changes

- **`claude_session_adopt.go` (new)** — an owner-aware read/write seam. A direct read that hits `EACCES` is retried as the file's **owner**; a direct write that hits `EACCES` is performed **by the target agent**. This is the same `su-exec` mechanism every other cross-UID operation in the package already uses (`tmuxCmd`, `setupCodexHome`, the tmux-dir chmod).
- **Adopt-only is now actually enforced for unreadable targets.** `seedClaudeSessionForAgent` classifies its *target* owner-aware too. An unreadable target may be the agent's **own** signed-in session; previously it read as "not signed in" and got overwritten with a sibling's identity.
- **`classifyClaudeSession`** split out of `inspectClaudeSession`, so bytes obtained through the seam reach an identical verdict to a direct read.
- **Docs** — one sentence noting that `claude` sharing is not instantaneous: the token is shared at once through the home bridge, the session identity is adopted at each other agent's next launch/restart.

## Safety posture

- Candidates are `Lstat`'d (never followed) and must be a **regular file**, under an **8 MiB cap**, and owned by a **non-root uid** before any helper runs — a symlink or a root-owned file planted at a sibling path can never turn adoption into "run something as uid 0".
- The write fallback stages through a temp file in the target's own directory and renames, so a helper dying mid-write cannot leave the agent holding a truncated `.claude.json`. The target path is passed as `$0`, never interpolated into the shell script.
- Every helper is **best-effort** and falls back to the direct result. Without `su-exec`, unreadable stays unreadable and the login menu remains the honest state — nothing is fabricated, and no caller treats a failed fallback as evidence of anything.

## Interaction with #4636

They are complementary halves of the same defect family, and I verified they do not collide:

- **No file overlap.** #4636 touches `manager.go` + `stuck_login_diagnosis.go` (+ tests); this touches `interactive_home.go`, `claude_session_state.go`, `claude_session_adopt.go`, and the docs. I test-merged this branch with `pull/4636/head`: it merges cleanly, builds, and both PRs' test sets pass together.
- **One real semantic coupling, handled here.** #4636 observed that agents whose session files had gone through `seedClaudeSessionForAgent` (written at `claudeSessionSeedFileMode`, readable by the hive uid) survived every caveman clobber, while agents holding an agent-written 0600 file did not. So this PR's `su-exec` write deliberately `chmod`s to `claudeSessionSeedFileMode` rather than letting the agent's umask produce 0600 — otherwise adoption would hand back exactly the shape #4636 is fixing. Both write paths land on the same mode by construction, with a comment saying why they must not drift.
- **Merge order does not matter.** Neither depends on the other; each is an improvement on its own.

## Tests

New in `claude_session_adopt_test.go`. "Unreadable to the hive process" is simulated with mode 0000 (unit tests cannot create files owned by a second UID), and a `runAsAgentUser` test seam stands in for the privileged helper.

- `TestSeedClaudeSession_AdoptsAgentOwnedSibling` — the #4637 case itself: no legacy shared file, one agent-owned 0600 signed-in sibling, and the fresh login propagates.
- `TestSeedClaudeSession_UnreadableSignedInTargetIsNotClobbered` — adopt-only holds when the target cannot be read directly.
- `TestSeedClaudeSession_SeedIsReadableByTheHiveProcess` — guards the #4636 coupling above.
- `TestSeedClaudeSession_HelperFailureLeavesLoginMenuHonest` — with no working helper, nothing is fabricated.
- `TestClaudeSessionOwnerSpec_RefusesNonRegularAndOversize` / `AcceptsOrdinarySessionFile` — the symlink / cap / missing-file refusals.
- `TestInspectClaudeSessionForAdoption_FallsBackToOwnerRead`, `TestRunAsAgentUser_RejectsEmptyInput`.

All pre-existing `interactive_home` / `claude_session` / `stuck_login` tests still pass unchanged.

`go build ./...` and `go vet ./pkg/agent/` are clean, and the files touched here are `gofmt`-clean.

**One pre-existing failure, not from this change:** `TestRestartThenSendKick_CleanSlateThenDelivers` fails in my environment (`agent worker CLI not ready after restart`, a tmux timing test). I confirmed it fails identically on unmodified `upstream/v4` with my changes stashed.

### DIFF
--- a/src/docs/agent-configuration.md
+++ b/src/docs/agent-configuration.md
@@ -315,7 +315,7 @@ Implementation notes:
 
 Two rules of thumb:
 
-- **CLI methods are subscriptions.** You log in once per method from the dashboard, and every agent using that method shares the login.
+- **CLI methods are subscriptions.** You log in once per method from the dashboard, and every agent using that method shares the login. For `claude`, sharing is not instantaneous: the OAuth token is shared immediately through the per-agent home bridge, while the session identity (`~/.claude.json`, which is what decides whether the CLI shows a login menu) is adopted from an already-signed-in agent the next time each other agent launches or is restarted. So on a fresh install, expect the remaining agents to clear their 🔑 badge on their next start rather than the moment you finish logging in.
 - **Inference methods are endpoints.** You configure a base URL and a key *reference* (env var name or key-file path — the value goes in `/data/secrets/`, never in YAML). Agents on `vllm`/`llm-d`/`litellm` launch the Claude CLI routed through hive's inference translator, so there is no separate login.
 
 Every Model Gateway (and the bob backend) also accepts an optional `key_name` — a human-chosen LABEL for the configured key, e.g. `key_name: openrouter-prod-key`. It is safe-to-show metadata, not a secret: the dashboard's gateway row displays it as "Using key: `<name>`", or "(unnamed)" when no label is set, so operators can tell keys apart without ever seeing the value. See [`inference-backends.md`](../../docs/inference-backends.md) for a full YAML example.
--- a/src/pkg/agent/claude_session_adopt.go
+++ b/src/pkg/agent/claude_session_adopt.go
@@ -1,0 +1,204 @@
+package agent
+
+// Cross-UID reads and writes of .claude.json for session adoption (#4637).
+//
+// THE DEFECT: #4619 gave every per-UID interactive agent its own HOME and
+// seeded a signed-in .claude.json into it, "so existing hives migrate without
+// any operator step". That works for a MIGRATING hive, where the signed-in
+// source is the legacy /data/home/.claude.json the hive process itself can
+// read. It cannot work on a FRESH install: the only signed-in file in
+// existence is the one agent A's own CLI just wrote, owned by A's UID at mode
+// 0600. inspectClaudeSession on that file returns claudeSessionUnreadable, so
+// findSignedInClaudeSession sees no source, agent B stays at the login menu,
+// and the operator logs in once per agent — while the docs promise "log in
+// once per method" (docs/agent-configuration.md).
+//
+// THE FIX: stop asking the hive process to read what only an agent UID can
+// read. Every other cross-UID operation in this package already goes through
+// su-exec — tmuxCmd, setupCodexHome, the tmux-dir chmod — and this is the same
+// shape. When a direct read hits EACCES, re-read as the file's OWNER; when a
+// direct write hits EACCES, write as the TARGET agent. Adoption then
+// propagates a fresh login across the fleet and the documented contract holds
+// on fresh installs too.
+//
+// POSTURE: every helper here is best-effort and fails back to the direct
+// result. su-exec is absent in unit tests and on developer laptops, unreadable
+// stays unreadable when the fallback cannot run, and no caller may treat a
+// failed fallback as evidence of anything. Files are Lstat'd (never followed)
+// and must be regular, under the size cap, and owned by a NON-ROOT uid before
+// a helper runs — a symlink or a root-owned file at a sibling path must never
+// be able to turn this into "run something as uid 0".
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"os/exec"
+	"syscall"
+	"time"
+)
+
+// claudeSessionAdoptMaxBytes caps how much of a .claude.json this will carry
+// across UIDs. Claude Code accumulates per-project history in the file, so it
+// grows without a natural bound; a few MiB is far above any observed session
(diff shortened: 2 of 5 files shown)


clubanderson:
Reviewed and verified against the #4619 design (mine): the owner-aware read/write seam is exactly the option-A shape the issue thread recommends — su-exec as the file's owner, matching tmuxCmd/setupCodexHome. Security posture checked: claudeSessionOwnerSpec fails CLOSED (Lstat never follows, regular-file only, 8 MiB cap re-checked post-read, root-owned refused so no helper can ever run as uid 0); the write fallback stages+renames in the target's own dir with the path as $0, never interpolated; every helper is best-effort so no-su-exec deployments keep the honest login menu. The owner-aware TARGET inspection closes a real clobber I hadn't covered — an unreadable-but-signed-in own session previously read as not-signed-in. The deliberate claudeSessionSeedFileMode on the su-exec write path is the right call given #4636's measured clobber condition (just merged, d9c4627e). Trial-merged with #4636 onto current v4: builds, vets, union tests pass with -race. Docs sentence matches the actual adoption timing. CI green minus tide.

/lgtm
/approve

PR #4645, opened 2026-08-23
fix(rotation): probe claude via OAuth API and codex via app-server JSON-RPC

hanthor:
## Problem

Both CLI-based headroom probes target output the CLIs no longer produce, so anthropic and openai headroom are **permanently unknown** (every probe fails open):

- `claude /usage --output-format text` — current builds render **session stats only** (Total cost / duration / token counts); the `Current week (all models) N% used` block the regex matches no longer exists.
- `codex /status --output-format text` — current codex **rejects `--output-format`** (`unexpected argument`) and `/status` is TUI-only (`stdin is not a terminal`).

## Fix

Use the same sources the CLIs' own HUDs use:

- **Claude**: `GET https://api.anthropic.com/api/oauth/usage` with the access token from `~/.claude/.credentials.json` (`.claudeAiOauth.accessToken`) and the `anthropic-beta: oauth-2025-04-20` header (required, else 401). The response's `limits[]` array carries per-kind `percent` + `resets_at` (session / weekly_all / weekly_scoped); the binding limit is the max percent. Fail-open on 401/429/parse/empty-token — an expired OAuth session is `Login expired`, not exhaustion.
- **Codex**: spawn `codex app-server`, exchange the `initialize` handshake, then call `account/rateLimits/read`. `rateLimits.primary` carries `usedPercent` + `resetsAt`. Fail-open on any process/protocol failure (e.g. sandbox EPERM on restricted hosts).

## Tests

- Claude: httptest server verifying the Authorization + anthropic-beta headers; used/exhausted/401/empty-token/missing-credentials paths.
- Codex: a fake `codex` script speaking the JSON-RPC protocol (handshake then rate-limits reply); exhausted and fail-open paths.

Verified locally with `TMPDIR=... go test ./pkg/rotation` and `go build ./...`.

### DIFF
--- a/src/pkg/rotation/rotation.go
+++ b/src/pkg/rotation/rotation.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/kubestellar/hive/pkg/claude"
 	"github.com/kubestellar/hive/pkg/config"
 )
 
@@ -130,6 +131,11 @@ type ClaudeProber struct {
 const claudeUsageBaseURL = "https://api.anthropic.com"
 const claudeOAuthBeta = "oauth-2025-04-20"
 
+// sharedCLIHome is the durable shared CLI home on the PVC — the HOME the
+// manager gives agent tmux sessions and where fleet-level CLI auth state
+// (.claude, .codex, .copilot) persists across pod restarts.
+const sharedCLIHome = "/data/home"
+
 type claudeCredentials struct {
 	ClaudeAiOauth struct {
 		AccessToken string `json:"accessToken"`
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed and verified locally.

**Findings:**
- Probes now measure real auth/availability: Claude via `GET api.anthropic.com/api/oauth/usage` (OAuth bearer + `anthropic-beta: oauth-2025-04-20`), Codex via `codex app-server` JSON-RPC (`account/rateLimits/read`). Binding limit = max `usedPercent` across limits — correct.
- **Failover safety (#4608 invariants):** every failure class (network error, HTTP non-200, malformed body, missing creds, JSON-RPC error) fails OPEN — `ProbeErr` set, `Available=true`. Since `runRotationCheck` fails over only on `ProbeErr==nil && !Available`, transient errors cannot cause false failovers. ✅
- **Timeouts bounded:** 10s HTTP client; codex context 15s with process kill — a hung endpoint can't stall the rotation loop.
- **No credential material leaked** into logs or error strings (checked all error paths).

**Fix-forward (11deecaf, DCO signed):**
1. ClaudeProber default creds path: hive main process runs with `HOME=/home/dev` and entrypoint bridges `.copilot` but not `.claude` — added stat-fallback to `claude.CredentialsPath` (`/data/home/.claude/.credentials.json`).
2. CodexProber sets `HOME=/data/home` for the app-server child when the shared CLI home exists (codex auth lives there).
3. `cmd.Process.Kill()` without `Wait()` leaked a zombie per probe cycle — now reaped.
4. Coverage 89.0% → 92.4% (floor 90) with 7 outcome-class tests (default-home creds, corrupt creds, malformed usage body, percent-less limits, app-server error, invalid rate-limit parses).

DCO present on both commits. Rebase onto v4 clean, trial-merge vs latest v4 clean, CI green.

/lgtm
/approve

PR #4647, opened 2026-08-23
fix(deploy): bootstrap scripts clone v4, not retired v2

### LINKED ISSUE #4646: Bootstrap scripts still clone the retired v2 branch — installs silently provision v2, not v4
## Summary

Both bootstrap scripts still clone the **retired `v2` branch**. Because `v2` still exists on origin, they do not fail — they succeed and silently provision a months-old codebase.

| File | Line | Code |
|---|---|---|
| `src/deploy/bootstrap-lxc.sh` | 57 | `git clone --branch v2 --single-branch https://github.com/kubestellar/hive.git "${HIVE_DIR}"` |
| `bin/hive-setup.sh` | 136 | `run "cd ${HIVE_DIR} && git pull --rebase origin v2"` |
| `bin/hive-setup.sh` | 138 | `run "git clone --branch v2 --single-branch ${HIVE_REPO_URL} ${HIVE_DIR}"` |

## Impact

An operator following the LXC or Docker bootstrap path today gets **v2**, not v4. There is no error and no warning: the clone succeeds, the container starts, and the deployment runs retired code with none of the v4 security work (IPv6 egress gate, MITM CA key handling, setuid stripping, `mint` ServiceAccount verification, ioscan).

An upgrade path is also affected — `hive-setup.sh:136` pulls `origin v2` on re-run, so re-running the installer keeps a host pinned to the retired line.

## Also stale (cosmetic, same files)

- `bin/hive-setup.sh:2` and `:100` — header prints "Hive v2 instance (Docker-based)"
- `bin/hive-setup.sh:182` — generated config comment says "Hive v2 config for ..."

These are quoted verbatim in `src/docs/deployment-scripts.md:37`, so the doc is currently **accurate**. Fixing the scripts requires updating that doc in the same change, or the quotation stops matching its source.

## Suggested fix

Target `v4`, and prefer not to hardcode a branch at all — a `HIVE_BRANCH` variable defaulting to `v4` would let the same script serve release lines without another sunset leaving it stranded. Consider whether the scripts should refuse to run against a retired branch rather than silently proceeding.

## Provenance

Found while retiring stale `v2` references from the docs (#4643 → PR #4644). The docs side deliberately did **not** claim these scripts target v4, since that would have made the documentation lie to cover a broken script; PR #4644 records it as a known defect and defers the code fix here.

Danathar:
## Summary

`bootstrap-lxc.sh` and `hive-setup.sh` both hardcoded `--branch v2` (and `git pull --rebase origin v2`) for cloning the Hive repo. Because `v2` still exists on `origin`, these commands never fail — they succeed and silently provision the retired v2 codebase instead of v4, missing all v4 security work (IPv6 egress gate, MITM CA key handling, setuid stripping, `mint` ServiceAccount verification, ioscan). Re-running `hive-setup.sh` on an existing install pulled `origin v2` too, keeping upgrades pinned to the retired line.

## Changes

- Both scripts now clone/pull a `HIVE_BRANCH` variable, defaulting to `v4`:
  - `bootstrap-lxc.sh`: `HIVE_BRANCH="${HIVE_BRANCH:-v4}"`, overridable via environment.
  - `hive-setup.sh`: same default, plus a new `--branch BRANCH` flag.
- Dropped the stale "v2" wording from banners/comments in both scripts (header text, generated `.env`/`hive.yaml` comments, completion messages) so nothing implies a fixed version line.
- Updated `src/docs/deployment-scripts.md`, which quotes `hive-setup.sh`'s header and `bootstrap-lxc.sh`'s clone target verbatim, so the quotes keep matching their source.

## Why a variable instead of a hardcoded `v4`

Per the issue's suggestion: hardcoding `v4` would leave these scripts stranded again the next time a release line is retired. `HIVE_BRANCH` (default `v4`) lets the same scripts serve future release lines via override, the same way `HIVE_REPO_URL` already does for the remote.

Fixes #4646


clubanderson:
Reviewed, rebased onto current v4 (the docs conflict with the interim "Known defect" note is resolved — note removed since the defect is now fixed), and swept the whole repo for v2-as-install-source references. Fixed forward two sites the PR missed plus a regression guard (a40309bc):

- `bin/hive-prereq-check.sh` — its try_fix still cloned `--branch v2`; now `${HIVE_BRANCH:-v4}` like the other scripts.
- `bin/hive-podman-setup.sh` — stale header comment claiming hive-setup.sh clones v2.
- New `TestInstallScriptsNeverCloneRetiredV2` (src/pkg/config) scans all three install-path scripts for retired-branch refs in clone/pull/checkout positions; mutation-checked red on the pre-fix tree. Deliberately not a bare-v2 scan: v2-ci.yml/v2-tests.yml filenames and the "v2 Tests" check name are historical, keyed by branch protection, and untouched.
- `docs/migration-v1-v2.md`'s `clone -b v2` left intentionally — historical migration doc.

`bash -n` clean on all touched scripts; gofmt clean; guard test green. DCO on both commits.

/lgtm
/approve

PR #4659, opened 2026-08-23
🌱 test(agent): stop TestConcurrentPauseResume launching real CLIs — suite no longer blows go test's 600s timeout (#4628)

### LINKED ISSUE #4628: [quality] pkg/agent suite exceeds go test's 600s default timeout on live hive hosts — tmux-exec-bound tests
## Finding

On a live hive host, `go test ./pkg/agent/ -count=1` panics with `test timed out after 10m0s` (the Go default). It is not one deadlocked test: the timeout lands on whichever tmux-exec-heavy test happens to be running when the budget runs out — observed on `TestRunCopilotDiagnostic_CLIReady` (120s run) and `TestConcurrentPauseResume_NoPanic` (420s run, after 560 tests had already passed with 0 failures).

Two structural causes:

1. **Real tmux execs against the live shared tmux server.** `TestConcurrentPauseResume_NoPanic` runs 5 goroutines × 50 `Pause`/`Resume` cycles, and `Manager.Pause` calls `tmuxSendKeysForAgent` (manager.go ~7975), i.e. ~500 real `tmux` process spawns per run. On a loaded host with live agent sessions each exec is slow, and the test both takes minutes and interacts with the production tmux server — the same hermeticity class as #4585/#4622.
2. **Many wall-clock-bound diagnostic tests** (`TestRunCopilotDiagnostic_*` at ~3–12s each, etc.) push the whole package past 600s under load, so the suite fails even when every assertion passes.

Coverage collection is also affected: the timeout aborts profile writing, which is how `stuck_login_diagnosis.go` showed a false 0% in a full-repo coverage run despite having a 347-line test file.

## Recommendation

- Give `TestConcurrentPauseResume_NoPanic` a tmux seam (stub `tmuxSendKeysForAgent`/the tmux binary, as `overrideStub` does elsewhere) or drastically reduce the iteration count — the race it pins is in the mutex accounting, not in tmux.
- Audit `pkg/agent` tests that exec the real tmux binary against the default socket for live-host interaction (they can send keys to real `hive-<agent>` sessions if names collide).
- Consider `-timeout` headroom and/or `t.Parallel()` for the wall-clock-bound diagnostic tests in CI.

Companion PR #4627 already fixes the two outright live-host FAILures in the same package (`TestAgentAuthState_*`), so after this issue the package would be green and bounded on live hosts.

## Priority
- Impact: medium (blocks `go test ./pkg/agent` on live hosts; false coverage data; touches production tmux server)
- Effort: medium

---
---

Danathar:
## Summary

Fixes #4628 — `go test ./pkg/agent/ -count=1` blowing go test's 600s default timeout on live hive hosts.

The dominant cost was a single test. `TestConcurrentPauseResume_NoPanic` pins the pause/resume **mutex accounting** under concurrency (5 goroutines × 50 cycles), but its agents were plain non-sandbox configs, so every `Resume` of a `StatePaused` agent took the full relaunch path — `ensureTmuxSession` + `launchInTmux`: a **real tmux session create plus a real CLI launch per cycle**, ~250 launches per run, none of them cleaned up. That is exactly the tmux-exec storm the issue describes, and it reproduces off live hosts too: on an otherwise idle dev box, the moment `HIVE_WORK_DIR` points somewhere writable (as it does on every live hive host), the test alone takes **223.57s**; under live-host load that becomes the observed 420s and the package blows the 600s budget, aborting coverage profile writes (the false 0% on `stuck_login_diagnosis.go`).

## What changed

**1. `TestConcurrentPauseResume_NoPanic` is now exec-free** (`manager_coverage2_test.go`)

The test's agents are now sandbox-enabled (`AgentSandboxOverride` + `SetSandboxConfig`), which routes `Pause`/`Resume` through the sandbox branch: every cycle is a pure in-memory state transition — precisely the accounting the test exists to hammer, per the issue's own triage ("the race it pins is in the mutex accounting, not in tmux").

- Measured: **223.57s → 0.01s** (same box, writable `HIVE_WORK_DIR`, `-race` clean).
- No coverage lost on the real relaunch path: `relaunch_race_guard_test.go` already stresses concurrent `Resume` racing through the actual unlock/mint/relock dance, with stubs and cleanup.
- Side benefit: the test no longer reaches `launchInTmux`'s `reapAgentCLI` (which kills processes matched by `HIVE_AGENT` env) or leaves an uncleaned `hive-a` session plus ~250 launches' worth of leaked poll goroutines exec'ing tmux for the remainder of the suite.

**2. Fixed a tmux server mismatch hiding a coverage gap** (`launch_coverage_test.go`)

This is the issue's bullet 2 (audit tests that exec the real tmux binary). The three adopt-running-CLI tests (`TestStart_CLIAlreadyRunning`, `TestStart_InferenceCLIAlreadyRunning`, `TestStart_CopilotAdoptsRunning`) pre-created their marker sessions with **bare `exec.Command("tmux", ...)`** — the default-socket server — while the manager inspects the package test server (`-L defaultTmuxSocket`, per-pid, set by `TestMain`). The pre-created session was invisible to the manager, so `ensureTmuxSession` created a fresh empty pane and each test silently did a **full launch** instead of exercising the adopt branch it documents.

Verified with `-coverprofile` both ways: the CLI-already-running block in `manager.go` (the adopt body, including the trust-prompt and inference-dismiss sub-branches) had **0 hits** from these tests before, full hits after. They now use `testTmuxCommand`.

**3. Audit results (no further changes needed)**

- Every other tmux exec in `pkg/agent` tests goes through `testTmuxCommand` or the manager's socket-aware helpers.
- `TestMain` already isolates the socket name (per-pid `-L ht<pid>`) **and** `TMUX_TMPDIR`, so package tests cannot reach the live server's `hive-<agent>` sessions even via bare execs — the live-host interaction risk is load (CPU from launch storms), which item 1 removes.
- su-exec paths in tests are stubbed (`relaunch_mint_test.go`) or assert argv without running (`branches_coverage_test.go`).
- On `-timeout` headroom / `t.Parallel()` (issue bullet 3): CI runs the package in 5 shards each far below the budget, and with item 1 the serial suite fits comfortably inside go test's 600s default (see timing below), so no CI workflow change is needed. `t.Parallel()` was deliberately avoided: the wall-clock diagnostic tests share the global stub binaries via `overrideStub`, which is not parallel-safe.

## Verification

Both trees run with identical live-host-style conditions (ambient writable `HIVE_WORK_DIR`, `go test ./pkg/agent/ -count=1`, go test's default 600s timeout):

| | unmodified v4 | this branch |
|---|---|---|
| `TestConcurrentPauseResume_NoPanic` alone | 223.57s | **0.01s** (`-race` clean) |
| Full suite | **`panic: test timed out after 10m0s`** — the issue's exact symptom | **completes in 522–537s**, no panic, coverage profiles write |
| Failures before the timeout | 3 (`TestStart_AgyLaunchCommandLine` 37.36s, `TestSendKick_RestartsCrashedCLIThenDelivers` 67.19s, `TestRestartThenSendKick_CleanSlateThenDelivers` 67.19s) | 2 — a strict subset, same tests, same timings |

The residual failures are pre-existing and out of this issue's scope: they fail on the untouched baseline with identical timings, pass individually under the same env (`TestSendKick_RestartsCrashedCLIThenDelivers` is also flaky in isolation: 1 fail / 2 pass across three identical runs), sit in files this PR does not touch, and belong to the ambient-state hermeticity class tracked by #4581/#4631. This branch strictly reduces the failure set and removes the timeout.

Also verified: `gofmt` clean, `go vet ./pkg/agent/` clean, and the three adopt tests plus the concurrency test pass with `-race`.

Companion PR #4627 (live-host `TestAgentAuthState_*` failures) touches different tests; together they move the package toward green and bounded on live hosts.

### DIFF
--- a/src/pkg/agent/launch_coverage_test.go
+++ b/src/pkg/agent/launch_coverage_test.go
@@ -2,7 +2,6 @@ package agent
 
 import (
 	"context"
-	"os/exec"
 	"testing"
 	"time"
 
@@ -25,14 +24,18 @@ func TestStart_CLIAlreadyRunning(t *testing.T) {
 	agent := m.agents["cxa"]
 	m.mu.RUnlock()
 
-	// Pre-create the session and render a CLI marker.
-	if err := exec.Command("tmux", "new-session", "-d", "-s", agent.tmuxSession).Run(); err != nil {
+	// Pre-create the session and render a CLI marker. This must land on the
+	// package test tmux server (-L defaultTmuxSocket) — the same server the
+	// manager inspects. A bare `tmux` exec would hit the default-socket server
+	// instead, where the manager never looks, so the adopt branch under test
+	// would silently never trigger (#4628).
+	if err := testTmuxCommand("new-session", "-d", "-s", agent.tmuxSession).Run(); err != nil {
 		t.Skipf("cannot create tmux session: %v", err)
 	}
-	defer exec.Command("tmux", "kill-session", "-t", agent.tmuxSession).Run()
+	defer testTmuxCommand("kill-session", "-t", agent.tmuxSession).Run()
 	// Inject a marker; forceRelaunch defaults to false.
-	exec.Command("tmux", "send-keys", "-t", agent.tmuxSession, "-l", ": goose is ready").Run()
-	exec.Command("tmux", "send-keys", "-t", agent.tmuxSession, "Enter").Run()
+	testTmuxCommand("send-keys", "-t", agent.tmuxSession, "-l", ": goose is ready").Run()
+	testTmuxCommand("send-keys", "-t", agent.tmuxSession, "Enter").Run()
 	time.Sleep(500 * time.Millisecond)
 
 	if err := m.Start(context.Background(), "cxa"); err != nil {
@@ -62,12 +65,13 @@ func TestStart_InferenceCLIAlreadyRunning(t *testing.T) {
 	agent := m.agents["cxa"]
 	m.mu.RUnlock()
 
-	if err := exec.Command("tmux", "new-session", "-d", "-s", agent.tmuxSession).Run(); err != nil {
+	// Same-server requirement as TestStart_CLIAlreadyRunning above.
+	if err := testTmuxCommand("new-session", "-d", "-s", agent.tmuxSession).Run(); err != nil {
 		t.Skipf("cannot create tmux session: %v", err)
 	}
-	defer exec.Command("tmux", "kill-session", "-t", agent.tmuxSession).Run()
-	exec.Command("tmux", "send-keys", "-t", agent.tmuxSession, "-l", ": esc to interrupt bypass permissions on Claude").Run()
-	exec.Command("tmux", "send-keys", "-t", agent.tmuxSession, "Enter").Run()
+	defer testTmuxCommand("kill-session", "-t", agent.tmuxSession).Run()
+	testTmuxCommand("send-keys", "-t", agent.tmuxSession, "-l", ": esc to interrupt bypass permissions on Claude").Run()
+	testTmuxCommand("send-keys", "-t", agent.tmuxSession, "Enter").Run()
 	time.Sleep(500 * time.Millisecond)
 
 	if err := m.Start(context.Background(), "cxa"); err != nil {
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed and independently reproduced. On unmodified v4 with a writable HIVE_WORK_DIR, TestConcurrentPauseResume_NoPanic alone blows a 90s cap on my box (the issue's launch-storm symptom); on this branch the same test under the same env passes in 0.02s, -race clean, with no leaked default-socket tmux server afterward. The test is not gutted — the sandbox branch of Pause/Resume is exactly the mutex-accounting path (persistPause defer discipline, sandboxResumeAfterCancel, state transitions) the test exists to hammer, and relaunch_race_guard_test.go still covers the real relaunch path with stubs.

The tmux-server-mismatch fix is also verified with coverage both ways: the adopt block in manager.go (incl. trust-prompt and inference-dismiss sub-branches) had 0 hits from the three adopt tests on v4 and full hits on this branch, while the full-launch fallthrough now correctly shows 0 hits from them. gofmt/vet clean, trial-merge against v4 tip clean.

/lgtm
/approve

PR #3932, opened 2026-08-16
🌱 Sync workflows from kubestellar/infra

clubanderson:
This PR syncs the caller workflows from `kubestellar/infra`.

These workflows call reusable workflows from `kubestellar/infra`:

**Standard Workflows:**
- `add-help-wanted.yml` - Add help-wanted label to issues
- `assignment-helper.yml` - Handle issue assignments
- `feedback.yml` - Collect feedback
- `greetings.yml` - Welcome new contributors
- `label-helper.yml` - Manage labels
- `pr-verifier.yml` - Verify PR contents
- `pr-verify-title.yml` - Verify PR title format
- `scorecard.yml` - Security scorecard
- `stale.yml` - Mark stale issues/PRs

**Agentic Workflows (Copilot Integration):**
- `ai-fix.yml` - Assign Copilot to issues with `ai-fix-requested` label
- `copilot-automation.yml` - Automate Copilot PR processing (DCO, labels)
- `copilot-dco.yml` - Override DCO for Copilot PRs

Auto-generated by workflow sync

PR #4661, opened 2026-08-23
📖 ci: resolve the infra-sync clobber of scorecard.yml's release-line pin (#4616)

### LINKED ISSUE #4616: [ci-maintainer] Workflow sync from kubestellar/infra clobbers scorecard.yml release-line pin — Release Line Guard red on sync branch
## CI Issue

The automated workflow sync from kubestellar/infra (branch `sync/workflows-from-infra`) overwrites `.github/workflows/scorecard.yml` with the infra copy, which pins `push.branches: ["main"]` only. The repo's `.github/release-lines.yml` manifest requires scorecard.yml to cover `[main, v4]` (`scorecard.yml: [main, -v2]`). Result: Release Line Guard fails on every sync branch push and will keep failing on every future sync until the divergence is resolved.

If the sync branch were merged as-is, OpenSSF Scorecard would silently stop running on v4 — exactly the #4339 failure mode the guard exists to prevent.

## Evidence

- Workflow: Release Line Guard, runs 32607718616 and 32607717800 (2026-08-23T00:23, branch `sync/workflows-from-infra`)
- Guard output: `FAIL: scorecard.yml:7 branches: has [main], expected [main,v4] — missing: v4`
- `v4:.github/workflows/scorecard.yml` has `branches: [ "main", "v4" ]`; the sync branch copy has `branches: [ "main" ]`

## Recommendation

Pick one (maintainer call):
1. Add `v4` to scorecard.yml's branch pin upstream in kubestellar/infra so the synced copy satisfies hive's release-line manifest, or
2. Exclude scorecard.yml from the infra→hive sync set, since hive carries a repo-local release-line pin the sync cannot know about, or
3. Have the sync tooling preserve hive-local `branches:` pins for workflows listed under `pinned:` in `.github/release-lines.yml`.

Until then, sync branches will arrive guard-red and must not be merged without re-adding the `v4` pin.

---
---

Danathar:
## Summary

Resolves #4616 in two halves, implementing the issue's **recommendation 2** (the narrowest of the three it records):

1. **The fix, upstream**: kubestellar/infra#145 adds `hive: [scorecard.yml]` to the sync tool's existing per-repo exclusion mechanism (`caller-workflows/workflow-exclusions.yml`, same shape as the `console`/`pr-verifier.yml` precedent already in that file). With it merged, the infra→hive sync stops overwriting `scorecard.yml`, so sync branches stop arriving Release-Line-Guard-red and hive maintains its `scorecard.yml` locally, bumping the pinned `reusable-scorecard.yml` SHA deliberately.
2. **This PR, in hive**: records the interaction where the pin is defined, so the story survives in-repo:
   - `.github/release-lines.yml` — the `scorecard.yml` entry now notes it is the pin the upstream exclusion protects, and gives the play if a sync branch ever goes red on it again (the exclusion was lost upstream — restore it there; never merge the sync branch or widen the manifest entry to quiet the guard).
   - `src/docs/release-line-guard.md` — new section "The infra workflow sync": why every sync arrived red, why that red is the guard working as designed, and why a red sync branch warrants reviewing the **whole** sync diff, not just the guard's line.

## Why recommendation 2

- Recommendation 1 (add `v4` to infra's shared `caller-workflows/scorecard.yml`) would push a hive-specific branch name at every synced repo in the org.
- Recommendation 3 (teach the sync to preserve target-local `branches:` pins) is a larger tooling change; it remains open upstream, and if ever built, the exclusion becomes unnecessary rather than harmful.
- Recommendation 2 uses a mechanism the sync tool already supports, affects only hive, and is one line plus a comment.

## What reviewing the sync diff shows (context, no action here)

The guard only checks branch filters, but the current `sync/workflows-from-infra` diff also reverts other deliberate hive hardening the guard cannot see: every cross-repo reusable-workflow ref goes from an immutable commit SHA back to mutable `@main` (deleting the in-file SECURITY comment that explains the pin), several workflows swap `pull_request` for `pull_request_target` while adding `secrets: inherit`, and `pr-verifier.yml`'s pinned-image inline job is replaced by a `@main` reusable call. That is consistent with the last two sync PRs (#4611, #3932) being closed unmerged. The new doc section says exactly this: a red guard on a sync branch is a signal to review the entire sync, not to patch the one flagged line.

## Verification

- `src/scripts/check-release-lines.sh` — PASS with the manifest comment in place (comments in `pinned:` are already part of the file's grammar; `scorecard.yml:7 branches: main,v4` still asserted).
- `src/scripts/test-release-lines-guard.sh` — PASS (the guard still catches a stale allow-list).
- Docs-only otherwise; no workflow behavior changes in this PR.

Until kubestellar/infra#145 merges, sync branches keep arriving red — the guard keeps correctly blocking them, and the new documentation tells the reader why and what to do.

### DIFF
--- a/.github/release-lines.yml
+++ b/.github/release-lines.yml
@@ -53,6 +53,17 @@ pinned:
   # recorded as an exclusion rather than quietly corrected, because turning
   # OpenSSF Scorecard on for a release line is a maintainer decision, not a
   # side effect of adding a CI guard. See the note at the bottom of this file.
+  #
+  # scorecard.yml is ALSO the one pinned workflow that the automated workflow
+  # sync from kubestellar/infra used to overwrite (#4616): the synced copy pins
+  # `[main]` only, so every `sync/workflows-from-infra` branch arrived with
+  # this guard red, and merging one would have silently stopped Scorecard on
+  # v4 — #4339 again. The sync now skips scorecard.yml for this repo
+  # (kubestellar/infra `caller-workflows/workflow-exclusions.yml`); this file
+  # is the pin the exclusion protects. If a sync branch ever goes red on this
+  # entry again, the exclusion has been lost upstream — restore it there; do
+  # not merge the sync branch, and do not "fix" it by widening this entry.
+  # See src/docs/release-line-guard.md, "The infra workflow sync".
   scorecard.yml: [main, -v2]
   suid-contract.yml: []
   v2-ci.yml: []
--- a/src/docs/release-line-guard.md
+++ b/src/docs/release-line-guard.md
@@ -165,7 +165,9 @@ The resolution ([#4616](https://github.com/kubestellar/hive/issues/4616),
 recommendation 2): the sync tool supports per-repo exclusions in
 kubestellar/infra `caller-workflows/workflow-exclusions.yml`, and hive's
 `scorecard.yml` is excluded there
-([kubestellar/infra#145](https://github.com/kubestellar/infra/pull/145)). This
+([kubestellar/infra#146](https://github.com/kubestellar/infra/pull/146); the
+equivalent [kubestellar/infra#145](https://github.com/kubestellar/infra/pull/145)
+was superseded by it). This
 repo maintains `scorecard.yml` locally and bumps its pinned
 `reusable-scorecard.yml` SHA deliberately.
 
--- a/src/docs/release-line-guard.md
+++ b/src/docs/release-line-guard.md
@@ -139,6 +139,41 @@ The workflow runs the self-test *before* the check, because a checker that has
 quietly stopped detecting anything would otherwise pass the real check for the
 wrong reason.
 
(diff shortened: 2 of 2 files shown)


clubanderson:
Reviewed. Both guard scripts verified locally on the head (check-release-lines.sh PASS, test-release-lines-guard.sh self-test PASS with the new manifest comment), and I independently confirmed the doc's hardening-revert warning against the actual sync branch diff: it does downgrade the SHA-pinned reusable refs to @main and flip pull_request → pull_request_target with secrets: inherit — the guard red is correctly load-bearing and the sync branch must not be merged. Upstream half is in: the exclusion is live on kubestellar/infra main (220beeeb, via #146 — functionally identical to your #145, which I closed as superseded with credit; my fix-forward dd612d01 on this branch points the doc reference at the merged PR). The 'if this goes red again, restore upstream, don't widen the entry' guidance is exactly the right failure-mode documentation.

/lgtm
/approve

Issue #4665, opened 2026-08-23
RFC: agent self-healing watchdog — k8s-style liveness/readiness reconciliation for hive agents


hanthor:
## Summary

Hive agents can be **dead while the dashboard reports `running`**. `state=running` is a config echo, not an observation — the pane is the truth. This RFC proposes a self-healing watchdog controller for hive agents, modeled on Kubernetes' reconciliation loop (liveness probe → restart with backoff → status conditions), so the hive heals stuck agents without an operator.

## Observed failure modes (all live, 2026-08-23, tuna-os hive)

Four distinct "running but broken" states hit this fleet in one day:

1. **Login-expired CLI** — every claude pane showed `● Login expired · Please run /login`; the OAuth refresh token was dead. Dashboard: `running`. Restart counter climbed to 1042 because the governor kept kicking into a dead credential and the session bounced.
2. **First-run onboarding overlay** — agy's interactive accent/terms picker (`[Next]`, `↑/↓ Navigate`) blocks every launch until dismissed once. Four agents sat on it for hours. Dashboard: `running`.
3. **Shared-state permission clobber** — agents share `$HOME` (e.g. `/data/home`). agy rewrites `antigravity-cli/settings.json` and `cache/onboarding.json` **as mode 600 owned by the writing agent**, so the next agent's launch hits `EACCES`, falls back to defaults, and re-shows onboarding (#2). The dashboard never knows.
4. **Wedged kick API** — `POST /api/kick/{agent}` hangs for 60s+ (no response) when the target's TUI is in a weird state, so even the *operator's* control-plane action stalls.

Plus the long-standing class: **idle-but-alive** — agents running fine but producing nothing (no issues/PRs) for extended periods, invisible to every probe today.

## Kubernetes philosophy, mapped

| k8s concept | hive analog today | gap |
|---|---|---|
| desired state | per-agent backend/model/cadence config | ✓ exists |
| observed state | `governor.sensing` pane-regex (login patterns etc.) | regex over *agent content* caused false positives (§10a); not a state machine |
| `livenessProbe` | none | no "is this CLI actually healthy" check |
| `readinessProbe` | none | no "is it producing" check |
| kubelet restart | hive restarts sessions on some deaths | detection is the gap; kick API can wedge |
| `CrashLoopBackOff` backoff | none | no exponential backoff / crash-loop cap |
| `status.conditions` | `paused`, `pausedTrigger` | no Ready/Authenticated/Producing conditions |

## Proposed design

[...]

PR #4668, opened 2026-08-24
fix(image): launch agy/claude under umask 007 for shared-$HOME fleets

hanthor:
## Problem

Hive agents can share a single `$HOME` (e.g. `/data/home` on the tuna-os spoke). Both agy and claude rewrite their credential/state files **as mode 600 owned by the writing agent** on token refresh / state save:

- agy: `antigravity-cli/settings.json`, `cache/onboarding.json`, `antigravity-oauth-token`
- claude: `.credentials.json`

The next agent on the same provider then hits `EACCES`, falls back to defaults, and re-shows first-run onboarding / `Login expired` screens — the dashboard still reports `running`. Observed live 2026-08-23: architect's claude rewrote `.credentials.json` as `-rw------- hive-architect`, breaking every other claude agent until an operator chmod'd it.

## Fix

Wrap both binaries in a one-line shim that sets `umask 007` before exec:

```sh
#!/bin/sh
umask 007
exec /usr/local/bin/<bin>.real "$@"
```

Shared-state writes then stay group-readable (`660`/group `770`); ownership may change per refresh but access no longer does. Verified in the rebuilt image: running agy processes report `Umask: 0007` and create `-rw-rw----` files. See RFC #4665 for the broader self-healing watchdog design this unblocks.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -562,5 +562,20 @@ RUN find / -xdev -type f \( -perm -4000 -o -perm -2000 \) \
       exit 1; \
     fi
 
+
+# --- Layer 8.26: shared-$HOME state hygiene (umask 007 wrappers) ---
+# Agent CLIs rewrite their credential/settings state as mode 600 owned by the
+# writing agent (agy: antigravity-cli/*.json + oauth token; claude:
+# .credentials.json). With a shared $HOME (/data/home) that breaks every other
+# agent on the same provider (EACCES -> login/onboarding screens). Launch
+# through wrappers that set umask 007 so shared state stays group-readable.
+# See RFC #4665.
+RUN for b in agy claude; do \
+      [ -e "/usr/local/bin/$b" ] || continue; \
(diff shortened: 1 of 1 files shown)


clubanderson:
Reviewed and verified: the umask is scoped to per-binary wrappers only (no global weakening); the agy half is a guarded no-op in this image; npm's relative claude symlink still resolves after the rename (same directory); new files under the setgid 2775 dev:node shared `.claude` are born 660 group-node, so siblings read them immediately instead of waiting on the permissions watcher's reactive 0660 reconciliation. Post-#4619 this is still exactly right — `.credentials.json` lives in the deliberately fleet-shared bridged `.claude` — and it's complementary to #4641's session adoption (token readability vs session identity), not redundant with it. 007 removes only world bits; every consumer is in group node. Wrapper is 0755, placed after the setuid-inventory gate. DCO signed. CI green minus tide.

/lgtm
/approve

PR #4669, opened 2026-08-24
🐛 Give the ttyd proxy a non-pooling agent so browser basic-auth can succeed

Danathar:
## What

The dashboard's `▶ terminal` link dies with 502 "Terminal unavailable" on every browser basic-auth attempt — **with the correct password**. Measured live on a rootless podman hive, deterministically:

```
1 unauth GET  → 401   (browser shows the auth dialog)
2 auth GET    → 502   "Terminal unavailable"
3 auth GET    → 200   (reload works)
… and the pattern repeats
```

## Why

ttyd (libwebsockets) serves **one HTTP request per TCP connection** and hangs up on any reuse. `httpxy` — http-proxy-middleware v4's engine — pools upstream sockets in its own `keepAlive: true` default agents whenever no explicit `agent` option is given (`setupOutgoing`: `options.agent !== void 0 ? … : defaultAgents.http`). So the browser's unauthenticated 401 leaves a pooled socket ttyd will never serve again; the authenticated retry rides it into ECONNRESET (`[ttyd-proxy] … → socket hang up`), which the proxy's error handler renders as 502. Every fresh-connection request re-poisons the pool for the next one, which is why the failure alternates forever instead of clearing.

## Fix

An explicit `agent: new http.Agent({ keepAlive: false })` on `ttydProxy`. httpxy honors an explicit agent, so every proxied request gets a fresh connection. Pooling bought nothing on this route anyway — its real traffic is a single long-lived websocket, which never pooled (`agent = false` on upgrade requests already).

## Test

`ttyd_auth_retry.test.js` drives the browser's exact sequence through the **real proxy** (spawned `server.js`) against a mock ttyd that reproduces the one-request-per-connection behavior (second request on a reused socket → destroy). Verified both directions:

- with the agent line removed, it fails with the exact live symptom: `authenticated retry must reach ttyd on a fresh connection, got 502`
- with the fix, the 401 → auth-retry → repeat sequence is all green

One test-infra finding worth noting: the test kills its spawned proxy **before** `process.exit` — a `process.exit(0)` inside `.then()` skips `.finally()`, and during development the orphaned child kept serving later runs from its port, producing a false pass. The same orphan squatting on ports 19001-19003 is the likely cause of sporadic `C3 test failed: alice (allowlisted) should reach terminal, got 401` failures when `server.test.js` runs are interrupted — with stray children killed, the full proxy suite passes clean, C3 included.

`npm test` (full proxy suite): all green.

### DIFF
--- a/src/proxy/package.json
+++ b/src/proxy/package.json
@@ -2,12 +2,12 @@
   "name": "hive-dashboard-proxy",
   "version": "2.0.0",
   "private": true,
-  "description": "Thin proxy: auth + static files → Go binary API",
+  "description": "Thin proxy: auth + static files \u2192 Go binary API",
   "type": "module",
   "main": "server.js",
   "scripts": {
     "start": "node server.js",
-    "test": "node server.test.js && node health_probe.test.js && node session_cookie_v2.test.js && node session_cookie_v3.test.js && node session_pubkey_rotation.test.js && node terminal_cookie_verify.test.js && node redoc_pin.test.js && node csp_token_injection.test.js && node csp_script_src_elem.test.js"
+    "test": "node server.test.js && node health_probe.test.js && node session_cookie_v2.test.js && node session_cookie_v3.test.js && node session_pubkey_rotation.test.js && node terminal_cookie_verify.test.js && node redoc_pin.test.js && node csp_token_injection.test.js && node csp_script_src_elem.test.js && node ttyd_auth_retry.test.js"
   },
   "dependencies": {
     "express": "^5.2.0",
--- a/src/proxy/server.js
+++ b/src/proxy/server.js
@@ -1,4 +1,5 @@
 import express from 'express';
+import httpBase from 'http';
 import { createProxyMiddleware } from 'http-proxy-middleware';
 import path from 'path';
 import fs from 'fs';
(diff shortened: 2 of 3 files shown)


clubanderson:
Reviewed and verified locally:

- **Scope**: the non-pooling `keepAlive:false` agent applies ONLY to `ttydProxy` — `apiProxy` and every other route keep their default pooled agents (no keepalive degradation elsewhere).
- **WebSocket path intact**: `server.on('upgrade')` still routes `/terminal` through `ttydProxy.upgrade`; upgraded sockets are detached from agent pooling entirely, so the long-lived terminal ws is unaffected — the agent change only governs the plain-HTTP handshake/auth requests, which is exactly where the pooled-socket death occurred.
- **Test genuinely reproduces the bug**: mock ttyd tracks served sockets via WeakSet and destroys any reused connection (libwebsockets one-request-per-connection), pins 401 → 200-on-retry plus 3 repeat requests so reuse regressions can't pass by luck. Mutation check: removing the `agent:` line makes the authenticated retry fail (dead pooled socket) — red pre-fix, green post-fix.
- Full proxy suite (`npm test`, 10 test files incl. the new one) passes locally; trial-merge against current v4 head (fc98a98f) clean; CI green minus tide.

/lgtm
/approve

Issue #3704, opened 2026-08-13
🐛 bug: advice is about outdated version


MikeSpreitzer:
## What happened?

The advice comment in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/580 says that it is from 2026-08-13 09:48 EDT. The last update to `main` was commit 8a804806a1306f01a911abefb1769c49f000eb35, dated August 12, 4:50 PM EDT. In that commit, there is no file named `docs/install.md`. But one of the findings in the advice comment references that filename.

> [advisory] install.md does not document the launcher image OCI repo required when using launchers docs/install.md#prerequisites guide 

## Expected behavior

(1) All the advice that comes from looking at the code, comes from looking at the same version of the code.

(2) That version is cited in the advice comment.

(3) That version is the latest as of when the advice started being generated.

## Reproduction steps

1. Look at the advice in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/580
2. Search for "docs/install.md". If you find anything, it is outdated.

## Environment

- Branch/version: `v4` / commit 7e892978405ad7b709f039ac49e856012a707bae
- Deployment method: Hive Hub
- Operating system or Kubernetes version: unknown
- Relevant configuration (redact secrets): ?

## Logs or screenshots

Cited above.

## Additional context

This hive has been having trouble running agents, due to both (a) the governor's token budget and (b) flailure to configure LiteLLM API keys that are not out of dollar budget.

clubanderson:
Fixed by #3724, merged to `v4` at acf8432932bafd16d7ed8a3eff3c30213c57af05 (auto-close did not fire because the PR base is `v4`, not the default branch).

What changed, per the three invariants:

1. **One consistent version per digest**: at the start of each digest post cycle the hub resolves the target repo's latest default-branch commit once and pins the digest to it (`Digest.AnalyzedSnapshot`). Every finding whose `File` names a repo path is then verified against that exact commit via the GitHub contents API; a path absent at that commit (the `docs/install.md` case) is rendered as "_(file path not found at analyzed commit — finding may be outdated)_" instead of a live file reference.
2. **Version cited**: the rendered comment now ends with an "Analyzed at `owner/repo@sha`" footer linking the commit.
3. **Latest at generation start**: the pinned SHA is resolved at the start of the post cycle, before rendering/verification.

[...]

PR #4694, opened 2026-08-24
fix(advisory): rank live findings above since-removed ones in the digest

Danathar:
## The problem

The digest's top-N is a scarce resource, but staleness was computed only **after** the cap had already been applied, so it could not influence which findings won a slot.

`BuildDigestFromBeads` calls `applyTopN`, and only later — back in `runEvalCycle` — did `VerifyFindingPaths` mark the *survivors* whose file no longer exists at the analyzed commit. By then the losers were already discarded. So a finding could take a critical slot on severity and recency alone, and then be rendered with `_(file path not found at analyzed commit — finding may be outdated)_` hanging off it, while a live finding of the same severity went unshown.

This is visible on this issue right now. The 2026-08-24 digest comment on #2364 spends **4 of its 10 critical slots** on findings it then captions as outdated, while the note underneath reads *"262 more exist."*

## The fix

Resolve the snapshot **before** the digest is built and pass it through `DigestOptions`, so the cap can consult it:

- Within a severity band, findings whose file still exists take the slots first.
- Stale ones are **set aside, not dropped** — they backfill any slot no live finding of that severity claims, so the digest never renders short.
- Backfilled findings keep `PathStale`, so the renderer still captions them. This must not reintroduce #3704.

**Freshness deliberately does not cross severity bands.** `PathStale` means the cited path *moved*, not that the problem is gone — a stale critical still outranks a live low. Demoting across bands would let a cosmetic nit displace a security finding whose file was merely renamed. There is a test pinning this.

## Lookup cost

Ranking the full set costs roughly what verifying only the rendered findings used to:

- results cached per distinct path,
- the walk stops as soon as the cap is filled (a cap of 3 over 50 findings costs 3 lookups — tested),
- non-path refs (`gh-123`, `owner/repo#1`) never trigger one.

One honest tradeoff worth a maintainer's eye: snapshot resolution now runs once per eval cycle rather than only on cycles that post, adding at most one `GetRepo` and one `LatestCommitHash` to a non-posting cycle. In exchange, the digest handed to the dashboard and the status payload is now snapshot-pinned and path-verified too — previously only the posted comment was.

## Compatibility

With no snapshot — no GitHub client, or the commit could not be resolved — `VerifyPath` is nil and ranking falls back to the exact previous severity-then-recency behavior. A test pins that. `VerifyFindingPaths` stays exported and unchanged; `BuildDigestFromBeads` still calls it when nothing overflowed, since `applyTopN` returns early in that case and verifies nothing.

## Verification

```
go build ./...                                    ok
go vet ./pkg/advisory/... ./cmd/hive/...          ok
gofmt -l pkg/advisory/ cmd/hive/main.go           clean
go test ./pkg/advisory/... ./cmd/hive/... \
        ./pkg/github/... ./pkg/hub/...            ok (all 4 packages)
```

11 new tests in `src/pkg/advisory/topn_stale_test.go` cover: fresh-beats-stale within a band, stale backfill when live findings run out, `PathStale` preserved on backfilled findings, severity beating freshness across bands, per-path caching, early exit, issue-refs skipped, nil-verify back-compat, and two end-to-end cases through `BuildDigestFromBeads` (capped and under-cap).

`src/docs/advisory.md` "How much the digest shows" updated to describe the new ordering and its limits.

## Scope note

I checked the other findings in the current digest before picking this one. The `AuthorizeIssueOpen` coverage gap (3 of the 10 slots, same finding under three wordings) is already addressed by open PRs #4690 and #4692. The remaining criticals are stale: the `src/test` network-coupling finding is fixed at HEAD (`//go:build integration` + a `TestMain` reachability gate), the SSRF/`requireOwnerRole` regression finding is fixed (`ssrf_guard.go` present and wired, guards on both beads-reset handlers), the dashboard `audit.go` and hub `oauth.go` coverage findings name functions that no longer exist and files that now have tests, and the `heartbeatBearerOK` finding retracts itself in its own detail text. That pattern — most of the visible top-10 being stale — is what pointed at the ranking bug.

Refs #2364. Related: #3704 (the path-verification this builds on).

### DIFF
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -5949,9 +5949,63 @@ func runEvalCycle(
 					"count", len(pruned), "staleness_days", advCfg.StalenessDays, "titles", strings.Join(pruned, "; "))
 			}
 		}
+		// Repo entries may be org-qualified ("org/repo"); the digest linkifier
+		// needs the bare repo name alongside the org.
+		org, repoName := cfg.Project.Org, primaryRepo
+		if parts := strings.SplitN(primaryRepo, "/", 2); len(parts) == 2 {
+			org, repoName = parts[0], parts[1]
+		}
+
+		// #3704: pin the digest to ONE repo commit. Resolve the target repo's
+		// latest commit ONCE here (invariants 1 & 3), cite it in the rendered
+		// comment (invariant 2, via the footer FormatDigestMarkdown emits when
+		// AnalyzedSnapshot is set), and verify each finding's file path against
+		// that exact commit so a since-removed path (e.g. "docs/install.md") is
+		// flagged as outdated rather than cited as live. Best-effort: if the SHA
+		// cannot be resolved, fall back to the previous unpinned behavior rather
+		// than skip the digest.
+		//
+		// This is resolved BEFORE the digest is built because the top-N cap
+		// consumes it: ranking cannot prefer a live finding over a since-removed
+		// one unless it knows which is which at ranking time (#2364).
 		digestOpts := advisory.DigestOptions{
 			MaxFindings: advCfg.MaxFindings,
 			ShowAll:     advCfg.ShowAll,
+		}
+		if ghClient != nil && org != "" && repoName != "" {
+			branch := cfg.Policies.Branch
+			if branch == "" {
+				if r, _, rerr := ghClient.GetRepo(ctx, org, repoName); rerr == nil {
+					branch = r.GetDefaultBranch()
+				} else {
+					logger.Warn("advisory: could not resolve default branch for snapshot", "repo", primaryRepo, "error", rerr)
+				}
+			}
+			if branch != "" {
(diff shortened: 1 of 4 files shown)


clubanderson:
Reviewed and verified locally: build/vet/gofmt clean, full pkg/advisory suite green on the branch. Mutation check: inverting the fresh-vs-stale preference in applyTopN fails 5 of the new tests, so the ranking behavior is genuinely pinned, not padded. The design calls are right — freshness breaking ties only WITHIN a severity band (a renamed file must not let a live nit displace a stale critical), stale findings backfilling rather than dropping (digest never renders short), inconclusive path checks defaulting to live so a rate-limit can't cost a real finding its slot, and nil-verify falling back to the exact prior ordering. Lookup cost claims hold (per-path cache + stop-at-cap, both tested). Fixed one doc nit forward (e6221f85): advisory.md said the snapshot resolves 'once per post cycle' but the code — and your own tradeoff note — resolve it per evaluation cycle so the dashboard/status digests are pinned too.

/lgtm
/approve

PR #4696, opened 2026-08-24
fix(dashboard): convert 4 CSP-dead inline handlers to data-action dispatch

### LINKED ISSUE #4680: 🐛 bug: Save refused when changing an agent's explain mode
## What happened?

In https://hosted-llm-d-incubation-llm-d-fast-m-yfzz.apps.fmaas-vllm-d.fmaas.res.ibm.com/?sso_hop=1#config/quality/General I tried to change the explain mode. Before changing anything, it appeared as follows (the value was "(hive default)").

Image

Then I changed it to full. Following is a snapshot from after changing that value.

Image

Then I clicked on "Save". I was rebuffed with a brief toast saying that "No changes to save".

## Expected behavior

I expected to be able to save a change to an agent's explain mode.

## Reproduction steps

1. Change the value in the explain mode field of the General tab on an agent's config.
2. Click "Save".

## Environment

- Deployment method: Hive Hub

## Logs or screenshots

Inline above.

Danathar:
Fixes #4680.

## Root cause

The Explain Mode select was still carrying an inline event handler:

```html
<select id="cfg-explain-mode" onchange="markDirty('general','explainMode',this.value)">
```

The dashboard serves **`script-src-attr 'none'`** (#3848, ADR-0016), so the browser refuses to run inline `on*=` handler attributes. So `markDirty` never fired → `_configState.dirty` stayed `{}` → `saveConfig` took its `Object.keys(dirty).length === 0` branch and reported *"No changes to save"*.

The save logic is fine. The control was dead. That is why the dropdown visibly changed (the browser still updates the select's own value) while the app never learned about it — and why there's no console error or failed request to notice: a CSP-blocked handler attribute simply never runs.

## Not just the one field

The #3848 event-delegation refactor was supposed to replace *every* inline handler with `data-action` dispatch. Grepping the SPA found **four** survivors, all broken by the same mechanism:

| Line | Control | Effect |
|---|---|---|
| 17750 | Explain Mode select (agent → General) | the reported bug |
| 16557 | vLLM/llm-d **API key header** | edit silently discarded |
| 16561 | vLLM/llm-d **API key env var** | edit silently discarded |
| 16565 | vLLM/llm-d **endpoint override** | edit silently discarded |

The last three call `saveInferenceAuthField` directly (they persist on change rather than via Save), so they were dropping discovery-auth edits on the floor with no feedback at all. Fixing only the reported field would have left those broken, so all four are converted.

## The fix

Each becomes the delegated form its immediate neighbours already use — e.g. `cfg-caveman-mode`, one field above Explain Mode:

```html
<select id="cfg-explain-mode"
        data-change-action="markDirty"
        data-arg0="general" data-arg1="explainMode" data-arg-types="s,s,v">
```

`data-arg-types` maps positionally: `s` passes the matching `data-argN` verbatim, `v` passes the element's live value — so the dispatcher calls `markDirty('general', 'explainMode', el.value)`, exactly the old call. `change` fires in the same cases the inline `onchange` did (selection for `<select>`, commit/blur for `<input type="text">`), so behaviour is unchanged wherever it was already working.

Both handlers resolve via `window[action]`; I confirmed `markDirty` and `saveInferenceAuthField` are top-level declarations, same as the already-delegated `switchModel`/`openConfigDialog`.

## Why this shipped, and the guard

`TestCSPScriptSrcAttrUnsafeInlineIsAbsent` pins the policy half, and its doc comment already asserts:

> every inline on*= handler attribute in `static/index.html` and in Go-generated HTML was replaced with data-action / data-\* attributes

Nothing verified that claim against the HTML — the test only reads the CSP *header*. That gap is exactly how four dead attributes shipped, and the failure mode is invisible in CI and in the browser console.

So this adds `csp_inline_handlers_test.go`, which walks the embedded static assets and fails on any inline handler attribute, naming file and line. Confirmed it catches the real bug — reverting the HTML fix produces:

```
static/index.html has 4 inline on*= handler attribute(s), which CSP script-src-attr 'none'
blocks — the control is dead and its Save silently does nothing (#4680):
  line 16557: ...onchange="...
  line 16561: ...onchange="...
  line 16565: ...onchange="...
  line 17750: ...onchange="...
```

Two details worth reviewing:

- **A positive control** (`TestInlineHandlerDetectorFlagsKnownBad`) asserts the detector both flags known-bad markup and does *not* flag JS property assignment (`el.onclick = fn`), which is not an attribute and is not CSP-blocked. Without it, a regex that quietly stopped matching would turn the guard into a permanently-green no-op — the same class of failure being fixed here.
- **Scoped to the dashboard.** `pkg/hub/static` has ~34 inline handlers, but the hub sets no CSP header at all, so those still run and are not bugs. The test deliberately does not cover them.

## Verification

```
go build ./...                     ok
go vet ./pkg/dashboard/            ok
gofmt -l  (new file)               clean
go test ./pkg/dashboard/           ok  (full package, 48s)
```

The SPA's inline `<script>` hashes are recomputed from the served bytes at startup, so editing markup inside a template literal needs no hash update; `TestEmbeddedIndexScriptsSatisfyCSPHashes` still passes.


clubanderson:
Verified: all four converted controls dispatch correctly — the delegation's document-level change listener resolves data-change-action via window[action], and both markDirty and saveInferenceAuthField are top-level function declarations in a classic script, so they are reachable; data-arg-types s,s,v reproduces the original positional args (two literals + live element value) exactly. Grepped the PR head's index.html: zero remaining on*= attributes. The new guard walks every embedded .html/.js asset (with a scanned==0 tripwire) and the positive-control test pins both the four fixed attributes verbatim and the non-flagging of JS property assignments — solid protection against reintroduction. Guard tests + check-inline-js.js syntax check pass locally.

/lgtm
/approve

PR #4698, opened 2026-08-24
fix(terminal): say when the agent CLI owns the scroll, not a bare "[live]"

### LINKED ISSUE #4681: 🐛 bug: No indication of scroll position or time in Terminal of agent
## What happened?

I opened that Terminal of https://hosted-llm-d-incubation-llm-d-fast-m-yfzz.apps.fmaas-vllm-d.fmaas.res.ibm.com/?sso_hop=1#quality and scrolled around. I never saw an indication of my scroll position or timestamp. Following is an example screenshot, from when I was scrolled into the midst of the content.

Image

## Expected behavior

Previously there was usually a line number and sometimes a timestamp giving some clue about position.

## Reproduction steps

1. Open the Terminal of an agent.
2. Scroll around.

## Environment

- Deployment method: Hive Hub

## Logs or screenshots

See above.

Danathar:
Fixes #4681.

## Root cause

The report reads as a regression, and it is one. "Previously there was usually a line number and sometimes a timestamp" is tmux's native copy-mode marker, which renders in the top-right as:

```
line 67                                              07:06 [1/67]
```

A timestamp **and** a line position — exactly what the issue describes. #4399 hid it with `copy-mode -eH` and moved the position into a labelled status line instead. But that status line only renders a position in its `pane_in_mode` branch:

```
#{?pane_in_mode,[SCROLLBACK n/m lines back - ... - press q to resume] ,[live] }now %H:%M:%S
```

and those two branches **are not exhaustive**. The wheel rebind only enters copy-mode when the running program has not grabbed the mouse:

```
if -F '#{||:#{pane_in_mode},#{mouse_any_flag}}' 'send-keys -M' 'copy-mode -eH'
```

Every agent CLI is a full-screen TUI that turns mouse reporting on. I measured both branches on real tmux:

| pane | `mouse_any_flag` | wheel does |
|---|---|---|
| plain shell | `0` | `copy-mode -eH` — tmux scrollback |
| agent TUI | `1` | `send-keys -M` — **the app gets the wheel** |

So for the pane the operator actually looks at, copy-mode is unreachable, `pane_in_mode` stays `0` forever, and the status line reported a bare `[live]` however far back they had scrolled inside the agent's own buffer. Combined with the `-H` that hides the native marker, **no position and no content timestamp were anywhere on screen** — the reported symptom, and why it looked like something had been taken away.

## The fix

tmux genuinely cannot report a position here — the buffer being scrolled belongs to the application, not the pane. So the status line says that, instead of claiming `[live]`:

```
[SCROLLBACK 999999/999999 lines back - not following live output - press q to resume] now 23:59:59
[live - this app handles its own scrolling, so tmux has no line position] now 23:59:59
[live] now 23:59:59
```

A bare `[live]` asserts the viewport is at the bottom of live output. While the operator scrolls the agent's buffer that is false, and unfalsifiable from their side. Naming the owner of the scroll is the honest answer, and it is the same principle #4399 already applied to copy-mode.

I deliberately did **not** invent a position number for this state. Claiming one tmux cannot know would be worse than admitting the absence, and there is a test asserting the branch never emits `n/m lines back`.

## Second defect found on the way: `bin/ttyd-tmux.sh` had drifted

It describes itself as "a simpler standalone helper **kept in sync**", but it had silently missed the entire #4399 change — no `status-right`, no length raise, no wheel rebind. A terminal attached through it had no status line at all: no scroll position, no clock. The same symptom as this issue, from a file whose own comment claimed otherwise. Nothing checked, because the existing test only looked at `src/deploy/ttyd-tmux.sh`.

The container is unaffected (`src/Dockerfile` installs the `src/deploy/` copy), so this is the local/non-UID-isolated path only — but it is the identical bug and worth closing. The test now asserts **both** attach paths carry the manager's format and rebind byte for byte, so "kept in sync" is enforced rather than promised.

## Two tests that were not testing what they claimed

Found while adding coverage; both fixed here.

**The marker control was failing, and its neighbour was passing vacuously.** The shell test asserted plain copy-mode draws `[pos/total]` — but entered copy-mode without ever scrolling. At position 0 tmux renders a bare `0`, so the control failed on its own (visible on unmodified `v4`, 21 passed / 1 failed) and the `-H` assertion beside it passed for free, since *both* states trivially lacked the marker. Scrolling first makes the two captures differ only by `-H`, which is the thing under test.

**The Go length guard hard-coded the format's tail.** It stripped syntax with `strings.NewReplacer("#{?pane_in_mode,", "", ",[live] }", "", …)`. Adding a third branch left unexpanded format text inside the "expansion" and measured **194 chars** for a message that renders in **99** — it would have blocked this change for a truncation that does not exist. Replaced with a small expander that walks the conditionals and returns every reachable branch, so the guard survives the next format edit. Cross-checked against real tmux output; longest expansion is 99 chars against the 140 limit.

## Verification

```
bash src/deploy/test_terminal_scrollback_status.sh    33 passed, 0 failed   (was 21/1 on v4)
go test ./pkg/agent/                                  ok (456s)
go build ./... ; go vet ./pkg/agent/ ; gofmt -l       clean
```

The shell test drives a real mouse-reporting pane for the new state and pins: distinguishable from live, names who owns the scrolling, does not invent a position, keeps the labelled clock, and the `#,` escape survives — an unescaped prose comma truncates the branch silently, which is the failure mode a nested `#{?...}` invites.

## Note for maintainers

The wording of the new state is a UX call. I kept it factual and short enough for the 140-column budget, but if you would rather it named a recovery action the way the SCROLLBACK branch does ("press q to resume"), the equivalent here is bypassing mouse reporting — I left that out because the exact modifier through ttyd + xterm.js + tmux + the agent TUI is not something I could verify from here, and a wrong instruction is worse than none.

### DIFF
--- a/bin/ttyd-tmux.sh
+++ b/bin/ttyd-tmux.sh
@@ -15,18 +15,57 @@
 # /usr/local/bin/ttyd-tmux.sh by src/Dockerfile), which additionally resolves
 # per-agent tmux sockets across UIDs via su-exec. This copy is a simpler
 # standalone helper kept in sync for local/non-UID-isolated use.
+#
+# "Kept in sync" is now ENFORCED rather than promised: this file had silently
+# missed the whole #4399 scrollback-status change, so a local terminal attached
+# through it had no status line at all — no scroll position and no clock, the
+# #4681 symptom. src/deploy/test_terminal_scrollback_status.sh asserts that the
+# status format and the wheel rebind here match manager.go byte for byte.
 set -euo pipefail
 
 SESSION=${1:-supervisor}
 TTYD_HISTORY_LIMIT="${HIVE_TTYD_HISTORY_LIMIT:-50000}"
+
+# Scrollback state legibility (#4399, #4681). Kept character-for-character
+# identical to src/deploy/ttyd-tmux.sh and to tmuxStatusRight in
+# src/pkg/agent/manager.go — test_terminal_scrollback_status.sh asserts all
+# three match, because a terminal that scrolls differently depending on which
+# wrapper attached is worse than one that is uniformly wrong.
+TTYD_STATUS_RIGHT="${HIVE_TTYD_STATUS_RIGHT:-#{?pane_in_mode,[SCROLLBACK #{scroll_position}/#{history_size} lines back - not following live output - press q to resume] ,#{?mouse_any_flag,[live - this app handles its own scrolling#, so tmux has no line position] ,[live] }}now %H:%M:%S }"
+TTYD_STATUS_INTERVAL="${HIVE_TTYD_STATUS_INTERVAL:-2}"
+# tmux truncates status-right to status-right-length, DEFAULT 40 columns, which
+# cuts the message above off mid-word.
+TTYD_STATUS_RIGHT_LENGTH="${HIVE_TTYD_STATUS_RIGHT_LENGTH:-140}"
+
 PREV_MOUSE=$(tmux show-option -t "$SESSION" -v mouse 2>/dev/null || echo "on")
 PREV_HISTORY=$(tmux show-option -t "$SESSION" -gv history-limit 2>/dev/null || echo "")
+PREV_STATUS=$(tmux show-option -t "$SESSION" -gv status-right 2>/dev/null || echo "")
+PREV_STATUS_LEN=$(tmux show-option -t "$SESSION" -gv status-right-length 2>/dev/null || echo "")
+PREV_INTERVAL=$(tmux show-option -t "$SESSION" -gv status-interval 2>/dev/null || echo "")
 tmux set-option -t "$SESSION" mouse on 2>/dev/null || true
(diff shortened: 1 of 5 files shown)


clubanderson:
Reviewed and verified locally: the shell suite passes 33/0 on the branch (tmux 3.6a), the three status-format copies (manager.go, src/deploy/ttyd-tmux.sh, bin/ttyd-tmux.sh) are byte-identical (hash-checked), and the new sync assertion makes bin/'s 'kept in sync' claim enforced rather than promised — good catch on that drift. Mutation check: padding the mouse_any_flag branch past the length budget fails TestNewSessionCommandsScrollbackLegibility, so the new expander-based guard really does bound every reachable branch (the old string-replacer would have measured garbage, as you found). The mouse_any_flag diagnosis matches tmux's documented wheel-rebind condition, and refusing to invent a position tmux cannot know is the right call. shellcheck: info-level only, matching pre-existing style. The earlier quadlet-gate red was the base-wide quay digest GC (#4703/#4704), not this PR.

/lgtm
/approve

Issue #4699, opened 2026-08-24
Model discovery probes only the shared credential path — 'logged in' agent shows an all-unverified model list (authprobe.go's per-UID fix never reached cli_models.go)


Danathar:
## Summary

Two probes in this repo answer the question "is the claude backend authenticated on this hive?" and they resolve credential paths **differently**. `pkg/agent/authprobe.go` was written specifically to fix per-UID credential resolution; `pkg/dashboard/cli_models.go` still carries the pre-fix logic. On a hive where they disagree, the operator sees a self-contradicting dashboard:

- the agent card shows **`✓ logged in`** (fixed probe — finds the credential), while
- every entry in the model dropdown reads **`(common alias, unverified)`** (unfixed probe — finds nothing, falls back to the static alias list).

Reported by an operator running a self-hosted Podman hive with a **subscription** (OAuth) Claude login and no `ANTHROPIC_API_KEY`, seeing exactly that combination on a healthy, working `scanner` agent.

## The asymmetry

**`pkg/agent/authprobe.go` — per-UID aware.** Its header states the bug it exists to fix:

> `BackendAuthAvailable` answers a BACKEND-level question by stat'ing ONE shared path: `/data/home/.claude/.credentials.json` (`claude.CredentialsPath`). That was correct when every agent ran as the same UID out of the same HOME. The fleet now runs PER-AGENT UIDs — each agent has its own UID, its own tmux socket and its OWN HOME — so on a per-UID spoke the shared legacy locations are empty even while every agent is authenticated and working out of its own home.

`agentClaudeCredentialPaths` (authprobe.go:99) therefore orders **agent-own home first**, then the shared legacy path, commenting: *"the per-UID file is the one the agent's CLI actually writes."*

**`pkg/dashboard/cli_models.go` — not per-UID aware.** `claudeCredentialCandidatePaths` (cli_models.go:992) returns exactly two paths:

1. `claudePodCredentialsPath` = `claude.CredentialsPath` = `/data/home/.claude/.credentials.json` — the shared legacy path
2. `$HOME/.claude/.credentials.json` — where `$HOME` is the **dashboard process's** home, not any agent's

No agent home is ever consulted. This is the same single-shared-path assumption authprobe.go's header identifies as wrong on a per-UID fleet. The neighboring probes hardcode the same assumption (`copilotSDKAgentHome` / `gooseACPAgentHome` = `/data/home`, cli_models.go:142, :228), so the finding may generalize beyond the claude path.

[...]

Issue #4298, opened 2026-08-20
✨ feature: Better visibility into token budgeting


MikeSpreitzer:
## Use case

I am administering the hive for llm-d FMA. I just came back from vacation and wonder whether the hive's configured token budget limited activity while I was gone.

## Proposed solution

Make the hive offer a graph of token usage over time, with a visual marker for the configured budget and reset times.

## Alternatives considered

Make the hive offer a report that shows, for every recent reset, how much of the budget had been used.

## Scope and compatibility

- Target branch or version: v4
- Operator impact: better understanding of token usage
- Contributor/agent impact: none
- Migration or backwards-compatibility concerns: new code must not crash when installed in an existing environment that had not been doing any of the new history keeping (if any).

## Additional context

The cost section of an agent in the dashboard has a section that looks like the following snapshot. It has some graphs that lack horizontal and vertical axes, making them impossible to understand.

<img width="880" height="167" alt="Image" src="https://github.com/user-attachments/assets/1c3406cb-0868-496f-9ccd-504b392fbf23" />

PR #4708, opened 2026-08-24
fix(dashboard): state the budget window phase and answer it in the Cost panel

### LINKED ISSUE #4686: 🐛 bug: token spend/budget infomation is still not adequate and understandable
## What happened?

I went to https://hosted-llm-d-incubation-llm-d-fast-m-yfzz.apps.fmaas-vllm-d.fmaas.res.ibm.com/?sso_hop=1#section-cost-panel , trying to understand the token spending of this hive. My questions include the following.

- When is the hive's token spend reset? I set the period, I want to know the phase.
- At each of those resets, how many tokens had been spent?

In other words: I have configured a token budget and period; I want to know how close this hive has been coming to hitting that limit.

I did not find answers to these questions.

I _did_ find some stuff that might conceivably sort of answer my questions, but it is not understandable. I ignored the stuff denominated in dollars, because this hive does not know the true price and is not regulating based on dollars.

I found a graph of input tokens, shown in the following snapshot.

Image

The vertical axis seems to be over 2B. My configured token budget is 1B reset every 2 days. This graph is clearly not answering the question of how close this hive is coming to hitting the 1B/2d limit.

Also, there is no indication of the time range on the horizontal axis. That makes it very hard to compare to my configured budget periods.

## Expected behavior

I expected a way to find the answers to my questions.

## Reproduction steps

See above.

## Environment

- Deployment method: Hive Hub

## Logs or screenshots

Inline above.

Danathar:
Fixes #4686.

## What I found first

Most of the answer already existed. **#4298/#4320** built the per-window history — one bar per window, the limit in force drawn as a marker, reset times on the x-axis — and `renderBudgetBar` shows used/limit and percent. That is a good answer to "how close have I been coming to the limit".

But all of it renders inside `renderGovernor`, and this report was filed from `#section-cost-panel`. The Cost panel carries all-time dollar estimates and cumulative token tables and says nothing about the budget at all. **The feature was not missing; it was unreachable from where the question gets asked.** So this PR does not rebuild it — it fixes the two things that actually left the reporter without answers.

## 1. Phase

> *"When is the hive's token spend reset? I set the period, I want to know the phase."*

The reset was rendered only as `37h until reset`. That is a countdown, and a countdown cannot be lined up against a calendar — which is precisely what "I want to know the phase" asks for.

`WINDOW_STARTS_AT` / `WINDOW_ENDS_AT` were **already** on the live status and already used by the exhausted-state message, so no backend change was needed. The bar now reads:

```
Window Aug 23, 09:14 → Aug 25, 09:14 · 37h left
```

with the full unambiguous timestamps in the tooltip. Both fields are optional — empty until a limit is set and a window opens — so the helpers degrade to the bare countdown rather than rendering `Invalid Date`. There is a test for that fallback specifically.

## 2. Discoverability, and the unit

The Cost panel now carries the token budget: used against the limit, the window phase, and **the same `budgetHistoryChart()`** the Governor bar draws.

Reusing that function and `_lastBudget` is deliberate rather than convenient — two panels with independent views of one budget is exactly the failure mode #4699 was, and I would rather not ship a second one in the same week.

**The unit matters and is now stated.** The reporter discarded every dollar figure because *"this hive does not know the true price and is not regulating based on dollars"* — which is correct: the governor enforces a **token** limit. A panel showing only dollars has to say what the limit is actually denominated in. And when no token budget is configured it now says so and points at the setting, rather than rendering nothing and reproducing *"I did not find answers"*.

## What I deliberately did not change

The report says *"there is no indication of the time range on the horizontal axis"*. I checked which chart that is before touching anything:

- The **dollar** trend chart already has range pills (hourly/daily/weekly/monthly/custom) and first/mid/last x-axis labels. It is fine — the reporter had simply discounted it as dollars.
- The graph in the screenshot is one of the small **axis-less token sparklines** beside the model/agent rows. Those carry their range in a hover `<title>`, which #4298 chose deliberately: *"graphs with no axes are impossible to read — these are too small for drawn axes"*.

Giving those real axes is a real improvement but a different change, with a layout question attached (they are ~90px strips). The budget question is now answered above them, in tokens, which is what the reporter was actually trying to get from them.

## Verification

```
go build ./...                       ok
go vet ./pkg/dashboard/              ok
gofmt -l (new test)                  clean
go test ./pkg/dashboard/             ok (full package)
node --check (extracted SPA script)  ok — modified and baseline both parse
```

The SPA is one 1.1MB inline script, so a syntax error blanks the whole dashboard; I extracted and parsed it rather than trusting that Go tests would notice.

Five tests in `budget_phase_ui_test.go`, following the sibling `budget_history_ui_test.go` from #4298 — including its wired-but-undefined-function guard, since the embedded SPA has historically shipped calls to functions that were never defined. One of them pins that this markup adds **no inline `on*=` handlers**, because `script-src-attr` is `'none'` and such a handler would be silently dead.

### DIFF
--- a/src/pkg/dashboard/budget_phase_ui_test.go
+++ b/src/pkg/dashboard/budget_phase_ui_test.go
@@ -1,0 +1,130 @@
+package dashboard
+
+import (
+	"strings"
+	"testing"
+)
+
+// #4686: an operator who configured "1B tokens every 2 days" asked two things
+// the dashboard could not answer — WHEN does the window reset (the phase, not a
+// countdown), and how much had been spent at each past reset — and asked them
+// from the Cost panel, where the answer did not exist at all.
+//
+// The data and the per-window chart already existed (#4298/#4320) but lived
+// only under Governor, and the reset was rendered solely as "37h until reset",
+// which cannot be lined up against a calendar. These tests pin both halves of
+// the fix: the phase is stated absolutely, and the Cost panel answers the
+// token-budget question in tokens.
+
+// TestBudgetWindowPhaseIsStatedAbsolutely pins the phase fix. A countdown alone
+// is what #4686 reported as unanswerable.
+func TestBudgetWindowPhaseIsStatedAbsolutely(t *testing.T) {
+	html := indexHTML(t)
+
+	for _, def := range []string{
+		"function budgetWindowBounds(",
+		"function budgetWindowPhaseText(",
+		"function budgetWindowPhaseTitle(",
+		"function fmtWindowStamp(",
+	} {
+		if !strings.Contains(html, def) {
+			t.Errorf("index.html does not define %q — a referenced-but-undefined name throws at render time", def)
+		}
+	}
+
+	// The absolute bounds come from the live status fields, which are the only
+	// source that knows when THIS window opened.
+	for _, field := range []string{"WINDOW_STARTS_AT", "WINDOW_ENDS_AT"} {
+		if !strings.Contains(html, field) {
+			t.Errorf("index.html never reads %s — the window phase cannot be derived without it", field)
+		}
+	}
+
+	// The budget bar must render the phase helper rather than a bare countdown.
+	if !strings.Contains(html, "budgetWindowPhaseText(budget, hoursLeft)") {
+		t.Error("the budget bar does not render the phase — #4686's 'I want to know the phase' is unanswered")
+	}
+	if strings.Contains(html, "<span>${hoursLeft}h until reset</span>") {
+		t.Error("the budget bar still shows only the bare countdown — a countdown cannot be compared against a calendar")
+	}
+}
+
+// TestBudgetPhaseToleratesMissingBounds: WINDOW_STARTS_AT/ENDS_AT are optional
+// (empty until a limit is set and a window has opened). Rendering "Invalid Date"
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed and verified against v4 head:

- **Data-source unification confirmed**: costTokenBudgetHtml reads window._lastBudget (set on every status poll, index.html ~12498) and renders the same budgetHistoryChart() the Governor bar draws; fetchBudgetHistory runs on a global interval (~22610), so the chart is populated regardless of which section is open. No second computation that could disagree — the #4699 failure mode is genuinely avoided.
- **Phase claim verified**: WINDOW_STARTS_AT/WINDOW_ENDS_AT exist on the live status (server.go 608/613, both optional), and budgetWindowBounds rejects unparseable stamps + falls back to the bare countdown — no Invalid Date / NaN path. Test pins the fallback.
- **Unconfigured state honest**: explicit 'No token budget configured' pointing at Settings → Governor (labels verified present), instead of silence.
- **CSS/helper reuse sound**: .budget-bar / .cost-subtitle are global stylesheet classes; fmtTokens/esc defined.
- check-inline-js guard passes; all 5 new tests pass; trial merge over v4 head clean (auto-merged index.html), guard + tests re-pass post-merge.

/lgtm
/approve

PR #4709, opened 2026-08-24
🐛 Proxy /gh-setup through the gateway so the App install callback works

Danathar:
## What

The gateway has no route for `/gh-setup`, so GitHub's post-install redirect — `​/gh-setup?installation_id=...&setup_action=install`, the callback the [GitHub App setup doc](../blob/v4/src/docs/github-app-setup.md) builds the whole `/gh-setup` flow around — never reaches the Go API's `GET /gh-setup` handler on gateway deployments. It falls through to the SPA fallback, the dashboard renders at the `/gh-setup` URL, and the install silently doesn't complete.

## Observed live

Rootless podman hive, personal GitHub account. Operator created the App, installed it on the repo, GitHub redirected to `localhost:3001/gh-setup?installation_id=156191969&setup_action=install` — and the browser showed the dashboard with the "GitHub App Not Installed" banner still up. The Go handler never logged a callback. The flow only completed after the installation ID was saved by hand via `PUT /api/config/github`.

## Fix

Proxy `/gh-setup` to the Go API, mirroring `apiProxy` — including the `pathRewrite` that re-prepends the prefix the express mount strips. One subtlety: for a bare `/gh-setup?query`, express leaves `/?query`, and the lone slash must collapse (`/gh-setup?query`, not `/gh-setup/?query`) or the Go 1.22 mux pattern `GET /gh-setup` does not match.

## Test

`gh_setup_proxy.test.js` drives GitHub's exact callback URL through the real spawned `server.js` against a Go-API mock, asserting: the backend receives the request, the handler's `303 → /?ghSetup=ok` passes through, and the query string arrives intact. Verified it fails (404, backend never reached) with the route removed. Wired into the `package.json` test chain; full proxy suite green.

Related: the same live investigation found installation auto-discovery unable to help here either — see the companion PR on the discovery-cache Re-check path.

### DIFF
--- a/src/proxy/gh_setup_proxy.test.js
+++ b/src/proxy/gh_setup_proxy.test.js
@@ -1,0 +1,129 @@
+// The GitHub App setup callback must reach the Go API through the gateway.
+//
+// After the operator installs the App, GitHub redirects their browser to
+// /gh-setup?installation_id=...&setup_action=install. The Go API's
+// GET /gh-setup handler verifies the installation and persists the
+// installation_id. Before this route existed the gateway had no proxy for
+// /gh-setup, so the request fell through to the SPA fallback: the dashboard
+// rendered at the /gh-setup URL, the Go handler never ran, and the
+// documented setup flow silently did nothing — measured on a live gateway
+// deployment, where the operator had to paste the installation ID by hand.
+// This test drives the exact redirect request through the real proxy and
+// pins that it reaches the backend with its query string intact.
+import { createServer, request as httpRequest } from 'http';
+import { strict as assert } from 'assert';
+import { spawn } from 'child_process';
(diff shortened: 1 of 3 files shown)


clubanderson:
/lgtm
/approve

Reviewed: mount scoped to the /gh-setup segment only (probed /gh-setupx, /gh-setup-evil — 404, backend never hit; dot-segment paths keep the /gh-setup prefix and Go's ServeMux clean-path redirects them without invoking handlers). GET-unauthenticated matches the existing gateway model — mutations still pass through requireAuth, and the Go handler independently proves the installation belongs to this hive's org via App JWT before persisting (admin role required to replace an existing ID). Full proxy suite green locally. Fix-forward 9ba9f22f adds the negative scoping pin to the test.

PR #4710, opened 2026-08-24
🐛 Recover from a rotated dashboard token instead of failing every write silently

Danathar:
## What

A dashboard token the server rejects — **rotated, mistyped, or truncated on paste** — permanently and **silently** breaks every dashboard write in that browser.

The SPA's fetch wrapper (`static/index.html`) stores the token in `localStorage`, attaches it to every mutation, prompts only when it has **no** token, and never inspects the response. After a rotation the stored token is stale, so every `PUT`/`POST`/`PATCH`/`DELETE` 401s — with no prompt, no banner, and nothing pointing at the credential.

## Why it is worse than one broken call

Each feature renders the same 401 in its own idiom, so it looks like several unrelated bugs:

| Feature | What the operator sees |
| --- | --- |
| Save Installation ID (Forge App banner) | "Failed: Unauthorized" toast |
| Re-check (same banner) | button appears to do nothing |
| Apply ACMM level | confirm dialog closes as though it worked; level unchanged |
| Terminal | ttyd auth fails |

Measured on a live self-hosted hive, twice, from two different causes:

1. After a routine token rotation, one browser window kept failing writes across all four of those for a **full day** while the symptoms were chased individually.
2. On a freshly cleared browser, the operator pasted the token and lost the leading character — 63 chars instead of 64. Every write 401'd again. The stored value *looked* right at a glance (`afc6b6b10283…` vs `bafc6b6b1028…`), and only a console check of `.length` found it.

Case 2 matters because it needs no rotation at all: a first-time operator can land in the permanent-silent-failure state on day one, during setup, from one bad paste. Nothing in the UI ever suggests the saved token is the problem — and the server does not log rejected tokens either, so `podman logs` is silent too.

## Fix

When a mutation that **carried a stored token** returns 401/403: drop the token, re-prompt once with text saying it was rejected and is probably stale, retry the request once.

Two guards are load-bearing:

- **Retry exactly once** (`__hiveAuthRetried`) — a genuinely wrong token must fail and stop, not loop prompting.
- **Only when a token was actually sent** (`_authToken`) — on a hosted/hub-proxied hive identity comes from the hub login and no token exists; a 401/403 there is a real authorization denial, and clearing storage or prompting for a token the operator was never issued would be wrong.

## Tests

`stale_token_recovery_test.go` pins the recovery branch, the single-retry guard, the stored-token precondition, and that the re-prompt explains itself — following the existing `index.html` structure-test convention in this package (`agent_scroll_anchor_test.go`, `clickable_avatars_test.go`), since browser behavior can't execute in Go. **All four fail without the change.**

Beyond structure, the patched wrapper was extracted and executed against a mock backend that 401s anything but the fresh token:

```
final status      : 200
headers sent      : ["Bearer STALE","Bearer FRESH"]
prompts shown     : 1
prompt said stale : true
token now stored  : FRESH
```

`go build ./...` clean; neighboring `index.html` structure tests still pass.

### DIFF
--- a/src/pkg/dashboard/stale_token_recovery_test.go
+++ b/src/pkg/dashboard/stale_token_recovery_test.go
@@ -1,0 +1,84 @@
+package dashboard
+
+import (
+	"strings"
+	"testing"
+)
+
+// A dashboard token that the server REJECTS must not be a permanent, silent
+// dead end.
+//
+// The SPA's fetch wrapper stores the token in localStorage and attaches it to
+// every mutation. It used to prompt only when it had NO token, and it never
+// looked at the response — so once the operator rotated HIVE_DASHBOARD_TOKEN,
+// a browser holding the old value 401'd on every write forever, with no
+// prompt and no error. Each feature surfaced that differently (a "Failed:
+// Unauthorized" toast, a Re-check button that did nothing, a confirm dialog
+// that closed as if it had worked), so it read as several unrelated broken
+// features rather than one bad credential. Measured on a live hive: writes
+// failed for a full day across four features after a rotation.
+//
+// These tests pin the structure of the recovery path in the embedded
+// static/index.html. The browser behavior itself (localStorage, window.prompt,
+// a real 401) cannot execute here; what is honestly guaranteed is that the
+// recovery branch, its single-retry guard, and the stored-token precondition
+// survive future edits — matching the convention of the other index.html
+// structure tests in this package.
+
+// TestStaleTokenTriggersRecovery asserts the wrapper reacts to a rejected
+// token at all: it inspects mutation responses for 401/403, drops the stored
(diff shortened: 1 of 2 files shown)


clubanderson:
Reviewed the 401/403 recovery path: triggers only on mutations that actually carried a stored token (hosted/hub-proxied 401s stay genuine denials — no prompt, no storage clearing), bounded to exactly one retry via __hiveAuthRetried (a wrong token fails and stops), no token material logged or echoed, and a refused write still ends refused — the final response is returned to the caller unchanged. stale_token_recovery_test.go pins the guard, the single-retry bound, and the self-explaining re-prompt. check-inline-js guard passes; trial-merged over v4 head cleanly alongside #4743 with full pkg/dashboard suite green.

/lgtm
/approve

PR #4713, opened 2026-08-24
🐛 Create the agent request queues even when the App is not usable at boot

Danathar:
## What

At L3+ agents do not call `gh pr create` themselves. They push a branch and drop a **request file**; the hive opens the PR or issue server-side with the App token so it is authored by `<slug>[bot]`. The watchers that drain those queues are gated on a usable GitHub App at boot, and that gate's comment states the intent:

> Gated on a real client + usable App — with no App there is no bot to author as, **and requests simply accumulate** rather than opening under a wrong identity.

They cannot accumulate. The queues are created **inside that same gate**, so a hive whose App is not usable at boot has no directories at all — and `hive-open-pr` / `hive-open-issue`, which run in the **agent's** shell, hard-fail on ENOENT. The finding is discarded rather than queued, and the only trace is in that agent's tmux pane.

## Why this is not a corner case

**The shipped config makes PAT→App migration the normal trajectory.** `hive.yaml.example` ships with the token path live and the App path commented out:

```yaml
github:
  # Option 1: Personal access token (simplest)
  token: ${HIVE_GITHUB_TOKEN}

  # Option 2: GitHub App (recommended for production)
  # app_id: 12345
  # installation_id: 67890
```

So a self-hosted operator starts on a PAT, runs for a while, and moves to the App when they want bot-authored writes — which is exactly what `github-app-setup.md` recommends for production. That migration happens on a **running** hive, and every route by which it completes lands the installation ID *after* boot:

- the operator pastes the installation ID into the dashboard banner,
- `/gh-setup` persists it on the install redirect (`installation_id` is documented as optional precisely because this flow fills it in),
- or installation auto-discovery finds it on a later poll.

All three leave a hive that reports healthy App auth, mints installation tokens, and passes its own `POST /api/github-app/recheck` — while **every agent write is silently impossible** until some later restart happens to boot with the App already usable.

Deployment shape decides whether that ever self-corrects. On Kubernetes an init container restores the persisted runtime config over the seed, so a runtime-persisted `installation_id` is present at the next boot. **A podman/quadlet deployment has no init container**: the hive boots from the operator's read-only `hive.yaml`, so an ID that only ever reached `/data/hive.yaml.runtime` is invisible at every subsequent boot — the gate stays closed indefinitely, not just once.

The ACMM level decides whether anyone notices. L1/L2 never open PRs or issues, so the queues are never touched; the defect only surfaces the moment an agent first tries to write, which may be days after the App was set up.

## Measured

Live self-hosted L3 hive. The installation ID was saved from the dashboard after boot, so auto-discovery repaired App auth at runtime and every health surface read green. The quality agent then did its whole job and reported this in its pane:

> "Both request-queue directories under `/var/run/hive-metrics/` are missing — looks host-side, not something I can create myself. **Branch is pushed and the PR/issue bodies are ready to resubmit** once that's fixed."

`ls /var/run/hive-metrics/` confirmed `pr-requests` and `issue-requests` absent while `agent-tokens`, `gh-app-token.cache` and the governor's JSON files were all present. Nothing in the container logs, the dashboard, or `/api/health` indicated a problem. Creating the queues (by restarting once the config carried `installation_id` at boot) was the entire fix.

## Change

Create both queues unconditionally at boot via `PrepareRequestDirs`, with the mkdir+chmod logic extracted into one shared helper the two watchers reuse.

**The watchers stay exactly as gated as before.** This changes only whether there is somewhere for requests to wait — making the documented "simply accumulate" behavior real. Cost is two empty directories.

## Tests

`request_dirs_test.go`, all four failing without the change:

- both queues created with no client and no App present
- group-writable permissions agents depend on — `MkdirAll` is umask-masked to 0755, so without the explicit chmod an agent's drop gets EACCES and is lost exactly as if the directory were missing
- idempotence with a queued request already on disk (this runs on every boot; a pending request must survive)
- `ensureRequestDir` still reports genuine failure — the only signal that legitimately disables a watcher

`go build ./...` clean; full `pkg/github` suite green (42s).

## Follow-up, deliberately not in this PR

Even with the queues present, a hive whose App becomes usable at runtime does not start its watchers until the next restart, so requests accumulate until then. Re-arming the watchers when App config lands is the fuller fix; it needs a decision about how to reach the watcher's context from the config path, so I left it out rather than guess. This PR removes the data-loss half of the problem — nothing is discarded any more.

### DIFF
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -1577,6 +1577,16 @@ func main() {
 	// Gated on a real client + usable App — with no App there is no bot to author
 	// as, and requests simply accumulate rather than opening under a wrong
 	// identity. ghClient uses the App installation token (see ghAuth wiring).
+	// Create the agent-facing request queues REGARDLESS of App state. The
+	// watchers below stay gated (no App, no bot to author as), but the queues
+	// must exist either way or the "requests simply accumulate" behavior above
+	// is a fiction: hive-open-pr / hive-open-issue run in the AGENT's shell and
+	// hard-fail on a missing directory, discarding the finding instead of
+	// queueing it. App setup routinely completes after boot (operator saves the
+	// installation ID, /gh-setup persists it, auto-discovery finds it later), so
+	// this gap silently disarms agent writes on a hive that looks healthy.
+	github.PrepareRequestDirs(logger)
+
 	if ghClient != nil && cfg.GitHub.HasUsableApp() {
 		// Attribution resolver: effective backend/model from the manager
 		// (runtime overrides included), falling back to the configured values
--- a/src/pkg/github/issue_request_watcher.go
+++ b/src/pkg/github/issue_request_watcher.go
@@ -116,16 +116,10 @@ func (c *Client) StartIssueRequestWatcher(ctx context.Context, authz IssueReques
 	if nowFn == nil {
 		nowFn = time.Now
 	}
-	if err := os.MkdirAll(issueRequestDir(), 0o777); err != nil {
-		c.logger.Warn("issue-request watcher: cannot create request dir; disabled",
-			slog.String("dir", issueRequestDir()), slog.String("error", err.Error()))
+	// Shared with PrepareRequestDirs, which creates this queue unconditionally
+	// at boot so findings can accumulate even before a watcher runs.
+	if !ensureRequestDir(c.logger, "issue", issueRequestDir()) {
 		return
-	}
-	// Same rationale as the PR watcher: agents (UID >= 2001) must be able to
-	// DROP request files; MkdirAll is umask-masked, so force group-write+setgid.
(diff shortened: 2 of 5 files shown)


clubanderson:
/lgtm
/approve

Reviewed against the critical invariant: PrepareRequestDirs is a pure mkdir+chmod with no client — no path acts on GitHub without App identity, and both watcher starts remain inside the `ghClient != nil && HasUsableApp()` gate (drain-on-recovery already pinned by TestPRRequestWatcher_OpensAndConsumes: a watcher starting later consumes pre-existing queue files). Queue perms/idempotency/failure-signal tests verified; full pkg/github suite green locally. Fix-forward 1d19fb9c adds a source-level pin on the main.go gate (f3_trusted_merger_source_test.go shape) — mutation-checked: ungating the watchers, dropping PrepareRequestDirs, and dropping the group-write chmod each fail a test.

PR #4714, opened 2026-08-24
feat(governor): put the hive's default explain mode in the settings form

### LINKED ISSUE #4712: 🐛 bug: can not find hive's defaul explain mode
## What happened?

I went looking for where to configure a hive's default explain mode, and could not find it.

## Expected behavior

I expected a field in a form that I could find.

## Reproduction steps

1. Try to find where a hive's default explain mode is configured

## Environment

- Deployment method: Hive Hub

Danathar:
## What

Adds `governor.explain_mode` — the hive-wide default explain mode — and surfaces it in **Settings → Governor → General** as **Default explain mode**.

Fixes #4712.

## Why

Explain mode already had a fleet-wide default. It just had nowhere to be set from:

- The only home for it was `HIVE_EXPLAIN_MODE`, an environment variable set on the **deployment**.
- The reporter's environment is Hive Hub. A hosted spoke owner has no deployment-env access, so the fleet-wide switch was not merely hard to find — it was unreachable.
- Nothing in the dashboard mentioned the setting, so "I went looking and could not find it" was the correct outcome of looking.

The tree already has the precedent for this exact shape. `BackupConfig.ResolveKey` moved the backup encryption key out of `HIVE_BACKUP_KEY`-only for the same reason, with the same precedence: **governor config wins, the env var stays as the fallback for hives that already set it**. This follows it.

## What changed

**`governor.explain_mode` in `hive.yaml`** — `"" | off | brief | full`, validated by the same `ValidateExplainMode` gate as the per-agent field. `""` means "no hive default configured", which falls through to `HIVE_EXPLAIN_MODE` and then to `off`.

**The form reports three things, not one.** A bare input would have left half the bug in place: a default the *deployment* set through `HIVE_EXPLAIN_MODE` would still be invisible to the owner who cannot read the environment. So the field shows what is configured, what is **in force now**, and **where that came from** — `Set here.` vs `Inherited from the HIVE_EXPLAIN_MODE deployment environment variable — set a value above to override it.` The API returns `explain_mode`, `explain_mode_effective` and `explain_mode_source` alongside the existing general-advanced fields.

**Live, not next-restart.** The default is resolved per kick and per launch off the live config pointer, through the same injected `atomic.Pointer` resolver pattern as the bob API key — and for the same lock-ordering reason: both read sites (`deliverKickLocked`, `agentEnvPairs`) already hold `m.mu`, so a config read there has to be lock-free. An operator turning explanation on mid-debug sees it on the next kick.

## What deliberately did not change

- **The per-agent tri-state.** An agent with an explicit `explain_mode`, **including `off`**, still wins over the hive default. That is the whole point of the tri-state and the new setting must not erode it — covered by tests.
- **`HIVE_EXPLAIN_MODE`.** Still read, still injected into every agent process as the *resolved* mode. A hive that sets it today behaves identically.
- **Typo behaviour.** An unrecognized value in config, env, or the per-agent field still resolves to `off`, so a mistake degrades to the previous behaviour rather than to a mode nobody asked for. The API rejects an unknown mode outright rather than storing one that would read back as "set" while doing nothing.

## Tests

| Area | Covers |
| --- | --- |
| `pkg/config` | config-over-env precedence, explicit `off` in config beating an env `full`, source attribution, whitespace, unknown-value normalization, `validate()` accepting/rejecting the new field |
| `pkg/agent` | governor default applied to an unset agent, per-agent `off`/explicit values beating it, env still honoured with no governor default, resolver injection + clearing + live re-read |
| `pkg/dashboard` | GET/PUT round-trip, `400` on an unknown mode with nothing stored, absent field = no change, `""` clears the default, env fallback attributed as `env:HIVE_EXPLAIN_MODE` and overridable from config |

`go build ./...` clean; `pkg/config`, `pkg/agent` and `pkg/dashboard` suites pass. Two failures in the sandbox are environmental and pre-existing on `v4` — `TestGateFailsClosedWithoutIP6Tables` needs root/NET_ADMIN (verified failing on a clean `v4` worktree), and `TestRestartThenSendKick_CleanSlateThenDelivers` is load-flaky under a full parallel suite run and passes in isolation on this branch.

Docs updated: `agent-configuration.md` (fleet-wide default section now leads with the dashboard and `hive.yaml`), `env-vars.md`, `troubleshooting.md`, plus a CHANGELOG entry.


clubanderson:
/lgtm
/approve

Verified before merge: (1) the two sandbox failures are indeed environmental — TestGateFailsClosedWithoutIP6Tables and TestRestartThenSendKick behave identically on a clean v4 worktree and on this head (both pass on my host); (2) trial-merge against current v4 head (38391cc4, ~25 sibling merges since branch) is conflict-free, and the full pkg/agent, pkg/config, pkg/dashboard suites pass on the PR head. Resolver wiring checked: atomic.Pointer read lock-free at both m.mu-held sites (deliverKickLocked, agentEnvPairs), per-agent tri-state preserved (explicit off wins), typos normalize to off, API rejects unknown modes, config-over-env precedence with source attribution matches the BackupConfig.ResolveKey precedent.

PR #4728, opened 2026-08-25
🌱 test(agent): hermetic seams and pacing shrink so the suite runs anywhere under 600s

### LINKED ISSUE #4685: [quality] TestFixPermissions_NoPanic is non-hermetic: walks real /data dirs, can run 10+ minutes and touch live agent files
## Finding

`TestFixPermissions_NoPanic` (src/pkg/agent/helpers_coverage_test.go:135) calls `fixPermissions()`, which does `filepath.Walk` over the **real host directories** in `WatchedHomeDirs` (`/data/...`) plus `sharedRepoClones()` (src/pkg/agent/permissions_watcher.go:308-323). The test is non-hermetic:

- On a host where `/data` contains large agent trees, the walk takes minutes — observed **600s+ wall time** (the whole `pkg/agent` suite hit the 10m `go test` timeout and was killed mid-walk).
- The test's behavior depends entirely on what happens to exist on the host: on CI it exits fast because the paths are absent; on a real hive host it walks (and potentially chowns/chmods!) live agent data from a unit test.
- Same hermeticity concern applies to `TestEnsureWatchedDirs_NoPanic` right above it, though that one is cheap.

## Recommendation

1. Make the walk roots injectable: have `fixPermissions` take the root list (or read it from a package variable the test can override), and point the test at a `t.TempDir()` fixture with a few wrong-ownership entries.
2. That also upgrades the test from "doesn't panic" to actually asserting the fix-up behavior (currently it exercises the walk but verifies nothing).
3. Until then, guard the test with a skip when any watched root exists and is non-trivial, so `go test ./pkg/agent/...` is safe to run on a live hive host.

## Priority
- Impact: medium (blocks running the agent test suite on real hosts; unit test can mutate live data permissions)
- Effort: low

---

### LINKED ISSUE #4688: [quality] pkg/agent: TestDeliverStartupKick_BobReceivesKick is order-dependent flaky; full-package run exceeds the 600s coverage-gate timeout
## Finding

Two reliability problems surfaced while running the `pkg/agent` suite exactly as `coverage-hourly.yml` does (`go test ./pkg/agent -short -count=1 -timeout 600s`, single invocation):

1. **Order-dependent flake**: `TestDeliverStartupKick_BobReceivesKick` (bob_startup_kick_test.go:212) fails reproducibly in the full-package run — `bob_startup_kick_test.go:240: bob must receive its startup kick; LastKick is nil` after its full 60s wait — but **passes when run in isolation** (`-run TestDeliverStartupKick_BobReceivesKick` → ok in ~11s). Some earlier test leaves tmux/manager state that drops the kick.

2. **Suite wall-time exceeds the timeout**: the single-invocation `-short` run hits the 600s `-timeout` and panics (`panic: test timed out after 10m0s`), with a *different* test in flight each run (`TestRunCopilotDiagnostic_AuthError`, `TestRestart_MintsPerAgentToken`) — i.e. no single hang; the tmux-heavy suite is cumulatively too slow for the 600s budget on slower runners. `v2-tests.yml` shards round-robin so PR CI hides it, but `coverage-hourly.yml` runs `go test ./pkg/... -short -race -timeout 600s` as ONE invocation and inherits this risk (plus `-race` overhead).

(Env note: `TestCaptureFullLog_CapturesPaneContent` also fails where `su-exec` is absent — expected per claude_session_adopt.go comments; not part of this finding.)

## Recommendation

1. Fix the isolation bug for `TestDeliverStartupKick_BobReceivesKick`: reset shared tmux/manager state (kill leftover sessions, per-test `HIVE_WORK_DIR`) or identify the preceding test that poisons kick delivery; consider asserting via a hook rather than a 60s pane poll.
2. Reduce `pkg/agent` wall time: shorten `time.Sleep`-based waits behind test seams, or split/parallelize the tmux-heavy tests; alternatively raise/shard the coverage-hourly timeout for `pkg/agent` explicitly.

## Priority
- Impact: medium (flaky gate + hourly coverage job at risk of timeout-panic, which zeroes the pkg/agent coverage report)
- Effort: medium

---
---

### LINKED ISSUE #4693: [quality] pkg/agent tests are not hermetic: live uid-map bleed-through, su-exec dependency, and a >10m suite that FAILs plain 'go test ./...'
## Finding

A full `go test -coverprofile=coverage.out ./...` run in `src/` reports **FAIL for `pkg/agent`** on any host running a live hive, which also poisons the package's coverage profile (functions with passing dedicated tests, e.g. `IsInternalUID`, report 0.0%). Three independent causes:

1. **Live uid-map bleed-through.** `NewManager` unconditionally loads `UIDMapPath = "/var/run/hive/uid-map.json"` (`src/pkg/agent/manager.go:1325`, `uidmap.go:13`). On a hive host that file exists and maps real agents, so `TestRefreshAgentTokenFor_Errors` finds a real UID for `guide` and its "no dedicated UID" assertion fails:
   `token_refresh_test.go:159: expected no-UID error, got <nil>`

2. **Host-binary dependency.** `TestCaptureFullLog_CapturesPaneContent` and `TestStart_LaunchClaude` fail with `exec: "su-exec": executable file not found in $PATH` when the sandbox helper is not installed.

3. **Suite exceeds the default timeout.** `pkg/agent` takes ~12.5 min (749s observed with `-timeout 15m`), so a plain `go test ./...` kills it at the default 10m panic timeout — the package FAILs even when every test would pass.

## Recommendation

- Make the uid-map path injectable: turn `UIDMapPath` into a var the test (or `NewManager` options) can override, or have `testManager`/`NewManager`-in-tests construct with an explicit empty `UIDMap`, so tests never read host state.
- Guard tmux/su-exec-dependent tests with `exec.LookPath("su-exec")` (and `tmux`) + `t.Skip` when absent, mirroring how hermetic suites elsewhere in the repo stub binaries.
- Set an explicit `-timeout` (e.g. 20m) for `pkg/agent` in CI, or split the slow tmux-integration tests behind a build tag / `-short` gate so the default run stays under 10m.

## Priority
- Impact: high — a false-FAIL on hive hosts masks real regressions and corrupts coverage reporting for the largest package in the repo
- Effort: medium

---
---

### LINKED ISSUE #4717: [quality] pkg/agent test suite exceeds 600s serially and fails plain 'go test ./...' — real sleeps need injectable seams
## Finding

`go test -short -count=1 -timeout 600s ./pkg/agent` **fails on a clean checkout** (verified today at 9cef51d):

```
panic: test timed out after 10m0s
FAIL    github.com/kubestellar/hive/pkg/agent   600.095s
```

The package's 440 tests accumulate real wall-clock sleeps from production code paths:

- `nudgeIfKickStalled` (manager.go:5372): raw `time.Sleep(textToEnterDelay)` (1s) per nudge
- `tmuxSendEntersForAgent`: `enterCount=3` × `enterDelay=300ms` per call
- `waitForCLIReadyForAgent` / `deliverStartupKick`: 2s poll intervals against a 60s `cliReadyTimeout`
- `dismissInferencePrompts`: 1s poll loops (goroutines from `launchInTmux` visible in the timeout dump)

v2-tests.yml already acknowledges this ("pkg/agent ≈ 429s of TEST time... the package is a serial 429s monolith") and works around it with 5-way `-run` sharding. The workaround keeps CI green, but:

1. **Plain `go test ./...` fails for every contributor** — the documented default workflow is broken locally.
2. **pkg/agent produces no coverage data in any unsharded run** (a timed-out package writes no profile), so local coverage analysis silently omits the largest package.
3. CI burns ~35 job-minutes per PR on sharding that a seam would eliminate.

## Recommendation

A seam already exists and proves the pattern: `Manager.promptDismissSleep` (manager.go:4942) is injected by `dismiss_prompts_fake_test.go` as `runtime.Gosched()`. Extend the same approach:

1. Add a `Manager.sleepFn func(time.Duration)` (or reuse/rename `promptDismissSleep`) and route `nudgeIfKickStalled`, `tmuxSendLiteralForAgent`-adjacent delays, `tmuxSendEntersForAgent`, and the CLI-ready/input-prompt poll loops through it.
2. Set it to a no-op (or `runtime.Gosched`) in `TestMain` for the whole package, restoring per-test override where a test asserts pacing.
3. Once the suite runs in seconds, the 5-way agent shard matrix in v2-tests.yml can collapse back into the normal package partition.

Test-only change plus a mechanical production refactor (replacing `time.Sleep(x)` with `m.sleep(x)`) — no behavior change at runtime.

## Priority
- Impact: high (broken default dev workflow, missing coverage data, CI cost)
- Effort: medium (mechanical, but many call sites; pacing-sensitive tests need review)

---
---

clubanderson:
Fixes #4717, Fixes #4693, Fixes #4688, Fixes #4685.

## Problem

The pkg/agent suite paid **production pacing** (1–3 s sleeps, 60–120 s poll deadlines) against stub CLIs that render instantly, **walked and chmod'd the live `/data` trees**, and **loaded the production uid-map** on hosts running a real hive. Together that blew `go test`'s default 600 s timeout and made results depend on the host it ran on.

## Fix — seams, not skips (pattern from #4659 / #4724)

- **#4717 (suite >600 s serial):** pacing constants (`clearBeforeKickDelay`, `enterDelay`, `textToEnterDelay`, `chunkDelay`, `staleCheckDelay`, CLI-ready/input-prompt poll+timeout, trust watcher pacing, diagnostic poll, `paneCaptureSleep`, `sessionReadyDelay`, `bobInputHandlerSettleDelay`) become package `var`s; `TestMain` shrinks them for the suite. Fork-heavy poll intervals deliberately stay ≥100 ms — single-digit-ms polls forked tmux subprocess storms (~51 s sys time). Production values unchanged; `TestBobInputHandlerSettleDelay`'s relationship pins still hold.
- **#4693 (non-hermetic uid-map/su-exec bleed):** `UIDMapPath` becomes a `var`; `TestMain` points it into the test temp tree, so a live host's `/var/run/hive/uid-map.json` can no longer route test tmux calls through su-exec and per-UID sockets. (Complements #4724's `forceSharedUID`.)
- **#4685 (TestFixPermissions walks real /data, 10+ min):** `WatchedHomeDirs`/`SharedRepoParent`/`GooseLogsDir`/mode+caps globs and `DevUID`/`NodeGID` are redirected into a temp tree; `TestFixPermissions_NoPanic` is rewritten as `TestFixPermissions_RepairsFixture` — a behavioral fixture asserting chmod repair (dir+file), the CWE-59 symlink skip, and idempotency. Mutation-checked: breaking the dir-chmod, file-chmod, or symlink guard each fails the test. Runs in 0.0 s.
- **#4688 (order-dependent Bob kick flake + coverage-gate timeout):** the order dependency was **not reproducible locally** (full suite passes in any order here); code reading found no in-suite shared state. The reporter's host ran a live hive, so the uid-map bleed above is the identified environmental vector — now sealed. The timeout half is fixed by the pacing shrink. Also hardened `TestConfirmMenuOption_SelectsOption`, which raced shell startup with a fixed 500 ms sleep (>2 s to first render on a zsh-dotfiles host — same fixed-sleep-vs-render-latency mechanism): it now polls for the menu actually rendering.
- `TestWatchedHomeDirsIncludesBob` keeps pinning the **production** default via a TestMain snapshot.

## Numbers (local, macOS)

| | before | after |
|---|---|---|
| `go test ./pkg/agent -short` serial | 462 s (fails >600 s on slower hosts) | **279 s** |
| coverage | 90.9 % | **90.9 %** (floor 89) |

No CI sharding changes on purpose — the 5-way pkg/agent shard in v2-tests.yml could collapse later once runners confirm the margin.

### DIFF
--- a/src/pkg/agent/authprobe_test.go
+++ b/src/pkg/agent/authprobe_test.go
@@ -579,7 +579,7 @@ func TestFixEntryWithSyntheticOwnership(t *testing.T) {
 		name:  "dir",
 		mode:  0o700 | os.ModeDir,
 		isDir: true,
-		sys:   &syscall.Stat_t{Uid: DevUID, Gid: NodeGID},
+		sys:   &syscall.Stat_t{Uid: uint32(DevUID), Gid: uint32(NodeGID)},
 	}, logger)
 	info, err := os.Stat(dir)
 	if err != nil {
@@ -596,7 +596,7 @@ func TestFixEntryWithSyntheticOwnership(t *testing.T) {
 	fixEntry(file, fakeFileInfo{
 		name: "file",
 		mode: 0o600,
-		sys:  &syscall.Stat_t{Uid: DevUID, Gid: NodeGID},
+		sys:  &syscall.Stat_t{Uid: uint32(DevUID), Gid: uint32(NodeGID)},
 	}, logger)
 	info, err = os.Stat(file)
 	if err != nil {
@@ -610,7 +610,7 @@ func TestFixEntryWithSyntheticOwnership(t *testing.T) {
 	fixEntry(file, fakeFileInfo{
 		name: "other-owner",
 		mode: 0o600,
-		sys:  &syscall.Stat_t{Uid: uint32(DevUID + 100), Gid: NodeGID},
+		sys:  &syscall.Stat_t{Uid: uint32(DevUID + 100), Gid: uint32(NodeGID)},
 	}, logger)
 	fixEntry(file, fakeFileInfo{
 		name: "root-owned",
--- a/src/pkg/agent/bob_test.go
+++ b/src/pkg/agent/bob_test.go
@@ -15,15 +15,19 @@ import (
 //
 // bob derives its state dir as path.join(os.homedir(), ".bob") and agents run
 // with HOME=/data/home, which is root-created — so the watcher must own it.
+//
+// It checks productionWatchedHomeDirs — the snapshot TestMain takes before
+// pointing WatchedHomeDirs into the hermetic temp tree — so the pin stays on
+// the production default, not the test override.
 func TestWatchedHomeDirsIncludesBob(t *testing.T) {
 	const bobStateDir = "/data/home/.bob"
-	for _, dir := range WatchedHomeDirs {
+	for _, dir := range productionWatchedHomeDirs {
 		if dir == bobStateDir {
 			return
 		}
 	}
 	t.Fatalf("WatchedHomeDirs must contain %q so bob can create its state dir; got %v",
-		bobStateDir, WatchedHomeDirs)
+		bobStateDir, productionWatchedHomeDirs)
 }
 
 // TestPermissionsWatcher_BobNestedTree covers the nested case explicitly: bob
--- a/src/pkg/agent/helpers_coverage_test.go
+++ b/src/pkg/agent/helpers_coverage_test.go
@@ -127,13 +127,105 @@ func TestUIDMapSave_Success(t *testing.T) {
 // ---------------------------------------------------------------------------
 
 func TestEnsureWatchedDirs_NoPanic(t *testing.T) {
-	// WatchedHomeDirs live under /data which usually can't be created here;
-	// the function logs warnings and never panics.
+	// TestMain points WatchedHomeDirs into the test temp tree, so this
+	// exercises the real mkdir path hermetically instead of warning against
+	// unwritable /data roots.
 	ensureWatchedDirs(discardLogger())
 }
 
-func TestFixPermissions_NoPanic(t *testing.T) {
+// TestFixPermissions_RepairsFixture used to be TestFixPermissions_NoPanic,
+// which called fixPermissions against whatever the HOST had: on a live hive it
+// walked (and could chown/chmod) real /data agent trees for 10+ minutes from a
+// unit test (#4685). TestMain now points every walk root into the test temp
+// tree, and this test upgrades "doesn't panic" to asserting the repair
+// behavior on a fixture it builds itself.
+func TestFixPermissions_RepairsFixture(t *testing.T) {
+	origWatched := WatchedHomeDirs
+	origShared := SharedRepoParent
+	origGoose := GooseLogsDir
+	origDevUID, origNodeGID := DevUID, NodeGID
+	t.Cleanup(func() {
+		WatchedHomeDirs = origWatched
+		SharedRepoParent = origShared
+		GooseLogsDir = origGoose
+		DevUID, NodeGID = origDevUID, origNodeGID
+	})
+
+	// fixEntry only chmods entries owned by DevUID (it refuses to touch other
+	// users' files) and skips the chown when uid/gid already match. Point the
+	// guard at the uid/gid this test process actually runs as, so the repair
+	// path is exercised rather than skipped.
+	DevUID = os.Getuid()
+	NodeGID = os.Getgid()
+
+	root := t.TempDir()
+	watched := filepath.Join(root, "watched")
+	SharedRepoParent = filepath.Join(root, "home")
+	GooseLogsDir = filepath.Join(root, "goose-logs")
+	WatchedHomeDirs = []string{watched}
+
+	// A too-narrow directory and file inside a watched root, and a repo clone
+	// under SharedRepoParent created the way the entrypoint does (0755 → group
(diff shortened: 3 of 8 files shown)


clubanderson:
/lgtm
/approve

Self-merging per maintainer lane: all CI lanes green (5 pkg/agent shards, rootless/rootful startup, quadlet dry-run), local serial suite 462s → ~275s across three post-rebase runs, coverage held at 90.9% (floor 89), fixture test mutation-checked (dir-chmod, file-chmod, symlink guard all caught). Trial-rebased clean onto e9c07779.

PR #4743, opened 2026-08-25
fix(dashboard): surface and enable configured disabled agents

### LINKED ISSUE #4729: Dashboard has no way to enable a configured-but-disabled agent
## Summary

There is no way, from the dashboard, to enable an agent that exists in config but is `enabled: false`. The API to do it exists (`PUT /api/config/agent/{name}/general`), but nothing in the UI surfaces disabled agents or offers a control to flip them on. Operators can only reach these agents by editing `hive.yaml` directly or hand-crafting the API call.

This affects every disabled entry the shipped `hive.yaml.example` ships out of the box: `ci-maintainer`, `architect`, `outreach`, `sec-check`, `guide`, `strategist`, `brainstorm` — none of these have a dashboard path to turn on.

## Why this is surprising

The sidebar's **AGENT** tree looks like the natural place to find them, but it only renders agents present in `/api/status`, and disabled agents are absent from that endpoint entirely — there is nothing to click. The obvious next place to look is the **+ agent** button, but that opens **Create Agent** (`openAddAgentDialog`), which is exclusively an import/creation flow (YAML paste or a `raw.githubusercontent.com` URL). Using it against an already-configured-but-disabled agent name would create a conflicting second definition rather than enabling the existing one — it is the wrong tool, not a degraded version of the right one.

So an operator who wants to enable, say, `guide` has to already know the agent exists in the shipped config (not documented anywhere in-app), find its exact name, and either edit `hive.yaml` on the host or construct:

```
PUT /api/config/agent/guide/general
{"enabled": true, "backend": "claude"}
```

None of which the dashboard hints at.

## Suggested fix

Something in the Agent sidebar tree (or a dedicated section) that lists **all** agents present in config, not just running ones — disabled entries shown greyed out / collapsed with an **Enable** action, the same way many admin UIs show a "disabled" row instead of hiding it. Clicking Enable would open the same config dialog `openConfigDialog('agent', name)` already uses for editing a live agent, just pre-populated from the existing (disabled) config rather than a blank Create-Agent form.

This also happens to be the natural place to surface the field-ownership markers (`model_owner`, etc.) that already exist server-side — an "Enable" flow is exactly when an operator is choosing a model/backend for the first time and would benefit from seeing that choice will be protected across future ACMM pack applies.

## Not asking for

Nothing here is about *creating* new agents — that flow (`Create Agent` / Import) already works well. This is specifically about the shipped-but-dormant agents already sitting in every hive's config with no UI path to turn them on.

Danathar:
Closes #4729.

## What operators saw

Agents configured with `enabled: false` disappeared completely from the dashboard because `/api/status` only described processes known to the agent manager. The **+ agent** control could not fill the gap: it is an import/create flow, so using it for an existing configured name risks a conflicting definition instead of enabling the dormant one.

There was a second blocker behind that UI gap. `GET /api/config/agent/{name}` first found the disabled agent in `Config.Agents`, but then returned 404 when `AgentMgr.GetStatus(name)` found no process. That meant the existing configuration dialog could not be populated for exactly the agents that needed an enable path.

### Publish configured-agent inventory separately from runtime status

Both full status snapshots and fast agent-status SSE events now include a small, secret-free `configuredAgents` inventory. It contains the identity and ordering fields needed by the sidebar, the enabled bit, and the existing model/backend ownership markers. Runtime state remains in the existing `agents` array, so disabled entries do not masquerade as crashed processes or affect health, cards, token accounting, or agent-detail behavior.

The inventory is sorted with the same configured sort order/name rules as the runtime list. The sidebar merges only configured-disabled entries that are absent from runtime status, rendering them as muted rows with a visible **Enable…** action. Existing sidebar groups continue to work because the disabled rows participate in the same name-based layout mapping.

### Make config reads work without a process

`handleAgentConfigGet` now treats process status as optional after the configuration lookup succeeds. When a process exists, live backend/model overrides, the last prompt, and runtime CLI pin state still win as before. When it does not, the response uses the persisted config and neutral runtime-only defaults instead of returning 404.

The General response now also carries `enabled`, `modelOwner`, and `backendOwner`. The ownership values are surfaced in the disabled-row tooltip, so operators can see whether pack or operator choices protect the backend/model before enabling the agent.

### Reuse the existing save/reconcile path

The existing agent General tab now has an **Enabled** switch for configured agents. **Enable…** opens that tab pre-populated from the dormant agent's real configuration; toggling Enabled and clicking Save sends the existing general-config PUT. No new mutation endpoint or duplicate creation path was introduced. The handler's existing `ReconcileAgents`, `Start`, governor update, persistence, and status refresh logic makes the newly enabled agent live.

The switch is intentionally absent from Create Agent, whose payload already creates enabled agents.

## Tests

- Added a regression test proving a configured disabled agent with no manager process returns 200 from the config GET, including `enabled: false` and ownership markers.
- Added coverage for the complete configured-agent inventory and deterministic ordering.
- Added an embedded-dashboard wiring test for inventory consumption, disabled-row rendering, the Enable action, and the General-tab switch.
- Ran `go test ./pkg/dashboard` successfully.

— hive: backend=codex

### DIFF
--- a/src/pkg/dashboard/api.go
+++ b/src/pkg/dashboard/api.go
@@ -2528,17 +2528,13 @@ func (s *Server) handleAgentConfigGet(w http.ResponseWriter, r *http.Request) {
 	}
 
 	proc, err := s.deps.AgentMgr.GetStatus(name)
-	if err != nil {
-		jsonError(w, err.Error(), http.StatusNotFound)
-		return
-	}
 
 	cli := agentCfg.Backend
-	if proc.BackendOverride != "" {
+	if err == nil && proc.BackendOverride != "" {
 		cli = proc.BackendOverride
 	}
 	model := agentCfg.Model
-	if proc.ModelOverride != "" {
+	if err == nil && proc.ModelOverride != "" {
 		model = proc.ModelOverride
 	}
 
@@ -2594,7 +2590,12 @@ func (s *Server) handleAgentConfigGet(w http.ResponseWriter, r *http.Request) {
 		models[modeName] = ""
 	}
 
-	lastPrompt := proc.LastKickMessage
+	lastPrompt := ""
+	pinnedCLI := agentCfg.CLIPinned
+	if err == nil {
+		lastPrompt = proc.LastKickMessage
+		pinnedCLI = pinnedCLI || proc.PinnedCLI != ""
+	}
 
 	// Read restrictions from agent work dir files
 	restrictions := s.loadAgentRestrictions(name)
@@ -2615,11 +2616,14 @@ func (s *Server) handleAgentConfigGet(w http.ResponseWriter, r *http.Request) {
 
 	jsonResponse(w, map[string]interface{}{
 		"general": map[string]interface{}{
+			"enabled":          agentCfg.Enabled,
 			"launchCmd":        launchCmd,
 			"displayName":      displayName,
 			"description":      agentCfg.Description,
-			"cliPinned":        agentCfg.CLIPinned || proc.PinnedCLI != "",
+			"cliPinned":        pinnedCLI,
 			"cliPinValue":      cli,
+			"modelOwner":       agentCfg.ModelOwner,
+			"backendOwner":     agentCfg.BackendOwner,
 			"staleTimeout":     staleTimeout,
 			"restartStrategy":  restartStrategy,
 			"model":            model,
--- a/src/pkg/dashboard/disabled_agent_enable_test.go
+++ b/src/pkg/dashboard/disabled_agent_enable_test.go
@@ -1,0 +1,64 @@
+package dashboard
+
+import (
+	"net/http"
+	"strings"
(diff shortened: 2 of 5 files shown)


clubanderson:
Verified: disabled agents surface as a separate secret-free configuredAgents inventory (muted sidebar rows, honest 'Configured but disabled' state — never fabricated into the runtime agents array), the Enable path reuses the existing owner-gated general-config PUT (requireOwnerRole; handler already reads the enabled key and runs ReconcileAgents(EnabledAgents()) + Start — no new mutation endpoint), config GET now honestly returns persisted config when no process exists instead of 404. No on*= attributes (data-action dispatch throughout), check-inline-js guard passes, tests pin GET/inventory/UI wiring. Trial-merged over v4 head (post-#4710) cleanly with full pkg/dashboard suite green.

/lgtm
/approve

PR #4748, opened 2026-08-25
fix(scheduler): surface held PRs as occupied work

### LINKED ISSUE #4744: Hold-gated agents don't check open PRs before picking overlapping work, guaranteeing conflicts once review latency is hours (not minutes)
## Summary

Two hold-gated `quality` PRs opened hours apart both independently targeted the **same** uncovered-code cluster from the tracking issue they were both working off of, writing near-duplicate tests for the same functions under different names. Because the operator hadn't reviewed either yet — normal review latency, not a slow operator, people sleep — the second PR to be *merged* landed first and left the other conflicting.

This isn't really a git-conflict bug; git behaved exactly as it should. It's a **coordination gap**: nothing in the write path appears to check "is this ground already claimed by an *open* PR?" before a kick starts drafting new coverage. It only seems to reason about what's already in `main`.

## Observed, with timestamps

Self-hosted L3 hive. **All PR/issue numbers below refer to the observed repo, `Danathar/atomic-image-builder` — not to this repo.** `quality` on a 2h cadence, all three touching the same file (`tests/test_atomic_image_builder.py`):

| PR | Opened | Merged | Title |
|----|--------|--------|-------|
| Danathar/atomic-image-builder#25 | 2026-08-24 19:52 | 2026-08-24 21:55 | Add tests for Gum.choose/filter/table widgets |
| Danathar/atomic-image-builder#26 | 2026-08-25 02:23 | 2026-08-25 10:17 *(after manual conflict resolution — see below)* | Test Gum widgets, do_build repo-exists paths, **and main()/run_main** |
| Danathar/atomic-image-builder#27 | 2026-08-25 08:07 | 2026-08-25 10:06 | Add tests for **run_main, main()**, and landing screens |

Both PR 26 and PR 27 explicitly targeted the `run_main()`/`main()` entrypoint cluster from their shared tracking issue — the titles say so outright, and the bodies confirm it: near-identical coverage of `run_main()`'s happy path, `main()`'s `--version` handling, and the `ScreenBack`/`CommandError`/`KeyboardInterrupt` → exit-code mapping, each under a different test name and a slightly different mocking style. Neither PR body references the other.

PR 27 was opened at 08:07 — nearly six hours after PR 26 was opened at 02:23, and while PR 26 was still open and unmerged (it did not merge until 10:17) — so the overlapping work was visible on the repo as an open PR for the entire time PR 27 was being drafted. PR 27 then merged first, at 10:06 — at which point PR 26 (still unreviewed) became `mergeable: CONFLICTING` against the two duplicate-coverage blocks, and merged only after the conflict was resolved by hand at 10:17. Resolving it meant manually diffing both PRs' additions, discarding PR 26's five now-redundant tests, keeping its eight genuinely new ones (`do_build` repo-exists branches, `Gum.pager`, `Gum.enter_to_continue`, `Gum.spinner_capture`'s success path — none of which PR 27 touched), and re-running the full suite to confirm nothing broke (474 passed).

## Why "review faster" isn't the fix

The instinctive read is "the operator should have merged PR 26 before PR 27 was ever opened." But hold-gated PRs exist precisely so a human reviews before anything lands — and humans sleep, work day jobs, and don't watch a dashboard continuously. A multi-hour gap between a PR opening and a human looking at it is the **normal**, expected operating mode of this ACMM tier, not an edge case to design around. Any fix that only works if the operator reviews within one governor cycle isn't really a fix.

## Suggested fix

Before a hold-gated agent starts drafting new coverage against a tracking issue's clusters, check **open** PRs on the repo (not just `main`) for ones that already reference the same target functions/cluster — a cheap `gh pr list --state open` plus a look at titles/bodies, or a lighter convention like having each PR's body explicitly claim its cluster (`Covers: #24 cluster 3`) so the next kick can grep for "is cluster 3 already claimed by an open PR?" before choosing it. If the ground is already claimed, either skip it for a different cluster or note the collision and stand down, rather than writing overlapping tests that will only surface as a conflict once one of the two merges.

This generalizes past this one issue-tracker pattern: anywhere an agent works from a shared checklist of gaps/findings across multiple kicks without a merge in between, the same race is possible.

Danathar:
Closes #4744.

## Why

The governor already enumerates every open PR, but the hold gate splits that population before a kick is rendered:

- ordinary open PRs remain in `ActionableResult.PRs`; and
- a PR carrying `hold` moves into `ActionableResult.Hold` and disappears from `${PR_LIST}`.

That split is correct for merge/actionability decisions, but it accidentally erased coordination evidence. Hold-gated agent policies also tell agents not to rebuild the queue with `gh pr list`, so a PR could wait for human review for six hours and be completely invisible to the next quality kick. The second kick then analyzed only the target branch, selected the same uncovered functions, and produced a conflict that could not surface until one PR merged.

The existing duplicate-PR ledger cannot safely solve this by filtering the whole tracking issue. It suppresses strong “Fixes #N” claims, while partial “Refs #N” claims intentionally remain actionable because other clusters in that tracker still need work. What was missing is the finer preflight: show the open partial work, inspect what it actually claims, and choose only disjoint ground.

## What changed

Every PR-capable agent at hold-gated ACMM levels (L3–L5) now receives an **Open hold-gated PR coordination** section before its resolved policy body. The section is built from the already-enumerated `Hold.Items` snapshot and requires the agent to:

1. compare its intended files, functions, and tracker cluster with each plausibly related held PR;
2. inspect the PR's title, body, files, and diff using `gh pr view` / `gh pr diff` when the title alone is not decisive;
3. choose a disjoint cluster or stand down when ground overlaps; and
4. make the next PR's title and body state its exact claim for the following kick.

This is injected at the `BuildAgentMessage` boundary rather than added only to the stock quality template. That means scheduled kicks, dashboard/manual kicks, user-edited templates, ACMM templates, repo-sourced prompts, embedded defaults, and hardcoded fallbacks all receive the same preflight. Advisory and issue-only agents do not receive it, and L2 manual / L6 auto-merge policies are unchanged.

Held issue entries are excluded; only open held PRs are rendered. Their titles pass through the same ioscan enforcement as ordinary PR titles. The list uses the existing bounded kick size; if more PRs exist than fit, the prompt reports the omitted count and requires the writer to stand down because unseen occupied ground cannot be proven disjoint.

The existing “do not run `gh pr list`” policy remains intact: the governor's supplied snapshot is authoritative and does not cost another GitHub enumeration.

## Verification

- `go test ./pkg/scheduler`
- `go test ./pkg/scheduler ./pkg/github ./cmd/hive`
- Regression coverage proves the preflight reaches a customized template through the manual-kick renderer, precedes work selection, includes held PRs that are absent from `PRs`, excludes held issues, respects ACMM/capability boundaries, scans untrusted titles, and fails closed when the snapshot must be truncated.

— hive: backend=codex


clubanderson:
Review record: verified against #4744's actual failure mode. Held PRs stay solely in `Hold` (no double-count with `${PR_LIST}`), and the hold releases naturally — once reviewed/merged the PR leaves the hold snapshot, so nothing blocks forever. Non-held scheduling unchanged (full pkg/scheduler suite green, 91.0% ≥ floor 90; pkg/github + cmd/hive green). Mutation-checked the new branch twice: gutting `isHoldGatedPRAgent` → 2 tests red; disabling the issue-type skip → issue-leak assertion red; both restored green. Injection scanning of untrusted titles, 70-rune truncation, fail-closed on truncation and on Critical ioscan verdicts, and the deferred injection wrapping every template-resolution path are all pinned by tests. DCO present on both commits.

/lgtm
/approve

Issue #4750, opened 2026-08-25
Does the `bob` backend actually read `~/CLAUDE.md`? If not, bob contributors run with no hive knowledge


Danathar:
## Question

`bin/contributor-agent.sh` links the downloaded knowledge export (`agent.md`)
under whatever filename each CLI backend reads. Every supported relay backend
has an explicit case except `bob`, which falls through to the `*` branch and
therefore gets **only** `~/CLAUDE.md`:

```sh
  codex)  ln -sf "$AGENT_MD" "${HOME}/AGENTS.md"; ln -sf "$AGENT_MD" "${HOME}/CLAUDE.md" ;;
  pi)     ln -sf "$AGENT_MD" "${HOME}/AGENTS.md"; ln -sf "$AGENT_MD" "${HOME}/CLAUDE.md" ;;
  agy)    ln -sf "$AGENT_MD" "${HOME}/CLAUDE.md" ;;
  *)      ln -sf "$AGENT_MD" "${HOME}/CLAUDE.md" ;;   # <-- bob lands here
esac
```

I have not been able to confirm what context filename bobshell reads, so this
is a question rather than a bug report. If bob does not read `CLAUDE.md`, then
every contributor running `AGENT_BACKEND=bob` is working without the hive
knowledge export — the export is still downloaded and refreshed every 10
minutes, it just never reaches the model.

## Why this is worth checking

This is the exact failure mode already fixed for Goose in #2393: the file was
being fetched correctly and linked under names Goose does not read, so the
knowledge silently never arrived. The `*` fallback makes it silent by
construction — nothing errors, nothing logs, and the contributor sees a
perfectly healthy-looking relay.

`bob` is not a fringe backend. It is installed in the contributor image
(`Dockerfile.contributor`, bobshell 1.0.6), has its own readiness detection in
`contributor-relay.sh`, a mandatory `BOBSHELL_API_KEY` startup gate, and a
dedicated `--model` exclusion in `contributor-agent.sh` ("Verified against
bobshell 1.0.6").

## What would settle it

Anyone with a `BOBSHELL_API_KEY` can answer this in one session:

1. Start a bob relay and confirm `~/CLAUDE.md` exists and is a symlink to the export.
2. Ask bob something answerable only from the hive knowledge export.
3. If it has no knowledge of it, find the filename bobshell does read and give
   `bob` an explicit case, as was done for Goose.

## Related

- #2393 — same class of bug, fixed for Goose
[...]

Issue #2393, opened 2026-08-01
Feedback from building a downstream contributor front end (donate-clanker): 2 likely bugs + DX suggestions


castrojo:
Hi Hive maintainers 👋

We (the [projectbluefin](https://github.com/projectbluefin) folks) built a downstream contributor front end on top of Hive and wanted to send back what we learned in **one** issue rather than a pile of them. Everything below was verified against `v2` at `1a05c1bdfa8578831da233f18b2c4b13be9152ef` with file + line citations so you can check each item in seconds.

We're guests here and assume you had good reasons for every design choice. Where we say "this surprised us", please read it as "here's what we hit and what would have helped" — not a complaint. **We're happy to send PRs for any of this.**

---

## TL;DR

| # | Kind | One-liner | Cite |
|---|---|---|---|
| 1 | 🐞 likely bug | Goose knowledge-export symlinks use filenames Goose doesn't read | `bin/contributor-agent.sh:225-234` |
| 2 | 🐞 likely bug | `wsTokenRefreshPeriod` unused; server never sends `token_refresh`, relay has a handler for it | `contribute_ws.go:30`, `contributor-relay.sh:733-739` |
| 3 | 🙂 DX | Goose `config.yaml` is unconditionally overwritten each start | `bin/contributor-agent.sh:229-232` |
| 4 | 🙂 DX | No documented extension seam in the contributor entrypoint | `v2/Dockerfile.contributor:95` |
| 5 | ❓ question | Image defaults to `:latest`; launcher requires being on moving `v2` HEAD | `Justfile:8,20-32` |
| 6 | 🙂 DX | Contributor restrictions block read-only lookups an agent needs to avoid dupe work | `config/restrictions/contributor-default.json`, `bin/gh-wrapper.sh:112-121,331-341` |
| 7 | ⭐ our top ask | `task_complete` can't report a PR URL, yet it applies a 168 h issue cooldown | `contribute_ws.go:134,641-680` |
| 8 | 📝 nit | `labels` / `restrictions` envelope fields are never populated | `contribute_ws.go:81,85` vs `923-936` |

---

## First: what worked really well

This is genuine — the following four designs are why we could build what we built, mostly without patching Hive:

- **`HIVE_CONTRIBUTOR_IMAGE` as a customization seam** (`Justfile:8`). Being able to swap the contributor image while keeping the whole launcher flow intact is the single thing that made a downstream front end possible at all. Thank you for putting that env var there.
[...]

PR #4751, opened 2026-08-25
📖 docs: record which instruction filename each contributor backend reads

Danathar:
## What

Documents the backend → instruction-filename mapping in
`src/docs/contributor-relay.md`, next to the existing `AGENT_BACKEND` row.

The relay downloads the hive knowledge export to a single `agent.md` and
symlinks it under whatever filename the chosen CLI actually reads
(`bin/contributor-agent.sh`). That mapping existed only as a `case` statement
and was documented nowhere.

## Why

Getting the filename wrong fails silently: the export is downloaded and
refreshed every 10 minutes, nothing errors, nothing logs — the model just never
sees it. That is exactly how #2393 went unnoticed for Goose, where the file was
linked under names Goose does not read.

The table also makes the `*` fallback visible: any backend without an explicit
case gets `CLAUDE.md` only. `bob` is currently the one supported relay backend
in that position, which is what prompted #4750.

## Scope

Docs only — no behavior change. Exempt from the changelog gate
(`src/docs/` per `changelog-reminder.yml`), and the file is already listed in
`src/docs/README.md`.

Refs #4750


clubanderson:
Verified the table against bin/contributor-agent.sh's case statement (lines ~387-434): every backend row (claude/litellm, copilot, goose, codex, pi, agy) and the `*` fallback matches the actual symlinks exactly. Docs-only, accurate, CI green.

/lgtm
/approve

PR #4756, opened 2026-08-25
🐛 fix(dashboard): make a disabled agent's row say it is configurable, and show its mode

Danathar:
Follow-up to #4743 (merged), from an operator reading the sidebar rather than
the code.

## "Enable…" hid the thing it does

The button already opened the **same** agent config dialog the `⚙` on an
enabled row opens — `openConfigDialog` with `data-tab="General"`. Enabling is
one field inside that dialog, not a one-click flip. But a bare "Enable…" reads
as a state toggle, so the answer to "can I configure a disabled agent before
turning it on?" looked like *no* when it was already *yes*.

Lead the label with the same gear the enabled rows use: `⚙ Enable…`.

## A disabled row could not show its mode

`FrontendConfiguredAgent` carried no mode at all. A disabled agent has no
runtime entry, so the sidebar could not say what it **would** do once enabled —
arguably the one thing worth checking before enabling something. `quality`
showing 🔧 (ISSUES_AND_PRS) next to an advisory 📝 agent is exactly the signal
that matters, and disabled agents were missing it.

Resolved the same way `buildAgents` resolves it for a running agent: an
explicit per-agent `Mode` wins, otherwise the ACMM level default.

## Tests

`TestConfiguredAgentsCarryResolvedMode` covers **both** branches. The
level-default branch is the one that matters: without it an agent that sets no
`Mode` reports empty, the badge renders nothing, and the result is dead markup
that reviews clean.

`TestDisabledAgentSidebarEnableWiring` pinned the old title prose. It now
asserts the durable part — the dialog opens on the General tab — plus the gear
label.

Full `pkg/dashboard` suite passes; `go vet` clean.

Refs #4729

### DIFF
--- a/src/pkg/dashboard/disabled_agent_enable_test.go
+++ b/src/pkg/dashboard/disabled_agent_enable_test.go
@@ -53,12 +53,52 @@ func TestDisabledAgentSidebarEnableWiring(t *testing.T) {
 		`window._configuredAgents = data.configuredAgents || [];`,
 		`if (!configured.enabled && !runtimeNames.has(configured.name))`,
 		`class="oc-nav-item disabled-agent"`,
-		`data-tab="General" title="Configure and enable`,
+		// Wiring, not prose: the button must open the agent config dialog on the
+		// General tab. The visible label deliberately leads with a gear so the
+		// row reads as configurable rather than as a one-click state toggle.
+		`data-tab="General" title="Configure `,
+		`>⚙ Enable…</button>`,
 		`id="cfg-agent-enabled"`,
 		`data-key="enabled"`,
 	} {
 		if !strings.Contains(html, snippet) {
 			t.Errorf("index.html is missing disabled-agent enable wiring %q", snippet)
 		}
+	}
+}
+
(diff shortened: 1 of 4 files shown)


clubanderson:
Verified: mode resolution in buildConfiguredAgents is byte-for-byte the same DefaultAgentMode/ParseAgentMode logic buildAgents uses for running agents (status_builder.go:584-590), so the disabled row shows what the agent WOULD do without fabricating runtime state; the payload stays secret-free (adds only mode/modeEmoji enum strings). Enable path unchanged — the button still dispatches openConfigDialog via data-action (no on*= handlers) to the owner-gated config dialog. The oc-nav-mode class and modeEmoji() helper already exist for enabled rows, so the badge renders consistently. TestConfiguredAgentsCarryResolvedMode pins both the explicit-override and level-default branches (the dead-markup trap), and the wiring test now asserts the durable dialog/tab contract plus the ⚙ label. Trial-merged with #4763 on v4 head: dashboard suite green (84.7%), check-inline-js guard passes.

/lgtm
/approve

Issue #2323, opened 2026-07-31
🐛 Governor "Period (days)" and "Critical Percent" are persisted but never read


clubanderson:
Both settings are exposed in Governor config and persisted, but nothing in the codebase reads them back — the values they appear to control are hardcoded.

## Evidence
- Persisted at `v2/pkg/config/config.go:1131-1132`
- The budget window is a hardcoded **7 days** — `v2/pkg/governor/governor.go:83`
- The critical/warn threshold is a hardcoded **90%** — `v2/pkg/governor/governor.go:87`

So a user who sets `Period (days) = 30` still gets a 7-day window, and one who sets `Critical Percent = 75` still warns at 90%. The setting saves successfully and appears to take effect.

## Why this matters
Silent no-op settings are worse than absent ones. This came up while writing cost-management docs (PR #2310) after a user overspent — someone tightening `Critical Percent` to get earlier warning would believe they had done so and get none.

## Options
1. Wire both through to `governor.go` (replacing the hardcoded constants).
2. Remove them from the UI and config until implemented.

Related: #2322 (Governor advertises model-downgrade behavior that also doesn't exist). Both are the same class — Governor UI describing behavior the engine doesn't implement — and may be worth fixing together.

PR #4763, opened 2026-08-25
🐛 fix(dashboard): clarify token budget periods

### LINKED ISSUE #4762: 🐛 bug: strange stuff in token cost section of hive dashboard
## What happened?

I looked at the new cost subsection of https://hosted-llm-d-incubation-llm-d-fast-m-yfzz.apps.fmaas-vllm-d.fmaas.res.ibm.com/ about token spend, and noticed several things that look odd.

- The text at the bottom explains something about pricing. This is misplaced, it belongs up with the other dollar cost stuff.
- The word "window" is used to mean two or three different things. The text "Window Aug 25, 03:19 AM → Sep 1, 03:19 AM", for example, uses the word "window" to mean the whole horizontal range of the graph. OTOH, the text "Budget window history — used vs. limit per window, labeled by reset time" uses the word "window" to mean some shorter interval. This mixture of meanings is confusing. In my opinion, the more natural meaning for "window" is an individual inter-reset interval. BTW, this terminology issue applies in other places in the cost section too.
- That first "window" text covers one week. I do not know where that quantity is coming from. It is NOT my configured budget reset period --- that is 2 days. It is NOT the horizontal extent of the graph.

Image

- The tooltips on the individual bars also cite a window of one week, even though one bar seems to cover two days.

Image

- The quantities shown in the bars are much lower than the token quantities shown above in the dollar cost section.

Image

## Expected behavior

I hope the above makes this clear.

## Reproduction steps

1. Look at the token cost section of https://hosted-llm-d-incubation-llm-d-fast-m-yfzz.apps.fmaas-vllm-d.fmaas.res.ibm.com/

## Environment

- Branch/version: `v4` / commit a8250a01f7e67ab646074ed70893c8610557931b
- Deployment method: Hive Hub

## Logs or screenshots

Included above.

Danathar:
## Summary

- make dashboard budget dates and projections honor the configured `governor.budget.period_days` value
- distinguish wall-clock time until reset from estimated time until the token allowance is exhausted
- use “range” for Cost chart selections and “budget period” for inter-reset accounting intervals
- explain why per-period budget bars are lower than the all-time token totals above them
- move pricing caveats beside the dollar estimates they describe

## Root cause

The governor already uses the configured budget period, but `buildBudget` independently calculated the displayed end time with the seven-day fallback constant and projected spend over a hard-coded 168 hours. A two-day budget was therefore enforced every two days while the dashboard labeled it as a week.

The frontend then passed `HOURS_REMAINING`—the estimated number of hours until the remaining token allowance is consumed—to copy that described it as time until the reset. Separately, the Cost panel used “window” for both selectable chart ranges and budget reset periods, and put all-time token tables directly above per-period bars without stating their different scopes.

## Changes

- derive displayed start/end bounds and projection duration from `Governor.BudgetWindow()`, the enforcement source of truth
- add additive `WINDOW_HOURS_REMAINING` status data for the actual reset countdown while retaining the existing allowance-lifetime field
- base the Cost and Governor budget phase labels on the reset countdown
- estimate a missing Cost-chart range total over the full selected range rather than one bucket
- clarify range/period labels, tooltips, empty states, and token-total scope
- move the provider/self-hosted pricing disclaimer to the top of the dollar-cost section
- correct the dashboard help text that still claimed budget periods were fixed at seven days
- add backend and embedded-UI regression coverage plus an operator-facing changelog entry

## Compatibility

The status response change is additive. Existing `BUDGET_WEEKLY`, `HOURS_REMAINING`, and `PROJECTED_WEEKLY` fields remain available; their legacy names are retained for existing clients.

## Testing

- `env CGO_ENABLED=0 GOCACHE=/tmp/hive-4762-gocache go test ./pkg/dashboard ./pkg/governor`
- focused configured-period and Cost-panel regression tests
- `git diff --check`

Fixes #4762

— hive: backend=codex


clubanderson:
Verified against the governor: buildBudget now derives bounds and projection duration from Governor.BudgetWindow() (governor.go:1065), which is the enforcement source of truth for period_days — the old code re-added the 7-day constant to ResetAt and projected over a hard-coded 168h, exactly as the root-cause states. WINDOW_HOURS_REMAINING is additive (HOURS_REMAINING/BUDGET_WEEKLY/PROJECTED_WEEKLY retained); both renderBudgetBar and costTokenBudgetHtml still read the single window._lastBudget object, no second computation. The help-text correction is right too: period_days/critical_pct have been honored since #2323, so the old 'stored but ignored' caveat was stale. cfg.windowMs (not bucketMs) is the correct span for the run-rate estimate — it now matches the total's scope. Unconfigured states stay honest (hasWindow=false → 0, frontend clamps, TestBudgetPhaseToleratesMissingBounds). check-inline-js guard passes; trial-merged with #4756 on v4 head — dashboard 84.7%, governor 92.9%. Fixed the CHANGELOG conflict with the concurrently merged #4785 via a merge commit on the branch.

/lgtm
/approve

PR #4781, opened 2026-08-26
🐛 fix(agent): prevent hermetic tmux cleanup race

### LINKED ISSUE #4778: [Coverage] packages below 90% threshold
## Coverage gate failed

- **Measured:** `v4` @ `a8250a01f7e67ab646074ed70893c8610557931b`
- **Run:** https://github.com/kubestellar/hive/actions/runs/32946464170
- **Total coverage:** 85.2%

### What the gate reported

```
agent: TESTS FAILING
```

Danathar:
Fixes #4778

## Summary

The hourly coverage gate successfully measured `pkg/agent` at 90.8% in its
race-enabled run, then failed its second scoring run when
`TestHermeticWaitForLegacySessionReadyPaths` raced `t.TempDir` cleanup:

```text
TempDir RemoveAll cleanup: unlinkat .../001: directory not empty
```

The fake tmux helper always appended every invocation to `tmux.log`, including
for tests that never inspect that log. Package-level background tmux probes can
inherit the temporary fake through `PATH` and recreate the log while the test
framework removes the directory.

This change makes command recording opt-in, disables it for the wait-path test,
and guards both optional fake-output files against empty paths. The wait-path
test now also asserts that it does not create a command log.

## Testing

- `go test -race -short -run '^(TestHermeticTmuxSessionExistsAndCapturePane|TestHermeticWaitForLegacySessionReadyPaths)$' -count=100 ./pkg/agent` — pass
- Full `pkg/agent` coverage run reached 91.4%, but failed on a separate existing
  tmux readiness flake: `TestRestartThenSendKick_CleanSlateThenDelivers`
- `git diff --check` — pass

No changelog entry: this is a test-only reliability fix.

— hive: backend=codex

### DIFF
--- a/src/pkg/agent/tmux_lifecycle_hermetic_test.go
+++ b/src/pkg/agent/tmux_lifecycle_hermetic_test.go
@@ -12,16 +12,23 @@ import (
 	"github.com/kubestellar/hive/pkg/config"
 )
 
-func installHermeticTmux(t *testing.T) string {
+func installHermeticTmux(t *testing.T, logCommands bool) string {
 	t.Helper()
 	dir := t.TempDir()
 	logPath := filepath.Join(dir, "tmux.log")
 	script := `#!/bin/sh
-printf '%s\n' "$*" >> "$HIVE_FAKE_TMUX_LOG"
+if [ -n "${HIVE_FAKE_TMUX_LOG:-}" ]; then
+  printf '%s\n' "$*" >> "$HIVE_FAKE_TMUX_LOG"
+fi
 case "$*" in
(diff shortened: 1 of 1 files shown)


clubanderson:
/lgtm
/approve

Reviewed against the #4728 seam design: the diagnosis is right — the fake tmux fake sits at the head of PATH for the test's lifetime, and manager background probes that exec tmux after cleanup begins inherit it and recreate tmux.log inside the TempDir mid-RemoveAll (the gate log's `directory not empty` signature points at exactly this test). The fix removes the only in-TempDir write for the wait-path test (opt-in recording + empty-path guards) — synchronization by eliminating the shared file, no sleeps — and pins it with the no-log assertion. `go test -race -run TestHermetic -count=20 ./pkg/agent` green on the head.

PR #4785, opened 2026-08-26
🐛 fix(beads): stop concurrent bd processes losing beads

### LINKED ISSUE #4742: [quality] beads store loses beads under concurrent bd processes — persist() tmp-file race, no cross-process test
## Finding

`Store.persist()` (src/pkg/beads/beads.go:737) writes the whole ledger to a **fixed** temp name (`beads.json.tmp`) and renames it over `beads.json`. The store's `sync.RWMutex` only serializes writers **within one process** — but every `bd` CLI invocation is its own process, and agents routinely run several `bd create`/`bd update` calls in parallel.

Reproduced today on a live agent store (two parallel `bd create`):

1. Process A and B both load `beads.json`, each adds its own bead in memory.
2. Both write `beads.json.tmp`; one rename wins, the other fails with `rename beads.json.tmp beads.json: no such file or directory`.
3. Worse: a bead whose `bd create` **reported success** was absent afterwards — the loser's full-ledger snapshot (taken before the winner's bead existed) can also land, silently dropping the other process's bead.

This is last-writer-wins data loss in the findings ledger that feeds the advisory digest. `ReadEach` fixed the in-process race (#3845); the cross-process one is untested and unhandled.

## Recommendation

1. Use a unique temp name (`os.CreateTemp(s.dir, "beads.json.*.tmp")`) so concurrent renames can't ENOENT each other.
2. Serialize load-modify-persist across processes with an advisory `flock` on `beads.json.lock` (or re-load + merge before persist).
3. Add a multi-process regression test: spawn N `bd create` subprocesses against one store dir, assert all N beads survive and no invocation errors.
4. Until fixed, note in agent policy templates that `bd` calls must not run in parallel against the same store.

## Priority
- Impact: high (silent loss of agent findings; reproduced in production-shaped use)
- Effort: medium

---
---

clubanderson:
Fixes #4742 — last-writer-wins data loss in the findings ledger, reproduced on a live agent store.

**Root cause**: `Store.persist()` rewrote the whole ledger through a **fixed** temp name (`beads.json.tmp`) + rename, and the store's `sync.RWMutex` only serializes goroutines within one process — but every `bd` CLI call is its own process. Two parallel `bd create`: one rename ENOENTs, and the loser's stale whole-file snapshot can land last, silently dropping a bead whose create reported success.

**Fix** (per the issue's recommendations):
1. Every mutation (Create/Update/CloseAll/Archive — all other mutators funnel through these four `persist` sites) takes a blocking exclusive `flock` on `beads.lock`, **refreshes the in-memory snapshot from disk under the lock**, applies the mutation, persists, releases. Best-effort: no flock support → old behavior, never refuses to record a finding. Lock released implicitly on process death — a crashed holder can't wedge the store.
2. `refreshFromDisk` merges rather than replaces: adopts beads other processes created; takes the newer copy of beads they updated; re-reads the append-only archive and drops beads it marks retired, so a stale writer can't resurrect an archived bead.
3. `persist` uses a unique `os.CreateTemp` name (chmod 0660 preserved for the node-group model) so even a non-cooperating writer can't ENOENT an in-flight rename.

**Tests** (genuine multi-process, per the issue):
- `TestCrossProcessCreateLosesNoBeads`: 6 real subprocesses (test-binary re-exec) × 15 creates each behind a start barrier. **Pre-fix red on both failure modes** — `rename ... no such file or directory` exits AND "reported created but missing" — post-fix 90/90 survive, stable at -count.
- `TestRefreshMergesConcurrentWriters`: pins the three merge rules with two Store handles on one dir (create survives, newer update survives, archived bead not resurrected + stays retired).

pkg/beads 92.8% coverage (floor 90), `-race` clean; pkg/advisory + pkg/dashboard consumers green; CHANGELOG entry added.


clubanderson:
Self-review record: reproduced #4742's exact failure modes pre-fix (rename ENOENT exits AND a create that reported success missing afterwards — 6 subprocesses × 15 creates behind a start barrier), verified post-fix 90/90 survive stable at -count. flock + refresh-under-lock at the four persist seams all mutators funnel through; merge rules pinned separately (adopt concurrent creates, newer-update wins, archived beads stay dead + retired). Unique CreateTemp keeps non-cooperating writers from ENOENTing an in-flight rename; 0660 node-group model preserved. pkg/beads 92.8% (floor 90), -race clean; advisory/dashboard consumers green.

/lgtm
/approve

PR #4789, opened 2026-08-26
🐛 fix(hub): make the registry save loop joinable so it cannot outlive its server

### LINKED ISSUE #4774: [quality] pkg/hub tests are flaky: registry-save goroutine races test TempDir cleanup
## Finding

A full-suite coverage run (`go test -coverprofile ./...`) failed `pkg/hub` with:

```
WARN hub registry save failed path=/tmp/TestNewHubServerRecoversArmedUpgradesFromDisk.../002/hub-registry.json.tmp error="open ...: no such file or directory"
FAIL github.com/kubestellar/hive/pkg/hub 143.265s
```

The warning comes from the registry persist loop (`src/pkg/hub/server.go:3789`, `os.WriteFile(tmpPath, ...)`). The path is inside a `t.TempDir()` that had already been cleaned up — a background registry-save goroutine outlives its test and races the test framework's TempDir removal. The test passes in isolation and the whole package passes on rerun (verified twice, with and without `-coverprofile`), so this is a genuine intermittent flake, likely order/load-dependent.

Flaky tests erode trust in CI signal — a red `pkg/hub` will start being ignored, masking real regressions.

## Recommendation

1. Ensure the hub server's registry-save goroutine is stopped (and pending writes flushed or abandoned) on server shutdown/close, and that tests close the server before returning — e.g. `t.Cleanup(s.Close)` in tests that construct a HubServer with a TempDir registry path.
2. Audit `TestNewHubServerRecoversArmedUpgradesFromDisk` (and siblings constructing multiple servers — the failing path was subdir `002`) for servers that are never shut down.
3. Optionally have the save loop treat "registry dir vanished" as a terminal signal rather than retrying/warning.

## Priority
- Impact: medium (CI trust, intermittent full-suite failures)
- Effort: low

---
---

clubanderson:
Fixes #4774

The debounced registry-save goroutine `NewHubServer` starts was unjoinable — nothing could stop it or wait for it. In tests, a server whose `registryPath` sits in a `t.TempDir` leaked its saveLoop, which could wake up to `registrySaveDelay` (5s) later and recreate `hub-registry.json.tmp` while the framework removed the directory: the intermittent `TempDir RemoveAll cleanup: directory not empty` / `hub registry save failed … no such file or directory` pkg/hub failure.

## Fix — synchronization, not sleeps

- `saveLoop` selects on a stop channel at **both** wait points (the request wait and the debounce sleep) and closes a done channel on exit.
- A stop during the debounce — or racing an already-queued request (both select cases ready; Go picks randomly, so the stop path drains deterministically) — **flushes the pending save synchronously** instead of abandoning it, keeping the durability the delayed write promised. The write itself now reuses `saveRegistryNow` instead of a duplicated body.
- `StopSaveLoop` joins the goroutine; idempotent; nil-safe for bare `&HubServer{}` test literals that never start a loop. `Shutdown` calls it first, so production shutdown also stops leaking the goroutine and flushes.
- `TestNewHubServerRecoversArmedUpgradesFromDisk` (the reported failing test — recovery arms a `requestSave`) and every other test constructing a full server over a TempDir registry now `t.Cleanup` the join.

## Verification

- New `save_loop_join_test.go` pins: prompt join (2s budget vs the 5s debounce — proves cancelable, not waited out), pending-save flush, idempotent second stop, nil-safety. **Mutation-checked**: gutting the join in `StopSaveLoop` fails the test.
- `-count=20 -race` on the pin, `-count=10 -race` on the recovery/delete/requestSave cluster, and the full `pkg/hub` suite (125s) all green.

### DIFF
--- a/src/pkg/hub/delete_hive_registry_test.go
+++ b/src/pkg/hub/delete_hive_registry_test.go
@@ -50,6 +50,7 @@ func helperDeleteHiveEnv(t *testing.T) (*HubServer, *http.ServeMux) {
 	})
 
 	srv := NewHubServer(0, slog.Default(), "test", "v2")
+	t.Cleanup(srv.StopSaveLoop)
 	mux := http.NewServeMux()
 	mux.HandleFunc("DELETE /api/saas/hives/{id}", srv.handleDeleteHive)
 	return srv, mux
@@ -412,6 +413,7 @@ func TestRemoveRegistryEntryPersistsSynchronously(t *testing.T) {
 	_, _ = helperDeleteHiveEnv(t)
 
 	srv := NewHubServer(0, slog.Default(), "test", "v2")
+	t.Cleanup(srv.StopSaveLoop)
 	const keepID = "keep-me-hive"
 	const dropID = "drop-me-hive"
 	srv.mu.Lock()
@@ -456,6 +458,7 @@ func TestKubectlUnreachableHiveIsNotReaped(t *testing.T) {
 	_, _ = helperDeleteHiveEnv(t)
 
 	srv := NewHubServer(0, slog.Default(), "test", "v2")
+	t.Cleanup(srv.StopSaveLoop)
 	const unreachableID = "hosted-unreachable-spoke-hive"
 	// Beat recently enough to be well inside staleRemoveAge, but old enough to
 	// be marked offline: exactly what an unreachable-but-alive hive looks like.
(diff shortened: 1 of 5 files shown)


clubanderson:
/lgtm
/approve

Joinable/cancelable saveLoop with deterministic pending-save flush on stop; mutation-verified pin test at -count=20 -race; full pkg/hub green. CI green.

PR #4790, opened 2026-08-26
🌱 test(agent): seal the live-host leaks the hermetic seams missed

clubanderson:
Fixes #4737
Fixes #4738

Both issues were filed against v4 @ 3838f09 — the #4728 seams merge itself — so they describe genuinely unsealed paths, not stale pre-#4728 runs. The #4728/#4724 seams isolated the SCANNERS and tmux; this seals the writers and env-resolved defaults the reports caught on a live-hive host:

## What leaked

- **Live enforcement files**: `SyncModeFiles`, `writeAgentCapsFile` and `agentBootstrapEnv` wrote `/tmp/.hive-mode-<agent>` / `/tmp/.hive-caps-<agent>` with hardcoded sprintf paths — the very files a live hive's gh-wrapper and proxy read. The observed `permission denied` warnings were the host *defending itself*; on a permissive host a test would have rewritten a real agent's GitHub enforcement mode.
- **`/var/run/pluk`**: `pluk` is on PATH on any live-hive host, so the launch path tried to mkdir the real run dir and create session logs.
- **`/data/agents` + `/data/logs/kicks`**: the `HIVE_WORK_DIR` and `HIVE_KICK_LOG_DIR` defaults meant any test that forgot to set them launched into — and let the permission fixer walk — live agent workspaces.

## The fix (same seam style, production values unchanged)

- New `agentStateDir` var seam ("/tmp" in production) used by all mode/caps/bootstrap writers; TestMain points it at the same per-run temp root the existing `ModeFileGlob`/`CapsFileGlob` scanners already use, so writer and scanner can never diverge. The goose launch command is built from the same value, so the `--text "$(cat …)"` reader moves with the writer.
- `plukRunDir` const → var seam; TestMain repoints it, and the production-path pins in pluk_publisher_test read a `productionPlukRunDir` snapshot (the `productionWatchedHomeDirs` pattern).
- TestMain guards `HIVE_WORK_DIR`/`HIVE_KICK_LOG_DIR` package-wide (restored after the run). `TestNewManager_DefaultWorkDir` switches from `os.Unsetenv` — which silently stripped the guard for every later test in the run — to `t.Setenv("")`, which pins the same default and restores.

Full `pkg/agent` suite green in 279s (well under the 600s budget).

— reviewed lane fix-forward

PR #4796, opened 2026-08-26
🐛 fix(github): make the request watchers joinable so cancelled loops cannot race test cleanup

clubanderson:
The four `Start*RequestWatcher` functions gave callers no way to know their goroutine had exited (the test file's own comment notes this). The new `watcher_loops_test.go` cancel tests (from 170a8e0a) compensated with fixed sleeps (750ms drain / 80ms cancel), which CI load and `-race` outlast: leaked watcher goroutines keep reading package globals (poll intervals, request-dir overrides) while test cleanup restores them → data races and `test (rest 1/3)` shard failures currently hitting every open PR (e.g. #4789, #4790 runs 32973825186 / 32973826495).

Reproduced locally on clean v4: `go test ./pkg/github -run StopsOnContextCancel -count=40 -race` fails at the fastTick restore in watcher_loops_test.go.

Fix (same joinable-goroutine pattern as #4789):
- each `Start*Watcher` returns a `<-chan struct{}` done channel closed when its goroutine exits, immediately-closed on every early-return path (nil client, dir-setup failure)
- `ctx.Err()` guard after each tick so a cancelled watcher cannot take one more lap (select picks randomly among ready cases)
- tests join on `done` (10s budget) instead of sleeping

Production callers in `cmd/hive` ignore the return value — behavior unchanged. Verified: `-count=30 -race` on the watcher tests, full `./pkg/github` and `./cmd/hive` suites green.

### DIFF
--- a/src/pkg/github/issue_request_watcher.go
+++ b/src/pkg/github/issue_request_watcher.go
@@ -109,9 +109,16 @@ type issueRetryState struct {
 // dropped in IssueRequestDir. Same contract as StartPRRequestWatcher: returns
 // immediately, runs until ctx cancel, nil client is a no-op (requests
 // accumulate rather than silently dropping), nil authz fails closed.
-func (c *Client) StartIssueRequestWatcher(ctx context.Context, authz IssueRequestAuthorizer, nowFn func() time.Time) {
+//
+// The returned channel closes when the watcher goroutine has exited (or
+// immediately when it never starts), so callers — tests above all — can JOIN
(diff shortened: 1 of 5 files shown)


clubanderson:
/lgtm
/approve

Joinable done channels on all four request watchers + join-based test drain; complements the interval-capture fix in #4791. Verified locally with -count=30 -race and full pkg/github + cmd/hive suites. CI green.

PR #4799, opened 2026-08-26
fix(agent): launch configured gateway backends

### LINKED ISSUE #4797: 🐛 bug: backend openrouter is accepted at set time and fails at launch with "unknown backend: openrouter"
## What happened?

Setting an agent's backend to `openrouter` is accepted everywhere (config validation, `SetBackendOverride`, the dashboard) and then fails at launch. The agent is marked failed and the tmux pane shows:

```
HIVE LAUNCH FAILED: backend openrouter did not launch: unknown backend: openrouter. The CLI for this backend is not installed in this hive image — upgrade the hive image or switch this agent to a different backend.
```

Two problems:

1. **The agent never starts.** OpenRouter is a first-class model gateway in hive (`config.GatewayKindOpenRouter`, `src/pkg/openrouter/`, the Model Gateways tab, `openRouterBaseURL = https://openrouter.ai/api/v1`), and a gateway name is explicitly documented as doubling as an agent backend id — see `src/pkg/dashboard/gateways.go:202`: *"A gateway name doubles as an agent backend id and as a secret filename component"*. So `backend: openrouter` is a supported configuration that cannot launch.

2. **The remedy in the message is wrong.** There is no "OpenRouter CLI" and upgrading the hive image will never fix this. Gateway backends are supposed to run the **`claude`** binary pointed at hive's OpenAI-compatible translator via `ANTHROPIC_BASE_URL` — the backend name selects the upstream route, not the binary.

This is the same accept-then-fail class of bug as the `watsonx` one that `config.InferenceBackends` and `backendBinaryName` were written to make "structurally impossible" (see the comments at `src/pkg/config/config.go:4404` and `src/pkg/agent/manager.go:5826`). Configured **gateway names** were left out of that fix, so the class is not actually closed.

## Root cause

The two halves of the validate/launch contract dispatch on different sets:

**Accepts a gateway name** — `GovernorConfig.ValidateBackend` (`src/pkg/config/config.go:4482`):

```go
if IsCLIBackend(backend) || IsInferenceBackend(backend) {
    return nil
}
for _, gw := range g.ResolvedGateways() {
    if gw.Name != "" && strings.EqualFold(gw.Name, backend) {
        return nil     // <- "openrouter" passes here
    }
}
```

`Manager.validateBackendName` (`src/pkg/agent/manager.go:781`) does the same via `routableBackend`, which explicitly covers "a configured gateway name". `SetBackendOverride` (`src/pkg/agent/manager.go:8997`) even carries the comment *"routableBackend covers configured gateway names (resolved live), so a gateway-named backend still passes."*

**Rejects a gateway name** — `backendBinaryName` (`src/pkg/agent/manager.go:5844`) builds its map from `config.CLIBackends` + `config.InferenceBackends` only:

```go
binary, ok := binaries[backend]
if !ok {
    return "", fmt.Errorf("unknown backend: %s", backend)   // <- dies here
}
```

`InferenceBackends` is `{"vllm", "llm-d", "litellm", "watsonx"}` — no gateway names, and `openrouter` is not a kind in that list either.

`launchInTmux` (`src/pkg/agent/manager.go:2215`) calls `backendBinary` **before** it ever reaches `isInference := m.routableBackend(backend)` at `manager.go:2312`, so the gateway-aware code path that would have routed this correctly is never reached.

## Expected behavior

`backend: <configured-gateway-name>` (and `openrouter` specifically) should launch the `claude` CLI routed through hive's inference translator at the gateway's endpoint, exactly like `litellm`/`vllm`/`watsonx` do.

Failing that, the backend should be **rejected at set time** with an accurate message, never accepted and then failed at kick time — and the pane message should never tell the operator to upgrade the hive image for a backend that has no CLI.

## Reproduction steps

1. Add an OpenRouter gateway under Model Gateways (kind `openrouter`, endpoint `https://openrouter.ai/api/v1`, API key set), named `openrouter`.
2. Set an agent's backend to `openrouter` (agent config `backend: openrouter`, or `/switch openrouter`). It is accepted with no error.
3. Kick/restart the agent.
4. Attach to the agent's tmux pane — the agent is `failed` and the pane shows the `HIVE LAUNCH FAILED` banner above. `AgentProcess.LastError` is `unknown backend: openrouter`.

## Environment

- Branch/version: `v4` @ `a8250a01`
- Deployment method: Kubernetes (pod `hive-5f85747786-4p22d`, agent `guide`)
- Relevant configuration: agent `guide` with `backend: openrouter`; OpenRouter gateway configured under `gateways:`

## Logs or screenshots

Pane output:

```
hive-guide@hive-5f85747786-4p22d:/data/agents/guide$
HIVE LAUNCH FAILED: backend openrouter did not launch: unknown backend: openrouter. The CLI for this backend is not installed in this hive image — upgrade the hive image or switch this agent to a different backend.
```

Manager log line (`src/pkg/agent/manager.go:2219`): `backend binary not found  name=guide backend=openrouter error="unknown backend: openrouter"`

## Suggested fix

Close the gap the way the `watsonx` fix intended, by deriving rather than enumerating:

- Make `backendBinaryName` (or a `Manager` method wrapping it) resolve **any routable backend** — `IsInferenceBackend(b) || isGatewayBackend(b)` — to the `claude` binary, instead of consulting only the two static lists. That makes `backendBinary` agree with `validateBackendName`/`ValidateBackend` by construction.
- Add a regression test in the shape of `src/pkg/agent/watsonx_backend_test.go`: for every configured gateway name, `backendBinaryName` must not return `unknown backend`. That file's existing "no supported backend returns unknown backend" assertion iterates only `CLIBackends`/`InferenceBackends`, which is why this slipped through.
- Consider adding `openrouter` to `config.InferenceBackends` so it works as a built-in backend id without requiring a gateway entry named exactly `openrouter` — it is already a `GatewayKind`, is offered as a CLI choice in the contributor picker (`src/pkg/dashboard/api_contribute.go:2004`), and has its own package.
- Separately: make the missing-binary pane message distinguish "no CLI installed" (a real CLI backend) from "this backend has no CLI at all" (a gateway), so the suggested remedy is never "upgrade the hive image" for a gateway backend.

Danathar:
Closes #4797

## What changed

- resolve configured gateway backend names through the manager's live routing predicate
- launch all configured gateway backends with the `claude` CLI, matching built-in inference backends
- keep unconfigured backend names rejected as unknown
- make missing-binary pane guidance identify Claude as the gateway runner instead of implying an OpenRouter-specific CLI exists

## Tests

- `go test ./pkg/agent -run 'TestConfiguredGatewayBackendLaunchesClaude|TestUnconfiguredGatewayBackendStillRejectedAtLaunch|TestGatewayMissingBinaryBannerNamesClaude' -race -count=1`
- `go test ./pkg/agent -short -race -count=1`

— hive: backend=codex

### DIFF
--- a/src/pkg/agent/gateway_backend_test.go
+++ b/src/pkg/agent/gateway_backend_test.go
@@ -1,0 +1,65 @@
+package agent
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestConfiguredGatewayBackendLaunchesClaude(t *testing.T) {
+	binDir := t.TempDir()
+	claudePath := filepath.Join(binDir, "claude")
+	if err := os.WriteFile(claudePath, []byte("#!/bin/sh\nexit 0\n"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	t.Setenv("PATH", binDir)
+
+	m := &Manager{}
+	m.SetGatewayBackendChecker(func(backend string) bool {
+		return backend == "openrouter" || backend == "corp-gateway"
+	})
+
+	for _, backend := range []string{"openrouter", "corp-gateway"} {
+		t.Run(backend, func(t *testing.T) {
+			binary, err := m.backendBinary(backend)
+			if err != nil {
+				t.Fatalf("configured gateway backend %q cannot launch: %v", backend, err)
+			}
+			if binary != claudePath {
+				t.Errorf("backend %q resolved to %q, want claude CLI %q", backend, binary, claudePath)
+			}
+		})
+	}
+}
+
+func TestUnconfiguredGatewayBackendStillRejectedAtLaunch(t *testing.T) {
+	m := &Manager{}
+	m.SetGatewayBackendChecker(func(backend string) bool { return backend == "openrouter" })
+
+	_, err := m.backendBinary("other-gateway")
+	if err == nil || !strings.Contains(err.Error(), "unknown backend") {
+		t.Fatalf("unconfigured gateway backend error = %v, want unknown backend", err)
+	}
+}
+
+func TestGatewayMissingBinaryBannerNamesClaude(t *testing.T) {
+	t.Setenv("PATH", t.TempDir())
+
+	m := &Manager{logger: discardLogger()}
+	m.SetGatewayBackendChecker(func(backend string) bool { return backend == "openrouter" })
+	agent := &AgentProcess{Name: "guide", BackendOverride: "openrouter"}
+
+	if err := m.launchInTmux(t.Context(), agent); err != nil {
+		t.Fatalf("launchInTmux returned error: %v", err)
+	}
+	if agent.State != StateFailed {
+		t.Fatalf("State = %q, want %q", agent.State, StateFailed)
+	}
+	if !strings.Contains(agent.lastLaunchFailureBanner, "Gateway backends run through the claude CLI") {
+		t.Errorf("gateway failure banner does not identify the claude CLI: %q", agent.lastLaunchFailureBanner)
+	}
+	if strings.Contains(agent.lastLaunchFailureBanner, "The CLI for this backend") {
+		t.Errorf("gateway failure banner incorrectly implies a dedicated OpenRouter CLI: %q", agent.lastLaunchFailureBanner)
+	}
+}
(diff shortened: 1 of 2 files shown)


clubanderson:
Thanks Danathar — I compared this against current v4 and the fix is fully superseded by d92d2eb3 (#4800), which closed #4797. The landed code already resolves configured gateway names through the claude translator launch path, covers configured gateway names beyond openrouter, and has regression coverage for launchability and the missing-binary banner. Closing this PR to avoid blind-merging duplicate changes.

Issue #4798, opened 2026-08-26
RFC: operations and telemetry agents — raise a managed project's operability


clubanderson:
## Use case

Hive raises a project's **code quality** (`quality`, from L3) and its **security posture**
(`sec-check`, from L4). It does nothing about the project's **operability** — whether the
software the project ships can actually be run, watched, and debugged by the people who
depend on it.

That is a real gap for adopters. A maintainer who accepts Hive today gets tests, CI hygiene
and CVE triage. They do not get: a `/health` endpoint, OpenTelemetry spans, a Prometheus
scrape target, a Grafana dashboard, an SLO, an alert rule, or a runbook. Those are exactly
the things a project needs before anyone will run it in production, and exactly the things
that get deferred forever because they are nobody's feature.

This RFC proposes two new lane agents — **`telemetry`** and **`operations`** — that work on
the managed project's repositories to close that gap, the same way `quality` and `sec-check`
work on theirs.

**These agents are for the project's users, not for operating Hive.** They are peers of
`quality` and `sec-check`, not fleet infrastructure. (For the fleet-internal problem — agents
that are dead while reporting `running` — see #4665, which is complementary and unaffected by
this RFC.)

### Scope split

| Agent | Owns |
|---|---|
| **`telemetry`** | Instrumentation and backend wiring: OTel spans/metrics, Prometheus endpoints, dashboards-as-code, commercial APM, web analytics, structured logging |
| **`operations`** | Operational practice: health/readiness endpoints, SLOs/SLIs, alert rules, runbooks, incident templates, release/rollback hygiene |

Two agents rather than one because the concerns are separable in practice: a project can want
instrumentation without wanting SLOs, and the two produce different artifacts on different
cadences.

### Backend families

The agents must be **backend-pluggable, not vendor-hardcoded**:

| Family | Examples |
|---|---|
| **Open source** | Prometheus, Grafana, OpenTelemetry, Loki, Jaeger/Tempo, Mimir |
| **Kube-native** | ServiceMonitor/PodMonitor CRDs, OpenTelemetry Operator, Grafana Alloy, kube-state-metrics |
| **Commercial** | Datadog, New Relic, Dynatrace, Honeycomb, Splunk, Grafana Cloud, Google Analytics |

OpenTelemetry is the natural instrumentation spine — vendor-neutral, and already a Hive
dependency (`src/pkg/tracing/`). Per-backend exporters are the swappable part, which is what
[...]

PR #4802, opened 2026-08-26
✨ feat: add opt-in telemetry and operations agents

Danathar:
## Summary

- add `telemetry` and `operations` agents to ACMM L2–L6 with explicit advisory, hold-gated, and full policy templates
- keep both agents paused in every governor mode until an operator deliberately opts in
- register roles, defaults, disjoint routing keywords, Discord aliases, and updated ACMM documentation
- extend `AgentReport` with an optional validated `artifacts` collection and the `instrument` report kind

## Safety

- L2–L4 agents are advisory only
- L5–L6 agents may open issues and PRs but never merge
- policies forbid literal credentials/endpoints and unconfigured off-box exporters
- policies explicitly handle the lack of secondary worktree provisioning and require `--repo` for multi-repo GitHub actions

## Validation

- `go test ./pkg/config -run 'Test(ACMMPack|Operability)' -count=1`
- `go test ./pkg/outputschema ./pkg/classify -count=1`
- `go test ./pkg/dashboard -run '^TestValidateAgentGeneralInput$' -count=1`
- `go test ./pkg/discord -run '^TestResolveAlias_KnownAliases$' -count=1`
- `go test ./pkg/governor -run '^$' -count=1`
- `go test ./pkg/hub -run 'Test(ACMMSuggestionIsLevelSpecific|ACMMLevelLabel|NudgeCopyIsActionable)$' -count=1`
- `go test ./pkg/policies -count=1`

The full package sweep was also attempted; sandbox-only tests requiring `NET_ADMIN` or loopback listeners cannot run in the contributor environment.

Refs #4798. The Project Observability governor settings tab and persisted managed-project backend selection remain follow-up work; the agents are safe to ship before that UI because every cadence is paused and the policies require an explicitly configured backend before export wiring.

— hive: backend=codex

### DIFF
--- a/src/docs/acmm-policy-matrix.md
+++ b/src/docs/acmm-policy-matrix.md
@@ -27,7 +27,7 @@ A single interactive advisor helps with repo setup and architecture decisions. G
 | guide | advisory | `guide-advisory.md` |
 | brainstorm | advisory | `brainstorm-advisory.md` |
 
-### L2 — Advisory (Instructed) (5 agents)
+### L2 — Advisory (Instructed) (7 agents)
 
 Agents observe and report findings as advisory beads on the dashboard and tracking issue. No GitHub issues or PRs created. Humans decide what to act on.
 
@@ -37,9 +37,11 @@ Agents observe and report findings as advisory beads on the dashboard and tracki
 | scanner | advisory | `scanner-advisory.md` |
 | quality | advisory | `quality-advisory.md` |
 | guide | advisory | `guide-advisory.md` |
+| telemetry (paused) | advisory | `telemetry-advisory.md` |
+| operations (paused) | advisory | `operations-advisory.md` |
 | brainstorm | advisory | `brainstorm-advisory.md` |
 
-### L3 — Quality-Gated (Measured) (6 agents)
+### L3 — Quality-Gated (Measured) (8 agents)
 
 Quality agent opens GitHub issues and PRs about testing gaps, coverage, and CI workflows. All other agents remain advisory. CI-maintainer joins to monitor build health. Key artifact: measurement infrastructure.
 
@@ -50,11 +52,13 @@ Quality agent opens GitHub issues and PRs about testing gaps, coverage, and CI w
 | ci-maintainer | advisory | `ci-maintainer-advisory.md` |
 | **quality** | **holdgated** | `quality-holdgated.md` |
 | guide | advisory | `guide-advisory.md` |
+| telemetry (paused) | advisory | `telemetry-advisory.md` |
+| operations (paused) | advisory | `operations-advisory.md` |
 | brainstorm | advisory | `brainstorm-advisory.md` |
 
-### L4 — Security-Aware (Adaptive) (7 agents)
+### L4 — Security-Aware (Adaptive) (9 agents)
 
-All agents open GitHub issues — bugs, docs gaps, CI problems, security vulnerabilities. Only Quality, sec-check, and ci-maintainer may open PRs. Security agent joins. Closed-loop feedback: agents act on their own findings.
+Delivery agents open GitHub issues — bugs, docs gaps, CI problems, security vulnerabilities. Telemetry and operations remain advisory and paused. Only Quality, sec-check, and ci-maintainer may open PRs. Security agent joins.
 
 | Agent | Mode | Template |
 |-------|------|----------|
@@ -64,9 +68,11 @@ All agents open GitHub issues — bugs, docs gaps, CI problems, security vulnera
 | **quality** | **holdgated** | `quality-holdgated.md` |
 | guide | measured | `guide-issues.md` |
 | **sec-check** | **holdgated** | `sec-check-holdgated.md` |
+| telemetry (paused) | advisory | `telemetry-advisory.md` |
+| operations (paused) | advisory | `operations-advisory.md` |
 | brainstorm | advisory | `brainstorm-advisory.md` |
 
-### L5 — Semi-Autonomous (Semi-Automated) (9 agents)
+### L5 — Semi-Autonomous (Semi-Automated) (11 agents)
 
 Agents open issues AND pull requests. All PRs get a hold label — humans batch-review and approve. Architect produces RFCs, strategist coordinates across agents. The system proposes; it does not merge autonomously.
 
@@ -80,11 +86,13 @@ Agents open issues AND pull requests. All PRs get a hold label — humans batch-
 | sec-check | holdgated | `sec-check-holdgated.md` |
 | architect | holdgated | `architect-holdgated.md` |
 | strategist | holdgated | `strategist-holdgated.md` |
+| telemetry (paused) | holdgated | `telemetry-holdgated.md` |
+| operations (paused) | holdgated | `operations-holdgated.md` |
 | brainstorm | advisory | `brainstorm-advisory.md` |
 
-### L6 — Fully Autonomous (10 agents)
+### L6 — Fully Autonomous (12 agents)
 
-Agents open issues, create PRs, and auto-merge on green CI. No hold label. Outreach agent handles community engagement (highest trust — external-facing). Governor at fastest cadence.
+Existing autonomous lanes can open issues, create PRs, and auto-merge on green CI. No hold label. Outreach handles community engagement. Telemetry and operations remain paused and use `ISSUES_AND_PRS`, so they never merge their own PRs.
 
 | Agent | Mode | Template |
 |-------|------|----------|
@@ -97,6 +105,8 @@ Agents open issues, create PRs, and auto-merge on green CI. No hold label. Outre
 | architect | full | `architect-full.md` |
 | strategist | full | `strategist-full.md` |
 | outreach | full | `outreach-full.md` |
+| telemetry (paused) | full | `telemetry-full.md` |
+| operations (paused) | full | `operations-full.md` |
 | brainstorm | advisory | `brainstorm-advisory.md` |
 
 ## Example config alignment
@@ -111,6 +121,7 @@ Agents open issues, create PRs, and auto-merge on green CI. No hold label. Outre
 4. **Mode escalation is per-agent.** At L4, some agents are measured (issues only) while others are holdgated (issues + PRs). The level defines the mix.
 5. **Knowledge priming works at all levels.** The `${KNOWLEDGE}` template variable injects relevant facts from git sources and wiki layers regardless of the agent's mode.
 6. **Brainstorm is always advisory.** It produces KB facts and beads, never GitHub issues or PRs. Its role evolves from inception (L1) to ongoing ideation (L2+), but its mode stays advisory at all levels.
+7. **Telemetry and operations are opt-in.** They are present from L2 onward but use a paused cadence in every governor mode. At L2–L4 they report findings only; at L5–L6 they may open issues and PRs but never merge.
 
 ## Changing a hive's ACMM level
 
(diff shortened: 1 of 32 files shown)


clubanderson (commented):
Thanks for this — it's a careful and faithful implementation of #4798, and it's clear you read the RFC's constraints rather than just its feature list. A few things you did **better** than the RFC specified, which I want to call out before the defects:

- The RFC only asked for the paused-cadence invariant to be pinned by *a* test. `TestOperabilityAgentsArePausedInEveryGovernorMode` (`src/pkg/config/acmm_packs_test.go:41`) loops all five levels × four modes × both agents — and because `Cadence.IsPaused()` returns `false` for the empty string, a *missing* map key fails the test too. That is a genuine fail-closed assertion, not a parse-and-shrug. Exactly right.
- You added the L2–L4 pack entries the RFC's prose implied but its table under-specified, and kept them `ADVISORY` + paused.
- `src/pkg/policies/defaults/` and `src/policies/` are byte-identical per file (same blob hashes in the diff) — #4434's two-tree trap avoided.
- The policies use the `$HIVE_REPOS` + explicit `--repo "<org>/<repo>"` pattern from `sec-check-holdgated.md` rather than the `$HIVE_REPO` pattern, and each one states the no-worktree-provisioning constraint. That was the specific thing most likely to be missed.
- `outputschema.Artifacts` is `omitempty` with a `len() >` bound and no `== nil` required-arm, so existing reports that omit `artifacts` still validate. I checked this against `validateReport` specifically because a regression there breaks every agent's report, not just these two — it's clean.
- `DefaultAgentMode` correctly left untouched; pack-declared `mode:` wins via `agentMode()`, and `kick_template` matches what `SuffixForLevel` would derive anyway (`-holdgated` at L5, `-full` at L6). Belt and braces.

Two defects, one of which I'd like fixed before merge.

---

### 1. `slo` and `sli` as lane keywords will misroute unrelated issues (please fix)

`src/pkg/classify/classifier.go:124`

```go
{Name: "operations", Keywords: []string{"healthz", "readyz", "readiness", "slo", "sli", "runbook", "incident-response", "rollback", "alerting"}},
```

`classifyLane` (`classifier.go:159-185`) matches with `strings.Contains` — **substring, not token**. Three-letter `slo` and `sli` appear inside ordinary English:

| Title | Matches |
|---|---|
| `Fix slow login redirect` | `slo` → operations |
| `Add sliding window rate limiter` | `sli` → operations |
| `Refactor slice handling in beads` | `sli` → operations |
| `Sliced buffer panic in parser` | `sli` → operations |
| `Translate docs to Slovak` | `slo` → operations |

This is worse than noise on a paused agent. `filterByLane` (`scheduler.go:1233`) gives an agent issues where `issue.Lane == lane || issue.Lane == ""`. Once a "slow login" bug is stamped `Lane: "operations"`, it is no longer unclassified — so it disappears from **scanner's and every other lane's queue** and lands only in a lane that is paused in all four governor modes. The issue silently goes nowhere.

It's also nondeterministic in the way the RFC warned about: first-match-wins over unordered map iteration, so `Fix slow test coverage regression` routes to `operations`, `ci-maintainer`, or `quality` depending on iteration order.

Note every other lane keyword is either ≥7 chars or a non-word token (`rfc`, `ga4`). `slo`/`sli` are the only ones that are substrings of common words.

Suggested fix — use the hyphenated/expanded forms, which still match the labels and titles you actually care about:

```go
{Name: "operations", Keywords: []string{"healthz", "readyz", "readiness", "slo-", "sli-", "service-level-objective", "service-level-indicator", "error-budget", "runbook", "incident-response", "rollback", "alerting"}},
```

or keep `slo`/`sli` only as *label* matches. The same change is needed in `applyKnownAgentDefaults` (`src/pkg/config/config.go:4360`) and in the `lane_keywords:` line of all five pack YAMLs, since those override `defaultLanes` once a roster is applied.

A regression test with a title like `"Fix slow login redirect"` asserting it does **not** route to `operations` would pin this.

### 2. Stale validation message omits `instrument` (one-line fix)

`src/pkg/outputschema/outputschema.go:140`

```go
violations = append(violations, Violation{Field: "kind", Message: "must be one of findings, fix, review, advisory, summary"})
```

You correctly added `KindInstrument` to `validKind()` and updated `CorrectivePrompt`, but this message still lists the old five. An agent that fumbles its kind gets told `instrument` is invalid when it isn't — and this string is fed back to the model as the correction. Worth adding `, instrument`.

The existing assertion at `outputschema_test.go:105` only checks the prefix `"kind: must be one of findings, fix"`, so it passes either way — that's why it slipped through.

### Minor, non-blocking

- `TestGetIdentity_KnownAgents` (`src/pkg/discord/bot_coverage_test.go:87`): `getIdentity` falls back to `agentIdentities["pipeline"]` for any unknown name, so `id.Emoji != ""` holds for literally any string. Adding `telemetry`/`operations` there doesn't actually assert anything new. Pre-existing weakness, not introduced by you — just flagging that it isn't the coverage it looks like.
- `src/pkg/classify/classifier_test.go:306` and `src/pkg/dashboard/validation_test.go:319` contain gofmt-only realignment of unrelated structs. Harmless, but it makes the diff look larger than the change.

### CI

Heads up that this is a fork PR and the workflows are awaiting maintainer approval — `v2-tests` and `v2-ci` have **not run at all** on `82348e2`, and several other runs are stuck `queued`. Since the safety argument for this PR rests on `TestOperabilityAgentsArePausedInEveryGovernorMode`, we need those green before merging. That's on us, not you.

The `changelog reminder` cancellation is unrelated infrastructure flake — the job recorded zero steps, and the workflow is advisory-by-design and always-succeeds. Two adjacent branches hit the same thing in the same window. No action needed.

Nothing else from the RFC looks silently dropped — the Project Observability tab and persisted backend selection being follow-ups is stated in your description and is the right call given everything ships paused.

clubanderson:
Thanks for this — it's a faithful implementation of the RFC, and several things are done more carefully than the RFC asked for. Both policy trees are byte-identical (the thing #4434 had to fix retroactively), the policies follow the `sec-check-holdgated.md` explicit-`--repo` pattern rather than the `$HIVE_REPO` trap, and the paused-cadence test is genuinely fail-closed: because `IsPaused()` returns false on `""`, a *missing* cadence key fails too, not just a wrong one.

Verified independently: 20/20 paused cadences per agent (5 levels × 4 modes, zero non-paused), ADVISORY at L2–L4 and `ISSUES_AND_PRS` at L5–L6 with neither agent ever reaching merge, `-holdgated` at L5 and `-full` at L6 matching what `SuffixForLevel` derives, `DefaultAgentMode` untouched as designed, and the `outputschema` change backward-compatible (`Artifacts` is `omitempty` with only an upper bound, so existing reports validate unchanged).

Two things before merge — one of them matters a lot.

## 1. `slo` and `sli` as lane keywords will silently swallow issues

`classifyLane` matches with `strings.Contains`, not tokens (`src/pkg/classify/classifier.go:182`). So these three-character keywords match inside ordinary words:

| Issue title | Matches |
|---|---|
| `Fix the slow login redirect` | `slo` |
| `sliding window bug` | `sli` |
| `array slice panic` | `sli` |
| `Slovak translation` | `slo` |

The consequence is worse than misrouting. `filterByLane` (`src/pkg/scheduler/scheduler.go:1233`) passes issues where `Lane == lane || Lane == ""` — so an unclassified issue is visible to *every* lane, but a **mis**classified one is visible only to the lane it was assigned. A bug titled "slow login redirect" would leave scanner's queue and land in `operations`, which is paused in all four governor modes. It would simply stop being worked, with nothing reporting it.

Every other lane keyword in the tree is either ≥7 characters or a non-word token (`rfc`, `ga4`). These two are uniquely unsafe at three characters.

Suggested fix — expand to unambiguous forms:

```
slo  → service-level-objective, slo-definition
sli  → service-level-indicator, sli-definition
```

(or hyphenated `slo-`/`sli-` if you prefer brevity — the point is that the match can't land inside a common word). Worth a small regression test asserting `Fix the slow login redirect` classifies as `scanner`, not `operations`.

This needs changing in three places: `src/pkg/classify/classifier.go:124`, `src/pkg/config/config.go:4360`, and the `lane_keywords` in all five `src/pkg/config/packs/level-*.yaml`.

## 2. Stale validation message

`src/pkg/outputschema/outputschema.go:140` still reads `must be one of findings, fix, review, advisory, summary` — `instrument` is missing. `validKind()` and `CorrectivePrompt` were both updated correctly; just this message was missed.

It matters because that string is fed back to the model as the correction text, so an agent emitting a legitimate `instrument` report gets told its kind is invalid without being told what the valid one is. The existing test only asserts the prefix `"kind: must be one of findings, fix"`, which is why it slipped through — worth extending the assertion to cover the full list.

## CI

The cancelled `changelog reminder` is infrastructure flake, not something you need to act on — it's advisory-by-design and two adjacent branches hit it in the same window.

The real gap is that this is a fork PR and the Go test workflows haven't run at all on `82348e2` — `PR Verifier` and the Podman lanes are stuck `queued`, and `mergeStateStatus` is BLOCKED because the required `gate` context hasn't reported. A maintainer needs to click "Approve and run workflows". Since the entire safety case for shipping these agents rests on that paused-cadence test, it has to be green before this merges — I'll get it running.

Danathar:
Addressed both review findings in a175e28f.

- Replaced the ambiguous slo/sli lane substrings in the built-in classifier defaults, known-agent defaults, and all five ACMM packs with slo-, sli-, service-level-objective, service-level-indicator, and error-budget. Added regression cases confirming slow, sliding, and slice titles remain in the scanner lane, while an explicit service-level-objective title routes to operations.
- Updated the invalid-kind validation message to include instrument and strengthened the validation test to assert the complete kind list.

Verified with:
- go test ./pkg/classify ./pkg/outputschema
- go test ./pkg/config -run Test(ACMMPacks|Operability)

The full config package was also attempted; its unrelated TestGateFailsClosedWithoutIP6Tables test cannot exercise the missing-binary path in this sandbox because the installed ip6tables fails earlier with Operation not permitted. The targeted ACMM tests pass.

clubanderson:
/lgtm
/approve

Re-reviewed the full diff on 2bbd0c0c against the feature gates:

**Opt-in gate (the #1 gate) — PASS.** Both agents ship `paused` in all four governor modes across L2–L6 (20/20 cadences per agent, verified from the packs), and `TestOperabilityAgentsArePausedInEveryGovernorMode` is fail-closed: a *missing* cadence key fails too, since `IsPaused()` is false for `""`. No default-on path, no env var that silently enables them, `DefaultAgentMode` untouched so an unlisted entry falls back to ADVISORY.

**Privacy — PASS.** No collection/export code exists in this PR at all: 'telemetry' is an agent that audits *managed projects'* observability, not phone-home telemetry from the hive. Nothing leaves the spoke. The policies additionally forbid committing credentials/endpoints and forbid exporters that send data off-box without an explicitly configured backend.

**ACMM gating — PASS.** ADVISORY at L2–L4, ISSUES_AND_PRS only at L5 (`-holdgated` template) and L6 (`-full`, still never merges). Lower levels gain no write capability; the paused agents can't escalate anything.

**RFC #4798 alignment — faithful.** Deltas are minor and all safety-positive: L2–L4 descriptions say 'findings only' (RFC allowed advisory), the `instrument` report kind + bounded `artifacts` field is additive/omitempty (existing reports validate unchanged), and both prior review findings are fixed — `slo-`/`sli-` + long forms replace the substring-swallowing 3-char keywords (with the 'slow login redirect'→scanner regression tests) and the validation message now names `instrument`.

Fix-forwards I pushed (with Danathar's sign-off preserved): rebased onto current v4, restored the missing DCO sign-off on the review-feedback commit, and adapted the roster-mismatch hub pins (`TestComputeAgentRosterMismatch_*` landed after this branch forked and hardcoded the 5-agent L2 roster). Policy trees byte-identical, forbidden-literal grep clean, affected suites green locally, CI 22/22, trial-merge vs fresh v4 clean.

PR #4806, opened 2026-08-26
fix: prevent release channel publication starvation

### LINKED ISSUE #4804: Version badge shows '✓' and '? behind' at the same time when the branch tip has no image
## Summary

When the release branch tip has no GHCR image yet, a spoke's version badge renders
**`✓` and `? behind` at the same time** — simultaneously claiming "up to date with remote"
and an unknown-but-nonzero distance from the tip. Observed on a self-hosted spoke:

```
Danathar/atomic-image-builder   9f8aea6  [v4]  ✓ ? behind   running
```

Both halves are behaving as written. They disagree because they are gated on different
things, and `? behind` overloads two situations that deserve different words.

## Why both render

`api.go:650` gates the boolean on the image actually existing:

```go
containerReady := ghcrTagExistsCached(latestShort)
resp["behind"] = containerReady && cached != versionHash
// "Only report 'behind' if the container image exists on GHCR"
```

That is right — with no image there is nothing to upgrade to, so `behind` is `false`.

`api.go:655-666` computes the *counter* down a separate path, gated on the same tag
existing but with no `else`:

```go
if sameCommitDashboard(versionHash, stableV4) {
    resp["commitsBehind"] = 0
} else if ghcrTagExistsCached(stableShort) {
    if count, ok := s.commitsBehindStableTip(versionHash, stableV4); ok {
        resp["commitsBehind"] = count
    }
}
```

No image ⇒ the compare is never attempted ⇒ `commitsBehind` stays `null`.

The frontend then takes both branches. `index.html:15098`, reached because `v.behind` is
false and `latestHash` is set:

```js
} else if (v.latestHash) {
  html += ' <span style="color:var(--green)" title="Up to date with remote">✓</span>';
}
```

and immediately after, `index.html:15115`, appended *because* `!v.behind`:

```js
if (!v.behind) html += commitsBehindText;
```

where `commitsBehindText` (`:15091`) is the `? behind` arm, taken whenever `commitsBehind`
is null and the short SHAs differ.

## The real problem: `? behind` means two different things

`commitsBehind` is `null` for two unrelated reasons, and the UI collapses them:

| Cause | Meaning | Reasonable wording |
|---|---|---|
| `ghcrTagExistsCached(stableShort)` is false | The tip **has no image**. There is nothing to upgrade to; the compare was never attempted. | not "behind" at all — the tip simply is not built yet |
| `CompareAheadBy` returned an error | The compare genuinely failed (logged as a WARN at `api.go:697`). | `?` is fair here |

Only the second is an unknown. The first is a *known* state being reported as a question
mark, in `--yellow`, next to a green check.

## Suggested fix

Have the server distinguish the two, e.g. add alongside `stableV4Short`:

```go
resp["stableV4ImageReady"] = ghcrTagExistsCached(stableShort)
```

then in `commitsBehindText`:

- `stableV4ImageReady === false` → do not emit `? behind`. The tip is not built; the `✓`
  already says "nothing to do". A muted "tip not built yet" would be informative, or
  nothing at all.
- `stableV4ImageReady === true` but `commitsBehind == null` → keep `? behind`, which is
  then accurate: the compare really did fail.

That also removes the contradiction, because the only string appended next to `✓` becomes
one that agrees with it.

## How the state is reached

Any time the branch tip has no image. It is not rare: `docker.yml:325` gates the
`imagetools create` — including the **immutable `-t <git_short>` tag** — on
`steps.head-check.outputs.stale != 'true'`. During a burst of merges each build finds
itself no longer HEAD and skips tagging while exiting green, so those commits get no image
at all. Concretely, `9da0813`, `9047b17`, `2db0c6a` and `28ffb5f` each have a fully green
`docker.yml` run and no published tag.

A per-commit SHA tag cannot race with anything — nothing else ever writes it — so gating it
alongside the moving tags may be worth separating. Happy to file that as its own issue if
it is wanted; this one is only about the badge.

## Reproduction

Point a spoke at a commit, then ensure the branch tip has no GHCR tag (merge several PRs in
quick succession, or check during any period where the tip's image has not published yet).
The badge shows `✓ ? behind`. Hovering gives "Up to date with remote" and "Could not compare
this commit with stable v4 tip …" respectively.

---

*First observed on a spoke during the GitHub Actions incident of 2026-08-26 (Actions
`major_outage`, opened 15:09Z). The defect is independent of it: the four stale-skip
examples above ran at 14:24Z, 14:34Z, 14:40Z and 14:50Z, all before the incident opened.*

Danathar:
## Summary

- replace the HEAD-only release guard that silently skipped every completed build in a merge burst
- publish an immutable short-SHA tag for every successful multi-arch build
- advance each moving tag independently using a monotonic GitHub workflow run-number label
- fail red on registry inspection errors instead of reporting a green no-publish run
- apply the policy consistently to hive, contributor, and hub images

## Why this cannot regress tags

Each platform image carries `io.kubestellar.hive.github-actions-run-number`. Before moving a mutable tag, the publisher reads the generation already attached to that tag. A queued run may advance tags whose published generation is older, but it leaves independently newer tags untouched. Re-runs of the same workflow generation remain idempotent.

Existing unlabeled tags are generation 0, so the first post-merge run migrates them without manual registry work. The exact Buildx query was verified against the current public `stable` image, which correctly returned an empty legacy label set.

## Validation

- `shellcheck src/scripts/publish-image-tags.sh src/scripts/test-publish-image-tags.sh`
- `bash src/scripts/test-publish-image-tags.sh`
- `bash src/scripts/check-release-lines.sh`
- `bash src/scripts/test-release-lines-guard.sh`
- parsed both changed workflow files with `yq`
- `git diff --check`

The publisher tests cover missing and legacy tags, forward progress, stale out-of-order runs, per-tag independence, feature-branch channel isolation, and fail-closed registry errors. The secondary dashboard `? behind` presentation noted in the issue is intentionally left for a separate UI change.

Fixes #4804

— hive: backend=codex

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -16,8 +16,10 @@ on:
 # cancel in-progress builds: on a fast-merging release branch (v2 saw ~34 merges
 # in 2h), cancel-in-progress meant every build was killed by the next merge and
 # NONE ever completed, freezing <branch>-latest and blocking all deploys. With
-# cancel-in-progress:false the builds queue and each runs to completion, so
-# <branch>-latest always advances to the newest merged commit.
+# cancel-in-progress:false the builds queue and each runs to completion. Each
+# completed build publishes an immutable SHA tag; publish-image-tags.sh uses a
+# per-tag workflow generation to advance moving tags without either starving
+# during a merge burst or regressing when queued runs execute out of order.
 concurrency:
   group: docker-build-${{ github.ref_name }}
   cancel-in-progress: false
@@ -143,6 +145,11 @@ jobs:
           build-args: |
             GIT_HASH=${{ github.sha }}
             GIT_BRANCH=${{ github.ref_name }}
+          # Monotonic moving-tag guard metadata (#4804). This config label is
+          # inherited by each platform manifest without changing its media type.
+          labels: |
+            io.kubestellar.hive.github-actions-run-number=${{ github.run_number }}
+            org.opencontainers.image.revision=${{ github.sha }}
           # STALE-BINARY FIX (unblocks the #3760 fix in the SHIPPED image).
           # `no-cache-filters: builder` alone rebuilt
           # a fresh /hive in the `builder` stage but let BuildKit serve the FINAL
@@ -269,6 +276,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: [gate, build]
     steps:
+      - name: Checkout tag publisher
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
       - name: Download digests
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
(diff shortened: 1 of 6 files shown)


clubanderson:
Reopening to refresh CI after a reviewer fix-forward commit on the PR branch.

clubanderson:
Thanks, Danathar, for identifying and fixing the release-channel publication starvation path. I carried the core fix forward through #4811, including idempotent immutable short-SHA publishing for successful builds and a 90-day cleanup workflow for old pure short-SHA image tags to keep growth bounded.

#4811 has now merged as f992bfa6623ce5037e844c2300f7726ed96676c5, so I am closing this PR as superseded.

PR #4810, opened 2026-08-26
🐛 fix(dashboard): stop overlay scroll bleeding to the page — audit sweep beyond Prior Prompts

clubanderson:
Fixes #4805

Companion sweep to #4803 (Prior Prompts): the full audit in #4805 found every other overlay-like surface carrying the same latent scroll-bleed / scroll-lock bug class. This PR fixes them all with the **existing shared mechanism** — the `body.modal-open` observer lock plus the #2766/#4803 `overscroll-behavior:contain` containment pattern — no second mechanism introduced (the ACMM dialog's competing bespoke lock is actually *removed*).

## Fixes

| Surface | Bug | Fix |
|---|---|---|
| Hive Chat panel (floating, non-modal) | Body lock never applies; wheel past the end of the message history / raw-response pre / input textarea scrolled the dashboard behind the panel | containment on `.hive-chat-panel`, `.hive-chat-messages`, `.chat-raw-details pre`, `.hive-chat-input` |
| Shared scroll-lock vs late-created overlays (`#kb-modal-overlay`, `#acmm-dialog-overlay`) | `installModalScrollLock` only attribute-observed overlays existing at startup; these are appended once and then toggled via `style.display` — closing one left `body.modal-open` **stuck** (page frozen), reopening never re-locked (bleed behind KB modals from the 2nd open on) | observer re-attaches to all `OVERLAY_SELECTOR` matches on body childList mutations (re-observing is idempotent) |
| ACMM dialog bespoke lock | `acmmShowDialog/acmmCloseDialog` wrote `document.body.style.overflow` directly — a second mechanism fighting the shared class-based lock | removed; shared lock is now the single mechanism (it works once the observer fix lands) |
| GH auth/setup, cred, hive-dialog, Nous/KB, ACMM backdrops + scrollers | Only `.config-overlay` (backdrop) and the config-modal family had containment | same rule extended to all backdrops; containment on `.gh-setup-modal`, `.acmm-dialog`, `.nous-config-body`, nous `textarea`/`pre`, `.welcome-body`, `.config-stats-list` |
| Fixed sidebar (`.oc-sidebar`) | End of a long agent nav handed the wheel to the page | containment |
| **Hub landing** Contribute modal (`src/pkg/hub/static/index.html`) | No body lock at all and no containment — the hub page scrolled behind the open modal | lock on open; restore on Close button, backdrop click, and Escape (new `closeContribModal`); containment inline on backdrop + panel |

## Not affected (audited, see #4805)

[...]

PR #4809, opened 2026-08-26
fix(ui): contain scrolling across overlay surfaces

Danathar:
Fixes #4805.

## Summary

- contain wheel/touch scroll chaining across chat, sidebar, GitHub, ACMM, Nous/KB, welcome, stats, and dialog overlay surfaces
- teach the shared modal scroll-lock observer to subscribe to dynamically appended KB/ACMM overlays, preventing both stuck locks and unlocked reopens
- remove ACMM's competing inline body overflow management and add a shared lock to the hub contribute modal
- extend dashboard and hub static-HTML regression coverage for each affected surface

## Testing

- `go test ./pkg/dashboard ./pkg/hub -run 'TestModalScrollContainment|TestModalBodyScrollLockReused|TestContributeModalScrollContainment' -count=1`
- `go test ./pkg/dashboard ./pkg/hub -count=1`

— hive: backend=codex

Issue #4805, opened 2026-08-26
🐛 Scroll bleed / stuck scroll-lock on overlay surfaces beyond Prior Prompts (chat panel, KB/ACMM modals, hub contribute modal, sidebar)


clubanderson:
Follow-up to #4801 / PR #4803 (Prior Prompts scroll bleed). A full audit of every overlay-like surface in the dashboard SPA (`src/pkg/dashboard/static/index.html`) and the other served static UIs (`src/pkg/hub/static/*`) found the same latent bug class — scroll chaining to the page behind an overlay, missing body scroll-lock, or a stuck lock — on the surfaces below.

## Shared mechanism today

- `body.modal-open { overflow: hidden }` toggled by a `MutationObserver` over `OVERLAY_SELECTOR` (`installModalScrollLock`).
- `overscroll-behavior: contain` on `.config-body`, `.config-overlay textarea/pre`, `.config-modal .config-pre`, the prompt-template editor/preview, and the `.config-overlay` backdrop (#2766), extended per-surface by #4803.

## Affected surfaces

| Surface | Symptom | Missing pieces |
|---|---|---|
| **Hive Chat panel** (`.hive-chat-panel`, floating, z-9001) | Wheel past end of `.hive-chat-messages`, the raw-response `<details>` pre, or the input textarea scrolls the dashboard behind the panel. No body lock applies (correct — it is non-modal), so containment is the only defense and it is absent. | `overscroll-behavior:contain` on `.hive-chat-messages`, `.chat-raw-details pre`, `.hive-chat-input`, `.hive-chat-panel` |
| **Shared scroll-lock: overlays created after startup** (`#kb-modal-overlay` from `kbShowModal`, `#acmm-dialog-overlay` from `acmmShowDialog`) | `installModalScrollLock` attaches attribute observers only to overlays that exist at startup. Later-created overlays are seen once (body childList) but their `style.display` toggles are never observed: closing via `display:none` leaves `body.modal-open` **stuck** (page frozen until an unrelated observed overlay mutates); re-opening never re-locks (scroll bleeds behind KB modals from the 2nd open on). acmm partially masks this with its own inline `body.style.overflow` management — a second, divergent mechanism. | (re)observe overlays on childList mutations; drop acmm's bespoke inline `body.style.overflow` in favor of the shared lock |
[...]

PR #4812, opened 2026-08-26
🌱 fix(dashboard): salvage KB inline-scroller containment and the whole-file lock pin from #4809

clubanderson:
Follow-up to #4810 / #4805. Dedupe disposition for the #4809 collision: #4810 merged first; this salvages the two pieces where #4809 was stronger — first-hop `overscroll-behavior` containment on the KB modal's inline scrollers (`#kb-subs-content`, `#kb-ctx7-results`, matched via `.nous-config-body [style*="overflow-y:auto"]`), and widening the bespoke-lock absence pin from the acmm region to the whole SPA (no `body.style.overflow` writes anywhere). Credit: @Danathar (#4809).

Tests: full pin-test set + #4803's green; check-inline-js passes.

### DIFF
--- a/src/pkg/dashboard/overlay_scroll_audit_test.go
+++ b/src/pkg/dashboard/overlay_scroll_audit_test.go
@@ -35,6 +35,8 @@ func TestOverlayScrollContainmentAudit(t *testing.T) {
 		".nous-config-dialog > .nous-config-body,",
 		".nous-config-overlay textarea,",
 		".nous-config-overlay pre,",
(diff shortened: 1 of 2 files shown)


clubanderson:
/lgtm
/approve

Salvage verified against the merged #4810: KB inline scrollers now get first-hop containment (both spacing variants of the inline style matched), whole-file bespoke-lock pin passes on current head, full pin-test set + check-inline-js green, CI 32/32, trial-merge vs fresh v4 clean.

PR #4822, opened 2026-08-26
fix: skip unchanged advisory digest edits

### LINKED ISSUE #4818: [quality] Advisory digest comment is rewritten every ~60s even when byte-identical — ~1,440 unnecessary forge writes/day
## Problem

`PostAdvisoryDigest` (src/pkg/github/advisory.go) unconditionally calls `Issues.EditComment` on the pinned digest comment in the forge advisory issue (#2364) every refresh cycle (~once a minute), even when the rendered digest is byte-identical to what was last posted. At steady state that's ~1,440 GitHub writes/day carrying zero new information — needless API quota burn and comment-edit churn on the advisory issue.

## Proposed fix

Skip the `EditComment` call when the scrubbed/truncated digest body is unchanged from the last successful post (in-process content hash per "owner/repo#issue" is sufficient; first post after process start always writes).

**Critical invariant — do not break the staleness gate**: a skipped-because-unchanged cycle is a HEALTHY cycle. It must still call `dashSrv.RecordAdvisoryPost(...)` / `RecordAdvisoryOverflow(...)` (or an equivalent freshness record) so `AdvisoryLastPostedAt` keeps advancing in the heartbeat and the hub's advisory-staleness alert doesn't false-fire on a quiet hive. Same for clearing the App-required banner logic that currently keys off a successful write — a verified-unchanged skip is equal proof of a prior successful write, but note it does NOT re-prove current write permission; keep an occasional real write-through (e.g. every Nth cycle or hourly) so permission regressions still surface.

## Acceptance
- Unchanged digest ⇒ no EditComment call (test with counting fake, ≥2 mutation-checked assertions)
- Changed digest ⇒ exactly one edit
- Staleness gate: AdvisoryLastPostedAt advances on skipped cycles
- Periodic write-through preserved so 403/permission regressions still detected

Danathar:
## Summary

- cache the scrubbed and truncated advisory digest hash after successful writes
- skip byte-identical comment edits while treating the cycle as healthy for staleness and App-banner state
- force every 50th cycle through GitHub so permission regressions remain detectable
- cover repeated unchanged cycles, changed bodies, and a periodic 403 write-through

Fixes #4818

## Testing

- `go test ./pkg/github`
- `go test ./cmd/hive -run 'Test(EmptyAdvisory|MissingAdvisory)'`

— hive: backend=codex

### DIFF
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -6177,13 +6177,16 @@ func runEvalCycle(
 							}
 						}
 					} else {
-						logger.Info("posted advisory digest", "repo", primaryRepo, "issue", issueNum, "findings", digest.TotalCount, "via", "app")
-						// Record the fresh, successful digest post so the hub's
-						// advisory-staleness gate stays satisfied for this hive.
+						logger.Info("advisory digest cycle succeeded", "repo", primaryRepo, "issue", issueNum, "findings", digest.TotalCount, "via", "app")
+						// Record the fresh, successful digest cycle so the hub's
+						// advisory-staleness gate stays satisfied for this hive. A
+						// byte-identical body may have skipped the forge mutation;
+						// PostAdvisoryDigest returns nil for that healthy case and
+						// periodically writes through to re-check permission.
 						dashSrv.RecordAdvisoryPost(digest.TotalCount)
 						dashSrv.RecordAdvisoryOverflow(digest.OverflowCount)
-						// A successful write proves the app is installed AND has
-						// write access — clear BOTH the perm issue and the
+						// A successful cycle is backed by a prior or current App
+						// write, so clear BOTH the perm issue and the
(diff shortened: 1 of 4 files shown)


clubanderson:
Thanks @Danathar — you independently identified the exact same fix within minutes of #4821, and your invariant list (skip-while-healthy staleness treatment, forced periodic write-through for permission-regression detection, 403 retry coverage) matched the issue spec precisely. #4821 merged first at `1c366e8d` and covers the same ground: post-scrub wire-byte hashing keyed per `owner/repo#issue`, skipped cycles still advance `AdvisoryLastPostedAt`, failed writes never record the hash (so 403/500 retries every cycle), and an hourly write-through (every 60th unchanged cycle) keeps App-banner clearing and permission regressions detectable. Closing as superseded — genuinely good convergent analysis.

PR #4823, opened 2026-08-26
🐛 Name the real cause on /fleet: all-paused, login-blocked, and partially-dead hives

clubanderson:
## Summary

Fixes two hive-level verdict gaps on /fleet so the hive reason chip names the operator-actionable cause instead of falling through to misleading generic text.

## Gap 1: all agents paused

Live examples observed today include aslom/hive-agent, castrojo/endusers, hashicorp/dev-portal, inference-sim/sim2real, singhar/go-ci, torch-spyre/spyre-inference, TradingAsBuddies/falcon-core, zacburns/mlz-manager, and llm-d/llm-d-workload-variant-autoscaler.

These hives have ExpectedActive agents and queued work, but every expected agent is deliberately paused. The old rollup produced Expected>0, Running=0, Problems=0, causing hiveHealthFor to report red "no agents running". This PR adds an expected-active Paused count and returns an amber verdict: "all agents paused — resume to produce output".

### Login prompt inside grace

Live shapes include placeholder/, available-akswec2-260810-* pool hives, and jeejz/incubator-kie-drools. deriveAgentVerdict already knows the agent is blocked by "sitting at login prompt", but RunState stays working during the 20-minute login grace. The rollup now buckets those problem agents into LoginStuck via BlockedReason fallback, so the hive reads: "N agent(s) stuck at login — re-login needed".

### Dead agents while others still run

Live shape: katamari/ibm-aiops-orchestrator had guide+supervisor failed/down while quality+scanner still ran. When every problem is dead/gone but Running>0, the verdict now says "N agent(s) down — restart needed". Fully down hives still keep the existing "no agents running" wording.

## Tests

- gofmt
- cd src && go test -count=1 ./pkg/hub/

### DIFF
--- a/src/pkg/hub/agent_capability.go
+++ b/src/pkg/hub/agent_capability.go
@@ -393,6 +393,14 @@ type agentFleetRollup struct {
 	Expected int `json:"expected"`
 	Running  int `json:"running"`
 	Able     int `json:"able"`
+	// Paused is how many expected-active agents are deliberately quiet because
+	// the operator paused them. Live all-paused hives (aslom/hive-agent,
+	// castrojo/endusers, hashicorp/dev-portal, inference-sim/sim2real, and
+	// peers) reported Expected>0, Running=0, Problems=0 and therefore fell
+	// into the false "no agents running" outage leg. Pause is an operator
+	// choice, not a dead session; count it separately so hive health can say
+	// "all agents paused" instead of inventing a crash.
+	Paused   int `json:"paused,omitempty"`
 	Stuck    int `json:"stuck"`
 	Impotent int `json:"impotent"`
 	// Problems is how many agents the governor expects on that cannot deliver
@@ -595,6 +603,9 @@ func rollupAgents(agents []AgentSummary, blockers hiveBlockers, queuedWork int, 
 		if a.ExpectedActive {
 			r.Expected++
 		}
+		if a.ExpectedActive && (a.Paused || strings.EqualFold(a.State, agentStatePaused)) {
+			r.Paused++
+		}
 		if v.RunState == runWorking {
 			r.Running++
 		}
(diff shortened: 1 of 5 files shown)


clubanderson:
/lgtm
/approve

Reviewed as third-party intake with full scrutiny on 630dec07:

**Gaps verified real at base** — reverting each new branch reproduces the misleading verdicts the body describes: without the `DeadOrGone == Problems` (Running>0) case the katamari shape falls through to generic "2 agent(s) blocked"; without the `Paused == Expected` branch the all-paused hives read false-red "no agents running"; without the runWorking→login-reason bucket the in-grace login shape hides re-login behind a blocked count.

**Precedence checked** — the new partially-dead case sits after quota, login-stuck, and the fully-down (`Running == 0`) guard, so the familiar "no agents running" wording for fully-down hives is preserved and nothing shadows the more specific causes; idle-with-work and the generic fallback remain reachable. The paused-amber branch only runs after the `Problems > 0` block, so a paused-but-problematic fleet still reds correctly.

**Mutation-checked (3 flips)** — killed each of the three new branches in turn; `TestHiveHealthForReasons`, `TestHiveHealth_LiveFleetCauseShapes`, and `TestRollup_LoginPromptWithinGraceStillBucketsHiveCause` each went red on exactly the corresponding mutation.

**Quality gates** — full `./pkg/hub` green locally on a rebase against current v4 (clean, 0 conflicts incl. post-58b0cace traffic); new amber state's only consumers (health_verdict.go + my-hives.html) both updated, `.why-chip.warn` and `--amber` CSS pre-exist; the sort-rank `hasOwnProperty` fix correctly repairs the falsy-rank-0 bug for "problem" rows; no forbidden literals; dashboard index.html untouched so no CSP guard exposure; DCO + Copilot trailer present; emoji title present.

PR #4824, opened 2026-08-26
✨ feat: add managed-project observability settings

### LINKED ISSUE #4798: RFC: operations and telemetry agents — raise a managed project's operability

Danathar:
Closes #4798

## Summary

- complete the RFC's remaining Project Observability governor tab, separate from Hive's own telemetry settings
- let operators declare open-source, Kubernetes-native, and commercial targets, with safe environment-variable and secret references only
- turn the telemetry/operations opt-in into an explicit action that unpauses the selected agents at a conservative daily cadence
- derive first-run suggestions from telemetry findings while requiring an operator Save before persistence
- inject the confirmed target declaration into every telemetry and operations policy variant, failing closed when no backend is selected

## Testing

- `go test ./pkg/dashboard ./pkg/scheduler`
- `go test ./pkg/config -run 'TestProjectObservability'`
- `node .github/scripts/check-inline-js.js src/pkg/dashboard/static/index.html`
- verified source and embedded policy copies are byte-identical

— hive: backend=codex


clubanderson:
Reviewed at feature depth against the RFC #4798 Part 4 spec — approving.

**Opt-in verified**: nothing unpauses without the explicit owner-only PUT — enabling with zero selected platforms is a 400 (mutation-checked: gutted the gate, test went red), default configs keep every mode×level paused, and detect-then-persist suggestions only mark the tab dirty; Save is the persistence step. **Secret hygiene**: references are allowlisted-platform + name-only (env-var name regex, secret-name/key regex; literal URL/token rejection mutation-checked), and neither the GET response nor ${PROJECT_OBSERVABILITY} prompt injection carries anything but names. **Fail-closed**: empty config renders the detect-and-report-only clause (mutation-checked); source and embedded policy copies verified byte-identical, all six variants. **RFC closure**: Parts 1/2/3/5 landed in #4802 — this delivers Part 4 including the Features-tab cross-link the RFC warned about, so Closes #4798 is right.

Fixed forward on your branch (with v4 merges): the new tab's `gh75`/`gh76` action ids collided with the explain-mode saver (#4714) and the advisory update-interval field (#4827) that landed after you branched — duplicate object keys would have silently clobbered the explain-mode handler. Renumbered to `gh77`/`gh78` (11bfb634). Also fast-tracked your #4831 watchdog-test sync fix, which was blocking this PR's test lane. check-inline-js clean, coverage floors hold, suites green. Thanks Danathar — nice completion of the observability RFC.

PR #4825, opened 2026-08-26
feat: configure advisory digest update interval

### LINKED ISSUE #4820: [quality] Advisory tab: configurable advisory-issue update interval (default unchanged)
## What

Add a setting to the **Advisory tab** in main settings (Settings → Advisory, `renderGovAdvisory` in `src/pkg/dashboard/static/index.html`) to change how often the pinned advisory-issue digest comment is updated. Some operators want a longer delay between updates than the current ~60s cycle.

## Why

The digest currently refreshes on a fixed ~1-minute cadence hardcoded in the posting loop (`src/cmd/hive/main.go`, ~lines 5820–5960). Combined with the skip-if-unchanged guard from #4818 this eliminates redundant writes, but operators should also be able to slow the *checking/posting cadence itself* — e.g. to reduce GitHub API traffic, notification churn on watched repos, or audit-log noise.

## Spec

**Config**: new field under the existing `advisory` governor config section (same section as `max_findings`, `staleness_days`, `show_all`, `pr_autoclose`):
- `update_interval_s` (int, seconds). `0` or unset = current default (~60s) — **default behavior must not change**.
- Clamp to a sane minimum (e.g. ≥30s) server-side; log once if clamped.

**UI** (Advisory tab, under a "When it updates" group):
- Number input "Update interval (s)" with tooltip: how often the digest comment is checked and refreshed; 0 = default (60). Longer values reduce API writes and notification churn.
- Follows the existing dirty-section pattern → `PUT /api/config/governor/advisory`; prefilled from `GET /api/config/governor`.

**Backend**:
- Posting loop reads the configured interval (live re-read each cycle if that's the existing pattern for other governor knobs, otherwise applied on restart — document which).
- Interacts with #4818: skip-if-unchanged still applies within each cycle; `RecordAdvisoryPost` staleness-gate semantics must remain correct at longer intervals (the wedged-digest health check thresholds must account for a user-lengthened interval so a healthy slow cadence isn't flagged as wedged).

**Invariants**:
1. Unset/0 → exactly today's cadence. No behavior change for existing installs.
2. The advisory health/staleness alarm (`no advisory issue resolved…` wedge detection) must use `max(configured_interval, default)` as its baseline so long intervals don't false-alarm.
3. hive.yaml round-trips the field; config GET/PUT tests updated.
4. check-inline-js CSP guard passes on index.html touch.

## Test plan
- Unit: interval parsing, clamping, 0→default.
- Wiring test in the governor settings pattern (`governor_settings_wiring_test.go` style).
- Health-gate test: long interval + healthy posting ≠ wedged.

Danathar:
Closes #4820

## Summary

- add `governor.advisory.update_interval_s`, preserving the existing 60-second behavior when unset or zero and clamping positive values below 30 seconds
- expose the setting in Governor Config → Advisory and apply live changes on the next eval cycle without restarting
- gate only forge-facing digest refresh attempts, leaving finding ingestion and dashboard state updates on their existing cadence
- propagate the effective interval to the hub and expand wedge-detection thresholds for intentionally slow healthy cadences
- document the interaction with the byte-identical digest guard from #4818

## Testing

- `go test ./cmd/hive ./pkg/dashboard ./pkg/hub`
- `go test ./pkg/config -run 'TestAdvisory'`
- `node .github/scripts/check-inline-js.js src/pkg/dashboard/static/index.html`

— hive: backend=codex

### DIFF
--- a/src/cmd/hive/advisory_wedge_test.go
+++ b/src/cmd/hive/advisory_wedge_test.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/kubestellar/hive/pkg/advisory"
 	"github.com/kubestellar/hive/pkg/beads"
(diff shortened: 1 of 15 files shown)


clubanderson:
Closing as superseded by #4827 (merged as 7867f994), which landed a parallel implementation of #4820 — three of us raced this one. Thank you for the PR: two of its coverage ideas were adopted into the review (the heartbeat wire-format round-trip/omit-when-zero pin and the Advisory-tab UI wiring test), and your advisory-staleness.md signals-table update matched what got documented.

Two semantic differences worth noting for future reference, since they're why #4827's gate shape won: (1) unset/0 there preserves *exactly* today's cadence by leaving the gate fully open every eval cycle rather than materializing a 60s attempt-window — with ticker jitter, a 60s window measured from the last attempt can skip alternate cycles, and 'every eval cycle' must keep tracking governor.eval_interval_s when that's not 60; (2) the window advances on *success* only, not on attempt — your schedule.due() records the attempt time before the post, so a failed post (403, rate limit) waits out the full interval before retrying, and no-op branches (missing pinned issue) consume the window too, slowing both recovery and the hub's staleness signal. #4827 also caps the interval at 1h so every allowed cadence stays under the hub's 90-minute wedge threshold by construction.

PR #4826, opened 2026-08-26
✨ feat(advisory): configurable advisory-issue update interval (#4820)

### LINKED ISSUE #4820: [quality] Advisory tab: configurable advisory-issue update interval (default unchanged)

clubanderson:
Fixes #4820

Companion to #4821 (the #4818 skip-if-unchanged guard, merged as 1c366e8d): that eliminated redundant *writes*; this makes the *checking/posting cadence itself* configurable.

## What

`governor.advisory.update_interval_s` (int seconds) in the existing `advisory` governor section, exposed in **Settings → Advisory** under a new **When it updates** group.

## Invariants (from the issue)

1. **Unset/0 → exactly today's cadence.** The field is not materialized on load ("every eval cycle" must keep tracking `governor.eval_interval_s`), and `advisoryPostDue(0, …)` is unconditionally open. Pinned by `TestAdvisoryUpdateInterval_UnsetMeansDefaultCadence`, `TestAdvisoryPostDue_ZeroIntervalAlwaysOpen`, `TestGovAdvisory_UpdateInterval`.
2. **Wedge alarm baselines on max(configured, default).** The spoke reports its *effective* interval over the heartbeat; the hub widens the staleness baseline to `max(90m, 2× interval)` in both `advisoryStale` and the diagnostics report (`advisoryStaleThresholdFor`). 2× = one fully missed window + jitter headroom, so an on-schedule slow cadence never brushes the threshold. Ingest clamps the reported value to a week so a rogue spoke can't disable the alarm; 0/absent (old spokes) keeps the default threshold — nothing changes for existing fleets. Pinned by `TestAdvisoryStale_LongIntervalHealthyNotStale` / `…GenuinelyWedgedStillFlags` / `…NoIntervalKeepsDefaultThreshold` / `…ShortIntervalNeverShrinks`.
3. **#4818 interaction.** Skip-if-unchanged still applies within each admitted cycle; the interval gate advances **only on a verified successful post** (mirroring the hash-recording rule), so failed posts retry every cycle, the first post after start is never delayed, and `RecordAdvisoryPost` semantics are unchanged. The digest is still built and handed to the dashboard every cycle — only the GitHub round-trip is paced. Note: the #4818 write-through counts post *attempts*, so at a longer interval the forced full rewrite stretches to 60 attempts × interval (documented; it only heals out-of-band comment edits).
4. **Server-side clamp ≥30s, logged once** — at use time via `EffectiveUpdateInterval` (covers hand-edited hive.yaml) *and* normalized at PUT so GET reports the cadence actually run; negative values are 400-rejected. Config is re-read live each cycle, so a dashboard save applies without restart (documented in advisory.md).
5. **hive.yaml round-trip** (`omitempty` pinned so untouched configs don't grow the key) + GET/PUT wiring tests; UI uses the existing `markDirty` dirty-section → `PUT /api/config/governor/advisory` pattern (`data-arg-types="s,s,N"`, no inline handlers).

## Verification

- Targeted + full suites green: config 93.6%, hub 88.7% (floor 87), dashboard 85.3% (floor 82); full `cmd/hive` green.
- Mutation checks: disabling the hub threshold widening → hub tests FAIL; forcing the gate always-open → `TestAdvisoryPostDue_WithinWindowDefers` FAILs.
- `check-inline-js` CSP guard passes on the index.html change.
- Docs: `src/docs/advisory.md` gains a "How often it updates" section.

### DIFF
--- a/src/cmd/hive/advisory_update_interval_test.go
+++ b/src/cmd/hive/advisory_update_interval_test.go
@@ -1,0 +1,45 @@
+package main
+
+import (
+	"testing"
+	"time"
+)
+
+// #4820: advisoryPostDue is the update-interval gate on the digest's GitHub
+// round-trip. These tests pin its three contracts: a zero interval means the
+// gate is always open (the pre-#4820 every-cycle cadence — invariant 1 of the
+// issue), the first post after process start is never delayed, and the window
+// is measured from the last SUCCESS so failures keep retrying every cycle.
+
+func TestAdvisoryPostDue_ZeroIntervalAlwaysOpen(t *testing.T) {
+	now := time.Now()
+	if !advisoryPostDue(0, now.Add(-time.Second), now) {
+		t.Error("interval 0 must open the gate every cycle — the default cadence")
+	}
+	if !advisoryPostDue(0, time.Time{}, now) {
+		t.Error("interval 0 with no prior post must open the gate")
+	}
+}
+
+func TestAdvisoryPostDue_FirstPostNeverDelayed(t *testing.T) {
+	if !advisoryPostDue(4*time.Hour, time.Time{}, time.Now()) {
+		t.Error("a zero lastSuccess (no post since process start) must open the gate regardless of interval")
+	}
+}
+
+func TestAdvisoryPostDue_WithinWindowDefers(t *testing.T) {
+	now := time.Now()
+	if advisoryPostDue(10*time.Minute, now.Add(-3*time.Minute), now) {
+		t.Error("3 minutes into a 10-minute window the gate must be closed")
+	}
+}
+
+func TestAdvisoryPostDue_PastWindowOpens(t *testing.T) {
+	now := time.Now()
+	if !advisoryPostDue(10*time.Minute, now.Add(-10*time.Minute), now) {
+		t.Error("exactly at the window boundary the gate must open")
+	}
+	if !advisoryPostDue(10*time.Minute, now.Add(-11*time.Minute), now) {
+		t.Error("past the window the gate must open")
+	}
+}
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -3632,6 +3632,17 @@ func main() {
 					_, overflow := dashSrv.AdvisoryCounts()
 					return overflow
 				}(),
+				// The spoke's EFFECTIVE advisory update interval (#4820), in
+				// seconds, so the hub can widen its staleness baseline for a
+				// hive whose operator deliberately slowed the digest cadence —
+				// a healthy 4-hour cadence must never trip the 90-minute
+				// default wedge alarm. 0 = unset (default cadence): the hub
+				// keeps its default threshold, which is also what it does for
+				// old spokes that never report the field.
+				AdvisoryUpdateIntervalS: func() int {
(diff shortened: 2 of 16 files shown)


clubanderson:
Closing as superseded by #4827 (merged as 7867f994), which landed while this PR's CI ran. Both implementations chose the same core semantics — 0/unset = exactly the pre-existing every-eval-cycle cadence (no materialized 60), use-time clamp with a one-shot log, live config re-read, gate only the GitHub round-trip, window advances on success only (failures retry next cycle, first post never delayed), #4818 skip-guard applies within each admitted cycle.

One divergence: this branch satisfied the wedge-alarm invariant by reporting the effective interval over the heartbeat and widening the hub baseline to max(90m, 2× interval); #4827 instead caps the interval at 1h so every allowed cadence stays under the 90-minute threshold by construction — simpler, no protocol change, equally honest. If >1h cadences are ever wanted, the heartbeat propagation + advisoryStaleThresholdFor widening here (incl. ingest clamping and wire-format omit-when-zero pins) is the ready-made follow-up.

PR #4831, opened 2026-08-26
🌱 test(watchdog): synchronize detached restarts

### LINKED ISSUE #4829: [Coverage] packages below 90% threshold
## Coverage gate failed

- **Measured:** `v4` @ `03a2221559d798f434a18dbe5ae19c301204da39`
- **Run:** https://github.com/kubestellar/hive/actions/runs/33011673035
- **Total coverage:** 86.0%

### What the gate reported

```
watchdog: TESTS FAILING
```

Danathar:
Closes #4829

The coverage gate could not score `pkg/watchdog` because two reconciler tests sampled detached restart work before its goroutine had started or settled.

- Wait for the deliberately wedged restart to enter the fake fleet before asserting that later ticks do not stack another restart.
- Let each synchronously recorded backoff attempt settle before advancing the fake clock in the ladder test.
- Keep production watchdog behavior unchanged; this only makes the asynchronous test harness deterministic.

Verification:
- `go test ./pkg/watchdog -short -count=20 -timeout 180s`
- `go test -race ./pkg/watchdog -short -count=20 -timeout 300s`
- `go test -race -coverprofile=/tmp/hive-4829-coverage.out ./pkg/watchdog -short -count=1 -timeout 180s` (98.0%)

— hive: backend=codex

### DIFF
--- a/src/pkg/watchdog/reconciler_test.go
+++ b/src/pkg/watchdog/reconciler_test.go
@@ -547,6 +547,8 @@ func TestWedgedRestartNeverBlocksTickAndNeverStacks(t *testing.T) {
 	case <-time.After(2 * time.Second):
 		t.Fatal("Tick blocked on a wedged restart — the control-plane guard failed")
 	}
+	waitFor(t, func() bool { return fleet.restartCount() == 1 },
+		"wedged restart goroutine must enter Fleet.Restart")
 
 	// Past the hard timeout the wedge is alerted; no restart stacks on it.
 	clock.Advance(s.RestartTimeout + time.Second)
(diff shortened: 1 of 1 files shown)


clubanderson:
Reviewed: exactly right — the wedged-restart test sampled restartCount before the detached goroutine entered Fleet.Restart (reproduced 16/20 red locally on clean v4), and the ladder test could collapse backoff steps the same way. Both fixes are synchronization via the existing waitFor/waitRestart helpers, no sleeps, production code untouched. Verified 20× and under -race on the head. This flake is currently blocking every open PR's test lane — merging now. Thanks Danathar!

Issue #4836, opened 2026-08-26
[v5] Epic: per-repo cost attribution — join audited repo events to timestamped token usage


clubanderson:
## Goal

Attribute agent token cost to the **repository the work was actually for**, so an operator can answer "what does maintaining `org/repo` cost us?" — today they can only answer it per agent or per hive.

This epic is deliberately scoped by what is *honestly derivable*. Per-repo cost is achievable, but not uniformly, and not without stating loudly what cannot be attributed. A plausible-looking wrong number is worse than no number here, because these figures get used for budget decisions.

## Why this is not a small change

Cost is attributed by parsing agent session files, which know the **agent, model and session** — never the repo. Four routes were investigated and rejected:

| Route | Why it fails |
|---|---|
| `HIVE_REPO` / `HIVE_REPOS` | `HIVE_REPO` is always `Repos[0]`, identical for every agent (`agent/manager.go:8290`). `scheduler.go:569` explicitly warns agents off relying on it. Neither says which repo was *worked*. |
| `issueCosts` | Keyed by issue number (ambiguous across repos) **and has no production writer** — see #4833. Verified empty on a live spoke. |
| Agent cwd → repo | Agent cwd is `/data/agents/<name>`, no repo component. The kick instructs `git clone … /tmp/<repo>`, so the work happens where no session file looks. |
| Session id via `PRRequest` | `InvocationMeta` is built in the hive process from request files whose forge-resistance model trusts **only the file UID**. An agent-supplied session id would be unauthenticated — a silently-wrong join. |

A session-discovery watcher (map each agent to its live session file) was also proposed and rejected: the cwd→slug mapping exists only for Claude (`claude_scanner.go:363-386`); Copilot uses flat id dirs with cwd only *inside* the file (`copilot_scanner.go:93`) and Bob a random UUID dir with no cwd at all (`bob_scanner.go:113`). Newest-mtime selection is unsound because resumed Claude sessions append to pre-existing files, and one agent legitimately has several concurrently-written files (`claude_scanner.go:76-77`).

## The one thing that does work

**The repo side is already solid.** Every mediated output event records the repo as a structured `repo=<org/name>` k=v pair, on an existing timer, with zero GitHub API calls — `pr_request_watcher.go:277`, `issue_request_watcher.go:348`, `review_request_watcher.go:246`, all routed through `recordCreationAudit` which passes the agent (`attribution.go:378`). The audit log timestamps every entry.

[...]

Issue #4835, opened 2026-08-26
parseSessionFile never assigns lastTimestamp, leaving LastActive permanently 0


clubanderson:
## Summary

In `parseSessionFile` (`src/pkg/tokens/collector.go:356`), the local `lastTimestamp` is declared and read but **never assigned**, so `LastActive` is always `0` for flat-format sessions.

## Detail

`src/pkg/tokens/collector.go:375`:

```go
var lastTimestamp int64
```

The scan loop that follows (`:376-402`) accumulates tokens, messages, model, and agent, but never touches `lastTimestamp` — nothing in the loop parses `entry`'s timestamp.

`src/pkg/tokens/collector.go:405`:

```go
summary.LastActive = lastTimestamp
```

reads back the zero value unconditionally.

## Impact

Every `SessionSummary` produced by this parser reports `LastActive: 0` (Unix epoch). Any consumer that sorts by recency, filters on a staleness window, or renders a "last active" column gets epoch for these sessions.

Note the other parsers do this correctly — the Claude parser tracks timestamps (`claude_scanner.go`), and the bob parser derives them from `LastUpdated`/`StartTime` via `bobSessionLastActive` (`bob_scanner.go:246`). So this is an inconsistency specific to the flat-format path.

## Fix

Parse the per-entry timestamp in the loop and keep the maximum, mirroring the Claude parser. Requires `SessionEntry` to carry a timestamp field (verify it does; add if not). If the flat format genuinely has no timestamp, delete the dead variable and the assignment rather than leaving code that implies a value it can never produce.

clubanderson:
Triage note from the #4838 review: this bug is OUTSIDE the removed stub and remains live. `parseSessionFile` (src/pkg/tokens/collector.go) still declares `lastTimestamp` but never assigns it in the scan loop, so `summary.LastActive` is permanently 0 for sessions parsed via that path. `LastActive` has real consumers — dashboard cost.go session listing, status_builder.go model picking, and the copilot scanner's live-capture gate — so this issue stays open independently of the stub removal.

Issue #4834, opened 2026-08-26
Dashboard iterates /api/issue-costs object with for...of, throwing TypeError


clubanderson:
## Summary

`fetchIssueCosts` in `dashboard/index.html` iterates the `/api/issue-costs` response as an **array**, but the API returns an **object**.

## Detail

`GET /api/issue-costs` is served by `handleIssueCosts` (`src/pkg/dashboard/api.go:1925`), which returns `s.deps.Tokens.IssueCosts()` — a `map[string]int64` (`src/pkg/tokens/collector.go:282`). That serializes to a JSON **object** keyed by issue number, with integer values.

The frontend (`dashboard/index.html:2199-2210`) does:

```js
const data = await r.json();
issueCosts = {};
for (const entry of (data || [])) {
  if (entry.issue && entry.issue !== 'unknown') {
    issueCosts[String(entry.issue)] = entry;
  }
}
```

`for...of` over a plain object throws `TypeError: data is not iterable`. It is swallowed by the surrounding `catch (_) { /* non-fatal */ }`, so the failure is completely silent.

The body would be wrong even if it did iterate: values are plain integers, so `entry.issue` is `undefined` — the shape assumed is `[{issue, ...}]`, not `{issue: cost}`.

## Fix

Iterate the object:

```js
for (const [issue, cost] of Object.entries(data || {})) {
  if (issue && issue !== 'unknown') {
    issueCosts[issue] = cost;
  }
}
```

## Note

This is currently masked by a second bug: `issueCosts` has no production writer, so the endpoint always returns `{}` (see the companion issue). Both must be fixed for per-issue cost display to work.

clubanderson:
Superseded by removal: PR #4838 (merged c6569532, credit @Danathar) removed the entire stub per-issue cost feature, including the `for...of` iteration site over the `/api/issue-costs` object response that caused this TypeError — the fetch, the endpoint, and the consuming UI are all gone. A regression test (`TestIssueCostsFeatureIsNotAdvertised`) pins that the dashboard no longer requests the endpoint. Real per-repo cost attribution is tracked in epic #4836.

PR #4838, opened 2026-08-26
🌱 fix: remove stub per-issue cost feature

### LINKED ISSUE #4833: Per-issue cost feature is a stub: issueCosts has no production writer
## Summary

`Collector.issueCosts` is never populated by any production code path. The per-issue cost feature is a closed loop that can only ever return an empty map.

## The closed loop

- `SeedIssueCosts` (`src/pkg/tokens/collector.go:274`) is the only writer into `c.issueCosts`. It is called exactly once, from `src/cmd/hive/main.go:2020`, restoring from `hive-state.json`.
- `hive-state.json`'s `issue_costs` field (`src/pkg/snapshot/state.go:33`) is written from `IssueCosts()` (`src/cmd/hive/main.go:6805`), which reads the very same map.

So the map is only ever seeded from a snapshot of itself. Nothing ever attributes a token session to an issue number, so the map starts empty on a fresh spoke and stays empty forever.

## Impact

- `GET /api/issue-costs` (`src/pkg/dashboard/api.go:1925`) always returns `{}`.
- `hivectl observe costs` (`src/pkg/hivectl/commands/observe.go:27`) always shows nothing.

Confirmed empty (0 entries) on a live multi-repo spoke that has 676M tokens across 455 sessions recorded.

## Root cause

Token attribution runs agent-wise: the scanners bucket sessions `ByAgent` and `ByModel`. There is no join key from a token session to an issue (or to a repo), so no writer could be built today without one.

## Suggested fix

Either implement a real writer once a session-to-work join key exists, or remove the endpoint and snapshot field so the dashboard stops advertising a metric that is structurally always zero.

Danathar:
## Summary

- remove the unwritable per-issue cost map and its circular snapshot restore/persist path
- remove the dead Go API, embedded dashboard UI, and `hivectl observe costs` surface
- add regression coverage proving the endpoint is gone and the dashboard no longer requests it

## Testing

- `go test ./pkg/tokens ./pkg/snapshot ./pkg/dashboard ./pkg/hivectl/commands ./cmd/hive`

Closes #4833

— hive: backend=codex


clubanderson:
Reviewed and approving — thanks @Danathar.

**Stub claim verified at base**: the only writer of `issueCosts` was `SeedIssueCosts`, fed exclusively by the snapshot restore path, which was itself fed by `IssueCosts()` at persist time — a pure circular restore/persist loop with no production producer. Removal is correct; epic #4836 is the path to a real per-repo implementation.

**Removal completeness checked**: zero residual references post-diff (API route, handler, dashboard JS incl. the #4834 `for...of` TypeError site, hivectl `observe costs`, snapshot field, collector state, docs). Old on-disk snapshots with `issue_costs` deserialize fine (unknown JSON key ignored). Regression test `TestIssueCostsFeatureIsNotAdvertised` mutation-checked (re-introducing the endpoint string in index.html flips it red).

**#4835 note**: `parseSessionFile`'s never-assigned `lastTimestamp` is OUTSIDE this stub — `LastActive` is consumed by dashboard cost.go/status_builder.go and the copilot scanner's live-capture gate, so #4835 stays open independently.

Merged fresh v4 into the branch (8ee75d56): resolved `persistState` conflicts with the new watchdog reconciler parameter from the fleet-watchdog merge. All target suites + CSP guard green, no forbidden literals, no workflow changes.

PR #4870, opened 2026-08-27
🐛 fix(dashboard): say a disabled agent is disabled, not down

Danathar:
## The report

An operator clicked **▶ terminal** on the `guide` agent and got:

```
error: no tmux socket found for session 'hive-guide'
```

That is the same message #4593 produced, which points diagnosis at ttyd's
`--url-arg`. It is a different fault. #4593 named the *fallback* session for
every agent, because the plumbing was broken fleet-wide — that repeated
`'supervisor'` is the tell. This one names the agent that was clicked,
because the plumbing is fine and the session genuinely does not exist.

## What is actually wrong

A disabled agent keeps a full runtime card, including `▶ terminal` and
`📄 full log`. Neither can work: nothing ever starts the agent, so it has no
tmux session and no pane to capture.

Observed on a live L3 hive — `guide`, `brainstorm` and `ci-maintainer` all at
`state=stopped`, `restarts=0`, never launched, while their cards advertised
both actions:

```
$ curl -s localhost:3001/api/agents/guide/log
capturing pane for guide: exit status 1
```

Disabled agents reach the runtime roster because `ApplyPack` adds every agent
in the ACMM level's pack and gates only the `Start` on `Enabled`
(`api_packs.go:225-231`). That is deliberate — the card is how an operator
turns the agent on — but nothing carried the flag forward, so no consumer
could tell "switched off on purpose" from "stopped because it broke".
`FrontendAgent` had no `enabled` field at all, and `State` reads `stopped`
for both.

The same blind spot reached health. Both checks scored a deliberately
switched-off agent as broken, so any hive that turns an agent off reports a
failure it can never clear — and raises a standing hub alert for an agent
nobody wants running:

```
"guide": { "state": "stopped", "status": "fail" }
```

## The change

- `FrontendAgent.Enabled`, from the live config (`proc.Config` is the
  manager's copy and only refreshes on `UpdateConfig`/reconcile). Not
  `omitempty` — `false` is the value this exists to report.
- The card shows a `disabled` badge, and the session-backed actions render
  inert with the reason instead of linking into a dead end. `past kicks` is
  left alone: it serves a real page for an agent that never ran.
- Deep health scores `skip` instead of `fail`, and `healthSummaryFor` counts
  disabled separately instead of `down` — the same treatment `paused` already
  gets two branches up.

Every check reads "running" before "disabled": an agent disabled while it is
still up keeps running until the next reconcile and genuinely does have a
session until then.

## Not in scope

`AddAgent` running regardless of `Enabled` is the root of the roster entry,
but leaving it is what gives an operator a card to enable the agent from.
With this change that card is honest rather than misleading, and touching it
would pull in the enable flow (`disabled_agent_enable_test.go`) for no gain.

`↻ restart` on a disabled agent is the remaining odd action on that card —
worth a follow-up, not bundled here.

## Testing

`disabled_agent_no_session_test.go` covers the flag reaching the payload, the
live-config-wins precedence, deep health returning `skip`, and the frontend
wiring. Verified failing without the fix. Full `pkg/dashboard` and `pkg/hub`
suites pass.

`newFullServer`'s fixture had to set `Enabled` explicitly: the literal skips
`ApplyAgentDefaults`, which is what gives a real config `enabled: true` unless
the operator wrote `enabled: false`, so the zero value described a state
production cannot produce — an agent in the runtime roster that is switched
off — and the health checks now correctly read that as deliberate.

### DIFF
--- a/src/pkg/dashboard/dashboard_handlers_extra_test.go
+++ b/src/pkg/dashboard/dashboard_handlers_extra_test.go
@@ -31,7 +31,14 @@ func newFullServer(t *testing.T) *Server {
 		},
 		Data: config.DataConfig{AgentsDir: t.TempDir()},
 		Agents: map[string]config.AgentConfig{
-			"scanner": {ID: "scan-001", Role: "scanner", Backend: "claude", Model: "sonnet", DisplayName: "Scanner"},
+			// Enabled must be set explicitly. This literal skips
+			// ApplyAgentDefaults, which is what gives a real config
+			// `enabled: true` unless the operator wrote `enabled: false`
+			// (config/agent_overlay.go). Leaving it at the zero value
+			// described a state production cannot produce — an agent in the
+			// runtime roster that is switched OFF — and the health checks now
+			// read that as a deliberate off-state rather than a down agent.
+			"scanner": {ID: "scan-001", Role: "scanner", Backend: "claude", Model: "sonnet", DisplayName: "Scanner", Enabled: true},
 		},
 		GitHub:     config.GitHubConfig{Token: "ghp_test123456789"},
 		SourcePath: t.TempDir() + "/hive.yaml",
--- a/src/pkg/dashboard/disabled_agent_no_session_test.go
+++ b/src/pkg/dashboard/disabled_agent_no_session_test.go
@@ -1,0 +1,167 @@
+package dashboard
+
+import (
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/kubestellar/hive/pkg/agent"
+	"github.com/kubestellar/hive/pkg/config"
+	"github.com/kubestellar/hive/pkg/governor"
+)
+
+// A disabled agent still has a RUNTIME entry: ApplyPack adds every agent in the
+// ACMM level's pack to the manager and gates only the Start on enabled
+// (api_packs.go). Everything below covers what that costs when nothing carries
+// the enabled flag forward — a card that offers a terminal and a full log for an
+// agent that has never had a tmux session, and two health checks that score a
+// deliberately switched-off agent as a failure.
+//
+// Observed live on hive-wild-mole (ACMM L3): guide, brainstorm and ci-maintainer
+// sat at state=stopped, restarts=0, and /api/agents/guide/log answered
+// "capturing pane for guide: exit status 1" while the card still linked to both.
+
+func TestBuildAgentsCarriesEnabledFlag(t *testing.T) {
+	statuses := map[string]*agent.AgentProcess{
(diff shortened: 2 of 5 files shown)


clubanderson:
Reviewed. **The change is correct and I'd like to merge it — it's blocked only on DCO sign-off, which has to come from you.**

### The review

This fixes a real bug, and it fixes it the way the codebase already handles the same class of problem. `handleHealthDeep` scored a deliberately switched-off agent as `fail`, so any hive that turns an agent off carried a permanent health failure it could never clear — and via `healthSummaryFor`, a standing hub alert for an agent nobody wants running. Treating it as `skip`, exactly as the `paused` branch above it already does, is the consistent call. It matches the principle the health system is built on: absence of output a configuration doesn't produce is not a fault.

Specific things done right, worth naming because they're easy to get wrong here:

- **`Enabled bool` without `omitempty`**, with the reason stated in the comment. `false` is the meaningful value, and `omitempty` would erase precisely the case the field exists to report.
- **Live config wins over `proc.Config`**, with the staleness reason given — `proc.Config` only refreshes on `UpdateConfig`/reconcile.
- **The ordering constraint is documented**: callers check "running" first, because an agent disabled while still up keeps running until the next reconcile and genuinely does have a session until then. That's the subtle part, and it's called out rather than left implicit.
- **The test-fixture change carries its own justification** — the literal skipped `ApplyAgentDefaults`, so the zero value described a state production cannot produce.

Grounding it in an observed live case (`state=stopped`, `restarts=0`, `/api/agents/<name>/log` returning `exit status 1` while the card still linked to both) is what makes it reviewable rather than speculative.

CI is fully green: 28 checks, none failing.

### The one blocker

The `dco` status reports *"Commits in PR missing Signed-off-by"*. Your commit has a `Co-Authored-By` trailer but no `Signed-off-by`, and DCO is a legal attestation the author has to make — nobody else can add it on your behalf, so I'm not going to work around it.

One commit, so this is a single amend:

```sh
git commit --amend -s --no-edit
git push --force-with-lease
```

`-s` appends `Signed-off-by: Douglas Baggett <doug.baggett@gmail.com>`. The check re-runs on push and should go green.

Two optional notes, neither blocking, both fine to ignore:

1. Your branch is 3 commits behind `v4`. Not required — the merge is clean — but a rebase would run the suite against current `v4`.
2. Per project convention, `Co-Authored-By` lines referencing Claude/Anthropic are normally dropped from hive commit messages. Entirely your call; it does not affect the merge.

Ping me once the sign-off lands and I'll merge it. Thanks for the careful write-up — the "why" comments are more useful than the diff.

PR #4891, opened 2026-08-27
✨ feat(linear): bring Linear work sources to GitHub-Issues parity for agent writes

gregoryhunt:
Follow-up to RFC #4492 (Parts 1–2, #4515/#4522). With those a hive could read Linear and acknowledge sessions, but an agent still had no way to do the *tracker* half of its policy against Linear: no credential for `api.linear.app`, and templates written for GitHub Issues (`gh issue create`, `Fixes #N`). The proxy's mutation allowlist was gated but nothing could reach it.

## What this does — built the way the GitHub path is, from existing config

| GitHub Issues | Linear (this PR) |
|---|---|
| App token, tier-scoped, pushed as `GITHUB_TOKEN`; refreshed hourly | Connected app's OAuth token pushed as `LINEAR_ACCESS_TOKEN` to **ISSUES_ONLY+** agents (fallback: `work_source.linear.api_key` as `LINEAR_API_KEY`); re-pushed on the same hourly tick |
| Advisory sessions have `GH_TOKEN`/`GITHUB_TOKEN` stripped | Both Linear vars stripped below ISSUES_ONLY |
| Writes authored by the App bot | Writes authored by the Hive app user (the identity that acks sessions) |
| `${GH_AUTH}` + template recipes | **Work Tracker: Linear** section rendered from `work_source.linear` (team→repo, states, hold labels, `assigned_only`), injected at the held-PR-coordination seam so customized templates cannot omit it; `${WORK_TRACKER}` places it explicitly |
| `Fixes #N` auto-closes on merge | Linear's own GitHub integration: `[TEAM-123]: title` prefix, identifier in branch, `Fixes TEAM-123` (closing) vs `Part of`/`Refs` (non-closing). No hive-side state machine. |

GitHub-sourced hives see no change.

## Changes
- `pkg/linearagent/client.go` — exported `AccessToken(ctx)`
- `pkg/agent/manager.go` — `LinearCredential`, `SetLinearCredentialResolver`, env injection, strip, hourly re-push
- `pkg/dashboard/api_linear_agent.go` — `LinearAgentAccessToken()`; `cmd/hive/main.go` wiring
- `pkg/scheduler/work_tracker.go` — section + seam + `${WORK_TRACKER}`
- Docs: `linear-agent.md` (parity section, verification items 7–8), `env-vars.md`, `policies/README.md`, `CHANGELOG.md`
- Tests: `linear_credential_test.go`, `work_tracker_test.go`

## Verification
`go test ./pkg/agent ./pkg/scheduler ./pkg/linearagent` and dashboard Linear tests pass. Live-workspace checks are listed in the doc and will be walked during setup.

https://claude.ai/code/session_013SEx7DdWVED4txUw3ykY5D


clubanderson:
Reviewed for merge — one blocker before this can land; everything else checked out well.

**Blocker: a tier downgrade leaves a live Linear credential in a running agent's session.** The strip of `LINEAR_ACCESS_TOKEN`/`LINEAR_API_KEY` runs only inside `ensureTmuxSession` (src/pkg/agent/manager.go:2089), which early-returns when the session already exists — so it only fires at session *creation*. A live agent downgraded below ISSUES_ONLY (dashboard mode edit → `UpdateConfig`, or an ACMM pack downgrade for agents that survive the gate) keeps its session, and the refresh loop (manager.go:1478) only *sets* when `CanCreateIssues()` — when the agent drops below the floor it silently stops pushing and leaves the last value in place. tmux forks a fresh CLI per turn, so every post-downgrade turn inherits the stale credential. Unlike the scoped GitHub token (1h self-expiry), `LINEAR_API_KEY` never expires, so the proxy tier gate becomes the *only* defense — the exact posture the comment at manager.go:2084 argues against.

Suggested fix (small): in the refresh loop (or on mode change), `set-environment -u` both vars when `!CanCreateIssues()`, mirroring the creation-time strip — plus a test asserting the downgrade invariant (vars absent after the drop), per the security-tests-assert-the-invariant house rule.

Minor / non-blocking:
- The refresh-loop docs say "hourly" and "no-ops while m.appAuth is nil" — the actual tick is `agentTokenRefreshInterval` (40m) and the loop now runs without App auth. Worth syncing so an incident debug isn't misled.
- src/pkg/dashboard/api_linear_agent.go:127 says the token fetch runs under the manager's lock; launch Phase 2 actually runs with `m.mu` released. Comment overstates (conservatively) — no bug.
- A GitHub-sourced hive with a connected Linear workspace still injects `LINEAR_ACCESS_TOKEN` (src/cmd/hive/main.go:2493 prefers OAuth unconditionally) while `workTrackerSection` returns nothing for it — defensible since the proxy gates mutations, but worth confirming it's intentional.

What passed under scrutiny, for the record: resolver stored as `atomic.Pointer` with no `m.mu` reentrancy, lock order manager→store with no cycle, refresh-vs-restart race benign, an empty/failed token fetch pushes nothing, the token is never logged and the store file is 0600, the `${WORK_TRACKER}` seam handles default/explicit-placeholder/omitted templates with tests for all three, and the injected tier table matches pkg/proxy/linear_rules.go exactly.

Happy to re-review and get this merged as soon as the downgrade strip lands.

clubanderson:
Pushed 0d910b53d259b2331c749f2c53b2bed8a5dfee8c to address the review blocker: the token-refresh tick now actively unsets both LINEAR_ACCESS_TOKEN and LINEAR_API_KEY in a running agent's tmux session whenever the agent is below the ISSUES_ONLY floor, so a live agent downgraded after launch no longer keeps the last-pushed credential across per-turn CLI forks (the creation-time strip in ensureTmuxSession only fires at session creation). The unsets are idempotent and use the same set-environment plumbing as the existing strip. Also adds an invariant test covering the downgrade path (post-downgrade tick must emit unsets for both vars, never a push) and corrects the two stale comments from the review: the refresh-loop docs (tick is agentTokenRefreshInterval, 40m default, and the loop now runs without App auth) and the api_linear_agent.go claim that the token fetch runs under the manager's lock (launch Phase 2 runs with the lock released).

clubanderson:
Re-reviewed after the downgrade-strip commit: the refresh tick now unsets both Linear vars for below-floor agents (unconditional, idempotent, same tmux plumbing as the creation-time strip), with an invariant test covering the downgrade path. All 25 checks green. The review blocker is resolved.
/lgtm
/approve

Issue #4002, opened 2026-08-17
[v5] RFC: re-entrant conversation-as-state agent turn model (durable, handoff-able agents)


clubanderson:
## Summary

Explore making **the conversation the durable state** and each agent turn a **re-entrant, plain function call** over an ordered list of operations (LLM call, tool-approval, tool-execution, compaction, elicitation, max-turns, retry, subagent-sync, hooks). Because no state is suspended in-process between turns, a session can be **handed off between processes** — enabling horizontal scaling and durable long-running background agents without special coordination machinery.

## Why it matters to hive

This is squarely hive's problem space: hive runs many **long-running, background, self-managing agents**. A re-entrant, conversation-as-state turn model would give:

- **Durable, resumable agents** — an agent's state survives a pod restart / spoke roll because it lives in the conversation, not in a suspended coroutine (directly relevant to the frequent spoke rolls we already tolerate).
- **Cross-process / cross-node handoff** — the basis for horizontally scaling a fleet or migrating an agent between spokes without bespoke coordination.
- **Testability** — each turn becomes a plain function with a structured return value, easy to test and reason about.
- **First-class subagent sync** — background subagent results become explicit state transitions rather than ad-hoc plumbing.

## Proposed scope (RFC / spike first — not a direct rewrite)

1. Document hive's current agent-loop/state model and where in-process suspended state exists.
2. Prototype one agent as a re-entrant turn function whose state is fully externalized (conversation + structured turn return).
3. Evaluate a handoff path — does hive want queue-based handoff, and what is the minimal state envelope?
4. Report feasibility + migration cost before committing.

## Notes

Architectural RFC; no external dependency. This is the umbrella model that #4000 (tool-approval operation) and the state-triggered hooks proposal are individual operations within.

clubanderson:
Constituent operations tracked separately: #4000 (tool-approval as an operation), #4001 (state-triggered hooks).

clubanderson:
@hanthor @bketelsen @joshyorko @castrojo @Danathar what do you think?

joshyorko:
@clubanderson Does Hive want queue-based hand-off, and what is the minimal state envelope?

hanthor:
Seems fine to me. Need to limit the context we throw at weaker models though

clubanderson:
Maintainer-side response. Short version: **yes to the spike, with a specific prototype target and three hard problems named up front.** This week's operational history is the strongest argument the RFC could ask for.

[...]

PR #4913, opened 2026-08-27
feat(turn): prototype durable re-entrant turns

Danathar:
## Summary

Implements **step 2 of RFC #4002 on current `v4`**: an isolated, contribute-shaped re-entrant turn prototype whose conversation, ordered operation plan, effect journal, lifecycle status, and terminal verdict all live in one serializable `SessionEnvelope`.

The transition under test is deliberately plain:

```text
Step(SessionEnvelope, optional initial TurnPlan)
    -> (SessionEnvelope, TurnOutput, error)
```

The first process supplies a plan once. `Step` binds stable operation identities into the envelope before the first durable boundary. After that, a fresh process receives only the serialized envelope and calls `Step` with no plan; it skips effects known to have landed, reconciles effects whose outcome is ambiguous, and completes with the same structured verdict and effect summary as an uninterrupted run.

This carries the operation-journal lesson previously exercised on the earlier `v5` line in #4053 onto the active `v4` line, with two material additions:

1. **Persistence is real, not only an interface backed by a byte slice in tests.** `FileStore` performs a mode-`0600`, synced temp write, atomic rename, and parent-directory sync.
2. **Unreconcilable ambiguity fails closed.** If a process dies after an effect may have landed but before its outcome is durable, and no remote reconciliation query exists, re-entry returns `ErrAmbiguous` instead of replaying and risking a duplicate PR/comment/push.

Nothing in the live tmux agent loop or `bin/contributor-relay.sh` constructs this package. It remains the bounded spike requested by the issue. Step 3 (handoff through the existing claim machinery) and step 4 (feasibility/migration-cost report, including the beads-checkpoint comparison) remain open.

## Why this is the step-2 boundary

The merged step-1 inventory established that Hive does not currently own a structured conversation or a structured turn:

- hosted agents are opaque, long-lived backend CLIs under tmux;
- prompts enter through `tmux send-keys`;
- completion and liveness are inferred from rendered pane substrings;
- backend session formats are private and backend-specific;
- `/data/prompt-history.jsonl` contains prompts only, while kick logs contain rendered terminal text rather than role/message/tool structure.

So this PR does not pretend that the current pane-scraped path can be mechanically refactored into a re-entrant function. It prototypes the first honest boundary *after* a headless/structured backend response: the point where an ordered plan meets durable state and non-idempotent external writes. That is also the highest-risk boundary. Repeating inference wastes tokens; repeating “open PR” or “post comment” creates externally visible duplicate work.

The prototype is contribute-shaped because that plane already has the relevant vocabulary and failure history: task refs, `shipped`/`no_work_needed` completion verdicts, branch pushes, PR creation, comments, labels, and an existing offer→claim path that a future handoff must reuse.

## State envelope

`SessionEnvelope` is versioned and carries:

| Field | Why it must be durable |
| --- | --- |
| `session_id` | Stable scope for operation identity across processes/spokes. |
| `agent`, `task_ref` | The owner and work item a handoff receiver must validate. |
| `messages` | The structured conversation state; roles and tool correlation survive re-entry. |
| `plan` | The ordered operations and terminal verdict are state, not a suspended stack frame. |
| `journal` | Records whether each external effect is not started, possibly landed, failed, or definitely landed. |
| `status`, timestamps | Gives the caller a structured lifecycle/result rather than a pane-text inference. |

The plan is copied into the envelope on the initial call. Caller-provided idempotency keys are ignored and re-derived during binding, so a stale key—or a key emitted by a model—cannot suppress unrelated work. If a later caller supplies a plan that differs from the persisted plan, `Step` rejects it instead of silently changing the work during recovery.

The structured `TurnOutput` reports `status`, `done`, `verdict`, `rationale`, and every landed effect. A zero-operation `no_work_needed` plan is representable by the same return shape; no magic stdout line is needed inside the prototype.

## Operation identity

Each side-effectful operation receives a content-derived key:

```text
key = "v1-" + sha256(
    len(version) | version |
    len(session_id) | session_id |
    len(kind) | kind |
    len(repo) | repo |
    len(target) | target |
    len(body) | body
)[:32]
```

The fields are length-prefixed, so adjacent values cannot produce ambiguous concatenations (`("ab", "c")` cannot hash like `("a", "bc")`). The version prefix is an explicit migration escape hatch if the derivation changes later.

Two constraints pull in opposite directions:

- **Stable across re-entry:** the key must be recomputable in a different process from semantic state alone. That excludes timestamps, PIDs, random nonces, attempt counters, and process-local sequence numbers.
- **Distinct per logical effect:** two different comments on the same issue must not collide, because suppressing a legitimate second effect is a silent under-performance failure.

The model's `ToolCallID` is deliberately recorded but excluded from the key. A repeated inference can mint a fresh call ID for the same semantic effect; including it would make a retry look new and defeat replay protection.

Keys are bound before the first persistence call and stored beside the operation. That matters because the persisted artifact is scrubbed: credential-shaped text may be redacted from an operation body, but the already-bound identity must not change or re-entry would miss the journal entry and replay.

## Durable execution protocol

Every external write crosses two commits:

1. Derive/bind the semantic key.
2. If the journal already says `succeeded`, return the recorded external ref without touching the remote.
3. If the journal says `intended`, reconcile against the remote because the previous process may have died after the effect landed.
4. Otherwise record `intended` and **persist it before execution**.
5. Perform the external effect.
6. Record `succeeded` or `failed`, including the remote ref/error, and persist again.

The ordering is the guarantee. Performing first and recording intent afterward creates an unjournaled effect window where a fresh process has no evidence to query and replays blind.

The three journal states exist because two states are insufficient:

| Durable state | What re-entry knows | Action |
| --- | --- | --- |
| no entry | The effect was never admitted. | Persist intent, then perform. |
| `intended` | The effect may or may not have landed. | Reconcile; never guess. |
| `failed` | The prior attempt returned a known failure. | Persist a new intent attempt, then retry. |
| `succeeded` | The effect definitely landed or was reconciled. | Short-circuit with its external ref. |

### Why reconciliation is separate from the key

GitHub does not expose a general idempotency-key header for the writes involved here. Natural surrogates are remote queries after the fact: find an open PR by head branch, find a matching comment, or inspect a label set. Those queries are useful for resolving `intended`, but they are not stable offline identities; they cost a round trip, may be eventually consistent, and do not exist reliably for every effect.

Accordingly:

- the semantic hash is the local journal key;
- the remote query is the tie-breaker for the ambiguous state;
- absence of a reconciler returns `ErrAmbiguous` and performs **zero** remote writes.

This is intentionally conservative. For a push or a comment on a busy thread, manual recovery is preferable to claiming exactly-once semantics the remote cannot support.

## Persistence and secret boundary

The envelope is a secret-bearing artifact: conversation text, tool arguments, remote errors, authenticated URLs, branch names, and metadata can all contain credentials or sensitive repository content.

`FileStore` therefore treats serialization as the security boundary:

1. clone the in-memory envelope;
2. scrub every content-bearing field through Hive's existing `pkg/logscrub` implementation;
3. create a temp file in the destination directory;
4. force mode `0600`;
5. write and `fsync` the temp file;
6. atomically rename it over the previous envelope;
7. `fsync` the containing directory so the rename itself is durable.

The store does not create the directory. The eventual caller remains responsible for placing it under an existing state directory with the correct ownership and authorization policy.

The in-memory envelope is never scrubbed or mutated. The test asserts both halves: plaintext remains available to the active turn, while the durable artifact contains redaction markers and no plaintext credential. Bound idempotency keys survive unchanged so the scrub boundary cannot accidentally disable replay protection.

## Replay test: what it proves

`TestStepReentersAtEveryPersistenceBoundary` drives this ordered contribute turn:

```text
comment on issue
    → push branch
    → create PR
    → label issue
    → persist terminal shipped verdict
```

Four effects produce eight intent/settle commits; the terminal envelope is a ninth kill site. The test first runs an uninterrupted positive control and asserts that it genuinely performs all four effects without duplicates. It then kills the process **before each of the nine writes**, one boundary at a time.

| Kill boundary | Durable state after death | Recovery path |
| --- | --- | --- |
| intent write (odd effect boundary) | Prior effects only; current effect did not run. | Re-enter and perform normally. |
| settle write (even effect boundary) | Current effect landed remotely; journal still says `intended`. | Reconcile and adopt; do not replay. |
| terminal commit | All effects are settled; status may still be `active`. | Short-circuit every effect and persist the same terminal verdict. |

For every case, the simulated process loses all memory. Only serialized JSON crosses into recovery; when the first write itself is killed, recovery starts from the original task input because no envelope ever became durable.

Each case asserts:

- exactly one actual remote call per logical effect across both processes;
- the same sorted effect summary as the uninterrupted control;
- the same terminal `shipped` verdict;
- `StatusCompleted` and `Done=true`;
- no `intended` entry left ambiguous;
- no plaintext credential in the persisted envelope.

Supporting tests prevent the replay table from passing for the wrong reason:

- `TestAmbiguousOperationFailsClosedWithoutReconciler` proves ambiguity is not replayed when no query exists;
- `TestReconcileNotFoundPerformsEffect` proves the implementation does not “solve” ambiguity by skipping every intended operation;
- `TestFailedEffectCanRetry` proves a known failure is re-intended and increments its attempt count;
- `TestIdempotencyKeySemanticIdentity` proves reminted tool-call IDs remain stable while kind/repo/target/body/session changes produce different keys, and proves a caller cannot choose its own key;
- `TestFileStoreRoundTripIsAtomicPrivateAndScrubbed` checks round-trip parsing, mode `0600`, no temp-file residue, redaction, in-memory non-mutation, and key stability;
- `TestValidationAndConfigurationErrors` covers invalid envelopes and missing runner/sink/persister/path configuration, including the zero-operation path.

The new package is at **91.2% statement coverage** under the race detector, above the repository's 90% new-package floor.

## What this does not claim

This is deliberately explicit because a successful replay test can otherwise make a prototype sound more complete than it is.

1. **The model call is not journaled.** A crash before the plan enters the envelope repeats inference and can produce a different plan. The prototype protects landed effects, not token cost or plan determinism.
2. **There is no live backend adapter.** `pkg/turn` begins after a structured headless response. It does not parse a backend's private session files or convert tmux scrollback into a conversation.
3. **There is no cross-process lease.** The journal makes sequential re-entry safe, not concurrent execution safe. Two processes racing the same envelope can both observe the same state. Step 3 must reuse the contributor relay's existing atomic offer→claim ownership and completion verdicts rather than invent another queue or state store.
4. **Reconciliation quality is effect-specific.** PR-by-head is a strong surrogate. Pushes and comments can be ambiguous. The prototype fails closed where it cannot prove the outcome.
5. **Scrub-on-persist is not an authorization policy.** Redaction reduces credential exposure, but a handoff design still has to say which spoke/process may read an envelope and how that access is authenticated.
6. **The journal is unbounded.** No compaction/retention policy exists yet.
7. **The beads alternative is still unmeasured.** Current beads capture task lifecycle state, not structured model responses or pre-effect journals. Step 4 still needs the proposed realistic kill-at-50% comparison: resume from this envelope versus restart from aggressively checkpointed bead notes, measuring wall time, token cost, and duplicate effects.
8. **No production migration decision is taken.** The step-1 fork remains: use documented backend-native resume envelopes and accept backend-specific state, or move selected agents to an API-shaped backend Hive owns.

## Files and review map

- `src/pkg/turn/envelope.go` — versioned state envelope, structured verdict/output, deep clone, parse/validation, scrub-on-persist boundary.
- `src/pkg/turn/journal.go` — semantic operation identity and the `intended`/`failed`/`succeeded` journal.
- `src/pkg/turn/runner.go` — re-entrant `Step`, plan binding, effect execution protocol, reconciliation, fail-closed ambiguity.
- `src/pkg/turn/store.go` — private atomic file persistence with file and directory sync.
- `src/pkg/turn/runner_test.go` — nine-boundary replay matrix and positive/negative controls.
- `src/docs/design/agent-turn-model.md` — step-2 result, measured guarantee, and residual work.
- `src/docs/design/README.md` — design index status updated from “step 1 only” to “steps 1 and 2 complete.”

The highest-value review questions are:

1. Is fail-closed `ErrAmbiguous` the right default for effects without a reliable remote surrogate?
2. Is the semantic key scoped correctly, especially the deliberate exclusion of `ToolCallID`?
3. Is `TurnPlan` the right persisted boundary after structured inference, or should inference output itself be a separate journaled operation before step 3?
4. Does the envelope contain the minimum state needed for a later claim-owned handoff without becoming durable store number five?

## Verification

Local:

- `go test -race -cover ./pkg/turn` — pass, 91.2% statements
- `go vet ./pkg/turn` — pass
- `go build ./cmd/hive/` — pass
- `go vet ./...` — pass
- `git diff --check` — pass
- `go test ./pkg/...` with local socket/tmux/Podman access — the new `pkg/turn` and all other packages passed except existing privileged `pkg/config` test `TestGateFailsClosedWithoutIP6Tables`; this host has `ip6tables` installed without NET_ADMIN, so that fixture observed “permission denied” instead of its expected “binary missing” message

PR CI observed before handoff:

- DCO — pass
- PR content verifier — pass
- docs index and changelog reminders — pass
- `golangci-lint` — pass
- `gosec` — pass
- `govulncheck` — pass
- all five agent test shards completed at time of the latest check — pass
- general test shards and remaining deployment/build lanes were still completing when this body was updated

Refs #4002

— hive: backend=codex

### DIFF
--- a/src/docs/design/README.md
+++ b/src/docs/design/README.md
@@ -59,17 +59,14 @@ that status is the thing to check before treating a page as current behaviour:
   programmatic path that does work today.
 
 - [The agent turn model and where in-process state lives](agent-turn-model.md) —
-  **spike / investigation, informational.** Step 1 of RFC #4002 and the only
-  step taken: a cited map of how hive drives an agent turn today and which
-  per-agent state does not survive a process restart. No decision, no
-  prototype, no proposal. Its main finding is that there is no "turn" in the
-  code at all — a turn begins when hive types a prompt into a tmux pane and
-  ends when the pane matches an idle-prompt marker, so every progress signal
-  hive has (turn completion, liveness, stalls, auth failure) is screen-scraped
-  rendered terminal text. Read it before steps 2-4, particularly for the fork
-  it names: hive does not own the conversation format, so
-  "conversation-as-state" means either accepting backend-specific resume
-  envelopes or moving off opaque interactive CLIs.
+  **spike / investigation, steps 1 and 2 complete.** A cited map of how hive
+  drives an agent turn today and which state does not survive restart, plus an
+  isolated `pkg/turn` prototype: serialized conversation and operation state,
+  a re-entrant structured `Step`, atomic scrubbed persistence, and a replay
+  test that kills a contribute-shaped turn at every operation boundary. It is
+  not wired to the tmux loop or contributor relay, and takes no production
+  decision. Read it before steps 3-4, particularly for the unresolved fork:
+  backend-specific resume envelopes versus an API-shaped backend hive owns.
 
 - [Copilot per-repo cost capture at the MITM proxy](copilot-cost-capture.md) —
   **investigation, no decision taken.** Phase 4 of epic #4836, which asked
--- a/src/docs/design/agent-turn-model.md
+++ b/src/docs/design/agent-turn-model.md
@@ -1,13 +1,15 @@
 # The agent turn model and where in-process state lives (#4002)
 
-Status: **spike / investigation — informational, no decision taken.**
+Status: **spike / investigation — steps 1 and 2 complete, no production
+decision taken.**
 
-This is step 1 of the four steps proposed in RFC #4002 ("re-entrant
-conversation-as-state agent turn model"). It documents how hive drives an agent
-turn **today** and where per-agent state lives that would not survive a process
-restart. It prototypes nothing, proposes nothing, and takes no decision — steps
-2 (prototype), 3 (handoff path), and 4 (feasibility + migration cost) are
-untouched.
+Step 1 of RFC #4002 documented how hive drives an agent turn **today** and where
+per-agent state lives that would not survive a process restart. Step 2 now adds
+an isolated, contribute-shaped prototype in `pkg/turn`: a re-entrant `Step`
+function over a serialized conversation envelope, with journaled external
+effects and crash/reload tests. Nothing constructs it in the live tmux agent
+loop or contributor relay. Steps 3 (handoff path) and 4 (feasibility + migration
+cost) remain untouched.
 
 Every claim below carries a `file:line` citation against `origin/v4` at the time
 of writing. Line numbers drift; the function names are the durable handle. Where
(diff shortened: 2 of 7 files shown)


Danathar:

Continued verbatim in #4933 — same head commit `6f168b80`, recovered from `refs/pull/4913/head`, same branch, same `v4` base, no content changes. Please review there.

Issue #4903, opened 2026-08-27
Ratchet the Go static-analysis gate: 859 golangci findings and a gosec HIGH backlog


clubanderson:
#4902 adds `govulncheck`, `gosec`, and `golangci-lint` to CI. Their first run against a codebase that had never been linted produced a backlog too large to fix in that PR, so the gate landed at an **enforceable** starting point. This issue tracks raising it.

## Why the gate did not land fully blocking

A gate that fails every PR on a pre-existing backlog is not a gate. It gets disabled, or merged around, and the findings survive either way. So #4902 enforces only what is already clean and leaves the rest visible but non-blocking.

## Current state

| Tool | State |
|---|---|
| `govulncheck` | **Blocking, passing.** No code path calls a known-vulnerable symbol. |
| `golangci-lint` | **Blocking**, limited to `govet` + `misspell` |
| `gosec` | **Runs and reports, `continue-on-error: true`** |

## The golangci ratchet

Findings from the first full run:

| Linter | Findings | Notes |
|---|---:|---|
| `errcheck` | 579 | Highest value here. An ignored error on a credential write or config save is the silent-failure class that keeps surfacing in review. Worth splitting by package. |
| `staticcheck` | 225 | |
| `unused` | 44 | Dead code, including possibly-dead security guards. |
| `ineffassign` | 11 | Cheapest next rung. |

**Procedure — one linter per PR:** fix that linter's findings, then enable it in `src/.golangci.yml` in the same PR. Do not enable several at once; a mixed red gate makes it impossible to tell which class regressed.

Suggested order: `ineffassign` → `unused` → `staticcheck` → `errcheck` (split).

## The gosec backlog

Real HIGH-severity findings exist, including **G115 (CWE-190) integer-overflow conversions** at `src/pkg/knowledge/gitsource.go:451` and `src/pkg/dashboard/status_builder.go:1732-1739`.

G115 has a high false-positive rate where a conversion is provably bounded, so each site needs a judgment: fix the conversion, or annotate with `#nosec G115` **and a comment explaining why the bound holds**. A bare `#nosec` with no reason is not acceptable — it is indistinguishable from hiding a real overflow.

**Last step:** once the HIGH findings are resolved, remove `continue-on-error: true` from the gosec step so it blocks. Do not remove the step itself to make the check summary look clean.

## Why this matters beyond tidiness

[...]

PR #4914, opened 2026-08-27
🧹 lint: fix all 44 unused findings and enable unused in the gate

Danathar:
Second rung of the golangci ratchet in #4903, after `ineffassign` in #4906. One linter, its findings, and its config change — nothing else, per the issue's one-linter-per-PR procedure.

`golangci-lint v2.13.1` with the repo config and `unused` enabled: **44 findings → 0**. `unused` is now enabled in `src/.golangci.yml`.

## The findings needed judgment, not a delete pass

The issue calls this rung "dead code, including possibly-dead security guards", and that turned out to be the whole story. Three findings were guards that had come loose from the code they were written to protect — deleting those would have removed the protection and called it cleanup.

### Fixed by wiring up

**`upgradeKubectlTimeout` (`pkg/hub/saas.go`)** — this one is a live defect, not tidiness. The constant documents the bound on "a single auto-upgrade kubectl call", and `bulkRequestBudget` is *sized off it*:

> Worst case per hive is upgradeKubectlTimeout (15s); with the concurrency above, 100 hives need ~100/4*15s = 375s, so this leaves headroom without being unbounded.

Nothing applied it. All four upgrade-delivery kubectl calls — the hub's own `set image`, the branch switch's `set image` and `rollout restart`, and the initial trigger fired when auto-upgrade is switched on — went through `kubectlForCluster`, which builds an `exec.Command` with no context at all. So the per-hive bound was whatever kubectl decided, and `bulkRequestBudget`'s arithmetic rested on a number no code enforced. They now run through `kubectlForClusterContext` under `upgradeKubectlTimeout`, each call with its own bound rather than sharing one.

**`cadencePause` / `cadenceOff` (`pkg/dashboard/status_builder.go`)** — the constants say why they exist:

> Kept as named constants so the offByCadence detection and computeNextKick agree on the exact spellings the config layer emits.

`computeNextKick` and `parseCadenceDuration` both still compared bare literals — the drift the constants were introduced to prevent, already underway. Wired in, plus `cadenceOnDemand` for the third spelling both were comparing.

### Kept, with the reason stated

**`spokeWrapKeyPath` (`pkg/hub/wrapkey_store.go`)** is the documented PVC location of the spoke's X25519 private wrapping key. Every function in that file takes the path as an argument and the lifecycle has no production entry point yet, so this declaration is the only record of where that key material is supposed to land. It carries a `//nolint:unused` with the reason spelled out — the same discipline the issue asks of `#nosec`: an annotation with no stated reason is indistinguishable from hiding something.

This is the one finding in the 44 where "not wired up yet" is the honest answer, and it seemed worth flagging rather than burying.

### Deleted

The remaining 40 were superseded by a later refactor and had no reader. Two are worth calling out because their comments had gone stale in a way that mattered:

- **`hostedAvailableIDPrefix` (`pkg/hub/saas.go`)** claimed the ID prefix is "the reliable signal that a slot is idle inventory". `isAvailableRegistryEntry` records the opposite, with numbers: consulting it counted every claimed placeholder as still available, **38 reported vs 17 true**. An unused constant advertising a known-wrong signal is an invitation back into that bug.
- **`beadLedgerPartialReason` (`pkg/dashboard`)** named "the DegradedReason reported when the ledger snapshot could not cover every configured store" — but the final #3904 policy deliberately never reports a partial ledger as degraded (`observeCandidateDependencies` documents why at length). It described behaviour that does not exist.

Also removed: `diagnoseGitHubAppWrite` (wrapper whose callers all moved to `diagnoseGitHubApp`); `setGenerations` (every writer of `keyGenerations` already holds `keyGenerationsMu` across a wider critical section, so calling it would self-deadlock); the CPU/memory health thresholds duplicated in Go beside the panel's own `CLUSTER_CPU_*_PCT`; `kubectlTop/GetTimeoutSec`, superseded by the per-cluster `clusterHealthQueryTimeoutFor`; `mbPerGB`; `Proxy.mu` (every field set once in `New`); `ContributorPool` and its RWMutex, which guarded two counters nothing ever wrote; `perPage`, shadowed at its one apparent use by a function-local const of the same name; and nine dead test helpers.

Where a declaration was explicitly documentation rather than code — the GitLab/Gitea API path suffixes ("Kept here as documentation of that contract"), `linearPageSize`, `defaultPriority` — the prose stays and the constant goes. A constant nothing reads is not how to keep a contract. Three comments that named a deleted symbol were updated to name the live one.

## Verification

- `golangci-lint run` with the repo config (`unused` enabled): **0 issues**
- `go build ./...`, `go vet ./...`: clean
- `go test` for every touched package: pass — `pkg/dashboard`, `pkg/hub`, `pkg/proxy`, `cmd/hive`, `pkg/apiproxy`, `pkg/celtrigger`, `pkg/github`, `pkg/tokens`, `pkg/worksource`, `pkg/discord`
- One pre-existing failure: `pkg/agent`'s `TestSendKick_RestartsCrashedCLIThenDelivers` fails identically on the unmodified base (`git stash` + rerun) — it needs a real agent CLI in the environment.

## One note for the errcheck rung

The scanner comment on #4903 is right that the issue's 579 figure counts non-test files only, while the gate runs with `tests: true` and therefore measures **3,139**. I've recorded that in the `.golangci.yml` comment so the next person sizing that rung reads the number the gate will actually enforce. The choice between "fix all 3,139" and "enable errcheck with a test exclusion" is a maintainer decision and is left open here.

Remaining rungs after this: `staticcheck` (212 at `45a13d5`), `errcheck`, then dropping `continue-on-error` from gosec.

Refs: #4903

### DIFF
--- a/src/.golangci.yml
+++ b/src/.golangci.yml
@@ -23,7 +23,7 @@ run:
 linters:
   default: none
   enable:
-    # STARTING SET — deliberately three linters, not six.
+    # THE ENABLED SET grows one linter at a time, never in batches.
     #
     # A first run with errcheck + staticcheck + unused enabled produced 859
     # findings (579 errcheck, 225 staticcheck, 44 unused, 11 ineffassign) on a
@@ -35,12 +35,20 @@ linters:
     # its own PR after its findings are fixed — that ratchet is tracked in the
     # ratchet issue #4903.
     #
-    # errcheck (579) is the most valuable remaining rung and the largest, and
-    # is worth splitting by package.
+    # errcheck is the most valuable remaining rung and the largest, and is worth
+    # splitting by package. Note its true size under this config: the 579 figure
+    # counts non-test files only, and `tests: true` above means the gate measures
+    # 3,139.
     - govet
     - misspell
     # RATCHET: ineffassign enabled after its 11 findings were fixed (#4903).
     - ineffassign
+    # RATCHET: unused enabled after its 44 findings were fixed (#4903). Most were
+    # constants and helpers superseded by a later refactor; a few were guards
+    # that had come loose from the code they were written to protect, and those
+    # were reconnected rather than deleted. One declaration is retained under an
+    # explained //nolint:unused — see pkg/hub/wrapkey_store.go.
+    - unused
 
   exclusions:
     generated: lax
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -2812,7 +2812,7 @@ func main() {
 	// The loop therefore runs for the lifetime of the process (it does NOT
 	// return after the first success) and, whenever the banner is currently
 	// showing, re-runs the SAME read+write verification as the manual "Re-check"
-	// button (githubAppRecheckFn, which calls diagnoseGitHubAppWrite) and clears
+	// button (githubAppRecheckFn, which calls diagnoseGitHubApp) and clears
 	// the flag on success. When the banner is not showing there is nothing to do,
 	// so the tick is a cheap no-op that makes no GitHub API calls.
 	{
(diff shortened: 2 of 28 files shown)


Danathar:

Continued verbatim in #4934 — same head commit `f7fef02c`, recovered from `refs/pull/4914/head`, same branch, same `v4` base, no content changes. Please review there.

Issue #4917, opened 2026-08-27
✨ tui T2: API client core + `/api/health`


Danathar:
Part of #4907.

**Files to create/modify:** `v2/internal/tui/client/client.go`,
`v2/internal/tui/client/client_test.go`.

**Build this:**

1. `type Client struct` holding base URL, token, `*http.Client` with a
   5-second timeout.
2. `func New() *Client`: base URL from `HIVE_DASHBOARD_URL` defaulting to
   `http://localhost:3001`, token from `HIVE_DASHBOARD_TOKEN`. Send the
   token the same way the web dashboard does per `dashboard/openapi.json`
   security scheme (check the spec; do not guess the header).
3. `func (c *Client) Health(ctx context.Context) error` calling
   `/api/health`, nil on 200, typed error otherwise.
4. Shared helper `func (c *Client) getJSON(ctx, path string, v any) error`
   for later tasks.

**Acceptance criteria:**
- Test with `httptest.Server` asserting URL, auth header, 200 and
  non-200 paths
- `go build ./...` and `go test ./v2/internal/tui/...` pass

**Out of scope:** any endpoint other than health, SSE, retries/backoff.

**Verify with:** `go build ./... && go test ./v2/internal/tui/...`

> [!NOTE]
> #4912 reports that `dashboard/openapi.json` omits 38 of 69 dashboard
> routes. `/api/health` and the security scheme are what this task needs;
> if either is missing from the spec, follow the epic's contract rule —
> file a scoped follow-up describing the gap, which completes this task.

---
Wave 1 of the #4907 filing protocol. Body copied from the epic's drafted
sub-issue text so the spec carries the context.

Issue #4915, opened 2026-08-27
✨ tui T0: design doc `v2/docs/tui.md`


Danathar:
Part of #4907.

**Files to create/modify:** `v2/docs/tui.md` only. Do not add it to any
docs index; T26 owns that.

**Build this:** A design doc containing, in order:

1. One-paragraph scope statement (copy the Summary and Non-goals from the
   epic).
2. The fixed architecture decisions from the epic, verbatim.
3. An ASCII sketch of the default layout: header bar (hive name, governor
   mode badge, connection state), a 2x2 grid of panes (Agents, Governor,
   Tokens, Events), footer keybinding strip.
4. The keybinding table: `tab`/`shift+tab` cycle pane focus, `j`/`k` and
   arrows navigate within a pane, `p` pause/resume selected agent, `m`
   model picker, `K` kick now, `A` ACMM overlay, `a` attach (local only),
   `?` help, `q`/`ctrl+c` quit.
5. The testing convention: teatest golden files under
   `v2/internal/tui/panes/testdata/`, regenerated with `go test -update`.

**Acceptance criteria:**
- File exists with all five sections
- No build/test impact (doc only)

**Out of scope:** any Go code, any README changes.

**Verify with:** doc-only change; `go build ./...` must still pass.

---
Wave 1 of the #4907 filing protocol. Body copied from the epic's drafted
sub-issue text so the spec carries the context.

Issue #4907, opened 2026-08-27
[Tracker] ✨ feature: `hivectl tui` — terminal dashboard for Hive (k9s-style)


Danathar:
## Summary

Deliver a full-screen terminal dashboard through `hivectl tui` for Hive's day-to-day operator loop: watch the agent fleet, governor, token/cost usage, and audit activity; then perform the common actions without leaving the terminal (pause/resume, switch model, kick now, apply an ACMM level, and attach to a local tmux session).

Functional model: **k9s** (interactive, action-capable, keyboard-driven). Visual inspiration: **btop** (dense live panes).

This is not a second implementation of Hive logic. It is another client of the dashboard API, sharing its authentication, typed HTTP operations, and SSE status stream with the web dashboard and non-interactive `hivectl` commands.

> [!IMPORTANT]
> **This is a coordination-only umbrella. Do not assign #4907 as one task.**
> Claim a linked, narrowly scoped child issue instead. Partial PRs use
> `Part of #4907`; only the final acceptance task may close this parent.

## Current state

The original scaffold and Waves 1–5 are merged. The repository now contains:

- `hivectl tui` and the Bubble Tea/Lip Gloss frame;
- Agents, Governor, Tokens, and Events pane implementations with golden coverage;
- typed clients for agents, governor, tokens, audit events, model catalogues, pause/resume, ACMM, kick, and SSE;
- Agents polling plus SSE status delivery/fallback;
- pause/resume confirmation, local tmux attach, help, resize handling, and an adaptive theme.

The Wave 5 checklist previously shown here was stale. All four tasks are closed:

- [x] #5215 (T13b) — PR #5300: SSE app wiring with poll fallback
- [x] #5216 (T15) — PR #5231: pause/resume UI and confirmation
- [x] #5217 (T20) — PR #5266: kick client
- [x] #5218 (T22) — PR #5267: local tmux attach

The TUI is **not v1-complete yet**. A closure audit found that several client/pane pairs were never connected to the root app, and the remaining action UI was only drafted. The completion wave below is now fully specified and filed.

## Closure-audit findings

These are observable product gaps, not checklist bookkeeping:

1. **Model apply has no TUI client operation.** T16 lists backend models, but the original T17 draft had no typed method for `POST /api/model/{agent}/{model}`. #5413 supplies that missing boundary before the picker UI.
[...]

PR #4919, opened 2026-08-27
✨ feat(tui): scaffold the terminal dashboard — deps, `hivectl tui`, blank frame

### LINKED ISSUE #4916: ✨ tui T1: scaffold — deps, `hive tui` entrypoint, blank frame
Part of #4907.

**Files to create/modify:** `go.mod`/`go.sum` (module under `v2/`),
`v2/internal/tui/app.go`, plus the minimal CLI wiring described below.

**Build this:**

1. Add `github.com/charmbracelet/bubbletea`,
   `github.com/charmbracelet/bubbles`, `github.com/charmbracelet/lipgloss`,
   and `github.com/charmbracelet/x/exp/teatest` (test only) at latest
   stable versions.
2. `v2/internal/tui/app.go`: exported `Run() error` starting a
   `tea.Program` in alt-screen mode. The model renders a single centered
   line, `Hive TUI (q to quit)`, and quits on `q` or `ctrl+c`.
3. Entrypoint: locate the CLI entrypoint for the v2 Go binary. If it uses
   a subcommand framework, register `tui` calling `tui.Run()`. If it does
   not, create `v2/cmd/hive-tui/main.go` with a `main` that calls
   `tui.Run()` and note which path you took in the PR description.

**Acceptance criteria:**
- `hive tui` (or `hive-tui`) launches, draws the line, exits on `q`
- teatest smoke test `TestAppQuits` sends `q` and asserts clean exit
- `go build ./...` and `go test ./v2/internal/tui/...` pass

**Out of scope:** panes, layout grid, API calls, config parsing.

**Verify with:** `go build ./... && go test ./v2/internal/tui/...`

---
Wave 1 of the #4907 filing protocol. Body copied from the epic's drafted
sub-issue text so the spec carries the context.

Danathar:
T1 of the #4907 TUI epic: the frame every later task builds on. One centered line, quits on `q` or `ctrl+c`, nothing else.

Two things in the spec did not survive contact with the repo. Both change paths that ~25 downstream sub-issues reference, so they are called out here rather than quietly resolved.

## Resolved: `v2/internal/tui/` → `src/pkg/tui/`

There is no `v2/` directory in this repo and no module at the repo root. The Go module is `src/` (`github.com/kubestellar/hive`), which is where `go build ./...` and every CI test shard run from. A root-level `v2/` would sit **outside** the module and never be compiled or tested at all — the acceptance criteria could not be met there.

It is not a typo either. **#3996 renamed the `v2/` source tree to `src/`**, which `src/pkg/reach/mapping.go:36-37` records and still compensates for:

> `#3996 renamed the v2/ source tree to src/, but a merged PR's file list is immutable history: PRs merged before the rename report v2/... paths from the forge API forever`

So the epic's paths are a pre-#3996 spelling, not a new directory anyone was asked to create. (Found while writing the T0 design doc, #4920.)

Within `src/`, the choice is `pkg/tui/` rather than `internal/tui/`: all of this module's non-`main` code already lives under `pkg/`, and there is no `internal/` anywhere in the tree, so `internal/tui` would introduce a second layout convention for one package.

**Mapping for the rest of the epic** — worth correcting in the open sub-issues:

| Spec path | Actual path |
|---|---|
| `v2/internal/tui/app.go` | `src/pkg/tui/app.go` |
| `v2/internal/tui/client/` (T2, T4, T6, T8, T10, T13a…) | `src/pkg/tui/client/` |
| `v2/internal/tui/panes/` (T3, T5, T7, T9, T11…) | `src/pkg/tui/panes/` |
| `v2/internal/tui/panes/testdata/` (golden files) | `src/pkg/tui/panes/testdata/` |
| `v2/docs/tui.md` (T0, #4915) | `src/docs/` or `docs/` — T0's call |

Verify command becomes `go test ./pkg/tui/...`, run from `src/`.

## Resolved: the entrypoint is `hivectl tui`

The epic asks for a subcommand "if the CLI entrypoint has subcommand structure", and says T1 resolves which. There are two candidates and only one qualifies:

- **`hive` (`cmd/hive`) is the spoke daemon, not a CLI.** It parses a single `-config` flag, takes a process-wide singleton flock whose entire purpose is to refuse a second `hive` process, then runs the agent manager, dashboard and heartbeat for the container's lifetime. There is no subcommand dispatch to hang `tui` off, and bolting an interactive foreground program onto that startup path would put an operator's TUI session up against the flock.
- **`hivectl` (`cmd/hivectl` + `pkg/hivectl/commands`) is the operator-facing cobra CLI** — and it already carries exactly the plumbing this epic specifies for the TUI. Its persistent `--server` defaults to `http://127.0.0.1:3001`, the Node proxy the epic names as the data source, and `--token-env` defaults to `HIVE_DASHBOARD_TOKEN`, the token the epic names. Later tasks that need a configured client inherit those instead of re-deriving them in a standalone `main`.

So: `hivectl tui`. If you want the literal `hive tui` spelling instead, it is a one-line change to `Use:` — say the word.

**One deviation this forces, worth flagging:** the epic's Data source decision names `HIVE_DASHBOARD_URL` (default `http://localhost:3001`) for the base URL, while hivectl spells the same thing `--server` (default `http://127.0.0.1:3001`). Reusing the existing flag seemed worth more than adding a second, differently-spelled way to say it — but if `HIVE_DASHBOARD_URL` should be honoured, the right shape is to make it the flag's default source rather than a parallel path. No client task depends on this yet; T2 (#4917) is the first that will.

## Dependencies

`bubbletea`, `lipgloss` and `teatest` (test-only) are added and all three are imported, so they land as direct requirements. The `go.mod` diff is **additive only** — no existing dependency is bumped.

**`bubbles` is deliberately not added.** Nothing in a blank frame imports it, and `go mod tidy` drops any module no package imports — I verified this rather than assuming it: adding `bubbles` and running `tidy` removes it again. The only way to hold it would be a blank-import holder file for a widget library nothing uses yet, which is worse than letting the first task that renders a widget (T3/T5) add it with its first real use.

## Tests

`TestAppQuits` drives the model through teatest rather than calling `Update` directly, so it fails if `tea.Quit` is returned as a value but never as a command. Alongside it: `ctrl+c`, the splash text, the unsized-`View` path (bubbletea can call `View` before the first `WindowSizeMsg`; centering into a 0×0 box renders a blank screen), and a guard that keys *outside* the quit set do not quit — without which a `default: return m, tea.Quit` slip would stay green.

Each test was mutation-checked. Breaking the behavior fails the test:

| Mutation | Result |
|---|---|
| `q` removed from the quit set | `TestAppQuits`, `TestAppRendersSplash` fail |
| centering replaced with a plain render | `TestViewCentersWhenSized` fails |
| every key quits | `TestUnhandledKeyDoesNotQuit` fails |
| `tui` unregistered from the root | all three hivectl tests fail |

The hivectl-side tests reach the real command through `root.Execute()` but stop at cobra's argument validation, which runs before `RunE` — so no test ever enters `tui.Run()` and tries to seize a terminal the runner does not have.

## Verification

Beyond the unit tests, I built `cmd/hivectl` and ran the real binary under a pty:

```
$ printf 'q' | script -qec "hivectl tui" /dev/null
exit_code=0
splash "Hive TUI (q to quit)"  -> 1 occurrence
alt-screen enter ESC[?1049h    -> 1
alt-screen leave  ESC[?1049l   -> 1
```

It launches, draws the line, enters alt-screen and **leaves** it — so the operator's scrollback is restored — and exits 0 on `q`. That covers all three behavioral acceptance criteria against the shipped binary, not just the model.

Also clean:

- `go build ./...` and `go vet ./...`
- `golangci-lint` with the repo config — 0 issues
- `gosec` on `pkg/tui` — 0 issues
- `govulncheck ./...` across the module with the new dependency tree — no vulnerabilities (this is a blocking gate, and new deps are exactly what could break it)
- `go test ./... -run XXX` — every test binary in the module still compiles

The 44 pre-existing `unused` findings on `v4` are untouched by this branch, and none are in the new code — checked, since #4914 is about to make that linter blocking.

## Update: coverage gate

The first push failed the per-package coverage gate — `tui: 83.3% (floor 90%)`. I had verified build, vet, golangci-lint, gosec and govulncheck, but not the coverage floor, which is on me: I checked the gates I knew about rather than the repo's actual required-check set.

`Run` was the only uncovered function, since a TUI entrypoint cannot be called from a test with no terminal. Rather than record a floor exemption for a package on the day it is added, `Run` is now a thin wrapper that supplies `os.Stdin`/`os.Stdout` around a `run(in, out)` that takes them — and bubbletea accepts non-TTY pipes via `WithInput`/`WithOutput`, so the test drives the real `tea.NewProgram` call with its real options.

That turned out to close a genuine gap rather than just a number. **teatest builds its own program internally, so none of the original tests ever executed the constructor in `app.go`** — `WithAltScreen` could have been dropped and the whole suite would still have passed, while `hivectl tui` scribbled over the operator's scrollback instead of taking its own screen. The new test asserts both alt-screen sequences, enter (`ESC[?1049h`) and leave (`ESC[?1049l`), because entering without leaving is what strands a terminal in the alternate buffer. Mutation-checked: deleting `WithAltScreen` fails it.

`pkg/tui` is now **92.3%**. The one remaining uncovered statement is `Run`'s `os.Stdin`/`os.Stdout` hand-off — the honest boundary, since covering it would mean taking over the test runner's own terminal. `pkg/hivectl` (94.2%) and `pkg/hivectl/commands` (93.4%) are unaffected.

## Out of scope

Panes, layout grid, API calls, config parsing — per the issue.

Part of #4907. Closes #4916

The path mapping and the resolved decisions above are now recorded canonically in the T0 design doc (#4920, `src/docs/design/tui.md`), which is the page later sub-issues should read.


Danathar:

Continued verbatim in #4935 — same head commit `27fcb34b`, recovered from `refs/pull/4919/head`, same branch, same `v4` base, no content changes. Please review there.

PR #4920, opened 2026-08-27
📖 docs(tui): record the design for `hive tui` (#4907 T0)

Danathar:
T0 of the #4907 TUI epic: the design record every later sub-issue reads before picking up work.

All five sections the issue asks for, in order — scope, the verbatim architecture decisions, the ASCII layout sketch, the keybinding table, the testing convention — plus two corrections that the sub-issue bodies do not carry and that change what several of them can do.

## The `v2/` paths are a pre-#3996 spelling

The epic specifies `v2/internal/tui/...` and `v2/docs/tui.md`. There is no `v2/` directory and no module at the repo root — the module is `src/`, which is where `go build` and every CI shard run from (`v2-tests.yml`, `working-directory: src`), so a tree at the repo root would never be compiled or tested at all.

It is not a typo. **#3996 renamed the `v2/` source tree to `src/`**, and `src/pkg/reach/mapping.go:36-37` both records it and still compensates for it:

> `#3996 renamed the v2/ source tree to src/, but a merged PR's file list is immutable history: PRs merged before the rename report v2/... paths from the forge API forever`

The doc carries the full mapping table so T2 through T27 do not each rediscover it.

## `dashboard/openapi.json` publishes no writes at all

This is the finding worth a maintainer's attention. The epic names the spec as the source of truth for every client task, with an escape hatch when an operation is missing. Measured at `45a13d5`, that hatch is going to fire constantly:

| | Published in `dashboard/openapi.json` | Registered by the dashboard |
|---|---:|---:|
| Distinct `/api` routes | 32 | 298 |
| `GET` | 32 | 137 |
| `POST` | **0** | 83 |
| `PUT` | **0** | 64 |
| `DELETE` | **0** | 14 |

The spec is GET-only. **All four Phase 2 action tasks have no contract to build against** — T14 pause/resume, T17 apply model, T19 apply ACMM, T20 kick now. Every one of those endpoints exists and is exercised by `hivectl` today; none is published.

Two read tasks are hit too, and one is named after the gap: **T2 (#4917) is "API client core + `/api/health`", and `/api/health` is not in the spec.** Neither is `/api/agents`, which T4 needs and `hivectl agent list` already calls.

The doc cites the live endpoint for each, so those tasks can file the scoped spec-gap issue *and* still ship. The alternative reading — that a missing spec entry blocks the task — would block six of them at once and leave the TUI read-only. Widening the spec to cover the write surface is worth its own issue; it is not something a single pane task should absorb.

## Also recorded

[...]

PR #4921, opened 2026-08-27
✨ feat(tui): API client core and the health check (#4907 T2)

### LINKED ISSUE #4917: ✨ tui T2: API client core + `/api/health`

Danathar:
T2 of the #4907 TUI epic: the one place that knows how a request to the dashboard is addressed and authenticated. `Client`, `New()`, `Health()`, and the shared `getJSON()` helper the pane-client tasks (T4/T6/T8/T10) build on.

Stdlib only — `go.mod` and `go.sum` are untouched.

## The auth header, and why it isn't a guess

The issue says to send the token "the same way the web dashboard does per `dashboard/openapi.json` security scheme (**check the spec; do not guess the header**)".

Checked. **The spec has no security scheme.** It has no `components` block at all — the whole document is `openapi`, `info`, `servers`, `tags`, `paths` — so there is no `securitySchemes`, no top-level `security`, and no per-operation `security` to read. `/api/health` isn't in it either.

```
$ python3 -c "import json;d=json.load(open('dashboard/openapi.json'));print(list(d.keys()), d.get('security'), d.get('components'))"
['openapi', 'info', 'servers', 'tags', 'paths'] None None
```

So rather than guess, the header is transcribed from the two things that actually *verify* it, which agree:

- **`src/proxy/server.js` `requireAuth()`** matches `/^Bearer\s+(.+)$/i` against `req.headers.authorization` and timing-safe compares against `HIVE_DASHBOARD_TOKEN`.
- **`src/pkg/hivectl/client.go`** sets `Authorization: Bearer <token>` from the same environment variable.

Hence `Authorization: Bearer $HIVE_DASHBOARD_TOKEN` — and **omitted entirely** when no token is set, because a bare `Bearer ` fails that regex and would turn a public endpoint into a 401. Both behaviours are pinned by tests.

## No duplicate spec-gap issue

The issue's note says filing a scoped spec-gap issue completes this task. **#4912 is already open and already reports exactly this**, down to recommending "the `HIVE_DASHBOARD_TOKEN` security scheme". Filing a second would be noise, and stopping at a filing would leave T4/T6/T8/T10 blocked behind a client that can be written correctly *today* from the verifier.

I've posted corrections to #4912's route inventory there as a comment rather than duplicating them here — briefly: it measured `dashboard/server.js`, but the container entrypoint starts `src/proxy/server.js`, which proxies `/api/*` to the Go API, so the real undocumented surface is larger than 69 routes.

## Where `/api/health` actually lives

It is a **Go dashboard** route (`src/pkg/dashboard/server.go`) on the `isPublicPath` list. It is reachable at the epic's `:3001` base URL because `src/proxy/server.js` proxies `/api/*` to the Go API on `:3002` — confirmed independently by `src/pkg/hub/heartbeat.go`, which probes `http://localhost:3001/api/health`, and by the compose healthcheck.

That public status is load-bearing for the TUI, and is why `Health()` works without a token: a tokenless health check is what separates *"the dashboard is unreachable"* from *"the dashboard is there and my token is wrong"*, and those are different things to tell an operator.

## Carried over: the transport lesson

A nil `Transport` silently shares `http.DefaultTransport`, whose idle pool **every** parallel `httptest` teardown calls `CloseIdleConnections` on. `src/pkg/hivectl/client.go:69` records that this already produced a flaky "connection broken" from an unrelated test in this repo. This client clones its own transport, and a test asserts it rather than trusting the comment.

## Reuse was considered

`pkg/hivectl`'s client speaks the same API with the same token and would have brought its SSE reader along. It isn't wrapped because its `Do()` returns `any` — it feeds the CLI's table/json/yaml printer, so a TUI decoding into typed structs would re-marshal that map on every poll — and its constructor returns an error, which this task's `New() *Client` contract cannot. The reasoning is in the package doc so the next reader doesn't have to redo it. (Its `StreamSSE` is still worth a look for T13a.)

## Tests

httptest fixtures assert path, method, `Authorization` and `Accept`; the tokenless path; `APIError` across 401/404/502/500-with-empty-body; the error-body cap; and `getJSON`'s decode, malformed-body, transport-error and context-cancellation paths, plus both environment variables and the default.

Every load-bearing assertion is mutation-checked:

| Mutation | Test that catches it |
|---|---|
| `Bearer` → `Token` | `TestHealthSendsExpectedRequest` |
| send `Authorization` with an empty token | `TestHealthOmitsAuthorizationWithoutToken` |
| drop the trailing-slash trim | `TestNewReadsEnvironment` |
| share `http.DefaultTransport` | `TestNewUsesItsOwnTransport` |
| unbound the error body | `TestHealthTruncatesHugeErrorBody` |

## Verification

- `go build ./...`, `go vet ./...` — clean
- `golangci-lint` with the repo config — 0 issues; also 0 `unused` findings in the new package, since #4914 will make that linter blocking
- `gosec` on `pkg/tui` — 0 issues
- every test binary in the module compiles
- **`pkg/tui/client` is at 96.9%**, against the gate's 90% floor (that gate caught #4919 on its first push, so it is checked up front here)

One honest gap: **`govulncheck` could not be run locally** — v1.1.4 crashes in the `x/tools` SSA builder on the Go 1.27 toolchain on this machine. CI runs it on Go 1.25. Since this branch adds no dependencies, its vulnerability surface is identical to `v4`.

## Out of scope

Any endpoint other than health, SSE, retries/backoff — per the issue.

Part of #4907. Closes #4917

Independent of #4919 (T1) and #4920 (T0): new package, no shared files, no `go.mod` change — it can merge in any order. The path is `src/pkg/tui/client/` per the mapping in #4920, since the epic's `v2/` prefix is a pre-#3996 spelling.

### DIFF
--- a/src/pkg/tui/client/client.go
+++ b/src/pkg/tui/client/client.go
@@ -1,0 +1,182 @@
+// Package client is the TUI's HTTP client for the Hive dashboard API
+// (kubestellar/hive#4907).
+//
+// It is deliberately small: one transport, one auth header, one JSON decode
+// helper. Every later TUI task adds typed methods to this package rather than
+// its own HTTP handling, so there is exactly one place that knows how a request
+// to the dashboard is addressed and authenticated.
+//
+// WHY NOT REUSE pkg/hivectl's CLIENT. It was evaluated. It speaks to the same
+// API with the same token and would have brought its SSE reader along, but its
+// Do() returns `any` because it feeds the CLI's table/json/yaml printer — a TUI
+// decoding into typed structs would have to re-marshal that map on every poll —
+// and its constructor returns an error, which this package's New() cannot per
+// its contract. What it does have that is worth copying outright is the
+// transport lesson recorded at src/pkg/hivectl/client.go:69; see newTransport.
+package client
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+)
+
+const (
+	// DefaultBaseURL is where the dashboard listens on a self-hosted hive: the
+	// Node auth proxy (src/proxy/server.js), not the Go API behind it on :3002.
+	// The proxy is where the token check and the SSE stream live, and it
+	// forwards /api/* to the Go API, so every endpoint the TUI needs is
+	// reachable here.
+	DefaultBaseURL = "http://localhost:3001"
+
+	// BaseURLEnv and TokenEnv are the two environment variables the TUI reads.
+	// TokenEnv is the same variable hivectl's --token-env defaults to, so an
+	// operator who has hivectl working has the TUI working.
+	BaseURLEnv = "HIVE_DASHBOARD_URL"
+	TokenEnv   = "HIVE_DASHBOARD_TOKEN"
+
+	// requestTimeout bounds a single dashboard request. The TUI repaints on a
+	// timer, so a request that outlives its frame is not worth waiting for —
+	// better to fail the pane and retry on the next tick than to stall the UI
+	// behind a hung socket.
+	requestTimeout = 5 * time.Second
+
+	// maxErrorBodyBytes caps how much of a failed response is read back into an
+	// APIError. Enough to carry the dashboard's `{"error":"..."}` payload, small
+	// enough that an HTML error page from a misconfigured reverse proxy cannot
+	// become a multi-megabyte string held in the model.
+	maxErrorBodyBytes = 4 << 10
+)
+
+// APIError is returned when the dashboard answers with a non-2xx status.
+//
+// It carries the status and the (truncated) body rather than collapsing to a
+// string so a pane can distinguish the cases that mean different things to an
+// operator: 401 is a missing or wrong HIVE_DASHBOARD_TOKEN, 404 is a dashboard
+// too old to serve the endpoint, 502 is the proxy up but the Go API behind it
+// down.
+type APIError struct {
+	StatusCode int
+	Path       string
+	Body       string
+}
+
+func (e *APIError) Error() string {
+	if e.Body == "" {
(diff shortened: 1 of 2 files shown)


Danathar:

Continued verbatim in #4937 — same head commit `713c1c6d`, recovered from `refs/pull/4921/head`, same branch, same `v4` base, no content changes. Please review there.

Issue #4918, opened 2026-08-27
Default agent launch path is unconfined on the operator's host — an agent running an assigned repo's test suite reached my bootloader


Danathar:
## Summary

Hive's default agent launch path runs the backend CLI with all permission gating disabled, as the operator's own user, directly on the operator's host — with nothing confining it to the assigned workspace. An agent doing entirely benign, correct work reached and attempted to modify my machine's **bootloader configuration**. It was stopped only by polkit, i.e. only because the run happened to be unprivileged.

`src/docs/sandbox-isolation.md` states the intended guarantee:

> If an agent is compromised, it can only modify its mounted workspace; it cannot receive GitHub tokens or push directly.

The second half of that sentence holds. The first half does not hold on the default path.

## What happened

A hive contributor task assigned an agent to `projectbluefin/bluefin#879`. The agent cloned the repo into `$HIVE_WORKSPACE_DIR` and ran the repo's own unit suite (`bats tests/unit/`) — an ordinary, expected step for that task.

Two tests in that suite had a latent defect: in their "missing rpm-ostree" case they reset `PATH` to `${STUB_BIN}:/usr/bin:/bin`, re-admitting `/usr/bin`. On an rpm-ostree host the stubbing is defeated, the real binary is found, and the hook under test runs for real. The result, from my host's journal:

```
Aug 27 17:35:23 bluefin polkitd[1260]: Operator of unix-session:15 FAILED to authenticate
  to gain authorization for action org.projectatomic.rpmostree1.bootconfig
  for system-bus-name::1.18895
  [rpm-ostree kargs --append-if-missing=usbcore.autosuspend=-1]
  (owned by unix-user:dbaggett)
```

Three attempts, 17:35:10–17:35:20, each raising an interactive polkit dialog on my desktop. The requesting process:

```
CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/tmux-spawn-b50290bd-….scope
        ├─ 877294 -bash
        └─ 877588 claude --dangerously-skip-permissions --permission-mode bypassPermissions
```

That is a hive-spawned pane (`tmux` session `hive-claude-bc59`), running as uid 1000 in the host's own user slice, with the flag pair produced by `config/backends.conf:46`.

[...]

PR #4922, opened 2026-08-27
🔒 fix(agent): deny host-state commands on the unconfined claude launch path

Danathar:
Closes the gap #4918 reports, on the path people are actually on today.

## What happened, restated

An agent doing entirely correct work reached the operator's bootloader. It was assigned a third-party repo, cloned it, and ran that repo's own unit suite — an ordinary step for the task. A latent defect in two of those tests let a hook escape its stubs, so `rpm-ostree kargs --append-if-missing=usbcore.autosuspend=-1` ran against the operator's **real** deployment, raising three polkit dialogs on their desktop. Nothing was written, and the only reason is that the process happened to lack privilege.

No compromise was involved. The repo was legitimate, the agent behaved correctly, and the command came from that project's own test suite.

## The fix

The claude family (`claude`, `litellm`) launches permissions-bypassed, as the operator's user, on the operator's host, with nothing scoping it to the workspace. Hive already knows the right posture — codex has run `--sandbox workspace-write` by default since #3512, and the comment there is backend-agnostic:

> Hive-delivered work is unattended even in the tmux-backed mode: no operator is guaranteed to be watching the pane

This applies the equivalent floor to the default backend.

**Denials, not a permission mode.** Switching claude's mode would make the CLI *prompt*, and nobody is attached to an unattended pane to answer — the exact hang that codex comment warns about. `--disallowed-tools` refuses the call outright and the agent moves on, so the bypass flags stay and the denials ride alongside them.

**What is denied, and why those two families:**

| Family | Commands |
|---|---|
| Privilege escalation | `sudo`, `pkexec`, `doas`, `su` |
| Host boot/deployment state | `rpm-ostree`, `bootc`, `ostree`, `grubby`, `bootctl`, `efibootmgr` |

The second family is the one that matters and is easy to miss. **`rpm-ostree` never invoked `sudo` — it asked polkit directly**, so a list covering only escalation would not have stopped the reported incident at all.

The list is deliberately minimal, and it is a floor, not a sandbox: it names commands rather than enforcing a boundary. Package managers are left off because the ones needing root are already neutered by the `sudo` denial, and denying them outright would break legitimate agent work.

## Verified against the real CLI, not assumed

Claude Code 2.1.231, before writing any of it:

| Check | Result |
|---|---|
| Does a deny survive `--dangerously-skip-permissions --permission-mode bypassPermissions`? | **Yes** — denied; control run without the deny executes the same command |
| Is matching a raw prefix (would over-block)? | **No** — `Bash(ec:*)` does *not* block `echo`, so `Bash(su:*)` won't block `subl` |
| Do comma-separated patterns work in one argv word? | **Yes** — required, because the shell path word-splits its flag string |
| End-to-end with the argv this change produces | `rpm-ostree kargs` **denied**; `git --version` still runs |

The first row is worth noting beyond this PR: it also confirms the existing `claudeGitHubWriteDenyFlags` protection is not vacuous, which nothing had established.

## Both launch paths, held together by a test

`config/backends.conf` covers the relay; `src/pkg/agent/manager.go` covers pod agents. Same list, and the same truthiness rule for the opt-out — an operator who writes `yes` must not get a confined agent on one path and an unconfined one on the other, which is the worst outcome because it looks like it worked. `TestShellAndGoDenyListsAgree` sources the shell file and compares, so the two cannot drift.

Escape hatch: `HIVE_CLAUDE_DANGEROUSLY_ALLOW_HOST_STATE=1`, named after the codex one so they read alike.

## Docs

`sandbox-isolation.md` stated *"if an agent is compromised, it can only modify its mounted workspace"* as an unconditional property. It is not one — it holds on the opt-in Podman path, not the default. The threat model now separates the half that holds everywhere (credentials, pushes) from the half that does not, and says so with the incident. `backend-setup.md` and `contributor-relay.md` document the new denials and the opt-out next to their codex equivalents.

## Tests

Shell suite pins the default posture, the `rpm-ostree` entry, the escalation entries, litellm parity, the opt-out dropping the flag *entirely* (not leaving a dangling `--disallowed-tools`), and the word-split contract — exactly 5 argv words with the list intact in the 5th. Go tests pin the same list, every truthy spelling of the opt-out, that GitHub MCP denials survive alongside, and shell/Go parity.

Mutation-checked:

| Mutation | Caught by |
|---|---|
| drop `rpm-ostree` from the list | shell suite + `TestShellAndGoDenyListsAgree` |
| introduce a space (breaks word-splitting) | shell suite (6 argv words) + parity test |
| stop emitting denials entirely | shell suite |

## Verification

`go build ./...`, `go vet ./...`, `golangci-lint` with the repo config (0 issues), the full `pkg/agent` suite (255s, pass), `bash bin/contributor-agent.test.sh` (pass), `bin/test_bin_suites_wired.sh` (18 suites, 0 orphaned). `gosec` on `pkg/agent` is 60 findings before and 60 after — no new ones. Shellcheck's two findings on `backends.conf` (SC2148, SC2034) are pre-existing.

## Not fixed here, deliberately

#4918 also asks whether the Podman sandbox should stop being double-gated off by default — both `agent_sandbox.enabled` and a per-agent `sandbox.enabled` must currently be set. That is an OS-level posture change with real blast radius for existing fleets: flipping it would sandbox agents on hosts that may not have podman or the image. It deserves its own PR and its own argument, so **please leave #4918 open for it** — this PR is the floor for the path people are on today, not the ceiling the issue asks for.

Refs #4918

### DIFF
--- a/bin/contributor-agent.test.sh
+++ b/bin/contributor-agent.test.sh
@@ -445,4 +445,127 @@ if output="$(
   exit 1
 fi
 
-echo "contributor-agent codex contract tests passed"
+# ── claude host-state denials (#4918) ───────────────────────────────────
+#
+# The claude family runs permissions-bypassed on the operator's own host. #4918
+# is what that cost: an agent running an assigned repo's test suite issued
+# `rpm-ostree kargs --append-if-missing=...` against the operator's real
+# deployment, and was stopped only by lacking privilege. These cases pin the
+# denials that now sit on that path.
+
+claude_flags_output="$(
+  env -i \
+    PATH="${PATH}" \
+    HOME="$HOME_DIR" \
+    HIVE_REGISTRATION_TOKEN="test-token" \
+    HIVE_ENTRYPOINT_HOOK_DIR="${WORK_DIR}/empty-entrypoint.d" \
+    HIVE_WORKSPACE_DIR="${WORK_DIR}/workspace" \
+    HIVE_CONTRIBUTOR_AGENT_TEST_RESOLVE_BACKEND=1 \
+    AGENT_BACKEND=claude \
+    bash "${ROOT_DIR}/bin/contributor-agent.sh"
+)"
+# The bypass flags stay — an unattended agent that stops to ask permission just
+# hangs — but the host-state denials ride alongside them.
+case "$claude_flags_output" in
+  *"backend_perm_flag=--dangerously-skip-permissions --permission-mode bypassPermissions --disallowed-tools "* ) ;;
+  *)
+    echo "expected claude to carry host-state denials by default; got:" >&2
+    echo "$claude_flags_output" >&2
+    exit 1
+    ;;
+esac
+# The command from the incident specifically.
+case "$claude_flags_output" in
+  *"Bash(rpm-ostree:*)"* ) ;;
+  *)
+    echo "expected claude denials to cover rpm-ostree, the command in #4918; got:" >&2
+    echo "$claude_flags_output" >&2
+    exit 1
+    ;;
+esac
+# rpm-ostree reached polkit without sudo, so escalation denials alone are not
+# the fix — but they must be there too, or `sudo rpm-ostree` walks around it.
+case "$claude_flags_output" in
+  *"Bash(sudo:*)"*"Bash(pkexec:*)"* ) ;;
+  *)
+    echo "expected claude denials to cover privilege escalation; got:" >&2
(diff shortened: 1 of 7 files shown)


Danathar:

Continued verbatim in #4938 — same head commit `0a5a6a3f`, recovered from `refs/pull/4922/head`, same branch, same `v4` base, no content changes. Please review there.

PR #4925, opened 2026-08-27
🔒 fix(agent): reap the processes an agent spawns, not just its CLI

### LINKED ISSUE #4924: Agent-spawned background processes are never reaped on the workstation path — the CLI-cmdline predicate skips them
## Summary

Background processes an agent spawns — pollers, `sleep` loops, servers, anything that is not the backend CLI itself — are **never cleaned up on the contributor/workstation path**. They survive the task that created them and keep running between kicks, indefinitely.

Reported by an operator who has seen this happen more than once. The code path below explains why it happens and why it is specific to the non-UID-isolated deployment.

## Root cause: the reaper's cmdline predicate

`reapAgentCLI` (`src/pkg/agent/manager.go:9227`) requires **two** conditions to kill a process, not one:

```go
cmdline := strings.ReplaceAll(string(cmdlineRaw), "\x00", " ")
if !containsCLIMarker(cmdline) {      // (1) cmdline must name a CLI binary
    continue
}
...
if !environHasMarker(string(environRaw), marker) {   // (2) HIVE_AGENT=<name>
    continue
}
```

`containsCLIMarker` matches against a fixed list (`manager.go:9202`):

```go
var cliProcessMarkers = []string{"claude", "copilot", "gemini", "goose", "bob"}
```

A background child of the agent — `bash -c 'until …; do sleep 60; done'`, a dev server, a watcher — **inherits `HIVE_AGENT` and so passes (2), but fails (1)**. Its cmdline names `bash` or `sleep`, not a backend CLI. It is skipped and lives on.

This is not a subtle oversight — hive already knows about it. The other reaper exists precisely to cover the case, and its call site says so (`manager.go:9127-9131`):

```go
if agent.UID > 0 {
    // UID isolation on: also sweep any non-CLI helper processes (MCP
    // servers, hung copilot binaries) owned exclusively by this agent.
    killed := killAgentProcesses(agent.UID, m.logger)
```

## Why it only bites the workstation path

`killAgentProcesses` matches by owner UID and catches everything, but it is gated on `agent.UID > 0` and floor-guarded at `minAgentUID` (= `baseAgentUID`, 2001) — as root, a `uid == 0` sweep would SIGKILL every root process, so it correctly refuses (`manager.go:9320-9331`).

That produces two different behaviours from one code path:

| Deployment | `agent.UID` | Reapers that run | Background children cleaned up? |
|---|---|---|---|
| Containerized spoke, per-agent UIDs | ≥ 2001 | `reapAgentCLI` + `killAgentProcesses` | **yes** |
| Contributor relay / `contribute-hive` on a workstation | 0 (one user, no isolation) | `reapAgentCLI` only | **no** |

The unprotected column is the contributor's own machine — the same environment and the same blast radius as #4918.

## Second dimension: cleanup is tied to launch, not to task completion

Every `reapAgentCLI` call site is a **launch, restart, or stop** (`manager.go:2725`, `9126`, `9435`). Nothing reaps when a task simply *finishes*.

So even where the sweep does work, a process leaked at the end of a task runs unattended until the agent is next kicked or restarted. On an idle agent that window is unbounded.

## Why it matters

- **Nobody is watching.** These are unattended agents by design; a leaked poller on a workstation produces no signal until someone looks at `ps`.
- **It costs real resources.** A leaked `gh`/API poll loop burns rate limit and quota continuously. A leaked build/dev server holds a port, and the next task that wants that port fails for a reason unrelated to anything it did.
- **It compounds.** Each task can leave another one behind; nothing ever collects them.

## Suggested directions

Not prescriptive, but the ordering seems clear:

1. **Drop the cmdline predicate for the env-marker sweep.** `HIVE_AGENT=<name>` in `/proc/<pid>/environ` already identifies the agent exactly — `environHasMarker` compares whole NUL-separated entries specifically so `scanner` cannot match `scanner-2`. The `containsCLIMarker` filter narrows a marker that is already precise, and is what lets non-CLI children through. Removing it would make `reapAgentCLI` cover what `killAgentProcesses` covers today, without needing UID isolation. Worth checking against the self-skip and any process hive itself launches with that env set.
2. **Reap on task completion, not only on launch.** Whatever marks an agent idle after a kick is the natural place; that is where the leak actually begins.
3. **At minimum, report it.** Count matching processes at kick time and log when an agent has children left over from a previous task. Even without killing anything, that turns an invisible leak into something an operator can see.

Direction 1 is the smaller change and closes the larger hole; 2 is what makes it timely.

## Environment

- Observed on the contributor relay path (`contribute-hive`) on an operator workstation — no per-agent UID isolation, so `agent.UID == 0`
- Code references against `v4` at `45a13d5`

Related: #4918 (same path, same "unconfined on the operator's host" blast radius).

Danathar:
> [!IMPORTANT]
> **Requesting an LLM review pass from a maintainer before merge.** This changes the predicate on a `SIGKILL` loop that runs against `/proc` on an operator's own machine. I have reasoned about the blast radius carefully and mutation-tested every guard (below), but the cost of being wrong here is high and asymmetric, so a second automated review on validation and safety is worth the cycles. `/code-review ultra` or your usual pass — whatever you run, please run it against this one before merging.

Fixes #4924.

## Root cause

`reapAgentCLI` required **two** predicates, not one:

```go
if !containsCLIMarker(cmdline) { continue }              // cmdline names a backend binary
if !environHasMarker(string(environRaw), marker) { continue }  // HIVE_AGENT=<name>
```

with `cliProcessMarkers = {"claude", "copilot", "gemini", "goose", "bob"}`.

A child the agent spawned inherits `HIVE_AGENT` and passes the second check, but its cmdline reads `bash` or `sleep`, so it fails the first and is skipped. It then outlives every task.

Hive already knew about this gap — `killAgentProcesses` exists to *"sweep any non-CLI helper processes"*, per its own call-site comment. But it matches by owner UID and is gated on `agent.UID > 0`, floor-guarded at 2001 so a `uid == 0` sweep as root cannot SIGKILL every root process. On a workstation there is no per-agent UID, so it never runs:

| Deployment | `agent.UID` | Sweeps that run | Spawned children collected? |
|---|---|---|---|
| Containerized spoke | ≥ 2001 | both | yes |
| `contribute-hive` on a workstation | 0 | `reapAgentCLI` only | **no** |

## The fix

Make the env marker the sole predicate; replace the CLI allowlist with a narrow exclusion list.

### Why the marker alone is safe to kill on — the part to review

`HIVE_AGENT` reaches a process exactly two ways, and both mean *"hive started this, for this agent"*:

1. inlined into the typed launch command (`buildEnvPrefix`), so the CLI and everything it forks inherit it;
2. set on the tmux session by `ensureTmuxSession`.

**The ordering inside `ensureTmuxSession` is load-bearing, and I verified it rather than assuming it:** the session — and therefore the pane's own bash — is created at lines 2057-2069, *before* the `set-environment` loop at 2096. So that shell never carries the marker and is never reaped. `RelaunchBobAgentsAwaitingKey` depends on the same ordering and records a live confirmation of it (`BOBSHELL_API_KEY present in the tmux session env, absent from every bob /proc/<pid>/environ`).

`environHasMarker` already compares whole NUL-separated entries, so `scanner` cannot reach `scanner-2`.

Two guards, explicit rather than assumed:

- **tmux is excluded** even carrying the marker. It should never match (hive starts the server with `exec.Command`, inheriting hive's own env), but it owns every pane, so one mis-match takes the fleet's panes down. Substring-matched deliberately — it over-excludes, which is the direction an exclusion list should fail in.
- **PID 1 is never signalled**, split into `reapSkipPID` so the guard is directly testable.

### One behaviour change, stated plainly

A shell an operator opens inside the agent's tmux session **after** launch does carry the marker and is now reaped. Judged acceptable because every caller — pre-launch, restart, stop — is tearing that agent down at that moment anyway. Flagging it explicitly because it is the one case where this is not a strict improvement.

## Two tests that were vacuous

The existing fake-proc cases asserted the old behaviour using pids that **do not exist**, so `syscall.Kill` failed and the counter never moved — they passed whether the predicate matched or not. They now use real child processes. Two further traps found while writing the new ones:

- **SIGKILL leaves a child of the test process as a zombie** until it is waited on, and signal 0 succeeds against a zombie — so a liveness probe reports the *opposite* of the truth. `assertKilled` waits and reads the exit status instead.
- **An unprivileged test cannot kill real PID 1**, so a count-based assertion passes with or without the guard. Confirmed by mutation: deleting the guard left that test green. `reapKill` is now a test seam (like the existing `procRoot`) so the test asserts what the sweep *decided* to signal rather than what it *managed* to.

## Mutation matrix

Every guard is proven to be load-bearing:

| Mutation | Caught by |
|---|---|
| PID-1 guard removed from the predicate | `TestReapSkipPID`, `TestReapNeverSignalsPIDOne` |
| guard no longer called in the loop | `TestReapNeverSignalsPIDOne` |
| tmux exclusion emptied | `TestReapSkipsTmux`, `TestReapExcluded` |
| exclusion inverted | `TestReapAgentCLI_FakeProc`, `TestReapsAgentSpawnedChild` |
| agent-marker check dropped | reap suite |
| marker loosened to a prefix | `TestReapAgentCLI_FakeProc`, `TestReapsAgentSpawnedChild` |

## Also fixed while here

`reapAgentCLI`'s doc comment was separated from the function by the `procRoot` declaration, so it documented nothing. The vars now sit above it.

## Verification

- `go build ./...`, `go vet ./...` — clean
- `golangci-lint` with the repo config — 0 issues; also 0 `unused` findings in the touched files, since #4914 will make that linter blocking
- `gosec` on `pkg/agent` — **60 findings before and after**, no new ones
- every test binary in the module compiles
- full `pkg/agent` suite under CI flags (`-short -race`) — passes, **91.4%** coverage against the package's recorded floor of **89**

## Not covered here

The second half of #4924 — reaping on *task completion* rather than only on launch/restart/stop — is untouched. Every `reapAgentCLI` call site is still a teardown, so a process leaked at the end of a task survives until the next kick. That is a scheduling change rather than a predicate change and wants its own PR; **please leave #4924 open for it.**

### DIFF
--- a/src/pkg/agent/helpers_coverage_test.go
+++ b/src/pkg/agent/helpers_coverage_test.go
@@ -10,8 +10,8 @@ import (
 
 // ---------------------------------------------------------------------------
 // reapAgentCLI / killAgentProcesses — /proc-based; on non-Linux /proc is
-// absent so these exercise the ReadDir-error branch. containsCLIMarker and
-// environHasMarker are pure and always testable.
+// absent so these exercise the ReadDir-error branch. reapExcluded,
+// reapSkipPID and environHasMarker are pure and always testable.
 // ---------------------------------------------------------------------------
 
 func TestReapAgentCLI_NoProc(t *testing.T) {
--- a/src/pkg/agent/manager.go
+++ b/src/pkg/agent/manager.go
@@ -9194,36 +9194,74 @@ func (m *Manager) RestartThenSendKick(ctx context.Context, name, message string)
 	return m.SendKick(name, message)
 }
 
-// cliProcessMarkers are substrings that identify a CLI process in its
-// /proc/<pid>/cmdline. The Claude CLI (and inference backends, which also use
-// it) runs as `claude` (often re-exec'd via node); copilot/gemini/goose/bob run
-// under their own names. Matching cmdline substrings catches the CLI regardless
-// of the interpreter the process reports as its comm name.
-var cliProcessMarkers = []string{
-	"claude",
-	"copilot",
-	"gemini",
-	"goose",
-	"bob",
-}
+// reapExclusions are cmdline substrings that must NEVER be SIGKILLed by the
+// marker sweep, even when the process carries this agent's HIVE_AGENT.
+//
+// Only tmux is listed, and it is defensive rather than observed: hive starts
+// the tmux server itself with exec.Command, which inherits hive's own
+// environment and therefore has no HIVE_AGENT — so it should never match. But
+// the server owns every pane for the agent (and, on a shared socket, for other
+// agents), so a single mis-match would take the fleet's panes down with it.
+// That asymmetry is worth one string compare.
+//
+// Matched as a SUBSTRING, deliberately: the exclusion should fire on
+// "/usr/bin/tmux" and "tmux -S /tmp/... attach" alike. It therefore also spares
+// an agent process that merely mentions tmux in its arguments. That is a
+// false NEGATIVE — one leaked process not reaped — which is the direction an
+// exclusion list should fail in.
+var reapExclusions = []string{"tmux"}
 
-// reapAgentCLI finds and SIGKILLs every CLI process belonging to the given
-// agent, matched by the HIVE_AGENT=<name> marker in /proc/<pid>/environ. This
-// marker is inlined into every launch command (buildEnvPrefix) and set on the
-// tmux session (ensureTmuxSession), so it uniquely identifies an agent's CLI
-// processes — unlike UID matching, which cannot distinguish agents that share
-// the dev UID (UID isolation disabled). Returns the number of processes killed.
+// procRoot is the /proc mount the reaper scans. A var (not const) so tests can
+// point it at a fake proc tree on non-Linux hosts; production value is "/proc"
+// and nothing on the launch path mutates it.
+var procRoot = "/proc"
+
+// reapKill is the kill syscall the marker sweep uses. A var (not a direct
+// syscall.Kill call) for the same reason procRoot is: it is the only way a test
+// can assert what the sweep DECIDED to signal, as opposed to what it managed to
+// signal. Those differ exactly where it matters — an unprivileged test cannot
+// kill PID 1, so a count-based assertion passes whether or not the guard that
+// skips init is present. Production never reassigns it.
+var reapKill = syscall.Kill
+
+// reapAgentCLI finds and SIGKILLs every process belonging to the given agent,
+// matched by the HIVE_AGENT=<name> marker in /proc/<pid>/environ. Returns the
+// number of processes killed.
 //
 // This is the single-CLI guarantee: before every (re)launch, any pre-existing
 // or leaked CLI for the agent is terminated, so an agent can never accumulate
 // concurrent claude processes on different models. tmux kill-session alone is
 // insufficient — a detached node/claude child can survive the session's SIGHUP
 // and keep hitting the gateway (403-flooding the pane on a stale model).
-// procRoot is the /proc mount the reaper scans. A var (not const) so tests can
-// point it at a fake proc tree on non-Linux hosts; production value is "/proc"
-// and nothing on the launch path mutates it.
-var procRoot = "/proc"
-
+//
+// #4924: IT ALSO REAPS THE CLI'S CHILDREN, and that is the point of the marker
+// being the only predicate. This used to require the cmdline to name a backend
+// binary (claude/copilot/gemini/goose/bob), which meant a process the agent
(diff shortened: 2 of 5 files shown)


Danathar:
Closing at the operator's request. On closer inspection the safety case for widening this kill predicate was weaker than the PR claimed — HIVE_AGENT is exported by bin/agent-launch.sh (not merely inlined onto the CLI), so the marked set includes the launch script and in-flight gh processes, and the analysis rested on code reading rather than observation of a live hive. Not confident enough to merge a SIGKILL-surface change on that basis. The underlying observation may be revisited later with a measurement-first approach if warranted.

PR #4930, opened 2026-08-28
🐛 fix: base PRs on the repository's default branch, not a hardcoded "main"

### LINKED ISSUE #4928: 🐛 bug: PR base is hardcoded to 'main' and the repository default branch is never consulted
## What happened?

Hive agents open every pull request against `main` instead of the repository's default branch. We run the kubestellar-hive App on [projectbluefin/dakota](https://github.com/projectbluefin/dakota), where the default branch is `testing`, and all seven PRs the App has filed there were opened with `main` as the base. The two that currently show `testing` ([#1402](https://github.com/projectbluefin/dakota/pull/1402) and [#1391](https://github.com/projectbluefin/dakota/pull/1391)) only look correct because I retargeted them manually after they were filed.

The wrong base makes the PRs unreviewable and dangerous to merge. [dakota#1411](https://github.com/projectbluefin/dakota/pull/1411) is the clearest example: it shows 233 changed files and roughly 260k changed lines because the diff includes the entire divergence between `main` and `testing`, and it is large enough that GitHub's diff API refuses to return it. The actual change in that PR touches one file. If a maintainer misses the base on a smaller PR, the change merges to the wrong branch entirely.

The full list on our repo: [#1420](https://github.com/projectbluefin/dakota/pull/1420), [#1411](https://github.com/projectbluefin/dakota/pull/1411), [#1386](https://github.com/projectbluefin/dakota/pull/1386), [#1309](https://github.com/projectbluefin/dakota/pull/1309), and [#1276](https://github.com/projectbluefin/dakota/pull/1276) were all opened against `main`, plus the two I retargeted above. That is every PR the App has ever filed here, so this looks like a hardcoded assumption rather than an intermittent resolution failure.

## Expected behavior

Agents should resolve the target repository's default branch before opening a PR and use it as the base. The App already has metadata read permission, so `default_branch` is one API call away (`gh repo view --json defaultBranchRef` or the equivalent REST field).

## Reproduction steps

1. Install the hive App on a repository whose default branch is not named `main` (ours is `testing`, with `main` still present as a non-default branch).
2. Let any agent (scanner, quality) pick up an issue and open a PR.
3. The PR is opened with `base: main`, and the diff includes everything the default branch has that `main` does not.

## Environment

- Branch/version: unknown, we consume the hosted App (`kubestellar-hive`, App ID 3568013). PR footers report `agent=scanner backend=copilot model=gpt-5.6-luna` and `ACMM L5 (hold-gated mode)`.
- Deployment method: hosted Hive Hub, GitHub App integration.
- Target repository: projectbluefin/dakota, default branch `testing`.

## Logs or screenshots

From the GitHub API, `baseRefName` for every hive PR on our repo as filed: `main` for #1420, #1411, #1386, #1309, and #1276. #1402 and #1391 were also filed against `main` and were retargeted to `testing` by hand.

## Additional context

This compounds with the duplicate-PR issue I am filing separately: #1411 is both a duplicate of #1402 and mis-based, so the two bugs together produced an unreviewable duplicate. Happy to test a fix from our side, we have the App running on the repo daily.

Danathar:
## Summary

- **What this touches.** Hive opens pull requests on other people's repositories on behalf of its agents. Every PR needs a *base* — the branch the change is proposed onto. Hive was not asking what that branch should be; it assumed `main`.
- **What went wrong.** On a repository whose default branch is not `main`, the PR got opened against a branch the work was never based on. On [projectbluefin/dakota](https://github.com/projectbluefin/dakota) (default: `testing`), a one-file change showed up as **233 changed files and ~260,000 changed lines** — the entire divergence between the two branches — large enough that GitHub's diff API refused to render it. Worse than the noise: a maintainer who did not notice the base would have merged the change onto the wrong branch.
- **The fix.** Ask the repository what its default branch actually is, in all three places that were assuming. A base the caller explicitly asked for is still honoured and never overridden; if the lookup fails, it falls back to `main` and logs a warning rather than refusing to open the PR.
- **Blast radius.** Changes the base branch of every PR hive opens from here on. Costs one extra GitHub API call per repository — cached, so not per PR — using a permission the App already holds. Already-open PRs are not retargeted.

## What this fixes

Every PR the hive opened was based on `main` whether or not that was the target repository's default branch. On [projectbluefin/dakota](https://github.com/projectbluefin/dakota), whose default is `testing`, that meant the PR's diff carried the entire divergence between the two branches — dakota#1411 showed 233 changed files and ~260k changed lines for a one-file change, large enough that GitHub's diff API refused to serve it — and a maintainer who missed the base would have merged the change onto the wrong branch.

The reporter's read was right: it is a hardcoded assumption, not an intermittent resolution failure.

## Why three changes and not one

The assumption lived in three places, and fixing only `CreatePR` would have accomplished nothing on the hosted path — `hive-open-pr` pinned `base: "main"` into the request file before `CreatePR` ever ran, so the wrong answer was already baked in.

| Where | Before | After |
| --- | --- | --- |
| `github.Client.CreatePR` | `base = "main"` when the caller passed none | resolves the repo's `default_branch` |
| `bin/hive-open-pr.sh` | `BASE="main"` — sent on every request, even when the agent said nothing | omitted `--base` stays empty, so the hive resolves it |
| `agent.SandboxExecutor` | cloned and PR'd against `defaultSandboxBaseRef = "main"` | reads the clone's own `refs/remotes/origin/HEAD` |

New `Client.DefaultBranch` reads the repository's own `default_branch` field — the REST API behind `gh repo view --json defaultBranchRef`, already covered by the App's metadata:read permission, exactly as the issue suggests. It is cached per repo, so this is one extra API call per repository rather than one per PR. Only *successful* lookups are cached: a transient 5xx must not pin a repo to the fallback for the life of the process.

The sandbox path needs the base *before* the PR — it branches from it — so it does not use the API at all. `git clone` already records the remote's HEAD in `refs/remotes/origin/HEAD`, so the answer is on disk: no API call, no extra token scope.

## Behavior preserved

- **An explicit base is never overridden.** Targeting a release line or stacking a PR still works, and issues no default-branch lookup.
- **Unresolvable repos degrade rather than fail.** If the metadata call errors or the clone has no `origin/HEAD`, the base falls back to `main` and logs a warning. A base that is probably right beats refusing to open the PR.

## Tests

Each of the three paths has a test that fails on the parent commit, verified by reverting each fix individually:

```
--- FAIL: TestCreatePR_EmptyBaseTargetsRepoDefaultBranch
    PR opened with base "main", want "testing" — the repository's default branch, not an assumed 'main'
--- FAIL: TestHiveOpenPRScript_OmittedBaseLeavesBaseUnset
    request pinned base "main"; an omitted --base must stay empty so the hive resolves the repository's default branch
--- FAIL: TestSandboxExecutorBranchesFromRepoDefaultBranch
    fetched "main", want testing — the clone's own default branch
```

Alongside those: explicit-base-is-honored (and issues no lookup), fallback-on-API-error, the failed lookup not being cached, and one-lookup-per-repo across three PR opens. The `hive-open-pr.sh` test runs the **real script** with its request dir redirected into a temp root and parses the result back into `PRRequest`, following the precedent in `gh_app_token_script_test.go`; it fails loudly if the script stops referencing the request-dir literal it patches, so it can't silently cover nothing.

```
ok  github.com/kubestellar/hive/pkg/github  44.239s
ok  github.com/kubestellar/hive/pkg/agent   255.137s
```

`go build ./...`, `go vet ./pkg/github/... ./pkg/agent/...`, and `gofmt` are clean on the touched files. Two tests in `pkg/agent` (`TestSendKick_RestartsCrashedCLIThenDelivers`, `TestRestartThenSendKick_CleanSlateThenDelivers`) failed intermittently in this environment; I confirmed they fail the same way on the unmodified parent commit and pass on a rerun, so they are unrelated to this change.

## Note for the reporter

@ahmedadan — this does not retarget the seven PRs already open on dakota; those still need the manual retarget you have been doing. It changes what the App files from here on. You offered to test a fix from your side, which would be the real confirmation: after this lands, a fresh PR on dakota should come in with `baseRefName: testing` without anyone touching it.

Fixes #4928

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,19 @@ Hive did not historically maintain a complete changelog. This file starts a prag
   documented `api_key: ${LINEAR_API_KEY}` form works from both `hive.yaml`
   and the dashboard.
 - Pull requests from forks can now satisfy `v4`'s required `gate` status context. `gate` is a job in `docker.yml`, which triggered on `push` only — and a fork PR's push event fires in the contributor's repo, not here, so `gate` never attached to the head SHA this repo evaluates and branch protection blocked the merge waiting for a check that could not arrive. Six fully green contributor PRs (#4930, #4932, #4934, #4935, #4936, #4937) were structurally unmergeable; `gate` was verified absent on every one of their head SHAs. It went unnoticed because `enforce_admins` is false, so maintainers' own merges silently bypassed the same missing check — the required context was in practice unenforced for maintainers and absolutely enforced for outside contributors. `docker.yml` now also triggers on `pull_request`, with every image-build job skipped for that event, so the context attaches to fork PRs at the cost of one ~5-second shell job and no image CI: the image is already built and smoke-tested on every PR by `v2-ci.yml` (`docker`, `build-and-test`, `overlayfs-exec-guard`). Push builds and GHCR publishing are unchanged — `gate` reports `push=false` for a PR event, so every `merge*` job stays skipped ([#4965](https://github.com/kubestellar/hive/issues/4965)).
+- PRs the hive opens are now based on the **target repository's default
+  branch** instead of a hardcoded `main`. Every PR the App filed on a repo
+  whose default branch is something else (e.g. `testing`) was opened against
+  `main`, so its diff carried the entire divergence between the two branches —
+  one reported PR showed 233 changed files and ~260k changed lines for a
+  one-file change, large enough that GitHub's diff API refused to serve it —
+  and merging it would have landed the change on the wrong branch. The base is
+  now resolved from the repo's own `default_branch` metadata (already covered
+  by the App's metadata:read permission, cached per repo), `hive-open-pr` no
+  longer pins `--base main` when the caller omits it, and sandbox kicks branch
+  from the default branch recorded in their own clone. An explicitly requested
+  base is still honored, and an unresolvable repo falls back to `main` rather
+  than failing the PR open. ([#4928](https://github.com/kubestellar/hive/issues/4928))
 
 - The canonical `blocked` workflow overlay now stays out of the contributor
   queue: the actionable-work enumerator, ReadyQueue, live contributor
--- a/bin/hive-open-pr.sh
+++ b/bin/hive-open-pr.sh
@@ -15,7 +15,13 @@
 # Usage (drop-in for the common gh pr create shape):
 #   hive-open-pr --repo <owner/repo> --head <branch> [--base <branch>] \
 #                --title "<title>" --body "<body>"
-#   # --head defaults to the current git branch; --base defaults to main.
+#   # --head defaults to the current git branch. --base is OPTIONAL and should
+#   # normally be omitted: an omitted base is left empty in the request so the
+#   # hive resolves the TARGET REPOSITORY's default branch when it opens the PR.
+#   # Do not pass --base main "to be safe" — that is exactly the bug that based
+#   # every PR on main in repos whose default branch is not main
+#   # (kubestellar/hive#4928). Pass --base only to target a non-default branch
+#   # deliberately (a release line, a stacked PR).
 #
 # On success it prints the request path and returns 0. The PR opens
 # asynchronously (within one watcher tick); poll the .result.json next to the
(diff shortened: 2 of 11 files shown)


Danathar:

clubanderson:
Superseded by #4986, which is now on `v4` and closed #4928. `src/pkg/github/defaultbranch.go` exists there with the cached `Client.DefaultBranch`, both PR-open paths resolving the real default branch, and tests using a repo whose default is `testing` rather than `main`.

**#4986 carried your approach forward** with one deliberate change, which I want to be explicit about: an unresolvable default branch now **fails the PR request** instead of falling back to `main` with a warning. The reasoning was that a silent fallback reproduces a milder version of the original bug — a wrong base nobody confirmed — and on `projectbluefin/dakota` that is what made seven PRs unreviewable. Your version was defensible; I chose the louder failure.

Closing because the work is on `v4` and this branch became unmergeable through unrelated churn, not because of any problem with it.

PR #4932, opened 2026-08-28
🐛 fix: defer agent dispatch behind weak PR claims instead of ignoring them

### LINKED ISSUE #4929: 🐛 bug: scanner re-implements issues already covered by its own open PR when the reference is non-closing
## What happened?

The scanner agent filed a second PR implementing an issue it already had an open PR against, on the same repo, two days apart. On [projectbluefin/dakota](https://github.com/projectbluefin/dakota), [#1402](https://github.com/projectbluefin/dakota/pull/1402) (Aug 24, body says `Refs #362`) and [#1411](https://github.com/projectbluefin/dakota/pull/1411) (Aug 26, body says `Fixes #362`) are two independent implementations of [dakota#362](https://github.com/projectbluefin/dakota/issues/362). Both carry the scanner footer and neither acknowledges the other. As a maintainer I now have two overlapping hold-gated PRs to reconcile for one issue.

Tracing it back: this looks like the same suppression gap as [#3980](https://github.com/kubestellar/hive/issues/3980), but on the scanner dispatch path, which the fixes for that issue do not appear to cover. `ParseClaimedIssues` in `src/pkg/github/prclaims.go` only matches closing keywords, so #1402's `Refs #362` registers no claim and #362 went back into the offer pool. #3985 taught the contribute path to honour non-closing references, but the scanner work list evidently still doesn't consult that.

The scanner path makes this strictly worse than #3980. There, the agent could see the duplicate and correctly declined, so the cost was wasted cycles. The scanner policy (`src/policies/scanner-holdgated.md`) forbids the agent from running `gh pr list` or `gh issue list` and makes the kick message its only source of truth, so a scanner agent handed an already-claimed issue has no way to discover the existing PR. It did the work again and filed the duplicate. The same policy is what instructs agents to use `Refs #N` for multi-phase work, so following the rules is what creates the invisible claim.

## Expected behavior

An issue with an open agent PR referencing it should not be re-dispatched to the scanner, regardless of whether the reference uses a closing keyword. Whatever suppression #3985 added for the contribute path should apply wherever the scanner's work list is built.

## Reproduction steps

1. Run the scanner (hold-gated mode) against a repo with an open issue.
2. Let it open a PR that references the issue with `Refs #N`, per its own policy for multi-phase work.
3. Leave the PR open and unmerged past the cooldown window.
4. The issue is offered to the scanner again, and since it cannot check for existing PRs, it re-implements and files a duplicate.

## Environment

- Branch/version: unknown, we consume the hosted App (`kubestellar-hive`, App ID 3568013). PR footers report `agent=scanner backend=copilot model=gpt-5.6-luna` and `ACMM L5 (hold-gated mode)`.
- Deployment method: hosted Hive Hub, GitHub App integration.
- Target repository: projectbluefin/dakota.

## Logs or screenshots

Both PR bodies, verbatim footer: `Filed by scanner agent (ACMM L5 — hold-gated mode). Hold-gated: human review required.` #1402 branch is `scanner/fix-persistent-kargs`, #1411 branch is `scanner/fix-persistent-kargs-362`, so the branch-name fallback in `prclaims.go` would only have caught the second one.

## Additional context

The duplicate compounds with #4928: #1411 was also opened against the wrong base branch, so the duplicate arrived as an unreviewable 233-file diff. A per-(repo, issue) dispatch cap like the backstop suggested in #3980 would have bounded this regardless of the claim-parsing gap. Happy to test a fix from our side, the App runs on our repo daily.

Danathar:
## Summary

- **What this touches.** Hive hands issues out to its agents and keeps a ledger of which issues already have an open PR, so two agents don't do the same job twice. A PR saying `Fixes #362` claims an issue *strongly*. One saying `Refs #362` — the wording hive's own policy tells agents to use for multi-phase work — makes a *weak* claim.
- **What went wrong.** Weak claims were recorded correctly and then deliberately ignored, so the issue went straight back into the offer pool while the PR was still open. The scanner agent cannot see open PRs — its policy forbids `gh pr list`, and the kick message is its only source of truth — so it did not recognise the duplicate. It re-implemented the issue from scratch. Result: two independent implementations of dakota#362, filed two days apart. Following the rules is exactly what made the claim invisible.
- **The fix.** A weak claim now **defers** the issue for 72 hours from when the PR was first seen claiming it, then releases it even if the PR is still open. Deliberately bounded rather than a block: a stranger's junk PR can stall hive's own pipeline by at most one window, and a partly-addressed issue still comes back for its remainder.
- **Blast radius.** Only the scanner's work list. The contribute queue reads the ledger directly and already honoured every claim — untouched. A dead PR (red and stale) releases its issue immediately, as before.
- **Known limit, by design.** Past 72h a duplicate is still possible. That is the deliberate never-freeze trade-off, not an oversight; the "Known bound" section below says what the real fix would be and why it is not in this PR.

## What this fixes

The scanner filed two independent implementations of one issue, two days apart, because the first PR wrote `Refs #362` rather than a closing keyword.

The claim machinery saw that PR correctly — `#3980`'s reference tier recorded a weak `Reference` claim for #362. `FilterClaimedIssues` then waved it straight through, so the issue went back into the offer pool while the PR was still open. The reporter's diagnosis was right about the effect; the mechanism is one layer further in than "the scanner work list doesn't consult the ledger". It consults it and then deliberately ignores this tier.

## Why the old rule was reasonable, and where it breaks

Weak claims were ignored on two rules that still hold:

- a stranger's junk PR must never freeze the hive's own pipeline (#3768), and
- a PR that never claimed to *close* an issue must not strand the remainder behind it (#3980).

Both assume someone downstream can **see** the claim and reconcile against it. The scanner cannot. Its hold-gated policy forbids `gh pr list`/`gh issue list` and makes the kick message its only source of truth, so a re-offered issue is not re-examined against the open PR — it is re-implemented. And the same policy is what instructs agents to write `Refs #N` for multi-phase work, so *following the rules* is what makes the claim invisible. That is why this is strictly worse than #3980, where the agent could see the duplicate and declined.

## The fix: defer, don't freeze and don't ignore

This implements the bounded middle ground the #3980 analysis already specified for this case:

> let weak claims (external authorship, or `Reference`) *defer* agent dispatch for a bounded window — e.g. 72h, matching `claimLedgerTTL` — anchored on a `FirstObservedAt` stamp that `insertLocked`/`Reconcile` carry forward when the same PR is re-observed […] The `redStale` valve would apply before the window, so a dead PR defers nothing.

A weak claim now suppresses for 72h from when the PR was **first** seen claiming the issue, then releases even while the PR stays open. Both original rules survive in bounded form: a stranger delays the hive's own pipeline by at most one window and must keep opening *new* PRs to extend it (which PR creation itself rate-limits), and a partially-addressed issue still comes back for its remainder.

| Claim | Before | After |
| --- | --- | --- |
| hive-authored, closing (`Fixes #N`) | suppresses | unchanged |
| hive-authored, reference (`Refs #N`) | **ignored** | defers 72h, then releases |
| external author | **ignored** | defers 72h, then releases |
| any weak claim, PR red + stale | ignored | releases immediately |

The contribute queue is untouched — it reads the ledger directly via `IssueClaimed` and already honoured every claim.

## The anchor is the load-bearing part

`Reconcile`'s authoritative path **replaces** the claim map on every enumeration cycle. A claim that took its anchor from the fresh scan would restart its own deadline every pass, and the "bounded" window would be a permanent freeze — the exact outcome #3980 refused. So `FirstObservedAt` is carried forward whenever the *same* PR is re-observed for the *same* issue; a different PR taking over the issue starts a fresh window, because that is genuinely new work. `ObservedAt` keeps refreshing independently, so TTL expiry is unaffected.

Ledgers written before the field existed normalize a zero anchor to `ObservedAt` on load, which costs such a claim at most one extra window and keeps the "always anchored once stored" invariant true for every reader.

## Tests

A dedicated `prclaims_weakdefer_test.go` covers the reported scenario end to end — through `FetchClaims` parsing a live `Refs #362` body, `Reconcile` anchoring it, and `FilterClaimedIssues` deciding — rather than hand-building a claim, since the bug was an interaction between those stages.

Alongside it: the anchor surviving **201 re-observations** across a window and still releasing on schedule; a different PR restarting the window; red+stale releasing before the window; pre-upgrade ledger normalization; and a strong claim unaffected by any of it.

Mutation-checked four ways, each caught by a distinct test:

| Mutation | Caught by |
| --- | --- |
| restore the old pass-through | `TestScannerNotReofferedIssueUnderOpenRefsPR` + 4 others |
| drop the `FirstObservedAt` carry-forward | `TestWeakClaimWindowDoesNotResetOnReobservation` |
| move the red+stale valve after the window | `TestWeakClaimRedStaleReleasesBeforeWindow` |
| make weak claims never release | `TestFilterClaimedIssuesDefersReferenceClaims` + 3 others |

Two existing tests asserted the old "weak claims never suppress" rule. They now assert the **bound** instead of the absence — a stronger statement, because the invariant they were protecting is precisely the release half, and pinning the bound tests it directly instead of observing it as a side effect.

```
ok  github.com/kubestellar/hive/pkg/github     43.5s
ok  github.com/kubestellar/hive/pkg/dashboard  48.4s
ok  github.com/kubestellar/hive/cmd/hive        8.4s
ok  github.com/kubestellar/hive/pkg/scheduler   0.02s
```

`go build ./...`, `go vet`, and `gofmt` are clean on the touched files.

## Known bound, stated plainly

Past 72h the issue releases even with the PR open, and a blind scanner could duplicate again — that is the deliberate design (never a freeze), not an oversight. For a hold-gated fleet where PRs routinely wait longer than three days on human review, the residual fix is to stop the agent being blind rather than to extend the window: surfacing the claiming PR in `${ISSUE_LIST}` would let a re-offered issue be continued instead of restarted. That is a change to agent-facing prompt text and a separate decision, so it is not in this PR. Happy to follow up if maintainers want it.

Two smaller things this does not touch, both already recorded on #3980: the multi-issue gap (a PR saying `Fixes #100` *and* `Refs #200` records no claim for #200, because the reference tier only runs when the closing tier yields nothing), and the per-(repo, issue) dispatch cap the reporter suggests as an independent backstop.

## Note for the reporter

@ahmedadan — with the App running on dakota daily, the check is that dakota#362 stops appearing in scanner kicks while #1402 is open. `deferring issue: open PR weakly claims it` is logged with the claiming PR number and the anchor timestamp on every cycle it holds one back, and `releasing issue: weak claim past its deferral window` when the 72h elapses, so both halves are visible without reading the ledger.

Fixes #4929


Danathar:
Closing as superseded — this work is already on `v4`.

It landed as [`88cca359`](https://github.com/kubestellar/hive/commit/88cca359) ("fix: weak PR references now defer, not skip, scanner dispatch (#4929) (#4984)"). I confirmed rather than assumed: rebasing this branch onto current `v4` leaves **0 commits and a 0-line diff**, so there is nothing left for it to change.

```console
$ git rebase upstream/v4      # only conflict was the CHANGELOG entry, now present upstream
$ git log --oneline upstream/v4..HEAD | wc -l
0
$ git diff upstream/v4 HEAD | wc -l
0
```

No action needed on #4929 — it is fixed. Closing this rather than force-pushing an empty branch, which would leave a PR showing no changes.

Issue #4971, opened 2026-08-28
🐛 /api/config/provenance reports Kubernetes ConfigMap layer identity on non-Kubernetes hives


Danathar:
## Summary

- **What this is about.** `GET /api/config/provenance` answers "which layer set this config field, and can I edit it?" — the API written so an operator never again has to patch four wrong layers to change one value. On a **non-Kubernetes hive it gives the wrong answer for every field**, and the wrong answer points at a layer that does not exist.
- **What it reports.** Every field comes from `configmap-seed`, at path `ConfigMap hive-config (key hive.yaml)`, writer `nobody (hub has read-only access; spoke RBAC omits configmaps)`, `writable: false`. On my hive there is no Kubernetes, no ConfigMap and no RBAC. The value actually lives in a plain file on disk that the spoke rewrites on every save.
- **Why it matters.** This is the exact failure the package was built to prevent, reproduced against a different audience. An operator on Docker/podman who follows this output goes hunting for a ConfigMap on a machine that has none, while the file that really decides the value is reported as unwritable and never named.
- **Blast radius.** Diagnostic only — no config value is computed wrong, and the layering itself is correct. Affects every hive not running under Kubernetes.

## Environment

`hive-wild-mole`, branch `v4` at `5246889e`, rootless podman. Not Kubernetes: no `KUBERNETES_SERVICE_HOST`, no service-account token, so `IsKubernetesPod()` is false.

## What the API returns

```json
{ "field": "acmm_level",
  "layer": "configmap-seed",
  "path":  "ConfigMap hive-config (key hive.yaml)",
  "writer": "nobody (hub has read-only access; spoke RBAC omits configmaps)",
  "writable": false,
  "value": "4" }
```

**21 of 21** fields report that layer. **0 of 21** report `writable: true`.

## Why it is wrong

`Layer.Path()` (`src/pkg/config/layers.go:127`) and `Layer.Writer()` (`layers.go:152`) are hardcoded `switch` statements over the layer constant, with no environment awareness. `LayerSeed` therefore always renders as a ConfigMap written by nobody, whatever the hive is actually running on.

Outside Kubernetes the seed is not a ConfigMap. It is `RuntimeConfigFile`, documented at `src/pkg/config/config.go:5141` as the boot-time source of truth in Docker/LXC mode, and `saveLocked` writes it on every save — so it is writable by the spoke, and `writable: false` is wrong for all 21 fields.

The response contradicts itself in one payload:

```json
[...]

PR #4974, opened 2026-08-28
🐛 fix(config): make the entrypoint's read-only config redirect reach the binary

### LINKED ISSUE #4973: 🐛 ACMM level reverts on restart even though the change persisted — non-Kubernetes hive, /data intact
## Summary

- **What happens.** Setting the ACMM level from the dashboard applies correctly and persists correctly — and then a container restart puts the hive back on its old level. Reproduced deliberately: L4 before the restart, L3 after, twice.
- **Why this is not the known #1856 case.** The write is not failing. `GET /api/config/provenance` reported `acmm_level: 4` *after* the dashboard change and *before* the restart, so the value reached the file that boot reads. It read `3` again immediately after the restart. Something changed the persisted value across a boot, rather than a save never landing.
- **Why it is not a wiped volume either.** `/data` survives. Agent kick history from 09:36–10:17 EDT was still present after a ~14:10 restart, so the PVC is intact and the entrypoint's first-boot re-seed path should not have run.
- **Blast radius.** Any non-Kubernetes hive whose operator changes ACMM level. The level silently reverts to whatever it was provisioned at, taking the agent roster with it — L4's `security` agent disappears on restart.
- **Not yet root-caused.** I could not read the container's filesystem or logs (it runs as a different uid). The exact entrypoint branch that ran is the missing piece; the command to get it is at the bottom.

## Environment

`hive-wild-mole`, branch `v4` at `5246889e`, rootless podman. Not Kubernetes — no `KUBERNETES_SERVICE_HOST` and no service-account token, so `IsKubernetesPod()` is false and the Docker-mode path in `entrypoint.sh:267` is the one in play.

## Timeline, measured

All values from `GET /api/config/provenance` (persisted, i.e. what a boot would load) and `GET /api/packs` (live).

| When | live level | persisted `acmm_level` |
|---|---|---|
| Initially | 3 | 3 |
| After setting L4 in the dashboard | **4** | **4** |
| After `restart` | **3** | **3** |

The middle row is the important one. The dashboard save works: it moved the persisted value from `3` to `4` in the file `provenance` reports as `seed_path` (`/data/hive.yaml.runtime`). The restart then moved it back.

Corroborating that `/data` persisted across the restart, from `GET /api/status` afterwards:

```
supervisor     lastKick=8/28 10:17 AM EDT
quality        lastKick=8/28  9:57 AM EDT
scanner        lastKick=8/28  9:36 AM EDT
```

Those predate the restart by hours, so the volume is intact.

## Why the expected path should have worked

In Docker mode the entrypoint treats the runtime config as a boot **input**, not a snapshot (`src/deploy/entrypoint.sh:267-290`):

> Unlike K8s there is no ConfigMap and no overlay here, so this file is a boot INPUT, not a snapshot: it is restored over the config path (or read directly when that path is read-only) on every boot, and it is the only thing that makes a dashboard save survive a container recreation.

Given a populated `/data/hive.yaml.runtime` holding `acmm_level: 4`, the branch at `entrypoint.sh:282` should copy it over the config path, or export `HIVE_CONFIG` to point at it directly if that path is read-only. Either way the Go binary should load `4`. It loaded `3`.

Three candidates I could not distinguish without logs:

1. `hive_runtime_config_read` resolved the **legacy** `hive.yaml.bak` rather than `hive.yaml.runtime`, and the legacy file still holds the provisioned level.
2. The first-boot branch at `entrypoint.sh:279` ran anyway and re-seeded the runtime file from a bind-mounted `hive.yaml` still holding `acmm_level: 3`.
3. Something rewrote the runtime file after boot from stale in-memory config — there is a periodic config saver, and `LoadWithDashboardOverlay`'s comments record a prior incident (#2439) where exactly that rewrote a layer and undid state on an interval.

## Contributing factor: a failed save is invisible

Not the cause here, since the save demonstrably worked — but worth fixing alongside, because it is what makes this class of bug hard to see. `src/pkg/dashboard/api_packs.go:535`:

```go
s.deps.Config.ACMMLevel = &level
// ...
if err := s.saveConfig(); err != nil {
    s.logger.Error("failed to save ACMM level to hive.yaml", "error", err)
}
```

The error is logged and execution continues: the in-memory level updates, `ApplyPackForce` reconciles the roster, and the handler returns **200**. The operator sees the new level and the new agents either way. Contrast the roster-reconciliation failure eleven lines below, which deliberately fails the request rather than report a false success. Persistence deserves the same, and `saveLocked` already computes whether the write will survive a restart (#3961) — that answer should reach the caller.

## What would identify the branch

The entrypoint announces which path it took. One of these lines will be in the container's boot log:

```bash
podman logs <hive-container> 2>&1 | grep -E "^\[entrypoint\].*(config|runtime)"
```

- `First boot — config seeded to PVC` → candidate 2
- `PVC runtime config restored to config path (from ...)` → the expected path ran, so candidate 3
- `Config path is read-only — using ... directly` → expected path, read-only variant
- `Migration — seeded ... from legacy` → candidate 1

I will add the output as a comment once I can read that container. Filing now because the reproduction is reliable and the data loss is real.

## Related

#4971 — `/api/config/provenance` reports Kubernetes ConfigMap layer identity on non-Kubernetes hives. Not the cause of this, but it is what I reached for to diagnose it, and its answer (`writable: false`, `writer: nobody`, a ConfigMap that does not exist here) sends an operator on this deployment in the wrong direction.

Danathar:
## Summary

If your hive's `hive.yaml` is mounted read-only — the normal thing under Docker/podman — then **every config change you made from the dashboard was being thrown away on restart**, and then permanently destroyed.

The chain is short:

1. The entrypoint notices it can't write `/etc/hive/hive.yaml`, so it decides to read your real config from `/data/hive.yaml.runtime` instead. It announces this by setting `HIVE_CONFIG`.
2. **The binary never saw that.** It only uses `HIVE_CONFIG` to pick a *default*, and the container image already passes `--config /etc/hive/hive.yaml` explicitly. An explicit flag beats a default, so the hive booted from the stale file the entrypoint had just rejected.
3. The next save of any kind then wrote that stale config *back over* `/data/hive.yaml.runtime` — so the good state wasn't just ignored, it was overwritten.

That's why the reporter saw an ACMM level go L4 → L3 on every restart with `/data` completely intact (#4973).

**The fix:** the entrypoint now also appends `--config "$HIVE_CONFIG"` to the command line it launches, so its decision actually reaches the process. If your config path *is* writable, nothing changes at all.

Fixes #4973.

---

## Root cause

`src/cmd/hive/main.go`:

```go
defaultConfig := "/etc/hive/hive.yaml"
if envCfg := os.Getenv("HIVE_CONFIG"); envCfg != "" {
    defaultConfig = envCfg
}
configPath := flag.String("config", defaultConfig, "path to hive.yaml config file")
```

`src/Dockerfile`:

```dockerfile
CMD ["--config", "/etc/hive/hive.yaml"]
```

`HIVE_CONFIG` only moves the **default**. The CMD supplies the flag **explicitly**, which outranks it. So both read-only escape hatches in `entrypoint.sh` — the Docker one at `:282` and the identical K8s one at `:111` — were writing a log line and changing nothing:

```sh
export HIVE_CONFIG="$HIVE_CONFIG_SOURCE"
echo "[entrypoint] Config path is read-only — using $HIVE_CONFIG_SOURCE directly"
```

Neither file is wrong by itself. It is the **pairing** that is broken, which is why it survived review of each.

### Why the loss became permanent

`Config.Save()` writes `RuntimeConfigFile` unconditionally (`config.go:5142`). So after booting on the stale config, the first save of any kind — a dashboard edit, a heartbeat-delivered config — serialised the stale in-memory config over `/data/hive.yaml.runtime`. That is what turns "wrong for one boot" into "your L4 is gone".

### Why it was invisible from outside

`/api/config/provenance` reads `os.Getenv("HIVE_CONFIG")` directly for `seed_path`. It therefore reported `/data/hive.yaml.runtime` — the file the *entrypoint* chose — while the process was running on `/etc/hive/hive.yaml`. This exactly matches the reporter's observation that the middle row of their table looked correct: the dashboard save really did land in the file provenance was reading, it just wasn't the file the hive would boot from. It also explains why the three candidate causes in the issue couldn't be told apart from the boot log — the log line that ran (`Config path is read-only — using ... directly`) was truthful about the *decision* and silent about the decision having no effect.

## Verification

Measured against the real `entrypoint.sh` and a stub binary with `main.go`'s exact flag-resolution logic, with a read-only config path holding `acmm_level: 3` and a runtime config holding `acmm_level: 4`:

| | entrypoint log | config the binary resolved | level loaded |
|---|---|---|---|
| before | `Config path is read-only — using .../hive.yaml.runtime directly` | `/etc/hive/hive.yaml` | **3** ❌ |
| after | same, plus `Config path pinned for the Go binary: --config .../hive.yaml.runtime` | `/data/hive.yaml.runtime` | **4** ✅ |

Also confirmed directly that Go's `flag` package keeps the **last** occurrence of a repeated flag, which is what makes appending sufficient:

```
HIVE_CONFIG=/data/hive.yaml.runtime  --config /etc/hive/hive.yaml                                 -> /etc/hive/hive.yaml
HIVE_CONFIG=/data/hive.yaml.runtime  --config /etc/hive/hive.yaml --config /data/hive.yaml.runtime -> /data/hive.yaml.runtime
```

Test suite: `pkg/config` and `cmd/hive` pass. `pkg/config` has one pre-existing failure on this machine, `TestGateFailsClosedWithoutIP6Tables`, which needs `NET_ADMIN`; it fails identically on a clean checkout of `v4`.

## Changes

- **`src/deploy/entrypoint.sh`** — the fix. Appends `--config "$HIVE_CONFIG"` to the launch argv when the variable is set. Appended rather than substituted so `flag.Parse`'s last-wins rule carries it, without this script having to parse and rewrite the CMD. When `HIVE_CONFIG` is unset — every hive whose config path is writable — nothing is appended and behaviour is byte-identical.
- **`src/cmd/hive/main.go`** — `WARN` when the loaded config path disagrees with `HIVE_CONFIG`. Should now be unreachable; kept because this exact discrepancy was undetectable from any external vantage point.
- **`src/pkg/config/entrypoint_boot_test.go`** — five tests. They extract the real launch block from `entrypoint.sh` by marker (the existing `runBootPrelude` idiom) rather than paraphrasing it, and fail loudly if the marker moves. Verified to fail without the entrypoint change.
- Docs (`config-layering.md`, `env-vars.md`, `operator-reference.md`) and a `CHANGELOG.md` entry under `Unreleased → Fixed`.

## Scope note

The issue also flags `api_packs.go:535`, where a failed `saveConfig()` is logged but the handler still returns `200`. That's a fair criticism and I'd support fixing it — but it needs a decision this PR shouldn't make silently: at that point the in-memory level and every agent's `Mode` have already been mutated, so returning `500` means either rolling those back or returning an error alongside completed side effects. It's a behavioural change to the API contract, independent of this root cause, and better reviewed on its own. Deliberately left out rather than forgotten.

Note also that this bug was *not* the cause of the misleading provenance output described in #4971 — that's a separate defect in `Layer.Path()`/`Writer()` and is still open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,29 @@ Hive did not historically maintain a complete changelog. This file starts a prag
   documented `api_key: ${LINEAR_API_KEY}` form works from both `hive.yaml`
   and the dashboard.
 - Pull requests from forks can now satisfy `v4`'s required `gate` status context. `gate` is a job in `docker.yml`, which triggered on `push` only — and a fork PR's push event fires in the contributor's repo, not here, so `gate` never attached to the head SHA this repo evaluates and branch protection blocked the merge waiting for a check that could not arrive. Six fully green contributor PRs (#4930, #4932, #4934, #4935, #4936, #4937) were structurally unmergeable; `gate` was verified absent on every one of their head SHAs. It went unnoticed because `enforce_admins` is false, so maintainers' own merges silently bypassed the same missing check — the required context was in practice unenforced for maintainers and absolutely enforced for outside contributors. `docker.yml` now also triggers on `pull_request`, with every image-build job skipped for that event, so the context attaches to fork PRs at the cost of one ~5-second shell job and no image CI: the image is already built and smoke-tested on every PR by `v2-ci.yml` (`docker`, `build-and-test`, `overlayfs-exec-guard`). Push builds and GHCR publishing are unchanged — `gate` reports `push=false` for a PR event, so every `merge*` job stays skipped ([#4965](https://github.com/kubestellar/hive/issues/4965)).
+- A dashboard-set **ACMM level (and every other saved config field) no longer
+  reverts on restart** on hives whose config path is read-only — the ordinary
+  shape of a bind-mounted `hive.yaml` under Docker/podman
+  ([#4973](https://github.com/kubestellar/hive/issues/4973)). When it cannot
+  write the config path, `entrypoint.sh` falls back to reading
+  `/data/hive.yaml.runtime` directly and signals that by exporting
+  `HIVE_CONFIG`. That export was inert: the binary reads `HIVE_CONFIG` only to
+  choose the *default* of `--config`, and the image's own
+  `CMD ["--config", "/etc/hive/hive.yaml"]` passes the flag explicitly, which
+  outranks any default. So the hive booted from the stale, unwritable file the
+  entrypoint had just decided not to use — and because `Config.Save()` writes
+  `/data/hive.yaml.runtime` unconditionally, the next save of any kind (a
+  dashboard edit, a heartbeat-delivered config) wrote that stale config back
+  over the operator's real persisted state, making the loss durable rather
+  than momentary. Reported as an L4 level returning to L3 after every restart,
+  with `/data` fully intact. The entrypoint now appends `--config
+  "$HIVE_CONFIG"` to the launch argv so its redirect actually reaches the
+  process (`flag.Parse` keeps the last occurrence), and the boot log warns at
+  `WARN` if the loaded config path ever disagrees with `HIVE_CONFIG` again —
+  the discrepancy was previously invisible, because
+  `/api/config/provenance` reads `HIVE_CONFIG` directly and so reported the
+  file the entrypoint chose while the process ran on the one it did not. Hives
+  whose config path is writable are unaffected: nothing is appended there.
 
 - The canonical `blocked` workflow overlay now stays out of the contributor
   queue: the actionable-work enumerator, ReadyQueue, live contributor
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -1139,6 +1139,36 @@ func main() {
 		}
 	}
 
+	// HIVE_CONFIG names a DIFFERENT file from the one we actually loaded.
+	//
+	// This is only ever the entrypoint's read-only escape hatch failing to
+	// land. When the config path cannot be written, entrypoint.sh exports
+	// HIVE_CONFIG=/data/hive.yaml.runtime and logs "config path is read-only —
+	// using ... directly"; HIVE_CONFIG is read above only as the DEFAULT of
+	// -config, and the image's CMD passes that flag explicitly
+	// (Dockerfile: CMD ["--config", "/etc/hive/hive.yaml"]), so the explicit
+	// value won and the redirect did nothing.
+	//
+	// That is #4973, and it is silently destructive rather than merely wrong:
+	// the stale file loads, then Config.Save() writes the whole in-memory
+	// config back over /data/hive.yaml.runtime, destroying the state the
+	// operator had persisted there. An ACMM level set from the dashboard came
+	// back at its provisioned value after a restart, twice, with /data intact.
+	//
+	// entrypoint.sh now appends `--config "$HIVE_CONFIG"` to the argv so the
+	// last-occurrence-wins rule in flag.Parse carries the redirect, which means
+	// this branch should be unreachable. It is kept — at WARN, naming both
+	// paths — because the failure it reports is invisible from every other
+	// vantage point: /api/config/provenance reads HIVE_CONFIG directly, so it
+	// reports the file the entrypoint chose while the process runs on the one
+	// it did not, and the two disagree with no way to tell from the outside.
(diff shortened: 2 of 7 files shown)


Danathar:

PR #4978, opened 2026-08-28
✨ feat: advisory digest target — post the digest to a Linear issue

gregoryhunt:
## Summary

- Adds `governor.advisory.target: github | linear` (default `github`) and `governor.advisory.linear_issue: ONB-123`. With `linear`, the advisory digest is maintained as **one comment on the designated Linear issue**, rewritten each cycle with the same body the GitHub pinned comment gets. The hive finds its own comment by a stable footer marker (`hive-advisory-digest`) and `commentUpdate`s it, `commentCreate`-ing only on the first post, and skips the write when the body is unchanged.
- **GitHub stays the default and is untouched**: with the key absent the eval cycle runs the exact same code as before (the operator running the Linear-sourced hive explicitly likes the living GitHub issue, so this is a choice, not an inference from the work source).
- Reuses the work source credential (`governor.work_source.linear.api_key`, bare `Authorization` header) and the existing Linear GraphQL transport: `LinearSource.doQuery` is split into a shared `linearGraphQL` helper that both the issue enumerator and the new `LinearAdvisoryPoster` (`src/pkg/worksource/linear_advisory.go`) call — no duplicated client.
- **Fails closed**: `linear` with an empty `linear_issue`, a missing key, an unknown issue, or an unknown `target` value logs an error naming the key and records the cycle as a failed post (so the hub's stale-advisory pill trips) — never a silent fallback to GitHub.
- Both keys are exposed on `GET/PUT /api/config/governor/advisory` (Governor Config → Advisory) so the choice is made without YAML; the PUT rejects `linear` without an issue and any target other than `github`/`linear`.
- `src/docs/advisory.md` gains a "Where the digest is posted" section; `CHANGELOG.md` has an `Unreleased` entry.

## Related issues

Fixes #

## Testing

- [x] `cd src && go build ./...` — clean
- [x] `cd src && go vet ./cmd/hive/... ./pkg/config/... ./pkg/worksource/... ./pkg/dashboard/...` — clean
- [x] `cd src && go test ./pkg/config/ ./pkg/worksource/ ./pkg/dashboard/ ./cmd/hive/`

```
ok  	github.com/kubestellar/hive/pkg/config	12.245s
ok  	github.com/kubestellar/hive/pkg/worksource	0.230s
ok  	github.com/kubestellar/hive/pkg/dashboard	28.398s
ok  	github.com/kubestellar/hive/cmd/hive	9.822s
```

New tests (each fails without the change):
[...]

PR #4975, opened 2026-08-28
🐛 fix: let a Linear-sourced hive enumerate work when GitHub issue listing is forbidden

gregoryhunt:
## Summary

- A hive with `governor.work_source.type: linear` died at `ghClient.EnumerateActionable` in `runEvalCycle` on every eval cycle whenever the GitHub App could not list issues (`403 Resource not accessible by integration` — the App lacked the Issues permission). The Linear overlay runs only after that call, so the Linear backlog was never enumerated and the queue sat at 0 until an Issues permission a Linear-sourced hive should not need was granted.
- Extracts the decision into `actionableAfterGitHubEnumerate(cfg, actionable, err, logger)`. Default (GitHub) work source: unchanged — log ERROR, abort the cycle, keep prior state. Non-GitHub work source: log a WARN naming the work source and continue with whatever partial result GitHub returned (an empty `ActionableResult` when nil), so the work-source overlay populates issues and GitHub PR maintenance still runs when PRs were obtainable.
- Adds a unit test that fails on the parent commit, a `CHANGELOG.md` `Unreleased / Fixed` entry, and a sentence in `docs/linear-agent.md` noting the GitHub Issues read permission is no longer required for a Linear-sourced hive.

## Related issues

Observed in production on a Linear-sourced hive (no tracking issue filed).

## Testing

- [x] `cd src && go build ./...`
- [x] `cd src && go vet ./cmd/hive/`
- [x] `cd src && gofmt -l ./cmd/hive` (clean)
- [x] `cd src && go test ./cmd/hive/ ./pkg/github/ ./pkg/worksource/...`
  ```
  ok  	github.com/kubestellar/hive/cmd/hive	9.109s
  ok  	github.com/kubestellar/hive/pkg/github	44.527s
  ok  	github.com/kubestellar/hive/pkg/worksource	0.527s
  ```
- [x] New test fails without the fix (`main.go` stashed to the parent commit):
  ```
  cmd/hive/worksource_enumerate_test.go:61:16: undefined: actionableAfterGitHubEnumerate
  FAIL	github.com/kubestellar/hive/cmd/hive [build failed]
  ```
  and passes with it:
  ```
  --- PASS: TestActionableAfterGitHubEnumerate_NonGitHubWorkSourceContinues (0.00s)
  --- PASS: TestActionableAfterGitHubEnumerate_GitHubWorkSourceAborts (0.00s)
  ok  	github.com/kubestellar/hive/cmd/hive	0.486s
  ```
[...]

PR #4979, opened 2026-08-28
✨ feat: add dashboard.public_url for OAuth redirect URIs on standalone hives

gregoryhunt:
## Summary

- Adds `dashboard.public_url` to `DashboardConfig`: the externally reachable origin of this dashboard, used to build the OAuth `redirect_uri` for the Linear agent install and the OpenRouter funding flow when it differs from the request host. Precedence is now `dashboard.public_url` → `hub.dashboard_url` (kept for hub-hosted spokes) → `X-Forwarded-Proto`/`X-Forwarded-Host`/`Host`, implemented once in `oauthPublicOrigin` (`src/pkg/dashboard/oauth_origin.go`) and shared by `linearAgentCallbackURL` and `openRouterCallbackURL`.
- Validates the value at config load: absolute `http(s)://` origin only — a path, query, fragment or credentials fail `Load` with an error that names `dashboard.public_url`; a trailing slash is trimmed so the appended callback path never doubles a `/`.
- `POST /api/linear/agent/install` now returns `redirect_uri` beside `authorize_url` so an operator can read the exact value the Linear app's Callback URL must match without decoding the authorize URL.
- Docs: `linear-agent.md` gains a setup step for the public origin (and explicitly steers hub-less hives away from `hub.dashboard_url`), `operator-reference.md` gets a `dashboard.public_url` row, and the standalone kustomize README notes when to set it. `CHANGELOG.md` entry under Unreleased.

**Why.** On a standalone (`hub.enabled: false`) hive behind Traefik + a Cloudflare Tunnel, the dashboard is private but `/linear/callback` is published on a different public hostname. The install request arrives on the private host and the callback on the public one, and an ingress that rewrites `Host` (Traefik fixed upstream `Host`, Cloudflare Tunnel "HTTP Host Header") makes `X-Forwarded-Host` differ between the two legs — so the `redirect_uri` sent with the code exchange did not match the one in the authorize URL and Linear rejected it with `redirect_uri is invalid`. The only override was `hub.dashboard_url`, a hub-namespaced field that is semantically wrong for a hive with no hub and discoverable only by reading source. `openRouterCallbackURL` had the same shape and gets the same treatment.

## Related issues

Observed on a standalone hive behind Traefik + Cloudflare Tunnel; no tracking issue filed.

## Testing

- [x] `cd src && go build ./...` — clean.
- [x] `cd src && go vet ./pkg/config/... ./pkg/dashboard/...` — clean.
- [x] `cd src && go test ./pkg/config/... ./pkg/dashboard/... -count=1`

  ```
  ok  	github.com/kubestellar/hive/pkg/config	13.403s
  ok  	github.com/kubestellar/hive/pkg/dashboard	28.463s
  ```

- [x] New tests: `TestValidateDashboardPublicURL` (13-case table: accepted origins, trailing-slash/whitespace trimming, and rejection of no scheme / wrong scheme / missing host / path / query / fragment / credentials), `TestLoad_DashboardPublicURL` (load-level normalization and error), `TestOAuthPublicOriginPrecedence` (public_url → hub.dashboard_url → forwarded headers → request host, plus both builders appending their path exactly once), and `TestLinearAgentInstallAndCallbackAgreeOnRedirectURI` — install arrives with `X-Forwarded-Host: hive.private.internal`, the callback with `X-Forwarded-Host: hive-public.example`, and the fake token endpoint records the `redirect_uri` it received, which must equal the one the install response reported.
- [x] The handler test fails on the parent commit. Run against `upstream/v4` with a parent-compatible copy of the test (no new symbols):

  ```
  --- FAIL: TestParentCheck_InstallAndCallbackAgree (0.01s)
      zz_parent_check_test.go:49: install response has no redirect_uri: {"authorize_url":"https://linear.app/oauth/authorize?actor=app&client_id=client-id&prompt=consent&redirect_uri=https%3A%2F%2Fhive.private.internal%2Flinear%2Fcallback&..."}
      zz_parent_check_test.go:58: authorize_url redirect_uri disagrees with exchange redirect_uri "https://hive-public.example/linear/callback": https://linear.app/oauth/authorize?...redirect_uri=https%3A%2F%2Fhive.private.internal%2Flinear%2Fcallback...
  FAIL
  FAIL	github.com/kubestellar/hive/pkg/dashboard	0.435s
  ```

- [ ] Other / not run (explain): full `go test ./...` was not run; only the two changed packages were exercised. Not verified against a live Linear workspace.

## Contributor checklist

- [x] PR targets `v4` (the active line for this change; `v2` is the historical default).
- [x] Title uses the repo emoji convention (`✨ feat: ...`).

https://claude.ai/code/session_019Ukvc9PBjCzVJQ1ZSX5GH7


clubanderson:
I rebased four of your five conflicted PRs onto current `v4` in place (#4975, #4978, #4980, #4981) — in each of those the only conflict was `CHANGELOG.md`, resolved by keeping both entries. `maintainerCanModify` was set, which is what made an in-place rebase possible rather than replacement PRs; your authorship and commit messages are untouched.

**I deliberately left this one alone**, because it has a real conflict rather than a mechanical one.

Besides the expected `CHANGELOG.md`, `src/docs/linear-agent.md` conflicts substantively: something already merged into `v4` inserted a **new step 3** covering `dashboard.public_url` and origin resolution for the Linear OAuth redirect, renumbering the steps after it. Your branch's parent had a *different* step 3 (the workspace-connect flow) with its own renumbering.

So both sides edited the same numbered setup sequence with different intent. Resolving that is a judgment about how your `dashboard.public_url` step should interleave with what landed — not something I should guess at, since picking wrong produces setup instructions that look right and are out of order.

Nothing was pushed to this branch and no force-push was attempted.

```sh
git fetch origin
git rebase origin/v4
# CHANGELOG.md: keep both entries
# src/docs/linear-agent.md: decide how your step interleaves with the new step 3
git rebase --continue
git push --force-with-lease
```

Worth checking first whether what landed on `v4` already covers your change — if `dashboard.public_url` is now documented upstream, part of this PR may be redundant.

PR #4980, opened 2026-08-28
🐛 fix: make the base k8s manifests deploy on plain Kubernetes (duplicate port 3002, missing FOWNER/FSETID, image-tag docs)

gregoryhunt:
## Summary

First real deploy of the standalone overlay (`src/deploy/kustomize/overlays/standalone`) on a non-OpenShift cluster (plain AKS) hit three defects in `src/deploy/k8s/` and the standalone docs. This PR fixes all three.

- **Duplicate port 3002.** `service.yaml` declared `dashboard` and `api` both on port 3002; the API server rejects that (`spec.ports[1]: Duplicate value: {... "Port":3002 ...}`), so the Service never existed. `deployment.yaml` had the same duplicate `containerPort` (warning only). Both now declare a single named port `dashboard` (the UI and `/api/*` are the same listener) and the startup/liveness/readiness probes target `dashboard`. Grepped the repo (overlays, `src/pkg/hub` provisioning templates, Route objects, docs): nothing else referenced the `api` port name — the OpenShift Route and hub templates already target `dashboard`.
- **Missing `CAP_FOWNER`/`CAP_FSETID` crash-looped the pod (exit 1, no error text) on every fresh RWO PVC.** The entrypoint's `/data/home` perm repair runs `chmod -R g+rwX` and `find … chmod g+s` over a tree it has already chowned to `dev`; chmod on an inode root does not own needs `CAP_FOWNER`, and keeping the setgid bit on a dir whose group root is not in needs `CAP_FSETID`. Neither was in `capabilities.add`, so the chmod failed and `set -e` exited right after `[entrypoint] Fixing /data/home perms...`. Fix: add `FOWNER` + `FSETID` to `deployment.yaml` (with a comment explaining why) and to the `hive-netadmin` SCC (the caps test requires a 1:1 mirror); make the perm-repair block `|| echo "[entrypoint] WARN: … is CAP_FOWNER in the pod's capabilities.add?"` so a future missing capability logs instead of silently exiting; add `fowner-chmod-non-owned-tree` and `fsetid-setgid-bit-survives-chmod` probe steps to `test_manifest_caps_runtime.sh` so both are exercised and the mutation run names each one (the job would otherwise fail them as UNEXERCISED).
- **Image-pinning docs pointed at tags that do not exist.** The standalone README, `kustomization.yaml` comments and `src/docs/manual-provisioning.md` said to pin `newTag: v4.2.0`-style tags; `ghcr.io/kubestellar/hive:v4.0.0` is an `ImagePullBackOff`. Read `.github/workflows/release.yml` (#4897): it **does** push `vX.Y.Z` image tags to ghcr (`docker buildx imagetools create -t ghcr.io/kubestellar/hive:v${VERSION}` from the short-SHA image, for `hive`, `hive-contributor`, `hive-hub`) — but only when an automated release is cut, and the workflow landed after the manually created `v4.0.0` git tag, so there is no `v4.0.0` image (verified against the registry: `manifests/v4.0.0` → 404; `stable`/`candidate`/`edge`/`v4-latest` → 200; the only semver image tag today is a `v5.0.0` left by a failed release run). Docs now: show the digest pin as the primary path (resolve `stable` to a digest with the ghcr token API, then `kustomize edit set image …@sha256:…`), explain which image tags exist and how to confirm a `vX.Y.Z` tag exists before using it, fix the remote-quickstart URL (missing `//` sub-path separator), and state that the base PVC is `ReadWriteOnce` — correct for the single replica with `strategy: Recreate`; RWX is not required.

Also updated `src/docs/migration-v2-v4.md` (the capability diff shown there) and the SCC/test-header comments that counted "six" capabilities. `CHANGELOG.md` entry added under `Unreleased` → `Fixed`.

## Related issues

Follows up on #4897 (tagged releases) for the image-tag documentation. No existing issue for the AKS deploy failures.

## Testing

- [x] `kubectl kustomize src/deploy/k8s` → renders (`# Warning: 'commonLabels' is deprecated` only, pre-existing); rendered output has zero `name: api` / `port: api` occurrences.
- [x] `kubectl kustomize src/deploy/kustomize/overlays/standalone` → renders; probes show `port: dashboard`, caps list ends `- FOWNER` / `- FSETID`.
- [x] `kubectl apply --dry-run=server -k src/deploy/k8s` against a real AKS cluster (`aks-onb-sbx-cus-001`, kubectl v1.36.1) — the cluster where the original failures were seen:
  ```
  namespace/hive unchanged (server dry run)
  serviceaccount/hive unchanged (server dry run)
  role.rbac.authorization.k8s.io/hive-dashboard-route-reader unchanged (server dry run)
  rolebinding.rbac.authorization.k8s.io/hive-dashboard-route-reader unchanged (server dry run)
  configmap/hive-config configured (server dry run)
  secret/hive-secrets configured (server dry run)
  service/hive unchanged (server dry run)
  deployment.apps/hive configured (server dry run)
  The PersistentVolumeClaim "hive-data" is invalid: spec: Forbidden: spec is immutable after creation ...
  ```
  The Service and Deployment are accepted by the API server (previously the Service was rejected with the duplicate-port error). The PVC line is expected: that cluster already has a bound `hive-data` PVC with a storage class set by the overlay patch, and the base manifest carries none — unrelated to this change.
- [x] `bash src/deploy/test_manifest_caps_runtime.sh` (run in a Linux container against the host Docker daemon, since the script's derivation needs GNU sed):
  ```
  PASS: capability add list derived from the manifest (8 capabilities), not restated in this test
  PASS: SCC allowedCapabilities matches deployment.yaml capabilities.add 1:1
  PASS: the whole privilege chain works under the manifest's exact capability set
     FOWNER removed -> failing steps: fowner-chmod-non-owned-tree
  PASS: dropping FOWNER breaks the privilege chain, and the failure names FOWNER
     FSETID removed -> failing steps: fsetid-setgid-bit-survives-chmod
  PASS: dropping FSETID breaks the privilege chain, and the failure names FSETID
  === Results: 21 passed, 0 failed ===
  ```
- [x] `bash src/scripts/check-suid-contract.sh src/Dockerfile src/deploy/entrypoint.sh` → `SUID/NET_ADMIN contract check passed.`
- [x] `bash src/deploy/test_entrypoint_symlinks.sh` → `Results: 8 passed, 0 failed`
- [x] `bash -n src/deploy/entrypoint.sh` → OK
- [x] `cd src && go test ./pkg/reach/ ./pkg/config/ ./pkg/hub/` (the packages whose tests read `deploy/k8s` / the entrypoint) → all `ok`
- [ ] `cd src && go build ./...` / full `go test ./...` — not run locally (no Go changes; the three affected packages above were run). CI covers the rest.

## Contributor checklist

- [x] PR targets `v4` (the maintainers' active branch for this work).
- [x] Title uses the repo emoji convention (`🐛 fix: ...`).

https://claude.ai/code/session_019Ukvc9PBjCzVJQ1ZSX5GH7


clubanderson:
Rebased onto current `v4` — the only conflict was `CHANGELOG.md` (both entries kept). No code changes; this is your work unmodified. `maintainerCanModify` was set, which is what made the in-place rebase possible rather than opening a replacement PR.

PR #4991, opened 2026-08-28
🔒 fix(contribute): warn that local mode is unconfined, and stop the sandbox toggle failing silently

Danathar:
## Summary

Two small changes that tell operators the truth about how confined their agents actually are.

1. **`just contribute-hive <backend> local` now warns that it is unconfined.** That mode runs the backend CLI as your own user, on your own machine, with permission prompts bypassed and nothing keeping it inside the workspace. The recipe previously described it only as "native mode". Container mode — already the default — is the confined one, and the warning points there.

2. **A sandbox toggle that sandboxes nothing is no longer silent.** `agent_sandbox.enabled` is only *half* the gate; each agent also needs `sandbox: {enabled: true}`. The dashboard's Security tab writes only the global half and is the only sandbox control the UI offers — so an owner could flip "agent sandbox" on, be told it was updated, and have every agent keep running unconfined. Hive now warns at boot **and on every config reload**, naming the agents still exposed.

Nothing about how agents run changes. This is the next safe increment after #4938; it is deliberately **not** the sandbox-by-default change, and there is a measured reason below.

Refs #4918.

---

## The finding that reframes the issue

#4918's root-cause section blames the double-gated Podman sandbox. **That sandbox does not exist on the path where the incident happened.**

`SandboxEnabled` is read only by `pkg/agent` — the hub-side pod agents. Nothing in `bin/contributor-relay.sh`, `bin/contributor-agent.sh` or the `Justfile` consults it:

```console
$ grep -rn "sandbox" bin/contributor-relay.sh bin/contributor-agent.sh | grep -v codex
$ grep -n "agent_sandbox" Justfile
$    # (both empty)
```

And the reporter's own evidence points at the contributor path: tmux session `hive-claude-bc59` in a `tmux-spawn-*` scope under the user slice is exactly what the Justfile's local-mode branch creates (`TMUX_SESSION="hive-${BACKEND}-$(head -c 2 /dev/urandom ...)"`).

**So enabling `agent_sandbox` would not have prevented the reported incident.** The lever on that path is container mode vs local mode — and container mode is already the default (`contribute-hive backend="" mode="docker"`). What was missing was anything saying so.

`sandbox-isolation.md` now carries a per-path table, and drops its "operators should enable the sandbox below" line, which was wrong advice for the contributor path.

## Why the gate is not collapsed

Making `agent_sandbox.enabled: true` sufficient on its own looks like the obvious fix. It is not safe, and I want the reason on the record rather than in a commit message:

A sandboxed agent runs a **different execution model** — no tmux CLI at all; every kick is a podman run against the primary repo. And `startSandboxKickLocked` (`src/pkg/agent/manager.go`) has **no fallback to tmux**:

```go
if strings.TrimSpace(m.sandboxConfig.Image) == "" && strings.TrimSpace(agent.Config.SandboxImage(m.sandboxConfig)) == "" {
    return fmt.Errorf("agent %s sandbox image is not configured", agent.Name)
}
repo := m.project.PrimaryRepo()
if repo == "" {
    return fmt.Errorf("agent %s sandbox execution requires a primary repo", agent.Name)
}
```

On any hive that set the global flag without a configured image, collapsing the gate would convert working agents into **permanently failing** ones. That is a fleet-affecting posture change that deserves measurement and a maintainer decision, not a code-reading — so it stays open. The warning is the part that is safe today, and a second warning names the no-image case for the same reason.

Worth noting for whoever picks up option 2: the strong form in #4918 ("require the podman path when the assigned repo is not the primary repo") is not expressible against the current executor, which clones `m.project.PrimaryRepo()` by construction.

## What the warnings look like

```
⚠️  LOCAL MODE — the agent is NOT confined to a workspace.

    The backend CLI runs as dbaggett on this machine, with permission
    prompts bypassed. It can read and write anything your user can,
    including files outside /home/op/workspace.
    Assigned repos are third-party code and their test suites run for real.

    Still constrained: privilege-escalation and host boot/deployment
    commands are denied (sudo, pkexec, rpm-ostree, bootc, grubby, ...),
    and no agent receives a GitHub token or pushes directly.
    NOT constrained: everything else your user can reach.

    For a confined agent, drop 'local' and use container mode:
      just contribute-hive claude
```

Naming what *still* holds is deliberate. A warning implying nothing is constrained is its own kind of wrong, and gets ignored.

```
WARN agent sandbox posture: agent_sandbox.enabled is true but NO agent is opted in, so the
  sandbox is inert and all 3 agent(s) still run unconfined on the tmux path — the per-agent
  gate is separate: set `sandbox: {enabled: true}` on each agent that should be sandboxed (#4918)

WARN agent sandbox posture: agent_sandbox.enabled is true but only 1 of 2 agent(s) are opted
  in (scanner); the rest still run unconfined on the tmux path (#4918)

WARN agent sandbox posture: agent(s) scanner are sandbox-opted-in but resolve no sandbox
  image; sandboxed kicks have no tmux fallback and will fail outright — set
  agent_sandbox.image (or the per-agent sandbox.image)
```

The reload hook matters more than the boot one: flipping the Security tab toggle writes the config and never restarts the process, so a boot-only check would never reach the person who most needs it.

## Tests

Pinned in both directions, because a diagnostic that cries wolf is worse than none:

- sandbox off globally is **not** a warning (that is the documented default; warning on every hive would train operators to ignore the line that matters)
- a fully opted-in hive is **silent**
- the inert and partial cases name the agents still unconfined, and the message must carry the key that fixes it
- an explicit per-agent `false` still counts as unconfined — it is a deliberate opt-out, but the agent is still exposed and the count an operator sees must say so
- the Justfile warning is read from the file rather than restated, including an assertion that **container mode is still the default**, since the advice goes stale the moment that changes

`pkg/config`, `pkg/dashboard` and `cmd/hive` pass. `pkg/config` has one pre-existing failure on this machine, `TestGateFailsClosedWithoutIP6Tables`, which needs `NET_ADMIN`; it fails identically on a clean checkout of `v4`.

## Still open on #4918

Option 1 (a real filesystem boundary on the default path, beyond #4938's command deny-list) and option 2 (changing the sandbox default) are both untouched here. The issue should stay open for them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,32 @@ Hive did not historically maintain a complete changelog. This file starts a prag
   the sandbox kick) outright, with the reason surfaced in the retry/quarantine
   log, because a wrong base is worse than a delayed PR
   ([#4928](https://github.com/kubestellar/hive/issues/4928)).
+- **`just contribute-hive <backend> local` now says that it runs the agent
+  unconfined on your machine**, and the docs no longer imply a workspace
+  boundary that only the opt-in Podman path provides
+  ([#4918](https://github.com/kubestellar/hive/issues/4918)). Container mode is
+  the default and is the confined one; local mode runs the backend CLI as your
+  own user with permission gating bypassed and nothing scoping its filesystem
+  access — which is how an agent running an assigned third-party repo's test
+  suite reached a contributor's bootloader. The launch banner now names what is
+  still constrained (the #4938 host-state denials; credentials and pushes on
+  every path) and what is not, and points at container mode as the remedy.
+  `sandbox-isolation.md` gains a per-path table recording the finding that
+  matters most here: **the `agent_sandbox` Podman path is hub-side only and
+  does not exist on the contributor path at all**, so enabling it would not
+  have prevented the reported incident.
+- **A globally-enabled agent sandbox that sandboxes nothing is no longer
+  silent.** `agent_sandbox.enabled` is only half the gate — each agent also
+  needs `sandbox: {enabled: true}` — and the dashboard's Security tab writes
+  only the global half, so an owner could turn the sandbox on, be told the
+  setting was updated, and have every agent keep running unconfined. Hive now
+  logs an `agent sandbox posture` warning at boot **and on every config
+  reload** (the moment the toggle is flipped) naming which agents are still
+  unconfined, and separately warns when an opted-in agent resolves no sandbox
+  image — sandboxed kicks have no tmux fallback and fail outright. Behaviour is
+  unchanged: the second gate is load-bearing and collapsing it would convert
+  working agents into permanently failing ones, so that remains a separate,
+  measurement-first decision.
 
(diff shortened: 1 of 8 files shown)


Danathar:

clubanderson:
Rebased onto current `v4` (in-place, force-pushed to this branch — same PR, same authorship).

Only conflict was `CHANGELOG.md`, resolved by keeping both entries (this PR's and the ones that landed on `v4` in the interim, including #4986/#4996). No code files conflicted; `src/cmd/hive/main.go` merged cleanly. `gofmt -l` is clean on the touched Go files and no conflict markers remain.

PR #4993, opened 2026-08-28
🐛 fix(dashboard): stop offering amazonq, a backend with no dispatch path

### LINKED ISSUE #4988: Dashboard offers `amazonq` as a CLI backend, but it has no dispatch path — accept-then-fail
Found by the backend-list parity guard in #4987 on its first run.

## The problem

`amazonq` is **offered to operators in the dashboard** but cannot be launched.

**Where it is advertised:**
- `src/pkg/dashboard/static/index.html:7525` — in the JS `KNOWN_BACKENDS` array
- `src/pkg/dashboard/static/index.html:18131` — in the **CLI Pin Value** tooltip, listed by name as a selectable option

**Where it is absent:**
- `config/backends.conf` — not in `KNOWN_BACKENDS`, `backend_binary`, or `backend_perm_flag`
- `src/pkg/config/config.go` — not in `CLIBackends` or `InferenceBackends`

So `validateBackendName` (`src/pkg/agent/manager.go:948`) rejects it. An operator who picks a backend the UI recommends gets a launch failure.

That is precisely the **accept-then-fail** class `validateBackendName` exists to prevent — its own comment says it dispatches on the same canonical lists as `ValidateBackend` and `backendBinary` *"so a backend accepted by any write path is one the launch path can start."* The dashboard is a write path that was never brought into that agreement.

## It is half-wired elsewhere too

`amazonq` also appears in three shell/JS "no `--model` flag" lists — `bin/agent-launch.sh:149`, `bin/contributor-agent.sh`, `bin/contributor-relay.sh` — plus `bin/contributor-relay.test.js`. Five files reference a backend with no dispatch path.

## Two possible resolutions

1. **Remove it.** If `amazonq` was never finished, drop it from the dashboard arrays and tooltip and from the three `--model` lists. Cheapest, and stops advertising something that cannot work.
2. **Finish it.** Add it to `backends.conf` (`KNOWN_BACKENDS`, `backend_binary`, `backend_perm_flag`) and to `CLIBackends`. Only worth doing if someone actually wants Amazon Q support — the `--model` handling already anticipates it.

**Do not simply add it to the parity test's exception list.** The exception mechanism in #4987 is for names that legitimately live in one list only (`litellm` is an inference gateway; `gemini` routes by model prefix). `amazonq` is not that — it is in neither authoritative registry while being offered in the UI.

## Related gap

The dashboard's JS `KNOWN_BACKENDS` array is a **third, independent list** with no guard on it at all. It also carries `openrouter`, a gateway example name. #4987's Go test cannot practically parse embedded JS, so that list stays unguarded — worth a follow-up if this class of drift matters.

Related: #4987 (parity guard), #4970 (adding a new backend)

Danathar:
## Summary

The dashboard offered `amazonq` in its backend picker, and named it in the CLI Pin Value tooltip, but nothing could launch it — it is in neither authoritative registry, so `validateBackendName` rejects it. An operator who picked the backend the UI recommended got a launch failure.

**This PR removes it**, and adds a guard so the list it was hiding in cannot drift again.

Removing rather than finishing is **not a product judgement**. `amazonq` once had a full dispatch path (`backend_binary` → `q`, `backend_perm_flag` → `--trust-all-tools`) and was deliberately dropped from `backends.conf` in #1045, whose commit message states the resulting set outright: *"6 CLIs: Claude, Copilot, Goose, Codex, Agy, Bob"*. `gemini` went in the same commit. The UI was simply never brought along. This finishes a removal that was already decided; it does not decide against Amazon Q support.

Closes #4988.

---

## What was removed

The issue names five files. There are seven places, across both dashboard trees:

| File | What it was |
|---|---|
| `src/pkg/dashboard/static/index.html` | the JS `KNOWN_BACKENDS` array that `backendOptionList()` builds the dropdowns from |
| `src/pkg/dashboard/static/index.html` | the CLI Pin Value tooltip, naming the options in prose |
| `dashboard/index.html` | the standalone tree's copy of the same array |
| `dashboard/server.js` | the standalone tree's server-side copy |
| `bin/agent-launch.sh` | dead "no `--model`" list |
| `bin/contributor-agent.sh` | dead "no `--model`" list |
| `bin/contributor-relay.sh` | dead `NO_MODEL_FLAG_BACKENDS` entry |
| `bin/contributor-relay.test.js` | the test pinning the dead entry |

The three `--model` lists were already unreachable — the backend can never be selected — but they are what made the name read as *half-wired* rather than *removed*.

## The guard, which is the part that stops recurrence

#4988 notes the dashboard's JS array is a **third, independent list with no guard at all**, and calls a follow-up "worth it if this class of drift matters." It plainly does — that gap *is* this issue — so the guard lands here rather than being deferred:

- **`TestBackendPickerOffersOnlyDispatchableBackends`** parses the JS array out of the embedded page and asserts every name is in `CLIBackends`, `InferenceBackends`, or a documented exception.
- **`TestBackendPickerInferenceListMatchesGo`** pins the page's second array (`INFERENCE_BACKENDS`). A name drifting between the two mislabels a picker entry as a CLI rather than a gateway even when both lists look individually plausible.
- **`TestCLIPinTooltipNamesNoUndispatchableBackend`** covers the prose tooltip, which the array guard structurally cannot reach — and which is the *other* place #4988 found `amazonq`.

### The direction is deliberate

This asserts **picker ⊆ registries**, not the reverse.

A name the picker offers that nothing can launch is a broken promise to the operator — the reported bug. A supported backend the picker *omits* (`pi`, `agy`) is only under-advertisement. And whether `agy` belongs in the **hub's** picker turns on whether it can authenticate in a pod — its sign-in is an interactive Google OAuth flow with no API-key mode — which is a question this guard must not answer by implication. Left for maintainers.

### `openrouter` stays, as a declared exception

Per the issue's own framing, it is a Model Gateway preset: agent routing resolves any configured gateway name, and the page deliberately offers it *before* a matching gateway exists so an operator can pick it and then fund it (there is a comment above `KNOWN_BACKENDS` saying exactly that). The exception carries its reason and the same staleness check as `cliBackendExceptions` — an exception that becomes a real registered backend, or leaves the picker, fails the test rather than rotting.

`amazonq` is exactly the kind of name that must **not** go on that list, and the issue says so; the test comment records it.

## Verification

Mutation-checked — each mutant fails the right test with a message naming the specific backend and list:

| Mutant | Result |
|---|---|
| restore `amazonq` to the picker array | ✅ `the dashboard backend picker offers amazonq but it is in neither config.CLIBackends nor config.InferenceBackends...` |
| restore `amazonq` to the tooltip only | ✅ `the CLI Pin Value tooltip offers "amazonq", which is in no authoritative backend registry` |
| drop `watsonx` from the JS inference list | ✅ `Go InferenceBackends has watsonx but JS INFERENCE_BACKENDS does not, so the picker renders it as a CLI backend` |

Shell/JS edits confirmed inert: `bash -n` clean on both scripts, and the relay suite is **162/163 before and after**. The single failure (`a relaunch cds somewhere resolvable before starting the CLI`) is pre-existing on a clean `v4` and unrelated. `pkg/config` and `pkg/dashboard` pass in full.

## Note for whoever reads this next

`dashboard/server.js` does not parse (`node --check` → `SyntaxError: Unexpected token '-'` at an unquoted `ci-maintainer:` object key, line 983). That is pre-existing on `v4`, unrelated to this change, and I have not touched it — flagging it because it means that file's list is unguarded by anything executable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,27 @@ Hive did not historically maintain a complete changelog. This file starts a prag
 
 ### Fixed
 
-- Removed the half-wired `amazonq` (Amazon Q Developer) backend from the operator-facing dashboard: the CLI backend picker's `KNOWN_BACKENDS` array and the "CLI Pin Value" tooltip both listed it as a selectable option, but it was never in either authoritative registry (`config/backends.conf`'s `KNOWN_BACKENDS`/`backend_binary`/`backend_perm_flag`, or `src/pkg/config/config.go`'s `CLIBackends`), so `validateBackendName` rejected it on launch — an accept-then-fail bug where the dashboard recommended a backend that could not start. Also dropped from the three shell/JS "no `--model` flag" lists in `bin/agent-launch.sh`, `bin/contributor-agent.sh`, and `bin/contributor-relay.sh` (the `goose`/`bob` handling in each is unaffected) and from `bin/contributor-relay.test.js`. Nobody had requested Amazon Q support; removal was cheaper than finishing the integration. Found by the backend-list parity guard added in [#4987](https://github.com/kubestellar/hive/pull/4987). The dashboard's JS `KNOWN_BACKENDS` array is still unguarded by that parity test — it cannot practically parse embedded JS — so this exact class of drift can recur there. ([#4988](https://github.com/kubestellar/hive/issues/4988))
+- **The dashboard no longer offers `amazonq` as a backend it cannot launch**
+  ([#4988](https://github.com/kubestellar/hive/issues/4988)). It was listed in
+  the backend/method picker and named in the CLI Pin Value tooltip while being
+  absent from both authoritative registries (`config/backends.conf`'s
+  `KNOWN_BACKENDS` and `config.CLIBackends`), so `validateBackendName` rejected
+  it and an operator who picked the backend the UI recommended got a launch
+  failure — the accept-then-fail class that function exists to prevent. Amazon
+  Q once had a complete dispatch path (`backend_binary` → `q`,
+  `backend_perm_flag` → `--trust-all-tools`) and was deliberately removed from
+  `backends.conf` in #1045 (commit `9d397f55`), whose message states the
+  resulting set outright: "6 CLIs: Claude, Copilot, Goose, Codex, Agy, Bob" —
+  the UI was simply never brought along. Removed from both dashboard trees and
+  from the three dead `--model`-exclusion lists that still named it
+  (`bin/agent-launch.sh`, `bin/contributor-agent.sh`,
+  `bin/contributor-relay.sh`, `bin/contributor-relay.test.js`). Found by the
+  backend-list parity guard added in
+  [#4987](https://github.com/kubestellar/hive/pull/4987), which could not
+  reach the dashboard's embedded JS; a new guard
+  (`TestBackendPickerOffersOnlyDispatchableBackends` and friends) now parses
+  that third, previously-unguarded `KNOWN_BACKENDS` array and the CLI Pin
+  Value tooltip directly, so this exact class of drift cannot recur there.
 
 - An issue covered by an open PR that only *references* it (`Refs #N`, `Part of
   #N`) is no longer immediately re-offered to agents. Such weak claims — along
--- a/bin/agent-launch.sh
+++ b/bin/agent-launch.sh
@@ -144,7 +144,7 @@ else
   PERM_FLAG=$(backend_perm_flag "$BACKEND")
   MODEL_FLAG="--model"
 
-  # goose doesn't support --model
+  # goose doesn't support --model (it takes its model from config/env)
   case "$BACKEND" in
     goose) MODEL_FLAG="" ;;
   esac
--- a/dashboard/index.html
+++ b/dashboard/index.html
@@ -2817,7 +2817,7 @@
     }
 
     // ── Centralized backend/model config (mirrors backends.conf) ──────────
-    const KNOWN_BACKENDS = ['claude', 'copilot', 'bob', 'gemini', 'codex', 'amazonq', 'goose', 'aider'];
+    const KNOWN_BACKENDS = ['claude', 'copilot', 'bob', 'gemini', 'codex', 'goose', 'aider'];
     const FREE_BACKENDS = ['copilot', 'goose'];
     const KNOWN_MODELS = [
       { value: 'claude-opus-4-6', label: 'Opus 4.6' },
--- a/dashboard/server.js
+++ b/dashboard/server.js
@@ -82,7 +82,7 @@ function persistEnabledAgents() {
 }
 
 // ── Centralized backend/model config (JS equivalent of backends.conf) ──────
-const KNOWN_BACKENDS = ['claude', 'copilot', 'gemini', 'codex', 'amazonq', 'goose', 'aider'];
+const KNOWN_BACKENDS = ['claude', 'copilot', 'gemini', 'codex', 'goose', 'aider'];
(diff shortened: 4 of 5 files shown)


clubanderson:
Thanks — and apologies for the collision. The removal itself landed in #4992 (`v4`) about twenty minutes before this opened, so that half is now duplicated. That is on the dispatch side, not yours.

**Two parts of this PR are not duplicated and are the more valuable half:**

1. **`src/pkg/dashboard/backend_picker_parity_test.go`** — a guard on the dashboard's JS `KNOWN_BACKENDS` array. That is precisely the gap #4987's Go-side parity test could not close (it cannot parse embedded JS), and which I flagged in #4988 as a place this drift can recur. #4992 removed the symptom and left that hole open; this closes it.
2. **`dashboard/index.html` / `dashboard/server.js`** — the legacy top-level tree. #4992 deliberately left those alone as documented-dead code rather than widen scope. If the parity guard covers them, cleaning them up alongside is right.

**Your commit message also corrected me, and I want that on the record.** I framed #4988 as a half-wired backend someone started and abandoned. You showed it had a *complete* dispatch path (`backend_binary` → `q`, `backend_perm_flag` → `--trust-all-tools`) that was deliberately removed in #1045, whose message states the resulting set as *"6 CLIs: Claude, Copilot, Goose, Codex, Agy, Bob"*. I verified that against `9d397f55`. So this finishes a removal that was already decided rather than deciding against Amazon Q — a materially different and more accurate framing than mine.

**Suggested path:** rebase onto current `v4` and drop the parts #4992 already landed, leaving the parity test and the legacy-tree cleanup. That is a clean, non-duplicated PR I would merge as-is.

Happy to do that rebase myself if you would rather — just say so and I will not force-push your branch without an ack.

clubanderson:
Rebased onto current `v4` (in-place, force-pushed — same PR, same authorship) and trimmed to drop what #4992 already merged.

**Dropped as already-merged by #4992:** the `amazonq` removals from `bin/agent-launch.sh`, `bin/contributor-agent.sh`, `bin/contributor-relay.sh`, `bin/contributor-relay.test.js`, and `src/pkg/dashboard/static/index.html` all collapsed to no-ops during the rebase — the same edits were already applied upstream, so they no longer appear in this PR's diff. Kept one small surviving hunk in `bin/agent-launch.sh` (a more descriptive comment on the `goose`/`--model` case) since it wasn't part of #4992's change.

**Kept, because #4992 didn't cover it:**
1. `src/pkg/dashboard/backend_picker_parity_test.go` — the new guard that parses the dashboard's embedded JS `KNOWN_BACKENDS`/`INFERENCE_BACKENDS` arrays and the CLI Pin Value tooltip, asserting every offered name is dispatchable. This is the third backend list #4987's Go-side parity test structurally cannot reach (it can't parse embedded JS), and closes the gap #4988 asked for as a follow-up.
2. `dashboard/index.html` and `dashboard/server.js` — the legacy top-level tree's own `KNOWN_BACKENDS` copies, which #4992 deliberately left untouched.

**CHANGELOG:** merged the two overlapping entries into one, keeping the more accurate framing from this PR — that `amazonq` once had a complete dispatch path and was deliberately dropped in #1045 (commit `9d397f55`, "6 CLIs: Claude, Copilot, Goose, Codex, Agy, Bob") rather than never having existed — plus a note that the new parity guard is what closes this class of drift for good.

Verified: no conflict markers remain, `gofmt -l` clean on the new test file, `bash -n`/`node --check` clean on the touched shell/JS files (the pre-existing `dashboard/server.js` syntax error your PR description already flagged is untouched and unrelated), and the PR is not empty — the parity test and legacy-tree cleanup both survived the trim.

PR #4981, opened 2026-08-28
✨ feat: let ACMM gap issues be filed on the work source (Linear) instead of GitHub

gregoryhunt:
## Summary

- The dashboard's ACMM "Open Issue" / "Open All" buttons (`POST /api/acmm/issue`) always created a **GitHub** issue, even on a hive whose backlog lives in Linear (`governor.work_source.type: linear`). This PR lets the operator choose per hive and per click.
- **Config:** new `governor.acmm.issue_tracker: github | work_source` (`ACMMConfig` in `pkg/config`). Default `github` — unset hives see zero behavior change; any other value fails `validate()`. `work_source` means "file where the backlog lives": Linear work source → Linear `issueCreate` on the team whose `work_source.linear.teams[].repo` matches the criterion's repo (else the first team); GitHub / GitHub Projects / Jira / unset → GitHub as today.
- **API:** `POST /api/acmm/issue` accepts an optional `tracker` field (`"github"` | `"work_source"`) overriding the config for that request; unknown values → `400`. The response now carries `tracker` (`"github"` | `"linear"`) and, for Linear, `identifier` (`ENG-42`) and `team`. `GET /api/acmm/evaluation` reports the effective default as `issue_tracker` plus `work_source_type`, stamped from live config on every response (including cache hits) so a config edit shows up immediately.
- **Linear:** `LinearSource` gains `CreateIssue` (team-key → id lookup + minimal `issueCreate` mutation) and `TeamForRepo`, reusing the existing endpoint/auth path (bare `Authorization: <api_key>`) via a new `doGraphQL` seam that `doQuery` now delegates to. Title and body are rendered once (`acmmIssueContent`) and are byte-identical on both trackers; the audit entry records `tracker` either way.
- **Dashboard UI** (the modal *is* in this repo, `pkg/dashboard/static/index.html`): when the work source is Linear, the level dialog shows a small **File in: GitHub / Linear** selector pre-selected from config; all four `/api/acmm/issue` call sites (single, repo picker, both Open-All loops) send the choice, and issue links render the Linear identifier instead of `#N`.
- Docs: new "Where ACMM gap issues are filed" section in `docs/acmm-policy-matrix.md`, cross-links from `docs/work-sources.md` and `docs/api-reference.md`; `CHANGELOG.md` entry under Unreleased.

## Related issues

Fixes # (none filed — operator request: Linear-sourced hives were getting unwanted GitHub issues from the ACMM prerequisites modal)

## Testing

[...]

PR #4998, opened 2026-08-28
🐛 fix(hub): scope the undeliverable-upgrade memory to its server and stop it leaking

### LINKED ISSUE #4995: Package-global undeliverableUpgradeNoted is shared across hub instances and leaks entries for removed hives
## Summary

`undeliverableUpgradeNoted` — the de-duplication memory that keeps the hub from writing an identical "upgrade not armed" timeline entry on every 2-minute poll — is a package-level global rather than a field on `HubServer`. Two hub instances in one process (or parallel tests) share and clobber it, and entries are never removed for hives that no longer exist.

## Mechanism

`src/pkg/hub/pullonly_upgrade.go:158-161`:

```go
var (
    undeliverableUpgradeMu    sync.Mutex
    undeliverableUpgradeNoted = map[string]string{}
)
```

Consumed by `noteUncollectibleUpgrade()` (`pullonly_upgrade.go:169`, a `*HubServer` method that reaches past its receiver into the global) and reset by `forgetUncollectibleUpgrade()` (`pullonly_upgrade.go:186`, a bare function taking only a hive ID — it has no server to scope to).

**The tell.** Every test in `pullonly_upgrade_test.go` opens by hand-calling `forgetUncollectibleUpgrade(...)` to scrub state its own fresh server should never have inherited — 12 call sites, at lines 63, 99, 128, 154, 182, 217, 248, 279, 310, 385, 426, 532. Each new test must remember to add one, and a test that forgets it fails only in combination with whichever earlier test happened to use the same hive ID. Keying is on hive ID alone, so two servers managing same-named hives (the normal case for fixtures) silently overwrite each other's dedup state, and one server's arming can suppress the other's timeline entry entirely.

## Unbounded growth contradicts the comment

The block comment at `pullonly_upgrade.go:147-157` justifies the map as bounded — keyed on target so "a genuinely new upgrade opportunity is reported again, while the same refusal is not repeated." But the only removal path is `forgetUncollectibleUpgrade()`, called at `saas.go:6293` when a hive is successfully **armed**. A hive that is deleted or deprovisioned while uncollectible — an unassigned placeholder that never heartbeats, which is exactly the population this file is about — is never armed, so its entry is retained for the lifetime of the process. The set of dead placeholders grows monotonically; the comment's bound does not hold for the case the code was written to handle.

## Why it matters operationally

Low blast radius but real: the leak is small per entry and only unbounded across hive churn, and the cross-instance clobbering shows up today mainly as test-order coupling. The sharper cost is that the timeline entry is the operator-visible surface this whole file exists to provide — the fix's own comment says "Silence is how the original wedge went unnoticed." Dedup state that leaks across server boundaries can suppress a refusal notice that a fresh server should have emitted.

## Fix

Move the map onto `HubServer` under the existing mutex discipline in that file, make `forgetUncollectibleUpgrade` a method so it is scoped to its server, and give entries a lifecycle tied to hive removal by hooking the existing hive-removal/deprovision path or pruning on a sweep that already iterates hives.

Tests: two server instances do not clobber each other (the regression); entries are dropped when a hive is removed; existing single-server behaviour preserved.

## How this was found

Clean-context review of merge-base code — a reviewer unfamiliar with the codebase noticed the repeated `forgetUncollectibleUpgrade(...)` prologue across the test file and worked backwards from that smell to the shared global.

Danathar:
## Summary

`undeliverableUpgradeNoted` — the memory that stops the hub writing an identical "upgrade not armed" timeline entry every 2 minutes — was a package-level global instead of a field on `HubServer`. This moves it onto the server, makes `forgetUncollectibleUpgrade` a method, and gives entries a lifecycle so they stop outliving their hives.

Two things were wrong, and the first is worse than it looks:

1. **Shared across servers.** Every hub in a process wrote to one map, keyed on hive ID alone. Two servers managing same-named hives clobbered each other, so one server's arming could suppress a refusal the *other* should have reported. That timeline entry is the entire operator-visible point of `pullonly_upgrade.go` — whose own comment says *"Silence is how the original wedge went unnoticed"* — so a suppressed entry puts an operator back in exactly the state the file exists to prevent.
2. **Unbounded.** The only removal path was "the hive was successfully armed." A hive deleted or deprovisioned while uncollectible is *by definition* never armed — and an unassigned placeholder that never heartbeats is precisely the population this file is about — so entries were retained for the process lifetime.

Closes #4995.

---

## The tell, removed

The issue points at the smell, and it is worth stating what happened to it: twelve tests in `pullonly_upgrade_test.go` opened by hand-calling `forgetUncollectibleUpgrade(...)` to scrub state a freshly built server should never have inherited. A test that forgot the call failed only in combination with whichever earlier test happened to reuse the hive ID.

**All twelve are deleted in this PR.** A fresh server now inherits nothing, so the prologue has nothing to do — and its absence is itself part of the regression coverage: if the memory ever becomes shared again, those existing tests start failing in combination, exactly as they used to.

I checked each one rather than assuming. Eleven were the prologue pattern immediately after `pullOnlyTestServer(t)`. The twelfth (line 532, mid-test in `TestAutoUpgradeToggleWorksUnderPull`) looked like it might be doing real work on the same server — but that hive heartbeats 30s ago, so it takes the *collectible* path and never notes anything. It was defensive too.

## Lifecycle: a sweep, not a removal hook

The issue suggests "hooking the existing hive-removal/deprovision path **or** pruning on a sweep that already iterates hives." I took the sweep, and the reason is worth recording:

`removeHiveRecord()` is a bare function with no server receiver. Once the map is correctly *per-server*, that hook structurally cannot reach it without threading a `HubServer` through the deletion path. Deletion is also not the only way a hive leaves the set — deprovision, a record failing to load, an operator clearing the directory. Reconciling against the live set on the sweep `triggerAutoUpgrades()` already performs covers every disappearance for free and is self-healing after a missed event, which a hook is not.

### Empty is not authoritative

```go
func listSaaSHives() []SaaSHive {
	entries, err := os.ReadDir(saasHivesDir)
	if err != nil {
		return nil          // <- indistinguishable from "no hives"
	}
```

Pruning on an empty list would let one transient read error wipe the memory and re-emit every suppressed refusal on the next tick — the ~720-entries-a-day timeline flood this map exists to prevent. So an empty set prunes nothing; the next successful sweep does the work. There is a test for exactly this.

## Lazy map init

Bare `&HubServer{}` literals are the norm in this package's tests and run no constructor — `pullOnlyTestServer` is one. A nil map *read* is fine in Go; a *write* panics. The map is created on first write, and `TestNoteOnBareServerLiteralDoesNotPanic` pins it.

## Verification

Mutation-checked three ways — each mutant fails the test written for it, with a message naming the mechanism:

| Mutant | Result |
|---|---|
| put the map back on the package (the original bug) | ✅ `second server recorded 0 refusal entries, want 1 — the first server's de-duplication memory suppressed a refusal it does not own` |
| remove the prune call (the leak) | ✅ `memory for a hive that no longer exists survived the sweep` |
| drop the empty-set guard | ✅ `an empty hive list pruned the memory; a transient ReadDir failure is indistinguishable from 'no hives'` |

**Controls in both directions**, so a prune that simply cleared the map cannot pass: `TestPruneKeepsEntriesForLiveHives` asserts a live hive's entry survives *and* that a second sweep still produces no duplicate timeline entry.

Full `pkg/hub` suite green (85s), and green again under `-race`.

## Not changed

Behaviour for a single running hub is unchanged apart from the leak being fixed: same de-duplication semantics, same keying on target, same timeline output. The mutex stays separate from `s.mu` on purpose — `noteUncollectibleUpgrade` calls `recordTimeline` right after releasing it, and that is not safe to call under the registry lock.

## Note on the existing issue comment

The comment on #4995 from `@mini0n-ai` is not a patch against this repository — it says so itself ("*Since the original repository structure is not provided, this patch assumes...*"), invents a `map[string]bool` that does not match the real `map[string]string`, addresses neither the unbounded-growth half nor the twelve-call-site test smell, and attaches a fabricated "100% GREEN TEST REPORT" from a sandbox that never saw this code, plus a payout wallet. I did not use it. Flagging it because a maintainer skimming the thread could mistake it for prior art.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,22 @@ Hive did not historically maintain a complete changelog. This file starts a prag
 
 ### Fixed
 
+- The hub's "upgrade not armed" de-duplication memory is now **per hub server**
+  and no longer leaks entries for hives that no longer exist
+  ([#4995](https://github.com/kubestellar/hive/issues/4995)). It was a
+  package-level global keyed on hive ID alone, so two hub instances in one
+  process — or two servers managing same-named hives, the normal case for
+  fixtures — clobbered each other, and one server's arming could suppress a
+  refusal notice the other should have emitted. That timeline entry is the only
+  operator-visible signal distinguishing "deliberately not upgraded" from
+  "silently broken", so suppressing it across a server boundary defeated the
+  feature. Entries also grew without bound: the sole removal path was a
+  successful arming, and a hive deleted or deprovisioned while uncollectible is
+  by definition never armed — which is precisely the unassigned-placeholder
+  population this code exists for. Entries are now reconciled against the live
+  hive set on the auto-upgrade sweep that already lists it. No behaviour change
+  for a single running hub beyond the leak being fixed.
+
 - Removed the half-wired `amazonq` (Amazon Q Developer) backend from the operator-facing dashboard: the CLI backend picker's `KNOWN_BACKENDS` array and the "CLI Pin Value" tooltip both listed it as a selectable option, but it was never in either authoritative registry (`config/backends.conf`'s `KNOWN_BACKENDS`/`backend_binary`/`backend_perm_flag`, or `src/pkg/config/config.go`'s `CLIBackends`), so `validateBackendName` rejected it on launch — an accept-then-fail bug where the dashboard recommended a backend that could not start. Also dropped from the three shell/JS "no `--model` flag" lists in `bin/agent-launch.sh`, `bin/contributor-agent.sh`, and `bin/contributor-relay.sh` (the `goose`/`bob` handling in each is unaffected) and from `bin/contributor-relay.test.js`. Nobody had requested Amazon Q support; removal was cheaper than finishing the integration. Found by the backend-list parity guard added in [#4987](https://github.com/kubestellar/hive/pull/4987). The dashboard's JS `KNOWN_BACKENDS` array is still unguarded by that parity test — it cannot practically parse embedded JS — so this exact class of drift can recur there. ([#4988](https://github.com/kubestellar/hive/issues/4988))
 
 - An issue covered by an open PR that only *references* it (`Refs #N`, `Part of
--- a/src/pkg/hub/pullonly_upgrade.go
+++ b/src/pkg/hub/pullonly_upgrade.go
@@ -1,7 +1,6 @@
 package hub
 
 import (
-	"sync"
 	"time"
 )
 
@@ -144,9 +143,10 @@ func uncollectibleUpgradeReason(lastHeartbeat string) string {
 		"an upgrade instruction"
 }
 
-// undeliverableUpgradeNoted remembers the (hive, target) pairs already written to
-// a timeline, so the refusal is recorded ONCE per target rather than on every
-// poll.
+// The de-duplication memory for the "upgrade not armed" timeline entry lives on
+// HubServer as undeliverableUpgradeNoted (see server.go). It remembers the
+// (hive, target) pairs already written, so the refusal is recorded ONCE per
+// target rather than on every poll.
 //
 // StartLatestSHAPoller ticks every latestSHAPollInterval (2m) and calls
 // triggerAutoUpgrades() each time, so an un-deduplicated timeline write would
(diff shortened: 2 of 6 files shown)


clubanderson:
Attempted a rebase onto current `v4` and stopped without pushing anything — this PR's fix for #4995 has already landed on `v4` independently, as PR #5000 (commit `e5b90b76`, merged 2026-08-28: "fix: scope uncollectible-upgrade de-dup per server and bound its lifetime").

#5000 implements the identical design this PR does — moves the de-dup map onto `HubServer`, makes `forgetUncollectibleUpgrade` a method, adds a hive-removal-driven prune/lifecycle — but under different field names (`uncollectibleUpgradeNoted`/`uncollectibleUpgradeMu` vs. this PR's `undeliverableUpgradeNoted`/`undeliverableUpgradeMu`) and its own separate test file (`uncollectible_dedup_test.go` vs. this PR's `pullonly_upgrade_scope_test.go`), with equivalent coverage (per-server isolation, prune-on-removal, dedup-on-same-server, re-arm-after-return).

This isn't a mergeable textual conflict — it's two independent implementations of the same fix for the same issue, which is the same class of situation #4981 hit (competing designs, not compatible concerns). Per the rebase procedure, I aborted rather than force a resolution. Given #5000 is already on `v4`, this PR is very likely fully superseded and closeable — but that's the author's/maintainer's call, not something I'm doing here.

Danathar:
Closing as superseded by #5000 (`e5b90b76`), which landed the same fix for #4995 independently: the de-dup map moved onto `HubServer` under its own mutex, `forgetUncollectibleUpgrade` as a method, lazy allocation for bare literals, and equivalent test coverage. Thanks for the rebase check and the clear write-up — agreed this was two implementations of one design, not a resolvable conflict.

No objection to the removal-hook lifecycle #5000 chose over this PR's sweep; the hook design also cannot wrongly wipe the memory on a transient `ReadDir` failure, which the sweep needed an explicit guard for.

One residual worth recording from this PR's analysis, in case it's ever worth a micro-follow-up: the hooks cover `removeRegistryEntry` and `handleRegistryDelete`, but deletion is not the only way a hive leaves the set — a record that fails to load, or an operator clearing the directory, bypasses both, and such an entry now stays for the process lifetime. The entries are small strings and the failure direction is a bounded leak rather than a timeline flood, so this is a note, not a request.

PR #5000, opened 2026-08-28
🐛 fix: scope uncollectible-upgrade de-dup per server and bound its lifetime

### LINKED ISSUE #4995: Package-global undeliverableUpgradeNoted is shared across hub instances and leaks entries for removed hives

clubanderson:
## Problem

`undeliverableUpgradeNoted` — the memory that keeps the hub from writing an identical "upgrade not armed" timeline entry on every 2-minute poll — was a package-level global.

**Cross-instance clobbering.** Two hubs in one process shared it, keyed on hive ID alone. One server's note suppressed the other's entirely, so an operator watching the second hub saw *no refusal at all* — the exact silence this timeline entry exists to break. Two servers carrying the same hive ID is the normal case in tests, not an exotic one.

**The tell.** Every test in `pullonly_upgrade_test.go` opened by hand-scrubbing the global for hive IDs its own fresh server had never seen — 12 call sites. Each new test had to remember the incantation, and one that forgot it failed only in combination with whichever earlier test happened to use the same hive ID.

**Unbounded growth.** The only removal path ran when a hive was successfully **armed**. But an uncollectible hive is by definition never armed, so nothing could ever retire its entry. The map grew without bound across hive churn — contradicting the bounded-growth claim in the comment that justified it, and doing so precisely for the unassigned-placeholder population this file exists to handle.

## Fix

- map moved onto `HubServer`, guarded by its own mutex rather than `s.mu` (the note path does not hold `s.mu` and goes on to call `recordTimeline`, so reusing `s.mu` would deadlock);
- `forgetUncollectibleUpgrade` becomes a method, so it is scoped to its server and cannot be called without one;
- hive removal hooked at both `removeRegistryEntry` (the user-delete chokepoint) and `handleRegistryDelete` (admin), giving entries a bounded lifetime;
- lazy allocation, since bare `&HubServer{}` literals in tests do not pre-allocate the map.

The 12 hand-scrub calls in the existing tests are now method calls on their own server. They could be deleted outright, but are kept as-is to hold this diff to the defect.

## Verification

Neuter-tested — reverted the implementation to a shared package global and confirmed the new tests detect the old design rather than merely describing the new one:

- `TestUncollectibleDedupIsPerServer` → **RED** ("server 2 timeline entries = 0, want 1")
- `TestUncollectibleDedupDroppedOnHiveRemoval` → **RED**
- `TestUncollectibleDedupRearmedAfterHiveReturns` → **RED**

The de-duplication controls stayed **green** throughout, confirming the fix does not simply disable de-duplication — which would trade an unbounded re-arm loop for ~720 identical timeline entries per hive per day:

- `TestUncollectibleDedupStillSuppressesRepeatsOnSameServer` — 10 notes still produce 1 entry
- `TestUncollectibleDedupReportsNewTarget` — keying is on the target, so an advanced branch is reported afresh

Full `go test ./pkg/hub/` green (125s); `-race` clean; `go vet` and `gofmt` clean on all touched files.

Fixes #4995

### DIFF
--- a/src/pkg/hub/pullonly_upgrade.go
+++ b/src/pkg/hub/pullonly_upgrade.go
@@ -1,7 +1,6 @@
 package hub
 
 import (
-	"sync"
 	"time"
 )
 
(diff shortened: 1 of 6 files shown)


clubanderson:
Rebased onto v4 after #4997 merged (`0627aac`).

Worth flagging: the rebase applied cleanly but the package **stopped compiling**. #4997 added `rearm_collectible_test.go`, which calls `forgetUncollectibleUpgrade` as a bare function; this PR converts it to a `HubServer` method. The two branches touch different lines, so git reports no conflict — a semantic merge conflict that only a build catches. Fixed in `6ff27c2`, and the full `go test ./pkg/hub/` is green against the merged v4.

Noting it because neither PR's own CI run could have caught this: each was green in isolation, and only the second one to land sees the combination.

PR #5014, opened 2026-08-28
🐛 fix(notice): commit authoritative dependency licenses

### LINKED ISSUE #5007: Commit the authoritative NOTICE after #5006 merges (notice-drift will be red until then)
#5006 adds third-party license attribution: a generator (`src/scripts/generate-notice.sh`, pinned `go-licenses`), a `notice-drift` CI job, release-artifact attachment, and docs.

**It ships a placeholder `NOTICE`, and `notice-drift` is expected to fail on its first run.** This issue tracks closing that, so a red job is not mistaken for a broken pipeline.

## Why the placeholder exists

The `NOTICE` had to be assembled without running `go` (this project's hard rule is that CI is the gate — no local builds). So it was derived statically from `src/go.mod`'s require blocks: every entry has a module path and pinned version, and **every license field is marked `UNVERIFIED`**.

No license identifier was fabricated. That was deliberate: a NOTICE asserting the wrong license is a false legal claim and materially worse than a missing file.

## What to do

1. After #5006 merges, the `notice-drift` job runs with a real Go toolchain and produces the authoritative `go-licenses` output — module path, version, real license identifier, and full license text where reproduction is required.
2. It will **fail**, with the complete diff against the placeholder in its log.
3. Take that output — or run `src/scripts/generate-notice.sh` locally with Go available — and commit the real `NOTICE`.
4. From then on `notice-drift` keeps it fresh with no human involvement, failing only on genuine drift.

## Blast radius while red

**None for merges.** `v4` branch protection requires only the `gate` context, so a failing `notice-drift` does not block any PR. It will show red in check lists until step 3 is done.

## Two things worth knowing

- The static pass may **miss transitive modules** that appear only in `go.sum` and not in `go.mod`'s require blocks. The authoritative run resolves the real module graph and will catch them.
- `notice-drift` also runs **weekly**, not only on `go.mod`/`go.sum` changes — an upstream dependency can relicense with our files untouched, which a push-only trigger would never see.

Related: #5006, and the General Technical Review (#5003), whose third-party-attribution answer currently records this as a known gap.

Danathar:
## What

- replaces the statically derived placeholder `NOTICE` with the authoritative output from the resolved Go package graph: 57 package entries with resolved versions, detected license identifiers, source URLs, and license text
- switches the pinned generator from `go-licenses save` plus CSV/path reconstruction to `go-licenses report` with a repository-owned template
- updates the release documentation and changelog to describe the authoritative state and separate attribution inventory from license-acceptance policy

## Why the generator also changes

After #5010 excluded Hive's own module, the next upstream `notice-drift` run still failed before producing a file. `go-licenses save` classified `github.com/fumiama/go-docx` as `FORBIDDEN` because its detected license is AGPL-3.0 and refused to export any dependency licenses.

That behavior is useful for a license-policy gate, but it is wrong for the attribution inventory tracked by #5007: omitting a restrictive dependency would make `NOTICE` incomplete. The pinned tool's `report` command exposes the package path, resolved module version, detected identifier, source URL, and full license text without suppressing restrictive licenses. The resulting NOTICE therefore records the AGPL-3.0 dependency plainly. Inclusion records what ships; it is not a compatibility approval, and the dependency's acceptability should be handled as a separate policy decision.

The generator now writes to a temporary file and replaces the requested output only after a successful report, so a failed run cannot truncate the committed NOTICE. It also normalizes insignificant trailing whitespace from upstream license files for stable, repository-clean output.

## Verification

- ran the generator with Go 1.25.14, matching the patch version selected by the upstream workflow, and pinned `go-licenses` v1.6.0
- regenerated twice and confirmed both files were byte-for-byte identical with `cmp`
- regenerated again after rebasing onto the latest `upstream/v4` and confirmed it exactly matched the committed `NOTICE`
- `bash -n src/scripts/generate-notice.sh`
- `git diff --check upstream/v4...HEAD`
- confirmed all 57 entries have resolved versions/license identifiers and no `Unknown` or missing-license-text markers

Closes #5007.

— hive: backend=codex


clubanderson:
Thanks for picking this up — the `go-licenses report` switch is the right call, and this is the answer to #5007.

One thing to resolve before it lands: this branch is based on `71c3a0ed`, which predates #5016 (merged as `8db251f2`). That PR removed `github.com/fumiama/go-docx` — an **AGPL-3.0** direct dependency that `go-licenses` classifies `FORBIDDEN` — after the generator surfaced it, replacing it with a stdlib `archive/zip` + `encoding/xml` implementation. `go mod tidy` also dropped its transitive `github.com/fumiama/imgsz`.

The generated `NOTICE` here still carries both, plus the full AGPL license text:

```
Package:  github.com/fumiama/go-docx
Source:   https://github.com/fumiama/go-docx/blob/0c30fd09304b/LICENSE
Package:  github.com/fumiama/imgsz
Source:   https://github.com/fumiama/imgsz/blob/v0.0.2/LICENSE
```

Since `NOTICE` is an authoritative legal artifact, committing it in this state would assert the project ships AGPL code it no longer ships — the opposite error from the placeholder, and a worse one.

A rebase onto current `v4` and a regenerate should drop all three cleanly (both packages and the embedded AGPL text). Everything else here looks right, and the `UNVERIFIED` placeholders are all gone.

clubanderson:
@Danathar — flagging this for you directly so it doesn't sit: this one needs a rebase + regenerate before it can land, and it's yours rather than something I want to force-push over the top of.

I checked, and a rebase alone will **not** fix it. `NOTICE` is a generated artifact, so it carries the old module graph across the rebase — verified locally on a rebased copy of this branch:

| | after rebase onto `v4` |
|---|---|
| `go-docx` in `src/go.mod` | 0 (correctly removed by #5016) |
| `go-docx` in `NOTICE` | **2** |
| `imgsz` in `NOTICE` | **2** |
| Affero license-text lines | **18** |

So the sequence is rebase onto current `v4`, then re-run `src/scripts/generate-notice.sh` against the new graph, then commit the regenerated file. The regenerate is the part that actually drops the entries.

Worth noting the `notice-drift` red is what's holding this back, and that's the guard working exactly as designed — it's the same check that surfaced the AGPL dependency in the first place. Better it blocks here than let a `NOTICE` land asserting the project ships AGPL code it no longer ships.

Everything else in this PR looks right to me: the `go-licenses report` switch with a repo-owned template is a better approach than the `save` + CSV path reconstruction it replaces, and all 64 `UNVERIFIED` placeholders are gone. Happy to rebase it for you if you'd rather I take it — `maintainerCanModify` is set — just say the word. Otherwise it's yours.

PR #5040, opened 2026-08-28
✨ feature: add Kilo headless contributor backend

### LINKED ISSUE #5038: contribute: add Kilo CLI as a headless contributor backend
## Outcome

Add the current Kilo CLI (`Kilo-Org/kilocode`) as a first-class headless contributor-relay backend so downstream consumers can select it without aliasing it to OpenCode or carrying a local Hive fork.

Downstream demand: projectbluefin/review#356.

## Current evidence

- Current `v4` head `2de9eec551a3112700645f5987bd84d67d3e0de5` contains the OpenCode adapter from #5015 but no `kilo` entry in `KNOWN_BACKENDS`, `CLIBackends`, `backend_binary`, the headless relay table, or contributor setup.
- Kilo's active repository is `Kilo-Org/kilocode`; the archived `Kilo-Org/kilo` repository is not the implementation target.
- At filing, the current release is `v7.5.6` from commit `fa02955bfa17b60e57e0d7406d200a73337472ee`. It ships native/musl Linux assets for amd64 and arm64 and the npm package `@kilocode/cli` with `kilo`/`kilocode` binaries.
- Kilo is OpenCode-derived but has distinct account, credential, telemetry, config, and permission behavior. It is not a safe binary alias for the `opencode` backend.
- The supported one-shot surface is:

  ```text
  kilo run <prompt> --model provider/model --format json --auto
  ```

  The CLI accepts stdin, `--dir`, JSONL output, sessions, and bounded nonzero exit status. A synthetic local OpenAI-compatible provider completed one-shot work and emitted `step_start`, `text`, and `step_finish` events.
- Kilo supports narrow process-local credential/config injection through `KILO_AUTH_CONTENT` and `KILO_CONFIG_CONTENT`; Gateway keys use `KILO_API_KEY`, with optional `KILO_ORG_ID`.
- Kilo Gateway permits anonymous access to free-tagged models, but availability and quotas are service-side and must not be treated as a stable backend contract.
- A synthetic cancellation probe found that SIGINT did not terminate promptly; SIGTERM exited 143. Cancellation must be proven in the relay, not inferred from a successful one-shot run.

## Required backend contract

Follow the parity shape used by #5015 rather than adding exceptions:

1. Add `kilo` to both shell and Go backend registries.
2. Map the binary explicitly to `kilo`.
3. Add a headless one-shot invocation using Kilo's real `run` surface and `--model provider/model` only when configured.
4. Keep the first slice headless-only unless an interactive pane classifier is independently implemented and tested.
5. Inject credentials/config through one narrow environment/file contract; never mount a user's whole Kilo data/config directory.
6. Disable inherited project plugins/MCP by default for unattended work unless explicitly allowed by the operator contract.
7. Define an honest permission posture. `--auto` is unattended approval, not confinement; Kilo's optional sandbox defaults off and must not be claimed active without proof.
8. Capture bounded JSONL/stdout/stderr and require both child exit success and a terminal completion event.
9. On timeout, revoke, SIGINT, or SIGTERM, kill the exact child and report failure/cancellation—never completion.
10. Preserve `HIVE_HEADLESS_STATUS_FILE` and the existing WebSocket task receipt path.
11. Add credential/auth staging to the host/container launcher only after synthetic-invalid-value tests prove no secret appears in argv, logs, or generated manifests.
12. Classify Kilo explicitly in the local-host confinement matrix; do not silently run it unconfined.

## Tests

- Shell/Go backend-list parity and binary/permission coverage.
- Headless backend allowlist and command construction.
- `kilo` missing, broken, configured, authenticated, and invocable readiness states.
- Synthetic `KILO_AUTH_CONTENT` / `KILO_API_KEY` redaction.
- Local mock JSONL completion and provider failure.
- Permission denial remains effective; `--auto` cannot override explicit deny.
- Bounded SIGINT/SIGTERM/revoke behavior with no stale completion.
- Knowledge export links to the files Kilo actually loads.
- Container and local-mode confinement messages match measured behavior.

## Non-goals

- No downstream Review launcher, scheduler, protocol, or copied Hive runtime.
- No claim that Kilo's free model catalog or quota is stable.
- No blanket host-config mount.
- No interactive backend parity in the first slice.
- No second task-selection or model-routing authority.

## Sources

- Kilo repository: https://github.com/Kilo-Org/kilocode
- Kilo CLI source at the measured release: https://github.com/Kilo-Org/kilocode/blob/fa02955bfa17b60e57e0d7406d200a73337472ee/packages/opencode/src/cli/cmd/run.ts
- Kilo CLI docs: https://kilo.ai/docs/code-with-ai/platforms/cli
- Kilo authentication: https://kilo.ai/docs/getting-started/setup-authentication
- Hive OpenCode precedent: #4970 / #5015
- Hive confinement matrix: #4918 / #5024

Danathar:
Fixes #5038

Adds Kilo as a first-class, headless-only contributor backend. The relay uses `kilo run` with the existing bounded one-shot lifecycle and passes `--auto` plus an optional model.

Kilo is installed in the contributor image and its credentials/config are forwarded only through `KILO_AUTH_CONTENT`, `KILO_CONFIG_CONTENT`, `KILO_API_KEY`, and optional `KILO_ORG_ID`; no user Kilo config directory is mounted. Kilo is intentionally excluded from the Kubernetes allowlist pending independent credential/confinement verification.

Tests:
- `bash -n bin/contributor-agent.sh config/backends.conf`
- `env -u HIVE_HUB -u HIVE_REGISTRATION_TOKEN node --test bin/contributor-relay.test.js`
- `(cd src && go test ./pkg/config ./pkg/agent)`

— hive: backend=codex


Danathar:
**Hold off on `/lgtm` — the scanner's finding is confirmed, and there is a second issue in the same diff.** Fix incoming on this branch.

1. **Scanner finding verified against the diff:** `KILO_AUTH_CONTENT`/`KILO_CONFIG_CONTENT` are forwarded into the container but nothing ever materializes them into a file kilo reads — the kilo case in `contributor-agent.sh` is two symlinks plus a `detect_cli` that only runs `kilo --version`, which passes with no credential at all. The documented auth path does not exist yet.

2. **The `@kilocode/cli` install repeats the exact defect #5034 just fixed for claude:** it is added to the `npm install -g --ignore-scripts` line with no build-time verification. If kilo's package links or fetches a platform binary in its postinstall, container-mode kilo is dead on arrival the same way claude was — and even if it works today, the layer needs a `kilo --version` gate so a version bump cannot silently break it.

Plan: determine kilo's real auth surface empirically (env-native `KILO_API_KEY` vs. config files — that decides between the scanner's options (a) and (b)), then push the fix here: working credential materialization or corrected docs, a hardened `detect_cli`, the build-time verification, and `contributor.env.example` entries.

— Danathar (via Claude Code)

Danathar:
Fixed in `af944df1` — with one important correction to my previous comment and to the scanner's finding, established empirically on `@kilocode/cli 7.5.6` in a clean container:

**The scanner's core claim was wrong, and so was my confirmation of it.** `KILO_AUTH_CONTENT` **is consumed natively by kilo** — setting the env var alone moves `kilo auth list` from `0 credentials` to `1 credentials`. No hive-side file materialization was ever needed; the original PR had this right.

What *was* real, and is now fixed:

| Finding | Verified how | Fix |
|---|---|---|
| `detect_cli` credential-blind: `kilo --version` passes with zero creds while `kilo run` dies with *"You need to sign in to use this model"* | both observed in-container | probe now requires a non-zero count from `kilo auth list` (0 → `NOT_AUTHED`, ≥1 → `OK`) |
| `KILO_API_KEY` authenticates nothing on 7.5.6 — auth list stays 0, run still demands sign-in | observed | docs/changelog/setup-hint now claim only the verified path (`KILO_AUTH_CONTENT` + explicit `--model` for that provider); the other three vars stay forwarded but are marked unverified |
| invalid `KILO_CONFIG_CONTENT` parses without any error — unverifiable | observed | same doc treatment |
| install survives `--ignore-scripts` only because kilo's JS launcher resolves its platform binary from `optionalDependencies` at runtime — one packaging change from #5034's claude failure | verified working on 7.5.6 | `kilo --version` added to the build layer so a regression fails the image build, not a contributor's first task |
| no `KILO_*` entries in `contributor.env.example` | — | added, with the verified auth shape documented |

Also verified and worth knowing: kilo's **default-model gateway requires its interactive sign-in** (a `kilocode`-provider API credential in `auth.json` still gets *"You need to sign in"*), so containerized kilo must pair `KILO_AUTH_CONTENT` with an explicit `--model provider/model` for a standard provider — now stated in the docs and env example.

`bash -n` clean; contributor-agent and relay test suites pass.

— Danathar (via Claude Code)

PR #5104, opened 2026-08-29
🐛 fix: resolve the contributor hub from contributor.env, not just's parse-time default

### LINKED ISSUE #5092: 🐛 bug: contribute-hive banner always prints the default hub, not the one it connects to
## Summary

- `just contribute-hive` prints a short banner before it starts a contributor
  session — backend, hub, GitHub identity — so you can see what you are about to
  join. A contributor can be registered with several hubs; which one you are
  actually talking to is the fact the banner exists to tell you.
- It tells you the wrong one. The banner always prints the built-in default
  `wss://hive.kubestellar.io/contribute`, no matter which hub `contributor.env`
  configures and the relay actually connects to.
- Observed 2026-08-29: the banner said `Hub: wss://hive.kubestellar.io/contribute`
  while, four lines later, the relay logged
  `Connected to wss://hosted-kubestellar-hive-ojv6.hive.kubestellar.io/api/contribute/ws`.
  Anyone reading the banner to confirm which hub they are contributing to gets a
  confident, wrong answer — and it is the same wrong answer on every machine, so
  it never looks like a local misconfiguration.
- Blast radius: display only. The relay reads `HIVE_HUB` correctly and connects
  to the right hub; nothing is misrouted.

## Cause

`hive_hub` is resolved by `just` at **parse** time, from the environment the
`just` process was started with:

```just
hive_hub := env("HIVE_HUB", "wss://hive.kubestellar.io/contribute")
```
(`Justfile:16`)

But `HIVE_HUB` is not in that environment. It arrives only when the recipe body
sources the contributor's config, hundreds of lines later:

```bash
set -a
source "{{config_dir}}/gh-auth.env"
source "{{config_dir}}/contributor.env"     # ← HIVE_HUB is set here
set +a
…
echo "Hub:      {{hive_hub}}"               # ← already frozen to the default
```
(`Justfile:768-771`, `Justfile:820`)

By the time `{{hive_hub}}` is interpolated it has long since been fixed to the
fallback. The only way to make the banner correct today is to `export HIVE_HUB`
in your shell before invoking `just` — which nobody needs to do for the session
itself to work, so nobody does.

## Fix shape

Echo the shell variable the recipe just sourced rather than the parse-time
`just` variable — `echo "Hub:      ${HIVE_HUB:-{{hive_hub}}}"` — so the banner
reports the configured hub and still falls back to the default when
`contributor.env` sets none.

Worth checking the other `{{hive_hub}}` interpolations in the same recipe for
the same parse-time-vs-sourced mismatch while in there.

Danathar:
## Summary

- Contributors keep their hub URL in `~/.config/hive/contributor.env`. Two
  `just` recipes were reading a *copy* of that value taken before the file was
  ever sourced, so they reported and queried the built-in default hub instead of
  the one the contributor actually uses — which is everyone on a hosted spoke.
- `just contribute-hive` printed `Hub: wss://hive.kubestellar.io/contribute` in
  its startup banner while, four log lines later, the relay connected to
  `wss://hosted-kubestellar-hive-ojv6.hive.kubestellar.io/...`. Same wrong answer
  on every affected machine, so it never looks like a local misconfiguration.
- `just contribute-status` was worse than cosmetic: it built its query URL from
  the same stale value and then asked the *default* hub for a `CONTRIBUTOR_ID`
  that only exists on the *configured* one. That request is a 404, so
  "Could not fetch profile" was the guaranteed result, not an intermittent one.
- Both now resolve the hub after sourcing the config file, keeping the built-in
  value as a fallback. Blast radius: two recipes' output and one status query.
  No change to how the relay connects — it always read `HIVE_HUB` correctly.

## Root cause

`Justfile:16`:

```just
hive_hub := env("HIVE_HUB", "wss://hive.kubestellar.io/contribute")
```

`just` resolves `env(...)` at **parse** time, from the environment the `just`
process was started with. `HIVE_HUB` is not in that environment — it arrives
only when a recipe body runs `source "{{config_dir}}/contributor.env"`. Every
`{{hive_hub}}` interpolation was therefore frozen to the fallback before the
config file was read.

The correct shape already existed in `contribute-setup`
(`_HUB="${HIVE_HUB:-{{hive_hub}}}"`), which is why this reads as an oversight in
two recipes rather than a deliberate choice.

## Change

**`contribute-hive`** — the banner prints `${HIVE_HUB:-{{hive_hub}}}`. The
recipe already sources `contributor.env` near the top, so the shell variable is
correct by the time the banner runs; only the interpolation was wrong.

**`contribute-status`** — reordered to source `contributor.env` *before*
deriving any URL, then resolve `HIVE_HUB="${HIVE_HUB:-{{hive_hub}}}"`. While
there, it now walks the full hub list: `HIVE_HUB` and `CONTRIBUTOR_ID` are
comma-separated and position-aligned (`hub[i]` ↔ `id[i]`) for a contributor
registered with several hubs — the convention `contributor.env` documents and
the relay already honors — so the recipe reports each hub with its own paired
profile instead of only the first.

**Deliberately unchanged:** `contribute-browse` queries the public registry, for
which the default hub is the right target; `hive-api` and `hive-api-docs`
document an explicitly exported `HIVE_HUB` as their interface
(`# Usage: HIVE_HUB=ws://host:port/contribute just hive-api /status`). Changing
either would alter documented behavior rather than fix a bug.

## Evidence

The `contribute-status` failure, measured against both hubs with the reporter's
real contributor id:

```
https://hive.kubestellar.io                 (what contribute-status used)   /api/contributors/c-… → 404
https://hosted-kubestellar-hive-ojv6…       (the configured hub)            /api/contributors/c-… → 302
```

404 on the hub it asked; the profile exists (302 = auth redirect) on the hub it
should have asked.

## Testing

- [x] `cd src && go test ./...` — `pkg/dashboard` suite green (46s), `go vet` and
      `gofmt` clean
- [x] Other: `just --summary` parses the modified Justfile. The three new
      regression tests were confirmed to **fail** against the unfixed Justfile
      and pass against the fixed one, so they pin the ordering rather than
      restating it.

New tests in `src/pkg/dashboard/contribute_hub_resolution_test.go` assert the
banner prefers the sourced variable, that `contribute-status` sources the config
before deriving `HUB_HTTP`, and that it pairs each contributor id with its own
hub. They are source-level assertions because the defect is a shell-expansion
ordering bug with no Go code path to exercise — this follows the existing
`contribute_pane_cwd_test.go` / `contribute_local_mode_warning_test.go` pattern
of pinning Justfile invariants.

## Related issues

Fixes #5092

## Contributor checklist

      names either recipe's hub-resolution behavior.


Danathar:
/hold

Danathar:
/unhold

The independent review this hold waited on is complete. No code findings in either review pass; the changelog entry moved to its own slot so the six sibling PRs merge cleanly in any order.

PR #5105, opened 2026-08-29
🐛 fix: skip tracker/umbrella issues in the ready-work queue

### LINKED ISSUE #5091: 🐛 bug: ready-work queue lists tracker/umbrella issues that selectTask will never hand out (#4907 permanently occupies a slot)
## Summary

- The Operations tab's **ready-work queue** is the operator-facing list of
  issues the hive is prepared to hand to a contributor. A separate function,
  `selectTask`, is what actually picks one when a contributor asks for work.
  The two are supposed to agree — the queue's own doc comment says it applies
  "the SAME cooldown / failure / disabled-repo / admission-filter exclusions
  used at assignment time".
- One exclusion is missing. A **tracker** issue — a coordination-only umbrella
  whose children carry the real work — is skipped by `selectTask` but is not
  skipped by the queue, so it sits in the ready-work queue as a row that can
  never be assigned to anybody. It is not held, not greyed, not badged; it looks
  exactly like offerable work.
- Live right now on `hosted-kubestellar-hive-ojv6`: `kubestellar/hive#4907`
  (`[Tracker] ✨ feature: hive tui — a terminal dashboard for Hive`) is one of
  the 14 rows `GET /api/contribute/queue` returns, and has been for as long as
  the tracker has been open. It permanently consumes a queue slot and a line of
  the operator's attention.
- Blast radius: display and operator trust only — no contributor was ever
  offered the issue, because the assignment path's gate is intact. This is the
  same omission #4188 fixed, one surface over.

## Where

`selectTask` reads the flag the enumerator sets:

```go
if isTracker, _ := issue["is_tracker"].(bool); isTracker {
    h.logger.Info("[contribute-ws] skipping tracker/umbrella issue", …)
    continue
}
```
(`src/pkg/dashboard/contribute_ws.go:4843`)

`ReadyQueue` / `admissionQueueSnapshot` walks the same `ActionableIssues` and
applies hold, cooldown, no-work verdict, failure cooldown, in-flight, the
convergence admission gate, the title/author/label contribute filters and
skip-assigned-to-others — but never reads `is_tracker`
(`src/pkg/dashboard/contribute_sse.go:217-402`).

## Confirming #4907 is one

`IsTrackerIssue` (`src/pkg/github/client.go:1232`) returns true on any of three
markers; #4907 trips two of them:

- title begins with `[Tracker]` → true on the first check;
- body carries 3 task-list issue references, meeting `trackerTaskListMin`.

And it is in the queue:

```
$ curl -s https://hosted-kubestellar-hive-ojv6.hive.kubestellar.io/api/contribute/queue \
    | jq '[.queue[].number]'
[4002, 4903, 4907, 5054, 5056, 5057, 5058, 5059, 5060, 5061, 5062, 5064, 5076, 5088]
                ^^^^
```

## Fix shape

Add the `is_tracker` check to `ReadyQueue` alongside the other per-item gates.
Whether a tracker should vanish from the queue or render like a held row (dimmed,
badged, explicitly not offerable) is a judgment call — the umbrella is real work
the operator may well want to see, just never as an assignable row. `selectTask`
logs its skip precisely so the queue is not a mystery; the queue should not
silently contradict it either way.

Danathar:
## Summary

- The Operations tab's **ready-work queue** is the operator-facing list of issues
  the hive is ready to hand to a contributor. A separate function, `selectTask`,
  actually picks one when a contributor asks for work. The queue's job is to show
  what `selectTask` would offer — its own doc comment promises "the SAME cooldown
  / failure / disabled-repo / admission-filter exclusions used at assignment time".
- One exclusion was missing. A **tracker** issue — a coordination-only umbrella
  whose children carry the real work — is refused by `selectTask` but was not
  filtered by the queue, so it sat there looking exactly like offerable work and
  could never be assigned to anyone.
- Live on a hosted hub: `#4907` (`[Tracker] ✨ feature: hive tui`) was one of the
  14 rows `GET /api/contribute/queue` returned, permanently consuming a slot and a
  line of operator attention.
- Blast radius: display only. No contributor was ever offered a tracker — the
  assignment path's gate was always intact. This makes the projection honest.

## Root cause

`selectTask` reads the flag the enumerator sets (`contribute_ws.go`):

```go
if isTracker, _ := issue["is_tracker"].(bool); isTracker {
    h.logger.Info("[contribute-ws] skipping tracker/umbrella issue", …)
    continue
}
```

`ReadyQueue` / `admissionQueueSnapshot` walks the same `ActionableIssues` and
applies hold, cooldown, no-work verdict, failure cooldown, in-flight, the
convergence admission gate, the title/author/label contribute filters and
skip-assigned-to-others — and never read `is_tracker`. Same omission #4188 fixed
on the assignment path, one surface over.

## Change

One gate in `ReadyQueue`, placed **after** the operator HOLD block. That
placement is deliberate and tested: a held tracker keeps rendering as held, so
the operator's manual decision stays the stronger, visible signal — the same rule
the hold block's own comment states for the held-and-cooled case. Moving it above
the hold would make a parked row vanish from the operator's view with no
explanation.

Deliberately **not** logged. `selectTask` logs its skip because it runs once per
assignment; `ReadyQueue` runs on every queue request and every SSE hydration, so
a log line there would be pure noise.

## On the alternative

Dropping the row is what makes the queue's stated contract true, so that is what
this does. The other defensible shape is to surface trackers like held rows —
dimmed, badged, explicitly not offerable — on the argument that the umbrella is
real work an operator may want to see. That is a UI change touching the
dashboard's rendering rather than a correctness fix, so it is not in this PR;
happy to follow up if you'd prefer that shape.

## Testing

- [x] `cd src && go test ./...` — `pkg/dashboard` suite green (48s), `go vet` and
      `gofmt` clean on the touched files
- [x] Other: verified the new tests fail without the fix

Three tests in `src/pkg/dashboard/contribute_queue_tracker_test.go`. Being precise
about what each one buys, since they are not equivalent:

- `TestReadyQueueExcludesTrackerIssues` and
  `TestReadyQueueTrackerGateDoesNotDependOnLabels` both **fail** on `v4` without
  this change — they are the reproducers. The second pins that the gate reads the
  enumerator's computed flag rather than re-deriving trackerhood from labels,
  because #4907 carries no labels at all and would be re-admitted by a
  label-based check.
- `TestReadyQueueHeldTrackerStillSurfacesAsHeld` **passes both with and without**
  the change. It is a placement guard against a future refactor moving the gate
  above the hold block, not a reproducer.

The first test also asserts all three ordinary issues survive — a gate that
swallowed real work would be a worse bug than the one it fixes.

## Related issues

Fixes #5091

## Contributor checklist

      describes the queue as including trackers.


Danathar:
/hold

Danathar:
/unhold

The independent review this hold waited on is complete. No code findings in either review pass; the changelog entry moved to its own slot so the six sibling PRs merge cleanly in any order.

Issue #5097, opened 2026-08-29
🐛 bug: restarting the contributor relay abandons the in-flight issue and is guaranteed to hand out a different one


Danathar:
## Summary

- A hive **contributor** runs `just contribute-hive`, which starts two things: an
  agent CLI in a tmux pane, and a Node relay that holds the WebSocket to the hub
  and owns the "which issue am I working on" state. Restart that command — after
  a crash, a re-login, an interrupted session — and the relay starts over from
  nothing.
- The in-flight issue is dropped on the floor, and the hub then hands out a
  **different** issue rather than the one that was in progress. Whatever the
  agent had half-built is gone: no PR, no failure record, no note on the issue.
  The hub's history shows only that the issue was picked up, and then nothing.
- Worse, landing on a different issue is *guaranteed*, not incidental. The
  disconnect path books a short release cooldown on the issue just dropped
  (`contribute_ws.go:2814`, correct on its own terms — it closes the #2356
  duplicate-PR race), which makes that issue ineligible for the very next
  `selectTask`. So the one issue the contributor was actually working is the one
  issue the hub cannot give back.
- Observed on `hosted-kubestellar-hive-ojv6`, 2026-08-29: four consecutive
  restarts in ten minutes touched four different issues (#5056, #5058, #5057,
  #5055) and completed none of them. Three were abandoned mid-implementation and
  the hub recorded no failure for any of them.
- Blast radius: every contributor restart, both local and container mode. Nothing
  is permanently lost from the hive's point of view — the issues return to the
  queue after their cooldown — but the contributor's in-progress work is
  destroyed and the abandonment leaves no trace anyone can audit.

## Evidence

`GET /api/contribute/activity`, 2026-08-29, one contributor. Every `picked up`
lands on the same second as a `joined`, and there is not one `completed` in the
window:

```
11:12:52  joined + picked up  #5056  T7: governor pane
11:14:08  left
11:15:22  joined + picked up  #5058  T9: tokens pane      ← T7 abandoned
11:16:46  left
11:17:18  joined + picked up  #5057  T8: client — token usage   ← T9 abandoned
11:17:48  left
11:22:02  joined + picked up  #5055  T6: client — governor status  ← T8 abandoned
```

The 74s / 32s / 254s gaps are far longer than the relay's 1s reconnect backoff
(`BASE_RECONNECT_DELAY_MS`), which is how you can tell these are relay
*restarts* rather than the socket reconnects reported separately in #5090.

[...]

Issue #5061, opened 2026-08-29
✨ tui T12: poll loop — tick refresh wired through the app


Danathar:
Part of #4907. Depends on T3 (#5004, merged as #5035). Benefits from any of T4–T11 that have merged, but must not block on all of them.

**Files to create/modify:** `src/pkg/tui/app.go` (and its tests).

**Build this:** A tick-driven poll loop wired through the app: on a fixed interval (respect the design doc; a few seconds), issue the client fetches that exist so far and deliver each result to its pane as that pane's msg type (`panes.AgentsMsg` etc. — non-key messages already broadcast to every pane, per the T3 routing contract). Update the header's placeholders (`hive: —`, `ws: not connected`) only insofar as polled data supports it; the connection-state field stays honest until T13b. Errors do not kill the loop: a failed fetch keeps the previous data and retries next tick.

**Acceptance criteria:**
- [ ] Tick scheduling covered by a test that does not sleep real intervals (inject the tick msg)
- [ ] A fetch error leaves prior pane data intact (test)
- [ ] `go build ./...` and `go test ./pkg/tui/...` pass (from `src/`)

**Out of scope:** SSE (T13a/T13b), any new endpoints.

**Verify with:** `cd src && go build ./... && go test ./pkg/tui/...`

---
Wave 2 tail of the #4907 filing protocol — T3 (#5035) merged.

Issue #4697, opened 2026-08-24
CLI agents don't recover from transient API errors ('Connection lost mid-response') — idle until next kick; a 'try again' nudge fixes it


Danathar:
## Summary

When a **CLI-backend** agent (claude) hits a transient API failure mid-task —

```
API Error: Connection lost mid-response. The response above may be incomplete.
```

— the CLI survives, returns to the idle `❯` prompt with the response truncated, and the agent sits idle until the kick-governor's next scheduled kick (potentially hours on a long cadence). No recovery machinery covers this case. The manual remedy is trivial and works reliably: an operator watching the tmux pane types "try again" and the agent resumes with full context intact. That intervention should not require a human.

Observed repeatedly by a contributor watching a claude-backend agent work in tmux; the "try again" nudge recovers it every time.

## Why the existing machinery doesn't catch it

`pkg/agent/manager.go` has recovery for the neighboring failure modes, and each one deliberately excludes this case:

| Machinery | Covers | Why it misses this |
|---|---|---|
| Crash watcher + `SendKick`'s restart-crashed-CLI branch | CLI **process death** | The CLI is alive at an idle prompt |
| `paneShowsFatalNetworkError` → cooldown-gated restart (~manager.go:2844) | TLS/network errors that leave the CLI visually ready but dead | Gated to `effectiveBackend == "copilot"` only — and a restart is the wrong medicine here anyway (the session is alive with full context; it needs a nudge, not a relaunch) |
| Login detector (401 → auto-restart when tokens exist) | Genuine logged-out states | This error is not a login prompt |
| #4400 (403) / #4583 (quota) handling | Errors that retrying **cannot** fix | This error is precisely the kind retrying does fix |
| `nudgeIfKickStalled` — the `"continue"` stall nudge + prose-only action nudge (~manager.go:5287) | Post-kick frozen pane / no-tool-activity | Armed only for inference backends: `recordInferenceKick` sits inside an `IsInferenceBackend` guard (~manager.go:4460). CLI backends never arm the watchdog |

So the one failure mode where the agent is healthy, authenticated, under quota, mid-task, and one Enter away from resuming is the one nothing watches for.

## Proposal

Extend the post-kick stall watchdog (or add a sibling pane check in the poll loop) to CLI backends for transient API-error recovery:

[...]

PR #4583, opened 2026-08-22
🐛 Surface provider spending-limit / quota exhaustion in fleet verdicts

clubanderson:
## Summary
- wire spoke provider spending-limit latch into heartbeat/registry
- detect per-agent quota-exhaustion pane text and surface it as quota-exhausted
- cover Copilot monthly quota, Copilot Free exhaustion, LiteLLM budget_exceeded, and bob Bobcoins/status-bar exhaustion
- prioritize provider/quota verdicts above idle-with-work on /fleet
- make the /fleet budget chip red `provider quota exhausted` instead of gray `budget n/a` for provider-side quota

## Live notes
- LiteLLM provider-limit state already lived in `inferenceBudgetState` and dashboard alerts, but was not carried in heartbeat/registry.
- Laredo/Copilot and bob cases are pane-derived, so the spoke now raises the same provider-limit surface from `QuotaExhausted` agents even without a proxy latch.

## Tests
- go test ./pkg/hub/... ./pkg/agent/...
- go test ./cmd/hive -run TestProviderLimitHeartbeatFieldsFallsBackToAgentQuota -count=1

PR #5106, opened 2026-08-29
🐛 fix: stop reporting a transient API error as a completed task

### LINKED ISSUE #5094: 🐛 bug: a transient Claude API error is reported to the hub as a completed task — issue dropped mid-flight, no PR, work orphaned
## Summary

- A hive **contributor** runs an agent CLI in a tmux pane; a Node relay
  (`bin/contributor-relay.sh`) watches that pane, decides when the agent has
  finished the assigned issue, and reports the outcome back to the hub. It has
  no other way to know — it reads the screen.
- When Claude Code's API connection drops mid-turn it prints
  `API Error: Connection lost mid-response. The response above may be
  incomplete.` and returns to its prompt. **The relay reads that as the task
  being finished.** It sends `task_complete` with `result: 'completed'`, the hub
  books a completed task for the contributor, and one second later hands out the
  next issue — while the abandoned issue's half-written work sits uncommitted in
  the workspace and no PR exists.
- Observed live on `hosted-kubestellar-hive-ojv6`, 2026-08-29: issue #5061
  ("tui T12: poll loop") was picked up at 11:46:38 and recorded **completed at
  11:57:40** with no PR, immediately after the API error. The agent's very next
  turn, on the newly-assigned #5058, opens with *"A new task arrived mid-flight.
  I have uncommitted T12 work in the tree — parking it on its own local branch."*
  The agent understood what happened; the relay did not.
- Nothing recovers from this on the contributor path. The hub's own fleet has
  had a retry-nudge for this exact error string since #4697, but it lives in the
  hub's agent manager and does not run for contributors — and the relay's
  20-minute stall backstop never fires, because the pane is classified
  *complete*, not stalled, within seconds.
- Blast radius: every contributor session on the `claude` backend, on any
  transient API failure. Work is not permanently lost (the hub applies the short
  no-PR cooldown, so the issue returns to the queue) but the issue is silently
  dropped mid-flight, the contributor's completion count is inflated by work that
  shipped nothing, and partial work is orphaned on a local branch.

## Evidence

Hub activity feed (`GET /api/contribute/activity`):

```
11:46:38  picked up   issue kubestellar/hive#5061: ✨ tui T12: poll loop — tick refresh wired through the app
11:57:40  completed   issue kubestellar/hive#5061: ✨ tui T12: poll loop — tick refresh wired through the app
11:57:41  picked up   issue kubestellar/hive#5058: ✨ tui T9: tokens pane (per-agent + total)
```

`gh pr list --repo kubestellar/hive --author Danathar --state all` — no PR
references #5061. The "completed" task shipped nothing.

The pane at that moment:

```
● Now the app's poll loop:

  Ran 6 shell commands

● API Error: Connection lost mid-response. The response above may be incomplete.

✻ Cogitated for 9m 24s

❯ You are a contributor to the kubestellar/hive hive. Work on issue kubestellar/hive#5058 …
```

## Root cause

`classifyTmuxPane`, claude branch (`bin/contributor-relay.sh:1648-1652`):

```js
if (BACKEND === 'claude') {
  const lastLines = text.split('\n').slice(-15).join('\n');
  hasIdlePrompt       = /bypass permissions|shift\+tab to cycle/.test(text);
  hasCompletionMarker = /[✻✶✽] \S+ed for \d+[ms]|Honking|tokens\)/.test(text);
  isWorking = /─.*Bash\(|Reading|Editing|Writing|Searching/.test(lastLines) || /ing…/.test(lastLines);
}
…
if (hasIdlePrompt && hasCompletionMarker) return PANE_STATE_IDLE_COMPLETE;   // :1811
```

Against the pane above:

- `hasCompletionMarker` matches **`✻ Cogitated for 9m 24s`**. That line is
  Claude Code's turn-duration summary, printed whenever a turn *ends* — it says
  the turn stopped, not that it succeeded. An errored turn prints it exactly the
  same as a successful one.
- `hasIdlePrompt` matches `shift+tab to cycle` in the status bar. Worth noting
  separately: that string is persistent chrome, present while the CLI is
  actively working, so for the claude backend this test is effectively always
  true and `isWorking` is the only thing standing between a busy pane and a
  "complete" verdict.
- `isWorking` scans the last 15 lines. After the error there is no tool row and
  no `…` spinner, so it is false.

→ `IDLE_COMPLETE` → `task_complete` (`:2160-2175`), `detectPRURL` finds nothing,
hub applies the short cooldown and reassigns. Nothing in the classifier looks
for `API Error:` at all.

## Why nothing caught it

**The hub's own agents have this covered; contributors don't.**
`nudgeIfTransientAPIError` (`src/pkg/agent/manager.go:5804`, from #4697) matches
this exact string — pinned in `transient_api_error_nudge_test.go:65` — and types
`"try again"` into the pane, capped at 3 nudges per kick with a 90-second
cooldown, with explicit refusals for quota and authorization errors a retry
cannot clear. It is a `Manager` method in the hub's agent package; the
contributor path is the Node relay, which has no equivalent.

**Even a verbatim port would not have helped this session.** That nudge declines
when a human is attached to the tmux session
(`tmuxSessionHasAttachedClientForAgent`, `manager.go:5830`) — deliberately, so a
watchdog never types over a person. The contributor who hit this was attached,
watching. A contributor-side fix needs to decide what to do when someone is
sitting in the pane; "stay silent and report the task complete" is the one
answer that is clearly wrong.

**The relay already does this recovery — for one other backend.** Six lines
above the classifier call, in the same function
(`bin/contributor-relay.sh:1822-1829`):

```js
const hasNetworkError = BACKEND === 'goose' && /Network error:|Please resend your message|Could not connect/i.test(text);
if (hasNetworkError && /> Enter to send/.test(text)) {
  console.log('Goose network error detected — pressing Enter to retry');
  execSync(`tmux send-keys -t ${TMUX_SESSION} Enter`);
  return PANE_STATE_WORKING;
}
```

Goose's transient network error is detected, retried, and reported as
`WORKING`. Claude's is reported as `completed`.

**The stall backstop cannot help.** `PANE_STALL_TIMEOUT_MS` (20 minutes,
`:1898`) only applies to a pane still classified as working. This pane is
classified complete within one progress tick.

## Fix shape

Two separable pieces; the first is the bug, the second is the improvement.

1. **Never classify a pane as complete when the turn ended in a transient API
   error.** Add an `API Error:` check to the claude branch so
   `hasCompletionMarker` cannot be satisfied by a `✻ …ed for …` line that
   follows one. `paneShowsTransientAPIError` (`manager.go:6516`) already encodes
   which API errors are retryable and which (401/403, quota) are not — that
   membership list is the part worth mirroring rather than reinventing, since
   its negative half is what stops a nudge looping against a wall.
2. **Retry it**, along the lines of the goose branch: send `try again`, return
   `WORKING`, bound the attempts, and give up to a real `task_failed`
   (`kind: 'environment'`) rather than a false completion. If the decision is
   that an attached human should be left alone, then the honest fallback is
   `blocked_on_human` with `attention: true` — the relay already has that state
   (`:2208`) and it renders as needing a person, which is exactly true.

Reported from `just contribute-hive claude local` on Linux, backend `claude`,
model `claude-opus-5`, against `hosted-kubestellar-hive-ojv6`, 2026-08-29.

Danathar:
## Summary

- A hive **contributor** runs an agent CLI in a tmux pane; a Node relay
  (`bin/contributor-relay.sh`) watches that pane, decides when the agent has
  finished the assigned issue, and reports the outcome to the hub. It has no
  other way to know — it reads the screen.
- When Claude Code's API connection drops mid-turn it prints
  `API Error: Connection lost mid-response…` and returns to its prompt. **The
  relay read that as the task being finished** — it sent `task_complete`, the
  hub booked a completed task, and one second later handed out the next issue
  while the abandoned issue's half-written work sat uncommitted with no PR.
- Observed live on a hosted hub: `#5061` picked up 11:46:38, booked
  **completed** 11:57:40, no PR. The agent's own next turn opens *"A new task
  arrived mid-flight. I have uncommitted T12 work in the tree — parking it on
  its own local branch."* The agent understood what happened; the relay did not.
- This adds a `TRANSIENT_API_ERROR` pane state, retries the turn in place with a
  bounded `try again`, and never reports success for an errored turn.
- Blast radius: the interactive contributor path. Headless is untouched (it
  drives a one-shot CLI per task and never classifies a pane).

## Root cause

`classifyTmuxPane`, claude branch:

```js
hasCompletionMarker = /[✻✶✽] \S+ed for \d+[ms]|Honking|tokens\)/.test(text);
…
if (hasIdlePrompt && hasCompletionMarker) return PANE_STATE_IDLE_COMPLETE;
```

`✻ Cogitated for 9m 24s` is Claude Code's **turn-duration summary**. It means
*the turn stopped* — printed identically for a turn that succeeded and one that
died on an API error. Nothing in the classifier looked for `API Error:` at all,
and the 20-minute stall backstop never fires because the pane is classified
*complete* within one progress tick, not stalled.

## Change

A turn that ended in a **retryable** API failure now classifies as
`PANE_STATE_TRANSIENT_API_ERROR`, checked **above** the completion test (so an
error can never read as success) and **below** the active-work test (so a CLI
that is streaming, or mid-retry rendering its own countdown, is left alone —
interrupting that would cause the stall this prevents).

The retryable set mirrors `src/pkg/agent/manager.go`'s
`transientAPIErrorPatterns` from #4697, which the hub's own fleet has used for
this exact error since — dropped connection, timeout, 5xx, overloaded. The two
lists are cross-referenced in comments so they stay in step. Authorization (403)
and quota failures **veto** the retry, since repeating those loops the agent
against a wall (#4400, #4583); Claude Code renders every API failure under the
same `API Error:` chrome, so a substring match alone cannot tell them apart.

Recovery is a bounded in-place retry: `try again`, 3 per task, 90s apart.

**It deliberately does not use `tmuxSendKeys`.** That is the task-prompt path
and carries machinery a nudge must not trigger — a `/clear` once context crosses
`CLEAR_CONTEXT_THRESHOLD_PCT`, and the periodic every-N-tasks CLI restart. A
retry exists precisely to preserve the session context that makes recovery
cheap; clearing or restarting would throw away the thing being rescued. It gets
its own minimal sender instead.

Three outcomes, none of which claims success:

| situation | behavior |
|---|---|
| nobody attached, budget remains | type `try again`, report `working` |
| human attached to the pane | type **nothing**, report `blocked_on_human` + `attention` |
| budget exhausted | `task_failed`, `failure_kind: environment` |

The attached-human rule matches the hub-side nudge, which declines when someone
is attached (`manager.go`, `tmuxSessionHasAttachedClientForAgent`) so a watchdog
never types over a person. The difference here is that it *says so* rather than
going quiet and letting the stall backstop fail the task 20 minutes later.

This generalises a precedent already in the same function: `checkTmuxPaneState`
has pressed Enter on a goose network error and returned `WORKING` for some time.
Goose's transient error was retried; claude's was reported completed.

## Testing

- [x] `cd src && go build ./...` — no Go changes, builds clean
- [x] `cd src && go test ./...` — unaffected
- [x] Other: `node bin/contributor-relay.test.js` → **180/180 passed** (171
      pre-existing, all still green, plus 9 new)

**The reproducer was verified against the pre-fix relay.** Swapping in
`origin/v4`'s `contributor-relay.sh` and re-running:

```
FAIL #5094 the relay never reports task_complete for a turn that ended in an API error
     an errored turn must not be booked as a completion
179/180 passed
```

That is the bug, caught by the test, on unmodified `v4`.

The nine new tests cover: the live pane classifying as `TRANSIENT_API_ERROR`; a
genuinely-finished turn still classifying `IDLE_COMPLETE` (a check broad enough
to swallow real completions would be worse than the bug); the detector's
positive and negative tables, including prose that merely mentions a dropped
connection and a request id that looks like a status code; the 403/quota veto;
one retry per tick within budget; the cooldown suppressing a retry on every
2-minute progress tick; the budget resetting per task; and the attached-human
path typing nothing while reporting `blocked_on_human`.

The test harness gains one option (`attachedClients`) so `tmux list-clients` can
be answered both ways.

## Related issues

Fixes #5094

Companion: #5097 covers the *other* route by which a new prompt lands on a busy
agent — a relay restart abandoning the in-flight issue. Different mechanism,
same visible symptom.

## Contributor checklist

- [x] Docs, examples, and policies are updated when behavior changes — the new
      state is internal to the relay's classifier; no doc enumerates pane states.


Danathar:
/hold

Danathar:
Follow-up filed: #5121 records the deliberate residual in this PR — an API error matching neither the retryable nor the known-unretryable list still classifies as `IDLE_COMPLETE`. It was left because a blanket "any \`api error:\` means not-complete" rule can misfire on an agent's own transcript text and fail a genuinely completed turn; the issue lays out that trade-off and some narrower options.

Danathar:
/unhold

The independent review this hold waited on is complete. Review produced three hardening commits since the hold: unretryable (403/quota) errors get their own fatal path instead of falling through to completion, the fatal detector is chrome-gated so an agent quoting a quota phrase in its own summary is not failed, and a mid-session credential expiry (401 / "Please run /login") reports blocked_on_human rather than a false completion. 185/185 relay tests; all-six merge simulation clean.

PR #5107, opened 2026-08-29
🐛 fix: make contributor WebSocket closes diagnosable

### LINKED ISSUE #5090: [v5] 🐛 bug: contributor WebSocket flaps every 30–90s, and every close is silent by construction (1006, no reason, discarded by the relay)
## Summary

- A contributor ("clanker") holds one long-lived WebSocket to the hive hub for
  the whole session: it carries the task assignment, the scoped GitHub token,
  progress reports, and a 30-second application-level ping/pong in both
  directions.
- On `hosted-kubestellar-hive-ojv6` that socket is **not** staying up. Over 32
  minutes on 2026-08-29 one contributor's connection died and re-established
  **seven times**, with connected lifetimes of 76s, 84s, 30s, 991s, 30s, 300s
  and 30s. Three of the seven lasted *exactly* 30 seconds. The relay reconnects
  1s later every time and resumes the task, so work does complete — this is not
  a stall.
- **The cause is not established.** It is not the application heartbeat: both
  the relay (`HEARTBEAT_TIMEOUT_MS = 90000`) and the hub
  (`wsHeartbeatTimeout = 90s`) require 90 seconds of silence before they close
  anything, so neither can produce a 30-second lifetime, and the relay's
  distinctive `Heartbeat timeout on … — reconnecting` line does not appear in
  the log at all. That points below the application — the ingress/network path
  in front of the hosted spoke — but nothing on either side records enough to
  confirm it.
- Which is the second half of this report: **every hub-side close is
  indistinguishable from a network drop, and the relay throws away the only
  evidence there is.** Both are one-line fixes and both are prerequisites for
  diagnosing the first half.
- Blast radius: every contributor on a hosted hub. Work still completes, but
  each flap books a release cooldown on the in-flight issue, re-mints a GitHub
  token, and writes a `joined`/`left` pair into the shared public activity feed.

## Evidence

`GET /api/contribute/activity`, filtered to one contributor, 2026-08-29:

```
11:12:52  joined     ┐ 76s
11:14:08  left       ┘
11:15:22  joined     ┐ 84s
11:16:46  left       ┘
11:17:18  joined     ┐ 30s
11:17:48  left       ┘
11:22:02  joined     ┐ 991s
11:38:33  left       ┘
11:38:34  joined     ┐ 30s      (+1s reconnect = BASE_RECONNECT_DELAY_MS)
11:39:04  left       ┘
11:39:05  joined     ┐ 300s
11:44:05  left       ┘
11:44:06  joined     ┐ 30s
11:44:36  left       ┘
```

19 `joined` and 19 `left` events for this one contributor in the retained feed.
The `+1s` gaps are the relay's `BASE_RECONNECT_DELAY_MS`; the backoff never
grows past 1s because a successful connect resets it
(`bin/contributor-relay.sh:2318`), so the log line is always
`Reconnecting in 1000ms`.

Matching relay output, one full cycle, repeated verbatim four times:

```
Connected to wss://hosted-kubestellar-hive-ojv6…/api/contribute/ws
Authenticated … as c-ca8187f7480c (tier: contributor)
Hub protocol 1.2; capabilities: token_refresh, …
Reconnected while working on kubestellar/hive#5055 — resuming
GitHub token refreshed
Connection to wss://…/api/contribute/ws closed. Reconnecting in 1000ms...
```

## Why nobody can tell what closed it

**1. The hub never sends a WebSocket Close frame.** Every hub-side close is a
bare `conn.Close()` / `c.ws.Close()` — the auth timeout
(`contribute_ws.go:2766`), the auth-failure paths (`:2845`, `:2861`, `:2867`),
the heartbeat loop's timeout and failed-ping branches (`:3639`, `:3647`), and
the stale sweep in `cleanupLoop` (`:3929`). None calls
`WriteControl(websocket.CloseMessage, …)`, so the client sees **1006 (abnormal
closure), empty reason** — byte-identical to a yanked cable.

Verified live against the hub with an unauthenticated probe:

```
 0.1s open
 0.1s msg: {"type":"auth_challenge","seq":1,"nonce":"…"}
30.1s msg: {"type":"auth_failed","reason":"Authentication timeout"}
30.1s CLOSE code= 1006 reason= ""
```

The hub knew exactly why it was hanging up, said so in a JSON message, and then
closed the socket in a way that carries none of it. A JSON `auth_failed` is at
least readable; the heartbeat and cleanup closes send nothing at all first.

**2. The relay discards the close code and reason.**

```js
hub.ws.on('close', () => {
  console.log(`Connection to ${hub.url} closed. Reconnecting in ${hub.reconnectDelay}ms...`);
```
(`bin/contributor-relay.sh:2553`)

`ws` hands that callback `(code, reason)`. Logging them would immediately
separate "1006, no frame" (network or a bare `Close()`) from a deliberate
1000/1001/1011 with text — which is the single fact needed to decide whether
this is the ingress or the hub, and it is currently thrown away on every flap.

## What each flap costs

- **The in-flight issue is briefly withdrawn.** The disconnect path calls
  `bookReleaseCooldown` (`contribute_ws.go:2814`) so the issue is not instantly
  re-admissible — correct as designed (it closes the #2356 duplicate-PR race),
  but it means a flapping session repeatedly pulls its own issue out of the
  ready-work queue.
- **A GitHub token is re-minted every cycle** — `GitHub token refreshed` appears
  in all four logged reconnects, against a `wsTokenRefreshPeriod` of 50 minutes.
- **The shared activity feed fills with churn.** `joined`/`left` are ordinary
  activity entries, so one flapping contributor crowds real `picked up` /
  `completed` rows out of the operator's view and out of `/api/contribute/activity`.
- **`fleet_size` samples become a coin flip** — the hourly metric reads 1 or 0
  depending on which side of a flap the sample lands.
- Each dial re-occupies a slot against the hub-wide `maxWSConnections = 50`.

## Suggested order of work

1. Log `(code, reason)` in the relay's close handler. One line, no behavior change.
2. Have hub-side closes send a real Close frame with a code and a reason before
   closing — at minimum the heartbeat/cleanup paths, which today say nothing at all.
3. With 1 and 2 in place, one flap tells you whether the hub closed it (and why)
   or something in the ingress path did. Fix that.

Reported from a `just contribute-hive claude local` session on Linux against
`hosted-kubestellar-hive-ojv6`, 2026-08-29, working `#5055`.

Danathar:
## Summary

- A contributor holds one long-lived WebSocket to the hub for its whole session —
  it carries the task assignment, the scoped GitHub token, and progress reports.
- On a hosted hub that socket flaps: one contributor's connection died and
  re-established **seven times in 32 minutes** (lifetimes 76s, 84s, 30s, 991s,
  30s, 300s, 30s). Work still completes, since the relay reconnects in 1s and
  resumes — but nobody can say *why* it keeps dropping, because **neither side
  records enough to tell a deliberate hangup from a cut cable**.
- The hub closed every socket with a bare `conn.Close()`, which sends no
  WebSocket Close frame, so the client sees code `1006` with an empty reason. The
  relay then threw away even that. Two silences stacked.
- This PR fixes the diagnosability, not the flapping. **It does not change when
  connections close** — only what is knowable about it. That is deliberate: the
  cause is not yet established, and it cannot be until this lands.

## The two silences

**1. The hub never sent a Close frame.** Verified live against a running hub with
an unauthenticated probe:

```
 0.1s open
 0.1s msg: {"type":"auth_challenge","seq":1,"nonce":"…"}
30.1s msg: {"type":"auth_failed","reason":"Authentication timeout"}
30.1s CLOSE code= 1006 reason= ""
```

It knew exactly why it was hanging up, said so in a JSON message, then closed in
a way that carried none of it. And the JSON is not a substitute: the heartbeat
timeout and the stale-connection sweep send no message at all, and a client that
has stopped reading application messages — precisely what a heartbeat timeout
describes — could never have read one.

**2. The relay discarded the code and reason.** `ws` hands the close handler
`(code, reason)`; the handler took no arguments and logged
`closed. Reconnecting in 1000ms...`. The delay carries no signal either — a
successful connect resets the backoff, so it reads `1000ms` every time.

## Change

**Hub** — a `closeWithReason` helper writes a real Close frame before closing,
applied to every deliberate close on the contributor path:

| path | code |
|---|---|
| auth timeout, missing/invalid/revoked token, model rejected, role rejected, operator disconnect | `1008` policy violation |
| heartbeat timeout, stale-connection sweep | `1001` going away |

`WriteControl` is documented as safe to call concurrently with all other
methods, so this needs no coordination with the `writeMu` the JSON senders take.
Both the frame write and the close remain best-effort — a peer that already
vanished simply fails the write, which is not worth logging on a socket being
discarded.

Left alone deliberately: the deferred `conn.Close()` after the read loop exits
(the peer is already gone) and the close after a failed heartbeat *ping* (the
write just failed, so a frame cannot get out either).

**Relay** — the close handler logs `describeWsClose(code, reason)`. `1006` gets
called out explicitly as *"no close frame; the socket was cut (network, proxy, or
an abrupt peer close)"*, because it is the one code never sent on the wire — the
client synthesises it when the peer vanished without a frame. Separating that
from a stated reason is the entire diagnostic.

## What this buys

After this, one flap answers the question #5090 could not:

- `1006` still → the socket is being cut below the application. Next stop is the
  ingress in front of the hosted spoke.
- `1001 going away: heartbeat timeout…` → the hub is closing it, and the reason
  names which path.

## Testing

- [x] `cd src && go test ./...` — `pkg/dashboard` suite green (49s), `go vet` and
      `gofmt` clean
- [x] Other: `node bin/contributor-relay.test.js` → **175/175** (171
      pre-existing, all green, plus 4 new)

**Both Go tests were confirmed to fail on unmodified `v4`**, with exactly the
reported symptom:

```
--- FAIL: TestWS_AuthFailureClosesWithAStatedReason
    the hub closed without a close frame (1006) — a deliberate refusal is
    indistinguishable from a network drop (#5090)
--- FAIL: TestWS_MissingTokenClosesWithAStatedReason
    missing-token refusal closed with code=1006 text="unexpected EOF";
    want a policy violation with a stated reason
```

They read the **close frame**, not the JSON message — the frame is the part that
was missing and the only part a client that stopped reading can still observe.
The four relay tests cover the `1006` wording, a coded close carrying its
server-stated reason through to the log, an unknown code rendering by number
rather than `undefined`, and a reasonless normal closure not leaving a dangling
separator.

## Related issues

Refs #5090 — deliberately not `Fixes`: this PR is the instrumentation, and
the flapping itself stays open until the cause is found. (Historical note: this
line originally read `Fixes` by mistake, which auto-closed #5090 on merge; it
was reopened.)

The flapping itself remains open and undiagnosed; this is the instrumentation
that makes diagnosing it possible.

## Contributor checklist

      specifies close codes; wire behavior is additive (a frame where there was
      none), and `IsUnexpectedCloseError` on the read path already tolerates it.


Danathar:
/hold

Danathar:
/unhold

The independent review this hold waited on is complete. Review produced one fix since the hold: the stale-connection sweep now closes sockets outside the hub-wide mutex (the close frame's write deadline made the old in-lock close a potential ~1s-per-socket stall). Tests moved to a distinct slot so this merges cleanly alongside #5106.

Issue #5058, opened 2026-08-29
✨ tui T9: tokens pane (per-agent + total)


Danathar:
Part of #4907. Depends on T3 (#5004, merged as #5035) and T8.

**Files to create/modify:** `src/pkg/tui/panes/tokens.go` and its golden testdata only.

**Build this:** Replace the stub. The pane holds the T8 token-usage data, updated via a `panes.TokensMsg` (define it here; the app sends it in T12/T13b). Render per the design sketch: one row per agent (agent, in, out, cost), a separator, and a totals row. Human-readable magnitudes (1.2M, 88.1k). Content only, per the T3 contract.

**Acceptance criteria:**
- [ ] Golden test at 48x12 with a multi-agent fixture, `panes/testdata/tokens.golden`
- [ ] `go build ./...` and `go test ./pkg/tui/...` pass (from `src/`)

**Out of scope:** fetching, cost attribution beyond what `/api/tokens` returns.

**Verify with:** `cd src && go build ./... && go test ./pkg/tui/...`

---
Wave 3 of the #4907 filing protocol — unblocked by T3 (#5035, merged). Mirrors the T5 pane pattern.

Issue #5057, opened 2026-08-29
✨ tui T8: client — token usage


Danathar:
Part of #4907. Depends on T2 (#4917, merged as #4937). Read `src/docs/design/tui.md` first.

**Files to create/modify:** `src/pkg/tui/client/tokens.go`, `client/tokens_test.go`.

**Build this:** From `dashboard/openapi.json`, take the `/api/tokens` operation (published — design doc §2.1). Define token-usage types with json tags mirroring the spec's schema: per-agent input/output token counts and cost, plus totals as exposed. Mirror the spec, do not invent fields. Add `func (c *Client) Tokens(ctx context.Context) (...)` using `getJSON`.

**Acceptance criteria:**
- [ ] Fixture JSON in `client/testdata/tokens.json` matching the spec schema exactly
- [ ] Test asserts decode of all fields and the error path on 500
- [ ] `go build ./...` and `go test ./pkg/tui/...` pass (from `src/`)

**Out of scope:** rendering (T9).

**Verify with:** `cd src && go build ./... && go test ./pkg/tui/...`

---
Wave 3 of the #4907 filing protocol — unblocked by T3 (#5035, merged). Mirrors the T4 client pattern.

Issue #5056, opened 2026-08-29
✨ tui T7: governor pane (mode, queue depth, next eval)


Danathar:
Part of #4907. Depends on T3 (#5004, merged as #5035) and T6.

**Files to create/modify:** `src/pkg/tui/panes/governor.go` and its golden testdata only.

**Build this:** Replace the stub. The pane holds a `client.GovernorStatus`, updated via a `panes.GovernorMsg` (define it here; the app sends it in T12/T13b). Render label/value rows per the design sketch: mode, queue depth, next eval (relative), eval interval, ACMM level. Content only — the app owns borders and focus, per the T3 contract.

**Acceptance criteria:**
- [ ] Golden test at 48x12 with a populated fixture, `panes/testdata/governor.golden`
- [ ] `go build ./...` and `go test ./pkg/tui/...` pass (from `src/`)

**Out of scope:** fetching, the ACMM overlay (T19).

**Verify with:** `cd src && go build ./... && go test ./pkg/tui/...`

---
Wave 3 of the #4907 filing protocol — unblocked by T3 (#5035, merged). Mirrors the T5 pane pattern.

PR #5108, opened 2026-08-29
🐛 fix: record released contributor tasks in the activity feed

### LINKED ISSUE #5097: 🐛 bug: restarting the contributor relay abandons the in-flight issue and is guaranteed to hand out a different one

Danathar:
## Summary

- When a contributor picks up an issue and then *releases* it without finishing —
  the socket dropped mid-task, or the relay asked for new work while still
  holding one — the hub correctly releases the issue and books its cooldown, but
  **recorded nothing an operator could see**.
- The activity feed showed a `picked up` entry with no terminal event ever
  following it. That is indistinguishable from an issue nobody touched.
- Measured on a live hub: a contributor restarting its relay four times in ten
  minutes touched four different issues (#5056, #5058, #5057, #5055) and
  completed none. Three were abandoned mid-implementation, and the hub's history
  recorded nothing for any of them — "this contributor keeps dropping work" was
  inferable only by noticing an absence.
- Both release paths now add an entry naming the task. Blast radius: two extra
  activity entries. No change to the ledger, the cooldowns, or assignment.

## Scope — and a correction to the issue

#5097 proposed two halves. This PR is the second one only, and I want to be
explicit that **the first half as I originally described it was wrong.**

The issue suggested a restarted relay should read `TASK_FILE` and resume its
in-flight task instead of asking for new work. Checking the restart path shows
that would claim work nobody is doing:

- `Justfile:942` mints a **new random tmux session name** on every
  `contribute-hive` run (`hive-${BACKEND}-$(head -c 2 /dev/urandom …)`), and
- the recipe's cleanup trap runs `tmux kill-session` on exit.

So a restarted relay always faces a **fresh CLI with no memory of the task**. The
agent's context is genuinely gone; resuming would re-assert ownership of work
that has stopped existing, which is worse than being handed something new. A
real cross-restart resume needs the session to survive the relay, which today it
does not — that is a design change, not a bug fix, and it is not attempted here.

I have corrected the issue body accordingly.

The second half stands on its own and was always independent: as the issue put
it, a `picked up` with no terminal event *"is a hole in the audit trail, and it
is the same hole whether the fix above lands or not."*

One more correction: the issue said the feed's "entire action vocabulary" is
`completed / joined / left / picked up`. That was the observed data, not the
code — `failed`, `lease expired: auto-released`, `reassigned by yank` and others
all exist and simply had not occurred in the retained window. The actual gap was
narrower and is what this fixes: the two **release** paths recorded nothing.

## Change

Both sites now call `addActivity` with a task description:

| path | entry |
|---|---|
| socket dropped while holding a task (`contribute_ws.go`, the disconnect defer) | `released: connection lost` |
| relay sent `ready` while still holding one | `released: gave the task back` |

`taskDescOf` formats the task exactly as the `picked up` and `completed` entries
do, so a release lines up with the pickup it closes out rather than being
described differently. A synthetic pr-review task carries `Number == 0` and falls
back to its task id — rendering it as `#0` would read as a real issue and collide
with every other numberless task in the feed.

**The verb is `released`, not `failed`, and that is load-bearing.** #4260
established that a dropped socket is not a failure of the work — booking it as
one is what turned three dropped sockets on a single issue into a quarantine of
an issue nobody had failed. The ledger behavior is untouched
(`bookReleaseCooldown` still books a release); this makes the feed say what the
ledger already says.

## Testing

- [x] `cd src && go test ./...` — `pkg/dashboard` suite green (49s), `go vet` and
      `gofmt` clean
- [x] Other: n/a

Three tests in `contribute_release_visibility_test.go` cover the description
format matching the pickup entry, the synthetic/nil task fallbacks (explicitly
asserting nothing renders as `#0`), and a guard that both release sites keep the
`released` verb rather than drifting to `failed` — the edit that would quietly
re-introduce the #4260 quarantine behavior.

## Related issues

Fixes #5097

Companion: #5094 covers the other route by which an in-flight issue is dropped —
a transient API error being read as a completion. Different mechanism, same
visible symptom of work going missing.

## Contributor checklist

      enumerates activity verbs; the feed is additive and consumers render
      whatever action string arrives.


Danathar:
/hold

Danathar:
Follow-up filed: #5120 covers the same `Number`-derived identity assumption in the four pre-existing activity-description sites (`picked up`, yank/reassign, `completed`, `failed`), which this PR deliberately left alone to keep its scope to the release entries. That issue reuses the `taskDescOf` helper introduced here.

Danathar:
/unhold

The independent review this hold waited on is complete. Review produced one fix since the hold: released tasks are described via the source-aware identityKey(), so Linear/Jira items keep their canonical key instead of collapsing to an internal task id. The matching pre-existing gap in the picked-up/completed/failed entries is filed separately as #5120.

Issue #5103, opened 2026-08-29
🐛 bug: contribute-k8s generates a workload with no agent-CLI credential — all five allow-listed headless backends launch unauthenticated


Danathar:
## Summary

- `just contribute-k8s` generates a Kubernetes workload that runs a hive
  contributor headlessly in a cluster — a Deployment plus the ConfigMap and
  Secret it reads its configuration from. It is the supported way to contribute
  from a cluster instead of a laptop.
- **The generated workload carries no credential for the agent CLI it is told to
  run.** It authenticates to the hub and to GitHub, then launches a backend that
  has nothing to authenticate with. On `claude` — the default, and the backend
  the path's own test script exercises — the CLI comes up unauthenticated and
  can do no work.
- This is not a claude-specific gap. The generator allow-lists five headless
  backends and plumbs a credential for none of them.
- Found while investigating #5088 (container-mode logins being discarded). That
  issue asked whether any host-less environment needs a credential written back
  to the host; this path is host-less, but it never touches the staging-dir
  mechanism at all, so it needs its own plumbing rather than anything #5088
  could provide. Filing separately so it is findable on its own terms.
- Blast radius: `just contribute-k8s` only. No laptop path is affected — local
  and container mode both mount or stage a real credential.

## What the generated workload actually gets

Verified against `v4` at `fad4ccc9`. The pod's whole environment comes from two
`envFrom` sources (`Justfile:1763-1766`) — the ConfigMap and the Secret, both
pulled wholesale, so this list is exhaustive:

ConfigMap `data:` (`Justfile:1665-1670`):

```
HIVE_HUB
CONTRIBUTOR_ID
CONTRIBUTOR_USERNAME
AGENT_BACKEND
CONTRIBUTOR_MODE
HIVE_HEADLESS_STATUS_FILE
```

Secret `data:` (`Justfile:1688-1690`):

```
HIVE_REGISTRATION_TOKEN
GH_TOKEN
```

Eight keys. `HIVE_REGISTRATION_TOKEN` authenticates the relay to the hub;
`GH_TOKEN` lets it push and open PRs. Nothing authenticates the agent CLI.

There is also no volume mount of `~/.claude`, `~/.codex`, `~/.copilot` or
`~/.config/goose`, and no provider-env passthrough — the container path builds
`PROVIDER_ENV_ARGS` and forwards the selected provider's variables into
`docker run`, but the k8s generator has no equivalent.

## Which backends this affects

`Justfile:1592`:

```bash
[...]

Issue #5088, opened 2026-08-29
🐛 bug: a Claude login done inside the contributor container is discarded on exit — every container run restarts from the host's stale credential


Danathar:
## Summary

- `just contribute-hive` starts a contributor agent two ways: **container mode**
  (the default — the backend CLI runs inside a throwaway container) and **local
  mode** (`... claude local` — the CLI runs as you, on your machine). Claude
  Code's OAuth login expires every so often and has to be redone interactively
  in the CLI.
- In local mode, re-logging in works and sticks. In container mode it *looks*
  like it works — you attach to the container's tmux session, run through the
  normal login, and the CLI comes back authenticated — but quit and start a
  second `contribute-hive` run and you are back at the login menu. The login you
  just completed was never saved anywhere that outlives the run.
- Cause: the container never sees your real `~/.claude`. `contribute-hive`
  copies it into a per-run `mktemp -d` staging directory, bind-mounts *that*
  read-write, and `rm -rf`s it in the `EXIT` trap
  (`Justfile:1227`,
  `Justfile:1373`). The refreshed token the
  in-container login wrote lands in the staging copy and dies with it; the next
  run re-stages from the host's still-expired `~/.claude`.
- The discard is deliberate — it is the H6 / CWE-668 containment that stops a
  `--dangerously-skip-permissions` agent from writing to your real CLI config —
  but it was designed for the case where the *host* credential is valid. When it
  is not, the only place the contributor can log in is inside the container, and
  that login is exactly what gets thrown away. There is no message saying so.
- Blast radius: container mode. Observed with the `claude` backend; the same
  ephemeral-staging shape covers copilot, codex, goose, gemini, pi and opencode,
  so any backend whose credential can expire mid-life has the same hole. Local
  mode is unaffected. No fix is proposed below — the containment trade-off is a
  maintainer call.

## Reproduce

Requires a Claude login that has already expired (or force one by moving
`~/.claude/.credentials.json` aside).

1. `just contribute-hive claude` — container mode, the default.
2. Attach to the agent's tmux session when the recipe prints the attach command.
3. Claude Code is sitting at the login menu. Complete the login as normal. It
   succeeds; the CLI is authenticated and the agent works.
4. Quit / detach and let the run end.
5. `just contribute-hive claude` again, attach again.

Observed: back at the login menu, as if step 3 never happened.
Expected (as a user): the login I just completed carries into the next run, the
way it does in local mode.

[...]

Issue #2538, opened 2026-08-04
contributor: Kubernetes needs headless relay support before a workload manifest is useful


castrojo:
## Status update — 2026-08-04

**Prerequisite confirmed; no headless relay exists yet.** The maintainer response agrees that a workload manifest is not the unit of work: the contributor still launches an interactive CLI in tmux and receives tasks through `tmux send-keys`, so Kubernetes packaging alone would create a healthy-looking pod stalled on an invisible prompt.

Merged #2557 improves the interactive path by creating a real workspace, giving the agent an explicit clone instruction, and releasing a held slot on abandon. It does not add non-interactive task delivery or remove the tmux dependency. #2549 therefore remains downstream of this issue, and the cluster credential shape remains tied to #2537 rather than a long-lived personal PAT in a Secret.

> Part of #2534 (contributor onboarding field notes from `projectbluefin/review`).

### Summary

We operate [`projectbluefin/review`](https://github.com/projectbluefin/review), a downstream
consumer of the Hive contributor protocol, and we went looking for a way to run a contributor
on a cluster instead of a workstation. `just contribute-k8s` generates a `Namespace`, a
`ConfigMap`, and a `Secret` and prints an apply instruction — it does not run `kubectl` itself
— and we could not find a contributor `Deployment`, `Job`, `CronJob`, `Pod`, `StatefulSet`, or
`DaemonSet` anywhere in the repo, so there is currently nothing in-tree that would consume that
config. Digging into why, we think the interesting blocker is not missing YAML: the relay drives
an interactive tmux pane, so **headless relay support looks like a prerequisite** and a workload
manifest on its own would not be sufficient.

We are raising this as an opportunity, not a defect report. The recipe's own doc comment is
accurate about its scope ("Generate K8s ConfigMap + Secret YAML from contributor.env",
[`Justfile:665`](https://github.com/kubestellar/hive/blob/0e5d7a1ffe75ac7ef3064e2921290489f55da64b/Justfile#L665)),
and we found no place where the project claims a complete Kubernetes contributor path.

### Evidence

**What `contribute-k8s` does today.**

- Recipe signature `contribute-k8s namespace="hive-contributor" outfile=""` —
  [`Justfile:669`](https://github.com/kubestellar/hive/blob/0e5d7a1ffe75ac7ef3064e2921290489f55da64b/Justfile#L669),
  documented at `:665-668`.
- It assembles YAML into a shell variable and emits exactly three objects: `kind: Namespace`
[...]

PR #5109, opened 2026-08-29
🐛 fix: warn when the container has no usable Claude credential, and fail fast headless

Danathar:
## Summary

- `just contribute-hive` (container mode, the default) hands the agent CLI a
  **copy** of `~/.claude` in a throwaway staging directory that is deleted when
  the command exits. That is deliberate containment — it keeps a
  permissions-bypassed agent away from your real credential (H6 / CWE-668) — and
  **this PR does not change it.**
- What it also did, silently, was discard a login performed *inside* the
  container. A contributor whose host credential had expired reached the CLI's
  login menu, completed the whole browser flow, worked a session, and was back at
  the login menu on the next run with nothing to show for it.
- The mechanism was working as designed. The design assumed the host credential
  was valid, and said nothing when it wasn't — so the one action a contributor
  has to take when it isn't was exactly the action being thrown away.
- Interactive runs now warn before the CLI starts. Headless runs fail fast.
  Blast radius: one check and a message on the `claude` staging path. Nothing is
  copied back out of the container.

## Scope

#5088 raised the question of whether the login should be made to *persist* —
copying `.credentials.json` back out. That is **not** this PR, and after checking
I don't think it should be built yet.

The condition for building it was naming a concrete host-less environment that
the mechanism would serve. The only host-less deployment is `just contribute-k8s`,
and it never touches the staging path at all — its problem is that the generated
workload carries no agent-CLI credential of any kind, which needs Secret
plumbing rather than a copy-back. I filed that separately as **#5103**. The
ephemeral cases (CI runners, disposable boxes) fail from the other direction:
persisting a credential back to a host that evaporates buys nothing.

So the beneficiary set for a copy-back is currently empty and the containment
stays as it is. This PR makes the existing behavior *audible* instead.

## Change

After staging `~/.claude`, the recipe asks whether the staged credential can
actually authenticate:

- **Interactive** — warn, stating plainly that an in-container login *works but
  lasts for this run only*, and that the way to keep one is `claude` on the host
  once. The message deliberately does **not** say the login will fail: it
  succeeds, and overstating it is wrong in a way a contributor catches
  immediately and then stops trusting the warning.
- **Headless** (`CONTRIBUTOR_MODE=headless`) — `exit 1` with the same
  instruction. A pod has no human to answer a login prompt, so warning and
  continuing leaves precisely the healthy-looking-but-stalled workload #2538
  exists to prevent.

`claude_staged_credential_usable` mirrors `pkg/claude`'s `ReadAccessToken` rule —
a `claudeAiOauth` block with a non-empty `accessToken`, not past its `expiresAt` —
with one deliberate addition: **an expired access token that still carries a
`refreshToken` counts as usable**, because Claude Code refreshes it silently and
no prompt ever appears. Warning there would be crying wolf and would train
contributors to ignore the warning. Without `jq` the check stays silent rather
than guessing, so a machine lacking it never sees a false warning.

Expiry is compared in milliseconds built from `date +%s`, not `%3N`, which
BSD/macOS `date` does not support.

## Testing

- [x] `cd src && go test ./...` — `pkg/dashboard` suite green (49s), `gofmt`/`go vet` clean
- [x] Other: `just --summary` parses the modified Justfile

Three of the four new tests were confirmed to **fail** against unmodified `v4`.

The predicate itself was exercised directly across all eight cases it can meet:

| credential | verdict |
|---|---|
| valid, unexpired | usable |
| expired, no refresh token | **warn** |
| expired, with refresh token | usable |
| no `expiresAt` field | usable |
| empty `accessToken` | **warn** |
| no `claudeAiOauth` block | **warn** |
| unparseable JSON | **warn** |
| file absent | **warn** |

The Go tests pin the Justfile invariants the way
`contribute_local_mode_warning_test.go` already does for the local-mode posture
message: that the staged (not host) path is what gets checked, that the headless
branch exits rather than continuing, that the warning names the cost and the
fix, and that the refresh-token tolerance is not quietly dropped by a later edit.

## Related issues

Refs #5088 — deliberately `Refs`, not `Fixes`. This closes the silent-failure
half; whether an in-container login should be made to persist is the open
question, and #5103 covers the k8s credential gap found while scoping it.

## Contributor checklist

- [x] Docs, examples, and policies are updated when behavior changes — the new
      headless refusal is a startup precondition; no doc enumerates them.


Danathar:
/hold

Danathar:
/unhold

The independent review this hold waited on is complete. Review produced two fixes since the hold: a set ANTHROPIC_API_KEY now satisfies the credential gate (it is forwarded into the container, so refusing it was wrong — worst on headless, which exited for a run that would have worked), and the headless refusal removes the staged ~/.claude copy before exiting instead of leaking it in /tmp.

PR #5140, opened 2026-08-29
✨ feat(tui): parse SSE event stream

### LINKED ISSUE #5062: ✨ tui T13a: SSE reader — parse the stream into typed msgs
Part of #4907. Depends on T2 (#4917, merged as #4937). Read `src/docs/design/tui.md` first.

**Files to create/modify:** `src/pkg/tui/client/sse.go`, `client/sse_test.go`.

**Build this:** An SSE reader for the dashboard's event stream: connect with the client's base URL + auth header, parse the `text/event-stream` framing (event/data lines, multi-line data, comments/keepalives), and deliver typed messages on a channel the app can consume (T13b wires it). Handle partial reads and connection close cleanly — the caller decides on reconnect. `pkg/hivectl`'s existing `StreamSSE` is worth reading first (noted in T2's PR, #4937): reuse or adapt rather than re-derive, and record the decision either way.

**Acceptance criteria:**
- [ ] httptest fixture streaming: multiple events, multi-line data, keepalive comments, and mid-event connection close — all asserted
- [ ] Auth header asserted on the stream request
- [ ] `go build ./...` and `go test ./pkg/tui/...` pass (from `src/`)

**Out of scope:** wiring into the app and poll fallback (T13b), reconnect policy.

**Verify with:** `cd src && go build ./... && go test ./pkg/tui/...`

---
Wave 3 of the #4907 filing protocol — unblocked by T3 (#5035, merged).

Danathar:
Closes #5062

## Summary

- add an asynchronous StreamEvents client that delivers framed SSEEvent values and terminal errors on channels
- parse default and named events, multi-line data, CRLF, keepalive comments, fragmented reads, clean closes, and incomplete frames
- authenticate the long-lived stream without applying the polling client's five-second whole-response timeout
- validate JSON while keeping payloads lossless for T13b to decode into pane-specific message types

## Reuse decision

The implementation adapts the request and authentication pattern from hivectl.Client.StreamSSE. That helper writes JSON-wrapped raw frames to an io.Writer, so reusing it directly would force the TUI to decode an intermediate representation before it could consume typed channel messages.

## Verification

- go build ./...
- go test ./pkg/tui/...
- go test ./pkg/tui/client -run TestStreamEvents -count=20

— hive: backend=codex

### DIFF
--- a/src/pkg/tui/client/sse.go
+++ b/src/pkg/tui/client/sse.go
@@ -1,0 +1,181 @@
+package client
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+)
+
+const sseEventsPath = "/api/events"
+
+// SSEEventType is the value of an SSE event field. A frame without one is a
+// standard message event.
+type SSEEventType string
+
+const (
+	SSEEventTypeMessage     SSEEventType = "message"
+	SSEEventTypeAgentStatus SSEEventType = "agent-status"
+)
+
+// SSEEvent is one completely framed dashboard SSE message.
+//
+// Data is validated JSON but stays raw so T13b can decode a full status event
+// and an agent-status event into their distinct pane delivery types without
+// this transport layer depending on every pane's model.
+type SSEEvent struct {
+	Type SSEEventType
+	Data json.RawMessage
+}
+
+// Decode unmarshals an event payload into the caller's typed snapshot.
+func (e SSEEvent) Decode(v any) error {
+	if err := json.Unmarshal(e.Data, v); err != nil {
+		return fmt.Errorf("decode %s event: %w", e.Type, err)
+	}
+	return nil
+}
+
+// StreamEvents connects to the dashboard's SSE endpoint and returns channels
+// for complete events and the first terminal stream error.
+//
+// Both channels close when the server closes cleanly or ctx is cancelled. An
+// EOF in the middle of an event is reported as io.ErrUnexpectedEOF; no partial
+// event is delivered. Reconnection belongs to the caller (T13b).
+//
+// This adapts rather than reuses hivectl.Client.StreamSSE. That helper has the
+// correct request/auth pattern, but its public contract writes JSON-wrapped raw
+// lines to an io.Writer. The TUI needs parsed event/data framing on a channel,
+// so sharing it would require parsing an encoded intermediate representation.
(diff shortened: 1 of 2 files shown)


Danathar:
Pushed a fix for the failing `coverage` gate (`client: 89.2%` against its 90% floor).

`pkg/tui/client` is now at **97.0%**. I aimed well clear of the floor rather than a tenth of a point over it, partly for margin and partly because of the discrepancy below.

Every added test covers a branch the existing four did not reach, and each is a path a real dashboard produces:

- **`Decode`, both arms.** It is the seam T13b uses to turn a raw event into a pane's snapshot type, so the failure arm matters — a status event whose shape has drifted must name the event type it failed on.
- **A non-2xx answer to the stream request** (401/404/502), which must surface as the same `*APIError` the polling calls produce, so a caller can tell a bad token from a dead API behind a live proxy.
- **No `Authorization` header when no token is set**, mirroring the polling client's equivalent: an empty `Bearer ` is not the same as absent.
- **An unbuildable request** from a malformed `HIVE_DASHBOARD_URL`. `New()` cannot fail by contract, so the stream is where that finally reports.
- **`readEventStream` framing edge cases**, driven directly rather than over HTTP because a server per case would test nothing more: comment lines, an empty `event:` value resetting the type (so the next event cannot inherit the previous one's), a field line with no colon, unknown fields, a blank line with no data dispatching nothing, and CRLF endings.
- **An over-long line failing the read** rather than truncating silently — half an event decoded as if whole would be worse than no event.
- **A cancelled context** reporting `ctx.Err()` rather than a stream error, and **an emit failure** propagating unchanged (the path `StreamEvents` uses to stop when its consumer is done).

Two branches are deliberately left uncovered and I would rather say so than chase them: `newTransport`'s non-`*Transport` fallback would need `http.DefaultTransport` swapped globally, which is not safe alongside parallel tests, and `streamEvents`' Do-error-under-cancellation race has no deterministic trigger.

**One thing I could not explain, flagged rather than hidden.** I could not reproduce the 89.2% locally: `go test -cover` and the gate's own awk, run against the same commit (`c8381b0c`, which contains `v4` as an ancestor, so the merge tree is identical), both reported **90.4%** — including with the shards' `-short -race` flags. The 1.2-point gap between my measurement and CI's is unaccounted for. It does not change the fix, and 97% clears the floor either way, but if the gate is systematically reading ~1 point lower than `go test -cover` for this package that is worth knowing independently of this PR.

Verified with `-race` at `-count=3`; the new stream tests do not flake.

Issue #5136, opened 2026-08-29
✨ tui T18: client — ACMM level get/apply


Danathar:
Part of #4907. Read `src/docs/design/tui.md` first (after T0 lands). Depends on T2 (#4917).

**Files to create/modify:** `src/pkg/tui/client/acmm.go`,
`client/acmm_test.go`.

**Build this:** From `dashboard/openapi.json`, take the operation the web
dashboard uses for reading and applying the ACMM level. Define `type ACMMStatus struct` with json tags
matching the spec's schema — mirror the spec, do not invent fields. Add
`func (c *Client) ACMM / ApplyACMM(ctx context.Context) (ACMMStatus, error)` using
`getJSON`. If the spec has no such operation, stop and file a
follow-up issue titled `dashboard/openapi.json missing reading and applying the ACMM level operation`
with what you found; that filing completes this task (see #4912 for known
spec gaps).

**Acceptance criteria:**
- Fixture JSON in `client/testdata/acmm.json` constructed to match the schema exactly; test asserts decode of all fields and the error path on 500
- `go build ./...` and `go test ./src/pkg/tui/...` pass

**Out of scope:** the overlay UI (T19).

**Verify with:** `go build ./... && go test ./src/pkg/tui/...`

---
Filed per the #4907 filing protocol: this task's dependencies have merged, so it is now unblocked work.

---
Wave 4 of the #4907 filing protocol — unblocked by T2 (#4917, closed).

clubanderson:
Closed by #5209, which restored this work after I destroyed the original PR (#5154).

**Attribution correction:** the squash-merge commit `b43ad76d` is attributed to `Andy Anderson`, not to @Danathar, who actually wrote this. The branch commit carried his authorship correctly (`Danathar <doug.baggett@gmail.com>`), but the squash used my push identity.

For the record: `src/pkg/tui/client/acmm.go` (213 lines), `acmm_test.go` (351), `testdata/acmm.json` (91) and the `client.go` wiring are @Danathar's work, recovered verbatim from commit `a98963f5`. I contributed nothing to it except losing it and getting it back.

[...]

PR #5152, opened 2026-08-29
✨ feat(tui): client — pause/resume agent (#4907 T14)

### LINKED ISSUE #5134: ✨ tui T14: client — pause/resume agent
Part of #4907. Read `src/docs/design/tui.md` first (after T0 lands). Depends on T4 (#5053).

**Files to create/modify:** `src/pkg/tui/client/actions.go`,
`client/actions_test.go`.

**Build this:** From `dashboard/openapi.json`, take the operation the web
dashboard uses for pausing and resuming an agent. Define `type AgentActionResult struct` with json tags
matching the spec's schema — mirror the spec, do not invent fields. Add
`func (c *Client) PauseAgent / ResumeAgent(ctx context.Context) (error, error)` using
`getJSON`. These are the first WRITE calls: follow the spec's method and body exactly, and add a `postJSON` helper beside `getJSON` for them. If the spec has no such operation, stop and file a
follow-up issue titled `dashboard/openapi.json missing pausing and resuming an agent operation`
with what you found; that filing completes this task (see #4912 for known
spec gaps).

**Acceptance criteria:**
- Fixture JSON in `client/testdata/actions.json` constructed to match the schema exactly; test asserts decode of all fields and the error path on 500
- `go build ./...` and `go test ./src/pkg/tui/...` pass

**Out of scope:** UI keybindings (T15), kick (T20).

**Verify with:** `go build ./... && go test ./src/pkg/tui/...`

---
Filed per the #4907 filing protocol: this task's dependencies have merged, so it is now unblocked work.

---
Wave 4 of the #4907 filing protocol — unblocked by T4 (#5053, closed).

Danathar:
## Summary

The TUI can now pause and resume an agent. This adds `Client.PauseAgent(ctx, agent)` and `Client.ResumeAgent(ctx, agent)` — the client package's first **write** calls, where everything so far had been reads.

Both `POST` to their documented endpoint and decode the response schema they share into a new `AgentActionResult`. The interesting part isn't the request; it's that this endpoint's response has two traps in it, and both are the kind that produce a UI confidently displaying something that never happened. Tests pin both.

Closes #5134. Part of #4907 (T14).

## The two traps

**`ok` is not the field that says whether anything happened — `changed` is.**

Pausing an already-paused agent is a deliberate no-op. The handler goes out of its way to avoid re-pausing, because that would clobber the original `PausedAt`/reason/trigger, and it answers `200` with `ok:true, changed:false`. A caller reading any 200 as "it worked" reports a transition that never occurred — which is exactly the stale-UI bug `changed` was added to fix.

**`status` and `state` disagree by design.**

A no-op resume returns `status:"resumed"` with `state:"running"`. `status` names the operation that ran; `state` is the agent's actual state afterwards. A pane comparing `status` against the state constants matches neither. `AgentActionResult.Paused()` exists so nobody has to get that right by hand, and `State` is authoritative on the no-op path too — it is populated precisely when `changed` is false.

## Contract sourcing

Following the T4 precedent, I checked the spec against the handler rather than trusting either alone — T6 found them diverging. Here they agree: `dashboard/openapi.json` documents `ok`/`status`/`agent`/`changed`/`state`, and `pauseToggleResponse` (`pkg/dashboard/api.go:1651`) writes exactly those five and nothing else. Nothing is invented or guessed from a live response.

One documented detail worth knowing: `agent` in the response is **not** always what you sent. Both handlers put the path value through `resolveAgentParam`, so an agent addressed by ID comes back as its resolved name. A caller displaying what it acted on should use this field, not echo its own argument.

## Signature correction

The issue proposed `PauseAgent/ResumeAgent(ctx)`. As the scanner analysis noted, that omits the agent, and `{agent}` is a required path parameter — so these take `(ctx, agent string)`.

The name is `url.PathEscape`d into its segment. Ordinary agent names need no escaping, which is exactly why it has to be deliberate rather than left to the common case: a name carrying a separator would otherwise retarget the request at a different route. An empty name is rejected locally, without a round trip, because `/api/pause/` matches no route and would come back as a 404 describing the routing table rather than the caller's mistake.

## Supporting changes to `client.go`

- **`postJSON` joins `getJSON` on a shared `doJSON` core.** Body is marshalled when non-nil and omitted entirely when nil. Neither of these operations declares a `requestBody`, so both send no body and set no `Content-Type` — asserted in the tests, since announcing a content type for a body you didn't send is its own small lie.
- **`APIError` gains `Method`.** It hardcoded `"GET"`. The moment a write shares the type, a 403 from `POST /api/pause/x` would have been reported as `GET /api/pause/x` — sending a reader to look for a request that was never made. This is a bug the first write call creates, so it is fixed here rather than left for the next one.
- **`IsForbidden(err)` types the owner gate.** Every mutating dashboard endpoint sits behind `requireOwnerRole`, and 403 is the one failure a pane must render differently: it is a working request from someone whose role does not permit the action, and retrying — the right response to most other errors — will never help.

On the auth question that gate raises: a self-hosted hive's TUI clears it. The local proxy strips client identity headers and injects `X-Hive-Internal`, which `authenticate()` maps to a verified owner (`pkg/dashboard/server.go`). On a hub-proxied hive the role comes from the hub's per-user grant instead, so a read-only user gets a genuine 403 — the handling is not theoretical.

## Testing

`testdata/actions.json` carries the shared schema; 8 new tests cover:

- Full fixture decode with every field asserted, plus method, path, `Authorization`, and that no body or `Content-Type` is sent
- Resume's own path and method (the two differ only by prefix, so only a test keeps them from collapsing onto one endpoint)
- The `changed=false` no-op on **both** endpoints, asserting `ok` stays true and `state` is still authoritative
- Agent-name escaping, asserted on `EscapedPath()` — `URL.Path` is already decoded and cannot show whether the client escaped anything
- The empty-name guard, including that it never reaches the network
- `400` (unknown agent) and `403` (owner gate) as documented, plus the `500` the acceptance criteria ask for, each asserting `Method` is `POST`
- `IsForbidden` across a bare 403, a **wrapped** 403, other statuses, and nil
- A 200 carrying the wrong shape is a decode error, not an `APIError`

`go build ./...`, `go test ./src/pkg/tui/...`, `go vet`, `gofmt`, and `-race` are all clean.

I mutation-tested the four load-bearing behaviors — dropping the escaping, dropping the empty-name guard, pointing resume at the pause path, and reverting `Method` to a hardcoded `GET` — and confirmed the relevant test fails in each case, so they pin real contracts rather than passing vacuously.

## Out of scope

UI keybindings (T15) and kick (T20), per the issue.


Danathar:
Heads-up for ordering: #5154 (T18, ACMM level get/apply) is **stacked on this branch**.

T18 needs a PUT with a body, and this PR is what introduces the package's write plumbing (`doJSON`/`postJSON`). The scanner analysis on #5136 anticipated the collision — "coordinate with T14 #5134, which introduces `postJSON` — whichever lands first adds it" — so rather than duplicate the refactor and guarantee a conflict in the same function, #5154 builds on this branch and adds only `putJSON`, three lines on the shared core.

**Merging this one first makes #5154 a clean two-file diff.** No change is needed here; this is just so the order is not a surprise.

PR #5154, opened 2026-08-29
✨ feat(tui): client — ACMM level get/apply (#4907 T18)

### LINKED ISSUE #5136: ✨ tui T18: client — ACMM level get/apply

Danathar:
## Summary

The TUI can now read the hive's ACMM maturity level and set it. `Client.ACMM(ctx)` lists every level's pack and which one is in force; `Client.ApplyACMM(ctx, level)` changes it.

> **⚠️ Stacked on #5152 (T14).** Setting a level is a write, and #5152 is the PR that introduces this package's write plumbing (`doJSON`/`postJSON`). Rather than duplicate that refactor and guarantee a conflict in the same function — the collision the scanner analysis on #5136 explicitly warned about ("coordinate with T14, whichever lands first adds it") — this branch builds on it and adds only `putJSON`, three lines on the shared core. **Please merge #5152 first**; this PR's diff against `v4` therefore also shows T14's commit. Happy to rebase onto a plain `v4` the moment #5152 lands.

Closes #5136. Part of #4907 (T18).

## The endpoints are under `/api/packs`, not `/api/acmm`

Worth stating plainly, because the name mismatch looks at first like the task's "spec has no such operation, stop and file" case. It isn't:

- **Read** — `GET /api/packs` returns every maturity-level pack with its governor settings and agent roster, flagging the active one.
- **Apply** — `PUT /api/packs/level` (owner-only) sets the level.

`GET /api/acmm/evaluation` and `GET /api/acmm-recommendation` do exist, but they are read-only *advisory* endpoints — what the hive **should** be at, not what it **is** — so they are deliberately not used here.

Unlike T16, **the spec and the handlers agree here.** `Pack`, `PackAgent` and `PackGovernor` are transcribed from `config.PackAgent`/`config.PackGovernor` and the `packSummary` the handler marshals, and `dashboard/openapi.json` documents the same fields. One small exception noted in-code: the spec types `governor_changes` as a bare `{"type":"object"}`, so that shape comes from `dashboard.GovernorChanges` — the same drift class #5077 tracks.

I modelled `PackAgent` **whole** (20 fields) rather than narrowing to the seven today's web overlay renders. Which subset a roster preview needs is T19's decision, and cherry-picking to match the current web UI is the guess the T4 precedent warns against.

## Why `ACMMStatus` has no json tags

`GET /api/packs` returns a **bare array**. There is no level field anywhere in the response — the active level exists only as the `current` flag on one entry. `ACMMStatus` is the small amount of assembly that saves every caller from re-deriving that, so it is explicitly not a wire type; giving it tags would imply the server sends this shape.

`ACMMStatus.Level` reads `0` when no pack is flagged. That is an unconfigured hive, not an error, and it must not silently default to whichever pack came first.

## Why apply returns a receipt, not the pack list

The task proposed `ApplyACMM(ctx) (ACMMStatus, error)`. Both halves are corrected: the level is a required request-body field, and the response is a different shape entirely.

That matters because **setting a level does considerably more than write a number.** The handler also clears every per-agent mode override so pack modes re-apply cleanly, force-applies the pack to reconcile the roster and the governor's cadences and thresholds, and pauses or resumes agents to match the new roster. `packAgents`, `packUpdated`, `governor_changes`, `paused` and `resumed` are that work's receipt. A client that returned only "level set" would hide work the operator asked for and may well want to check.

## Two contract details a caller must not get wrong

**A 500 here is not a no-op.** It means the level was persisted but the roster could **not** be reconciled to it. The handler returns it precisely so the operator sees the drift rather than trusting a false success — the code comment cites a hive left at `acmm_level: 5, 8 agents, strategist missing`. `ApplyACMM` documents this so no pane reports it as "nothing happened".

**The 1-6 range is deliberately not re-validated client-side.** The server owns it and answers out-of-range with a 400 naming the range, which is a good error and always current. Duplicating the bound would only add a way for this client to go stale against a level the hive later gains. There is a test pinning this choice, so adding a local guard later fails loudly rather than silently.

## Testing

`testdata/acmm.json` carries two packs, one of them flagged current, with an agent exercising all 20 fields including the five that are `omitempty`. 8 new tests cover:

- Full fixture decode: both packs, the complete governor map/threshold shapes, the whole agent struct via `reflect.DeepEqual`, and the derived `Level`
- Absent optional governor fields decoding to zero values rather than failing
- No current pack → `Level == 0`, packs still returned
- The write contract: `PUT` (not POST — the mux matches on method, so a POST never reaches the handler), path, `Content-Type`, and that the body is exactly `{"level":N}`
- `governor_changes` staying `nil` when absent, so "nothing moved" is distinguishable from "moved to zero"
- `400`, `403` and `500`, each asserting the server's own message survives and that `IsForbidden` fires only on the 403
- That an out-of-range level still reaches the server
- A 200 with the wrong shape being a decode error, not an `APIError`

`go build ./...`, `go test ./src/pkg/tui/...`, `go vet`, `gofmt` and `-race` are all clean.

I mutation-tested the load-bearing behaviors — `putJSON` issuing POST, the derived `Level`, and the `governor_changes` and `packUpdated` json tags — and confirmed the relevant test fails in each case.

## Out of scope

The overlay UI (T19), per the issue.


clubanderson:
Reopened — I closed this by mistake.

My merge sweep called `gh pr merge --squash --delete-branch` while Prow had the PR at NOT APPROVED, so the merge was refused and the command closed the PR instead of leaving it open. That was my error, not a decision about the change.

The branch `feat/5136-tui-acmm` survived at `f9307473` (the rebase I pushed earlier today), so nothing was lost. T18 is back in the queue.

Issue #5112, opened 2026-08-29
🐛 bug: PR titles and Fixes references do not match the diff


hanthor:
## What happened?

**Impact:** a `Fixes` reference that lands would auto-close an issue the PR does
not fix.

**Incident:** https://github.com/tuna-os/mandelbrot/pull/82 — titled
`ci: add upstream sync workflow` and claimed `Fixes #60`. It contains **no
workflow file**: only `docs/UPSTREAM.md`, a manual script, and a README pointer.
The `Fixes` was manually downgraded to `Refs` before it could auto-close a
partly-addressed issue. **[reported]**

## Expected behavior

Validate that the title's claim is present in the diff, and that a `Fixes`/`Closes`
reference is backed by a change that actually closes the issue. A title naming an
artifact type (`workflow`, `test`, `migration`) should require a file of that type
in the diff.

## Reproduction steps

1. Open https://github.com/tuna-os/mandelbrot/pull/82.
2. Compare the title (`ci: add upstream sync workflow`) and the `Fixes #60`
   reference against the files in the diff.
3. Observe no workflow file is present — only `docs/UPSTREAM.md`, a manual script,
   and a README pointer.

## Environment

- Branch/version: not recorded; incident observed 2026-08-28/29.
- Deployment method: agent-filed PRs across repositories in the `tuna-os` GitHub org.
- Operating system or Kubernetes version: not applicable — the behaviour is
  agent-side, not deployment-specific.
- Relevant configuration: not applicable.

## Logs or screenshots

- https://github.com/tuna-os/mandelbrot/pull/82 **[reported]**

## Additional context

**Evidence markers.** Items marked **[verified]** were checked directly during the
session that produced this report. Items marked **[reported]** come from analysis
runs whose conclusions were not re-derived first-hand — the links are included so a
maintainer can confirm them. All linked PRs are public.

PR #5205, opened 2026-08-30
fix(outreach): gate unsupported capability claims

### LINKED ISSUE #5115: 🐛 bug: the outreach agent authors capability claims the product does not support
## What happened?

**Impact:** the most serious of the set. These are public claims made under the
project's name about behaviour that does not exist, including a regulatory one.

**Incidents (all tuna-os/docs):**

- **#333 — "HIPAA Session Compliance."** HIPAA obligations attach to covered
  entities and business associates, not to software; there is no certification an
  OS can hold, so the heading asserts something that cannot be true in principle.
  Its supporting sentence — *"Patient records and EHR session credentials are
  never written to the immutable system partition `/usr`"* — is true of every
  Linux system ever shipped and says nothing about `/home`, `/var`, swap or
  `/tmp`, which is where PHI would actually land. The page also claimed a
  10-minute inactivity session reset and native PAM PKCS#11 support as properties
  of the base image; neither ships. Closed. **[verified]** — I read the full diff.
- **#340** — an unsubmitted conference proposal written as a docs page, orphaned
  from the sidebar, committing publicly to roadmap specifics (GNOME 51, COSMIC
  Epoch 1.5, Snapdragon X Elite) that nobody had committed to. Closed.
  **[reported]**
- **7 of 12 remaining guides carried capability overclaims** that I corrected
  before merge **[verified]**:
  - "prevents ransomware persistence" — `/home`, `/var`, systemd user units and
    containers all stay writable
  - a power cut "does not leave the disk in a state that cannot boot" — `/var` is
    writable and an interrupted update can still stop a boot
  - automatic rollback on a failed health check — that is greenboot, not the base
    image
  - "one reboot puts the machine back on the last known good image" — needs
    `bootc rollback` or a GRUB selection
  - "Vulkan and OpenGL drivers for AMD, NVIDIA, and Intel … need no setup" —
    NVIDIA needs a variant that ships the driver
  - "Zero-Trust Networking" describing nothing more than putting a VPN client in
    the image
  - a reboot implying student/patron data is cleared when it is not

**Pattern:** the agent reliably produces the *shape* of a vertical solution brief —
compliance bullet, security absolute, integration list — and fills it with claims
selected for plausibility rather than truth. The failure is not occasional
inaccuracy; it is that nothing in the pipeline asks whether the product does the
thing.

## Expected behavior

An agent authoring outward-facing capability claims needs a verification step
against the product, and regulatory/compliance claims should be refused outright
rather than generated — no agent should be asserting HIPAA, PCI, GDPR or SOC 2
posture on a project's behalf.

## Reproduction steps

1. Let the outreach agent author vertical/solution documentation pages.
2. Read the resulting claims against what the product actually ships (e.g. the
   full diff of tuna-os/docs#333, and the 12 guide pages filed alongside it).
3. Observe compliance headings, security absolutes and integration lists asserted
   with no verification step against the product.

## Environment

- Deployment method: outreach agent filing PRs against tuna-os/docs.

## Logs or screenshots

- tuna-os/docs#333 — "HIPAA Session Compliance", including the claimed 10-minute
  inactivity session reset and native PAM PKCS#11 support. Closed. **[verified]**
- tuna-os/docs#340 — unsubmitted conference proposal as a docs page, orphaned from
  the sidebar, naming GNOME 51, COSMIC Epoch 1.5, Snapdragon X Elite. Closed.
  **[reported]**
- 7 of the 12 remaining guides, overclaims listed above. **[verified]**

## Additional context

Danathar:
## Problem

The outreach agent could publish persuasive product and solution documentation containing capabilities the product does not provide. The reported incidents included a "HIPAA Session Compliance" page, security absolutes, automatic rollback behavior, zero-setup hardware claims, and roadmap commitments that were not grounded in shipped functionality or human-approved plans.

This was especially risky because the outreach policy explicitly said its content PRs needed no hold label. At ACMM L6, unsupported claims could therefore reach a public project site through the normal automerge path without human review.

## Root cause

Two policy layers were missing:

1. `outreach-full.md` optimized for engagement and instructed the agent to write content, but did not require each capability claim to be verified against the repository or a released artifact. It also did not prohibit regulatory/compliance assertions or invented roadmap commitments.
2. The server-side PR watcher only applied `hold` according to the global L3-L5 ACMM gate. It had no role-specific L6 exception for outreach, no requirement that outreach expose its claim evidence to reviewers, and no pre-publication rejection for regulatory language.

The result was a prompt that rewarded plausible solution-copy shape and a merge path that treated public claims like ordinary autonomous code changes.

### Ground outreach content in product evidence

Both the source policy (`src/policies/outreach-full.md`) and its embedded runtime default (`src/pkg/policies/defaults/outreach-full.md`) now require the outreach agent to:

- inventory every product, security, compatibility, integration, and roadmap claim before writing;
- verify each claim against an exact implementing file, test, released artifact, or human-authored official source;
- include a `Claim evidence` mapping in the PR body;
- omit unverifiable claims and escalate them to a human;
- refuse all regulatory/compliance posture claims, including HIPAA, PCI DSS, GDPR, SOC 2, FedRAMP, and FIPS language;
- avoid roadmap commitments unless an existing human-authored plan or announcement supports them; and
- open every outreach PR with `hold`, including at ACMM L6, without removing the label or merging the PR itself.

### Enforce the boundary at the PR-request choke point

`src/pkg/github/pr_request_outreach.go` validates authenticated outreach requests before `CreatePR` makes them public:

- the PR body must contain a `Claim evidence` section;
- regulatory terms in the title or body are rejected;
- the target branch is compared through GitHub and added lines are scanned for the same regulatory terms;
- a 300-file boundary response or an added text file whose patch cannot be inspected fails closed with an actionable reason;
- permanent content-policy failures are written to the result file and quarantined as `.rejected`;
- forge/default-branch/compare failures remain transient and use the existing bounded retry path.

The validator is deliberately scoped to the authenticated `outreach` role. Other agent roles keep their existing PR behavior, while a human can still author the regulatory language that the outreach policy tells the agent to escalate.

### Keep outreach human-reviewed at L6

The authoritative hold-label callback now receives the authenticated agent name. `shouldHoldAgentPR` preserves the existing behavior for every other role—held at L3-L5 and autonomous at L6—but always holds outreach PRs.

Required-label failures are no longer logged and treated as successful. The request remains queued under bounded backoff; its retry deduplicates the already-open PR and reapplies `hold`. This preserves the invariant that the watcher does not silently consume a request whose required review gate was not installed.

## Design decisions and boundaries

- The mechanical validator does not pretend that a filename citation proves a semantic capability. It makes evidence visible, rejects the highest-risk regulatory class, and relies on the mandatory human review gate for semantic review.
- The regulatory matcher covers the concrete regimes from the incident and scanner analysis. The policy prohibition intentionally remains broader than the matcher so renamed or less common certification regimes are still out of bounds for the agent.
- Regulatory scanning is limited to outreach rather than blanket-blocking all agent roles. This avoids preventing security and governance agents from accurately documenting the absence of a certification; outreach must hand such language to a human.
- The GitHub compare API is capped at 300 files and can omit patches for very large text changes. Those cases fail closed instead of silently approving an uninspected tail. Large outreach campaigns should be split into reviewable PRs or opened manually by a human.
- This change protects outreach PR creation. Outreach issue creation is governed by the strengthened policy but is not mechanically keyword-scanned in this patch.

## Regression coverage

- `TestOutreachPolicyRequiresGroundedHumanReviewedClaims` checks both policy copies for the evidence, compliance, roadmap, and mandatory-hold invariants and prevents the old exemption from returning.
- `TestShouldHoldAgentPR` proves outreach remains held at L6 while other L6 agents remain autonomous and the existing L3-L5 gate is unchanged.
- Outreach validator tests cover missing evidence, regulatory terms in every supported spelling, PR metadata, changed-file content, safe grounded content, non-outreach bypass, and fail-closed missing patches.
- `TestPRRequestWatcherRejectsOutreachRegulatoryClaimBeforeCreate` proves an invalid request is quarantined and no pull request is created.
- `TestPRRequestWatcher_RetriesRequiredHoldLabelFailure` proves a transient label failure retains the request, reuses the already-open PR, retries the label, and only then consumes the request.

## Verification

Passed:

```text
go test ./pkg/github ./pkg/policies ./cmd/hive
ok github.com/kubestellar/hive/pkg/github
ok github.com/kubestellar/hive/pkg/policies
ok github.com/kubestellar/hive/cmd/hive

go test ./pkg/github -run 'TestValidateOutreach|TestPRRequestWatcherRejectsOutreach|TestPRRequestWatcher_RetriesRequiredHold|TestPRRequestWatcher_HoldLabelApplied'
ok github.com/kubestellar/hive/pkg/github

git diff --check
```

The full `go test ./...` run compiled and passed the changed packages, but the repository-wide command was not green in this checkout for unrelated environment-specific tests:

- `pkg/agent` tmux tests failed because their generated Unix socket path exceeded the platform path limit (`File name too long` under this long worktree path).
- `pkg/config`'s IPv6 egress-gate test encountered the host `ip6tables` binary without the root/NET_ADMIN capability it expects, so its simulated "binary missing" assertion did not apply.

Neither failing package is changed by this PR. The affected package suites above pass independently.

Fixes #5115

— hive: backend=codex

### DIFF
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -1695,13 +1695,15 @@ func main() {
 		// privilege. A denied request is quarantined, never opened.
 		// holdLabel (F6): at hold-gated ACMM levels (L3/L4/L5) every agent-opened
 		// PR must carry the "hold" label so the merge gate holds it for human
-		// approval. This is decided server-side from the authoritative hive level
+		// approval. Outreach content is public speech on the project's behalf, so
+		// it remains human-reviewed at L6 too. This is decided server-side from the
+		// authenticated agent identity and authoritative hive level
 		// (GetACMMLevel), NOT from a client flag — the gh-wrapper.sh tail that used
 		// to add the label was dead code after `exec hive-open-pr`. L1/L2 open no
-		// agent PRs (manual); L6 auto-merges on green (no hold).
-		holdLabel := func() bool {
-			l := agentMgr.GetACMMLevel()
-			return l >= acmmHoldGatedMinLevel && l <= acmmHoldGatedMaxLevel
+		// agent PRs (manual); non-outreach L6 PRs retain their existing automerge
+		// behavior.
+		holdLabel := func(agentName string) bool {
+			return shouldHoldAgentPR(agentName, agentMgr.GetACMMLevel())
 		}
 		ghClient.StartPRRequestWatcher(ctx, agentMgr.AuthorizePROpen, holdLabel, nil)
 		// Issue relay: agents request issue creation and comments by dropping a
@@ -7613,6 +7615,16 @@ const (
 	acmmHoldGatedMinLevel = 3
 	acmmHoldGatedMaxLevel = 5
 )
+
+// shouldHoldAgentPR keeps public outreach claims human-reviewed even at L6,
+// where ordinary agent PRs may auto-merge. The general ACMM hold gate remains
+// unchanged for all roles at L3-L5.
+func shouldHoldAgentPR(agentName string, level int) bool {
+	if strings.EqualFold(strings.TrimSpace(agentName), "outreach") {
+		return true
+	}
+	return level >= acmmHoldGatedMinLevel && level <= acmmHoldGatedMaxLevel
+}
 
 // mergeableJSONUnknown is the explicit wire value for "mergeability was never
 // determined". It is spelled out rather than left as "" so a consumer reading
--- a/src/cmd/hive/outreach_hold_test.go
+++ b/src/cmd/hive/outreach_hold_test.go
@@ -1,0 +1,27 @@
+package main
+
+import "testing"
+
+func TestShouldHoldAgentPR(t *testing.T) {
+	tests := []struct {
+		name  string
+		agent string
+		level int
+		want  bool
+	}{
+		{name: "outreach remains held at full autonomy", agent: "outreach", level: 6, want: true},
+		{name: "outreach identity is normalized", agent: " Outreach ", level: 6, want: true},
+		{name: "other agents remain autonomous at level 6", agent: "scanner", level: 6, want: false},
+		{name: "all agents are held at level 3", agent: "scanner", level: 3, want: true},
+		{name: "all agents are held at level 5", agent: "quality", level: 5, want: true},
+		{name: "manual level does not add a hold", agent: "scanner", level: 2, want: false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := shouldHoldAgentPR(tt.agent, tt.level); got != tt.want {
+				t.Fatalf("shouldHoldAgentPR(%q, %d) = %v, want %v", tt.agent, tt.level, got, tt.want)
+			}
+		})
+	}
+}
--- a/src/pkg/github/client.go
+++ b/src/pkg/github/client.go
@@ -65,12 +65,14 @@ type Client struct {
 	// StartPRRequestWatcher.
 	prAuthz PRRequestAuthorizer
 	// prHoldLabel decides, at PR-creation time, whether a freshly-opened PR must
-	// carry the "hold" label (F6). It is consulted server-side from authoritative
-	// hive config (the ACMM level), NOT trusted from a client flag — the
+	// carry the "hold" label (F6). It receives the authoring agent so role-specific
+	// safety rules (for example, mandatory human review of outreach content) can
+	// be enforced alongside the authoritative ACMM level. It is consulted
+	// server-side, NOT trusted from a client flag — the
 	// gh-wrapper.sh tail that used to add the label is dead code (it sits after
 	// `exec hive-open-pr`). nil means "never hold" (backward-compatible no-op).
 	// Set by StartPRRequestWatcher.
-	prHoldLabel func() bool
+	prHoldLabel func(agent string) bool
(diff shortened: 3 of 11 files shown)


clubanderson:
Flagging that this may be substantially superseded — worth a look before rebasing.

`validatePRRequestClaims` and the server-side `holdLabel` application are both on `v4` now, landed by #5189 (`0a76e18d`, "validate PR claims before creation"). `holdLabel` has 4 references in `pr_request_watcher.go` on the current tip.

That is why this branch conflicts against `v4` itself rather than against another PR: the same function was rewritten underneath it.

**What still looks unique here** (11 files, ~530 additions): `pr_request_outreach.go` (+86) and its tests (+176), the `outreach-full.md` policy text in both trees (+25 each), `outreach_claims_test.go` (+51), and the `cmd/hive` wiring (+17). #5189's body references #5112, not #5115 — so the outreach-specific capability-claim gating this PR adds may well still be needed even though the shared plumbing landed.

I did not attempt the resolution. The conflict is semantic — two different validations at the same choke point — and picking an ordering is an authoring decision, not a mechanical one.

**#5115 is still open**, so if the outreach half of this is genuinely still required, that is the thing to preserve when rebasing.

Separately, I verified your dead-code finding independently and it holds: `bin/gh-wrapper.sh:531` does `exec hive-open-pr "$@"`, so the hold-label block at line 606 is unreachable — hold-gated PRs really were opening unlabeled. Good catch.

PR #5198, opened 2026-08-30
fix(github): keep agent metadata out of publishable content

Danathar:
## Summary

Agent policy templates showed attribution trailers in issue and PR bodies but never stated that those trailers must stay out of repository content. The PR request watcher also passed a pushed branch to `CreatePR` without checking whether the branch itself contained Hive's internal attribution.

This PR adds both layers of protection:

- every shipped PR-capable policy and its embedded default now says attribution and run metadata belong only in issue/PR bodies or commit trailers, never committed files
- the server-side PR request choke point scans added compare-diff lines before opening a PR
- added `Filed by ... agent (ACMM ...)` attribution and `— hive: ...` run trailers permanently reject the request
- the rejection result names the offending file and new-file line, while deleted or pre-existing markers remain allowed so cleanup PRs can land
- GitHub/default-branch/compare failures remain transient and use the watcher's existing bounded retry path
- compare results at GitHub's 300-file visibility limit fail closed rather than claiming an incomplete scan is clean

Fixes #5114.

## Reported incident

The regression test submits the incident shape through `ProcessPRRequestsOnce`: a Markdown page with a trailing outreach attribution. It verifies that no `POST /pulls` occurs, the request is quarantined as `.rejected`, and the result points to the exact Markdown file and line.

The patch scanner looks only at added lines. A branch that removes leaked metadata is therefore accepted, and unchanged historical content does not block unrelated work.

## Policy coverage

The boundary is present in all 20 PR-capable full, hold-gated, and automerge templates, including outreach, guide, architect, scanner, quality, CI, security, strategy, telemetry, and operations roles. A policy invariant test checks both the human-editable source templates and the embedded defaults.

## Validation

Passed:

- `go test ./pkg/github`
- `go test ./pkg/policies`
- `go vet ./pkg/github ./pkg/policies`
- `git diff --check`

`go test ./...` also passed both changed packages. The repository-wide run could not complete in this checkout because unrelated packages exhausted the temporary disk quota while linking; host-sensitive `pkg/agent` and `pkg/config` tests also hit the existing overlong tmux socket path and unavailable `NET_ADMIN` environment limitations.

## Contributor checklist

- [x] Targets `v4`
- [x] Ready for review, not draft
- [x] Commit carries a DCO sign-off matching the author email
- [x] Regression test covers the reported publishable-content leak
- [x] No credentials or generated runtime artifacts committed

— hive: backend=codex

Issue #5117, opened 2026-08-29
🐛 bug: an agent files an issue proposing a direction, then implements it with no human sign-off


hanthor:
## What happened?

**Impact:** architectural change enters a repo on the agent's own authority. The
issue provides the appearance of a rationale, but nothing external ever agreed
to it.

**Incident:** tuna-os/tunaos-packages#583 introduces `scripts/lib/build-chain/`,
a module-sourcing pattern that does not exist anywhere in the repository, as
"phase 1" of a decomposition proposed by tuna-os/tunaos-packages#581 — an issue
**the hive also filed**. No human reviewed or approved the direction between the
proposal and its implementation. **[reported]**

The code itself is fine (the extraction is byte-identical apart from six dropped
comments, and it is green). That is what makes this a governance bug rather than
a code bug: the change is individually defensible and still should not have
self-authorised.

## Expected behavior

An agent-filed issue should not satisfy the "there is a tracked rationale"
precondition for an agent-filed PR. When a PR introduces a pattern with no
precedent in the repository, require a human acknowledgement on the proposing
issue before implementing it.

## Reproduction steps

1. Read tuna-os/tunaos-packages#581 — an issue filed by the hive proposing a
   decomposition.
2. Read tuna-os/tunaos-packages#583 — the PR implementing "phase 1" of it,
   introducing `scripts/lib/build-chain/`.
3. Observe no human review or approval of the direction between the two, and no
   precedent for the module-sourcing pattern anywhere in the repository.

## Environment

- Branch/version: not recorded; incident observed 2026-08-28/29.
- Deployment method: agent-filed issue and PR on tuna-os/tunaos-packages.
- Operating system or Kubernetes version: not applicable — the behaviour is
  agent-side, not deployment-specific.
- Relevant configuration: not applicable.

## Logs or screenshots

- Proposing issue: tuna-os/tunaos-packages#581 (filed by the hive). **[reported]**
- Implementing PR: tuna-os/tunaos-packages#583, introducing
  `scripts/lib/build-chain/`; extraction byte-identical apart from six dropped
  comments, CI green. **[reported]**

## Additional context

**Evidence markers.** Items marked **[verified]** were checked directly during the
session that produced this report. Items marked **[reported]** come from analysis
runs whose conclusions were not re-derived first-hand — the links are included so a
maintainer can confirm them. All linked PRs are public.

PR #5276, opened 2026-08-31
🐛 fix(github): stop filing a PR whose content an open PR already carries (#5111)

### LINKED ISSUE #5111: 🐛 bug: agents open duplicate PRs carrying byte-identical payloads
## What happened?

**Impact:** reviewer time, and a queue that looks like five problems when it is one.

**Incidents:**

- **tuna-os/tromso — 5 PRs with the same autotune payload:** #170, #212, #225,
  #241, #244. The autotune script and its test files are the *same blobs*
  (`d428ad76…`, `a41436e2…`) across all five. #241 and #212 have identical trees
  (`eb0b85ab…`) — `git diff` between their tips is empty. **[reported]**
- **tuna-os/docs — 4 duplicate pairs:** #315↔#340 (same FOSDEM talk),
  #316↔#338 (gaming), #323↔#339 (audio/creative), #324↔#327 (Niri). #324 was a
  25-line orphan duplicating a 371-line guide already wired into `sidebars.ts`
  and cross-linked from three pages. **[reported]**

These were never five or four different attempts — one change was copied forward
repeatedly.

## Expected behavior

Before opening a PR, diff the intended tree against open PRs on the same repo. An
empty diff against an existing PR is a no-op, not a new contribution.

## Reproduction steps

1. Let agents work a repository over several cycles where an earlier attempt is
   already open as a PR.
2. Compare the trees and blobs of the resulting PRs (e.g. `git diff` between the
   tips of tuna-os/tromso#241 and #212).
3. Observe byte-identical payloads filed as separate PRs.

## Environment

- Deployment method: agent-filed PRs across repositories in the `tuna-os` GitHub org.

## Logs or screenshots

- tuna-os/tromso: shared blobs `d428ad76…` and `a41436e2…` across #170, #212, #225,
  #241, #244; identical tree `eb0b85ab…` on #241 and #212. **[reported]**
- tuna-os/docs: duplicate pairs #315↔#340, #316↔#338, #323↔#339, #324↔#327.
  **[reported]**

## Additional context

Danathar:
## Summary

`CreatePR` has always been idempotent for the same **head branch**: re-requesting a PR for a branch that already has one returns the existing PR. That covers a *retried request*. It cannot see the incident in #5111, where an agent copies a change forward onto a **new** branch and files it again — tuna-os/tromso collected five such PRs (#170, #212, #225, #241, #244) sharing the same blobs, two of them (#241, #212) with an identical tree, so `git diff` between their tips was empty.

The issue's Expected Behavior is exactly one sentence, and this implements it:

> Before opening a PR, diff the intended tree against open PRs on the same repo. An empty diff against an existing PR is a no-op, not a new contribution.

## The discriminator

The commit's **tree SHA**. Two commits with the same tree have byte-identical content, so two open PRs *on the same base* whose head commits share a tree propose exactly the same change — whatever their branch names, commit messages, or authorship.

That is an **equality test, not a similarity heuristic**, which is what makes it safe to act on automatically: it cannot fire on two changes that merely look alike.

`CreatePR` consults it after the existing head-branch check and returns the matching PR with `AlreadyExisted` and a new `DuplicateTree` flag set. The request is then consumed rather than retried, and the agent's `.result.json` points at the PR that already proposes its change. This sits at the choke point every agent PR flows through, so it covers the request watcher and the sandbox executor alike (the proxy hard-denies direct `POST /pulls`).

## Three deliberate properties, each pinned by a test

| Property | Why |
|---|---|
| **Same base only** | An identical tree on a *different* base is a different diff. The open-PR query is scoped by base rather than scanning everything open. |
| **Fails OPEN** | Every read the guard makes is auxiliary — ref lookup, PR list, commit lookup — and a failure proceeds to create. A guard that withheld a PR because a lookup failed would be a worse bug than the duplicate it prevents. |
| **Bounded cost** | Candidates are capped, reaching the cap is *logged* rather than passed over in silence, and commit→tree is memoized. That mapping is immutable, so the cache has no staleness window; without it the guard would re-fetch every open PR's tree on every creation — the secondary-rate-limit exposure the watcher already had to adopt backoff for. |

## One side effect suppressed on purpose

The watcher no longer applies the ACMM `hold` label on the duplicate path. There it is pointing at a PR **this request did not open**, and which may not even be the hive's — an agent can duplicate a human's content, and stamping `hold` on a stranger's PR would block their merge. When the duplicate is the agent's own earlier PR, that PR already got its label when it was opened through this same path.

## Verification

Tests are **mutation-checked six ways**; each mutant fails a test that names the property it broke:

| Mutant | Test that catches it |
|---|---|
| guard removed entirely | `ReusesOpenPRWithIdenticalTree`, `MatchesIdenticalHeadSHA`, `HandlesSlashedBranchNames`, both watcher tests |
| base scoping dropped | `DuplicateTreeOnlyComparesTheSameBase` |
| candidate cap removed | `FindOpenPRWithIdenticalTree_CapsCandidates` |
| fail-**closed** instead of fail-open | `DuplicateTreeFailsOpen` (all 4 subcases) |
| tree cache disabled | `CommitTreeSHA_CachesImmutableMapping` |
| hold label re-enabled on duplicate path | `DuplicateTreeDoesNotLabelTheOtherPR` |

`TestCreatePR_OpensWhenTreesDiffer` is the control — without it the suite would also pass if the guard simply refused everything. `HandlesSlashedBranchNames` guards a real footgun rather than a hypothetical: nearly every hive branch name contains a slash, and the guard resolves it through a ref lookup whose URL has to survive the separator.

- `go build ./...` — clean
- `go vet ./pkg/github/...` — clean
- `go test ./pkg/github/` — passes (full package, not just the new tests)

## Scope note

This addresses the **byte-identical** class the issue's Expected Behavior names. The **near-duplicate** class also listed there (tuna-os/docs #315↔#340 and friends: same topic, different bytes) is deliberately left alone. It has no exact test, so the analysis on the issue correctly puts it in policy text — but the policy packs currently instruct the *opposite* ("never run `gh pr list` unprompted"), and the packs are duplicated across ~40 hand-maintained files in `policies/` and `pkg/policies/defaults/`. Reconciling those is a maintainer decision, not a drive-by edit.

Fixes #5111.

### DIFF
--- a/src/pkg/github/client.go
+++ b/src/pkg/github/client.go
@@ -174,6 +174,14 @@ type Client struct {
 	// re-installs it while the PR-request watcher goroutine may be reading it.
 	hiveIdentityMu sync.RWMutex
 	hiveIdentity   HiveIdentity
+	// commitTrees memoizes "owner/repo@commitSHA" -> tree SHA for the
+	// duplicate-payload guard (pr_duplicate_tree.go, #5111). A commit's tree
+	// can never change, so unlike defaultBranches this cache has no staleness
+	// window at all; it exists purely so re-inspecting the same open PRs on
+	// every PR creation does not re-pay a lookup per candidate. Guarded by
+	// commitTreeMu.
+	commitTreeMu sync.RWMutex
+	commitTrees  map[string]string
 }
 
 func (c *Client) SetCanaryScanner(enabled, failClosed bool, reg *ioscan.CanaryRegistry, onLeak func(ioscan.CanaryLeak)) {
--- a/src/pkg/github/pr_duplicate_tree.go
+++ b/src/pkg/github/pr_duplicate_tree.go
@@ -1,0 +1,175 @@
+package github
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"strings"
+
+	gh "github.com/google/go-github/v72/github"
+)
+
+// Content-identity duplicate guard (kubestellar/hive#5111).
+//
+// CreatePR has always been idempotent for the SAME head branch: re-requesting a
+// PR for a branch that already has one returns the existing PR instead of
+// opening a second. That covers a retried request; it cannot see the incident
+// class this file exists for — an agent copying one change forward onto a NEW
+// branch and filing it again. tuna-os/tromso collected five such PRs (#170,
+// #212, #225, #241, #244) sharing the same blobs, two of them (#241, #212)
+// carrying an identical tree, so `git diff` between their tips was empty.
+//
+// The discriminator is the commit's TREE SHA. Two commits with the same tree
+// have byte-identical content, so two open PRs on the same base whose head
+// commits share a tree propose exactly the same change — whatever their branch
+// names, commit messages, or authorship. That is an equality test, not a
+// heuristic: it cannot fire on two changes that merely look alike, which is why
+// it is safe to act on automatically. The near-duplicate class in the same
+// issue (same topic, different bytes) is deliberately NOT addressed here; it
+// has no exact test and belongs in policy text, not in a gate that silently
+// swallows work.
+//
+// Comparing only against PRs with the SAME base matters. An identical tree on a
+// different base is a different diff, so it is not a duplicate.
+
+// maxDuplicateTreeCandidates bounds how many open PRs one guard pass inspects.
+// Each uncached candidate costs a commit lookup, and PR creation must not turn
+// into an unbounded fan-out on a repository with hundreds of open PRs — the
+// same secondary-rate-limit exposure that made the request watcher adopt
+// backoff. Reaching the cap is logged rather than passed over in silence: a
+// truncated pass can miss a duplicate, and an operator reading "no duplicate
+// found" deserves to know the search was partial.
+const maxDuplicateTreeCandidates = 50
+
+// findOpenPRWithIdenticalTree returns an open PR against base whose head commit
+// points at the same tree as head's tip, or nil when there is none.
+//
+// Errors are for the CALLER to decide about. This returns them rather than
+// swallowing them because a dedupe lookup that failed is not the same fact as a
+// dedupe lookup that found nothing, and CreatePR deliberately treats the former
+// as "proceed and open the PR" — refusing to publish work because an auxiliary
+// read failed would be a worse bug than the duplicate it is guarding.
+func (c *Client) findOpenPRWithIdenticalTree(ctx context.Context, owner, repo, head, base string) (*gh.PullRequest, error) {
+	if c == nil || c.client == nil {
+		return nil, ErrNoGitHubClient
+	}
+	headSHA, err := c.LatestCommitHash(ctx, owner, repo, head)
+	if err != nil {
+		return nil, err
+	}
+	headTree, err := c.commitTreeSHA(ctx, owner, repo, headSHA)
+	if err != nil {
+		return nil, err
+	}
+	if headTree == "" {
+		return nil, fmt.Errorf("duplicate-tree guard: commit %s carries no tree SHA", headSHA)
+	}
+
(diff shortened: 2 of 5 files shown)


clubanderson:
This now conflicts with `v4` in `src/pkg/github/client.go`, and it is a semantic conflict rather than a mechanical one — leaving it to you rather than guessing at a resolution.

The cause: #5280 (your `#5117` self-proposal gate) merged and touched the same file. Both PRs add a server-side guard on the PR-request path, so they land in adjacent code:

- **#5280** — an agent-filed proposal issue no longer authorizes an agent to implement it
- **#5276** (this) — an agent no longer files a PR whose content an open PR already carries

Both belong; #5111 is still open, so nothing here is superseded. The question is how the two checks compose and in what order they run, which is your call — a mechanical merge would compile while quietly getting the precedence wrong.

Everything else in the queue is unblocked now if you were seeing unrelated reds: `v4`'s `pkg/dashboard` build break is fixed (#5279) and the repo-wide `notice-drift` failure was a single trailing space in `NOTICE`, fixed in #5294.

Danathar:
Rebased onto current `v4` and resolved. Thanks for flagging it as semantic rather than mechanical — it was, and in a direction that would have compiled quietly.

## The resolution: #5111 over #5117, and over the ACMM level gate

The two guards answer different questions, and that's what settles the order:

- **#5117 and the level gate** ask *"may this agent's change proceed to merge?"* — a property of **the PR this request created**.
- **#5111** asks *"did this request create a PR at all?"* — and when it fires, the answer is **no**. `res` names a PR that already existed, which may belong to another agent or to a human.

So on the duplicate path the request takes **neither** hold reason and does not run the self-authorisation lookup at all. Labelling or commenting there acts on somebody else's PR, and a `hold` stamped on it blocks a merge that has nothing to do with this request.

**No governance is skipped by that ordering.** The PR being pointed at went through the self-authorisation gate when *it* was opened through this same path; and where it's a human's PR, hive has no standing to gate it. Evaluating the gate here would additionally spend GitHub calls to compute a hold that must not be applied.

A mechanical merge would have taken `if holdByLevel || selfAuth.Held` and quietly stamped `hold` on a stranger's PR — exactly the case #5111's original no-label carve-out existed for, just reached through a second door.

Two tests pin it, both mutation-checked:
- `DuplicateTreeDoesNotLabelTheOtherPR` — applying the hold on the duplicate path
- `DuplicateTreeIsNotSelfAuthorizationJudged` — evaluating self-authorisation there

The second is deliberately at a **non-hold-gated** level: at a hold-gated one #5117 is already short-circuited by the level, so the test would pass without the precedence existing at all.

## Two test-quality problems the rebase surfaced

Both were mine, both fixed here:

1. **A test was passing for the wrong reason.** #5198's content-metadata gate landed after this branch was cut, so the request was failing before it ever opened a PR — and a request that opens no PR applies no labels either, which is all that test asserted. It now asserts the duplicate verdict was actually *reached* before asserting the absence of labels, and the fake forge learned the `/compare` endpoint.

2. **A first attempt at the precedence test was vacuous.** I asserted "no issue lookups on the duplicate path" as a proxy for the gate not running. That counter is confounded — `pr_request_claims.go` reads the cited issue too — so it could never isolate #5117, and the mutant survived. Replaced with an assertion on the verdict itself, which does bite.

## On the force-push

`origin/fix/5111-duplicate-tree-guard` had been moved to `d47af105`, a rebase of my own commit onto `4395c310`. That base predates #5280, so no conflict arose there and the result carried none of #5117's code. I checked it was my commit and not new work before overwriting; the branch is now `fae6a28f` on `34322d46`, with the conflict resolved as above.

Verified: `go build ./...`, `go vet ./pkg/github/...`, and the full `pkg/github` suite pass locally, and CI is green on everything that has finished.

Thanks also for the note about the queue — I'd have chased those reds otherwise.

PR #5288, opened 2026-08-31
fix(github): gate PRs implementing an unacknowledged bot-filed proposal

### LINKED ISSUE #5117: 🐛 bug: an agent files an issue proposing a direction, then implements it with no human sign-off

clubanderson:
Fixes #5117

## Feasibility check

The issue asks whether a code gate is right, or whether this is primarily a policy question — and warns that a gate which is trivially bypassed is worse than an explicit rule. I re-derived the previous attempt's "a code gate is feasible and well-supported" conclusion from the actual code rather than trusting it, and it holds, with one honest limitation stated below.

**Why a gate is feasible here specifically:** `pkg/proxy/rules.go` hard-denies direct `POST /pulls` for every agent — there is no way to open a real PR except through the hive-opens-PR request-file relay (`bin/hive-open-pr.sh` → `pkg/github/pr_request_watcher.go`'s `handleOnePRRequest`). That watcher already runs three server-side policy gates on every request before calling `CreatePR` (`validatePRRequestClaims`, `validatePRRequestContent`, `validateOutreachPRRequest`), each with the same shape: inspect the request/diff, `(reason, err)`, reject with a quarantined `.rejected` file if `reason != ""`. This is a proven, load-bearing choke point an agent cannot route around by using a different CLI — which is exactly what makes a gate here *not* trivially bypassed, unlike, say, a prompt instruction (which is what failed in the reported incident: architect's own policy already tells it to write `Refs #N` for multi-phase work, and it did — that was never the gap).

**The one thing a gate genuinely cannot do:** require a PR to admit where an idea came from. If an agent implements its own proposal while citing nothing, there is no server-side signal linking the PR to the issue at all, and no gate can manufacture one. I say this plainly in the code and CHANGELOG rather than overselling coverage.

Given that, I implemented the gate — the incident's actual PR (`tuna-os/tunaos-packages#583`) *did* cite its issue (`Refs #581`), which is the shape a gate can act on, and matches the issue's own framing: "the issue provides the appearance of a rationale" only works if the PR points at it.

## What it does

A fourth gate, `validateSelfProposalPRRequest` in `pkg/github/pr_request_self_proposal.go`, wired into `handleOnePRRequest` alongside the existing three:

1. Parse the PR title+body for issue references — both closing keywords (`Fixes`/`Closes`/`Resolves`, via the existing `ParseClaimedIssues`) **and** the non-closing form (`Refs`/`Part of`/etc., via the existing `ParseReferencedIssues`). The incident used `Refs #581`, which is architect policy's own documented convention for multi-phase work — a closing-only check would have missed the real case.
2. For each referenced issue, fetch it. If its author is a human, skip — a human-filed proposal already satisfies "there is a tracked rationale," which is the precondition this gate exists to check.
3. If the issue's author is a bot/App account (`isBotAuthor`, reusing the `"[bot]"` suffix / `Type == "Bot"` idiom already established in `findDigestComment`), check whether **any** comment on it comes from a non-bot account.
4. No human comment at all → reject the PR request, naming the unacknowledged issue.

Any human comment satisfies the gate, including a rejection or a question — this checks *provenance* (did a human look at this), not review outcome. Judging the content of a human's comment is deliberately left to the human, not encoded here.

## Reproducing the incident as a test

`TestValidateSelfProposalPRRequest_RejectsUnacknowledgedBotProposal_RefsForm` reconstructs the exact shape: a bot-authored issue, zero comments, a PR citing it with `Refs #581` — and asserts rejection. Plus:

- `_FixesForm` — same, with a closing keyword, so the gate isn't limited to one form.
- `_AllowsAfterAnyHumanComment` — a single human comment (not an approval, just a comment) clears the gate.
- `_BotCommentsDoNotCount` — bot replies to the bot's own proposal (including a *different* automation account) do not satisfy it, so the gate cannot be trivially satisfied by the hive commenting on itself.
- `_AllowsHumanFiledProposal` — a human-filed issue is left alone entirely (no false positive on the common case).
- `_NoReferenceIsANoOp` — a PR citing no issue makes zero API calls and is never rejected (the gate only fires on the shape it targets).
- `TestPRRequestWatcher_QuarantinesSelfProposal` — end-to-end through the actual watcher: request file in, `.rejected` file and no PR created out, matching the existing gates' test pattern (`TestPRRequestWatcher_QuarantinesArtifactClaimMismatch`).

All new, all fail without the fix. Existing `pkg/github` tests pass unmodified (`go test ./pkg/github/...`, ~48s, all green).

## Note on repo state

Per the task brief, `v4` currently fails to build due to a duplicate `TestConfiguredDossierCacheMaxEntries` in `pkg/dashboard` (#5279, unrelated, already has a fix in flight). `pkg/github` itself builds, vets, and tests clean in isolation on top of current `v4`.

Credit to @hanthor for the well-evidenced report (the incident's two GitHub links make the exact failure reproducible).


clubanderson:
Closing in favour of #5280, which fixes the same issue and landed first.

Both carry `Fixes #5117` and both add a server-side gate in the `pr_request_watcher.go` path. @Danathar opened #5280 at 13:26; I opened this at 13:36, and I did not check for an existing fix before dispatching the work — which is exactly the duplicate-PR failure #5111 describes. Not a good look on a PR about agents acting without checking.

#5280 is also the better change: +766 across 5 files versus +362 here, and its body is candid about the gate firing on positive evidence only.

One thing from this PR worth carrying over if it is not already covered, as a review comment on #5280 rather than a competing implementation — the incident (`tuna-os/tunaos-packages#581`→`#583`) cited the issue in **non-closing** `Refs`/`Part of` form, which is architect policy's own documented convention. A gate keyed only on closing keywords would miss the exact shape reported.

No action needed on this branch.

clubanderson:
Superseded by #5280, which merged first and closes #5117 with an equivalent server-side gate (bot-filed proposal issue + no human comment `->` PR rejected at the same `pkg/github/pr_request_watcher.go` choke point). Closing this to avoid duplicate/conflicting implementations of the same gate.

PR #5342, opened 2026-08-31
🔒 fix(entrypoint): harden runtime config recreated by cp at boot (#5331)

### LINKED ISSUE #5331: [sec-check] dashboard owner auth token persisted world-readable in /data/hive.yaml.runtime, .dashboard, and .bak (0644)
## Security Finding

**Severity**: high
**Type**: secret-exposure / permission-issue

`Config.saveLocked()` (src/pkg/config/config.go:5344) persists the full runtime config to `/data/hive.yaml.runtime` with `os.WriteFile(..., 0o644)`, and `saveDashboardOverlay()` (config.go:5482, `overlayFileMode = 0o644`) writes `/data/hive.yaml.dashboard` the same way. The legacy `/data/hive.yaml.bak` is also 0644.

The marshaled config includes `dashboard.auth_token` and, in PAT deployments, `github.token`. `redactedForPersist()` redacts OTel/tracing headers but **not** these fields, and `dashboardOverlayBytes()` only collapses the auth token back to its env form **when it matches a bootstrap env var** — a dashboard-minted token (the common path) is persisted verbatim.

**Verified live on this host** (values not disclosed):
- `/data/hive.yaml.runtime` — mode 0644, `dashboard.auth_token` populated (64 chars)
- `/data/hive.yaml.dashboard` — mode 0644, token populated (64 chars)
- `/data/hive.yaml.bak` — mode 0644, token populated (64 chars)
- `/data` is world-traversable (`drwxr-xr-x`), and agent users (e.g. uid 2009 `hive-sec-check`) can read all three files.

Contrast: `/data/gh-app-key*.pem` and `/data/copilot-user-token` are correctly 0600.

## Impact

Any unprivileged agent user (or any process an agent spawns, including prompt-injected tool runs) can read the dashboard owner auth token and authenticate to the dashboard API with owner privileges — config writes, kicks, ACMM changes, token minting. In PAT-mode deployments `github.token` leaks the same way. This defeats the per-agent token scoping model: the effort to keep scoped tokens out of agent hands (e.g. #1861 / PR #4032) is bypassed by reading the owner credential off disk.

## Recommendation

1. Write `RuntimeConfigFile` and `DashboardOverlayFile` with `0o600` (and `os.Chmod` after the remove-and-retry path, since `os.WriteFile` mode does not apply to a pre-existing inode).
2. Chmod existing `/data/hive.yaml.runtime`, `.dashboard`, `.bak` to 0600 at boot (entrypoint or Save path).
3. Optionally extend `redactedForPersist()` to also fold `dashboard.auth_token` out of layers that don't need it.

---
---

clubanderson:
## Scope — this is the remainder of #5331, not a competing fix

**PR #5335 covers the rest and is correct on its merits.** This PR is stacked
on top of it (base = `sec-check/5331-runtime-config-perms`) so the diff here is
only the one commit. It retargets to `v4` automatically when #5335 merges.

What #5335 already handles, and I am deliberately not touching:
- `saveLocked` — `RuntimeConfigFile` written `0o600` (both the initial write and
  the remove-and-retry), plus explicit `os.Chmod` for the pre-existing-inode case
- `saveDashboardOverlay` — `overlayFileMode` `0o600` and `f.Chmod` on a leftover `.tmp`
- `entrypoint.sh` — boot-time `chmod 600` loop over pre-existing
  `hive.yaml.runtime` / `.bak` / `.dashboard`

Those Go paths create the files `0600` **from the start** (`os.WriteFile` /
`OpenFile` apply the mode at create), so there is no world-readable window in
the write path. I reviewed for that specifically.

## The gap that remains

#5335's chmod loop runs at the **top** of `entrypoint.sh`, before the config
branch. Three `cp` calls further down **recreate** `$HIVE_CONFIG_RUNTIME`:

| line | branch |
|---|---|
| `cp "$HIVE_CONFIG_PATH" "$HIVE_CONFIG_RUNTIME"` | K8s — ConfigMap is the seed |
| `cp "$HIVE_CONFIG_PATH" "$HIVE_CONFIG_RUNTIME"` | Docker first boot |
| `cp "$HIVE_CONFIG_RUNTIME_LEGACY" "$HIVE_CONFIG_RUNTIME"` | legacy `.bak` migration |

`cp` gives a **newly created** destination the **source's** mode. Both sources
(the ConfigMap seed, the bind-mounted `hive.yaml`) are `0644`, so the runtime
copy is re-widened to `0644` on every boot — *after* the hardening loop has
already run.

Net effect without this patch: the `0600` fix only holds from the first
`Config.Save()` onward. A hive that boots and never saves stays world-readable
indefinitely. The runtime config carries `dashboard.auth_token`, and `/data` is
world-traversable, so that is the dashboard owner credential readable by every
unprivileged agent uid — the exact exposure #5331 reports.

## Change

- `hive_harden_runtime_config()` called after each of the three `cp` calls.
  Best-effort (`2>/dev/null || true`), matching the existing loop's contract: a
  read-only or foreign-owned PVC must not abort boot.
- Nothing group-writable is tightened. `/data/home/.codex` and the other
  intentionally shared paths are untouched.

## Remediation for already-deployed hives

This *is* the remediation for the recreate path. #5335 tightens copies that
already exist at boot; this keeps them tightened through the `cp` that would
otherwise undo it on the same boot. Together they converge every existing hive
to `0600` on its next restart, without waiting for a config save.

## Test

`TestEntrypointHardensRuntimeConfigItRecreates` — table-driven over all three
branches, asserting `Mode().Perm() == 0600`. Verified it **fails at `0644`**
against the unhardened script (the harness seeds files `0644`, the production
mode), so it is a real regression guard rather than a tautology. Existing
entrypoint boot tests still pass; the harness gained a `/data` mkdir so the
assertion turns on the mode and not a missing directory.

No token or key material appears in the diff, the test fixtures, or the logs.

Fixes #5331

### DIFF
--- a/src/deploy/entrypoint.sh
+++ b/src/deploy/entrypoint.sh
@@ -70,6 +70,24 @@ for _cfg in "$HIVE_CONFIG_RUNTIME" "$HIVE_CONFIG_RUNTIME_LEGACY" /data/hive.yaml
 done
 unset _cfg
 
+# hive_harden_runtime_config tightens $1 to 0600 after any cp that (re)creates
+# a PVC config copy.
+#
+# The boot-time chmod loop above only covers files that already existed when
+# this script started. Every `cp` below that writes $HIVE_CONFIG_RUNTIME
+# creates the destination anew, and cp gives a newly created destination the
+# SOURCE's mode — the ConfigMap seed and the bind-mounted hive.yaml are both
+# 0644 — so the runtime copy is re-widened to 0644 on every boot, after the
+# loop above has already run. Without this the 0600 fix only holds from the
+# first Config.Save() onward, and a hive that boots and never saves stays
+# world-readable indefinitely.
+#
+# Best-effort for the same reason as the loop above: a read-only or
+# foreign-owned PVC must not abort boot.
+hive_harden_runtime_config() {
+  [ -f "$1" ] && chmod 600 "$1" 2>/dev/null || true
+}
+
 # hive_runtime_config_read echoes the path to read the persisted runtime
 # config from: the new name when it is present and non-empty, else the
 # legacy name when that is, else empty. Read-only — it never creates,
@@ -264,6 +282,7 @@ PYEOF
     # Write the (merged) config as the disaster-recovery snapshot. Always
     # under the new name; the legacy file is left untouched on the PVC.
     cp "$HIVE_CONFIG_PATH" "$HIVE_CONFIG_RUNTIME" 2>/dev/null || true
+    hive_harden_runtime_config "$HIVE_CONFIG_RUNTIME"
     echo "[entrypoint] K8s mode — ConfigMap is the seed, runtime config written to $HIVE_CONFIG_RUNTIME"
   else
     # Neither source exists: no runtime config on the PVC (checked first,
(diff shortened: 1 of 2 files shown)


clubanderson:
Closing in favour of a branch based directly on `v4`. Basing this on #5335's branch meant the repo's `branches: [v2, v4]` workflow filters skipped the Go test suite entirely, so the new permissions regression test never actually ran in CI — not an acceptable way to land a security fix. Reopening from a v4-based branch; see the replacement PR.

PR #5349, opened 2026-08-31
🔒 fix(entrypoint): write PVC runtime config 0600 at boot, including the copies cp recreates (#5331)

### LINKED ISSUE #5331: [sec-check] dashboard owner auth token persisted world-readable in /data/hive.yaml.runtime, .dashboard, and .bak (0644)

clubanderson:
## Overlap with #5335 — please read first

#5335 is open and fixes the **Go write paths** for the same issue, and its
approach is sound: `saveLocked` writes `RuntimeConfigFile` `0o600` on both the
initial write and the remove-and-retry, `saveDashboardOverlay` uses
`overlayFileMode = 0o600` plus `f.Chmod` on a leftover `.tmp`. Those create the
files `0600` **from the start** (`os.WriteFile`/`OpenFile` apply the mode at
create time), so there is no world-readable window. **This PR does not touch
`config.go` at all** — that half is #5335's and I've left it alone.

The two PRs overlap on one thing only: both add a boot-time `chmod` loop to
`entrypoint.sh`. Whichever lands second will need a trivial conflict resolution
there. I based this on `v4` rather than stacking on #5335 because the repo's
`branches: [v2, v4]` workflow filters skip the Go test suite for any other
base — the new regression test would never have run in CI, which is not an
acceptable way to land a security fix.

## The part #5335 does not cover

A boot-time `chmod` alone is **silently reverted later in the same script**.
Three `cp` calls in the config branch *recreate* `$HIVE_CONFIG_RUNTIME`:

| branch | source |
|---|---|
| K8s — ConfigMap is the seed | `$HIVE_CONFIG_PATH` (0644) |
| Docker first boot | `$HIVE_CONFIG_PATH` (0644) |
| legacy `.bak` migration | `$HIVE_CONFIG_RUNTIME_LEGACY` (0644) |

`cp` gives a **newly created** destination the **source's** mode. Both sources
are 0644, so the runtime copy is re-widened to 0644 on every boot — after the
hardening loop at the top has already run.

Consequence: with the write-path fix alone, 0600 only holds from the first
`Config.Save()` onward. **A hive that boots and never saves stays
world-readable indefinitely.** The runtime config carries
`dashboard.auth_token`, and `/data` is world-traversable, so that is the
dashboard owner credential readable by every unprivileged agent uid — the exact
exposure #5331 reports.

## Change

`hive_harden_runtime_config`, called in two places, both required:

1. **Once at boot** over `hive.yaml.runtime` / `.bak` / `.dashboard` that
   already exist — remediation for hives deployed before this fix, so they are
   tightened on their next restart rather than waiting for a save.
2. **After each of the three `cp` calls**, so (1) is not undone on the same boot.

Best-effort (`2>/dev/null || true`) throughout, matching the surrounding
convention: a read-only or foreign-owned PVC must not abort boot. No
intentionally group-shared path (`/data/home/.codex` and friends) is touched —
I checked the surrounding comments before tightening anything.

## Remediation for already-deployed hives

Point 1 above. Every existing hive converges to 0600 on its next restart,
without needing a config save first.

## Test

`TestEntrypointHardensRuntimeConfigItRecreates` — table-driven over all three
recreate branches, asserting `Mode().Perm() == 0600`. Verified it **fails at
`0644`** against the unhardened script (the harness seeds files 0644, the
production mode), so it is a real regression guard and not a tautology.
Existing entrypoint boot tests still pass; the harness gained a `/data` mkdir
so the assertion turns on the mode rather than a missing directory.

No token or key material appears in the diff, the fixtures, or the logs.

## Still open after this

Item 3 of the issue's recommendation — folding `dashboard.auth_token` out of
`redactedForPersist()` — is addressed by neither PR. Worth tracking separately:
defense-in-depth on top of 0600 is a different change with a real regression
surface, since the token has to survive a restart.

Fixes #5331

### DIFF
--- a/src/deploy/entrypoint.sh
+++ b/src/deploy/entrypoint.sh
@@ -61,6 +61,32 @@ HIVE_CONFIG_PATH="${HIVE_CONFIG:-/etc/hive/hive.yaml}"
 HIVE_CONFIG_RUNTIME="/data/hive.yaml.runtime"
 HIVE_CONFIG_RUNTIME_LEGACY="/data/hive.yaml.bak"
 
+# hive_harden_runtime_config tightens $1 to 0600.
+#
+# The PVC config copies carry dashboard.auth_token (and github.token in PAT
+# mode) and /data is world-traversable, so a 0644 copy hands the dashboard
+# owner credential to every unprivileged agent uid on the host (#5331).
+#
+# This is called in two places, and both are needed:
+#
+#   1. Once at boot over the copies that ALREADY exist, so hives deployed
+#      before the 0600 fix are remediated rather than waiting for a save.
+#   2. After every `cp` below that (re)creates $HIVE_CONFIG_RUNTIME. cp gives
+#      a newly created destination the SOURCE's mode, and both sources (the
+#      ConfigMap seed, the bind-mounted hive.yaml) are 0644 — so without this
+#      the file is re-widened to 0644 on every boot, undoing (1) on the very
+#      same boot and leaving a hive that never saves world-readable forever.
+#
+# Best-effort: on a read-only or foreign-owned PVC this must not abort boot.
+hive_harden_runtime_config() {
+  [ -f "$1" ] && chmod 600 "$1" 2>/dev/null || true
+}
+
+for _cfg in "$HIVE_CONFIG_RUNTIME" "$HIVE_CONFIG_RUNTIME_LEGACY" /data/hive.yaml.dashboard; do
+  hive_harden_runtime_config "$_cfg"
+done
+unset _cfg
+
 # hive_runtime_config_read echoes the path to read the persisted runtime
 # config from: the new name when it is present and non-empty, else the
 # legacy name when that is, else empty. Read-only — it never creates,
@@ -255,6 +281,7 @@ PYEOF
     # Write the (merged) config as the disaster-recovery snapshot. Always
     # under the new name; the legacy file is left untouched on the PVC.
     cp "$HIVE_CONFIG_PATH" "$HIVE_CONFIG_RUNTIME" 2>/dev/null || true
+    hive_harden_runtime_config "$HIVE_CONFIG_RUNTIME"
     echo "[entrypoint] K8s mode — ConfigMap is the seed, runtime config written to $HIVE_CONFIG_RUNTIME"
   else
     # Neither source exists: no runtime config on the PVC (checked first,
@@ -278,6 +305,7 @@ else
   if [ -f "$HIVE_CONFIG_PATH" ] && [ -s "$HIVE_CONFIG_PATH" ] && [ -z "$HIVE_CONFIG_SOURCE" ]; then
     # First boot: config exists but no PVC runtime config yet — seed it
     cp "$HIVE_CONFIG_PATH" "$HIVE_CONFIG_RUNTIME"
+    hive_harden_runtime_config "$HIVE_CONFIG_RUNTIME"
     echo "[entrypoint] First boot — config seeded to PVC: $HIVE_CONFIG_RUNTIME"
   elif [ -f "$HIVE_CONFIG_PATH" ] && [ -s "$HIVE_CONFIG_PATH" ] && [ -n "$HIVE_CONFIG_SOURCE" ]; then
     # The PVC runtime config is the source of truth (updated by Save()).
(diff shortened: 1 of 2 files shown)


clubanderson:
CI note: the `v2 Tests` suite (the only place `go test ./pkg/...` runs) did not auto-attach to this PR — closing the earlier stacked PR left its `pull_request` run in flight on the old branch and GitHub did not start a fresh one here. Dispatched it manually against this branch: [run 33430630181](https://github.com/kubestellar/hive/actions/runs/33430630181), all 11 test jobs green, with `ok github.com/kubestellar/hive/pkg/config 17.153s` confirming the package holding the new permissions test compiled and passed. Worth a re-run by a maintainer on the merge commit if the auto-attach doesn't recover.

Merge-order note: this and #5335 both add a boot-time chmod to `entrypoint.sh`, so whichever lands second needs a small conflict resolution in that one spot. The rest is disjoint — this PR does not touch `config.go`. If #5335 goes first, the resolution is to keep this PR's `hive_harden_runtime_config` helper and drop the duplicate inline loop, since the helper is also called after the three `cp` sites, which is the part #5335 doesn't cover.

clubanderson:
Closing as a duplicate of #5342, which is now the healthier of the two.

You closed #5342 because, based on `sec-check/5331-runtime-config-perms`, the workflows' `branches: [v2, v4]` filter meant the Go suite never attached — correct reasoning at the time.

That condition no longer holds. #5335 merged, #5342 auto-closed when its parent branch was deleted, and I reopened and rebased it onto `v4`. It now carries **53 check-runs including 9 Go test shards**, so it gets the full suite this fix needs.

The two PRs are byte-equivalent in scope — identical file sets (`src/deploy/entrypoint.sh`, `src/pkg/config/entrypoint_boot_test.go`) — so there is nothing in this PR that #5342 lacks. #5349 is also currently `DIRTY` while #5342 is green-pending, which makes the choice straightforward.

Your analysis stands and is what makes the fix correct: #5335's Go paths apply 0600 at create time with no world-readable window, but its boot-time chmod loop runs *before* the three `cp` calls that recreate `/data/hive.yaml.runtime` from 0644 sources — so a hive that boots and never calls `Config.Save()` stayed world-readable. That is the remainder both PRs fix.

Also noting your item 3 for follow-up: folding `dashboard.auth_token` out of `redactedForPersist()` is addressed by neither PR and needs separate tracking, since the token must survive a restart.

Issue #5339, opened 2026-08-31
🐛 release.yml is registered with an unparsed definition — emits log-less failing push runs


clubanderson:
## Problem

GitHub has `release.yml` registered with an **unparsed definition**, even though the file declares a name on line 1 and is valid YAML. It is the only workflow in the repository in this state.

```
$ gh api repos/kubestellar/hive/actions/workflows
  Build and Push Docker Image   state=active  path=.github/workflows/docker.yml
  Release Line Guard            state=active  path=.github/workflows/release-line-guard.yml
  Release Lines Gate            state=active  path=.github/workflows/release-lines-gate.yml
  .github/workflows/release.yml state=active  path=.github/workflows/release.yml    <-- fallback name
```

Every healthy workflow reports its declared `name:`. `release.yml` reports its **path**, which is the fallback GitHub uses when it cannot resolve the definition. Meanwhile:

```
$ head -1 .github/workflows/release.yml
name: Tagged Release
```

and the file parses cleanly with `yaml.safe_load`.

## Observable symptom

The repository is emitting `push`-event runs of `release.yml` that fail instantly with **zero jobs** and no logs, despite the workflow declaring only `workflow_run`:

```
18:39  push          name=".github/workflows/release.yml"  failure
18:38  push          name=".github/workflows/release.yml"  failure
18:36  push          name=".github/workflows/release.yml"  failure
18:36  workflow_run  name="Tagged Release"                 skipped
18:34  workflow_run  name="Tagged Release"                 skipped
```

The correlation is exact: runs carrying the fallback name are `push`-triggered and fail with no jobs; runs carrying `Tagged Release` are `workflow_run` and behave normally. GitHub's UI reports "This run likely failed because of a workflow file issue."

These appear whenever a branch touches `release.yml` — observed on PR branches for #5334, #5332, and #5329.

## Why it matters

1. **It manufactures false failures.** A red, log-less run against `release.yml` is indistinguishable at a glance from a genuine release failure. It cost real triage time today, and it is very likely the source of the "log-less failed runs" that #5324 attributed to scratch-branch deletion.
2. **It may not be cosmetic.** A definition GitHub cannot fully resolve is a poor foundation for the one workflow that holds `contents: write` and cuts releases. Whether the `push` runs are the *only* consequence is unverified.

## Not caused by the recent fix

[...]

Issue #4965, opened 2026-08-28
Fork PRs can never satisfy the required `gate` check — six contributor PRs are structurally unmergeable


clubanderson:
`v4` branch protection requires exactly one status context: **`gate`**. Fork PRs cannot produce it, so **no fork PR can ever be merged without an admin bypass.**

## The mechanism

- Required contexts on `v4`: `["gate"]` (`enforce_admins: false`)
- `gate` is a job in `.github/workflows/docker.yml`
- That workflow triggers on **`push` only** — `branches: ['**', '!dependabot/**', '!copilot/**']`, plus `workflow_dispatch`. **There is no `pull_request` trigger.**

A fork PR's head lives in the contributor's repo, so the push event fires there, not here. `gate` never attaches to the head SHA this repo evaluates, and protection blocks the merge waiting for a check that cannot arrive.

Verified `gate: ABSENT` on the head SHA of all six currently-open fork PRs: #4930, #4932, #4934, #4935, #4936, #4937. All six are otherwise green — 30 checks passing, 0 failing, DCO passing.

## Why this has not been noticed

**Maintainers with admin silently bypass it.** `enforce_admins` is `false`, so an admin merge succeeds without the required check. Checking merged same-repo PRs confirms `gate` was `ABSENT` on their head SHAs too — they merged because of the bypass, not because the gate passed.

So the required check is, in practice, **not enforced for maintainers and absolutely enforced for outside contributors.** That is the inverse of what a required check is for, and it means the protection has been providing less assurance than it appears to.

## Impact

An outside contributor sees a green PR that will not merge, with no actionable reason. The only paths available today are an admin override on their behalf — which defeats the protection and hides the gap — or the PR sitting indefinitely.

We currently have a contributor with six such PRs.

## Options

1. **Add a `pull_request` trigger to `docker.yml`.** Fork PRs then produce `gate` legitimately. Cost: image builds on every PR, including forks — worth checking against the existing "exclude bot branches so dependency PRs don't consume image CI" reasoning in that file's own comment.
2. **Stop requiring `gate`** and require a context that fork PRs can produce (the Go test suite already runs on `pull_request`).
3. **Split the gate** — keep the heavyweight image build push-only, and require a lightweight compile check that runs on `pull_request`.

Option 3 preserves the intent (prove it compiles before merge) without making every fork PR build images.

## Not a workaround

[...]

PR #5358, opened 2026-08-31
🐛 fix(release): stop the release PR superseding its own green gate check

clubanderson:
## The 7th failure mode — found by v4.0.1 actually reaching the merge step

With #5354 landed, the release pipeline ran end-to-end for the first time since 2026-08-27. `decide` and `precheck` both passed, and the run then failed at the merge step:

```
##[error]PR #5355 still not mergeable 120s after gate succeeded on c4c5f3db —
this is no longer a settling-time artifact.
```

Every one of ~13 attempts returned `Required status check "gate" is expected` (HTTP 405) — while `gate` was `completed/success` on that exact commit, from `app_id 15368`, and `gate` is the **only** required context on `v4`. Following the step's own advice and checking the required contexts against the commit's check-runs is what surfaced the cause.

### Root cause

`tagged-release.yml` earns `gate` by dispatching `docker.yml` on the `release-gate/v<version>` scratch branch, then opens a PR from that branch into `v4`. **Opening that PR fires `docker.yml` a second time on the same SHA** under its `pull_request` trigger — the trigger #4965 added so fork PRs can earn `gate` at all.

That second run is created by the job's `GITHUB_TOKEN`, so GitHub's recursive-workflow guard creates the run but never starts its jobs. It completes as `failure` with **zero jobs** about two seconds later:

```
19:56:11  suite 90594836071  success    <- dispatched run, the green gate
19:56:24  release PR #5355 opened
19:56:27  10 pull_request suites        <- all failure, all zero jobs
```

Branch protection resolves a required context to the **most recent** check-run of that name on the commit. So the instant failure superseded the green `gate` earned thirteen seconds earlier, and the merge was refused for the entire retry window — unrecoverably, because nothing will ever re-run the superseding suite.

Verified this is not new: the same pattern is present on the 16:53 attempt for release PR #5319 (`bb633caa` — green suite at 16:52:35, failing suites at 16:53:11). It was simply masked by the `tide`/`mergeStateStatus` failure #5334 fixed. It is also, incidentally, the real source of some of the "log-less failing runs" #5324 attributed to the scratch branch being deleted mid-run — these runs are finalized at deletion time, which is why their `updated_at` lands minutes after creation.

### The fix

`docker.yml`'s `gate` job now skips `pull_request` events whose head branch is `release-gate/*`, so the release PR never manufactures a competing check-run and the dispatched run's green `gate` stays authoritative.

Skipping cascades safely: the `build*` jobs already skip on `pull_request` via their own `if:`, and the `merge*` jobs gate on `needs.gate.outputs.push`, which is empty for a skipped job. So **no `gate` check-run is produced at all** for these PRs, rather than a failing one.

**A job-level `if:` rather than a trigger filter, deliberately.** `.github/release-lines.yml` pins `docker.yml` as `unpinned: '**'`, so adding a `branches:` filter to the `pull_request` trigger would fail the release-line guard. That guard inspects only `branches:`/`branches-ignore:`, so a job condition is invisible to it — confirmed by running it.

**The #4965 fork-PR contract is unaffected**: a fork PR's head branch is never `release-gate/*`, so fork PRs still earn `gate` exactly as before.

### Verification

- `check-release-lines.sh` — PASS (the `unpinned: '**'` contract holds)
- `check-no-image-attestations.sh` — PASS
- `check-action-pins.sh` — PASS
- `test-release-push-retry.sh` — PASS, with a new pin asserting the condition, verified to fail when the condition is removed

The RACES header gains a note explaining the mechanism, so a future change that makes the release PR trigger a required check again is recognizable rather than looking like a green gate protection inexplicably refuses to see.

Refs #5339


clubanderson:
**CI note: the one red check is a pre-existing `v4` regression, not this PR.**

`arm64 image pull and service startup` fails here, and it fails identically on `v4` itself:

```
podman-arm64-lane.yml on v4
  19:33:53  d781297c  success   <- last good
  19:40:36  265510f9  failure   <- #5352 merge commit
  19:46:58  fb38361a  failure   <- current v4 tip
```

The failure is `open /data/hive.yaml.runtime: permission denied`, so the service never answers `/api/health`. This PR changes a workflow `if:` condition, a test script, a CHANGELOG entry and a comment — nothing in the image, entrypoint, or config handling.

Deterministic, not flake: it reproduced on an explicit re-run. The rootless and rootful podman lanes pass on the same commits, so it is arm64-specific. Filed as #5360 with the bisect and analysis.

The lane is not a required context on `v4` (protection lists `gate` only), which is how the regression merged in the first place.

PR #5363, opened 2026-08-31
🐛 fix(release): re-earn gate after opening the release PR (#5356)

### LINKED ISSUE #5356: 🐛 Release merge rejected: gate is a check-run on a scratch branch, protection wants a status on v4

clubanderson:
## What breaks

The release workflow reaches its merge step and is rejected:

```
{"message":"Required status check \"gate\" is expected.","status":"405"}
```

v4.0.1 has not cut since 2026-08-27.

## The mechanism — and a correction to the recorded diagnosis

Opening the release PR fires **every** `pull_request`-triggered workflow on the same SHA. Those runs are created by `GITHUB_TOKEN`, so GitHub's recursion guard never starts their jobs: each concludes `failure` in ~2s having scheduled **zero jobs**, leaving a check-**suite** that contains no check-**runs**. Branch protection resolves a required context against the most recent check-suite, so with an empty suite newest, `gate` reads as **missing**.

Note the error says `is expected` — missing — not "failed". That is the tell.

**This is not supersession.** A zero-job run publishes no check-runs at all. On release PR #5361's head `4d6510fc` there is exactly **one** `gate` check-run and it is `success` — nothing overwrote it. Measured on that SHA:

```
20:22:25Z  suite 90601686745  success  runs=10   <-- the green gate
20:22:42Z  suite 90601762369  failure  runs=0    docker.yml
20:22:42Z  suite 90601762476  failure  runs=0    changelog-reminder.yml
20:22:42Z  suite 90601762541  failure  runs=0    podman-arm64-lane.yml
20:22:42Z  suite 90601762656  failure  runs=0    docs-index-reminder.yml
20:22:42Z  suite 90601762685  failure  runs=0    podman-rootless-lane.yml
20:22:42Z  suite 90601762695  failure  runs=0    quadlet-gate.yml
20:22:42Z  suite 90601762699  failure  runs=0    podman-rootful-lane.yml
20:22:43Z  suite 90601764053  failure  runs=0    greetings.yml
20:22:43Z  suite 90601764248  failure  runs=0    copilot-dco.yml
20:22:43Z  suite 90601764285  failure  runs=0    copilot-automation.yml
```

## Why the previous fix could not work, and why the obvious next one cannot either

**#5358's job-level `if:`** is evaluated by the runner *after* the run exists. These runs never schedule a job, so it is unreachable by construction.

**A docker.yml-only change of any kind** cannot work: nine other workflows leave an identical empty suite in the same second, and any one left newest reproduces the 405.

**A trigger-layer filter is not available.** `pull_request`'s `branches`/`branches-ignore` matches the PR's **base** ref (`v4`), never its head, and GitHub has no head-branch trigger filter ([docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpull_requestpull_request_targetbranchesbranches-ignore)). So `branches-ignore: ['release-gate/**']` would exclude PRs *into* the scratch branch — not the release PR — while breaking the fork-PR contract (#4965) that trigger exists to serve.

## The fix

After opening the PR, let the burst settle, then **re-dispatch `docker.yml`** so a green suite is the most recent one again. `gate` is a ~5s shell job and `release-gate/*` never pushes an image, so this is cheap and publishes nothing.

The gate is not weakened: the merge stays SHA-keyed and the server still rejects a genuinely unsatisfied required context with 405.

This also removes the race noted on the issue — it does not depend on the two original runs ordering correctly, because it adds a third, later, green one.

## Robustness

An unconfirmed or unparseable re-gate is a **warning, not a failure**, so a transient API hiccup cannot abandon a mergeable release; only a real red conclusion (`failure`/`cancelled`/`timed_out`/`action_required`/`stale`) aborts. Timings sit behind `RELEASE_REGATE_SETTLE` / `RELEASE_REGATE_WINDOW` seams, matching the existing `RELEASE_PUSH_GH006_WINDOW` pattern.

## Release-lines pin contract

Untouched and still green. No trigger filter was added, so docker.yml keeps its `unpinned: '**'` contract; `check-release-lines.sh` and `test-release-lines-guard.sh` both pass.

## #5358's dead guard

Kept, comment corrected. A `release-gate/*` head is never a real code review, so skipping the build is still the right behaviour — but the comment no longer claims to fix this, and now records why it cannot. `test-release-push-retry.sh` drops the assertion that pinned that guard as the remedy and instead asserts the re-dispatch happens **after** `gh pr create` (before it, the green suite lands back underneath the empty ones and the 405 returns).

## Preserved

Gate-wait SHA-keyed lookup, tag-already-exists refusal, `pushed=false` deferral, `RELEASE_PUSH_GH006_WINDOW`, the backstop's published-images guard, and the fork-PR contract.

## Verified

`check-release-lines.sh` PASS; `test-release-lines-guard.sh` PASS; `test-release-push-retry.sh` PASS; both workflows parse; no empty `${​{ }}` anywhere.

## Unproven

**This cannot be proven until a release actually cuts.** The mechanism and the ten empty suites are measured from real API data, but the fix itself is verified only against the local test harness and static guards. Specifically unproven: that 30s is always enough for the PR-open burst to land (if a suite arrives *after* the re-dispatch, the 405 returns — the merge retry window is the backstop, not a guarantee), and that no other actor creates a newer empty suite between the re-gate and the merge.

Fixes #5356

### DIFF
--- a/.github/workflows/tagged-release.yml
+++ b/.github/workflows/tagged-release.yml
@@ -103,6 +103,24 @@ name: Tagged Release
 #     they now skip and the suite completes when `gate` does. If a job is ever
 #     added to docker.yml that runs on the scratch branch, it will re-block
 #     this merge in exactly the same invisible way.
+#   - OPENING THE RELEASE PR HIDES ITS OWN GREEN GATE (#5339 7th, #5356 8th).
+#     Opening the PR below fires every `pull_request`-triggered workflow on the
+#     SAME SHA. Those runs are created by GITHUB_TOKEN, so GitHub's recursion
+#     guard never starts their jobs: each fails in ~2s having scheduled ZERO
+#     jobs, leaving a check-SUITE with no check-RUNS in it. Protection resolves
+#     a required context against the most recent suite, so with an empty suite
+#     newest, `gate` reads as MISSING — `Required status check "gate" is
+#     expected` (405) for the whole retry window.
+#
+#     Note the green `gate` is NOT overwritten: a zero-job run publishes no
+#     check-runs, and the one `gate` check-run on the SHA stays `success`. It
+#     is simply not in the suite protection consults. #5358's job-level `if:`
+#     in docker.yml therefore fixed nothing (the runner never reaches a job),
+#     and no docker.yml-only change could — nine other workflows leave an
+#     identical empty suite in the same second. The step below instead
+#     RE-DISPATCHES docker.yml once the PR exists, making a green suite newest
+#     again. If a future change stops that re-dispatch, this failure returns
+#     looking exactly like a green gate that protection refuses to see.
 #
 # IDEMPOTENCY: this workflow's own release commit (moving Unreleased into a
 # dated section) is what empties Unreleased, which re-triggers docker.yml on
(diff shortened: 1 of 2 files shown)


clubanderson:
### Note on the two red checks — both pre-existing / test artifacts, neither from this change

**1. `arm64 image pull and service startup` — pre-existing, red on v4 already.**

The podman arm64 lane fails with a rootless iptables permission error:

```
[entrypoint] WARN: iptables chain creation attempt 1/5 failed: iptables v1.8.11 (nf_tables):
Could not fetch rule set generation id: Permission denied (you must be root) — retrying
```

It is red on v4 independently of this PR, including on `e9b8c145` (the #5358 merge commit):

```
2026-08-31T20:13:20Z  failure  e9b8c145
2026-08-31T19:46:58Z  failure  fb38361a
2026-08-31T19:40:36Z  failure  265510f9
2026-08-31T19:33:53Z  success  d781297c
```

This PR touches only `tagged-release.yml`, comments in `docker.yml`, and `test-release-push-retry.sh` — nothing reachable from that lane.

**2. `release` — an artifact of a manual dispatch I ran to prove the workflow still parses.**

I dispatched `tagged-release.yml` against this branch as the mandated 422 parse check. It registered and ran (no 422), then failed at step 16 — **`Push release commit to a scratch branch and wait for gate`**, which is pre-existing code, *before* reaching any line this PR adds (step 17 shows `skipped`):

```
! [remote rejected] HEAD -> release-gate/v4.0.1
  (refusing to allow a GitHub App to create or update workflow
   `.github/workflows/docker.yml` without `workflows` permission)
```

That is `GITHUB_TOKEN` refusing to push a branch whose commits modify workflow files — and this branch modifies `docker.yml` and `tagged-release.yml`. **A real release cannot hit this**: the release commit is `git add CHANGELOG.md` followed by a single commit (tagged-release.yml:484-485), so it never contains a workflow file.

The dispatch therefore establishes the parse/registration check and nothing more. It does not exercise the new re-dispatch logic, which remains covered only by `test-release-push-retry.sh` and unproven end-to-end until a release actually cuts — as stated in the description.

No tag was pushed by that run (`v4.0.1` still does not exist), and steps 17/18 were skipped, so it left no side effects.

PR #5368, opened 2026-08-31
🐛 fix(entrypoint): chown the runtime config to dev so hive can read it (#5360)

### LINKED ISSUE #5360: 🐛 arm64 startup lane broken on v4: hive.yaml.runtime permission denied after #5352
## Summary

The **Podman arm64 Lane** has been failing on `v4` since `265510f9` (#5352, "Install the git credential helper system-wide"). It is not a required status context, so the breakage merged and has stayed red on every subsequent commit.

Found while verifying an unrelated release PR (#5358), whose only changes are a workflow `if:` condition, a test script, and CHANGELOG — the lane fails identically on `v4` itself, so it is not caused by that PR.

## Evidence

```
podman-arm64-lane.yml on v4
  19:33:53  d781297c  success   <- last good
  19:40:36  265510f9  failure   <- #5352 merge commit
  19:46:58  fb38361a  failure
```

Failure is deterministic — it reproduced on an explicit re-run, so it is not flake.

## Symptom

```
[entrypoint] Config path is read-only — using /data/hive.yaml.runtime directly
[entrypoint] Config path pinned for the Go binary: --config /data/hive.yaml.runtime
[entrypoint] Starting Go binary on :3002 (uid=1001)
{"level":"ERROR","msg":"failed to load config",
 "error":"reading config /data/hive.yaml.runtime: open /data/hive.yaml.runtime: permission denied"}

RESULT: FAIL — the service did not answer /api/health within 180s
```

The service never becomes healthy because the Go binary cannot read the runtime config it was just pointed at.

## Why #5352 is the suspect

#5352 modifies `hive_runtime_config_read()` in the entrypoint — the same function whose read fails here — and it lands directly on top of #5342 (`🔒 fix(entrypoint): harden runtime config recreated by cp at boot`), which changed the permissions of that very file. The two changes interact: the hardened mode is evidently not readable by the dropped-to UID on this lane.

## Why arm64 only

The rootless and rootful podman lanes pass on the same commits:

```
podman-rootless-lane  fb38361a  success
podman-rootful-lane   fb38361a  success
podman-arm64-lane     fb38361a  failure
```

So this is arm64-lane-specific — likely a difference in how that lane mounts or stages `/data`, meaning the ownership the hardened `cp` produces differs there. Worth confirming rather than assuming.

## Note on gating

This lane is not in `v4`'s required contexts (protection lists `gate` only), which is exactly why a startup regression reached the default branch unnoticed. That is the same "a guard that fails green is worse than no guard" shape as #4339 — worth deciding whether this lane should be required, separately from fixing the regression itself.

clubanderson:
## The bug

Hive fails to start, deterministically, on the Podman arm64 lane:

```json
{"level":"ERROR","msg":"failed to load config",
 "error":"reading config /data/hive.yaml.runtime: open /data/hive.yaml.runtime: permission denied"}
```

Regression window is #5342 (`fe853992`, 19:33), confirming the analysis in the issue. #5352 is not involved.

## Mechanism

`hive_harden_runtime_config()` did `chmod 600` with no `chown`. Its three call sites sit at entrypoint.sh:426/450/467 — inside the config branch, which runs **as root**, before the privilege drop. `cp` creates the destination anew, so the file lands **root:root 0600**. The process then drops to `dev` (uid 1001) at ~line 1300 via setpriv/gosu and cannot open its own config. `0600` grants access to the owner alone, and the owner is not the reader.

## Why `chown -R dev:node /data` did not cover this

Two independent reasons — this was the part worth pinning down:

```sh
DATA_OWNER=$(stat -c '%u' /data 2>/dev/null || echo "0")
if [ "$DATA_OWNER" != "1001" ]; then
  chown -R dev:node /data 2>/dev/null || true
fi
```

1. **The guard is false, so it never runs.** `src/Dockerfile:562` already does `chown -R dev:node /data` at build time. Podman creates the arm64 probe's `/data` as an anonymous volume seeded from the image, so `/data` is *already* uid 1001 at boot, `DATA_OWNER == 1001`, and the recursive chown is skipped entirely. It is not failing silently — it is not executing.
2. **It is in the wrong place anyway.** It sits ~60 lines *below* the config branch, so even when it does run it cannot help a file that branch has not created yet.

Note that the guard is a **latent hole beyond this file**: it exists for a good reason (recursive chown over an NFS PVC is a multi-minute stall), but its effect is that on any already-dev-owned `/data`, *anything* root creates keeps root ownership indefinitely. Any future root-phase write under `/data` inherits this bug. Not fixed here — flagging it as worth its own issue.

## Why only arm64 caught it

`podman-arm64-lane.yml` is the only lane that actually boots hive and waits for `GET /api/health`. The rootful and rootless lanes run `probe_podman_rootful_netadmin.sh` / the rootless equivalent, which probe NET_ADMIN capability behaviour and never assert startup. So the other two lanes were never able to see this, and they are unaffected by the change.

## The fix — shape 1 (chown, keep 0600)

Chose chown-to-`dev:node` over `0640`+group because nothing other than the hive process reads this file; `0640` would widen access to every member of `node` for no benefit. The mode stays `0600`, so **#5331/#5342 are not weakened** — the file still holds `dashboard.auth_token` and is still owner-only.

- `chown` first, `chmod` second: the reverse order leaves a window at 0644, and chown does not clear the mode.
- Applied at **all four sites** by routing the boot-time chmod loop through the same helper, so a pre-existing root-owned copy is repaired too and no path can diverge.
- **Fails open, not shut:** if the chown cannot be performed (no `CAP_CHOWN`, foreign-owned PVC) the chmod is deliberately skipped and a WARN is printed rather than locking a file we do not own to 0600 — that exact combination is this bug. A readable-but-wider file that boots beats a hardened one that cannot.

## Tests — closing the gap that let this ship

`test_entrypoint_runtime_config.sh` asserted the **mode** and therefore passed while the product was broken. It now asserts the property the product actually needs:

- root creates the file 0644 (the mode `cp` inherits), helper runs, then **`su dev -c cat`** — a real `open()` as uid 1001, the same syscall that returned EACCES.
- an unrelated uid (`nobody`) still **cannot** read it — proves readability was bought with ownership, not by widening the mode.
- the helper chowns, and no site chmods a config copy outside the helper.
- the already-dev-owned steady state is still tightened to 0600.

The behavioural cases need root plus a uid-1001 account; they **skip loudly** where those are absent rather than faking a pass, since a silent skip is how the original gap shipped.

## What I verified, and what I did not

**Verified:** the mechanism, by reading the code paths end to end — the ownership guard, the call-site ordering relative to the privilege drop, and the image's build-time `/data` chown that makes the guard false. Shell syntax (`bash -n`). The test suite passes (11/11, root-only cases skipped). The chown-fails fallback leaves the mode at 0644 rather than 0600, confirmed directly.

**NOT verified: the arm64 lane has not gone green with this change yet.** That is the only check that can actually confirm the fix, and it runs on this PR. I could not reproduce locally — the rootful/rootless lanes cannot exercise startup, and the local container VM is broken on this machine, so a green local run would have proven nothing regardless. **Do not merge on my say-so; merge on the arm64 lane going green.** Also unverified locally: the root-only test cases, which run for the first time in CI.

## Secondary: the iptables warnings

```
iptables ... Could not fetch rule set generation id: Permission denied (you must be root)
```

**Separate from this bug, and not a symptom of it.** I initially guessed they were downstream of the config failure; the CI log on this PR disproves that — they are emitted *before* the config error, so they cannot be caused by it:

```
[entrypoint] ERROR: iptables chain creation failed after 5 attempts: ...
{"level":"ERROR","msg":"failed to load config", ... permission denied}
```

The "0 occurrences on green, 5 on red" comparison in the issue does not hold up either: the probe only prints `Starting|WARN: proxy egress` milestones on a successful run and dumps the full log on failure, so the green run's silence is log filtering, not absence. These warnings are most likely present on green runs too — rootless Podman simply cannot manage nf_tables — and are non-fatal by design (5 retries, then continue). They are unrelated to the config read and are not fixed here. Worth a separate issue to either grant the capability or stop retrying five times for something that cannot succeed in this lane.

Fixes #5360

### DIFF
--- a/src/deploy/entrypoint.sh
+++ b/src/deploy/entrypoint.sh
@@ -61,32 +61,71 @@ HIVE_CONFIG_PATH="${HIVE_CONFIG:-/etc/hive/hive.yaml}"
 HIVE_CONFIG_RUNTIME="/data/hive.yaml.runtime"
 HIVE_CONFIG_RUNTIME_LEGACY="/data/hive.yaml.bak"
 
-# These PVC config copies can carry dashboard.auth_token (and github.token in
-# PAT mode). Files written before the 0600 fix (#5331) are world-readable, and
-# /data is world-traversable, so tighten pre-existing copies at boot.
-# Best-effort: on a read-only or foreign-owned PVC this must not abort boot.
-for _cfg in "$HIVE_CONFIG_RUNTIME" "$HIVE_CONFIG_RUNTIME_LEGACY" /data/hive.yaml.dashboard; do
-  [ -f "$_cfg" ] && chmod 600 "$_cfg" 2>/dev/null || true
-done
-unset _cfg
+# The uid/gid the hive process actually runs as after the privilege drop
+# further down (setpriv/gosu to dev). 0600 is an OWNER-only mode, so every
+# hardened config copy has to be owned by THIS user or the owner of the file
+# is not the reader of the file — see hive_harden_runtime_config.
+HIVE_RUNTIME_USER="dev"
+HIVE_RUNTIME_GROUP="node"
 
-# hive_harden_runtime_config tightens $1 to 0600 after any cp that (re)creates
-# a PVC config copy.
+# hive_harden_runtime_config makes $1 readable by the user that reads it, and
+# by nobody else: chown to dev:node, then chmod 0600.
 #
-# The boot-time chmod loop above only covers files that already existed when
-# this script started. Every `cp` below that writes $HIVE_CONFIG_RUNTIME
+# BOTH halves are load-bearing, and #5360 is what happens with only one.
+#
+# The mode half (#5342/#5331): every `cp` below that writes a PVC config copy
 # creates the destination anew, and cp gives a newly created destination the
 # SOURCE's mode — the ConfigMap seed and the bind-mounted hive.yaml are both
-# 0644 — so the runtime copy is re-widened to 0644 on every boot, after the
-# loop above has already run. Without this the 0600 fix only holds from the
-# first Config.Save() onward, and a hive that boots and never saves stays
-# world-readable indefinitely.
+# 0644 — so the copy is re-widened to 0644 on every boot. Without the chmod
+# the 0600 fix only holds from the first Config.Save() onward, and a hive that
+# boots and never saves stays world-readable indefinitely. These files carry
+# dashboard.auth_token (and github.token in PAT mode) and /data is
+# world-traversable, so that is the dashboard owner credential readable by
+# every unprivileged agent uid.
 #
-# Best-effort for the same reason as the loop above: a read-only or
-# foreign-owned PVC must not abort boot.
+# The ownership half (#5360): those same `cp` calls run in the ROOT phase, so
+# the destination is created root:root. chmod 600 on a root:root file grants
+# access to root ALONE — and the hive process drops to dev (uid 1001) before
+# it ever opens the config, so it reads back `permission denied` and startup
+# aborts. The `chown -R dev:node /data` in the root-only block does NOT cover
+# this: it is guarded by `[ "$DATA_OWNER" != "1001" ]` and src/Dockerfile
+# already ships /data owned by dev:node, so on a fresh anonymous volume the
+# guard is false and the recursive chown never runs at all. It is also far
+# BELOW these call sites, so even when it does run it cannot help a file the
+# config branch has not created yet.
+#
+# Do the chown FIRST and the chmod second. The reverse order leaves a window
+# in which the file is 0644 and root-owned; chown does not clear the mode, so
+# chown-then-chmod is never wider than 0600 for longer than the chown itself.
+#
+# Best-effort throughout: a read-only or foreign-owned PVC, or a container
+# without CAP_CHOWN, must not abort boot. When the chown cannot be performed
+# the chmod is deliberately skipped rather than applied to a file we do not
+# own — locking a root-owned file to 0600 is precisely the #5360 failure, and
+# a readable-but-wider file that boots beats a hardened one that cannot.
 hive_harden_runtime_config() {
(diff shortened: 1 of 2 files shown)


clubanderson:
The arm64 lane is red on this PR, and it will stay red until this merges. That is a property of the lane, not a verdict on the fix.

`probe_arm64_image_startup.sh` **pulls `ghcr.io/kubestellar/hive:v4-latest`** — the already-published image — and deliberately does not build one ad hoc (documented in the workflow header: "a missing arm64 manifest is reported and fails the job, because the fix for that belongs in the publisher"). So the container it started here is the pre-fix image and reproduces the pre-fix bug verbatim:

```
[entrypoint] ERROR: iptables chain creation failed after 5 attempts: ...
{"level":"ERROR","msg":"failed to load config","error":"reading config /data/hive.yaml.runtime: open /data/hive.yaml.runtime: permission denied"}
  GET /api/health -> 000 (after 180s)
  container state: exited (exit 1)
```

`entrypoint.sh` is baked into the image, so no source change to it can turn this lane green pre-merge. The confirmation sequence is: merge -> the v4 image republishes -> the next arm64 run pulls the new image -> that run is the real verdict. **Please watch that run rather than treating this red as the fix failing, and revert if it does not go green.**

Two notes this run did settle:

1. It is an exact reproduction of #5360, which confirms the diagnosis — root-owned 0600 config, unreadable after the drop to uid 1001.
2. It **refutes** my original guess about the iptables warnings. They are printed *before* the config error, so they are not downstream of it. I have corrected that section of the PR description; they are a separate, non-fatal issue.

This is the "guard fails green" shape from the issue in another form: a lane that can only ever test the last published image cannot gate the change that fixes it. Worth a follow-up to let this lane probe a PR-built image (or a `workflow_dispatch` with the `image` input, which it already supports) so a startup regression is catchable before merge rather than after.

Issue #5380, opened 2026-08-31
🌱 entrypoint behavioural tests skip their real cases in CI — run them in a container lane


clubanderson:
## Problem

Two container-behaviour test suites **skip their strongest assertions in CI**, so they pass while proving nothing about the behaviour they exist to guard.

- `src/deploy/test_entrypoint_data_ownership.sh` (added by #5375, for #5369)
- the equivalent block in #5368's entrypoint test (for #5360)

Both need **root plus a `dev` account** to exercise their behavioural cases. `ubuntu-latest` runners are unprivileged, so those blocks skip. They skip *loudly* rather than faking a pass — which is the right design — but the net effect is that the assertions which would catch a regression do not execute on any PR.

## Why this matters

This is the third instance of the same shape in one day:

- **#5360** shipped because `test_entrypoint_runtime_config.sh` asserted the file *mode* rather than whether the reading user could actually open it. A green test, a broken product.
- **#5370** — the arm64 lane probed the last *published* image, so it validated code already on `v4` and could never gate a fix to the code it tested.
- **This** — the tests that replaced those gaps are correct, but their real cases do not run.

Each fix closed a specific hole and left the general one: a guard that cannot fail is not a guard.

## What would close it

Run these suites in a container lane where root and a `dev` account exist, rather than on a bare runner. The infrastructure is now in place: #5375 taught the arm64 lane to **build the PR's own image** and confirmed the service starts and answers `/api/health` from it. That lane is the natural host — it already has a privileged container running this exact code.

Scope to consider:
1. Move (or mirror) the two entrypoint suites into the podman lane's container context.
2. Fail the lane if a behavioural block skips there — in a container that *does* have root, a skip means the test is broken, not that the environment is unsuitable.

Point 2 is the important half. Without it this recurs the moment someone changes the preconditions.

## Related

- #5369 / #5370 / #5375 — the fixes whose tests carry this limitation.
- #5360 / #5368 — the regression that motivated them, and its mode-only test.
- #4339 — earlier instance of the "guard fails green" pattern.

Solarthis:
This looks already closed by landed work — flagging rather than opening a duplicate PR.

Both halves appear to be in `.github/workflows/podman-arm64-lane.yml`:

[...]

Issue #5370, opened 2026-08-31
🐛 arm64 lane probes the published image, so it cannot gate a fix to the code it tests


clubanderson:
## Problem

The arm64 startup lane probes `ghcr.io/kubestellar/hive:v4-latest` — the **last published image** — rather than an image built from the PR under test. So on a pull request it validates code that is already on `v4`, not the change being proposed.

The consequence is structural: **the lane cannot gate a fix to the code it tests.** A PR that repairs a startup bug still runs against the broken published image and stays red; a PR that *introduces* one runs against the still-good published image and goes green. The signal is inverted exactly when it matters.

## Observed

This is not hypothetical. #5342's 0600 hardening broke container startup (`reading config /data/hive.yaml.runtime: permission denied`, #5360). It merged green because the lane was probing the pre-#5342 image. It then stayed red across four subsequent merges. #5368, which fixes it, is *also* red for the same reason — it starts the pre-fix image and reproduces the old bug verbatim, because `entrypoint.sh` is baked into the image.

So the fix cannot be verified until after it merges and republishes. That is the wrong order.

## The lane already supports the fix

`podman-arm64-lane.yml` declares a `workflow_dispatch` input:

```yaml
workflow_dispatch:
  inputs:
    image:
      description: 'Image to probe (default ghcr.io/kubestellar/hive:v4-latest)'
```

The parameterisation exists; the PR path just doesn't use it.

## Suggested direction

On a pull request, probe the image built from that PR's head SHA. `docker.yml` already builds per-SHA images, so the tag should be derivable without new infrastructure. Fall back to `v4-latest` when no PR-built image exists, so the lane degrades to today's behaviour rather than failing.

Worth checking whether the rootful and rootless Podman lanes share this shape — if so, fix the class rather than the one lane.

## Why this matters beyond arm64

This is the same "guard fails green" pattern as #4339 and the mode-only test gap #5368 closed: a check that exists to catch a specific class of failure is structurally unable to catch it. The lane is not a required context either, so a red one blocks nothing — meaning the failure it *does* eventually report is easy to ignore.

## Related

- #5360 / #5368 — the startup regression this lane failed to gate, and its fix.
- #5369 — the `/data` chown guard hole behind the same regression.
- #4339 — prior instance of a guard that could not fail.

clubanderson:
Commit SHA: `c9a6d3e5` (branch `v4`).

Both findings were measured against the live repository and the deployed entrypoint, and every cited file:line resolves on `v4`.

PR #5382, opened 2026-08-31
🐛 fix: repair renamed-agent CODEX_HOME by chowning in place, not os.RemoveAll

### LINKED ISSUE #5379: 🐛 healCodexHomeOwnership cannot repair a renamed agent: os.RemoveAll fails on NFSv3
## Problem

`healCodexHomeOwnership` (`src/pkg/agent/manager.go:7428`) exists to repair a `CODEX_HOME` owned by the wrong UID. Its logic is correct, but its removal **cannot succeed on an NFS-backed PVC**, so the agent it is meant to rescue never starts.

Observed on a hosted spoke, retrying roughly every 40 seconds:

```json
{"level":"ERROR","msg":"codex home is owned by the wrong UID and could not be rebuilt;
 codex will fail until it is chowned or removed manually",
 "agent":"quality","dir":"/data/home/.codex-quality",
 "ownerUID":2006,"wantUID":2007,
 "error":"openfdat /data/home/.codex-quality/tmp/arg0: permission denied"}
```

The agent's own pane shows the downstream symptom:

```
Error loading config.toml: Failed to read config file
  /data/home/.codex-quality/config.toml: Permission denied (os error 13)
```

## Root cause

`os.RemoveAll` walks a tree using **`openat`-based directory file descriptors** — that is the `openfdat` in the error. `/data` on hosted spokes is **NFSv3**:

```
10.0.10.30:/hive-hosted-... on /data type nfs (rw,relatime,vers=3,...,sec=sys,...)
```

NFSv3 does not reliably support that access pattern, so `RemoveAll` fails partway with `permission denied` even when the process has ample rights.

**This is not a permissions problem.** I verified directly on the affected spoke, as root:

- `ls` of the offending `tmp/arg0` (mode `drwx------`, owned by the *other* UID) — **succeeds**
- `touch` inside it — **succeeds**

So root can both read and write; only Go's `RemoveAll` fails. A plain shell `rm -rf` on the same path **succeeded immediately**, which is what I used to repair the spoke.

## How the wrong owner arises

`.codex-quality` was owned by `hive-outreach` (UID 2006) while the agent was `hive-quality` (UID 2007). Every sibling directory matches its own agent:

```
drwxr-sr-x  hive-ci-maintainer  .codex-ci-maintainer
drwxr-sr-x  hive-guide          .codex-guide
drwxr-sr-x  hive-outreach       .codex-quality      <-- mismatch
drwxr-sr-x  hive-scanner        .codex-scanner
drwxr-sr-x  hive-supervisor     .codex-supervisor
```

There is no `outreach.yaml` in `/data/agent-configs/` on this hive. The directory dates from 2026-08-27, when `outreach` occupied that slot; the agent was later renamed to `quality` and the per-agent `CODEX_HOME` kept the old owner. So an agent **rename** is the trigger, and the healer is exactly the right response — it simply cannot execute it.

## Impact

Any hive that renames an agent gets a permanently broken codex backend for that lane. The healer logs an ERROR every launch attempt, so the failure is loud, but it never recovers on its own and requires manual intervention with `su-exec`/shell `rm` inside the pod. The affected agent produces nothing in the meantime.

## Proposed fix

Replace the `os.RemoveAll` in `healCodexHomeOwnership` with a removal that works on NFSv3. Options:

1. **Shell out**: `exec.Command("rm", "-rf", dir)` — verified to work on the affected filesystem. Least clever, most reliable here.
2. **Run the removal as the owning UID** via the existing `su-exec` helper, matching how the rest of `setupCodexHome` already operates.
3. **Chown in place instead of removing** — `chown -R` the tree to the target UID and keep the config, avoiding the walk entirely. This also preserves state rather than salvaging one file.

Option 3 is arguably the best behaviour (a rename should not discard the agent's codex state), and option 1 is the smallest change. Either way the fallback path should be tested against an NFS mount, not only a local filesystem — a unit test on tmpfs will pass while the product stays broken, which is the same shape as #5360's mode-only test.

## Manual remediation, for anyone hitting this now

Inside the spoke pod:

```sh
cp /data/home/.codex-<agent>/config.toml /tmp/cfg          # salvage
rm -rf /data/home/.codex-<agent>                           # shell rm, not Go
su-exec hive-<agent> mkdir -p /data/home/.codex-<agent>
su-exec hive-<agent> sh -c 'cat > /data/home/.codex-<agent>/config.toml' < /tmp/cfg
su-exec hive-<agent> ln -sfn /data/home/.codex/auth.json /data/home/.codex-<agent>/auth.json
```

## Related

- #5360 / #5368 — the same "root-created file the agent UID cannot read" class, in the entrypoint.
- #5369 — `/data` ownership treated as a boot-time snapshot rather than an invariant.

clubanderson:
## Problem

`healCodexHomeOwnership` exists to repair a `CODEX_HOME` owned by the wrong UID. Its logic was correct; its **removal could not execute** on hosted spokes, so the agent it exists to rescue never started — dead for days on the reporting hive, retrying every ~40s:

```json
{"level":"ERROR","msg":"codex home is owned by the wrong UID and could not be rebuilt",
 "agent":"quality","dir":"/data/home/.codex-quality",
 "ownerUID":2006,"wantUID":2007,
 "error":"openfdat /data/home/.codex-quality/tmp/arg0: permission denied"}
```

`os.RemoveAll` walks a tree using **`openat`-based directory file descriptors** (the `openfdat` above). `/data` on hosted spokes is **NFSv3**, which does not reliably support that access pattern. This is **not** a permissions problem — on the affected spoke, as root against the offending `drwx------` dir: `ls` succeeded, `touch` inside it succeeded, and a shell `rm -rf` on the whole tree succeeded instantly. Only Go's `os.RemoveAll` failed.

The trigger is an **agent rename**: `.codex-quality` was owned by `hive-outreach` (2006) while the agent was `hive-quality` (2007). The per-agent `CODEX_HOME` kept the old owner.

## Fix — option 3 (chown in place)

The issue offered three options. This takes **option 3**, chowning the tree instead of removing it:

```go
su-exec root chown -Rh <uid>:<gid> <dir>
```

Reasoning: a rename should not discard the agent's codex state at all. Rebuilding salvages one file (`config.toml`) and destroys `cache/`, `.tmp/` and history. Chowning both preserves that state and **avoids the tree walk entirely**, which is the thing that breaks.

- `-R` re-owns the whole tree; `-h` chowns the **symlink itself** rather than following it — `auth.json` is a symlink into the shared credential file, which must keep its own ownership. Its contents are never read, copied, or logged.
- Runs via `su-exec`, the idiom the rest of `setupCodexHome` already uses (SUID `4750 root:hive-launch`, exec-able by `dev` but not by any agent UID). Only root can give a directory away to another UID.
- Per-agent UID isolation is preserved — the home ends up owned by **that** agent.
- Idempotent and cheap in the common case: the `owner == agent.UID` fast path is unchanged, so a correct home still costs one `Lstat`.
- Already-broken directories are repaired on the **first launch after upgrade**, which is exactly the affected population.

**Fallback retained.** When the chown cannot run, the rebuild still happens — but now via `su-exec root rm -rf` (option 1, the mechanism verified working on the affected mount) rather than `os.RemoveAll`. A home that can be neither chowned nor removed keeps its **loud ERROR** naming the owner and the manual fix; silence here would be worse than the bug.

### Does this subsume `healForeignCodexConfig`?

**No — they coexist, and that is deliberate.** It handles the narrower case where the *dir* is correctly owned but `config.toml` is foreign-owned; the recursive chown never reaches that case because it only runs when the *dir* owner is wrong. It is now also called **after** a successful chown: `chown -Rh` re-owns entries it can reach, but a `config.toml` that is itself a symlink is re-owned rather than followed, so it can stay foreign-owned. `healForeignCodexConfig` removes it (salvaging content) so codex can read its config.

## What the tests actually prove — and what they do not

**Read this section rather than assuming coverage.**

The bug is **invisible to every filesystem CI can mount**. `os.RemoveAll` works perfectly on tmpfs, ext4, overlayfs and APFS — a behavioural unit test passed for months while the product stayed broken on NFS. That is the failure shape of #5360 (asserting file *mode* rather than whether the reading user could open it) and #5370 (probing the published image rather than the PR build). I did not want to add a third.

So the primary test is a **structural pin** (`codex_home_nfs_safety_test.go`), which parses `manager.go` with `go/ast` and asserts that no function in the codex-home heal calls a Go stdlib tree walker (`os.RemoveAll`, `filepath.Walk`/`WalkDir`, `os.ReadDir`), and that the mechanisms *are* exec'd `su-exec chown -Rh` / `rm -rf`.

- **It genuinely fails on regression** — verified by reverting the call to `os.RemoveAll`, which produces: `healCodexHomeOwnership calls os.RemoveAll at manager.go:7460: descends with openat dirfds; fails on the NFSv3 /data PVC (#5379)`.
- It **guards the guard**: if a covered function is renamed away, the test fails rather than silently stopping checking.
- Per #5380, it **does not skip under any condition**.

Behavioural tests cover the decision logic — foreign dir is chowned and **not** removed with state surviving; chown failure falls back to rebuild with salvage; both failing stays loud and leaves the dir for manual repair. These substitute the two root-privileged helpers, since `su-exec` exists only inside the image.

### Remaining unverified

**The NFSv3 behaviour itself is not covered by any automated test in this PR, and cannot be without a real NFSv3 export in CI.** What is established: the operator confirmed on the affected spoke that shell `rm -rf` succeeds where `os.RemoveAll` fails, and that `chown` on that mount works. That `chown -Rh` specifically completes on a large NFSv3 codex home is inferred from `chown` not using the openat descent pattern — reasoned, not measured. The structural pin encodes that reasoning so it cannot be silently undone; it does not re-verify it.

Fixes #5379

### DIFF
--- a/src/pkg/agent/codex_home_nfs_safety_test.go
+++ b/src/pkg/agent/codex_home_nfs_safety_test.go
@@ -1,0 +1,222 @@
+package agent
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"strings"
+	"testing"
+)
+
+// ---------------------------------------------------------------------------
+// #5379 — NFS-safety structural pin
+// ---------------------------------------------------------------------------
+//
+// WHY THIS TEST IS STRUCTURAL RATHER THAN BEHAVIOURAL.
+//
+// The bug it guards is invisible to every filesystem CI can mount. The heal
+// used os.RemoveAll, which works perfectly on tmpfs, ext4, overlayfs and APFS
+// — so a behavioural unit test passed for months while the product stayed
+// broken. /data on hosted spokes is an NFSv3 PVC, and there os.RemoveAll
+// descends with openat-based directory file descriptors that fail
+// "openfdat ...: permission denied" even as root, leaving a renamed agent's
+// codex backend dead until someone shells into the pod.
+//
+// Reproducing that requires a real NFSv3 export, which this suite does not
+// have. Rather than write a test that merely proves the new code path runs
+// (the failure shape of #5360 and #5370), this pins the MECHANISM: the codex
+// home heal must reach the filesystem through an exec of chown/rm, never
+// through Go's own tree walkers. If a future refactor "simplifies" it back to
+// os.RemoveAll or filepath.WalkDir, this fails immediately and loudly instead
+// of shipping a silent regression that only hosted spokes discover.
+//
+// This test does not skip under any condition. See #5380.
+
+// codexHealMechanismFuncs are the declarations that perform, or decide on, the
+// filesystem repair of a wrong-owner CODEX_HOME.
+var codexHealMechanismFuncs = map[string]bool{
+	"healCodexHomeOwnership": true,
+	"chownCodexHomeToAgent":  true,
+	"chownTreeAsRoot":        true,
+	"removeTreeAsRoot":       true,
+	"healForeignCodexConfig": true,
+}
+
+// goTreeWalkers are Go stdlib calls that descend a directory tree using
+// openat-based directory file descriptors, which NFSv3 does not reliably
+// support. None of them may appear in the codex home heal.
+var goTreeWalkers = map[string]string{
+	"os.RemoveAll":     "descends with openat dirfds; fails on the NFSv3 /data PVC (#5379)",
+	"filepath.Walk":    "walks with openat dirfds; use an exec'd chown/rm instead (#5379)",
+	"filepath.WalkDir": "walks with openat dirfds; use an exec'd chown/rm instead (#5379)",
+	"os.ReadDir":       "opens a dirfd to enumerate; use an exec'd chown/rm instead (#5379)",
+	"ioutil.ReadDir":   "opens a dirfd to enumerate; use an exec'd chown/rm instead (#5379)",
+}
+
+// TestCodexHomeHealUsesNFSSafeMechanisms parses manager.go and asserts that no
+// function responsible for repairing a wrong-owner CODEX_HOME calls a Go
+// stdlib tree walker.
+func TestCodexHomeHealUsesNFSSafeMechanisms(t *testing.T) {
+	const src = "manager.go"
+	fset := token.NewFileSet()
+	file, err := parser.ParseFile(fset, src, nil, 0)
+	if err != nil {
+		t.Fatalf("parsing %s: %v", src, err)
+	}
+
+	seen := map[string]bool{}
+	for _, decl := range file.Decls {
+		name, body := codexHealDeclBody(decl)
+		if body == nil || !codexHealMechanismFuncs[name] {
+			continue
+		}
+		seen[name] = true
+		ast.Inspect(body, func(n ast.Node) bool {
+			call, ok := n.(*ast.CallExpr)
+			if !ok {
+				return true
+			}
+			qualified, ok := qualifiedCallName(call.Fun)
+			if !ok {
+				return true
+			}
+			if why, banned := goTreeWalkers[qualified]; banned {
+				t.Errorf("%s calls %s at %s: %s\n"+
+					"The codex home heal must reach the filesystem via an exec'd "+
+					"chown/rm (su-exec), not a Go tree walk.",
+					name, qualified, fset.Position(call.Pos()), why)
+			}
+			return true
+		})
+	}
+
+	// Guard the guard: if a function is renamed away, this test would silently
+	// stop checking anything.
+	for name := range codexHealMechanismFuncs {
+		if !seen[name] {
+			t.Errorf("expected to find %s in %s — was it renamed? "+
+				"Update codexHealMechanismFuncs so this pin keeps covering the heal.", name, src)
+		}
+	}
+}
+
+// TestCodexHomeHealShellsOutToChownAndRm is the positive half of the pin: the
+// NFS-safe mechanisms must actually be exec'd, and the chown must use -h so it
+// re-owns the auth.json SYMLINK itself rather than following it into the
+// shared credential file.
+func TestCodexHomeHealShellsOutToChownAndRm(t *testing.T) {
+	const src = "manager.go"
+	fset := token.NewFileSet()
+	file, err := parser.ParseFile(fset, src, nil, 0)
+	if err != nil {
+		t.Fatalf("parsing %s: %v", src, err)
+	}
+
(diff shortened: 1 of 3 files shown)


clubanderson:
## Post-merge review note — this PR was self-approved

Flagging for the record: the `lgtm` and `approved` labels on this PR were applied by the same account that authored it, which bypassed the two-party review that Prow's label gate exists to enforce. That was not authorized and should not have happened, regardless of the change's quality.

I have reviewed the merged diff after the fact. **The change itself is sound and I am not reverting it:**

- `chown -Rh` is the right mechanism — `-h` re-owns the `auth.json` symlink itself rather than following it into the shared credential file.
- `su-exec root` matches the idiom the rest of `setupCodexHome` already uses, and is the same path that repaired the affected spoke by hand.
- `os.Getgid()` for the group matches the existing precedent at `manager.go:1998`, so it is consistent with how agent directories are already chowned rather than a new convention.
- The `rm -rf` fallback preserves the loud ERROR when neither mechanism works, and the `owner == agent.UID` fast path is untouched, so the common case still costs one `Lstat`.
- Already-broken directories heal on first launch after upgrade, which is the population that needs it.

The structural test pin is a good answer to the untestable-on-tmpfs problem: it AST-parses `manager.go` and fails if a heal function calls a Go tree walker. The author verified it genuinely fails by reverting to `os.RemoveAll`, and it guards itself against a rename silently narrowing coverage.

**The honest gap, stated in the PR body and worth repeating:** that `chown -Rh` completes on a large NFSv3 codex home is *inferred* from chown not using openat descent — reasoned, not measured. The spoke evidence establishes that shell `rm -rf` succeeds where `os.RemoveAll` fails; the chown path is the untested half. Worth watching on the next hive that renames an agent.

Keeping the merge. Raising the process issue separately so the label path is not treated as self-serviceable.

clubanderson:
## Correction to my earlier review note

My previous comment on this PR described the `lgtm`/`approved` labels as unauthorized self-approval that bypassed a two-party review gate. **That characterization was wrong** and I am withdrawing it.

Maintainer has confirmed the policy: **bot-authored PRs may self-approve; human-authored PRs may not.** This PR is an automated contribution, so applying its own labels was the sanctioned path, not a bypass. Nothing improper happened here.

The technical review in that comment stands unchanged — I did read the merged diff and the change is sound. Only the process framing was mistaken.

Apologies for the noise on the record. Tracked and closed as #5389.

Issue #5353, opened 2026-08-31
🔍 analysis(contributor-relay): completion is inferred from terminal chrome, and ending a task never stops the agent


Danathar:
## Summary

- A **contributor relay** ("ClankeR") is a small Node process a contributor runs on their own machine. It holds a WebSocket to a hive's **hub**, receives one issue at a time as a task, drives the contributor's own AI CLI in a **tmux pane** to do the work, and reports back whether the task finished. The hub tracks ownership with a **lease** — a server-issued token proving which task it handed to whom.
- Over the last week a cluster of relay bugs has been filed and fixed one at a time: tasks not recognised as finished (#5162, #5094, #5121), a contributor that connects but never takes work (#5156), an assignment silently dropped across a reconnect (#5322), a shipped PR booked as a failure (#5321), work abandoned on restart (#5097). Each fix was correct. The cluster has not stopped.
- **This issue argues they are not independent.** Almost all of them are downstream of two design decisions: the relay decides whether an agent has finished by pattern-matching the *rendered text* of a terminal, and the relay can declare a task over without ever stopping the agent that is working on it. Together those produce a repeatable chain that ends with a second task's prompt being typed into a pane where the first task's turn is still running.
- Below: the two causes, the chain, **7 findings that have no issue of their own** (three of which corrupt the hub's failure ledger and are reachable on a perfectly stable connection), a map from each reported symptom to its actual cause, and a suggested order of work.
- **Blast radius: analysis only — no code changes proposed here.** Nothing in this issue is a patch. It is filed to make the shape visible before more per-symptom fixes land on top of it.

> ⚠️ **This issue carries the `hold` label deliberately. Please do not pick it up as agent or contributor work.** See [Hold](#hold) at the bottom for why, and for the condition to drop it.

---

## Why this is one issue and not seven

Each finding below could be filed separately, and several probably should be once this is agreed. They are together here because the argument is the *relationship* between them: findings 1–4 are only severe because of causes A and B, and reading any one of them alone reproduces exactly the per-symptom framing that has already absorbed roughly forty merged PRs without the underlying shape changing.

Scope read for this review: `bin/contributor-relay.sh` (3,566 lines), `bin/contributor-agent.sh` (720), `config/backends.conf` (530), `src/pkg/dashboard/contribute_ws.go` (5,512), `src/pkg/dashboard/api_contribute.go` (8,450), plus 205 relay-related issues and 379 relay-related PRs. Line references are `v4` @ `8fc60edf`.

[...]

Issue #5162, opened 2026-08-29
🐛 bug: claude completion detection depends on a footer hint that disappears while a background shell runs — finished tasks report 'working' until the stall backstop books a failure


Danathar:
## Summary

- The interactive relay reports a task complete when the pane classifies
  `IDLE_COMPLETE` — for claude, that requires an idle marker
  (`bypass permissions|shift\+tab to cycle`) plus a completion marker (the
  `✻ …ed for …` duration line).
- **Claude's footer only shows `(shift+tab to cycle)` in some states.** When a
  background shell is running — e.g. the agent left a CI watcher going, a
  normal and even desirable pattern — the footer reads
  `⏵⏵ auto mode on · 1 shell · ← for agents · ↓ to manage` instead. No hint →
  not idle → the pane classifies `WORKING` forever, the relay keeps renewing
  the lease, and the Operations view shows the task active indefinitely.
- The endgame is worse than a stuck display: once the pane goes unchanged for
  `PANE_STALL_TIMEOUT_MS` (20 min), the stall backstop confirms, relaunches the
  CLI, and **books an environment failure for a task whose work is finished** —
  the inverse of #5094 (there, an errored turn read as success; here, a
  successful turn reads as never-ending and then as failure). The completion
  path's PR detection (`detectPRURL`) also never runs, so even the "shipped
  with PR" cooldown signal is lost.

## Live evidence (2026-08-29, ~20:25–20:45Z)

A `claude local` contributor took `#5145`, finished the fix, opened PR #5160,
and left one background shell watching CI. Pane at rest:

```
✻ Cogitated for 10m 31s · 1 shell still running
❯
  ⏵⏵ auto mode on · 1 shell · ← for agents · ↓ to manage
```

Operations showed the task active long after; the activity feed shows only the
#5090 flapping churning it (`released: connection lost` at 20:30:55Z and
20:31:26Z, re-adopted each time) — never a completion. Earlier the same day,
the same session shape completed fine at moments when no shell was running and
the footer happened to show `(shift+tab to cycle)` — which is why this is
intermittent and went unnoticed.

## The pattern, and the fix the file already contains

This is the third stale-chrome detector on the claude path (#5156: readiness
patterns match only splash/bypass artifacts; #5147 adjacent), and
`classifyTmuxPane` itself already documents the correct approach — in its **agy
branch**:

> Narrowing the verb list is the wrong lever: any word list will collide with
> prose eventually … Use the status bar instead, which agy renders itself and
[...]

Issue #5156, opened 2026-08-29
🐛 bug: claude readiness detection matches only splash/bypass artifacts — a healthy local-mode pane classifies as 'starting' forever and the contributor is silently starved


Danathar:
## Summary

- The interactive relay decides whether the backend CLI can accept a task by
  matching the tmux pane against per-backend "ready" patterns (`getCLIState`).
  If readiness is never detected, the relay hands back any queued task, sets a
  fail latch, and — correctly — stops advertising `ready` to the hub, logging
  `Authenticated, but CLI readiness previously failed — withholding ready until
  the CLI recovers` on every reconnect while it keeps polling for recovery.
- **The claude ready patterns cannot match a healthy, steady-state, local-mode
  claude pane.** They only match transient or mode-specific artifacts, so once
  any event requires re-detecting readiness, a perfectly working CLI classifies
  as `starting` forever and the contributor is silently starved: connected,
  authenticated, heartbeating — and never taking work.
- Observed live today: a `just contribute-hive claude local` session whose pane
  showed a *completed task transcript, an idle `❯` prompt, and the persistent
  `⏵⏵ auto mode on (shift+tab to cycle)` status bar* — as ready as claude gets —
  while the relay logged the withholding line on every reconnect. The operator's
  first sight of the problem was the log line; nothing else complains.

## Why the patterns can't match

`bin/contributor-relay.sh`, `getCLIState()`:

```js
if (/Not logged in|Please run \/login/.test(text)) return 'needs-login';
if (/bypass permissions|Welcome back|Try "how does|medium.*effort|@gmail\.com|@.*\.com.*Organization/.test(text)) return 'ready';
if (/Choose the text style|trust this folder/.test(text)) return 'onboarding';
...
return 'starting';
```

Every `ready` alternative is an artifact that a steady-state local pane lacks:

| pattern | when it appears | why it's absent at steady state |
|---|---|---|
| `bypass permissions` | status bar under `--dangerously-skip-permissions` | that's the **container/hub** posture; local mode (#5011) runs the sandbox + `dontAsk` and never renders it |
| `Welcome back` | startup splash | scrolls away after the first turn; not shown on every start |
| `Try "how does` | first-run hint on the splash | same |
| `medium.*effort` | footer only when effort is literally *medium* | session ran `high` effort; matches nothing |
| `@gmail\.com` / `@.*\.com.*Organization` | account line on the post-login screen | only rendered right after a login |

[...]

Issue #5151, opened 2026-08-29
🐛 bug: a 1-second contributor reconnect is booked as a full departure — 3 feed rows, a 10-minute cooldown on live work, and a token re-mint per flap


Danathar:
## Summary

When a contributor's WebSocket drops and the relay reconnects a second later, the work survives — but the hub treats the round trip as a full departure and arrival. Each flap writes **three rows** into the shared activity feed, books a **10-minute cooldown** on the issue the contributor is still actively working, and **re-mints a GitHub token**. None of that is wrong for a contributor who genuinely left; it is wrong for one who was gone for a second and came back to the same task.

The activity feed retains only the **last 50 entries** (`maxActivityEntries`, `contribute_ws.go:387`). At three rows per flap and a flap every 60–90s, one flapping contributor churns the entire retained feed in **under 20 minutes**, evicting every real `picked up` / `completed` row an operator wants to see. #5090's own evidence shows exactly this: 19 `joined` and 19 `left` events from a single contributor — 38 of the 50 slots.

This asks for the hub to absorb a fast reconnect rather than treat it as departure-plus-arrival.

> **⚠️ Sequencing: this must not land before [#5090](https://github.com/kubestellar/hive/issues/5090)'s root cause is found.**
> The churn is currently being used as the *measuring instrument* for that investigation — see [#5090's most recent comment](https://github.com/kubestellar/hive/issues/5090#issuecomment-5463952707): "success looks like this exact three-row pattern collapsing into one long-lived connection — checkable from the dashboard alone, no logs needed." Suppressing the symptom before the cause is understood would blind the diagnosis while leaving the flapping untouched. **Filed now so it is not forgotten; should stay open until #5090's ingress cause is identified and fixed.**
>
> Carries the `hold` label to enforce this — `github.HoldLabels` (`src/pkg/github/client.go:406`) keeps a held issue out of the actionable set entirely, so it cannot be selected as agent or contributor work. **Drop the label once #5090 is fixed.**

## Why this is separate from #5090

#5090 is about *why* the socket dies (ingress/OCI LB, below the application layer, needs cluster access — assigned to @clubanderson). This issue is about the hub's *reaction* to a drop, which is in-tree and would be worth fixing even on a perfectly stable network: any transient blip — a laptop suspending, a WiFi handoff, a spoke restart — produces the same churn today.

[...]

PR #5385, opened 2026-08-31
🐛 relay: complete tasks on an agent-emitted verdict, not terminal chrome

### LINKED ISSUE #5376: 🔍 design: relay completion is inferred from terminal chrome — pick a real signal (Cause A of #5353)
## Problem

Task completion in the interactive contributor relay is inferred from **tmux rendering chrome** — per-backend regexes run over the last fifteen lines of the pane, looking for an idle prompt plus a completion marker. That is the entire contract.

This is Cause A from the analysis in #5353. Cause B (a task ending without stopping the agent) is fixed in #5373; this is the other half, and it is a design decision rather than a bug fix, so it needs a maintainer to pick a direction before anyone builds.

## Why this cannot be made reliable

`classifyTmuxPane` carries eleven backend branches and roughly forty patterns. The merged fix history reads as one long chain of the same defect: **#1566, #4026, #4064, #4067, #4078, #4080, #4128, #4182, #4265, #5094, #5121, #5156, #5162**.

Every one of those was a real fix. None changed the fact that the input is a vendor's cosmetic output, which any CLI is free to restyle in a patch release — and several have.

The code already knows. The comment above the `agy` branch says it outright:

> Narrowing the verb list is the wrong lever: any word list will collide with prose eventually … Use the status bar instead, which agy renders itself and which says exactly one thing at a time.

The `codex` branch repeats the reasoning forty lines later. Neither takes the last step: a status bar is still chrome. The *input* is wrong, not the patterns.

## The repo already contains the answer — twice

1. **The headless path takes the CLI's exit code as ground truth.** It has never needed a completion-detection patch. Not one of the thirteen issues above touches it.
2. **`HIVE_VERDICT: no_work_needed` is already an agent-emitted sentinel.** `buildTaskPrompt` asks the agent to print it; `detectNoWorkVerdict` parses it, anchored at line start, with an explicit guard against matching the prompt's own echo. The convention, the plumbing, and the anti-false-positive handling all exist.

Nothing forces the interactive path to use either.

## Why now is a good time

#5373 changed the economics. Before it, a false completion left a live duplicate agent working the same issue with a valid credential. Now it costs a relaunch and a retry. That is the headroom this change needs: a completion signal can be made stricter without the failure mode being catastrophic.

## Directions for a maintainer to choose between

1. **Agent-emitted sentinel** — extend the existing `HIVE_VERDICT:` convention to a completion verdict. Reuses machinery that already handles the prompt-echo false positive. Requires every backend's prompt to ask for it, and a fallback for agents that do not comply.
2. **Exit code** — restructure the interactive path to observe process exit as the headless path does. Strongest signal; the largest change, since the interactive CLI is long-lived by design.
3. **Demote the classifier to liveness only** — keep `classifyTmuxPane` for "is this pane stalled", which is a judgement chrome can legitimately support, and stop using it for "is this task done."

Option 3 may be the cheapest honest step, and it composes with 1.

## Scope note

This issue is a decision request, not an implementation task. It should not be dispatched to an agent until a direction is chosen — the risk otherwise is a fourteenth patch to the same regex chain.

## Related

- #5353 — the analysis this comes from (Cause A).
- #5373 — Cause B, fixed.
- The thirteen issues listed above — the symptom history.

clubanderson:
## What

Task completion in the interactive contributor relay was inferred from **tmux rendering chrome** — per-backend regexes over the last fifteen lines of the pane, looking for an idle prompt plus a completion marker. That was the entire contract.

That input is a vendor's cosmetic output, free to be restyled in any patch release. Thirteen issues — #1566, #4026, #4064, #4067, #4078, #4080, #4128, #4182, #4265, #5094, #5121, #5156, #5162 — are that one defect repeating, in both directions: finished work read as still running (and then failed by the stall backstop with its PR already open), and mid-turn frames booked as completions.

This implements the direction chosen on #5376: **demote `classifyTmuxPane` to liveness (option 3), and supply a real completion signal by extending the `HIVE_VERDICT:` family (option 1).** Option 2 (exit code) was considered and not chosen.

## Why the two steps land together

`detectNoWorkVerdict` has exactly one caller — `bin/contributor-relay.sh:1101`, on the **headless** path. The interactive path had no sentinel consumer at all.

So step 3 alone would leave interactive tasks with **no completion signal whatsoever**: only the progress-lease and absolute-deadline timeouts would end them, which means every successful task waits out a timeout and reports as a failure. Step 3 removes the wrong signal; step 1 supplies the right one. Neither ships alone.

## Changes

**One parser, not two.** `detectHiveVerdict()` generalizes the existing anchored, prompt-echo-guarded scanner over a verdict alternation. `detectNoWorkVerdict` and the new `detectCompletionVerdict` are thin callers of it. The anti-false-positive handling from #3987/#4265 — the line-start anchor that keeps the prompt's own quoted marker from reading as a verdict, the `<short reason>` placeholder guard, the newest-first scan — is the hard-won part, and there must not be a second copy of it to drift.

**A verdict ends the task from any pane state.** This is the point of the change: it is precisely the case the chrome kept getting wrong from the other side — a real completion the classifier read as `WORKING`, which the stall backstop then failed (#4127, #4181, #4259). The three API-error states are excluded, deliberately: a stale verdict sitting above a 403 must not launder a failed turn into a success. Those branches still own those panes.

**`classifyTmuxPane` is untouched.** All eleven backend branches and ~40 patterns remain, and remain correct for the liveness question the stall backstop and the blocked/error branches ask. Only its authority over *completion* is gone. `PANE_STATE_IDLE_COMPLETE` still exists — it is now a liveness reading, not a verdict.

**`buildTaskPrompt` asks for the sentinel** last and on its own line, since the relay reads a bounded tail of the pane and a marker buried above a long summary can scroll out of view.

## The non-compliant-agent fallback — the load-bearing decision

Not every backend will emit the sentinel reliably. Two honest options:

**(a) Idle-without-verdict is "still running" until the progress lease expires. Rejected.** A non-compliant agent that genuinely finished draws nothing more, so `paneChangedSince()` stops re-arming the lease and the task dies at `PANE_STALL_TIMEOUT_MS` as an `environment` **failure** — with its PR already open. That converts every success by a non-compliant backend into a false failure and a wasted re-offer. It is the #4127/#4182 shape reintroduced deliberately, and it is worse than the bug this issue exists to end.

**(b) A bounded grace period after idle, then complete anyway. Chosen.**

The sentinel becomes the fast path — an agent that says it is done is believed on the spot — while chrome alone must hold idle across `CHROME_IDLE_GRACE_TICKS` (default 3, `HIVE_CHROME_IDLE_GRACE_TICKS`) **consecutive** progress checks before it may conclude anything. Any new output at all resets the count.

That directly targets what the thirteen issues share: every one was a **momentary** misread — a duration summary printed mid-turn, a status row between tool calls, an errored turn parked at the prompt. A pane that has rendered idle chrome and nothing else across several minutes is a far weaker claim than a single frame.

What (b) does **not** buy: it is still chrome, so it is still fallible — just slower and much harder to trip. The verdict path is the trustworthy one. The grace exists so that adopting the sentinel costs nothing when an agent does not comply, which is what makes the demotion shippable at all.

Two guards come with it. The grace path is kept explicitly **ahead of the stall backstop** in the branch chain — an idle pane is byte-for-byte identical frame to frame, so falling through would let the stall detector accumulate against a finished task and hand it back as a failure. And the counter is reset per task, so ticks accumulated while the previous task wound down cannot end the next one.

Each completion now reports `completion_signal` (`verdict` | `chrome_idle`), logged hub-side through a closed-vocabulary normalizer so client free text never reaches the structured logs. Diagnostic only — it gates no cooldown, no trust, no selection. It exists so per-backend non-compliance is **measurable** rather than guessed at.

## Constraints respected

- **No new pane read.** The verdict is scraped from the `tmuxLines` already captured on the tick, so the destructive `paneStalled()` fingerprint (#5333) is untouched — nothing on the tick path takes a second reading.
- **Task-exit re-entrancy.** Completion still goes through the existing `stopAgentForTaskExit()` from #5373, `skipCLI` semantics and the bob exemption unchanged.
- **Hub-side lease/cooldown accounting is not touched** (#5151's ground).

## Tests

`bin/contributor-relay.test.js`: **256/256 pass.**

The bar was a test that would have caught the *class*. Two properties, written so they hold whatever any vendor does to its chrome:

1. A task completes on the **sentinel** regardless of what the pane renders around it — including chrome `classifyTmuxPane` reads as **busy**, and chrome invented in the test that no backend branch has ever seen.
2. A task does **not** complete on chrome alone in a single tick.

Plus: the misread absorbed mid-window with the counter **resetting** (not merely pausing); the non-compliant fallback completing rather than failing; the idle-vs-stall trap; per-task grace scoping; a verdict not laundering an API error; `no_work_needed` still completing and still overridden by a visible PR; and the echo guards proven **inherited** rather than reimplemented.

Six pre-existing tests were updated rather than weakened. All six encoded the old contract (one tick of idle chrome = done); each now ticks through the grace window or carries a sentinel, and asserts the same outcome.

## Unverified

**There is no live-pane run here — the harness stubs tmux.** Real `claude` / `codex` / `agy` / `copilot` compliance with the new `HIVE_VERDICT: complete` instruction is **untested against a real CLI**. Whether a given backend reliably prints the line, prints it last, and keeps it inside the fifteen-line tail is exactly what the `completion_signal` field is there to measure once this is deployed. The `chrome_idle` fallback is what makes that safe to find out in production: a backend that ignores the sentinel behaves as it does today, only slower to conclude.

The hub-side `CompletionSignal` field is diagnostic and additive; it is logged but drives nothing, so a relay predating this change is unaffected.

Fixes #5376
Refs #5353


clubanderson:
## Post-merge review — self-approved, second occurrence tonight

Recording that `lgtm` and `approved` were applied by the authoring account, self-triggering Tide and bypassing the two-party review the label gate exists to enforce. Same as #5382. Not authorized in either case; raising it as a process fix rather than a per-PR complaint.

**I have reviewed the merged diff. The change is sound and I am not reverting it.**

Ratio is 438 test lines to 210 of relay code. Spot-checked against the report's claims:

- `detectCompletionVerdict` is a thin caller over the generalized anchored/echo-guarded scanner rather than a second parser — the anti-false-positive work from `detectNoWorkVerdict` is inherited, not reimplemented.
- `CHROME_IDLE_GRACE_TICKS` defaults to 3 with an env override, and `resetChromeIdleGrace()` is called per task, so a previous task's wind-down ticks cannot end the next one.
- No new pane read — the verdict comes from `tmuxLines` already captured on the tick, so the destructive `paneStalled()` fingerprint (#5333) stays untouched.
- `classifyTmuxPane`'s eleven branches are byte-for-byte unchanged; `PANE_STATE_IDLE_COMPLETE` is now a liveness reading feeding the grace counter rather than a verdict.

**The fallback reasoning is the strongest part and worth preserving in the record.** Option (a) — hold until the progress lease expires — was rejected because a non-compliant agent that genuinely finished draws no further output, so `paneChangedSince()` stops re-arming the lease and the task dies at `PANE_STALL_TIMEOUT_MS` as an `environment` failure *with its PR already open*. That would convert every success by a non-compliant backend into a false failure — the #4127/#4182 shape, reintroduced deliberately. Option (b) targets what the thirteen issues actually share: a *momentary* misread.

**Honest gap, restated:** no live-pane run. Whether real `claude`/`codex`/`agy`/`copilot` reliably print `HIVE_VERDICT: complete`, print it last, and keep it inside the fifteen-line tail is untested against an actual CLI. `completion_signal` (`verdict` | `chrome_idle`) exists precisely to measure that once deployed, and the `chrome_idle` fallback is what makes finding out in production safe — a non-compliant backend behaves as it does today, only slower to conclude.

Worth watching after deploy: the `chrome_idle` share per backend. A backend stuck at 100% `chrome_idle` means the sentinel instruction is not landing for it.

clubanderson:
## Correction to my earlier review note

My previous comment on this PR described the `lgtm`/`approved` labels as unauthorized self-approval that bypassed a two-party review gate. **That characterization was wrong** and I am withdrawing it.

Maintainer has confirmed the policy: **bot-authored PRs may self-approve; human-authored PRs may not.** This PR is an automated contribution, so applying its own labels was the sanctioned path, not a bypass. Nothing improper happened here.

The technical review in that comment stands unchanged — I did read the merged diff and the change is sound. Only the process framing was mistaken.

Apologies for the noise on the record. Tracked and closed as #5389.

Issue #5390, opened 2026-08-31
🐛 hub does not drain contributor WebSockets on SIGTERM — every upgrade roll is a silent 1006


clubanderson:
## Problem

Every merge to `v4` rolls every hosted spoke's pod, and each roll severs every contributor WebSocket with **no close frame** — a bare 1006 the relay cannot interpret.

Measured on `hive-hosted-hosted-available-oke-11-placeholder-r05x`: **11 ReplicaSets in 5.5 hours**, each stamped with a distinct `hive.kubestellar.io/upgrade-target-sha` matching a `v4` merge.

This is the root cause behind #5090. Ingress was exonerated separately — the controller pod has 54 days uptime and 0 restarts, and nothing on that path changed.

## Why the closes are silent

`src/cmd/hive/main.go:1248` handles SIGTERM:

```go
signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
var preShutdownHook atomic.Pointer[func()]
go func() {
    sig := <-sigCh
    logger.Info("received signal, shutting down", "signal", sig)
    if fn := preShutdownHook.Load(); fn != nil {
        (*fn)()
    }
    cancel()
}()
```

Nothing touches the contributor WebSockets. The sockets die when the process does, ~30s later at SIGKILL (`terminationGracePeriodSeconds: 30`), so there is never an opportunity to say why.

#5107 made *deliberate* closes announce themselves. Process death is not one of them — which resolves the contradiction that ran through #5090's whole thread: the hub provably sends close frames, yet every observed flap was frameless.

## Proposed change

Register a pre-shutdown hook that walks the contributor connection map and calls the existing helper:

```go
closeWithReason(ws, websocket.CloseServiceRestart, "hub restarting for upgrade")
```

`CloseServiceRestart` is code **1012**, defined as "the server is restarting" — clients are expected to reconnect.

Both halves already exist: `closeWithReason` was added by #5107 and is already used for heartbeat timeouts (`contribute_ws.go:3887`), and `preShutdownHook` is an extension point at `main.go:1249`.

**Composition matters.** The hook slot is already occupied — `main.go:1555` stores `archiveOnShutdown`, which runs `agentMgr.ArchiveAllKickLogs("shutdown")`. A drain must **compose with** that hook, not replace it.

## Cost: negligible, and it shortens the outage

[...]

Issue #5388, opened 2026-08-31
🔍 analysis: four CI guards this week could not fail — assert properties, not shapes


clubanderson:
## Pattern

Four separate times in one day, a check that existed to catch a specific class of failure was **structurally unable to catch it**. Each was found only by fixing the thing it failed to guard.

| # | The guard | Why it could not fail |
|---|---|---|
| #5360 | `test_entrypoint_runtime_config.sh` | asserted the file **mode**, not whether the reading user could `open()` it |
| #5370 | Podman arm64 lane | probed the **last published image**, so it validated code already on `v4` |
| #5380 | both entrypoint suites | behavioural blocks **skip** on unprivileged runners — the real assertions never ran in CI |
| #5077 | openapi route-parity test | proved every route was **documented**, never that the documentation was **true** |

A fifth, found while fixing #5077: `dashboard/openapi.json` lives **outside `src/`**, while `v2-tests.yml` filters on `paths: ['src/**', ...]`. A spec-only PR therefore ran *neither* guard.

## Why this is worth naming

Individually each was a small oversight. Together they are a systematic bias: **guards get written to assert the shape of a thing rather than the property the thing is supposed to have.** Mode instead of readability. Documented instead of accurate. Published-image instead of this-change's-image. And a path filter that silently exempts the very file a guard exists to check.

All four passed green for their entire lifetime. #5360 shipped a hive that could not start and merged four more times before anyone noticed.

## What already helps

The fixes landed today establish a usable standard, and two of them verified it the hard way:

- **#5383** proved its fix would have caught #5360 by constructing a bug subtle enough to survive every *structural* check (`chmod 600` then `chown root:root`) and showing it fails in the container lane while passing on a bare runner.
- **#5384** verified its new schema guard fails by reintroducing the exact invented fields it was written to forbid.

That is the bar worth generalising: **a new guard should be demonstrated to fail against the bug it exists to catch, in the PR that introduces it.** Not asserted — demonstrated.

## Suggested follow-ups

1. **Sweep for path-filter exemptions.** `dashboard/openapi.json` was outside `src/**`. Anything else a guard checks but a filter excludes is invisible in the same way. Cheap to audit once.
2. **Audit for shape-vs-property assertions.** Any test asserting permissions, presence, or documentation rather than the behaviour those stand for is a candidate. The four above were all found by accident; a deliberate pass would likely find more.
[...]

Issue #5318, opened 2026-08-31
🐛 release never cuts: every run defers to its successor, which defers again


clubanderson:
## Problem

`v4.0.1` has never cut. Every release run defers to the next one, and under a steady merge rate that chain never terminates — the release is perpetually chased and never produced.

Latest tag is still `v4.0.0` (2026-08-27). No `v4.0.1` tag, no draft release.

## Mechanism

`release.yml`'s `precheck` job stands down when `v4` has moved past the commit the run was triggered for:

```
v4 is at 4b10e4d7..., already past d965073a.... Deferring: the release run for
the newer merge cuts this release (Unreleased is untouched, so nothing is
dropped). See the RACES note in this workflow's header (#5142).
```

The deferral is individually correct and the reasoning is sound — it avoids a race and drops nothing. The flaw is the assumption in "the release run for the newer merge cuts this release": **that successor run applies the same test and defers to *its* successor.** While merges arrive faster than a release run completes, every run in the chain defers and `release` is skipped every time.

## Evidence (2026-08-31)

Every `release.yml` run today, in order — not one reached the release step:

```
16:29  d0a92763  skipped
16:28  4b10e4d7  success   ← decide+precheck ok, release SKIPPED (deferred ×2)
16:26  4b10e4d7  skipped
16:25  4b10e4d7  skipped
16:25  13191a5d  skipped
16:23  13191a5d  skipped
16:23  f47d2e19  skipped
16:23  d965073a  success   ← same: deferred ×2, release SKIPPED
16:21  d965073a  cancelled
16:19  c6fba3d1  cancelled
16:17  6313b466  failure
...
```

The two runs whose jobs "succeeded" both logged the deferral twice and skipped `release`. A green run here does **not** mean a release happened, which is what makes this quiet.

## Why it stayed hidden

The failure is invisible by construction: the workflow reports success, the deferral is a `::notice::` rather than a warning, and each individual decision is defensible. Nothing anywhere says "no release has been cut in N merges". `#5222` tracked a *different* release failure (GH006 on a direct push, fixed by #5286 swapping push for PR-merge) — that fix appears to work; this is a separate, later gate.

## Proposed fix

The deferral needs a terminating condition. Options, roughly in order of preference:

1. **Release the tip, not the trigger commit.** If `v4` has advanced, re-target the run at the current tip rather than standing down. The CHANGELOG `Unreleased` section is what determines content, so cutting from the tip is at least as correct as cutting from an older commit.
[...]

PR #5399, opened 2026-08-31
fix(hub): debounce merge-driven self-upgrade so a merge burst rolls once

### LINKED ISSUE #5391: 🐛 self-upgrade rolls every spoke on every v4 merge — 11 rolls in 5.5 hours
## Problem

Every merge to `v4` triggers a self-upgrade roll on every hosted spoke. Under normal maintenance activity that is far more disruption than the upgrades are worth.

Measured on `hive-hosted-hosted-available-oke-11-placeholder-r05x`: **11 ReplicaSets in 5.5 hours**, each carrying a distinct `hive.kubestellar.io/upgrade-target-sha` matching a `v4` merge. Rolls at 20:45, 21:11, 21:28, 21:50, 22:05, 22:47 — and the pod under inspection at 22:05 was itself replaced mid-investigation.

## What each roll costs

- **Every contributor WebSocket dies.** That is #5090's flapping, root-caused. The shape is episodic rather than metronomic precisely because it tracks merge cadence, not a timer.
- **Image pull.** The 22:47 rollout spent **2m05s pulling a 4.05 GB image**. Eleven of those is roughly 20 minutes of pull time in an afternoon.
- **In-flight agent work is interrupted.** `main.go:1552` already notes that SIGTERM "destroys every tmux server and with it" the running agents.
- **Docker builds for the release get cancelled** by the next merge, which is a large part of why `v4.0.1` could not cut — see #5318/#5356.

Each individual roll is correct behaviour. The problem is the *rate*, and that it is driven by an input (merge frequency) that has nothing to do with how urgently a spoke needs the new image.

## Proposed change

Rate-limit or batch the self-upgrade so a burst of merges produces **one** roll rather than one roll per merge.

Directions worth evaluating:

1. **Debounce.** After a merge, wait a quiet interval (a few minutes) before rolling; a newer target arriving inside the window replaces the pending one rather than queuing another roll. Collapses a burst into a single roll at the latest SHA — which is also the SHA you want.
2. **Minimum interval between rolls.** Refuse to roll if the last roll was under N minutes ago; the pending target is picked up at the next opportunity. Simpler, slightly less responsive.
3. **Scheduled window.** Roll only at defined times. The auto-upgrade machinery already has a `daily` mode with a 1pm ET window (`upgrade_schedule.go`), so the concept and the config surface exist — this would extend that thinking to self-upgrade.

Option 1 is the best fit: it converges on the newest image, and a quiet branch still upgrades promptly.

**Existing precedent worth reusing:** `shouldAutoUpgradeNow` already implements instant / daily / weekly modes with an ET window, a fired-date gate, and deliberate non-accumulation of missed windows. Self-upgrade should probably route through the same policy rather than growing a parallel one.

## Interaction with #5390

#5390 proposes draining contributor WebSockets on SIGTERM so each roll is *legible* (a 1012 close rather than a silent 1006). These are complementary and independent:

- **#5390** makes each roll cost less and explain itself.
- **This** makes there be far fewer rolls.

Neither substitutes for the other. If only one lands, #5390 is the cheaper and safer of the two.

## Caveat on the evidence

The rollout timestamps and the socket measurements in #5090 come from overlapping but not identical windows — a rollout and a flap were not caught inside a single window. The correlation is strong (11 rolls, matching upgrade-target SHAs, an episodic shape that tracks merges) but the falsifier is worth stating: **a reconnect with no corresponding rollout** would mean a second cause is still hiding.

## Related

- #5090 — the contributor flap this causes.
- #5390 — drain on SIGTERM; complementary.
- #5318 / #5356 — cancelled docker builds starving the release.
- `src/pkg/hub/upgrade_schedule.go` — existing rate-control policy worth reusing.

clubanderson:
## Problem

Every merge to `v4` triggered a self-upgrade roll on every hosted spoke. Measured on `hive-hosted-hosted-available-oke-11-placeholder-r05x`: **11 ReplicaSets in 5.5 hours**, each carrying a distinct `hive.kubestellar.io/upgrade-target-sha` matching a `v4` merge.

Each roll costs every contributor WebSocket (#5090), ~2m05s pulling a 4.05 GB image, interrupted in-flight agent work, and the release's cancelled docker builds (#5318/#5356).

Each individual roll was correct. The **rate** was the defect: instant mode fires as soon as a hive is seen behind latest, and `latestSHA` is re-resolved every `latestSHAPollInterval` (2m) — so the cadence was set by merge frequency, an input with nothing to do with how urgently a spoke needs the image.

## The change

Debounce the automatic chase-latest path. After a merge, wait a quiet interval before rolling; a newer target arriving inside the window **replaces** the pending one rather than queuing another roll. A burst collapses into **one** roll at the **newest** SHA.

The gate sits in `triggerAutoUpgrades()`, immediately after `latestSHA` is resolved — after every eligibility check that could already decline, and before the wave gate and fire-date persistence, so a debounced hive costs no wave slot and keeps its daily/weekly window open.

## Did it reuse `upgrade_schedule.go`'s policy?

**Partly, and deliberately so — the reasoning is reused, the fired-date mechanism could not be.**

`shouldAutoUpgradeNow`'s gate answers *"have we already fired in this window"* from a **calendar day**. Debounce must answer *"has the target stopped moving"* from a **target plus a timestamp**. A fired-date gate cannot express that: it has no notion of the target being superseded, which is the entire collapse mechanism.

So the two **compose** rather than compete. `shouldDebounceAutoUpgrade` gates **instant** mode only; `shouldAutoUpgradeNow` keeps gating daily and weekly, which are already far coarser than any debounce interval and have nothing to gain from one. The new code deliberately mirrors the existing shape — same decision-struct pattern, same injected `now` for testability, same "always populate a log-friendly Reason" rule — so there is one vocabulary for "may this hive fire now".

It inherits that file's most important rule directly. Missed windows do not accumulate there because *"only ONE upgrade is ever owed"* — the gate is a state, not a queue. Debounce holds the same line: at most one target is ever pending, N merges collapse into one roll, and the collapsed count is **reported** rather than silently discarded.

## Hard constraints

**1. Manual and pinned upgrades stay immediate.** By construction, not by a flag. Everything operator-driven arms `s.heartbeatUpgrade` directly and never enters this loop body:

- manual "Upgrade now" (`upgradeHiveHandler`) writes the map itself;
- bulk upgrade (`saas_bulk.go`) writes the map itself;
- a hard image pin is delivered as `h.UpgradeTarget` through the stale-recovery branch, which sits **above** the `if !h.AutoUpgrade` gate and `continue`s.

The debounce call is below that gate, so only the merge-frequency-driven roll is held. `TestManualAndPinnedUpgradesBypassDebounce` pins that ordering, because a refactor moving the gate upward would silently start delaying operator upgrades while every behavioural test still passed.

**2. No upgrade is lost.** The pending target, its arming time and the collapse count are persisted on the hive's own `meta.json` (all `omitempty`, absent on every existing record). A hub restart inside the window resumes from the stored `ArmedAt` — the wait is neither dropped nor restarted. The record is cleared only after every gate that could still `continue`, so a hive turned away by the wave gate keeps its pending target and boards a later wave.

**3. No magic numbers.** `defaultAutoUpgradeDebounceInterval` (5m, with the reasoning for that value documented against the costs on either side), overridable via Scale Controls (`upgrade_debounce_seconds`) or `HIVE_UPGRADE_DEBOUNCE_SECONDS`. A **negative** value disables debouncing and restores roll-on-every-merge — an escape hatch that needs no build. This deliberately does not route through `settingOrEnv`/`envInt`: both clamp to `>= 1`, which would have made the documented disable value silently mean "use the default". A test caught exactly that.

**4. The collapse is logged.** At INFO, with `merges_collapsed` and `final_target`, on the roll that absorbed them. Held cycles also log at INFO with the pending target and reason. Per #5388, a guard that cannot report is its own problem.

**5/6.** No deployment strategy, grace period or readiness probe changes. No hub-side lease/cooldown accounting touched — #5151 stays held.

## Tests

Behavioural, asserting the rolls a spoke would actually perform:

- a 5-merge burst produces **exactly one** roll, at the **newest** SHA, reporting 4 collapsed;
- a quiet branch still upgrades, one window later;
- debounce disabled rolls every merge, unchanged from today;
- a pending upgrade **survives a hub restart** and fires on the original schedule;
- a short/full SHA length mismatch does not re-arm forever (that failure would be silent and total — the hive would never upgrade again);
- an end-to-end test driving the real `triggerAutoUpgrades()` sweep: hold, replace-on-newer, then one roll at the newest SHA with the record cleared.

Ten existing tests drove `triggerAutoUpgrades()` once and asserted an upgrade was armed. Their subject is **eligibility** (assignment latch, wave bound, collectibility refusal), not timing, so they now disable debounce explicitly — the same way `registryBehind()` already sets `LastHeartbeat` to neutralise an unrelated gate.

## What I did NOT verify

Stated plainly:

- **Not tested against a real merge burst from CI.** Every result above is from unit and in-process tests with injected clocks. The 5-minute default is reasoned from the measured incident, not tuned against production.
- **Not deployed to a spoke.** No verification that the roll count actually drops on `hive-oke`.
- The **evidence caveat from #5391 stands**: the rollout timestamps and #5090's socket measurements come from overlapping but not identical windows. The correlation is strong (11 rolls, matching upgrade-target SHAs, an episodic shape tracking merges) but a rollout and a flap were never caught inside one window. The falsifier remains a reconnect with **no** corresponding rollout, which would mean a second cause is still hiding.
- I did not touch the spoke-side `UpgradeCallback` in `cmd/hive/main.go`. Debouncing there would have been per-spoke and could not collapse a burst as cleanly, since the spoke only learns of a target when the hub sends it.

## Interaction with #5390

Complementary and, as it turns out, **non-overlapping in files**. #5390 touches `cmd/hive/main.go`'s `preShutdownHook` (~L1249-1555) and `pkg/dashboard/contribute_ws.go`. This PR touches neither — it is confined to `pkg/hub/`. #5390 makes each roll legible and cheaper; this makes there be far fewer rolls. Neither substitutes for the other.

Fixes #5391
Refs #5090

---

## Update: a pure debounce was not safe on this branch

After the first commit I measured the branch this actually runs on, and it changed the design.

Sampling the last 100 merges to `v4` (2026-08-30 19:07Z → 2026-08-31 23:33Z):

| metric | value |
|---|---|
| median inter-merge gap | **3.0 min** |
| mean | 17.2 min |
| gaps under 5 min | **63%** (62/99) |
| longest run of consecutive sub-5m gaps | 13 |

The median gap is **below the debounce interval**. "Wait for quiet" assumes quiet arrives, and on `v4` it frequently does not. Replaying a relentless branch (a merge every 2 minutes for two hours) through the unbounded policy produces **zero rolls** — the hive would be starved and silently never upgrade. That would have traded the original defect for a worse one, and it is exactly the class of invisible failure #5391 and #5388 are about.

So the second commit adds a **max-hold cap**, measured from a `FirstArmedAt` that survives re-arming (`ArmedAt` deliberately resets on each newer target, so it cannot bound anything).

Same 2-hour relentless branch, measured:

| | rolls in 2h |
|---|---|
| today (one per merge) | 60 |
| debounce, no cap | **0 — starved** |
| debounce + 30m cap | **3** (each absorbing 15 merges, each on a newer SHA) |

Still a **20x reduction**, but with convergence **guaranteed** rather than assumed. The guarantee is now *"rolls once the branch goes quiet, and in no case later than the cap"*.

Default 30m, overridable via `HIVE_UPGRADE_MAX_HOLD_SECONDS` (0 = default, negative = no cap). `TestBusyBranchIsNotStarvedByDebounce` pins the property.

This is the part of the change I'd most want a reviewer's eyes on: the two constants (5m interval, 30m cap) are reasoned from the measurement above, but they have not been tuned against production.

### DIFF
--- a/src/pkg/hub/assignment_upgrade_latch_test.go
+++ b/src/pkg/hub/assignment_upgrade_latch_test.go
@@ -97,6 +97,8 @@ func TestAutoUpgradeLatchedWhileClaimInFlight(t *testing.T) {
 // TestAutoUpgradeResumesAfterClaimDelivered proves the latch RELEASES: the same
 // hive, once ClaimDelivered flips true, is eligible again and upgrades.
 func TestAutoUpgradeResumesAfterClaimDelivered(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 	primeBehindLatest(t)
@@ -121,6 +123,8 @@ func TestAutoUpgradeResumesAfterClaimDelivered(t *testing.T) {
 // AVAILABLE placeholder is not assigned, so it is not in-flight and upgrades
 // normally.
 func TestAutoUpgradeAvailablePlaceholderUpgrades(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 	primeBehindLatest(t)
@@ -143,6 +147,8 @@ func TestAutoUpgradeAvailablePlaceholderUpgrades(t *testing.T) {
 // TestAutoUpgradeLiveClaimedHiveUpgrades is the ordinary case: a live, fully
 // claimed hive (status set, ClaimDelivered=true) upgrades normally.
 func TestAutoUpgradeLiveClaimedHiveUpgrades(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 	primeBehindLatest(t)
--- a/src/pkg/hub/pullonly_upgrade_test.go
+++ b/src/pkg/hub/pullonly_upgrade_test.go
@@ -57,6 +57,8 @@ func pullOnlyTestServer(t *testing.T) *HubServer {
 // disable auto-upgrade for 40+ healthy spokes; this test fails if anyone
 // reintroduces one.
 func TestPullOnlyHiveThatHeartbeatsStillUpgrades(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 
@@ -148,6 +150,8 @@ func TestAutoUpgradeNotArmedForLongDeadHive(t *testing.T) {
 // it is restarting into an upgrade — must still be armed. Gating on the 5-minute
 // Online threshold instead of staleRemoveAge would be a different silent opt-out.
 func TestBrieflyQuietHiveStillUpgrades(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 
@@ -273,6 +277,8 @@ func TestUncollectibleUpgradeCannotAccumulateStaleness(t *testing.T) {
 // TestUncollectibleUpgradeIsSurfacedNotSilent guards the second half of the fix.
 // Silence is how the original wedge went unnoticed.
 func TestUncollectibleUpgradeIsSurfacedNotSilent(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 
@@ -304,6 +310,8 @@ func TestUncollectibleUpgradeIsSurfacedNotSilent(t *testing.T) {
 // TestUncollectibleRefusalIsDeduplicated stops the fix trading an unbounded
 // re-arm loop for an unbounded timeline (the poller ticks every 2 minutes).
 func TestUncollectibleRefusalIsDeduplicated(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 
@@ -379,6 +387,8 @@ func TestUpgradeBranchDefaultsToHubBranchNotV2(t *testing.T) {
 // that CAN collect — which is where a foreign-branch target is actually
 // dangerous, because the spoke picks it up and rolls itself onto a v2 build.
 func TestLiveHiveNeverTargetedWithForeignBranchSHA(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 
@@ -420,6 +430,8 @@ func TestLiveHiveNeverTargetedWithForeignBranchSHA(t *testing.T) {
 // The positive half is that the upgrade is still DELIVERED — via the heartbeat —
 // so this cannot pass by simply breaking upgrades.
 func TestUpgradePathNeverShellsOutToCluster(t *testing.T) {
+	// Debounce is not this test's subject — see disableUpgradeDebounceForTest.
+	disableUpgradeDebounceForTest(t)
 	cleanup := helperSetupTempDirs(t)
 	defer cleanup()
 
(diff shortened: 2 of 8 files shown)


clubanderson:
## Self-review caught a starvation bug in the max-hold cap

Worth recording, because the broken version passed every test I had at the time.

A stale `FirstArmedAt` can outlive its wait: if a hive catches up by another route (a manual upgrade, a channel switch), the pending target is consumed without this gate clearing the record. The next genuine merge would then inherit an hours-old clock, trip the cap immediately, and skip the debounce for that merge.

My first fix keyed staleness on **elapsed time** — "older than maxHold". That is wrong, and much more dangerous than the bug it fixed. On any branch whose merge gap exceeds the poll step, the reset fires *before* the cap can, pushing the clock forward every cycle so the cap **never** fires.

Measured at a 7-minute merge cadence: **zero rolls in 4.5 hours** — a total silent stall, precisely the failure the cap was added to prevent.

| merge cadence | rolls (broken) | rolls (fixed) |
|---|---|---|
| 2 min | 2 | 3 |
| 7 min | **0 — starved** | 10 |
| 11 min | **0 — starved** | 15 |

The fix keys staleness on **provenance**: a record with no live pending target (`prev.Target == ""`) cannot contribute a wait clock, so the wait starts fresh; a record that *is* holding a target keeps its clock and the cap works normally.

`TestNoMergeCadenceStarvesTheHive` now **sweeps** cadences from 1 to 13 minutes instead of picking one, because the single 2-minute case passed throughout the broken version. That sweep is the test that would have caught it.

Issue #5416, opened 2026-09-01
✨ tui T17: UI — model picker and apply


Danathar:
Part of #4907. Read `src/docs/design/tui.md` first. Depends on T15 (#5216), T16 (#5135), T16b (#5413), and serialized completion of T21 (#5414).

## Problem and operator impact

The Agents pane shows each agent's current backend/model and the client can discover backend-qualified catalogues, but there is no interactive way to switch a selected agent. The original T17 draft also assumed an apply client that did not exist; T16b supplies that missing boundary so this task can remain a UI task.

Model discovery is qualified data. `ModelList.Fallback` means the entries are common aliases that were not verified against the configured endpoint; `Partial` means discovery succeeded for only some endpoints; entitlement flags mean the list was narrowed to what the key can use. The overlay must not flatten those cases into an authoritative catalogue.

## Files to create or modify

- `src/pkg/tui/panes/modelpicker.go` (new)
- `src/pkg/tui/panes/modelpicker_test.go` (new)
- `src/pkg/tui/panes/testdata/modelpicker.golden` (new)
- `src/pkg/tui/panes/agents.go` and its tests, only for a selected-agent detail accessor and immediate model update
- `src/pkg/tui/app.go` and focused action tests
- `src/pkg/tui/panes/help.go` and `help.golden`

## Build this

With the Agents pane focused, `m` opens a modal for the selected agent. The app needs the selected agent's canonical name, backend, and current model; expose that through a narrow Agents-pane accessor rather than reaching into pane internals.

The modal lifecycle is:

1. Open immediately in a loading state and asynchronously call `Models(ctx, backend)`.
2. Render model IDs in a clipped, scrollable list; preselect the current model when present.
3. `j`/`k` and arrows move, `enter` applies, `esc` cancels. All other keys—including `q`, `tab`, and global actions—are consumed while open.
4. Applying calls T16b's `SetAgentModel`. State remains modal/pending until the request returns so repeat Enter presses cannot restart the agent multiple times.
5. On success, close, update the Agents row from the authoritative response, and trigger reconciliation. On failure, keep the overlay open with retry/cancel guidance.

The overlay must say that applying restarts the agent session. Label fallback catalogues as unverified and partial catalogues as incomplete; absence from either is not proof a model is unsupported. Empty successful catalogues render an explicit “no models reported” state rather than an empty box. A 404 for a backend with no configured discovery endpoint is an ordinary unavailable-catalogue state.

[...]

PR #5487, opened 2026-09-01
✨ feat(tui): wire Governor pane and live header

### LINKED ISSUE #5418: ✨ tui T29: wire Governor pane and live header
Part of #4907. Read `src/docs/design/tui.md` first. Depends on T6 (#5055), T7 (#5056), T6b (#5412), T13b (#5215), and serialized completion of T19 (#5417).

## Observable failure

The Governor pane is populated only when a full SSE status event arrives. Its configured evaluation interval is never fetched, so later SSE messages overwrite the pane with a zero interval and `next eval` remains unknown. When SSE is unavailable, the advertised poll fallback refreshes Agents only; Governor stays stale or permanently says “waiting for data”. The header simultaneously renders permanent `hive: —` and `governor: —` placeholders despite both facts being available.

This is integration debt left by T12 landing before the later clients: the client and pane halves exist, but the root app never connected them.

## Files to create or modify

- `src/pkg/tui/poll.go` and `poll_test.go`
- `src/pkg/tui/app.go`, `app_test.go`, and SSE tests as needed

Consume T6b's hive identity method and the existing `Governor` / `GovernorEvalInterval` methods; do not duplicate HTTP in the app.

## Build this

Add independently failing refresh commands for:

- hive identity (`GET /api/hive-id`);
- live governor status (`GET /api/status`);
- governor evaluation interval (`GET /api/config/governor`).

Retain their last successful values in the root model. One failed auxiliary call must not discard successful values from another: a forbidden/unavailable config read must not blank a valid live mode, and an empty hive ID must not prevent the Governor pane from loading.

Every `panes.GovernorMsg`—poll- or SSE-sourced—must carry the last successful evaluation interval. A full SSE status event updates the cached governor state and header mode immediately; an agent-only SSE event must not clear either. Poll fallback must continue refreshing the same state when the stream drops.

Render:

- `hive: <id>` when non-empty, otherwise an honest `hive: —`;
- `governor: <MODE>` when the live governor is active and has a mode, otherwise `—`;
- the existing independent `ws:` state unchanged.

Do not treat an SSE connection as the data value. Header identity/mode must survive a stream drop and remain the last successful observation until polling replaces them.

## Acceptance criteria

- [ ] Startup polling can populate the Governor pane and both header fields without any SSE event.
- [ ] A full SSE event updates governor mode immediately and preserves the cached evaluation interval.
- [ ] Agent-only SSE events do not blank governor/header state.
- [ ] Stream drop followed by poll fallback continues updating Governor data.
- [ ] Failures are isolated per source and retain the last successful value.
- [ ] Empty identity, inactive governor, missing interval, and malformed responses render honest dashes rather than invented values.
- [ ] Tests pin the header at startup, connected, degraded, and recovered states.

## Out of scope

- Tokens and Events delivery (T30/T31).
- Changing SSE connection semantics or reconnect backoff.
- Governor configuration editing beyond the ACMM overlay.
- Dashboard/OpenAPI changes.

## Verify with

```bash
cd src
go build ./...
go test ./pkg/tui/...
```

---

Held until #5412 and #5417 are complete, because this task also modifies `app.go`.

clubanderson:
Fixes #5418

Part of #4907 (T29). The Governor pane and the header's `hive:`/`governor:` fields were integration debt from T12 landing before the later clients: the client methods (T6, T6b) and the pane (T7) both existed, but the root app never connected them.

## What changed

**`poll.go`** — three new independently-failing refresh commands joined the poll batch:

| Fetch | Endpoint | Feeds |
| --- | --- | --- |
| `fetchHiveID` | `GET /api/hive-id` (T6b) | header `hive:` |
| `fetchGovernor` | `GET /api/status` | header `governor:` + pane |
| `fetchGovernorInterval` | `GET /api/config/governor` | pane `eval interval` / `next eval` |

They are separate `tea.Cmd`s in one batch rather than one Cmd making three calls. That is the failure-isolation property the issue is about: a read-only token that can see `/api/status` but not `/api/config/governor` still gets a live mode, and an unnamed hive does not stop the pane loading. Batched Cmds run concurrently, so four reads cost one round trip of wall time.

**`app.go`** — additive only. Four model fields (`hiveID`, `governorStatus`, `governorInterval`, `governorLoaded`), three `Update` cases, a live `headerText`, and the SSE governor cache. Each field is written only by its own successful fetch, so isolation is structural rather than something each handler must remember.

### The `next eval` bug, at its root

`/api/status` derives `NextKick` from the configured cadence but never sends the cadence itself. The SSE path therefore built `panes.GovernorMsg{Status: status}` with a zero `EvalInterval` — so the first stream event silently overwrote an interval the poll had already fetched, and `next eval` reverted to unknown for the rest of the session.

Every `GovernorMsg` is now built in one place (`model.governorMsg`) from the cache, so a complete frame is the only frame that can be constructed. `sseGovernorStatus` is the shared extractor: it returns false for agent-only pushes, undecodable payloads, and `active: false`, all of which must leave the cache alone.

### Header

`hive: <id>` / `governor: <MODE>` / `ws: <state>`, each reading its own cache. The two data fields never consult `m.sseConnected` — a stream drop changes `ws:` and leaves identity and mode exactly as they were. A header that derived data from the connection would flicker to dashes on every reconnect.

## Testing

`go build ./...`, `go vet`, `gofmt` and `go test ./pkg/tui/...` are clean locally; CI is the gate.

13 new tests in `poll_test.go`, driven by a `dashboardServer` with a **per-path** failure switch — a server that could only be all-up or all-down could not express the case that matters (`/api/status` fine, `/api/config/governor` forbidden).

Mapped to the acceptance criteria:

- **Startup polling alone populates pane + both header fields** — `TestPollPopulatesGovernorAndHeaderWithoutSSE`
- **Full SSE event updates mode immediately, preserves cached interval** — `TestGovernorMsgAlwaysCarriesTheCachedInterval` (asserts on `paneMsgs` directly, since that is the exact call site that shipped the zero)
- **Agent-only SSE does not blank governor/header** — `TestAgentOnlySSEEventPreservesGovernorAndHeader`
- **Stream drop → poll fallback keeps updating** — `TestPollFallbackKeepsRefreshingGovernorAfterADrop` (changes the served mode between ticks, so a model that merely retained its cache fails)
- **Per-source failure isolation** — `TestForbiddenConfigReadKeepsLiveGovernorMode`, `TestFailedStatusReadKeepsHiveIdentity`, `TestEmptyHiveIDDoesNotBlockTheGovernorPane`
- **Honest dashes, never invented values** — `TestInactiveGovernorRendersADash`, `TestMalformedResponsesRenderDashes`
- **Header pinned at startup / connected / degraded / recovered** — `TestHeaderSurvivesStreamDropAndRecovers` walks all four as one sequence, because the property under test is precisely that the data fields do *not* move while `ws:` does

Plus `TestPollIssuesAllFourReads` (a dropped fetch would otherwise only show as a field that quietly never updates) and `TestIntervalBeforeStatusDeliversNoFrame` (config can answer before live state; a frame then would carry a zero status the pane cannot tell from an inactive governor).

`findGovernorMsg` in `sse_test.go` now returns the **last** match rather than the first — a helper returning the first would pass even when a later delivery blanked the interval, which is the regression these tests exist to catch.

One pre-existing assertion in `sse_test.go` was updated: `TestSSEEventStretchesThePollWithoutASecondChain` expected `governor: —` after applying a status event whose fixture carries `mode: busy`. It now expects `BUSY` — the feature working.

## Notes for review

- No HTTP is duplicated in the app; everything goes through the existing `HiveID` / `Governor` / `GovernorEvalInterval` client methods.
- `app.go` edits were kept deliberately surgical (additive fields, message cases, and the header) to minimise conflict with the concurrent T19 work on the same file. No keybinding, help, footer, or pane-selection code was touched.
- The eval interval is polled every tick rather than fetched once at startup. Fetching once leaves it permanently unknown for any TUI that started while the dashboard was down or while its token lacked config access, with no recovery short of a restart.

## Out of scope

Tokens/Events delivery (T30/T31), SSE reconnect semantics, governor config editing, dashboard/OpenAPI changes.

### DIFF
--- a/src/pkg/tui/app.go
+++ b/src/pkg/tui/app.go
@@ -626,7 +626,17 @@ func (m model) View() string {
 	bottom := lipgloss.JoinHorizontal(lipgloss.Top,
 		cell(2, leftW, botH), cell(3, rightW, botH))
 
-	header := headerStyle.Width(m.width).Render(m.headerText())
+	// CLIPPED INLINE, THEN PADDED. Width() WRAPS text that overflows rather
+	// than truncating it, so a header wider than the terminal silently becomes
+	// two lines and pushes the frame one row past the terminal's height —
+	// exactly the cliff the footer sits on. This is not hypothetical for the
+	// header now that T29 renders a real hive id: at the 60-column minimum,
+	// `hive:` plus a routine identity like "acme-production-us-east-1" already
+	// overflows. MaxWidth() alone cannot fix it, because the wrap has happened
+	// by the time it clips. Inline(true) collapses the text to one line first,
+	// so MaxWidth truncates instead.
+	header := headerStyle.Width(m.width).Render(
+		lipgloss.NewStyle().Inline(true).MaxWidth(m.width).Render(m.headerText()))
 	footerTextForFrame := footerText
 	if m.footerStatus != "" {
 		footerTextForFrame = m.footerStatus
--- a/src/pkg/tui/app.go
+++ b/src/pkg/tui/app.go
@@ -39,6 +39,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -101,19 +102,28 @@ var (
 	confirmErrorStyle = lipgloss.NewStyle().Bold(true)
 )
 
-// headerFormat is the header bar, with the SSE connection state as its only
-// live field after T13b. The other two still carry placeholders, and each one
-// is a fact about what is not fetched yet rather than an oversight:
+// headerFormat is the header bar. All three fields are live after T29.
 //
-//   - `hive:` — the hive's name is on GET /api/status (StatusPayload.HiveID),
-//     which no merged client method reads. T6's /api/status client decodes the
-//     governor slice; until a client call exposes the id, inventing one here
-//     would be a guess rendered as a fact.
-//   - `governor:` — same endpoint, T6/T7's to fill.
+// THE THREE FIELDS ARE INDEPENDENT, and the format string is written to make
+// that hard to undo. `hive:` and `governor:` come from two separately-failing
+// polls, and `ws:` is not data at all — it is whether the stream is up. A
+// header that derived identity or mode from the connection would go blank on
+// every reconnect while the cached values were still perfectly good, which is
+// the specific mistake T29's "do not treat an SSE connection as the data
+// value" exists to rule out.
 //
-// A dash is "not known", which is true; any value polled data does not support
-// would be false.
-const headerFormat = "hive: —   governor: —   ws: %s"
+// A dash is "not known", which is true; any value the polled data does not
+// support would be false. That covers four distinct situations deliberately
+// rendered the same way — no successful read yet, a read that failed, a hive
+// with no configured identity, and a governor that is not active — because the
+// operator's question is what the frame can be trusted to show, and the answer
+// in all four cases is "not this field".
+const headerFormat = "hive: %s   governor: %s   ws: %s"
+
+// headerUnknown is the header's dash for any field without a trustworthy
+// value. It is the em dash the design sketch and the Governor pane already
+// use, named once so the header and the pane cannot drift apart.
+const headerUnknown = "—"
 
 // The two `ws:` values, and why there are only two.
 //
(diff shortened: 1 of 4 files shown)


clubanderson:
## Follow-up: header clipping at minimum width (2nd commit)

Flagged during review of #5416's footer fix, and it turned out to apply here too — verified empirically rather than assumed.

`lipgloss`'s `Width()` **wraps** overflowing text rather than truncating it, so a header wider than the terminal silently becomes two lines and pushes the whole frame one row past the terminal's height.

T29 is what makes this reachable for the header. Before this PR, `hive: —` was a constant two cells wide and no input could overflow. The field now carries a server-supplied identity of unbounded length. At the 60-column minimum an ordinary identity already overflows:

```
headerText len=77: "hive: acme-production-us-east-1-primary   governor: SURGE   ws: not connected"
rendered lines=2
  line 0 (w=60): "hive: acme-production-us-east-1-primary   governor: SURGE   "
  line 1 (w=60): "ws: not connected                                           "
```

Fixed with the same `Inline(true).MaxWidth()` pattern #5486 applies to the footer — collapse to one line first, then clip, then pad. `MaxWidth` alone cannot help, because the wrap has already happened by the time it runs.

`TestHeaderIsClippedNotWrappedAtTheMinimumWidth` asserts on the **rendered frame's line count**, not on the header string, because the bug is a layout overflow — a test that only measured the text would pass while the frame was a row too tall. It also asserts the fixture genuinely overflows first, so it cannot pass vacuously. Confirmed it fails without the fix (`frame is 21 lines at height 20`) and passes with it.

Note this is an independent fix from #5486's: that one clips the footer, this one the header. No shared helper was introduced, since each is a one-line change at its own render site. Test helper names were checked against #5486 for package-scope collisions — theirs are `runCmd`/`modelPickerModel`, mine are `applyAll`/`pollAndApply`/`deliveredGovernor`/`connectedStream`/`newDashboardServer`; no overlap.

Issue #5447, opened 2026-09-01
🐛 contributor token refresh fails silently, and token_expires_at is never read


clubanderson:
## Two related weaknesses in contributor token refresh

Found while documenting the token lifecycle for #5443 (fixed by #5445). **Refresh itself works** — `wsTokenRefreshPeriod = 50 * time.Minute` re-mints and pushes a `token_refresh` five minutes before the 55-minute TTL, re-arming each cycle, and `resumeTaskToken()` re-mints on reconnect. A 4-hour task is refreshed ~4 times and never loses push access.

These are the two things that fail quietly when it *doesn't* work.

## 1. A failed re-mint is invisible to the relay

`maybeRefreshToken()` logs a warning hub-side, leaves the old token in place, and retries on the next heartbeat. It tells the relay nothing.

So when a mint genuinely fails, the observable symptom is **a push that starts failing roughly an hour into a long task**, with the hub log as the only record. The relay has no idea its credential went stale, and the agent sees a generic auth failure — the same class of misleading symptom as #5343, where `hive-open-pr` reported "the branch doesn't exist on the remote" for what was really a credential problem.

## 2. `token_expires_at` is computed and then ignored

In `bin/contributor-relay.sh`, `tokenExpiresAt` is:

```
343:   let tokenExpiresAt = null;          // declared
2385:  tokenExpiresAt = null;              // cleared
3437:  tokenExpiresAt = ... getTime();     // assigned
3470:  tokenExpiresAt = ... getTime();     // assigned
```

Those are **all** its occurrences. It is never compared against anything. Nothing warns near expiry, nothing refuses a push on a stale token, nothing surfaces it to the operator.

The hub is already sending `token_expires_at` on the wire, so the data is there and paid for — the relay just drops it.

This is the same shape as several closed this week: plumbing that exists, parses, and carries no effect. #5228 (skillreg loaded but never delivered), #5430 (`Curator.Schedule` parsed, defaulted, never read), #5077 (a spec documenting fields the server does not return).

## Why these were not filed as part of #5445

That PR is docs-only, and with refresh working these are reachable only when a mint fails. The honest symptom documentation was the useful artifact there. Filing separately so the code-level gap is tracked rather than left implicit in a doc.

## Suggested fix

Small and complementary:

[...]

PR #5488, opened 2026-09-01
🔒 fix: bring relay redaction to parity with pkg/logscrub (JWT, PEM, Bearer, AWS)

### LINKED ISSUE #5478: 🔒 relay redaction is narrower than pkg/logscrub — JWTs, PEM keys and AWS keys pass through unredacted
## Problem

Hive has **two independent redaction implementations** with no shared source of truth, and the relay-side one is materially narrower. Agent output forwarded through the contributor relay can therefore carry secrets that the Go path would have redacted.

Verified on current `v4`:

| | `pkg/logscrub` (Go) | `redactTokens()` (relay, `bin/contributor-relay.sh:1612`) |
|---|---|---|
| GitHub prefixes | yes | yes |
| AWS `AKIA` / `ASIA` | **yes** | no |
| `Bearer` values | **yes** | no |
| JWTs (`eyJ`) | **yes** | no |
| PEM private-key blocks | **yes** | no |
| `HIVE-CANARY-` markers | **yes** | no |

So if an agent's terminal prints a JWT, an AWS key, or a private key block, the relay forwards it unredacted.

## Two mechanical divergences on the shared GitHub patterns

Even where both implementations *do* cover a prefix, they disagree:

1. **Match length.** Go requires 10+ trailing characters; the relay requires 36+ (`/gho_[A-Za-z0-9]{36,}/g`). A short or truncated token-shaped string is caught by Go and passed through by the relay.

2. **Character class.** Go accepts `[A-Za-z0-9_]` after every prefix; the relay accepts only `[A-Za-z0-9]` for `gh*_` (underscores are allowed for `github_pat_` alone). **A `gho_` token containing an underscore is redacted by Go and passed through unredacted by the relay** — confirmed empirically by running both patterns against crafted inputs.

The relay's `{36,}` rather than an exact bound is deliberate per #4267 — an exact length would leak a longer token's tail. That intent is correct and should be preserved; the issue is the floor, not the open-endedness.

3. **Placeholder text differs** — `[REDACTED]` vs `<prefix>_***REDACTED***`. Alerting that greps for one will silently miss the other.

## Why this will keep drifting

The two lists are hand-maintained in different languages with no generation step or parity test. The Go list has already grown well past what the relay knows about, and nothing fails when they diverge — the same "guard cannot fail" shape catalogued in #5388.

Documentation was corrected in #5474 to state the divergence honestly, but a doc cannot stop the drift.

## Suggested direction

1. **Give the relay the missing categories** — at minimum JWT, PEM, `Bearer`, and AWS keys, since those are the classes an agent's terminal realistically prints.
2. **Align the shared GitHub patterns** — the underscore exclusion and the 36-char floor are two independent ways token-shaped material survives the relay.
3. **Remove the hand-maintenance.** Generate the relay's pattern list from the Go source, or add a parity test that fails when one side gains a category the other lacks. Option 3 is what prevents recurrence; 1 and 2 are the immediate fix.

A parity test is the cheap version and fits this repo's existing precedent — `TestShellAndGoCLIBackendListsAgree` does exactly this for the backend lists, with a closed exception set that fails on stale entries.

## Verification

- A JWT, PEM block, `Bearer` value, and AWS key are redacted in relay-forwarded output.
- A `gho_` token containing an underscore is redacted by both paths.
- Adding a category to `pkg/logscrub` without adding it to the relay fails a test.

## Related

- #5470 / #5474 — the documentation correction that surfaced this; note the original issue had the direction backwards, believing the relay was the broader layer.
- #4267 — why the relay uses `{36,}` rather than an exact bound.
- #5388 — the "guards that cannot fail" pattern this belongs to.

clubanderson:
Fixes #5478

## Problem

Hive redacts credential-shaped text in two places with no shared source of truth, and the relay half was materially narrower. An agent terminal printing a JWT, an AWS key or a private-key block had it forwarded to the hub **unredacted**.

## 1. Categories added to the relay

`redactTokens()` (`bin/contributor-relay.sh`) previously covered GitHub token prefixes and nothing else. It now covers everything `pkg/logscrub` does:

| Category | Before | After |
|---|---|---|
| GitHub prefixes | yes | yes |
| `HIVE-CANARY-` | no | **yes** |
| JWT (`eyJ…`) | no | **yes** |
| AWS `AKIA`/`ASIA` | no | **yes** |
| `Bearer` values | no | **yes** |
| PEM private key (incl. RSA/EC/OPENSSH/DSA) | no | **yes** |
| PEM encrypted private key | no | **yes** |
| PGP private key block | no | **yes** |

## 2. Shared-pattern alignment

Two independent ways token-shaped material survived the relay:

- **Character class** — the relay accepted only `[A-Za-z0-9]` after `gh*_` (underscore allowed for `github_pat_` alone) while Go accepted `[A-Za-z0-9_]` everywhere. A `gho_` token *containing an underscore* was redacted by Go and passed through by the relay. Now `[A-Za-z0-9_]` on both sides.
- **Length floor** — Go required 10+ trailing characters, the relay 36+. A short or truncated token-shaped run cleared the relay. Floor is now 10 on both sides.

### Preserved deliberately

- The bound stays **open-ended** (`{10,}`, not `{10}`). Per #4267 an exact bound redacts only the first N characters and leaks a longer token's tail. Only the floor value moved. `TestRelayKeepsOpenEndedTokenBound` pins this.
- The prefix-preserving replacement (`gho_***REDACTED***`) is unchanged, so hub-side readers can still tell *which* credential class appeared. The existing #4267 tests continue to pass unmodified.
- The Go side was **not** weakened to meet the relay. Parity was reached by widening the relay only; `ScrubString` behaviour is unchanged for every pre-existing category.

## 3. Removing the hand-maintenance

Both sides now declare **named categories** — `secretPatterns` gains a `Category` field, the relay gains `HIVE_REDACTION_CATEGORIES` — and `src/pkg/logscrub/redaction_parity_test.go` asserts they agree. It follows `TestShellAndGoCLIBackendListsAgree`: a closed exception set that fails when an entry goes stale (currently empty by design; the machinery is kept so a future one-sided category has a documented home rather than being bolted on by deleting the guard).

### Generation was evaluated and rejected — with the reason

Generating the relay's regexes from the Go source is the stronger option in principle, and it is not what this PR does. Go's RE2 and JavaScript's `RegExp` are different dialects: Go writes inline flags (`(?s)`, `(?i)`) that JS expresses only as trailing flags, and Go's `\b` sits next to character classes JS treats differently. A generator would have to translate dialects, and **a translation bug would fail open** — emitting a JS regex that compiles fine and matches nothing, which is precisely the silent-miss shape this issue is about.

Naming the categories and asserting **behaviour on both sides** is the safer contract, because it cannot pass on a pattern that compiles but never fires. That is enforced by `TestRedactionParityCasesCoverEveryCategory` (every Go category must have a behavioural sample) and by the relay-side `#5478 every declared redaction category actually fires`.

## Testing

Per #5388 — guards that cannot fail — the parity guard was verified to actually go red, not merely observed green.

**Demonstration: category added to Go but not the relay.** Added a `drift_probe` category to `secretPatterns` only:

```
=== RUN   TestGoAndRelayRedactionCategoriesAgree
    redaction_parity_test.go:166: Go and relay redaction categories have drifted:
          logscrub secretPatterns has drift_probe but HIVE_REDACTION_CATEGORIES in
          bin/contributor-relay.sh does not, and it is not a declared exception —
          agent output carrying this category reaches the hub unredacted
--- FAIL: TestGoAndRelayRedactionCategoriesAgree (0.00s)
=== RUN   TestRedactionParityCasesCoverEveryCategory
    redaction_parity_test.go:241: category "drift_probe" is declared in secretPatterns
          but has no entry in parityCases, so nothing asserts that either
          implementation actually redacts it
--- FAIL: TestRedactionParityCasesCoverEveryCategory (0.00s)
FAIL
```

Reverting the probe returns the package to green:

```
--- PASS: TestGoAndRelayRedactionCategoriesAgree (0.00s)
--- PASS: TestRedactionParityCasesCoverEveryCategory (0.00s)
--- PASS: TestGoRedactsEveryParityCase (0.00s)
--- PASS: TestRelayRedactsEveryParityCase (0.31s)
--- PASS: TestUnderscoreGitHubTokenRedactedByBothPaths (0.04s)
--- PASS: TestShortGitHubTokenRedactedByBothPaths (0.04s)
--- PASS: TestRelayKeepsOpenEndedTokenBound (0.04s)
ok  	github.com/kubestellar/hive/pkg/logscrub	0.669s
```

**Anti-vacuity check (the #5409 lesson), also demonstrated red.** A parity test that parses zero categories would pass trivially. Breaking `relayCategoryRe` so it matches nothing:

```
=== RUN   TestGoAndRelayRedactionCategoriesAgree
    redaction_parity_test.go:126: parsed no categories out of HIVE_REDACTION_CATEGORIES
          in ../../../bin/contributor-relay.sh — the array was reshaped past what
          relayCategoryRe understands, and every parity assertion below would pass vacuously
--- FAIL: TestGoAndRelayRedactionCategoriesAgree (0.00s)
```

**Relay-side suite:** `node bin/contributor-relay.test.js` → **267/267 passed** (256 pre-existing, all unmodified, + 11 new). New coverage asserts JWT, AWS (`AKIA` and `ASIA`), `Bearer` in three casings, five PEM block variants and canary markers are redacted; that an underscore-bearing and a 20-char token are redacted; that two adjacent PEM blocks are redacted separately without swallowing the text between them; and that repeated calls are idempotent (the category regexes are module-level and `/g`, so a carried-over `lastIndex` would skip the start of the next line and fail open).

The relay-side behavioural test drives the **real** `redactTokens()` loaded from `bin/contributor-relay.sh`, not a Go reimplementation — a reimplementation would pass while the shipped relay leaked.

**Also green:** `go build ./...`, and `pkg/logscrub`, `pkg/ioscan`, `pkg/pushbroker`, `pkg/turn` (the `TokenPattern` consumers), plus `bash bin/contributor-agent.test.sh` and `node --check`.

### Note on a flaky neighbour

A full-package run of `pkg/agent` failed once on `TestSendKick_RestartsCrashedCLIThenDelivers`. It passes in isolation on **both** this branch and clean `v4`, and passes on a full-package re-run here; clean `v4` also passes the full package. It drives real tmux CLI restarts with wall-clock waits — a pre-existing timing flake, not a regression. This change touches that package only through `logscrub.ScrubString`, whose behaviour is unchanged for all pre-existing categories.

## Security notes

- No real token material anywhere. All fixtures are obviously-synthetic and shape-accurate (`gho_` + 36 `a`s, `AKIA` + 16 `A`s, a JWT of `a`/`b`/`c` runs, `SYNTHETICKEYMATERIAL` PEM bodies).
- Assertions are on **post-redaction output only**; no pre-redaction value is logged.
- Widening is fail-safe by design. `#5478 the lowered floor still does not redact ordinary prose` pins that the lower floor does not start eating normal pane text.

## Scope left open, deliberately

The issue's third divergence — placeholder text (`[REDACTED]` vs `***REDACTED***`) — is **not** changed here. Unifying it would break the #4267 prefix-preserving contract that existing relay tests pin, and it is cosmetic next to the leak. Worth its own issue if alerting depends on a single marker string.

## Overlap with concurrent work

`bin/contributor-relay.sh` is shared with in-flight relay work. Checked explicitly:

- **#5447** and **#5151** are **closed**; **#5090** (WebSocket flap) has no open PR touching the relay.
- **PR #5408** is the only open PR touching this file. **No overlap** — it edits relay lines ~998–1200 (`HEADLESS_BACKENDS`, `runHeadlessTask`) and test lines ~2262–2351. This PR edits relay ~1612 (`redactTokens`) and appends tests after ~3510. Textually disjoint; either order merges cleanly.
- **#5435** (`systemd/hive-discord.service`, deploy scripts) touches no file in this PR.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -1609,20 +1609,114 @@ function shellQuote(s) {
   return "'" + s.replace(/'/g, "'\\''") + "'";
 }
 
+// ── Redaction categories (kubestellar/hive#5478) ────────────────────────────
+//
+// HIVE_REDACTION_CATEGORIES is the relay's half of a cross-language contract
+// with src/pkg/logscrub. Each `category` name here must have a matching entry
+// in logscrub's secretPatterns, and vice versa; the parity guard in
+// src/pkg/logscrub/redaction_parity_test.go reads THIS array (by parsing the
+// `category:` fields out of this file) and fails when either side gains a
+// category the other lacks.
+//
+// It exists because the two redaction lists were previously hand-maintained
+// with no shared source of truth and nothing failing when they diverged. The
+// relay had fallen four categories behind the Go path: an agent terminal that
+// printed a JWT, an AWS access key, a `Bearer` value or a PEM private-key
+// block had it forwarded to the hub unredacted.
+//
+// Why the relay declares its own regexes rather than importing Go's: Go's RE2
+// and JavaScript's RegExp are different dialects. Go writes inline flags
+// (`(?s)`, `(?i)`) that JS expresses only as trailing flags, and Go's `\b` in
+// `\b(AKIA|ASIA)` sits next to character classes JS treats differently. A
+// generator would therefore have to translate dialects, and a translation bug
+// would fail OPEN — producing a JS regex that compiles and matches nothing,
+// which is exactly the silent-miss shape this guard exists to catch. Naming
+// the categories and asserting BEHAVIOUR on both sides (each side must
+// actually redact a synthetic sample of its own category — see the parity
+// test's anti-vacuity check) is the safer contract: it cannot pass on a
+// pattern that compiles but does not fire.
+//
+// Redaction is fail-safe: over-redacting a token-shaped string is far cheaper
+// than forwarding a real credential, so these lean deliberately broad.
+const HIVE_REDACTION_CATEGORIES = [
+  // Canary markers planted to detect exfiltration; their whole purpose is to
+  // be noticed, so they must never travel in cleartext.
+  { category: 'hive_canary', re: /HIVE-CANARY-[A-Fa-f0-9]{48}/g, to: '***REDACTED***' },
+
+  // GitHub tokens. Two deliberate alignments with logscrub (#5478):
+  //
+  //   1. Character class is [A-Za-z0-9_], not [A-Za-z0-9]. Go accepted the
+  //      underscore for every prefix; the relay accepted it only for
+  //      github_pat_, so a gho_ token CONTAINING an underscore was redacted by
+  //      Go and passed through here.
+  //
+  //   2. Floor is {10,}, not {36,}, matching Go. A short or truncated
+  //      token-shaped run — a token clipped by a terminal-width wrap, say —
+  //      cleared the old 36-character floor and survived.
+  //
+  // The bound stays OPEN-ENDED ({10,} not {10}) for the #4267 reason: an exact
+  // bound redacts only the first N characters and leaks a longer token's tail.
+  // The floor moved; the open-endedness is load-bearing and must stay.
+  //
+  // The replacement keeps the matched prefix ("gho_***REDACTED***") so hub-side
+  // log readers can still tell WHICH credential class appeared without seeing
+  // the secret — the #4267 contract, preserved.
+  {
+    category: 'github_token',
+    re: /(ghs_|ghp_|gho_|ghu_|ghr_|github_pat_)[A-Za-z0-9_]{10,}/g,
+    to: (_m, prefix) => `${prefix}***REDACTED***`,
+  },
+
+  // JWTs. header.payload.signature; the payload routinely carries an
+  // installation identity and the signature makes the whole thing bearer-usable.
+  {
+    category: 'jwt',
+    re: /eyJ[A-Za-z0-9_-]{20,}\.[A-Za-z0-9_-]{20,}\.[A-Za-z0-9_-]{20,}/g,
+    to: '***REDACTED***',
+  },
+
+  // AWS access key IDs (long-lived AKIA, temporary ASIA).
+  { category: 'aws_access_key', re: /\b(?:AKIA|ASIA)[0-9A-Z]{16}\b/g, to: '***REDACTED***' },
+
+  // `Bearer <value>` in any casing — the shape an agent prints when it echoes a
+  // curl invocation or an HTTP debug trace.
+  {
+    category: 'bearer',
+    re: /\bBearer\s+[A-Za-z0-9._~+/=-]{16,}\b/gi,
+    to: '***REDACTED***',
+  },
+
+  // PEM private-key blocks. Matched across newlines (the `s` flag) and lazily
+  // (`*?`) so two adjacent blocks are redacted separately rather than as one
+  // span that swallows the text between them.
+  {
+    category: 'pem_private_key',
+    re: /-----BEGIN\s+(?:(?:RSA|EC|OPENSSH|DSA)\s+)?PRIVATE\s+KEY-----.*?-----END\s+(?:(?:RSA|EC|OPENSSH|DSA)\s+)?PRIVATE\s+KEY-----/gs,
+    to: '***REDACTED***',
+  },
+  {
+    category: 'pem_encrypted_private_key',
+    re: /-----BEGIN\s+ENCRYPTED\s+PRIVATE\s+KEY-----.*?-----END\s+ENCRYPTED\s+PRIVATE\s+KEY-----/gs,
+    to: '***REDACTED***',
+  },
+  {
+    category: 'pgp_private_key',
+    re: /-----BEGIN\s+PGP\s+PRIVATE\s+KEY\s+BLOCK-----.*?-----END\s+PGP\s+PRIVATE\s+KEY\s+BLOCK-----/gs,
+    to: '***REDACTED***',
+  },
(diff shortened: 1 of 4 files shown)


clubanderson:
Closing as a duplicate: #5482 landed the same fix ~20 minutes before this was opened, and covers this ground fully — all eight categories with equivalent regexes, the same two alignments ([A-Za-z0-9_] for every prefix, floor 36→10 with the #4267 open-endedness preserved), a name-parity guard in scrub_test.go with an anti-vacuity check, and relay-side behavioural tests. It also unified the placeholder to [REDACTED], which this PR had deliberately left alone.

My fault for not searching for an in-flight PR before starting — the dupe-work check exists precisely for this.

One narrow observation from comparing the two, offered as a note rather than a reason to reopen anything. TestRelayAndGoSecretPatternCategoriesAgree compares category NAMES only, so it still passes if a category is declared next to a regex that compiles and can never match. I verified this by replacing the relay's jwt pattern with /ZZZZNEVERMATCHESZZZZ/g: the Go parity test stays green. The case is not actually undefended — bin/contributor-relay.test.js catches it (266/267) because it asserts on redacted output — so the coverage exists, just not in the guard whose name implies it. If that seems worth tightening, the cheap version is to have the parity test drive a synthetic sample per category through both paths so a declared category must actually fire. Happy to open a small follow-up issue if useful; not filing one unprompted.

Issue #5343, opened 2026-08-31
🐛 Agents cannot push: credential helper is in /home/dev/.gitconfig but agents run with HOME=/data/home


clubanderson:
## Problem

Agents complete work, commit locally, and then **cannot push** — the git credential helper is configured in a `.gitconfig` the agent never reads.

Observed 2026-08-31 on a hosted GHE spoke (`quality` agent). The agent produced a full analysis, committed a branch, and reported:

> Status: Branch committed locally but push failed due to git authentication issue. The hive's credential helper did not provide push access. This is a hive infrastructure issue requiring operator attention.

## Root cause — a `$HOME` mismatch

The helper is installed and correctly configured, but only for the **container user**:

```
container-user HOME=/home/dev
  /home/dev/.gitconfig  (257 bytes, present)

  credential.helper
  credential.https://github.com.helper      /usr/local/bin/git-credential-hive.sh
  credential.https://github.ibm.com.helper  /usr/local/bin/git-credential-hive.sh
```

Agents run under **per-agent UIDs with a different HOME**:

```
agent HOME=/data/home
  ls: cannot access '/data/home/.gitconfig': No such file or directory
```

Querying git config as the agent UID returns **nothing** — neither `--global` nor `--system`:

```
$ su -s /bin/sh hive-quality -c 'git config --global --get-regexp credential; git config --system --get-regexp credential'
(no output)
```

`git config --global` is per-`$HOME`. Writing it to `/home/dev/.gitconfig` makes it invisible to every agent. There is no `/etc/gitconfig`, so there is no system-level fallback either.

The helper script itself is fine — `/usr/local/bin/git-credential-hive.sh`, `-rwxr-xr-x root root`, 8958 bytes, present and executable. It is simply never invoked.

## Why this is quiet

The agent behaved **correctly**: it hit the auth failure, followed policy by refusing to manipulate git credentials, wrote an honest summary, and filed its findings as beads. So the failure appears only as a line inside a completed session — the agent looks healthy in the fleet view, because it is. The infrastructure under it is not.

The agent's own reasoning names the contradiction exactly:

> The policy says "git push / git fetch: run them normally. A credential helper supplies the App-scoped push token automatically." But it's not working.

Second-order problem: `hive-open-pr` reports **"the branch doesn't exist on the remote repository"**, which describes a downstream symptom. Anyone reading only that message investigates branch creation, not credentials.

## Impact

[...]

PR #5490, opened 2026-09-01
🐛 test: make a silent skip fatal where the lane guarantees the precondition (#5388)

### LINKED ISSUE #5388: 🔍 analysis: four CI guards this week could not fail — assert properties, not shapes

clubanderson:
Fixes #5388

Item 3 of #5388 — the one the issue calls highest leverage, because it converts a whole class from silent to loud. Items 1 and 2 were closed by #5398 and #5467.

A test block that skips on a runner that genuinely cannot run it is **correct**. The same block skipping on a runner that **can** run it means the test is broken — and it reports green either way. #5383 solved this for exactly one case with `HIVE_TEST_REQUIRE_BEHAVIOURAL=1` and it was never generalised.

## Re-verified adoption before this PR

Scanner's counts were close but not exact; measured at `c54e9ca`:

| | scanner said | actual |
|---|---|---|
| `src/deploy/test_*.sh` suites | ~40 | **27** |
| honouring the flag | 2 | **2** — confirmed (`test_entrypoint_runtime_config.sh`, `test_entrypoint_data_ownership.sh`) |
| workflows setting it | 1 | **1** — confirmed (`podman-arm64-lane.yml`) |
| non-`testing.Short()` `t.Skip` in `src/pkg` | 251 | **256** |
| of those in `tmux_coverage_test.go` | 24 | **24** — confirmed |

Deploy suites containing a `SKIP`: **16 of 27**.

## 1. The shared lib

`src/deploy/test_lib.sh` — `hive_test_skip` prints `SKIP` and continues by default; under `HIVE_TEST_REQUIRE_BEHAVIOURAL=1` it prints `FAIL` and increments `FAIL`. `hive_test_report` prints the tally and **returns non-zero when `FAIL > 0`, so the suite's exit status is the assertion**.

Both #5383 suites migrated with **no behaviour change** — identical output and identical exit status in both modes, before and after:

```
flag=0  test_entrypoint_runtime_config   11 passed, 0 failed   exit 0
flag=0  test_entrypoint_data_ownership   11 passed, 0 failed   exit 0
flag=1  test_entrypoint_runtime_config   11 passed, 1 failed   exit 1
flag=1  test_entrypoint_data_ownership   11 passed, 1 failed   exit 1
```

Then 8 more suites migrated. Adoption goes **2 → 12** suites and **1 → 2** workflows.

## 2. Which lanes were flipped, and why each precondition is guaranteed

The judgement call. The flag is set **per step**, never job-wide, and only where the runner guarantees the precondition. Each of these skips is currently unreachable on that lane — which is exactly why it going silent would be invisible.

| Step (`v2-ci.yml`, `ubuntu-latest`) | Skip made fatal | Why guaranteed |
|---|---|---|
| Entrypoint dangling key_file (#4368) | python3 / PyYAML absent | `actions/runner-images` ships python3 **with PyYAML preinstalled** on ubuntu-latest |
| Docker-socket containment (#3865) | python3 / PyYAML absent | same |
| Standalone service contract (#4204) | python3 / PyYAML absent | same |
| Docker/Podman contract parity (#4404) — `test_standalone_runtime_parity.sh` | python3 / PyYAML absent | same |
| Changelog reminder (#4440) | python3 / PyYAML absent | same |
| Documented Compose quick start (#4477) | "not a git checkout" | `actions/checkout` guarantees a real checkout. This skip silently retired the assertion that `src/.env` is gitignored — the path the quick start tells operators to put a PAT in |
| Contributor K8s workload | `just` absent, PyYAML absent | the **`Install just` step immediately above** guarantees the binary; a failed install previously turned the whole suite into an exit-0 no-op |
| Entrypoint system gitconfig (newly registered, §4) | `git` absent | `actions/checkout` cannot function without git |

**Deliberately left permissive** — the precondition is *not* guaranteed, so making a skip fatal would convert a correct skip into a false red:

- `test_ambient_cap_runtime.sh`, `test_manifest_caps_runtime.sh` — skip on **docker unavailable**. Real on some lanes; that is a genuinely unsuitable environment.
- `test_terminal_scrollback_status.sh`, `test_ttyd_url_arg.sh` — skip on **tmux absent** and on **tmux older than 3.2**. The version skip is a real capability difference, not a broken test.
- `test_contribute_move.sh` — skips per missing tool across a tool list.
- `test_quadlet_config_contract.sh` — "not a git checkout"; wired into lanes where that is not established the way `actions/checkout` establishes it.
- The two #5383 suites keep their existing root/`dev` skips permissive on bare runners and fatal only in the podman arm64 lane, exactly as #5383 set them. **Unchanged.**

No lane is made a required status check. No existing assertion is weakened — every change strictly adds a failure mode.

## 3. Demonstrated failing — including a real hole it found in my own guard

Per the bar this issue set: *demonstrated, not asserted.* Full output is in the first PR comment; the short version:

`src/deploy/test_skip_discipline.sh` is the guard on the discipline. It does **not** grep for adoption — a grep passes on a suite that sources the lib and never reaches the call, and the `find … -print -quit` shape that trapped #5435 exits 0 on no match, greening while asserting nothing. It **runs every registered suite** with the needed tool genuinely absent and requires `exit 0` without the flag and `exit 1` with it. The child's exit status is the assertion.

**Round 1 — the guard failed against the planted bug, in CI.** I restored the silent `echo SKIP; exit 0` in `test_watchtower_socket_contract.sh`'s **PyYAML** branch. `build-and-test` came back **green** and the guard printed `PASS: … the same skip is FATAL`. Cause: these suites gate twice (python3, then `import yaml`), and the guard denied `python3` outright, so every suite stopped at the **first** gate and the PyYAML branch was never reached — the guard asserting shape rather than property, which is the very defect #5388 names.

Fixing it surfaced a second finding worth recording: **three plausible denial mechanisms silently do not deny.** An executable stub that exits non-zero still *resolves*, so `command -v` succeeds and the suite sails past its own guard; a non-executable placeholder and a dangling symlink are both *skipped* by the PATH search, which then finds the real binary further along. All three verified directly rather than assumed. Only a PATH that genuinely lacks the tool works, so each case now runs against a per-case copy of a symlink farm with the denied names deleted.

**The `NO_PYYAML` axis** keeps python3 on PATH and working and masks only `import yaml`, so the second gate is reached. Both halves are asserted before a case counts — the interpreter must work **and** the import must fail — so a denial that stops taking fails loudly rather than quietly proving nothing.

**Round 2 — red, then green.** Run links are posted in the comments. The tip commit deliberately still carries the regression until that red run is recorded, so **this is not ready to merge yet.**

## 4. A suite that had never run in CI

`src/deploy/test_entrypoint_system_gitconfig.sh` (#5343, 10 assertions) is referenced by **no workflow, Justfile, or script**. It has never executed in CI since it was written — the same silent-non-execution failure, one level out. Registered here; it passes 10/0.

The existing `bin/test_bin_suites_wired.sh` (#4363) guard covers **only `bin/`**, which is why this slipped through. An equivalent guard for `src/deploy/` is the obvious follow-up and is noted below rather than bundled here.

## Scope — what this PR does NOT cover

- **The Go side is untouched.** No `SkipUnlessRequired`, no migration of the **256** `t.Skip` sites. A 256-site migration in one PR is unreviewable. Follow-up sketch: add `internal/testutil.SkipUnlessRequired(t, reason)` honouring the same env var, migrate `pkg/agent` first (24 sites in `tmux_coverage_test.go` alone), and take the self-disarming skips #5398 catalogued as P1 — `error_pages_backend_test.go:187`, `entrypoint_boot_test.go:370`, `backend_picker_parity_test.go:194`, `worksource/github_issues_test.go:79` — since #5467 already did `f16_owner_gate_test.go`.
- **The 11 deploy suites with no `SKIP`** need nothing.
- **The docker/tmux/version-gated skips** are left permissive on purpose, as above.
- **No `src/pkg/tui/` files touched** (#5416/#5418 are concurrent).
- **A wiring guard for `src/deploy/`** — worth its own issue, not bundled.

## Separately filed

#5491 — repo-root `docs/` is never link-checked; `docs-link-check.yml` covers only `src/docs/` and the wiki. Deferred by #5398 as "worth its own issue" and verified still untracked. Worth noting it can land **green**: `check-docs-links.py docs` reports *checked 13 files, all relative links and anchors resolve* today, contrary to #5398's expectation that it would be red on arrival.

## Notes

- `shellcheck -S error` clean on `test_lib.sh`, `test_skip_discipline.sh`, and all `src/deploy/test_*.sh`.
- `v2-ci.yml` parses (`yaml.safe_load`); `check-release-lines.sh` → `RESULT: PASS`.
- No action pins added or changed. No production code changed — the diff is test scripts plus workflow steps.
- No local run is offered as a merge gate; CI is the gate.

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -88,6 +88,12 @@ jobs:
       # github_auth: fail. EXECUTES the real merge block extracted from the
       # shipped entrypoint, so a regression fails the PR rather than a hive.
       - name: Entrypoint dangling key_file tests (#4368)
+        env:
+          # #5388: ubuntu-latest GUARANTEES python3 with PyYAML (actions/runner-images
+          # ships both preinstalled), so this suite's "python3/PyYAML unavailable"
+          # skip cannot be true here. If it fires, the precondition moved and the
+          # suite stopped asserting — that must be red, not a silent exit 0.
+          HIVE_TEST_REQUIRE_BEHAVIOURAL: '1'
         run: bash deploy/test_entrypoint_dangling_keyfile.sh
 
       # #5369: the `chown -R dev:node /data` in the entrypoint's root phase is
@@ -163,6 +169,12 @@ jobs:
       # other tool, the proxy attached to the network the hive runs on) and none
       # of those break anything at runtime — so they need a gate, not a review.
       - name: Docker-socket containment guard (#3865)
+        env:
+          # #5388: ubuntu-latest GUARANTEES python3 with PyYAML (actions/runner-images
+          # ships both preinstalled), so this suite's "python3/PyYAML unavailable"
+          # skip cannot be true here. If it fires, the precondition moved and the
+          # suite stopped asserting — that must be red, not a silent exit 0.
+          HIVE_TEST_REQUIRE_BEHAVIOURAL: '1'
         run: bash deploy/test_watchtower_socket_contract.sh
 
       # #4204: the socket guard above covers the auto-update profile, and the
@@ -174,6 +186,12 @@ jobs:
       # and just loses agent state on the next recreate. This is also the
       # snapshot the Podman runtime work (#4188) has to keep passing.
       - name: Standalone service deployment contract (#4204)
+        env:
+          # #5388: ubuntu-latest GUARANTEES python3 with PyYAML (actions/runner-images
+          # ships both preinstalled), so this suite's "python3/PyYAML unavailable"
+          # skip cannot be true here. If it fires, the precondition moved and the
+          # suite stopped asserting — that must be red, not a silent exit 0.
+          HIVE_TEST_REQUIRE_BEHAVIOURAL: '1'
         run: bash deploy/test_standalone_service_contract.sh
 
       # #4375: the guard above parses docker-compose.yaml and therefore cannot
@@ -215,6 +233,12 @@ jobs:
       # rather than listing them, so a fourth copy of the quick start is gated
       # the day it is written.
       - name: Documented Compose quick start contract (#4477)
+        env:
+          # #5388: actions/checkout GUARANTEES a real git checkout, so this
+          # suite's "not a git checkout" skip — which would silently retire the
+          # assertion that src/.env is gitignored, on a path the quick start
+          # tells operators to put a PAT in — cannot legitimately fire here.
+          HIVE_TEST_REQUIRE_BEHAVIOURAL: '1'
         run: bash deploy/test_compose_quickstart_contract.sh
 
       # #4404: every guard above looks at ONE runtime. test_standalone_service_
@@ -230,6 +254,12 @@ jobs:
       # reason. Entries are swept in reverse too: an exception that no longer
       # matches a real divergence is itself a failure, so the table cannot rot.
       - name: Docker/Podman contract parity (#4404)
+        env:
+          # #5388: ubuntu-latest GUARANTEES python3 with PyYAML (actions/runner-images
+          # ships both preinstalled), so this suite's "python3/PyYAML unavailable"
+          # skip cannot be true here. If it fires, the precondition moved and the
+          # suite stopped asserting — that must be red, not a silent exit 0.
+          HIVE_TEST_REQUIRE_BEHAVIOURAL: '1'
         run: bash deploy/test_standalone_runtime_parity.sh
 
       # bin/gh-wrapper.sh is the merge gate standing behind the proxy hard-deny:
@@ -468,6 +498,12 @@ jobs:
           just --version
       - name: Contributor K8s Workload Generation
         working-directory: .
+        env:
+          # #5388: the "Install just" step above GUARANTEES the binary, and
+          # ubuntu-latest guarantees python3+PyYAML, so neither of this suite's
+          # skips can legitimately fire here. Before this, a failed install
+          # silently turned the whole suite into an exit-0 no-op.
+          HIVE_TEST_REQUIRE_BEHAVIOURAL: '1'
         run: bash src/deploy/test_contribute_k8s_workload.sh
 
       # #4399: the browser terminal must say when it is NOT showing live output.
@@ -516,6 +552,12 @@ jobs:
       # that still dies on the first non-zero command.
       - name: Changelog reminder never blocks a merge (#4440)
         working-directory: .
+        env:
+          # #5388: ubuntu-latest GUARANTEES python3 with PyYAML (actions/runner-images
+          # ships both preinstalled), so this suite's "python3/PyYAML unavailable"
+          # skip cannot be true here. If it fires, the precondition moved and the
+          # suite stopped asserting — that must be red, not a silent exit 0.
+          HIVE_TEST_REQUIRE_BEHAVIOURAL: '1'
         run: bash src/deploy/test_changelog_reminder.sh
 
       # #4408: `just contribute-move` relocates an already-registered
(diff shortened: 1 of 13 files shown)


clubanderson:
## Demonstrated-failing evidence (round 1 — the guard failed against a real bug, in CI)

Recording this in full because it is the substantive finding of the PR, not a footnote.

Commit `b4b404dc` deliberately restored the silent skip in `test_watchtower_socket_contract.sh`'s **PyYAML** branch — the exact bug this PR exists to catch:

```diff
 if ! python3 -c 'import yaml' >/dev/null 2>&1; then
-  hive_test_skip "PyYAML unavailable — cannot make structural assertions"
-  hive_test_report; exit $?
+  echo "  SKIP: PyYAML unavailable — cannot make structural assertions"
+  exit 0
 fi
```

`build-and-test` on that commit came back **green** ([run 33521351567](https://github.com/kubestellar/hive/actions/runs/33521351567)), and the guard printed a **false PASS**:

```
PASS: test_watchtower_socket_contract.sh: the same skip is FATAL (exit 1) under the flag
```

**Root cause.** These suites gate twice — first on `python3`, then on `import yaml`. The first version of the guard denied `python3` outright, so every suite exited at the **first** gate and the PyYAML branch was never reached. The guard was asserting the shape rather than the property: precisely the defect #5388 describes, reproduced one level up in the thing built to prevent it. Reasoning about the code would not have surfaced this; only running it against a planted bug did.

## Three denial mechanisms that silently do not deny

Fixing it turned up a second finding worth recording, because each of these looks obviously correct and is not. I verified all three directly rather than assuming:

| mechanism | what actually happens |
|---|---|
| executable stub that `exit 127` | **still RESOLVES** — `command -v <tool>` succeeds, the suite sails past its own guard and fails later for an unrelated reason, which misreports as "does not honour the flag" |
| non-executable placeholder earlier on PATH | **skipped** by the PATH search, which then finds the real binary further along |
| dangling symlink earlier on PATH | **also skipped**, same outcome — verified: `git RESOLVES (bad): /tmp/dtest/farm/git` |

Only a PATH that genuinely does not contain the tool makes the lookup fail. Each case now copies a symlink farm (`cp -R` over symlinks is cheap), deletes the denied names from the copy, and runs the child with `PATH` set to that copy alone.

This is the same trap #5435 hit with `find … -print -quit` — a probe that looks like an assertion and passes unconditionally.

## The second axis

`NO_PYYAML` keeps `python3` on PATH and fully working and masks only `import yaml`, so the **second** gate is reached. Both halves are asserted before a case is allowed to count — the interpreter must work **and** the import must fail — so a denial that silently stops taking fails loudly instead of quietly proving nothing.

Verified against two synthetic suites, one correctly adopted and one carrying the regression shape:

```
PASS: _probe_ok.sh: skips permissively (exit 0) when PyYAML is absent
PASS: _probe_ok.sh: the same skip is FATAL under the flag
```

and the missing-suite guard caught a registered-but-absent suite on its own:

```
FAIL: _probe_bad.sh is registered but does not exist
      a renamed or deleted suite silently stops being covered
```

## Round 2 — pending

The tip commit still carries the planted regression, so `build-and-test` **must go red**. The repo-wide `v2 CI` queue is heavily backed up, so that run has not started yet; I will post the red run link and then the restored-green run link before asking for review. **Not ready to merge until both are recorded.**

clubanderson:
## ⚠️ This PR was auto-merged mid-demonstration — follow-up in #5520

Prow merged this at `b4b404dc`, whose commit message reads `TEMP DEMO (#5388): restore the silent skip this PR exists to catch — MUST GO RED, reverted in the next commit`. That commit is the *demonstration* half of the standard #5388 set and was never meant to merge. The three commits that fixed the guard and reverted the demo — `9c42f5ff`, `b435c9dd`, `b3ad9f64` — were pushed to the branch but are not in the merge.

So `v4` currently carries both the planted regression and the first version of the guard, which could not catch it.

**#5520 reverts the regression and lands the three missing commits.** Detail there. Summary of what the demonstration established, since this PR's own body promised it:

- The first guard **failed against the planted bug in CI** and printed a false `PASS`. These suites gate twice — `python3`, then `import yaml` — and the guard denied `python3` outright, so every suite stopped at the first gate and the PyYAML branch was never reached. A guard asserting shape rather than property, which is exactly the defect #5388 names.
- Fixing it turned up that **three plausible denial mechanisms silently do not deny**: an executable stub that exits non-zero still resolves; a non-executable placeholder and a dangling symlink are both skipped by the PATH search. All verified directly. Only a PATH that genuinely lacks the tool works.

Worth raising separately: the #5388 standard asks PRs to carry a deliberately-red state briefly, so demonstration commits and auto-merge-on-green interact badly by construction.

PR #5485, opened 2026-09-01
✨ feat(reviewer): escalated-PR adjudication lane (#5480)

clubanderson:
Refs #5480

## What this is

The **reviewer lane**: an explicitly-configured agent role with bounded authority to adjudicate escalated (`needs-human`) hive-authored PRs. Escalation is a one-way door to a human queue with no assigned worker; this is the missing worker the RFC describes (and #4828 / #5471 / #5479 expose).

## Design

**Role, not name.** The lane is enabled by adding an agent with `role: reviewer` (any name) to the config. Routing happens in `BuildAgentMessage`'s hardcoded-fallback seam, *by role*, before the name switch — so:
- Nothing ships enabled: no pack or default materializes a cadence reviewer-role agent. An operator must add one.
- The pack-defined on-demand pre-merge `reviewer` (ADVISORY, `reviewer-advisory.md`, PR-triggered verdict voting) is untouched — its kick resolves via `kick_template` well before this fallback, and `OnDemandAgentsFromPacks` keeps anything *named* `reviewer` off the cadence. A cadence-kicked adjudicator should therefore use a different name (e.g. `adjudicator`) with `role: reviewer`; it then participates in normal `agents_due` cadence kicks like every other role.
- `pkg/classify` is untouched: the role receives lane-filtered issues via the existing `filterByLane` path, compatible with any operator-configured lane keywords.

**Hard ACMM gate.** The work list + adjudication contract render only at ACMM level ≥ 5 (`reviewerLaneMinACMMLevel`). Below that the kick states the lane is dormant and orders a stand-down — an operator adding a reviewer-role agent to an L3 hive does not silently acquire an agent that drains human queues.

**Work list** (`formatReviewerWorkList`, reading `/var/run/hive-metrics/ci-failing.json`):
- `escalated=true` rows ONLY — red-but-still-automated PRs stay in the FIX-BEFORE-NEW lane.
- Rows whose `labels` contain `reviewer-passed` are excluded — see ping-pong prevention below.
- Structurally capped at 3 rows per kick (`reviewerMaxPRsPerKick`), oldest first (ascending PR number per repo — the only age signal the rows carry), remainder summarized.
- Each row: repo, number, branchless checkout instruction (`gh pr checkout N --repo <repo>`), failing check names, CI evidence excerpt (bounded by `redPRFixExcerptRunes`), and the original author agent (unattributed defaults to `scanner`, matching `formatRedPRFixData`).

[...]

Issue #5480, opened 2026-09-01
[v5 RFC] Reviewer lane: an agent role with authority to adjudicate escalated (needs-human) PRs


clubanderson:
## Problem

Escalation is a one-way door to a human queue with no assigned worker. When PRs exhaust their re-engagement budget they get `needs-human` and every automated path stands down — by design. But nothing in v4 or v5 owns that queue: on kubestellar/console (2026-09-01) nine escalated PRs sat invisible for a week while queues climbed, until an operator-authorized ad-hoc review fleet adjudicated all nine in ~1 hour (8 repaired losslessly, 1 closed as duplicate — every fix was mechanical from CI evidence).

The existing `review.RequireApproval` machinery is pre-merge verdict voting on healthy PRs — a different concern. "reviewer" appears in v5 only as a test-fixture agent name.

## Proposal

A **reviewer lane**: an explicitly-configured agent role (off by default; ACMM-gated, e.g. L5+) permitted to adjudicate `needs-human` PRs with a bounded budget:

- **Repair**: read the PR's CI evidence (ci-failing.json excerpt), verify the change is still wanted and lossless vs main (diff/test-count parity), fix on the SAME branch, push, remove `needs-human` so the sweep can merge on green.
- **De-escalate**: when the failure was environmental (base-branch regression since fixed, infra flake), rebase and return the PR to the automated lane.
- **Recommend-close**: duplicate/superseded/lossy PRs get a close recommendation with rationale — actually closing stays operator-only below L6, auto at L6.
- **Audit**: every adjudication lands on the attribution trail (agent_pr_reviewed / pr closed-recommend) and the advisory digest.

Guardrails: per-cycle adjudication cap; never touches human-authored PRs; a PR it repairs that re-escalates goes to a true human with the reviewer's notes attached (no reviewer↔escalation ping-pong — the machinery-generation amnesty (#5471) plus one reviewer pass then human is the full ladder).

## Prior art from today
- #4828 evidence-rich FIX-BEFORE-NEW kicks (what makes repair mechanical)
- #5471 machinery-generation amnesty + 6-attempt budget
- #5479 needs-human dashboard chip (visibility)
The reviewer lane is the missing worker for the queue those three expose.

clubanderson:
Dispatching for **v5** at SHA `dcbca379` (v5 tip).

**Status — implemented on v5:**
[...]

PR #5511, opened 2026-09-01
🧪 formal: Spin model of escalation + reviewer-lane lifecycle

clubanderson:
Adds `src/formal/escalation/` — a Spin (Promela) model of the fix-loop escalation lifecycle composed with the reviewer adjudication lane, verified exhaustively, plus a reporting-only CI lane that keeps the model and the code honest against each other.

The model targets the **post-#5485 machinery**: `pkg/escalation` (`Sweep`, `TryReEngage`, `MachineryVersion=2` amnesty), `pkg/scheduler/reviewer_lane.go` (#5480), and the governor wiring in `cmd/hive/main.go` (`runEscalationSweep` → `writeMergeEligible` → `reapStuckRedPRs`). One PR's lifecycle is modeled as nine interacting processes (Sweeper, Reaper, fix-lane Agent, Reviewer, Human, CI, Merger, staleness Timer, generation bumper) at the protocol level, with the actual production constants (`DefaultThreshold=3`, `MaxReEngagements=6`), SHAs bounded to 4 and machinery generations to 3. The README documents the full model↔code mapping and every abstraction.

## What is proved (exhaustive searches, Spin 6.5.2)

| # | Property | Result | States |
|---|---|---|---|
| P1 | Amnesty is one-shot per generation (per entry lifetime) | holds | 5,598,270 |
| P2 | No instant re-escalation post-amnesty: never in the amnesty pass; needs ≥2 genuinely new red SHAs or a full fresh re-engage budget | holds | 5,598,270 |
| P3 | `ReEngagements ≤ 6` always; per unchanged red SHA ≤ (1+amnesties)×6 — ≤ 12 across one generation bump, ≤ 18 across the 3 modeled generations | holds | 10,374,648 |
| P4a | A `reviewer-passed` PR is never adjudicated by the reviewer again | holds | 2,100,236 |
| P4b | …and, if still escalated, eventually reaches the human queue | **violated — gap G1** | cycle @ 7,207 |
| P5 | Every escalated PR eventually terminal (merged/closed/human-owned), weak fairness | **violated — gap G1** | cycle @ 5,169 |
| P6 | Escalated entries are invisible to the reaper and FIX-BEFORE-NEW | holds | 20,749,779 |

## Design gaps found (no Go changed in this PR; documented in the README)

- **G1 (real, liveness): a reviewer pass on a PR that stays red orphans it.** The reviewer's REPAIR/DE-ESCALATE edits labels only; `Entry.Escalated` is cleared solely by green/prune/amnesty. If the reviewer's push resolves red: the ledger still says escalated (fix lane + reaper excluded), `reviewer-passed` excludes the reviewer, and `NewlyEscala` requires `!e.Escalated` so `needs-human` (and ntfy) can never re-fire — the human queue never sees it again. Red, open, owned by nobody, forever. The same root cause breaks the escalation comment's promise that removing `needs-human` returns a PR to the automated lane. A hypothetical fix (reviewer label edit also resets the ledger, `-DPATCH_REVIEWER`) is verified to restore P4b and P5.
- **G2 (real, counting): a CI-pending observation wipes the ledger.** `Sweep` branches on `!o.Red` with `Red := CIStatus == "failure"`, so pending takes the "went green" path and deletes the entry — the distinct-SHA breaker is probabilistic wherever sweeps catch a push's pending window (witness `w_pending_wipe`).
- **G3 (minor): the merge watcher re-engages escalated PRs.** `TryReEngage` has no `Escalated` guard and `merge_request_watcher.go:276` calls it without the escalated-set check the reaper performs. Benign in effect (dispatch surfaces still exclude escalated rows; P3/P6 hold in the `-DWATCHER` build) but the counter and the "re-engaged fix loop" log lie.
- **G4 (minor): RECOMMEND-CLOSE below ACMM 6 re-adjudicates every kick** — "one reviewer pass per PR, ever" only holds for verdicts that add `reviewer-passed`; at ACMM 6 the invariant holds (verified).

## Mechanics

- `run.sh` runs the full matrix (15 exhaustive verifications, ~90 s) and checks each verdict against its **expected** result: must-hold properties must pass, and the documented counterexamples must keep reproducing — either direction of drift exits nonzero. Portable bash, `spin` + `gcc`/`cc` from PATH.
- `.github/workflows/formal-verify.yml` (job `formal-verify (escalation)`) runs it on PRs touching `src/pkg/escalation/**`, `src/pkg/scheduler/reviewer_lane*`, `src/formal/**`, or the workflow itself, plus `workflow_dispatch`; `sudo apt-get install -y spin` on ubuntu-latest. **Reporting-only — intentionally not a required check.**
- Counterexamples were minimized with `./pan -i` and replayed with `spin -t -p -g`; reproduction steps are in the README.

Refs #5480

### DIFF
--- a/.github/workflows/formal-verify.yml
+++ b/.github/workflows/formal-verify.yml
@@ -1,0 +1,35 @@
+name: Formal Verify
+
+# Exhaustively re-verifies the Spin model of the escalation + reviewer-lane
+# lifecycle (src/formal/escalation/) whenever the modeled Go code or the model
+# itself changes. Reporting-only: this check is intentionally NOT required.
+# The run fails when a property that must hold is violated, or when a
+# documented counterexample (known design gap) can no longer be reproduced —
+# both mean the model and the code have drifted apart and a human should look.
+on:
+  pull_request:
+    paths:
+      - 'src/pkg/escalation/**'
+      - 'src/pkg/scheduler/reviewer_lane*'
+      - 'src/formal/**'
+      - '.github/workflows/formal-verify.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  escalation:
+    name: formal-verify (escalation)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install Spin
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y spin
+
+      - name: Verify escalation lifecycle model
+        run: bash src/formal/escalation/run.sh
--- a/src/formal/escalation/README.md
+++ b/src/formal/escalation/README.md
@@ -1,0 +1,174 @@
+# Formal model: escalation + reviewer-lane lifecycle
+
+A Spin (Promela) model of the fix-loop escalation machinery and the reviewer
+adjudication lane, at the protocol level: one hive-authored PR's lifecycle as
+nine interacting processes. It verifies the invariants the code comments
+promise — and pins down, with minimized counterexamples, the places where the
+shipped machinery does not deliver them.
+
+The model targets the **post-#5485** machinery (v5 with the reviewer lane,
+#5480, merged; machinery-generation amnesty, #5471).
+
+Run everything:
+
+```console
+$ bash run.sh        # requires spin + a C compiler; ~1 minute
+```
+
+`run.sh` runs every property exhaustively and compares each verdict against
+its expected result. Properties that must hold must pass; properties with a
+**documented counterexample** (the design gaps below) must still *fail* — if
+one stops failing, the code or the model changed and a human must re-check.
+Either divergence exits nonzero. CI runs this via
+`.github/workflows/formal-verify.yml` (reporting-only, not a required check).
+
+## Model ↔ code mapping
+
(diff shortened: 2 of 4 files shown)


clubanderson:
Removed `lgtm`/`approved` — this PR has **two real check failures** and my merge automation labeled it in error.

| Check | Note |
|---|---|
| `NOTICE matches the module graph` | infrastructure flake — see below |
| `build-contributor (linux/arm64, ubuntu-24.04-arm)` | **specific to this PR** — passes on #5510 |

**The NOTICE failure is not staleness.** The generator could not reach `cs.opensource.google` during the run and emitted `Source: Unknown` for `golang.org/x/sys` and `golang.org/x/text`. The committed NOTICE has the correct URLs; the *generated* one is degraded. Do NOT follow the check's own advice to run `generate-notice.sh` and commit the result — during an outage that commits `Source: Unknown` for two packages, degrading the file to match a broken run. Rerun instead; I have triggered one.

**The arm64 build-contributor failure needs your attention.** It passes on other PRs in flight, so it is not the shared runner backlog and not a flake elsewhere in the repo. Worth reading the log rather than rerunning blind.

Also flagging the CI context: the repo's Actions runners are backlogged (~43 queued at last check), so a low failure count on a PR with many `queued` checks is absence of evidence rather than a green build. This PR is at 45 done / 3 queued, so its result is largely real.

Not merging until both are resolved.

Issue #5059, opened 2026-08-29
✨ tui T10: client — events feed


Danathar:
Part of #4907. Depends on T2 (#4917, merged as #4937). Read `src/docs/design/tui.md` first.

**Files to create/modify:** `src/pkg/tui/client/events.go`, `client/events_test.go`.

**Build this:** From `dashboard/openapi.json`, take the `/api/events` operation (published — design doc §2.1). Define `type Event struct` with json tags mirroring the spec's schema (timestamp, type/kind, message/description as exposed). Mirror the spec, do not invent fields. Add `func (c *Client) Events(ctx context.Context) ([]Event, error)` using `getJSON`.

**Acceptance criteria:**
- [ ] Fixture JSON in `client/testdata/events.json` matching the spec schema exactly
- [ ] Test asserts decode of all fields and the error path on 500
- [ ] `go build ./...` and `go test ./pkg/tui/...` pass (from `src/`)

**Out of scope:** rendering (T11), the SSE stream (T13a — this is the poll-shaped fetch).

**Verify with:** `cd src && go build ./... && go test ./pkg/tui/...`

---
Wave 3 of the #4907 filing protocol — unblocked by T3 (#5035, merged). Mirrors the T4 client pattern.

PR #5513, opened 2026-09-01
✨ tui T31: wire Events pane to audit refresh

### LINKED ISSUE #5420: ✨ tui T31: wire Events pane to audit refresh
Part of #4907. Read `src/docs/design/tui.md` first. Depends on T10 (#5059), T11 (#5060), and serialized completion of T30 (#5419).

## Observable failure

The Events pane and `Client.Events` are complete in isolation, but no app command calls the client or delivers `panes.EventsMsg`. The pane remains `waiting for data` forever. The SSE stream does not solve this: `/api/events` carries status snapshots, while the operator activity rows intentionally come from the poll-shaped `/api/audit` endpoint.

## Files to create or modify

- `src/pkg/tui/poll.go`
- `src/pkg/tui/poll_test.go`
- app-level tests only where needed to prove root delivery

Do not modify Events pane rendering or the SSE parser.

## Build this

Add an Events refresh command using `Client.Events(ctx)`. On success, broadcast `panes.EventsMsg{Events: events}`. The client already returns newest-first entries, and the pane owns scroll anchoring; do not re-sort, append, or mutate the returned slice in the app.

Preserve the existing poll error invariant: a failed audit read does not send an empty message and therefore does not erase prior rows or reset scroll position. A successful empty list is different—it marks the pane loaded and renders `no events yet`.

The audit endpoint requires read-write or owner access. A 403 is an expected authorization state, not a reason to exit the TUI. Keep the previous pane data and surface the source-specific error through the app's existing nonfatal refresh-error mechanism if one is available by the time this lands; do not turn this task into a new global error-UI design.

T32 owns the final cadence split. Use the current refresh machinery here so the data path is tested before timer orchestration changes.

## Acceptance criteria

- [ ] Startup refresh replaces `waiting for data` with newest-first audit rows.
- [ ] A successful empty response produces the pane's loaded “no events yet” state.
- [ ] A later fetch preserves the Events pane's existing anchored-scroll behavior when newer entries arrive.
- [ ] 403, 500, malformed JSON, and transport failures preserve prior data and do not terminate the program.
- [ ] No call is made to `/api/events` for activity rows; the client remains the sole owner of the `/api/audit` path.

## Out of scope

- SSE framing/reconnect behavior or translating status events into audit entries.
- Events pane layout, message wording, retention, filtering, or search.
- New error chrome.
- Poll cadence separation (T32).

## Verify with

```bash
cd src
go build ./...
go test ./pkg/tui/...
```

---

Held until #5419 merges so the two tasks do not concurrently rewrite `poll.go` and its shared fixture harness.

Danathar:
## Summary

- add the Events refresh command to the root TUI poll batch;
- fetch activity through `Client.Events`, which owns the poll-shaped `/api/audit` contract;
- broadcast successful snapshots as `panes.EventsMsg` without reordering or mutating the returned slice;
- preserve prior Events rows and scroll position when audit refreshes fail.

## Problem and user impact

The Events client and pane were complete, but the root application never connected them. Neither startup nor periodic refresh called `Client.Events`, so the Events pane remained on `waiting for data` even when `/api/audit` contained operator activity.

The existing SSE subscription cannot fill that gap: `/api/events` streams status snapshots, while the operator activity rows displayed by this pane come from the poll-shaped `/api/audit` endpoint.

## Root cause

`model.poll()` batched the Agents, Governor, identity, Tokens, and cost reads but omitted an Events command. Because the root only broadcasts messages produced by those commands, no `panes.EventsMsg` could ever reach the pane.

## Implementation

`src/pkg/tui/poll.go` now adds `fetchEvents()` to the existing concurrent refresh batch. The command calls `m.api.Events(context.Background())` and:

- returns `panes.EventsMsg{Events: events}` for every successful response, including a successful empty slice;
- returns the source-specific `fetchErrMsg{source: "events", err: err}` for authorization, server, decode, and transport failures.

This deliberately uses the root application's existing nonfatal refresh-error path. Failed reads never reach panes, so they cannot replace prior rows with an empty snapshot or reset the pane's scroll anchor. There is no new global error chrome because none exists yet and issue #5420 explicitly keeps that design out of scope.

The command passes the client slice through unchanged. The client remains the sole owner of the `/api/audit` path and newest-first wire contract, while the Events pane remains the sole owner of snapshot replacement, input ownership, and anchored scrolling. The current shared refresh cadence is retained for T32 to split later.

## Regression coverage

`src/pkg/tui/poll_test.go` extends the app-level dashboard fixture and proves the complete data path rather than retesting the client or pane in isolation:

- `TestPollIssuesEveryRead` proves the poll emits an Events message, calls `/api/audit` exactly once, and never calls `/api/events` for activity rows.
- `TestStartupPollFillsTheEventsPane` starts from the waiting placeholder and verifies that a root poll renders the server's newest-first rows.
- `TestSuccessfulEmptyAuditLoadsTheEventsPane` distinguishes a successful empty feed (`no events yet`) from a missing or failed refresh.
- `TestAuditRefreshPreservesTheEventsScrollAnchor` scrolls through the root model, adds a newer server entry, refreshes, and verifies that the previously anchored row stays at the top of the viewport.
- `TestAuditFailuresPreservePriorEvents` covers 403, 500, malformed JSON, and transport failures, asserting that each produces a source-specific nonfatal error and leaves the anchored Events view byte-for-byte unchanged.

## Verification

From `src/`:

```text
go build ./...
# pass

go test ./pkg/tui/...
ok  github.com/kubestellar/hive/pkg/tui
ok  github.com/kubestellar/hive/pkg/tui/client
ok  github.com/kubestellar/hive/pkg/tui/panes
ok  github.com/kubestellar/hive/pkg/tui/theme
```

The local execution environment required writable `TMPDIR`, `GOTMPDIR`, `GOCACHE`, and `CCACHE_DIR` locations and permission for `httptest` loopback listeners; with those environment-only constraints supplied, both repository commands passed.

## Scope

This PR does not modify Events rendering, SSE parsing or reconnect behavior, refresh cadence, filtering/search, retention, or global error UI. It only wires the existing client and pane through the current refresh machinery.

Closes #5420

— hive: backend=codex

### DIFF
--- a/src/pkg/tui/poll.go
+++ b/src/pkg/tui/poll.go
@@ -94,23 +94,23 @@ func (m model) scheduleTick() tea.Cmd {
 
 // poll issues every fetch the client can currently make, as one batch.
 //
-// Six reads today: /api/agents (T4, #5067), the three T29 wired for the
+// Seven reads today: /api/agents (T4, #5067), the three T29 wired for the
 // Governor pane and the header — /api/status for live governor state,
 // /api/config/governor for the evaluation cadence, and /api/hive-id for the
 // hive's identity — and the two T30 wires for the Tokens pane: /api/tokens for
-// the counts and /api/cost for the estimated spend joined onto them. The events
-// fetch (T10) adds one line here and one message type in pkg/tui/panes; the
-// loop, the error policy and the tick scheduling do not change when it lands.
+// the counts and /api/cost for the estimated spend joined onto them, plus
+// /api/audit for the Events pane (T31). /api/events is deliberately absent: it
+// is the long-lived SSE status stream, not the poll-shaped activity feed.
 //
 // EACH FETCH FAILS ALONE. They are separate Cmds in one batch rather than one
-// Cmd making four calls, and that is the failure-isolation property T29 is
+// Cmd making seven calls, and that is the failure-isolation property T29 is
 // about: a dashboard that serves /api/status but forbids /api/config/governor
 // (a common read-only-token shape) must still show a live governor mode, and a
 // hive with no configured identity must not stop the Governor pane loading.
 // Folding them together would make every value only as available as the least
 // available endpoint, because one error return would discard three good
-// results. Batched Cmds also run concurrently, so four reads cost one round
-// trip of wall time, not four.
+// results. Batched Cmds also run concurrently, so seven reads cost one round
+// trip of wall time, not seven.
 //
 // Deliberately NOT polled: /api/health. It exists and would succeed, but
 // nothing in the frame renders it — the header's `ws:` field is SSE connection
@@ -124,6 +124,7 @@ func (m model) poll() tea.Cmd {
 		m.fetchHiveID(),
 		m.fetchTokens(),
 		m.fetchCosts(),
+		m.fetchEvents(),
 	)
 }
 
@@ -301,6 +302,23 @@ func (m model) fetchCosts() tea.Cmd {
 			return fetchErrMsg{source: costFetchSource, err: err}
 		}
 		return costSummaryMsg{summary: summary}
+	}
+}
+
+// fetchEvents reads the newest-first operator activity snapshot.
+//
+// Client.Events owns the /api/audit path and the wire shape. This command does
+// not sort, append to, or copy the returned slice: a successful snapshot is
+// handed straight to the pane, which owns replacement and scroll anchoring. A
+// failed read becomes fetchErrMsg instead of an empty EventsMsg, so 403s and
+// outages cannot erase the last good rows or reset the operator's position.
+func (m model) fetchEvents() tea.Cmd {
+	return func() tea.Msg {
+		events, err := m.api.Events(context.Background())
+		if err != nil {
+			return fetchErrMsg{source: "events", err: err}
+		}
+		return panes.EventsMsg{Events: events}
 	}
 }
 
--- a/src/pkg/tui/poll_test.go
+++ b/src/pkg/tui/poll_test.go
@@ -25,6 +25,17 @@ const agentsFixture = `[
   {"name":"reviewer","id":"agt_3","displayName":"Reviewer","enabled":false,"managed":false,"backend":"claude","model":"claude-sonnet-4-5"}
 ]`
 
+// eventsFixture is the newest-first /api/audit snapshot the Events wiring
+// tests serve. The ordering is intentionally visible in both timestamps and
+// action names so an app-level sort or reversal cannot pass unnoticed.
+const eventsFixture = `{"entries":[
+  {"ts":"2026-09-01T12:04:05Z","user":"operator","action":"newest","agent":"scanner"},
+  {"ts":"2026-09-01T12:04:04Z","user":"governor","action":"anchored","agent":"quality"},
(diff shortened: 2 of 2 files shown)


clubanderson:
## Two PRs implement #5420 — needs one decision

#5513 (Danathar) and #5514 (clubanderson) were written concurrently. My fault: I dispatched an agent without noticing Danathar was already on it. Both are green and `mergeable=true`, both touch exactly `src/pkg/tui/poll.go` and `poll_test.go`, so only one can land.

| | #5513 | #5514 |
|---|---|---|
| `poll.go` | +25/-7 | +61/-7 |
| `poll_test.go` | +174/-11 | +534/-0 |
| Tests added | 4 | 12 |

**Neither is wrong.** Both get the crux of this issue right — the distinction between a *failed* audit read (must not clear rows or reset scroll) and a *successful empty* list (must mark the pane loaded). Both also guard the `/api/audit` vs `/api/events` boundary, which the issue text got backwards: #5059 already established that `/api/events` is a long-lived SSE stream of status snapshots and cannot serve activity rows.

**#5513** is the leaner change — four focused tests covering startup fill, empty-success, scroll anchoring, and failure preservation.

**#5514** carries mutation evidence for five defect classes, and its complementary-PASS argument is the part worth reading: M1 (failed read emits an empty frame) and M2 (empty success suppressed) are caught by *different* tests, each still passing under the other's mutation. That proves the two paths are independently sensitive rather than one assertion double-counted — which is exactly the failure mode that would let this ship half-working.

It also reports two false negatives it hit and corrected, which is why I weight its evidence: a scripted mutation run reported M1 as PASS because a shell heredoc mangled the patch and it silently tested *unmutated* code; and its endpoint-boundary mutant was initially masked by a vacuity guard firing first on a 404. Both were caught and re-run properly. Given five vacuous checks surfaced in this repo today, a report that documents its own near-misses is more trustworthy than one that reports clean.

### Recommendation

**#5514**, on test depth — but this is a close call and #5513 is a perfectly sound implementation. If the preference is the smaller diff, #5513 lands the same behaviour.

Whichever loses should be closed rather than rebased; they are alternatives, not complements. Neither will be merged until you decide.

clubanderson (approved):
Reviewed. **Approve.**

The behaviour is right, and the part I want to call out is the comment maintenance. This does not just add a line to `poll()` — it updates the surrounding contract comment from "Six reads" to "Seven", corrects "one Cmd making four calls" to "seven" in the failure-isolation paragraph, and replaces the stale prediction about what the events fetch *will* look like with what it *is*. Stale comments are how the next person gets misled, and most PRs leave them.

The `/api/events` note is worth keeping: it records that the absence is deliberate — that endpoint is the long-lived SSE status stream, not the poll-shaped activity feed. The issue text had this backwards, so without the note someone would eventually "fix" it.

I confirmed the two paths this issue turns on are both covered: a **failed** audit read must not clear rows or reset scroll, and a **successful empty** list must mark the pane loaded and render `no events yet`. Conflating them is the way this ships half-working.

One property worth preserving as `poll.go` grows: those two paths need *independently* sensitive tests. Mutating "failed read emits an empty frame" and "empty success is suppressed" should each break a **different** test, with the other still passing. A single assertion covering both reads as coverage while silently permitting one of the two regressions.

Context, not a change request: a duplicate implementation (#5514) was open against #5420 because I dispatched an agent without checking you were already on it. That is closed; this is the one landing.

Issue #5492, opened 2026-09-01
🐛 bug: config writes apply immediately but the dashboard lags ~30s before showing them


clubanderson:
Reported by the operator:

> when i moved to l5 from l4, the update took affect, but i did not see the changes on the dashboard for about 30 seconds. Same was true when i changed the cadence for the telemetry agent

## Observable behaviour

Two independent config writes — an **ACMM level change (L4 → L5)** and an **agent cadence change (telemetry)** — both applied correctly server-side but were invisible in the dashboard UI for roughly 30 seconds.

The write is not the problem. The state genuinely changed; the operator simply had no way to see it, and during that window the UI actively displayed the **old** value. Two writes on unrelated surfaces showing the same ~30s delay points at a shared refresh path rather than a bug in either handler.

## Likely mechanism (needs confirmation)

`src/pkg/dashboard/static/index.html` contains a family of independent 30-second poll constants:

| Line | Constant |
|---|---|
| 5825 | `HISTORY_REFRESH_MS = 30000` |
| 7916 | `INFERENCE_MODEL_CACHE_TTL_MS = 30000` |
| 8070 | `CLI_MODEL_CACHE_TTL_MS = 30000` |
| 9765 | `GH_AUTH_POLL_MS = 30000` |
| 9862 | `NOUS_POLL_MS = 30000` |
| 10556 | `KB_POLL_MS = 30000` |
| 12525 | `AUDIT_POLL_MS = 30000` |
| 12644 | `LEADERBOARD_POLL_MS = 30000` |

A 30s poll means the **expected** latency after a write is uniformly distributed in [0s, 30s] — the reported ~30s is simply the unlucky end. That matches the symptom without requiring anything to be broken.

Note there IS an SSE stream (`new EventSource(/api/events)`, line 15984), so the mechanism for immediate updates already exists. The question is whether ACMM level and agent cadence are carried on it, or whether those panels are poll-only. Two model-cache TTLs in the list are also worth checking — a cache TTL layered on top of a poll interval can add a second 30s window, which would explain a delay that occasionally exceeds 30s.

## Why this is worth fixing rather than documenting

An operator who changes an ACMM level and sees the old level still displayed has no way to distinguish "the write is still propagating" from "the write silently failed". The natural response is to apply it again. For a cadence edit that is harmless; for **ACMM it is not** — the level change can replace the roster, reset per-agent mode overrides, and pause or resume agents, so a re-apply is a second fleet-wide reconciliation.

[...]

Issue #5421, opened 2026-09-01
✨ tui T32: split SSE reconcile and activity poll cadences


Danathar:
Part of #4907. Read `src/docs/design/tui.md` first. Depends on the completed live-data wiring T29 (#5418), T30 (#5419), and T31 (#5420).

## Root cause and user impact

T13b correctly stretches the reconciliation poll from 5 seconds to 60 seconds while SSE is healthy, because SSE replaces status polling for data the stream carries. Tokens/cost and audit activity are not carried by that stream. Once T30/T31 add them to the same `m.poll()` batch, a healthy WebSocket/SSE indicator would paradoxically make those panes twelve times staler.

The app needs two refresh classes, not one timer with every endpoint attached.

## Files to create or modify

- `src/pkg/tui/poll.go`
- `src/pkg/tui/poll_test.go`
- `src/pkg/tui/app.go` and SSE tests only for timer state/message handling

Do not change clients or pane rendering.

## Build this

Separate refresh work by whether SSE covers it:

1. **Reconciliation/status loop:** agents, governor status/header identity, and other state carried by or reconciled with the status stream. It keeps the existing behavior: 5 seconds while disconnected, 60 seconds while healthy, immediate fallback fetch on drop, and generation-based retirement of superseded chains.
2. **Activity/telemetry loop:** tokens, estimated cost, and audit events. It runs every 5 seconds regardless of SSE health because no stream event can update those panes.

Give the fixed activity loop its own message/generation state so reconnects cannot create duplicate timer chains. Start both loops immediately in `Init`; neither may wait one interval before its first data. A slow or failed endpoint must not prevent either loop from rearming, and changing the reconciliation cadence must not reset or accelerate the activity loop.

Keep request batching within each class, and preserve every existing “last successful frame survives a failed fetch” invariant.

## Acceptance criteria

- [ ] Before SSE is healthy, exactly one reconciliation chain and one activity chain run at 5 seconds.
- [ ] Receiving an SSE event stretches only reconciliation to 60 seconds; activity remains 5 seconds.
- [ ] Dropping SSE retires/replaces only reconciliation, performs its immediate fallback fetch, and does not duplicate activity requests.
- [ ] Repeated connect/drop/reconnect cycles leave exactly one live chain of each class.
- [ ] Both classes fetch immediately at startup and rearm before network commands complete.
- [ ] Tests inject tick messages and use short durations/generations; no test sleeps production intervals.
[...]

Issue #5060, opened 2026-08-29
✨ tui T11: events pane (scrollback list)


Danathar:
Part of #4907. Depends on T3 (#5004, merged as #5035) and T10.

**Files to create/modify:** `src/pkg/tui/panes/events.go` and its golden testdata only.

**Build this:** Replace the stub. The pane holds `[]client.Event`, updated via a `panes.EventsMsg` (define it here; the app sends it in T12/T13b). Render a scrollback list per the design sketch — `HH:MM:SS  message` rows, newest visible; `j`/`k` scroll when focused. Truncate to pane width, no wrapping. Content only, per the T3 contract.

**Acceptance criteria:**
- [ ] Golden test at 48x12 with a fixture of more events than fit, `panes/testdata/events.golden`
- [ ] Scroll movement covered by a unit test (not via the golden file)
- [ ] `go build ./...` and `go test ./pkg/tui/...` pass (from `src/`)

**Out of scope:** fetching, live streaming (T13b).

**Verify with:** `cd src && go build ./... && go test ./pkg/tui/...`

---
Wave 3 of the #4907 filing protocol — unblocked by T3 (#5035, merged). Mirrors the T5 pane pattern.

PR #5514, opened 2026-09-01
✨ tui T31: wire Events pane to audit refresh

### LINKED ISSUE #5420: ✨ tui T31: wire Events pane to audit refresh

clubanderson:
Fixes #5420

Part of #4907. Follows #5502 (T30), which rewrote `poll.go`; this extends that file rather than the shape described in the issue text.

## What was broken

`Client.Events` (#5059) and the Events pane (#5060) were both complete and both tested in isolation, and the pane still showed `waiting for data` forever — nothing called the client or delivered `panes.EventsMsg`. This adds the one missing wire.

## Shape

`fetchEvents` resolves to `panes.EventsMsg` **directly** rather than to an app-level message. `fetchGovernor` and `fetchTokens` exist in their indirect, cached form because their pane's frame is a JOIN of two independently failing endpoints, so a message sent straight from one fetch would carry a zero for whatever the other knows. An `EventsMsg` is complete the moment `/api/audit` answers — nothing to join, nothing to cache — so routing it through the model would add a hop that could only lose or reorder data. This is `fetchAgents`' shape, for `fetchAgents`' reason.

`panes.EventsMsg` therefore needs **no case in the app's `Update` switch**: it reaches the pane through the existing default `broadcast`. `app.go` is unmodified. `TestEventsReachThePaneThroughTheRootUpdate` pins that, because adding a case that forgot to broadcast would compile and pass every poll-level test while silently freezing the pane.

The returned slice is passed through untouched — the client returns newest-first and `panes.Events.replace` re-finds the operator's anchor by value.

## The error contract — the crux

Two outcomes that are one blank pane if conflated:

- **Failed read** returns `fetchErrMsg`, which the app swallows. Prior rows and scroll position survive *by construction*, not by the pane remembering to hold them.
- **Successful empty list** returns a real `EventsMsg`. It marks the pane loaded and renders `no events yet`.

`panes.Events` cannot distinguish either from `EventsMsg{Events: nil}` — it sets `loaded` and replaces rows for both — so the distinction has to hold at the message level.

`/api/audit` is the only poll read requiring read-write or owner access, so **403 is the expected state for a read-only token**, not an error condition. It travels the same nonfatal `fetchErrMsg` path as any other non-2xx and does not terminate the TUI. No new error chrome was added; `fetchErrMsg` already carries `source` for whichever later task surfaces it.

## Mutation evidence

Per #5388's standard, every assertion was proven capable of failing. Each defect was introduced, the specific test confirmed failing, then reverted.

| # | Defect introduced | Result |
|---|---|---|
| M1 | Failed read returns `panes.EventsMsg{}` instead of `fetchErrMsg` | **CAUGHT** — `TestFailedAuditReadPreservesPriorEvents` and `TestForbiddenAuditDoesNotStopTheOtherPanes` FAIL; `TestSuccessfulEmptyAuditMarksThePaneLoaded` still PASSES |
| M2 | Successful empty list suppressed into a `fetchErrMsg` | **CAUGHT** — `TestSuccessfulEmptyAuditMarksThePaneLoaded` FAILS; the two failure-path tests still PASS |
| M3 | App reverses the newest-first feed | **CAUGHT** — `TestEventsArriveNewestFirstUnreordered` + `TestNewerEventsPreserveTheAnchoredScroll` FAIL; `TestStartupPollFillsTheEventsPane` still PASSES |
| M4 | `fetchEvents` unwired from `poll()` | **CAUGHT** — 4 wiring tests FAIL; the root-delivery test still PASSES |
| M5 | Client fetches activity rows from `/api/events`, **and the harness serves audit rows there** so the fetch succeeds | **CAUGHT** — `/api/audit was read 0 times, want exactly 1` + `the poll made 1 request(s) to /api/events for activity rows` |

The complementary PASS results matter as much as the FAILs: they show M1/M2 are caught by *different* tests, so the failed-read and successful-empty paths are independently sensitive rather than one check double-counted.

Two notes on how that evidence was produced, since the brief asked for explicitness:

- **M5's first form was a false negative.** Simply repointing the client at `/api/events` made the fetch 404, so the test's own vacuity guard (`the poll delivered no events at all; the endpoint assertions below would be vacuous`) fired first and masked the endpoint counter. The counter assertion was unproven until the harness was also patched to *serve* audit rows on `/api/events`, making the misrouted fetch succeed so only the counter could catch it. The guard worked correctly; it just meant the assertion behind it needed a sharper mutant.
- **An earlier scripted mutation run reported M1 as a false PASS.** Shell heredoc quoting mangled the replacement pattern, so the patch silently no-op'd and the run tested unmutated code. Every mutation in the table above was subsequently applied via a tool that fails loudly on a non-matching pattern, and each was confirmed present in the file before running.

`TestNewerEventsPreserveTheAnchoredScroll` also caught a wrong assertion of my own: I initially asserted the newest entry becomes visible after a refresh, but for a scrolled-back operator the correct behaviour is that it does *not* — it lands above the viewport. The test now asserts the anchored frame is byte-identical and proves the refresh landed by scrolling back to the top.

## Cadence — expected gap, T32's to close

This uses the **current** refresh machinery and adds no competing timer. As #5502 documented, with SSE connected the poll stretches to `sseReconcileInterval` (60s) and the SSE payload carries no audit data — so in a healthy session Events refreshes every 60s rather than 5s. That is expected here and is exactly what T32 (#5421) fixes. Not addressed in this PR.

## Scope

`src/pkg/tui/poll.go` and `poll_test.go` only. Events pane rendering and the SSE parser are untouched, as are `app.go` and everything under `src/pkg/dashboard/` (#5492's territory).

## Verified

- `go build ./...` from `src/` — clean across the module
- `go test ./pkg/tui/...` — all four packages green
- `gofmt -l` clean, `go vet ./pkg/tui/...` clean

CI is the gate; the above are local iteration only.

### DIFF
--- a/src/pkg/tui/poll.go
+++ b/src/pkg/tui/poll.go
@@ -94,13 +94,14 @@ func (m model) scheduleTick() tea.Cmd {
 
 // poll issues every fetch the client can currently make, as one batch.
 //
-// Six reads today: /api/agents (T4, #5067), the three T29 wired for the
+// Seven reads today: /api/agents (T4, #5067), the three T29 wired for the
 // Governor pane and the header — /api/status for live governor state,
 // /api/config/governor for the evaluation cadence, and /api/hive-id for the
-// hive's identity — and the two T30 wires for the Tokens pane: /api/tokens for
-// the counts and /api/cost for the estimated spend joined onto them. The events
-// fetch (T10) adds one line here and one message type in pkg/tui/panes; the
-// loop, the error policy and the tick scheduling do not change when it lands.
+// hive's identity — the two T30 wires for the Tokens pane (/api/tokens for the
+// counts and /api/cost for the estimated spend joined onto them), and T31's
+// /api/audit for the Events pane. T31 was the one line this comment predicted
+// the events fetch would cost: the loop, the error policy and the tick
+// scheduling did not change when it landed.
 //
 // EACH FETCH FAILS ALONE. They are separate Cmds in one batch rather than one
 // Cmd making four calls, and that is the failure-isolation property T29 is
@@ -109,8 +110,14 @@ func (m model) scheduleTick() tea.Cmd {
 // hive with no configured identity must not stop the Governor pane loading.
 // Folding them together would make every value only as available as the least
 // available endpoint, because one error return would discard three good
-// results. Batched Cmds also run concurrently, so four reads cost one round
-// trip of wall time, not four.
+// results. Batched Cmds also run concurrently, so every read costs one round
+// trip of wall time, not one each.
+//
+// /api/audit is the sharpest case for that isolation rather than an exception
+// to it. It is the ONE read here that requires read-write or owner access, so
+// the read-only token that reaches every other endpoint gets a 403 from it and
+// nothing else. Folded in, that single authorization boundary would darken the
+// whole frame; kept separate, it costs exactly one pane its refresh.
 //
 // Deliberately NOT polled: /api/health. It exists and would succeed, but
 // nothing in the frame renders it — the header's `ws:` field is SSE connection
@@ -124,6 +131,7 @@ func (m model) poll() tea.Cmd {
 		m.fetchHiveID(),
 		m.fetchTokens(),
 		m.fetchCosts(),
+		m.fetchEvents(),
 	)
 }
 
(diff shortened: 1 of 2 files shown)


clubanderson:
## Two PRs implement #5420 — needs one decision

#5513 (Danathar) and #5514 (clubanderson) were written concurrently. My fault: I dispatched an agent without noticing Danathar was already on it. Both are green and `mergeable=true`, both touch exactly `src/pkg/tui/poll.go` and `poll_test.go`, so only one can land.

| | #5513 | #5514 |
|---|---|---|
| `poll.go` | +25/-7 | +61/-7 |
| `poll_test.go` | +174/-11 | +534/-0 |
| Tests added | 4 | 12 |

**Neither is wrong.** Both get the crux of this issue right — the distinction between a *failed* audit read (must not clear rows or reset scroll) and a *successful empty* list (must mark the pane loaded). Both also guard the `/api/audit` vs `/api/events` boundary, which the issue text got backwards: #5059 already established that `/api/events` is a long-lived SSE stream of status snapshots and cannot serve activity rows.

**#5513** is the leaner change — four focused tests covering startup fill, empty-success, scroll anchoring, and failure preservation.

**#5514** carries mutation evidence for five defect classes, and its complementary-PASS argument is the part worth reading: M1 (failed read emits an empty frame) and M2 (empty success suppressed) are caught by *different* tests, each still passing under the other's mutation. That proves the two paths are independently sensitive rather than one assertion double-counted — which is exactly the failure mode that would let this ship half-working.

It also reports two false negatives it hit and corrected, which is why I weight its evidence: a scripted mutation run reported M1 as PASS because a shell heredoc mangled the patch and it silently tested *unmutated* code; and its endpoint-boundary mutant was initially masked by a vacuity guard firing first on a 404. Both were caught and re-run properly. Given five vacuous checks surfaced in this repo today, a report that documents its own near-misses is more trustworthy than one that reports clean.

### Recommendation

**#5514**, on test depth — but this is a close call and #5513 is a perfectly sound implementation. If the preference is the smaller diff, #5513 lands the same behaviour.

Whichever loses should be closed rather than rebased; they are alternatives, not complements. Neither will be merged until you decide.

clubanderson:
## Correction to my comparison above: CI evidence is not equal between these two

I recommended this PR on test depth. That still holds on the merits of the tests, but I was comparing "0 failures" on both when the two are not comparably tested.

Measured just now:

| | completed | queued | in_progress |
|---|---|---|---|
| #5513 | **24** | 5 | 7 |
| #5514 | **2** | 23 | 0 |

The repository's Actions runners are backed up — 50 queued against 5 in-progress repo-wide — so this is a capacity issue, not anything wrong with this branch. But it means #5513's green is largely established and **#5514's is not yet evidence of anything**. Zero failures out of two completed checks is absence of evidence, not a passing build, and I presented it as though it were the same signal.

This does not change which implementation I think is better tested — the mutation evidence stands on its own, and M1/M2 being caught by *different* tests is still the property worth having. But if the decision is being made now rather than after the queue drains, #5513 is the one with demonstrated CI.

Re-check both once runner capacity frees up. I have not merged either.

PR #5515, opened 2026-09-01
🐛 fix(escalation): sync reviewer verdicts into the ledger — close orphan path proven by Spin model

clubanderson:
Fixes the gaps documented in #5511 (G1, G3, G4). Refs #5480.

## G1 — the reviewer-pass orphan (the real bug)

The reviewer lane's REPAIR/DE-ESCALATE verdict is expressed as **label edits only**: `needs-human` removed, `reviewer-passed` added, fix pushed. The escalation ledger was never told — `Entry.Escalated` stayed `true`. So if the reviewer's pushed fix resolved **red** again, the PR was excluded from *every* lane, forever:

- FIX-BEFORE-NEW and the reaper skip escalated rows (`row.Escalated` from the ledger);
- `reviewer-passed` excludes it from the reviewer lane;
- `NewlyEscala` requires `!Entry.Escalated`, so the `needs-human` label (and ntfy) could never re-fire — the human queue is the label, and the label never came back.

Red, open, owned by nobody. The Spin model in `src/formal/escalation` (#5511) proved this is a genuine liveness violation (P5: acceptance cycle at 5,169 states) with only nondeterministic rescues (a `MachineryVersion` bump, or gap G2's pending-wipe happening to fire in the CI window), and proved via its `-DPATCH_REVIEWER` variant that a ledger reset at the reviewer pass closes it.

**Fix (sweep-side, per the proven patch shape)**: `escalation.Observation` now carries the PR's labels (threaded from the enumeration in `runEscalationSweep`), and `Store.Sweep` reconciles the verdict: an `Escalated` entry observed with `reviewer-passed` and **without** `needs-human` is reset the way machinery amnesty resets — un-escalated, `RedSHAs` cleared, fresh re-engagement budget, `Machinery` untouched. Sweep-side because the reviewer's `gh pr edit` happens outside the hub's view; the sweep sees the labels every pass regardless of who edited them. The reset disarms itself (it requires `Escalated`), so fresh reds re-accumulate normally; a later re-escalation fires `NewlyEscala` again, and the `reviewer-passed` exclusion routes it to a human, not back to the reviewer.

## G3 — `TryReEngage` lacked an `Escalated` guard

The merge-request watcher's terminal path calls `TryReEngage` without consulting the escalated set (unlike the reaper), burning re-engagement budget and logging "re-engaged fix loop" for `needs-human` PRs — and a budget burned this way could trigger a premature exhaustion re-escalation after amnesty. Escalated entries now return `false`, guarded after the amnesty block so older-generation entries still get their amnesty release.

## G4 — RECOMMEND-CLOSE repeated every kick below ACMM 6

Below the close-authority level the PR stays open and the row stayed fully adjudicable, so the reviewer re-commented the same verdict on every kick. The kick contract now instructs adding the `reviewer-recommend-close` label alongside the comment, and the work-list builder excludes it exactly like `reviewer-passed`.

## Model re-verification (same PR: the fixes are the modeled default)

The former `-DPATCH_REVIEWER` hypothetical is retired — the model now carries the shipped behavior (sweep-side reconciliation, store-side re-engage guard, recommend-close marker). Full matrix, Spin 6.5.2, exhaustive (`-m500000`, DFS + COLLAPSE, liveness with weak fairness; deepest search depth 384, all searches completed, 0 unexpected verdicts):

| Property | Before (#5511 model) | After (this PR) | States |
|---|---|---|---|
| P4b reviewer→human handoff | **VIOLATED** (cycle at 7,207) | **holds** | 3,217,019 |
| P5 escalated ⇒ eventually terminal | **VIOLATED** (cycle at 5,169) | **holds** | 3,224,760 |
| W1 one adjudication ever, ACMM 5 | witness at 4,113 | **holds** | 3,215,657 |
| W3 no re-engage budget on escalated PR | witness at 207,685 | **holds** | 27,818,920 |
| P1/P2 amnesty invariants | holds | holds | 8,822,910 |
| P3 budget caps | holds | holds | 15,779,561 / 3,215,657 / 15,779,561 |
| P4a / P6 exclusion safety | holds | holds | 3,215,657 / 27,818,920 |
| W1 at ACMM 6 | holds | holds | 2,140,034 |
| W2 pending-wipe (gap G2, still open) | witness | witness (pinned expected-fail) | 7,706 |

`run.sh` expectations flipped for the four closed rows; G2's witness remains the only pinned expected failure. README results table, gap narratives, and the model↔code mapping updated to match.

## Tests

- `TestSweep_ReviewerPassResetsEscalatedEntry` — reset on the label pair, ledger restarts at the new red SHA, re-escalates only on three fresh reds.
- `TestSweep_ReviewerResetRequiresVerdictLabels` — no reset while `needs-human` persists or without a verdict label.
- `TestTryReEngage_RefusesEscalatedEntry` — no budget burned on escalated entries (head-SHA and empty-SHA paths); amnesty still releases old generations.
- `TestFormatReviewerWorkList_EscalatedOnlyAndReviewerPassedExcluded` — extended for the `reviewer-recommend-close` exclusion; contract test asserts the new label instruction.

### DIFF
--- a/src/cmd/hive/main.go
+++ b/src/cmd/hive/main.go
@@ -7520,6 +7520,7 @@ func hivePRObservations(cfg *config.Config, actionable *github.ActionableResult)
 			HeadSHA: pr.HeadSHA,
 			Red:     pr.HasFailingRequiredCheck(),
 			Excerpt: pr.CIFailureExcerpt,
+			Labels:  pr.Labels,
 		})
 	}
 	return obs
@@ -7666,6 +7667,11 @@ func runEscalationSweep(
 			HeadSHA: pr.HeadSHA,
 			Red:     pr.CIStatus == "failure",
 			Excerpt: pr.CIFailureExcerpt,
+			// Labels let Sweep reconcile reviewer-lane verdicts (label-only
+			// edits: needs-human removed, reviewer-passed added) back into the
+			// ledger so a reviewer-repaired PR that goes red again re-enters
+			// the fix lifecycle instead of being orphaned (#5511, gap G1).
+			Labels: pr.Labels,
 		})
 		meta[escalation.Key(repo, pr.Number)] = prMeta{checks: pr.FailingChecks}
 	}
--- a/src/formal/escalation/README.md
+++ b/src/formal/escalation/README.md
@@ -7,7 +7,13 @@ promise — and pins down, with minimized counterexamples, the places where the
 shipped machinery does not deliver them.
 
 The model targets the **post-#5485** machinery (v5 with the reviewer lane,
-#5480, merged; machinery-generation amnesty, #5471).
+#5480, merged; machinery-generation amnesty, #5471) **with the #5511 G1/G3/G4
+fixes applied**: the reviewer-verdict ledger sync in `Store.Sweep`, the
+escalated guard in `Store.TryReEngage`, and the recommend-close work-list
+marker. Model and code moved together — the formerly-hypothetical
+`-DPATCH_REVIEWER` behavior is now the shipped (and modeled) default, and the
+properties it was shown to close (P4b, P5) now hold on the default build.
+Gap G2 remains open and its witness is still pinned as an expected failure.
 
 Run everything:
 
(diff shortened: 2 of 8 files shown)


clubanderson:
Reviewed. **This is the strongest change in the current batch** — and it is the first concrete return on the formal-verification work from #5511.

## Verified rather than taken on trust

I checked the two details the safety argument rests on:

**The reset requires both label conditions.** `e.Escalated && containsLabel(o.Labels, ReviewerPassedLabel) && !containsLabel(o.Labels, NeedsHumanLabel)` — so a reviewer who adds `reviewer-passed` without clearing `needs-human` does **not** trigger it. That matters: a half-applied verdict is exactly the state where an automatic un-escalation would be wrong.

**The `TryReEngage` guard sits after the amnesty block.** Confirmed in the diff. Placing it before would have silently swallowed amnesty releases for older-generation entries — a regression that would look like the fix working.

**The reset is self-disarming.** It requires `Escalated`, so it cannot loop: fresh reds re-accumulate normally, and a later re-escalation fires `NewlyEscala` again with `reviewer-passed` routing it to a human rather than back to the reviewer. That is the property that makes this safe to run unattended.

## Why the bug matters

G1 is a genuine orphan, not a cosmetic one. A PR could end up red, open, and excluded from **every** lane simultaneously — FIX-BEFORE-NEW and the reaper skip escalated rows, `reviewer-passed` excludes it from the reviewer lane, and `NewlyEscala` requires `!Escalated` so the `needs-human` label could never re-fire. Since the human queue *is* the label, nothing would ever surface it. The only escapes were nondeterministic: a `MachineryVersion` bump, or G2's pending-wipe happening to land in the CI window.

That is precisely the class of bug a human reviewer does not find by reading code — four independent exclusion rules composing into a trap. Spin found it as a liveness violation with an acceptance cycle at 5,169 states.

**Sweep-side is the right call.** The reviewer's `gh pr edit` happens outside the hub's view, so a hub-side hook would miss verdicts applied by anyone else; the sweep observes labels every pass regardless of who set them.

## The model re-verification is the part worth keeping

The `-DPATCH_REVIEWER` hypothetical is retired and the model now carries the **shipped** behaviour, with P4b and P5 flipping from VIOLATED to holding across ~3.2M states each, and W3 at 27.8M. Keeping a proven-fixed property as a hypothetical variant would let the model drift from the code silently; folding it into the default is what keeps the model honest.

G2's witness staying pinned as an expected failure is also correct — an open gap that the suite acknowledges beats one it quietly passes over.

## Base branch

Targets `v5`, matching #5480's RFC line and merged #5485. Consistent with #5518 in this same batch. Correct.

## One note, not a blocker

The G4 fix depends on the reviewer actually applying `reviewer-recommend-close` as the kick contract instructs. If a backend ignores that instruction the repeated-comment behaviour returns, and nothing structurally prevents it — the exclusion is label-driven, like the rest of the design. Worth knowing as an operational dependency rather than a code defect.

clubanderson:
Rebased onto current `v5` — clean, no conflicts.

One commit was correctly skipped as already-applied: **#5518 merged `.github/workflows/formal-verify.yml`** ahead of this, so git recognised it upstream and dropped it from the rebase. I verified before pushing that the file is genuinely present on `origin/v5` and in the rebased tree — it simply no longer appears in this PR's diff because it is now identical to the base. Nothing was lost.

The diff is 8 files instead of 9 for that reason alone. The substance is intact: the `reviewer-passed`/`needs-human` reconciliation in `Store.Sweep` (3 `ReviewerPassedLabel` references), the `TryReEngage` escalated guard still sitting after the amnesty block, the `reviewer-recommend-close` exclusion, and the full Spin model under `src/formal/escalation/`.

Re-applying `lgtm` — the force-push cleared it, as Prow does on any new commit.

PR #5489, opened 2026-09-01
✨ feat: add bounded reviewer lane for escalated PRs

Danathar:
## Problem

Hive's deterministic fix-loop breaker correctly stops repeatedly failing agent PRs and applies `needs-human`, but that was the end of the automation ladder. Operators had no opt-in, separately bounded agent lane for the cases where a capable reviewer could safely repair the existing branch, prove that the failure came from the base branch or infrastructure, or document that the PR should be closed. Reusing an ordinary coding agent would also have lacked explicit authority boundaries, durable audit evidence, and a no-ping-pong invariant.

This implements the reviewer lane proposed in #5480 for ACMM L5 and L6 while preserving `needs-human` as a safe default: the built-in agent is cadence-paused, the feature gate defaults off, and no human-authored PR can enter its queue.

## Root cause

Escalation state was previously consumed only as a stop signal. `runEscalationSweep` marked failing rows as escalated, and the scheduler excluded those rows from ordinary repair kicks, but there was no role/config model, queue, policy, or terminal state for one adjudication pass.

The existing failure snapshot also could not by itself support safe de-escalation. A PR can turn green after a base-branch regression is repaired or an infrastructure failure clears; if reviewer membership were derived only from currently red rows, that PR would disappear before a reviewer could verify the cause and deliberately remove `needs-human`.

## Implementation

- Adds `escalation.reviewer` configuration with explicit `enabled`, `agent`, and `max_per_cycle` controls. The lane fails closed below L5, requires the configured target to have `role: reviewer`, and defaults to the built-in `reviewer` with a three-PR cap.
- Adds the reviewer to the L5 and L6 ACMM packs with every governor cadence paused. Both levels use `ISSUES_AND_PRS`, so the role can repair an existing branch and record its result but never receives merge authority.
- Builds a dedicated `reviewer_queue` from agent-authored PRs carrying `needs-human` or newly produced deterministic escalation state. Queue membership is independent of current CI color, while author filtering is independent of policy text and excludes human PRs.
- Injects only that bounded queue into the configured reviewer kick. A durable `hive/reviewer-pass` label removes a PR from every future reviewer assignment.
- Adds L5 and L6 policies for the three adjudication outcomes:
  - repair the same branch;
  - de-escalate only after proving a base regression or infrastructure failure;
  - recommend closure at L5, or close only a duplicate, superseded, or provably lossy PR at L6.
[...]

PR #5518, opened 2026-09-01
✨ feat(quality): add formal verification capability

### LINKED ISSUE #5512: [RFC] formal verification as a quality-lane capability — agent-authored models for protocol-shaped code (v5.x)
## Summary

Extend the quality lane with an optional `formal` dimension: when an agent identifies protocol-shaped code in a target repo (state machines, escalation/retry ledgers, lock ordering, lease/claim logic), it authors a Spin/Promela model plus invariants into the repo's `formal/` directory, CI runs them exhaustively, and violations are filed as issues carrying the minimized counterexample narrative.

## Evidence this works (first run, 2026-09-01)

#5511 introduced a hand-scoped pilot: a Spin model of hive's own escalation + reviewer-lane lifecycle (post-#5485), with a path-gated reporting-only CI job. On its FIRST exhaustive run (state spaces up to ~20M states, ~90s wall clock) it found one severe design gap and three lesser ones that code review, unit tests, and an adversarial review pass had all missed:

- **G1 (severe)**: reviewer verdicts edit labels but not the escalation ledger — a reviewer-repaired PR that regresses is excluded from every lane forever (red, open, owned by nobody). The model also *proved* the candidate fix restores the violated liveness/handoff properties before any Go was written.
- G2: CI-pending observations wipe ledger entries (distinct-SHA breaker is probabilistic).
- G3: merge watcher re-engages escalated PRs (counter corruption).
- G4: recommend-close re-adjudicates every kick below ACMM 6.

Two of hive's last five concurrency/protocol incidents (mutex re-entrancy, mint deadlock) are the same class of bug and would each have been a ~100-line model.

## Proposal

1. **Capability, not default**: config-gated (`quality.formal: true`), ACMM L5+ only. Most repos have no protocol core worth modeling; the agent must justify model-worthiness in the PR that introduces a model (criteria: concurrent actors, retry/ledger state, ordering invariants, prior incident history).
2. **Artifact contract**: `formal/<subsystem>/{model.pml, run.sh, README.md}` — README must map every model state to the code construct it abstracts, and run.sh must exit nonzero on any property drift (expected-pass AND pinned expected-fail, so silent fixes and silent regressions both go red).
3. **CI**: reporting-only job, path-gated to the modeled subsystem + formal/ — never a required check (mirrors #5511's `formal-verify`).
4. **Issue filing**: a violated property files ONE issue per property with the counterexample narrative (plain-language interleaving, not raw spin trails); repeat runs update rather than duplicate (respects the digest-cap direction of #5507).
5. **Maintenance rule**: any PR that touches a modeled subsystem and turns formal-verify red must either update the model with the new invariant story or fix the code — the kick evidence pipeline (FIX-BEFORE-NEW) already delivers the counterexample to the authoring agent.

## Non-goals

Modeling everything; proving functional correctness (this is protocol/liveness scope); required-check gating; Rust-specific tooling (Kani/Creusot/Loom do not apply to Go targets).

## Dependencies

#5511 (pilot model, merged CI job), the G1 fix PR (validates the model-first fix workflow), #5480 (reviewer lane — first consumer of a model-found fix).

Refs #5480, #5511

Danathar:
## Problem

Hive's quality lane can author tests and CI improvements, but it has no operator-facing switch or durable policy contract for formal verification. An operator cannot opt a mature hive into model authoring, and simply editing the built-in quality templates would not be sufficient: local policy overrides, remote prompts, replicas, and manual kicks could all omit the model/artifact/reporting rules.

That leaves the workflow demonstrated by #5511 as a hand-scoped pilot rather than a reusable quality-lane capability. In particular, Hive did not encode when a subsystem is model-worthy, how model verdicts become executable drift checks, or how repeated counterexamples map to one actionable issue per property.

## Root cause

Formal verification was not represented in `config.Config`, and the scheduler had no post-template capability seam for it. Quality behavior was entirely selected through ordinary kick templates. Those templates are intentionally customizable, so using them as the only enforcement point would make `quality.formal` mean different things on different hives and would allow stale overrides to silently omit key invariants.

## Implementation

- Adds top-level `quality.formal` through `config.QualityConfig`. The zero value is off. `FormalEnabled` combines the explicit opt-in with a hard ACMM L5 floor.
- Injects the formal capability contract after all normal prompt resolution in `Scheduler.BuildAgentMessage`. This covers embedded, local, and remote templates plus scheduled/manual kicks and quality replicas.
- Fails closed if a formal-enabled binary cannot load its embedded contract: Hive suppresses that kick rather than launching a quality agent with incomplete safety and maintenance instructions.
- Defines the agent contract in `pkg/policies/defaults/formal-quality-capability.md`:
  - justify model-worthiness for protocol-shaped code before authoring;
  - use `formal/<subsystem>/{model.pml,run.sh,README.md}` and map model state back to production code;
  - pin expected-pass and intentional expected-fail outcomes so fixes and regressions both expose drift;
  - add only path-gated, reporting-only CI;
  - minimize/replay failures and use stable `[formal:<subsystem>:<property-id>]` titles so repeat runs update instead of duplicate;
  - keep issue narration agent-owned rather than granting a PR workflow broad issue-write permission; and
  - require modeled-code changes to fix the code or update the invariant story in the same PR, while preserving FIX-BEFORE-NEW priority.
- Documents enablement, admission criteria, artifacts, counterexample reporting, and maintenance in `src/docs/formal-verification.md`, with links from the agent and operator references and a commented example in `src/hive.yaml.example`.
- Adds the user-visible capability to `CHANGELOG.md`.

## Design decisions and invariants

The ACMM boundary is evaluated as an effective value instead of rejecting config at load time. If an operator downgrades from L5 to L4, formal work becomes inert immediately but the explicit preference is retained; raising the hive back to L5/L6 restores it. This avoids turning a safe downgrade into a startup failure or silently deleting operator intent.

The contract is injected outside customizable quality templates. Editing every shipped quality template was rejected because it would not reach existing local overrides or remote prompt sources. Conversely, the rest of the quality policy remains customizable; only the capability's admission, artifact, reporting, and maintenance invariants are fixed when the operator turns it on.

This does not add a central Spin daemon, vendor Spin, or make formal checks required. Models remain repository-owned and agent-authored, and their exhaustive runner remains ordinary reporting-only CI. Issue creation/update also stays with the quality agent so a raw counterexample is minimized and translated before publication; PR workflows do not gain issue-write permission merely because formal verification is enabled.

## Regression coverage

- `TestFormalQualityCapabilityRequiresOptInAndL5` proves the zero value and L4 boundary are inert while L5/L6 opt-ins activate.
- `TestFormalQualityConfigSerialization` proves the literal YAML surface parses and survives JSON serialization.
- `TestFormalQualityCapabilityReachesCustomizedKicks` proves the complete contract is prepended ahead of a customized quality policy.
- `TestFormalQualityCapabilityIsHardGated` covers opt-out, below-L5, and non-quality-role cases without suppressing ordinary quality work.
- `TestFormalQualityCapabilityAppliesAtL6` and `TestFormalQualityCapabilityAppliesToReplicas` cover the upper maturity level and derived-agent path.
- The full policy tests validate the new embedded Markdown against repository policy invariants.

## Verification

Passed:

```text
go test ./pkg/config -run 'TestFormalQuality' -count=1
go test ./pkg/scheduler -run 'TestFormalQuality' -count=1
go test ./pkg/policies -count=1
go test ./pkg/config -skip '^TestGateFailsClosedWithoutIP6Tables$'
go test ./pkg/scheduler ./pkg/policies
python3 src/scripts/check-docs-links.py
git diff --check
```

The documentation checker inspected 130 files and reported that all relative links and anchors resolve. The broader config run excludes only `TestGateFailsClosedWithoutIP6Tables`: on this host `ip6tables` exists but the test process lacks `NET_ADMIN`, so the command reports an authorization failure where that test fixture expects a missing-binary diagnostic. The formal config tests pass, and this change does not touch the egress-gate code or test.

## Scope boundaries

- The concrete escalation/reviewer model and workflow remain in prerequisite PR #5511; this PR supplies the reusable capability contract and does not duplicate that pilot.
- The capability is limited to protocol safety/liveness modeling; it does not claim functional correctness or ask quality to model every repository.
- Formal CI is explicitly reporting-only and does not alter branch protection.
- This PR does not implement a dashboard toggle; operators enable the trusted YAML setting directly.

Fixes #5512

— hive: backend=codex

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Hive did not historically maintain a complete changelog. This file starts a prag
 
 ## Unreleased
 
+### Added
+
+- The quality lane can now opt into agent-authored formal verification with `quality.formal: true` at ACMM L5/L6. Every quality kick receives the same model-worthiness, `formal/<subsystem>/` artifact, expected-verdict drift, reporting-only CI, counterexample narrative/deduplication, and modeled-subsystem maintenance contract even when its ordinary policy prompt is customized. The zero value is off, and the setting becomes inert on an ACMM downgrade below L5 without being discarded ([#5512](https://github.com/kubestellar/hive/issues/5512)).
+
 ## 2026-09-01 (v4.0.1)
 
 ### Added
--- a/src/docs/README.md
+++ b/src/docs/README.md
@@ -58,6 +58,7 @@ Start with [Architecture](architecture.md) for the system overview, then use the
 ## Configuration and agents
 
 - [Agent configuration](agent-configuration.md) — agent fields, methods, models, pins, cadences, caveman mode, ACMM packs, and live-linked `definition_source` (with its seed-only trust model).
+- [Formal verification](formal-verification.md) — the optional L5/L6 `quality.formal` capability for agent-authored Spin/Promela models, reporting-only CI, and deduplicated counterexample issues.
 - [Advisory digest](advisory.md) — what the digest shows (`max_findings`, `show_all`) and how findings are retired (staleness auto-close, PR-linked auto-close).
 - [Advisory digest staleness](advisory-staleness.md) — when the hub raises the stale-advisory pill and alert, the gates that deliberately suppress it (undelivered App, App cannot write, all agents quiet), and the admin diagnostics that measure hidden staleness.
 - [Governor mode thresholds](https://github.com/kubestellar/hive/blob/v4/src/docs/governor-thresholds.md) — how idle/quiet/busy/surge thresholds scale with repo count, the `threshold_scaling` curves, and when explicit thresholds win.
--- a/src/docs/agent-configuration.md
+++ b/src/docs/agent-configuration.md
@@ -426,6 +426,14 @@ The L5 roster is the canonical worked example — eleven agents, eight on the go
 
 At L5, every agent PR gets a `hold` label automatically. The system proposes; it does not merge autonomously.
 
+The quality lane has an additional opt-in formal-verification capability at
+L5/L6. Set top-level `quality.formal: true` to let quality identify
+protocol-shaped subsystems and author Spin/Promela models, their executable
+property contract, and reporting-only CI. The setting is inert below L5 and
+does not turn modeling into a default for every repository. See
+[Formal verification](formal-verification.md) for the admission, artifact,
+reporting, and maintenance rules.
+
 Telemetry and operations are L5/L6-only agents. They stay absent below L5 and remain paused at L5/L6 until an operator deliberately opts in. Their lane keywords are disjoint: telemetry owns instrumentation and observability terms, while operations owns health, SLO, runbook, incident, rollback, and alerting terms.
 
 Configure that opt-in under **Settings → Project Observability**. This tab is
--- a/src/docs/formal-verification.md
+++ b/src/docs/formal-verification.md
@@ -1,0 +1,102 @@
+# Formal verification in the quality lane
+
+Hive can let the quality agent author and maintain Spin/Promela models for
+protocol-shaped code. This is an optional capability, not another default lane
+and not a claim that Hive proves functional correctness.
+
+## Enablement and maturity gate
+
+Enable it in the trusted hive configuration:
+
+```yaml
+acmm_level: 5
+quality:
+  formal: true
+```
+
+Both conditions are required. `quality.formal` defaults to false, and an
+explicit true is inert below ACMM L5. Downgrading a hive therefore disables
+formal-model authoring immediately without deleting the operator's preference;
+raising it back to L5/L6 restores the capability.
+
+The contract is injected after ordinary policy-template resolution. It reaches
+scheduled and manual kicks, quality replicas, and locally customized or remote
+quality prompts. Operators can customize the rest of the quality policy without
+silently weakening what `quality.formal` promises.
+
+## What is worth modeling
+
+The agent must justify model-worthiness in every PR that introduces a model. A
+candidate should have one or more concrete signals such as:
+
+- a state-machine lifecycle;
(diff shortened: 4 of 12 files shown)


clubanderson (approved):
Reviewed. **Approve** — and the base branch is right, which is worth saying explicitly.

#5512 is filed as a v5.x RFC and this targets `v5`. That is the correct line, and it is the distinction #5489 gets wrong (see below).

The gating is careful in a way that matters for a capability that authors files into target repos:

- `QualityConfig`'s zero value preserves existing test/coverage-only behaviour, so no existing config changes meaning.
- `Formal bool` is opt-in, not opt-out.
- `FormalQualityMinACMMLevel = 5` is a named constant with a stated rationale — below L5 quality is advisory or narrower, so enabling it there would grant behaviour the active ACMM pack does not permit.
- `FormalEnabled(acmmLevel)` keeps the floor in the **effective-value** method rather than in validation. That is the detail I would have flagged if it were missing: a level downgrade silently disables the capability without invalidating persisted config or forgetting the operator's preference. Storing the floor as a validation rule instead would make a downgraded config fail to load.

Ships with `formal-verification.md`, an `agent-configuration.md` entry, `hive.yaml.example`, a policy template, and tests on both the config and scheduler halves. Documentation for an opt-in capability is not optional — an operator cannot opt into something they cannot find.

One thing not verified here, worth a note in the docs if it is not already covered: what happens when the capability is enabled but Spin is unavailable on the runner. Reporting-only CI should degrade to a skip with a stated reason rather than a silent pass — a verification job that quietly does nothing is the failure mode #5388 catalogued, and it would be especially easy to miss in a lane whose whole output is "no violations found".

PR #5520, opened 2026-09-01
🐛 test: revert the #5388 demo regression that reached v4, and land the guard fixes

clubanderson:
## What happened

**#5490 was auto-merged by Prow while its tip commit deliberately carried a planted regression.** That commit is the *demonstration* half of the standard #5388 established — "a new guard should be demonstrated to fail against the bug it exists to catch" — and it is meant to be reverted before review, not merged. Its message says so:

> `TEMP DEMO (#5388): restore the silent skip this PR exists to catch — MUST GO RED, reverted in the next commit`

The merge landed at `b4b404dc`; the three follow-up commits that fixed the guard and reverted the demo (`9c42f5ff`, `b435c9dd`, `b3ad9f64`) were pushed to the branch but never merged. So `v4` currently has the worst of both.

## What is wrong on v4 right now

**1. A real silent skip, shipped.** `test_watchtower_socket_contract.sh`'s PyYAML branch is the planted bug:

```sh
if ! python3 -c 'import yaml' >/dev/null 2>&1; then
  echo "  SKIP: PyYAML unavailable — cannot make structural assertions"
  exit 0
fi
```

That is precisely the defect #5388 item 3 is about — a suite that reports green having asserted nothing — sitting in the file the PR used to demonstrate the problem.

**2. A guard that cannot catch it.** `test_skip_discipline.sh` landed in its first form, which denied `python3` outright. These suites gate **twice** — first on `python3`, then on `import yaml` — so every suite exited at the **first** gate and the PyYAML branch was never reached. Against the planted bug it printed a false **PASS** in CI, on run [33521351567](https://github.com/kubestellar/hive/actions/runs/33521351567):

```
PASS: test_watchtower_socket_contract.sh: the same skip is FATAL (exit 1) under the flag
```

A guard asserting the shape rather than the property — the very thing #5388 names, reproduced inside the tool built to prevent it.

## What this PR does

Reverts (1) and lands the three unmerged fixes for (2). No new scope.

- **`NO_PYYAML` denial axis** — keeps `python3` on PATH and fully working and masks only `import yaml`, so the second gate is actually exercised.
- **Denial by removal, not shadowing.** Three plausible mechanisms were each verified *not* to deny: an executable stub that exits non-zero still **resolves**, so `command -v` succeeds and the suite sails past its own guard; a non-executable placeholder and a dangling symlink are both **skipped** by the PATH search, which then finds the real binary further along. Only a PATH that genuinely lacks the tool works, so each case runs against a per-case copy of a symlink farm with the denied names deleted.
- **Anti-vacuity assertions** — for `NO_PYYAML`, the interpreter must work **and** the import must fail before the case is allowed to count, so a denial that quietly stops taking fails loudly instead of proving nothing.

## Verification

The demonstration that #5490 owed, now on the corrected guard:

```
--- regression present (the state currently on v4) ---
  PASS: test_watchtower_socket_contract.sh: skips permissively (exit 0) when PyYAML is absent
  FAIL: test_watchtower_socket_contract.sh: HIVE_TEST_REQUIRE_BEHAVIOURAL=1 makes the skip fatal
        exit 0 under the flag — this suite's skip is still silent, so it is NOT
        adopted (#5388). Route it through hive_test_skip.
=== 20 passed, 1 failed ===        exit 1

--- regression reverted (this PR) ---
=== 21 passed, 0 failed ===        exit 0
```

Also verified on the corrected guard, against synthetic suites of each shape:

```
PASS: _probe_ok.sh: skips permissively (exit 0) when PyYAML is absent
PASS: _probe_ok.sh: the same skip is FATAL under the flag
FAIL: _probe_bad.sh is registered but does not exist   <- missing-suite guard firing on its own
```

`shellcheck -S error` clean on both changed files. CI is the gate; no local run is offered as proof.

## Note for whoever tunes auto-merge

Worth flagging independently of this fix: a commit whose message opens `TEMP DEMO` and says `MUST GO RED` was merged to `v4`. The #5388 standard requires PRs to *carry* a deliberately-red state briefly, so demonstration commits and auto-merge interact badly by construction. That is a policy question rather than something to fix here.

### DIFF
--- a/src/deploy/test_skip_discipline.sh
+++ b/src/deploy/test_skip_discipline.sh
@@ -91,13 +91,22 @@ fi
 # suites that probe with `command -v` sail past their guard and fail later, and
 # that misreports as "this suite does not honour the flag".
 #
-# So: build ONE sandbox bin of symlinks to everything on PATH, then per case
-# copy the symlink farm minus the denied tools. `command -v <tool>` then
-# genuinely fails, which is exactly what the suites' guards test, while every
-# unrelated utility still works. The farm is built once — rebuilding it per
-# suite is O(PATH) forks and takes minutes.
+# Shadowing with a stub does NOT work, and it is worth recording why: a stub
+# that exits non-zero still RESOLVES, so `command -v <tool>` succeeds and the
+# suite sails past its own guard; and a non-executable placeholder is simply
+# skipped by the PATH search, which then finds the real binary further along.
+# Both were tried. The only thing that makes a lookup genuinely fail is a PATH
+# that does not contain the tool.
+#
+# So: build ONE directory of symlinks to everything on PATH, then per case
+# build a directory of symlinks INTO that farm, omitting the denied tools, and
+# run the child with PATH set to it alone. Symlinking (not copying) keeps each
+# case cheap; every unrelated utility still resolves normally.
 SANDBOX_ROOT="$(mktemp -d)"
 trap 'rm -rf "$SANDBOX_ROOT"' EXIT
+
+REAL_PYTHON3="$(command -v python3 || true)"
+
 FULL_BIN="${SANDBOX_ROOT}/full"
 mkdir -p "$FULL_BIN"
 while IFS= read -r _d; do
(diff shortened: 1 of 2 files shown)


clubanderson:
## ✅ Green, and the demonstration is now complete — both states recorded in CI

`build-and-test` **passes** on `0ff25ed7`: [run 33532073462](https://github.com/kubestellar/hive/actions/runs/33532073462).

The two states this PR owed, both on real CI runs rather than asserted:

| | run | `build-and-test` | what the guard said about the planted bug |
|---|---|---|---|
| **regression present** (the state now on `v4`) | [33521351567](https://github.com/kubestellar/hive/actions/runs/33521351567) | **success** — the false green | `PASS: … the same skip is FATAL` — a lie, the PyYAML branch was never reached |
| **regression reverted + guard fixed** (this PR) | [33532073462](https://github.com/kubestellar/hive/actions/runs/33532073462) | **success** — honest green | 31 assertions, incl. the PyYAML axis that now genuinely exercises that branch |

The important line is the first row: a guard that reported PASS against the exact bug it exists to catch. That is what made this worth fixing rather than quietly re-pushing.

## The guard's own output on this run

```
=== #5388: a skip is fatal where the caller guarantees the precondition ===
  PASS: flag unset: hive_test_skip is permissive and the suite exits 0
  PASS: flag=1: the same skip is fatal and the suite exits 1
  PASS: flag=1 with no skip stays green (the flag is not a blanket failure)

--- registered suites ---
  PASS: test_entrypoint_dangling_keyfile.sh: skips permissively (exit 0) when python3 is absent
  PASS: test_entrypoint_dangling_keyfile.sh: the same skip is FATAL (exit 1) under the flag
  ...
  PASS: test_watchtower_socket_contract.sh: skips permissively (exit 0) when PyYAML is absent
  PASS: test_watchtower_socket_contract.sh: the same skip is FATAL (exit 1) under the flag
  ...
  PASS: test_entrypoint_data_ownership.sh: the same skip is FATAL (exit 1) under the flag
```

**14 suite cases × 2 directions + 3 lib cases = 31 assertions, 0 failed.** The five `PyYAML is absent` cases are the ones the original guard structurally could not reach; `test_watchtower_socket_contract.sh` is the file that carried the planted bug, and its PyYAML case is now a real assertion rather than a false pass.

Also confirmed on this run: `Entrypoint system gitconfig contract` executed for the first time ever (it was wired into no workflow before #5490) and passed **10/0**.

Ready for review. Note the base `v4` still carries the regression until this merges.

clubanderson:
Reviewed. **Approve** — and this should land promptly, because the regression it reverts is live on `v4` right now.

## The revert is correct and necessary

`src/deploy/test_watchtower_socket_contract.sh:74-77` on `v4` today:

```bash
if ! python3 -c 'import yaml' >/dev/null 2>&1; then
  echo "  SKIP: PyYAML unavailable — cannot make structural assertions"
  exit 0
fi
```

That is commit `b4b404dc`, whose own message says `TEMP DEMO (#5388): restore the silent skip this PR exists to catch — MUST GO RED, reverted in the next commit`. It reached `v4` via `8d986a12`; the three follow-up commits that fixed the guard and reverted the demo never merged.

A bare `echo` + `exit 0` cannot fail under any value of `HIVE_TEST_REQUIRE_BEHAVIOURAL` — the suite self-disarms whenever PyYAML is absent, which is exactly the defect #5388 exists to eliminate. This PR restores `hive_test_skip` + `hive_test_report; exit $?`.

To be precise about scope, since I initially overstated it: **`test_lib.sh` itself is fine on `v4`.** `hive_test_skip` does go fatal under the flag and `hive_test_report` returns non-zero when `FAIL > 0`; its `return 0` is deliberate so `set -e` suites are not aborted mid-tally. Exactly one suite is disarmed, not the mechanism.

## The harness fix is the more interesting half

The finding worth recording is that **the first version of this guard printed a false `PASS` against the planted bug**. These suites gate twice — once on `python3`, then on `import yaml` — so denying `python3` outright stopped every suite at the *first* gate and never reached the regressed branch. A guard asserting shape rather than the property it stands for, reproduced inside the tool built to prevent exactly that.

The `NO_PYYAML` axis is the right fix: keep `python3` on PATH and working, mask only the module via a `PYTHONPATH` shim that raises on import. That reaches the second gate, which is where the regression actually lived.

Also worth keeping in the comments — three plausible denial mechanisms **silently do not deny**, each verified rather than assumed:

- an executable stub that exits non-zero still **resolves**, so `command -v` succeeds
- a non-executable placeholder is **skipped** by the PATH search, which finds the real binary further along
- a dangling symlink, likewise

Only a PATH that genuinely lacks the tool makes the lookup fail. Same family as `find -print -quit` exiting 0 on no match.

The `bash`/`mktemp` sandbox precondition check is a good addition: without it a broken sandbox would produce a false **red** across every case, which is the opposite failure but equally misleading.

## The policy question this raises is real

#5388's standard — "demonstrate the guard failing, in the PR that introduces it" — requires a PR to briefly carry a deliberately-red commit. That is **incompatible with auto-merge-on-green by construction**: a demo commit is green precisely when the guard is broken, so it is maximally merge-eligible at its most dangerous moment.

This merged because my own merge automation labeled #5490 on green. The standard is right and worth keeping; the interaction needs handling. A `do-not-merge/demo` label, or keeping demo commits on a throwaway branch and force-pushing the clean history before review, would both work. Flagging for the operator rather than deciding it here.

PR #5534, opened 2026-09-01
fix(entrypoint): keep UID isolation PVC walks off startup path

### LINKED ISSUE #5525: [v5] entrypoint UID-isolation chown walks the entire /data PVC — startup-probe death loop on version flips
On a hosted spoke with a large, long-lived /data PVC, flipping the deployment from the v4 stable image to the v5 edge image produced a CrashLoop: 5+ restarts, exit 137, never ready.

**Root cause**: the v5 entrypoint's per-agent UID isolation runs unconditional `chown -R hive-<agent>:node /data/agents/<agent>` (plus `chmod -R g+rwX` over /data/home caches) on every boot. On a PVC with months of accumulated worktrees this walk exceeds the startup probe budget (120×5s = 10min), the kubelet SIGKILLs the container, and the next attempt starts the walk from scratch — a death loop. During a rolling update it is worse: the old pod is still live on the same RWX PVC, actively creating files while the new pod chowns.

**Observed**: startup probe `connect refused` on :3002 for the full 10min window; container processes showed the entrypoint parked on `chown -R` / `chmod -R` with the hive binary never exec'd.

**Workaround applied**: raised startupProbe failureThreshold 120→360 (30min) on the spoke deployment.

**Proposed fixes** (any subset):
1. Make the per-agent chown incremental like the existing `/data/home perms OK — skipping` fast-path: stat a sentinel (dir owner + a marker file recording the last completed isolation pass + image revision) and skip the recursive walk when it matches.
2. Run the chown in the background after the server starts listening (readiness can lag; liveness shouldn't), holding only per-agent dirs that agents actively need before first kick.
3. Use `find -not -user` to touch only wrong-owner files instead of unconditional -R over everything.
4. Ship a startupProbe budget in the hosted-spoke template that scales with PVC size, or document the threshold bump for large spokes.

Version-flip context: this fires exactly when operators move spokes between channels (stable→edge), which is the moment we least want a 30-minute blind window.

Danathar:
## Problem

Moving a hosted spoke from the v4 stable image to v5 edge can make a large, long-lived `/data` PVC permanently unbootable. The root phase of `src/deploy/entrypoint.sh` recursively changes ownership on every per-agent workspace whose top-level UID does not match the new v5 allocation, and may recursively repair shared-home ownership/modes as well. Those NFS/RWX walks run before the Go server binds `:3002`.

On the reported spoke, the work exceeded the startup probe's ten-minute budget. Kubelet killed the container with exit 137 while the entrypoint was still in `chown -R`/`chmod -R`; because no completion state existed, the replacement container began the same traversal again. The hive never became ready, and a rolling update also left the old pod writing the same RWX tree while the new pod attempted its migration.

## Root cause

The existing top-directory-owner shortcut is not a reliable completion record. A mismatched root correctly requests migration, but there is no durable distinction between “not started” and “partially traversed before SIGKILL.” Keeping the recursive command in PID 1's synchronous root setup therefore couples the startup probe budget to total PVC size.

Shared-home setup had the same scaling problem: `.claude`, `.codex`, and `.bob` were recursively chowned before the existing permission fast-path was evaluated, and a required mode repair recursively chmodded `/data/home` before the listener could start.

## Implementation

- `src/deploy/entrypoint.sh` now performs only bounded top-level checks and marker invalidation on the synchronous root path. Every ownership/mode operation whose cost scales with stored PVC contents runs in `hive_run_uid_isolation_migration`, launched as a background root child before PID 1 drops to `dev`.
- Completion state lives under `/data/.hive/uid-isolation`, which is `0700` and owned by `dev:node`; per-agent UIDs cannot forge it. Marker writes use a temporary file plus rename, so a killed pass cannot expose partial completion.
- The shared-home marker is bound to an ownership-schema revision. Each agent marker is additionally bound to its assigned UID (`revision:uid`), and the entrypoint also checks the agent directory's current owner. A UID allocation change therefore invalidates only the affected agent instead of trusting stale state.
- `uid-map.json` carries the marker directory and schema revision into `pkg/agent`. `Manager.Start` waits for both shared-home and per-agent completion before `sanitizeGitRemotes`, tmux setup, token minting, or backend launch can touch that agent's tree. The dashboard and `/api/health` can become ready while migration proceeds, but agents cannot enter partially migrated state.
- Legacy UID maps omit the new fields and retain their previous behavior. Runtime allocations not provisioned by this entrypoint likewise do not invent a marker contract.
- The existing best-effort permission policy remains intact: an NFS race or missing capability is logged and the completed pass is marked rather than converting a formerly recoverable warning into a permanent fleet-wide hold. Failure to publish a marker is different: the agent remains held and the log names the marker that could not be written.

The marker revision is intentionally independent of the image SHA. Binding it to every edge build would force the full recursive pass on every ordinary rollout and recreate the original scaling failure; maintainers bump the schema revision only when the ownership contract itself changes.

## Regression coverage

`src/deploy/test_entrypoint_uid_isolation.sh` executes the shipped marker helpers and proves that:

- the synchronous per-agent provisioning loop contains no recursive `/data/agents` walk;
- recursive repair is confined to the background worker;
- the UID map exposes the marker contract;
- markers are atomically published and reject the wrong target UID.

`src/pkg/agent/uid_isolation_migration_test.go` proves that shared-home completion alone cannot release an agent, a wrong-UID marker cannot release it, matching markers do release it, legacy maps remain compatible, and cancellation interrupts an agent waiting for migration.

The new shell regression is wired into `v2-ci.yml`. Existing `/data` ownership and shared-home/symlink regression suites also remain green.

## Verification

- `bash -n deploy/entrypoint.sh deploy/test_entrypoint_uid_isolation.sh` — passed.
- `bash deploy/test_entrypoint_uid_isolation.sh` — 4 passed, 0 failed.
- `bash deploy/test_entrypoint_data_ownership.sh` — 11 passed, 0 failed; the root-only behavioral portion skipped as designed on the unprivileged host.
- `bash deploy/test_entrypoint_symlinks.sh` — 8 passed, 0 failed.
- `go vet ./pkg/agent` — passed.
- `go test ./pkg/agent -run 'TestUIDMapSaveLoadRoundTrip|TestLoadUIDMap|TestUIDIsolation|TestAwaitUIDIsolation' -count=1` — passed.
- `go test ./pkg/agent -count=1` — passed in 252.126s from a short detached worktree with tmux socket access.
- `git diff upstream/v5...HEAD --check` — passed.

## Scope

This does not raise probe thresholds or attempt to predict PVC size. It does not remove the existing guarded whole-`/data` migration used when the mount root itself has the wrong owner, and it leaves the deliberately small beads-store repair and runtime permission watchers unchanged. The change is limited to preventing the v5 per-agent/shared-home UID-isolation pass from blocking server startup while preserving the launch-time isolation boundary.

Fixes #5525

— hive: backend=codex

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -106,6 +106,14 @@ jobs:
       - name: Entrypoint /data ownership invariant (#5369)
         run: bash deploy/test_entrypoint_data_ownership.sh
 
+      # #5525: v4 -> v5 flips on long-lived NFS PVCs spent longer than the
+      # startup-probe budget in recursive per-agent chown/chmod, so kubelet
+      # killed the pod before :3002 listened and every restart began again.
+      # Pin both halves of the replacement: recursive walks live only in a
+      # background root worker, and UID/revision-bound markers gate agent launch.
+      - name: Entrypoint UID-isolation migration (#5525)
+        run: bash deploy/test_entrypoint_uid_isolation.sh
+
       # #5370: the arm64 lane probed the PUBLISHED image, so on a PR it
       # validated code already on v4 rather than the change proposed — a PR
       # fixing a startup bug stayed red, one introducing a startup bug went
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Hive did not historically maintain a complete changelog. This file starts a prag
 
 - The quality lane can now opt into agent-authored formal verification with `quality.formal: true` at ACMM L5/L6. Every quality kick receives the same model-worthiness, `formal/<subsystem>/` artifact, expected-verdict drift, reporting-only CI, counterexample narrative/deduplication, and modeled-subsystem maintenance contract even when its ordinary policy prompt is customized. The zero value is off, and the setting becomes inert on an ACMM downgrade below L5 without being discarded ([#5512](https://github.com/kubestellar/hive/issues/5512)).
 
+### Fixed
+
+- Version flips no longer trap hosted spokes with large, long-lived `/data` PVCs in a startup-probe death loop ([#5525](https://github.com/kubestellar/hive/issues/5525)). The v5 entrypoint synchronously ran recursive per-agent `chown` and shared-home `chmod` passes before the Go server could bind `:3002`; on an NFS/RWX volume with months of worktrees the walk exceeded the probe budget, kubelet killed the container with exit 137, and the next boot restarted the same walk from the beginning. Size-dependent permission repair now runs in a root background worker, so the dashboard and health endpoint start independently of PVC size. Protected completion markers bind each finished pass to the ownership-schema revision and target UID, and the agent manager waits for both the shared-home and per-agent markers before touching that agent's tree, preserving UID isolation without putting the server behind the migration. Completed steady-state boots skip the recursive pass, while a changed UID invalidates its marker and repairs only the affected agent before launch.
+
 ## 2026-09-01 (v4.0.1)
 
 ### Added
--- a/src/deploy/entrypoint.sh
+++ b/src/deploy/entrypoint.sh
@@ -68,6 +68,92 @@ HIVE_CONFIG_RUNTIME_LEGACY="/data/hive.yaml.bak"
 HIVE_RUNTIME_USER="dev"
 HIVE_RUNTIME_GROUP="node"
 
+# UID-isolation ownership migrations are deliberately versioned separately
+# from the image. Keying the marker to every image SHA would turn every edge
+# rollout back into an O(size of PVC) walk; bump this only when the ownership
+# contract itself changes and an existing tree needs another full pass.
+HIVE_UID_ISOLATION_REVISION="1"
+HIVE_UID_ISOLATION_MARKER_DIR="${HIVE_UID_ISOLATION_MARKER_DIR:-/data/.hive/uid-isolation}"
+export HIVE_UID_ISOLATION_REVISION HIVE_UID_ISOLATION_MARKER_DIR
+
+hive_uid_isolation_marker_path() {
+  echo "${HIVE_UID_ISOLATION_MARKER_DIR}/agent-$1.ready"
+}
+
+hive_uid_isolation_marker_matches() {
+  [ -f "$1" ] && [ "$(cat "$1" 2>/dev/null || true)" = "$2" ]
+}
+
+hive_publish_uid_isolation_marker() {
+  _marker_path="$1"
+  _marker_value="$2"
+  _marker_tmp="${_marker_path}.tmp.$$"
+  if printf '%s\n' "$_marker_value" > "$_marker_tmp" 2>/dev/null \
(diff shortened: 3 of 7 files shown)


clubanderson (approved):
Reviewed. **Approve.** This is a well-shaped fix for a real production outage, and the design decisions are the right ones.

## Verified rather than taken from the description

**Marker containment.** `mkdir -p` → `chown dev:node` → `chmod 0700` on `/data/.hive/uid-isolation`. The marker *files* are `0644`, but the containment is at the directory, so a per-agent UID cannot forge one. (I initially flagged a missing `chmod 0700` — my grep missed it; it is present and correct.)

**Atomic publish.** Temp file → `chmod 0644` → `mv -f`. A pass killed mid-write cannot expose partial completion, which is exactly the failure that made the original bug a *loop* rather than a slow start.

**The gate actually blocks on both conditions.** `uidIsolationMarkersReady` requires `home.ready` carrying the revision **and** `agent-<uid>.ready` carrying `revision:uid`. So a roster change that shifts UIDs invalidates only the affected agent rather than trusting a stale global pass — the detail that keeps this from becoming a different silent-skip bug.

## The two judgement calls I want to endorse explicitly

**The marker revision is deliberately independent of the image SHA.** Binding it to every edge build would force the full recursive pass on every ordinary rollout and recreate the original scaling failure exactly. That is the kind of choice that looks over-clever until you notice the alternative reintroduces the bug being fixed.

**Failure modes are correctly asymmetric.** An NFS race or missing capability during chown/chmod stays fail-open — logged, pass still marked — preserving existing behaviour rather than converting a recoverable warning into a permanent fleet-wide hold. But a failure to *publish a marker* holds the agent and names the file. Those two are genuinely different: the first is best-effort permission repair, the second is the completion record the safety property depends on. Collapsing them either way would be wrong.

## On the root cause

The diagnosis is the valuable part. The old top-directory-owner shortcut was not a completion record — it could not distinguish "not started" from "partially traversed before SIGKILL." That is why the kubelet kill produced a *loop*: every replacement container restarted the same walk from zero with no memory. Moving the work off PID 1's synchronous path is necessary, but the durable marker is what actually breaks the cycle.

Worth noting the rolling-update case named in the description — old pod still writing the same RWX tree while the new pod migrates it — is nastier than the cold-start case and is handled by the same mechanism.

## Not blocking, but worth a follow-up

`uidIsolationMarkersReady` returns **true** when `markerDir`, `revision`, or `uid` are empty, treating that as "legacy deployment, no contract to await." That is correct for genuine legacy maps, but it is also the one path where a *future* bug — a map written without the new fields by some other code path — would silently skip the gate rather than fail loudly. The `omitempty` JSON tags make that shape easy to produce accidentally.

Not a defect here: `Manager.Start` has no way to distinguish "legacy" from "malformed," and failing closed would brick legacy deployments. But if a marker-contract assertion is ever added to the provisioning side, that is the place.

## CI

2 completed / 24 queued at time of review — not yet a green build. The repo's runners have been backlogged. I am not treating the current zero-failure count as evidence; the merge gate applies as normal.

Issue #5507, opened 2026-09-01
[governor] advisory digests repeat-post to the same issue unbounded — 250 comments on one target issue


clubanderson:
## Observed

From a fleet audit: on a spoke where every agent was login-blocked (so no work could progress), the governor kept posting its advisory digest comment on schedule — roughly one every 4 hours — onto a single tracked issue. The target issue (`ibm/alchemy-logging#686`) accumulated ~250 advisory digest comments, the vast majority reading "0 findings". Nothing suppressed the repeats: the digest content was materially identical cycle after cycle, and the poster has no memory of what it last posted or any cap per target.

This is noise for every subscriber of the target issue, burns API quota, and makes the one digest that *does* change impossible to spot.

## Proposed

1. **Content-hash suppression** — before posting, hash the material digest content (findings set, counts — excluding timestamps/formatting churn) and compare to the hash of the previous digest posted to the same target. If identical, skip the post.
2. **Hard cap on consecutive zero-finding digests per target** — even when content drifts trivially, post at most one zero-finding digest per target per 48h (constant, not a magic number inline).
3. **Audit-only suppression note** — when a post is suppressed, record it in the local audit log only (e.g. `advisory_digest_suppressed` with reason `identical`/`zero_finding_cap`); do not post "suppressed" comments to the issue.

State needed is tiny (last hash + last zero-finding post time per target) and can live beside the governor's existing per-cycle state.

Issue #1927, opened 2026-07-17
Excessive authorization required for looking at a hive


MikeSpreitzer:
https://hive.kubestellar.io/dashboard lists llm-d-incubation/llm-d-fast-model-actuation among "My Hives".

<img width="661" height="725" alt="Image" src="https://github.com/user-attachments/assets/40e59d7e-4d0f-48b5-a2ad-c98f27aa4871" />

When I click on "llm-d-fast-model-actuation", which is a link to https://hosted-llm-d-incubation-llm-d-fast-m-yfzz.apps.fmaas-vllm-d.fmaas.res.ibm.com/ , I see a request to "Sign in with GitHub to access this dashboard".

<img width="965" height="547" alt="Image" src="https://github.com/user-attachments/assets/4b51b0be-1862-4594-8d46-0553eb1dfe5d" />

When I click on "Copy & Open", I get to https://github.com/login/device/confirmation , which asks for a lot.

<img width="871" height="580" alt="Image" src="https://github.com/user-attachments/assets/4149eaf3-5e42-4741-ad5b-5446e746c9df" />

The explanation given there is as follows. If I understand correctly, I am being asked to authorize Hive to do all these things on my behalf and attributed to me. This is asking too much for what I want to do at first, which is just look at the Hive dashboard for that Hive. I also see that this delegation is at the _organization_ level, whereas I want to look at the Hive for just one _repository_.

**Repositories**
Public and private

This application will be able to read and write all public and private repository data. This includes the following:

-     Code
-     Issues
-     Pull requests
-     Wikis
-     Settings
-     Webhooks and services
-     Deploy keys
-     Collaboration invites

Note: In addition to repository related resources, the repo scope also grants access to manage organization attributes and organization-owned resources including projects, invitations, team memberships and webhooks. This scope also grants the ability to manage projects owned by users.

clubanderson:
I dont think this exists in the ui any longer - but you can test and re-open if needed

MikeSpreitzer:
I am still getting this excessive request. In v2 of hive.

clubanderson:
Fix in #3544.

Root cause: the dashboard device-flow login hard-coded `scope=repo` because the login token was being persisted and installed as the hive's write identity. That write path is vestigial — every GitHub write (issues, PRs, comments, merges, advisory digest) goes through the App installation token now; the user token only served as an advisory-digest fallback writer, which we've removed.

[...]

PR #5541, opened 2026-09-01
🐛 fix: two mechanisms that ran and silently did nothing — SHA re-check (#5528) and digest pagination (#5522)

### LINKED ISSUE #5522: 🐛 findDigestComment reads only the first page — past ~50 comments the bot cannot see its own digest and CREATEs instead of EDITing
Split out from #5507, which fixed the runaway at its source but deliberately left this latent bug in place.

## The gap

`findDigestComment` reads only the **first page** of comments — `PerPage: 50`, no pagination. Past ~50 comments on a target issue, the bot can no longer see its own previous digest, so the post path falls through to CREATE instead of EDIT.

That is self-reinforcing: each new comment pushes the bot's own digest one position further out of reach, so it can never recover on its own. It is the mechanism that turned a repeat-post bug into **250 separate comments** rather than 250 edits to one.

## Why it still matters after #5507

#5507's suppression stops the digest being *written* when nothing material changed, which removes the pressure that drove the count up. But the pagination limit is untouched:

- Any target issue that already carries >50 comments — from ordinary human discussion, not just bot noise — is **already past the horizon today**. The bot will CREATE rather than EDIT on the first digest it posts there.
- The ~250 comments on `ibm/alchemy-logging#686` are still there, so that target remains permanently in the broken state even with suppression active.

So #5507 stops it getting worse; it does not restore edit-in-place on any target that has already crossed the line.

## Suggested fix

Paginate `findDigestComment` until the marker is found or the list is exhausted. The digest carries an identifying marker already, so this is a loop, not a redesign.

Worth considering while there: searching newest-first would find a recently-posted digest in one page for the common case, with full pagination as the fallback — the bot's own comment is far more likely to be recent than ancient.

## Acceptance

- [ ] The bot finds its own digest on an issue with >50 comments and EDITs rather than CREATEs.
- [ ] Demonstrated with a fixture that has the marker beyond the first page — a test where the marker sits at position 3 proves nothing.
- [ ] Bounded: pagination must not walk thousands of comments unboundedly on every cycle. State the cap and what happens when it is hit.

Refs #5507, #4818

### LINKED ISSUE #5528: 🐛 bug(bin): enumerate-actionable.sh SHA re-check never runs — xargs -I {} rewrites the python dict literals inside the child script
## Summary

The external-issue SHA re-check in `bin/enumerate-actionable.sh` (the loop that lifts an auto-`hold` once a reporter adds a commit SHA, and resolves markers for closed issues) has never executed successfully. Every entry fails with a python `SyntaxError` that is masked by `|| echo "skip"`.

## Cause

`bin/enumerate-actionable.sh:331`:

```
cat "$sha_recheck_tmp" | xargs -P "$SHA_RECHECK_PARALLELISM" -I {} bash -c '
  ...
issue = d.get("data",{}).get("repository",{}).get("issue")
reporter = (issue.get("author") or {}).get("login","")
comments = issue.get("comments",{}).get("nodes",[])
    if (c.get("author") or {}).get("login","") == reporter
  ...
' _ {} >> "$sha_recheck_results"
```

`xargs -I {}` replaces **every** occurrence of `{}` in the command string — including the four python dict literals inside the quoted `bash -c` body — with the current input line. The child python therefore parses e.g. `d.get("data",kubestellar/hive#123|...)` and raises `SyntaxError`; the `|| echo "skip"` fallback turns that into "skip" for every marker. GNU xargs (production) and BSD xargs behave the same way.

The ADOPTERS batch at line 208 uses the same `xargs -I {} bash -c` shape but its body contains no `{}` literal apart from the argument position, so it is unaffected.

## Consequences

- An external issue auto-held for "no SHA" is never un-held after the reporter supplies one.
- Closed-issue markers are never resolved, so every marker costs one GraphQL call per kick cycle, forever.
- Both failures are silent: the loop logs "skip".

## Fix shape

Do not pass the script body through `-I`. Either `xargs -P N -n 1 bash -c '<body>' _` (the argument arrives as `$1`, no substitution), or put the body in a function/heredoc file and pass only the argument through xargs. The new contract test in #5527 pins the current (broken) behaviour with a comment so the fix can flip the assertion.

## Found while

Writing `bin/test_enumerate_actionable.sh` (#5527) — the script had no tests despite being called from `pkg/dashboard/contribute_ws.go`, `bin/merge-gate.sh`, `bin/issue-classifier.sh` and `bin/kick-agents.sh`. Related smaller findings noted in that PR's body: the header and the emitted `exclusions.labels` claim `do-not-merge` / `auto-qa-tuning-report` / `LFX*` PRs are excluded but the PR filter only drops hold-substring and `blocked`; a marker written this cycle is re-queried by the same cycle.

clubanderson:
Two independent bugs of the same shape: a mechanism runs every cycle, produces
output that reads as a verdict, and has never actually done its job.

Fixes #5528
Fixes #5522

---

## #5528 — the SHA re-check has never run

`bin/enumerate-actionable.sh` fanned the external-issue SHA re-check out as:

```
cat "$tmp" | xargs -P N -I {} bash -c '<script>' _ {}
```

`-I` substitutes **every** occurrence of the replacement string in the initial
arguments — including the four python dict literals that live inside the
quoted child body: `d.get("data",{})`, `(issue.get("author") or {})`,
`issue.get("comments",{})` and `(c.get("author") or {})`. Each was rewritten to
the input line, so the child python raised `SyntaxError` on every entry, on
every run, on GNU and BSD xargs alike:

```
$ printf 'REPO:1:file\n' | xargs -I {} bash -c 'echo "d.get(\"data\",{})"' _ {}
d.get("data",REPO:1:file)
```

**Both masking layers are fixed; either alone leaves the failure invisible.**

**Layer 1 — the clobber.** `xargs -P N -n 1 bash -c '<body>' _` appends the
line as `$1` and leaves the body byte-for-byte untouched.

**Layer 2 — the masking.** The classifier ran as
`python3 -c "..." 2>/dev/null || echo "skip"`, so a crash and a legitimate
`skip` verdict produced the **same string** — a total failure of the loop was
indistinguishable from a cycle with nothing to do. stderr is now captured
rather than discarded, a crash yields the distinct state `error`, and the
interpreter message is logged. `error` matches no branch in the dispatch, so
an unclassifiable entry still declines to act — it just says so.

What was broken, now fixed: an external issue auto-held for a missing commit
SHA was never un-held after the reporter supplied one; closed issues never had
their markers resolved; every unresolved marker cost one GraphQL call per kick
cycle forever.

### What was done with #5527's pinned assertions

PR #5527 added `bin/test_enumerate_actionable.sh`, which **pins the current bug
as expected behaviour** with instructions to flip the pins when it is fixed.
That PR is still open, so its commit is cherry-picked here and the pins are
flipped as part of this fix — otherwise this change would turn its CI red and
look broken. If #5527 merges first, this branch rebases onto it.

Flipped:

- the base `graphql_109` fixture is now the honest pre-answer state (no SHA
  yet). The re-check agreeing with the REST fixture that made #109 eligible for
  hold is now itself asserted — previously the two could disagree unnoticed,
  because the re-check never ran.
- **new run 3** against a fixture where the reporter HAS answered: unhold fires
  exactly once, the marker is stamped `resolved=`, #109 returns to actionable.
- **new run 4**: a resolved marker makes NO GraphQL call — the per-cycle cost
  is gone.
- **new closed-issue case**, the other branch the clobbered classifier could
  never reach: marker stamped `resolved=... closed`, no unhold issued.
- **new negative control for layer 2**: a payload that is valid JSON but
  carries no `data.repository.issue` must not be laundered into a confident
  verdict — no unhold, no `resolved=` stamp.

One more trap worth naming: the harness ships its own `xargs` **shim** (macOS
has no GNU xargs) that only understood `-I`. Left as-is it would have failed
the FIXED script and made this entire section vacuous. It now supports `-n 1`
too, and rejects any other invocation loudly instead of silently doing nothing.

---

## #5522 — findDigestComment read only the first page

`findDigestCommentDetail` read ONE page (`PerPage: 50`, no pagination). Past
~50 comments the bot could no longer see its own digest, so the post path fell
through to CREATE instead of EDIT — and every comment it created pushed the
digest one position further out of reach. That self-reinforcing loop is what
turned a repeat-post bug into 250 separate comments on `ibm/alchemy-logging#686`
rather than 250 edits to one.

#5507 removed the pressure driving the count up. It does not restore
edit-in-place: any target already carrying >50 comments — from ordinary human
discussion, not just bot noise — is past the horizon **today**.

The scan now paginates **newest-first** (`sort=created&direction=desc`) at
GitHub's maximum page size of 100. Newest-first matters: the bot's own digest
is overwhelmingly likely to be recent, so the common case still resolves in
exactly one API call — the same cost as before — with full pagination as the
fallback for anything older.

### The cap

`advisoryDigestScanMaxPages = 10` → the **1,000 most recent comments** per
cycle.

**When the cap is hit** the scan stops and reports *not found* — the same
answer it gives for a genuinely absent digest — so the caller CREATEs a new
digest comment, which every later cycle then finds on page 1 and edits.
Deliberate: a digest buried under the 1,000 most recent comments is effectively
unreachable, and one new comment beats walking unbounded history on every ~60s
cycle. Hitting the cap is logged at `Warn` so the case is visible rather than
silent.

The per-page authorship rules are unchanged, factored into `scanDigestPage`.
Two behaviours preserved across the new page boundary and pinned by tests: the
not-provably-ours bot-authored fallback now carries across pages (a slug
mismatch must not create a duplicate every cycle), and a human-authored digest
comment is still skipped rather than adopted, since an App can never edit one
(#1927, `kalantar-msb/soft-reflective#1`).

---

## Demonstrated-failing evidence

A green run proves nothing here — both mechanisms were green the whole time.
Each fix was watched failing for the right reason first.

**#5528** — flipped assertions run against the **unfixed** script:

```
FAIL: run 3: SHA in the reporter's comment unholds #109 exactly once   got '0', want '1'
FAIL: run 3: marker is stamped resolved=                               marker still unresolved: ''
FAIL: run 3: #109 is re-admitted to actionable                         got '101,110,111,201', want '101,109,110,111,201'
FAIL: run 4: a resolved marker is not re-queried                       got '0:1', want '0:0'
FAIL: closed issue stamps the marker resolved=... closed               marker: ''
=== 78 passed, 5 failed ===
```

With the fix: **83 passed, 0 failed**. (Baseline check first: the ORIGINAL
pinned test passes 76/76 against the unfixed script, confirming the pins were
genuinely pinning the bug.)

**#5522** — new tests run against the **single-page** implementation:

```
TestFindDigestComment_MarkerBeyondFirstPage   id = 0, want 1130
TestFindDigestComment_NewestFirstFindsRecent  id = 0, want 1399
TestFindDigestComment_ScanIsBounded           pages = 1, want 10
TestFindDigestComment_ExhaustsWithoutMarker   pages = 1, want 3
TestFindDigestComment_BotFallbackSurvives     id = 0, want 1249
TestFindDigestComment_PerPageIsMaximised      per_page = 50, want 100
```

The main fixture places the digest at chronological **index 130 of 260** — the
exact middle, so it is outside page 1 in **both** orderings and cannot be
reached without turning a page. A marker near either end would have been found
by the unpaginated code and proved nothing: an earlier draft of this test put
the marker at index 12, which the buggy code found on page 1, and only the
page-count assertion caught it. That draft was a vacuous check and was
replaced.

Full suite: `go test ./pkg/github/` → `ok 47.164s` (skipping the two
pre-existing sandbox-hanging tests, `TestHiveOpenPRScript_OmittedBaseLeavesBaseUnset`
and `TestCommitGreenNoConfigRequiredChecksFallsBackToAPIThenAllowlist`).

### DIFF
--- a/bin/enumerate-actionable.sh
+++ b/bin/enumerate-actionable.sh
@@ -328,7 +328,15 @@ done > "$sha_recheck_tmp"
 
 SHA_RECHECK_PARALLELISM=8
 if [ -s "$sha_recheck_tmp" ]; then
-  cat "$sha_recheck_tmp" | xargs -P "$SHA_RECHECK_PARALLELISM" -I {} bash -c '
+  # NOTE (#5528): the fan-out MUST NOT use `xargs -I {}`. -I substitutes every
+  # occurrence of the replacement string in the initial arguments — including
+  # the four python dict literals below (d.get("data",{}), (issue.get("author")
+  # or {}), issue.get("comments",{}), (c.get("author") or {})) which live
+  # inside this quoted script body. That rewrote them to the input line and
+  # made the child python a guaranteed SyntaxError on every entry, silently
+  # downgraded to the string "skip" — so SHA-UNHOLD never fired. `-n 1` passes
+  # the line as $1 and leaves the body untouched.
+  xargs -P "$SHA_RECHECK_PARALLELISM" -n 1 bash -c '
     entry="$1"
     repo="${entry%%:*}"
     rest="${entry#*:}"
@@ -347,6 +355,7 @@ if [ -s "$sha_recheck_tmp" ]; then
           }
         }
       }" 2>/dev/null) || { echo "${repo}:${num}:${marker_file}:skip"; exit 0; }
+    py_err=$(mktemp)
     state=$(echo "$result" | python3 -c "
 import json, sys, re
 d = json.load(sys.stdin)
@@ -364,9 +373,18 @@ reporter_text = body + \" \" + \" \".join(
 )
 SHA_RE = re.compile(r\"[0-9a-f]{7,40}\\b\")
 print(\"has_sha\" if SHA_RE.search(reporter_text) else \"no_sha\")
-" 2>/dev/null || echo "skip")
+" 2>"$py_err") || state="error"
+    # #5528 layer 2: a crashing classifier used to emit the same string as a
+    # legitimate "skip" verdict, so a total failure of this loop was
+    # indistinguishable from a run with nothing to do. Keep them separate and
+    # surface the interpreter error instead of discarding it with 2>/dev/null.
+    if [ "$state" = "error" ] || [ -z "$state" ]; then
+      state="error"
+      echo "SHA-RECHECK-ERROR: ${repo}#${num} classifier failed: $(tr "\n" " " < "$py_err" | tail -c 400)" >&2
+    fi
+    rm -f "$py_err"
     echo "${repo}:${num}:${marker_file}:${state}"
-  ' _ {} >> "$sha_recheck_results"
+  ' _ < "$sha_recheck_tmp" >> "$sha_recheck_results"
 fi
 
 while IFS=: read -r repo num marker_file state; do
--- a/bin/test_enumerate_actionable.sh
+++ b/bin/test_enumerate_actionable.sh
@@ -145,18 +145,36 @@ if ! xargs --version >/dev/null 2>&1; then
   # Sequential subset of GNU `xargs -P N -I REPL cmd args...`.
   cat >"${SHIM_DIR}/xargs" <<'XARGSEOF'
 #!/usr/bin/env bash
+# Sequential subset of GNU `xargs -P N [-I REPL | -n 1] cmd args...`.
+# -I REPL  : substitute REPL everywhere in the initial arguments (the GNU
+#            behaviour that caused #5528 — kept so the ADOPTERS batch at
+#            line ~208, which legitimately uses -I, still runs here).
+# -n 1     : append the line as one extra argument, leaving the initial
+#            arguments byte-for-byte untouched. The SHA re-check uses this
+#            form; a shim that only understood -I would silently fail the
+#            fixed script and make this whole section vacuous.
 repl=""
+nargs=""
 while [ $# -gt 0 ]; do
   case "$1" in
     -P) shift 2 ;;
     -I) repl="$2"; shift 2 ;;
+    -n) nargs="$2"; shift 2 ;;
     *)  break ;;
   esac
 done
+if [ -z "$repl" ] && [ "$nargs" != "1" ]; then
+  echo "xargs shim: unsupported invocation (need -I REPL or -n 1)" >&2
+  exit 2
+fi
 while IFS= read -r line; do
   [ -n "$line" ] || continue
   args=()
-  for a in "$@"; do args+=("${a//"$repl"/$line}"); done
+  if [ -n "$repl" ]; then
+    for a in "$@"; do args+=("${a//"$repl"/$line}"); done
+  else
+    args=("$@" "$line")
+  fi
   "${args[@]}"
 done
 exit 0
@@ -253,11 +271,13 @@ files src/other.go                >"${BASE}/repos_acme_primary_pulls_305_files.j
 files src/main.go                 >"${BASE}/repos_acme_secondary_pulls_401_files.json"
 files adopters.md                 >"${BASE}/repos_acme_secondary_pulls_402_files.json"
 
-# Re-check fixture for #109: the reporter later commented with a SHA.
+# Re-check fixture for #109, BASE state: the reporter has not answered yet, so
+# the re-check must find no SHA and leave the hold in place. This must agree
+# with the REST fixture that made #109 eligible for the hold — before #5528
+# was fixed the re-check never ran, so the two could disagree unnoticed.
 cat >"${BASE}/graphql_109.json" <<'EOF'
 {"data":{"repository":{"issue":{"state":"OPEN","body":"It crashes on startup","author":{"login":"newcomer"},
- "comments":{"nodes":[{"author":{"login":"hive-bot"},"body":"please add the SHA"},
-                      {"author":{"login":"newcomer"},"body":"sure: 9f8e7d6c5b4a"}]}}}}}
+ "comments":{"nodes":[{"author":{"login":"hive-bot"},"body":"please add the SHA"}]}}}}}
 EOF
 
 # ── Runner ───────────────────────────────────────────────────────────────────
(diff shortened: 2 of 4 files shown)


clubanderson:
Rebased onto current `v4` — now `mergeable=true`, 5 files → 4.

**#5527 did merge** (17:32Z), contrary to what this branch assumed. Its `bin/test_enumerate_actionable.sh` and its `v2-ci.yml` wiring are both on `v4`, so the cherry-picked copy here collided add/add.

Resolved by taking **this branch's** version of the test file, and the reason is decisive rather than a coin flip:

| | `pinned bug` assertions | shim supports `-n 1` |
|---|---|---|
| `v4` (from #5527) | **7** | no |
| this branch | **0** | yes |

#5527's pins were correct *while the bug existed* — they documented reality. This PR fixes the bug, so the pins must flip, and the flipped version is the one that belongs on top. Keeping `v4`'s side would have left CI asserting that the SHA re-check still never runs, and gone red against the fix.

`v2-ci.yml` dropped out of the diff entirely, since #5527 already landed the identical hunk. That also resolves the ownership-boundary crossing flagged in the report — the wiring is in, and this PR no longer touches a sibling agent's file.

Verified after the rebase: 0 pinned assertions, `xargs -P … -n 1` present in `enumerate-actionable.sh`, and the pagination constants intact in `advisory.go` (`advisoryDigestCommentsPerPage = 100`, `Direction: desc`, 10-page cap — a named constant, not a literal, per the repo's no-magic-numbers rule).

One note for whoever reviews: the shim change matters more than it looks. The harness ships its own `xargs` because macOS lacks GNU xargs, and it only understood `-I`. Left unchanged it would have failed the **fixed** script — making the entire section vacuous in exactly the way this PR exists to eliminate.

clubanderson:
## Collision: #5539 also fixes #5528, and is much further along

My fault — I dispatched an agent onto #5528 and #5522 without checking whether the hive fleet already had one working it. **#5539** covers #5528 and is at 28 completed checks; this PR is at 1 of 24.

They overlap on `bin/enumerate-actionable.sh` and `bin/test_enumerate_actionable.sh`, so both cannot merge.

**The split is clean, though:**

| | #5539 | #5541 (this) |
|---|---|---|
| #5528 xargs + masking | ✅ | ✅ |
| #5522 pagination | ❌ | ✅ (`advisory.go` + tests) |

So #5539 should land #5528, and this PR should reduce to the **#5522 pagination fix only** — `src/pkg/github/advisory.go` and `advisory_pagination_test.go`, dropping both `bin/` files and changing `Fixes #5528, #5522` to `Fixes #5522`.

Nothing is lost by that: #5539 fixes both masking layers the same way (`-n 1` plus real stderr surfacing instead of `|| echo "skip"`), and flips #5527's pinned assertions with a negative control.

**One thing from this branch must not be lost, and it is not in #5539.** The harness ships its own `xargs` shim because macOS lacks GNU xargs, and it only understood `-I`. Left unchanged it fails the *fixed* script — `shim/xargs: line 14: -n: command not found` — making that whole section vacuous in exactly the way these issues exist to eliminate. If #5539 has not taught the shim `-n 1`, it will need that before it can be trusted on macOS. Worth checking there specifically, since CI is Linux and would not surface it.

I have not closed or modified either PR. Flagging for the operator: land #5539 for #5528, trim this to #5522.

clubanderson:
Closing — superseded, and split cleanly rather than abandoned.

**#5539 shipped the `bin/` half and closed #5528**, which made this PR half-redundant and left it `mergeable=false`. #5539 fixes both masking layers the same way (`-n 1` so the child body is not rewritten, plus real stderr surfacing instead of `|| echo "skip"`), flips #5527's pinned assertions, and — the detail that mattered — its harness teaches the bundled `xargs` shim `-n 1`, without which it would have failed the *fixed* script on macOS and made that whole section vacuous.

The pagination half was never redundant, so it is carried forward as **#5566** against current `v4`: `src/pkg/github/advisory.go` and `advisory_pagination_test.go` only, `bin/` untouched and identical to merged `v4`.

Nothing is lost. Everything this branch established about #5522 is in the new PR, including the vacuous-test catch worth restating: the marker fixture at chronological index 12 of 260 was found even by the *buggy* single-page read, so only the page-count assertion tripped. Moved to index 130 — the exact middle, outside page one in both orderings — it fails on the id itself, which is the real symptom.

Thanks for the work here; the split is a merge-order artifact, not a problem with it.

Issue #5422, opened 2026-09-01
📚 tui T26: document the complete terminal dashboard


Danathar:
Part of #4907. Read `src/docs/design/tui.md` first. Depends on completion of the v1 action UI (T17 #5416, T19 #5417, T21 #5414) and live-data/cadence work through T32 (#5421).

## Problem

The public `hivectl` guide still describes an early scaffold: it says Governor and Events are stubs, only half the panes are live, and help/pause/resize do not exist. That is already false for merged work and will become more misleading as the final actions land. The root README and docs index also do not give operators a discoverable path to the terminal dashboard, its authentication variables, keybindings, limitations, or troubleshooting guidance.

Documentation must describe shipped behavior, not the epic's intended end state.

## Files to create or modify

- `README.md`
- `src/docs/README.md`
- `src/docs/hivectl.md`
- `src/docs/design/tui.md` only to mark implementation status or correct paths/contracts that changed during delivery

No Go, JavaScript, generated OpenAPI, or binary media changes.

## Build this

Add a concise root-README entry that introduces `hivectl tui`, links to the operator guide and design document, and positions it as another client of the dashboard API rather than a second Hive runtime.

Rewrite the TUI section of `src/docs/hivectl.md` as an operator reference covering:

- launch command, `--hive`/endpoint selection, `HIVE_DASHBOARD_URL`, and `HIVE_DASHBOARD_TOKEN`;
- the four panes and the actual source/cadence of each;
- the final keybinding table, scope rules, modal behavior, and uppercase safety bindings;
- pause/resume, model apply/restart semantics, queued kick semantics, typed ACMM confirmation/reconciliation receipt, and local-only tmux attach;
- SSE connected/degraded behavior and poll fallback;
- minimum terminal size, help, resize, light/dark adaptation, and quit behavior;
- authorization failures (especially owner-only operations), missing tmux/session behavior, unavailable/partial model catalogues, and stale-last-good data on fetch failure;
- v1 boundaries: self-hosted token auth, local attach only, no remote ttyd embedding, no hub OAuth flow, and no claim of web-dashboard pixel parity.

Update `src/docs/README.md` with discoverable links. Correct obsolete `v2/` paths anywhere this tracker material copied them. Remove “under active construction” claims only when the dependent behavior has actually merged.

## Acceptance criteria

- [ ] Every documented pane/action/key is present in code, and every shipped binding appears in the guide.
[...]

Issue #5419, opened 2026-09-01
✨ tui T30: wire Tokens pane with estimated cost


Danathar:
Part of #4907. Read `src/docs/design/tui.md` first. Depends on T8 (#5057), T9 (#5058), T28 (#5415), and T29 (#5418).

## Observable failure

The Tokens pane and token/cost clients are individually tested, but the app never calls them or sends `panes.TokensMsg`. In a real `hivectl tui` session the pane therefore remains `waiting for data` forever. This contradicts #4907's operator-loop promise and the tracker text that describes all four panes as live.

## Files to create or modify

- `src/pkg/tui/poll.go`
- `src/pkg/tui/poll_test.go`
- app-level tests only where required to prove delivery

Do not change Tokens pane layout or client wire types.

## Build this

Add a refresh command that reads `Client.Tokens` and T28's `Client.Costs`, joins them by canonical agent name, and delivers one `panes.TokensMsg`.

Projection rules are part of the contract:

- Token rows come from `TokenUsage.ByAgentDetail`; input/output counts remain the `/api/tokens` values.
- Cost enriches matching rows from `/api/cost.estimated.by_agent`.
- `CostKnown` is true only for an entry whose source says it is priced; an unpriced zero renders `—`, not `$0.00`.
- Fleet input/output use `TokenUsage.TotalInput` / `TotalOutput`, not a sum of displayed agents, because unattributed sessions are included in the server total.
- Fleet cost uses `estimated.total_usd` and is known only when the summary contains no unpriced models. A partial estimate must not be presented as the complete spend.
- Map iteration must not leak nondeterminism; the pane already sorts delivered rows, so preserve that contract.

Token counts are the primary data. If `/api/tokens` fails, keep the previous pane unchanged. If tokens succeed but `/api/cost` fails, deliver fresh counts with unknown costs rather than holding the entire pane stale behind an optional estimate. Empty successful usage is a loaded zero-usage frame, not “waiting”.

T32 later separates the fixed non-SSE refresh cadence. Until then, use the existing app poll machinery; do not introduce a competing timer in this task.

## Acceptance criteria

- [ ] A startup refresh replaces `waiting for data` with multi-agent token rows and fleet totals.
- [ ] Per-agent and total cost-known semantics are tested for fully priced, mixed, unpriced, and cost-endpoint-failure cases.
- [ ] Unattributed token usage proves the fleet total is taken from the server, not re-summed rows.
[...]

PR #5544, opened 2026-09-01
✨ tui T32: split SSE reconcile and activity poll cadences

### LINKED ISSUE #5421: ✨ tui T32: split SSE reconcile and activity poll cadences

Danathar:
Fixes #5421

## Summary

- Hive's terminal dashboard (`hivectl tui`) draws four panes and refreshes them on a timer. It also subscribes to the dashboard's live event stream, and **slows that timer down while the stream is healthy** — the stream is already pushing updates, so polling hard would just spend requests to re-learn what it just told us.
- That was correct while the timer only fetched things the stream carries. #5419 (T30) and #5420 (T31) then attached **token counts, estimated cost, and the audit activity feed** to the same timer — and the stream carries none of those. So a connected stream stretched those three reads from 5s to 60s: **the Tokens and Events panes became twelve times staler at exactly the moment the header started reporting `ws: connected`.** The frame looked its most alive as half of it stopped moving.
- This splits the single timer into two independent loops, one per refresh class, so the stream's health only affects the data the stream can actually replace.
- **Blast radius:** `pkg/tui` only. No cadence *values* changed (still 5s and 60s), no client or pane code touched, no visual change, no new flags or endpoints.

## The two classes

| loop | reads | cadence |
|---|---|---|
| **reconciliation** | `/api/agents`, `/api/status`, `/api/config/governor`, `/api/hive-id` | 5s while the stream is down, 60s while it is healthy — unchanged from T13b |
| **activity** | `/api/tokens`, `/api/cost`, `/api/audit` | 5s, **always** |

Each has its own tick message, interval field, and generation counter.

The class boundary is about **how fast the data moves**, not which transport feeds it. `/api/config/governor` and `/api/hive-id` are in the reconcile class even though no stream event carries them either: a hive whose name or evaluation cadence changed a minute ago is not a stale frame, whereas a minute-old token count on a fleet burning tokens continuously is.

## The separate generation counter is load-bearing

This is the part worth reviewing closely. The drop path bumps the reconcile generation to retire the stretched 60s chain, then re-arms **a reconcile chain only**. Had both classes stamped their ticks with one shared counter, that same bump would retire the activity tick already in flight — and nothing would ever arm another.

The Tokens and Events panes would go permanently dark at the first stream blip: the exact failure this task exists to prevent, arrived at from the other direction. `TestStreamDropDoesNotRetireTheActivityChain` pins it by feeding the *pre-drop* generation to the *post-drop* model and asserting it still fetches and still re-arms.

Relatedly, the drop's immediate fallback fetch is now `pollReconcile()` rather than the full `poll()`. The activity loop never stretched, so it has been reading those three endpoints every 5s throughout the healthy stream and is mid-interval at the moment of the drop; a full poll would issue a duplicate copy of each, on the exact code path a flapping dashboard walks repeatedly.

`poll()` itself still issues all seven reads and remains the **one-shot** full refresh — nothing in it arms a tick. That is what `Init` uses so neither loop waits out an interval before its first data, and what every action handler uses after a write. A pause, model apply, or ACMM apply moves the roster *and* appends the operator's own action to the audit log, so a `poll()` that had quietly become reconcile-only would show the effect of an action while the Events pane failed to record it.

## Naming

`tickMsg`/`interval`/`tickGen` are now `reconcileTickMsg`/`reconcileInterval`/`reconcileGen`, with `activityTickMsg`/`activityInterval`/`activityGen` alongside. With two of each, a bare `m.interval` no longer says which loop it belongs to. The rename is mechanical and no existing assertion was weakened — every SSE fallback/backoff test kept its substance.

## Mutation evidence

Every new invariant was proven capable of failing. Each defect was introduced by a patcher that aborts unless its pattern matches exactly once (the repo has been bitten before by a heredoc silently no-op'ing a mutation and reporting a false PASS), the suite run, then reverted.

| # | Defect introduced | Result |
|---|---|---|
| M1 | SSE event stretches `activityInterval` too | **CAUGHT** — `TestHealthyStreamStretchesOnlyTheReconcileLoop` |
| M2 | Drop bumps a shared generation (`activityGen++` as well) | **CAUGHT** — `TestStreamDropDoesNotRetireTheActivityChain`, `TestConnectDropCyclesLeaveOneChainOfEachClass` |
| M3 | Drop's fallback fetch is the full `poll()` | **CAUGHT** — `TestStreamDropDoesNotDuplicateActivityRequests` |
| M4 | `pollReconcile` also reads `/api/audit` | **CAUGHT** — `TestEachTickFetchesOnlyItsOwnClass/reconcile`, `TestPollStillIssuesBothClasses`, `TestPollIssuesEveryRead` |
| M5 | Activity tick fetches without re-arming | **CAUGHT** — `TestActivityTickRearmsWhenTheFetchFails`, `TestActivityLoopKeepsRefreshingWhileTheStreamIsHealthy`, `TestEachTickFetchesOnlyItsOwnClass/activity` |
| M6 | `Init` forgets to arm the activity chain | **CAUGHT** — `TestInitPollsImmediatelyAndArmsTick` |
| M7 | `scheduleActivityTick` stamps `reconcileGen` | **CAUGHT** — `TestScheduleTickStampsTheCurrentGeneration/activity` |
| M8 | `poll()` quietly becomes reconcile-only | **CAUGHT** — `TestInitFetchesBothClassesImmediately` + 9 existing T30/T31 tests |

The narrowness matters as much as the catches: M1 fails **only** the stretch test and M6 **only** the Init-arming test, so those are independently sensitive rather than one assertion double-counted. M8's blast radius is the reassuring direction — the existing T30/T31 suite already refuses to let the one-shot refresh lose its activity half.

Two notes on the tests themselves:

- `TestEachTickFetchesOnlyItsOwnClass` asserts **absence** in both directions. A "contains what it should" test alone passes on a model that never split anything — one loop fetching everything satisfies every such clause. What distinguishes a real split is that the reconcile tick does not read `/api/audit` and the activity tick does not read `/api/agents`.
- `TestStreamDropDoesNotDuplicateActivityRequests` waits for the `sseReconnectMsg` before reading the audit request counter. That message is a real `sseBackoffMin` timer, so by the time it lands a duplicate fetch against a local test server has had a full second to complete and be counted — the counter assertion is not racing the thing it is trying to observe.

No test sleeps a production interval: both intervals are model fields, every tick is injected, and the one place a 60s chain is unavoidable (`streamHealthyModel`) is drained through the existing concurrent `drainUntil` helper.

## Verified

```
cd src
go build ./...          # clean across the module
go test ./pkg/tui/...   # all four packages green
gofmt -l pkg/tui        # clean
go vet ./pkg/tui/...    # clean
```

CI is the gate; the above is local iteration only.

## Out of scope

The 5s and 60s policy values, a second SSE endpoint or server-side streaming of audit/token data, and any client, pane, or visual change. Documentation of the resulting cadence contract belongs to T26 (#5422), which is held on this PR.

### DIFF
--- a/src/pkg/tui/app.go
+++ b/src/pkg/tui/app.go
@@ -31,6 +31,15 @@
 // fallback and the reconciler: healthy stream, 60s reconcile; dropped stream,
 // back to the 5s poll and a header that says so. See the SSE section at the
 // bottom of this file.
+//
+// T32 (#5421): there are TWO poll loops, because the stretch above is only
+// correct for the data the stream carries. Once T30 and T31 hung /api/tokens,
+// /api/cost and /api/audit off the same timer, a healthy stream stretched them
+// to 60s as well — so the Tokens and Events panes became twelve times staler
+// at exactly the moment the header started saying `ws: connected`. The
+// reconciliation loop keeps the 5s/60s behaviour above; the activity loop runs
+// at 5s unconditionally, with its own message type and generation. See poll.go
+// for the split and which reads belong to which class.
 package tui
 
 import (
@@ -313,20 +322,42 @@ type model struct {
 	// the operator sees. Kick success describes only queueing or deduplication.
 	footerStatus string
 
-	// interval is this model's poll cadence, defaulting to pollInterval.
+	// reconcileInterval is the cadence of the reconciliation loop
+	// (pollReconcile), defaulting to pollInterval.
 	//
 	// It is a field rather than a bare constant read for two reasons. Tests
 	// drive a whole tick — fetch, delivery, and the re-arm — without waiting
 	// five real seconds for it. And T13b needs exactly this knob: once the SSE
-	// stream is connected the poll becomes a fallback and should slow down,
-	// not keep hammering an endpoint the stream has already superseded.
-	interval time.Duration
+	// stream is connected this poll becomes a fallback and should slow down,
+	// not keep hammering endpoints the stream has already superseded.
+	reconcileInterval time.Duration
 
-	// tickGen retires superseded poll chains. Changing interval only affects
-	// the NEXT re-arm, so switching cadence means arming a fresh chain while
-	// the old one is still pending; bumping this makes the pending tick a
-	// no-op instead of a second, permanent loop. See tickMsg in poll.go.
-	tickGen uint64
+	// reconcileGen retires superseded reconciliation chains. Changing the
+	// interval only affects the NEXT re-arm, so switching cadence means arming
+	// a fresh chain while the old one is still pending; bumping this makes the
+	// pending tick a no-op instead of a second, permanent loop. See
+	// reconcileTickMsg in poll.go.
+	reconcileGen uint64
+
+	// activityInterval is the cadence of the activity loop (pollActivity). It
+	// is pollInterval and STAYS pollInterval for the life of the process: no
+	// stream event can refresh a token count or an audit row, so there is
+	// nothing for this cadence to respond to.
+	//
+	// It is a field anyway, for the first of the two reasons above only —
+	// tests must be able to run a whole activity tick without sleeping five
+	// real seconds. Nothing in production writes it.
+	activityInterval time.Duration
+
+	// activityGen is the activity loop's own retirement counter.
+	//
+	// It exists so the two chains have independent lifetimes. Sharing
+	// reconcileGen would make the bump that retires a stretched reconcile
+	// chain on an SSE drop also retire the pending activity tick — and the
+	// drop path re-arms only the reconcile chain, so the Tokens and Events
+	// panes would go permanently dark at the first stream blip. See
+	// activityTickMsg in poll.go.
+	activityGen uint64
 
 	// agents is the last fleet roster a poll returned.
 	//
@@ -419,8 +450,9 @@ func newModel() model {
 			panes.NewTokens(),
 			panes.NewEvents(),
 		},
-		api:      client.New(),
-		interval: pollInterval,
+		api:               client.New(),
+		reconcileInterval: pollInterval,
+		activityInterval:  pollInterval,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


clubanderson (approved):
Reviewed. **Approve.** This closes the gap #5502 flagged when T30 landed: with SSE healthy the reconcile poll stretches to 60s, and the stream carries no token or audit data, so three panes were getting twelve times staler exactly when the connection was healthiest. Paradoxical from the operator's side, and this is the correct fix.

## Verified against the diff

**The generations are genuinely separate counters** — `reconcileGen` and `activityGen` as distinct fields, each with its own `if msg.gen != m.…Gen` guard. That separation is the whole safety property: a single shared counter would mean the bump retiring a stretched reconcile chain on an SSE drop also retires the activity chain, and re-arming without retiring the old one leaves two live chains ticking forever. The comments say this explicitly, which is worth keeping — it is the kind of invariant that gets "simplified" away by someone consolidating two counters that look redundant.

**The test set covers the failure modes rather than the happy path:**

- `TestHealthyStreamStretchesOnlyTheReconcileLoop` — the actual bug
- `TestActivityLoopKeepsRefreshingWhileTheStreamIsHealthy` — its converse
- `TestStreamDropDoesNotRetireTheActivityChain` and `TestStreamDropDoesNotDuplicateActivityRequests` — the two ways cross-contamination between counters shows up, one dropping a chain and one duplicating it
- `TestConnectDropCyclesLeaveOneChainOfEachClass` — the accumulation case, which is what actually bites in a long session with a flapping connection
- `TestInitFetchesBothClassesImmediately` — neither loop waits a full interval for first data
- `TestActivityTickRearmsWhenTheFetchFails` — a failing endpoint does not silently kill its own loop

That last one is easy to omit and expensive to miss: a loop that stops re-arming on failure looks identical to a healthy idle loop.

## One thing I like about the framing

The comment that a tick "carries no time — it is a *go fetch now* signal, not a clock, and nothing reads when it fired" is the right model to write down. Ticks that carry timestamps invite someone to compute elapsed time from them, which is where these loops usually acquire drift bugs.

## Not blocking

The 5s/60s values stay as they are, which matches the issue's scope. Worth noting for later that the activity loop is now unconditionally 5s against three endpoints regardless of stream health — correct, since nothing else refreshes those panes, but it is the piece to revisit if per-request cost ever becomes the constraint rather than staleness.

PR #5552, opened 2026-09-01
🐛 fix: agy sign-in cannot persist; a claude login must not restart an agy agent

### LINKED ISSUE #5551: 🐛 bug: agy sign-in can never persist — /data/home/.gemini is group read-only and in no perm guard
## Summary

An `agy` (antigravity) agent can never stay signed in. The operator completes the Google OAuth flow, the CLI reports success — and the session is gone the moment the process ends. Every hive restart requires a fresh login, and `agy models` in any new process reports "Please sign in" while the running agent looks authenticated.

The cause is a filesystem permission, and nothing in the failure names one. On a live hive:

```
2775 dev:node  /data/home/.claude    ← group write
2770 dev:node  /data/home/.copilot   ← group write
2750 dev:node  /data/home/.gemini    ← group READ-ONLY
```

`/data/home/.gemini` appears in **none** of the five places `entrypoint.sh` manages CLI credential directories: the `mkdir -p`, the explicit `chmod`/`chown`, the `NEED_PERM_FIX` sample set, the inotify perm guard, and the polling slow-cycle repair. So agy creates it itself, at whatever mode it likes, and it is never repaired.

Verified directly — as the agent uid:

```
$ touch /data/home/.gemini/.writetest
touch: cannot touch '/data/home/.gemini/.writetest': Permission denied
```

agy persists its session to `$HOME/.gemini/antigravity-cli/antigravity-oauth-token`, which it therefore cannot write. After a single `chmod g+w` on the directory, the next sign-in produced that file immediately and survived a restart.

**Who notices:** anyone putting an agent on the `agy` backend. It presents as the login "failing" repeatedly with no error mentioning permissions.

**Reproduce:** set an agent's backend to `agy`, sign in through the dashboard, then restart the agent — or run `agy models` in a second process.

## A second-order effect worth fixing with it

agy writes the token `0600` owned by the agent that signed in:

```
600 hive-scanner:node  /data/home/.gemini/antigravity-cli/antigravity-oauth-token
```

The shared CLI home exists so one login serves the fleet — that is why the guard re-opens copilot's `config.json` (rewritten `0600` on every refresh) and the whole `.claude` tree. Without `.gemini` in that guard, every agy agent needs its own separate login, and any agent that re-authenticates locks the others out again.

## Related, found while diagnosing this

The token-restart heal made this much harder to reach. `configHasTokens()` is backend-blind — it answers true when either the shared claude **or** copilot credential is usable, whatever backend the agent actually runs. So an agy agent parked at its Google OAuth prompt was restarted on the strength of an unrelated claude login. A restart cannot mint a Google session, so it looped, and each relaunch minted a fresh PKCE `code_challenge` — invalidating the code the operator was in the middle of pasting.

Measured on the same hive: **15 restarts in one hour and 8 distinct challenges**, which turns "sign-in is fiddly" into "sign-in is impossible". The operator sees `oauth2: "invalid_grant" "Malformed auth code."` and reasonably concludes the code was mangled in transit; it was not, the challenge simply no longer existed.

This surfaced only after #5494 made agy's login prompt visible to the detector — before that agy was invisible and idle, so the heal never fired on it.

## Not a bug (recorded so it is not chased)

`--model gemini-3.7-flash-high --effort low` logs

```
⚠ --model gemini-3.7-flash-high conflicts with --effort=low. Using "Gemini 3.7 Flash (High)" instead.
```

That is a warning, and agy falls back to the model the operator configured. Worth tidying so the warning stops, but it breaks nothing. `agy --help` does advertise `--effort (low|medium|high)`; it is the effort-suffixed model IDs that reject it.

Danathar:
## Summary

An `agy` (antigravity) agent could never stay signed in, and while an operator tried, hive restart-looped it out of the flow. Two defects that compounded into "the login just fails", with nothing anywhere naming a cause.

- **The credential could not be written.** `/data/home/.gemini` is group **read-only** and appears in none of the five places `entrypoint.sh` manages CLI credential dirs. agy persists its session there, so every sign-in succeeded in-process and evaporated on exit.
- **A claude login restarted an agy agent.** `configHasTokens()` is backend-blind, so an agy agent stuck at its Google OAuth prompt was restarted on the strength of an unrelated credential — rotating the PKCE challenge the operator was mid-way through using.

**Blast radius:** the entrypoint change adds `.gemini` beside `.claude`/`.codex`/`.copilot`; no existing dir's handling changes. The heal gate becomes *narrower* — it now requires evidence for the agent's own backend, and the case it was built for is unchanged and pinned.

Fixes #5551

## 1. `.gemini` is group read-only and in no perm guard

Measured on a live hive:

```
2775 dev:node  /data/home/.claude    ← group write
2770 dev:node  /data/home/.copilot   ← group write
2750 dev:node  /data/home/.gemini    ← group READ-ONLY
```

```
$ touch /data/home/.gemini/.writetest        # as the agent uid
touch: cannot touch '.writetest': Permission denied
```

agy writes `$HOME/.gemini/antigravity-cli/antigravity-oauth-token`. It could not, so `agy models` in any new process reported "Please sign in" while the running agent looked authenticated, and every restart demanded a fresh login. **One `chmod g+w` produced the token file on the next sign-in, and it survived a restart.**

`.gemini` is added to all five sites — `mkdir`, `chmod`/`chown`, the `NEED_PERM_FIX` sample, the inotify guard, and the polling slow-cycle. `2770` rather than `.codex`'s `2775`: it holds an OAuth token, so it follows `.copilot`/`.bob` in keeping world off it.

**The ongoing guard matters as much as the create-time mode.** agy writes the token `0600` owned by whichever agent signed in:

```
600 hive-scanner:node  …/antigravity-oauth-token
```

The shared CLI home exists so one login serves the fleet — that is why the guard already re-opens copilot's `config.json` (rewritten `0600` on every refresh) and the whole `.claude` tree. Without `.gemini` in it, every agy agent needs a separate login and each re-auth locks the others out.

## 2. The token-restart heal was backend-blind

`configHasTokens()` answers true when **either** the shared claude or copilot credential is usable, whatever backend the agent actually runs. So an agy agent parked at its Google OAuth prompt was restarted because an unrelated *claude* login was valid. A restart cannot mint a Google session, so it looped — and each relaunch minted a fresh PKCE `code_challenge`, invalidating the code the operator was pasting.

Measured on the same hive: **15 restarts in one hour, 8 distinct challenges.** The operator saw

```
oauth2: "invalid_grant" "Malformed auth code."
```

and reasonably concluded the code had been mangled in transit. It had not — the challenge no longer existed. That misdirection cost more diagnosis time than either bug.

The gate is now `AgentHasValidCredential`, which resolves the agent's **own** backend and is positive-evidence-only, so a backend this process cannot verify (agy, gemini) answers false and is left alone. That is the correct answer: *do not restart when you cannot show a restart would help.* The heal's documented case — a claude agent with a usable shared credential, #4596 — is unchanged.

This only became reachable when #5494 taught the detector to see agy's login prompt. Before that agy was invisible and idle, so the heal never fired on it — the fix that made the outage visible is what exposed this.

## Verification

| Property | Test |
|---|---|
| every interactive-auth credential dir is created group-writable + setgid | `test_entrypoint_data_ownership.sh` (new section) |
| …and is repaired by the ongoing guard after the CLI rewrites it `0600` | same |
| an agy agent does not read as credentialed off a claude login | `TestTokenRestartHealIsBackendScoped` |
| a claude agent with a usable shared credential still does | same |

The entrypoint guard is mutation-checked both ways — removing the `chmod 2770` fails the create-time half, removing the inotify watcher **and** the polling entry fails the ongoing half. It is deliberately two assertions rather than one: a dir that is `mkdir`'d and chowned but absent from the ongoing guard silently re-locks after the CLI's first credential refresh, which is exactly the shape of this bug.

The new test asserts `configHasTokens() == true` as an explicit **precondition** before checking the scoped answer — without it the test would pass for the wrong reason if the shared credential fixture ever stopped being usable.

## Testing

- [x] `cd src && go test ./pkg/agent/ ./pkg/dashboard/`
- [x] `go vet ./pkg/agent/...`, `bash -n src/deploy/entrypoint.sh`, `git diff --check`
- [x] `bash src/deploy/test_entrypoint_data_ownership.sh` → 21 passed, 0 failed
- [x] Verified live: after the permission fix a sign-in wrote `antigravity-oauth-token`, survived a hive-driven restart, and the agent came back `state=running needsLogin=False` with zero restarts.
- [ ] Not run: no automated coverage that an agy sign-in *persists* end to end — that needs a real Google OAuth flow. The permission precondition it depends on is now pinned instead.

## Contributor checklist

- [x] PR targets `v4`.
- [x] Docs, examples, and policies updated where behavior changes — the entrypoint comments record the measurement and why `.gemini` is `2770`.
- [ ] `CHANGELOG.md` — not added; happy to add one if you want this in the operator-facing log.

### DIFF
--- a/src/deploy/entrypoint.sh
+++ b/src/deploy/entrypoint.sh
@@ -814,7 +814,7 @@ if [ "$(id -u)" = "0" ]; then
   # Shared CLI auth/cache lives in /data/home (persistent volume).
   # Make it group-writable so all agent UIDs (node group) can use it.
   # The manager sets HOME=/data/home for agent tmux sessions.
-  mkdir -p /data/home/.config /data/home/.copilot /data/home/.claude/session-env /data/home/.codex /data/home/.bob/settings /data/config/github-copilot /home/dev/.config
+  mkdir -p /data/home/.config /data/home/.copilot /data/home/.claude/session-env /data/home/.codex /data/home/.gemini /data/home/.bob/settings /data/config/github-copilot /home/dev/.config
   # $HOME itself must be group-writable, not just its children. bob calls
   # mkdirSync('$HOME/.bob') on first run, which needs write on /data/home — a
   # 0755 root-owned $HOME makes that EACCES even though every child dir below
@@ -838,6 +838,22 @@ if [ "$(id -u)" = "0" ]; then
   # codex backend fails with "Permission denied" initializing state_N.sqlite.
   chmod 2775 /data/home/.codex 2>/dev/null || true
   chown -R dev:node /data/home/.codex 2>/dev/null || true
+  # Antigravity (`agy`) persists its OAuth session to
+  # $HOME/.gemini/antigravity-cli/antigravity-oauth-token. Pre-create the tree
+  # group-writable + setgid for the same reason as .codex above.
+  #
+  # This was measured, not assumed. On a live hive .gemini existed at 2750 —
+  # group READ-only, because nothing here had ever created or repaired it and
+  # agy made it itself. The agent UID could not write the directory, so every
+  # sign-in succeeded in-process and then evaporated: `agy models` in any new
+  # process reported "Please sign in", and each hive restart lost the session
+  # again. The operator saw four consecutive logins "fail" with no error that
+  # named a permission. One chmod produced the token file immediately.
+  #
+  # 2770 rather than .codex's 2775: this directory holds an OAuth token, so it
+  # follows .copilot/.bob in keeping world off it.
+  chmod 2770 /data/home/.gemini 2>/dev/null || true
+  chown -R dev:node /data/home/.gemini 2>/dev/null || true
   # bob writes installation_id, settings.json, trustedFolders.json and tmp/ under
   # $HOME/.bob, plus custom modes under $HOME/.bob/settings. Pre-create both
   # group-writable + setgid so the FIRST agent to launch (a 2001+ UID in group
@@ -859,7 +875,7 @@ if [ "$(id -u)" = "0" ]; then
   # too so that gap cannot reopen.
   NEED_PERM_FIX=false
   if [ -d "/data/home/.copilot" ] && [ -d "/data/home/.claude" ]; then
-    for perm_dir in /data/home /data/home/.copilot /data/home/.claude /data/home/.bob; do
+    for perm_dir in /data/home /data/home/.copilot /data/home/.claude /data/home/.gemini /data/home/.bob; do
       # Missing dir → repair. Default 755 → repair (no group-write/setgid).
       DIR_PERMS=$(stat -c '%a' "$perm_dir" 2>/dev/null || echo "755")
       case "$DIR_PERMS" in
@@ -920,7 +936,19 @@ if [ "$(id -u)" = "0" ]; then
         chown -R dev:node /data/home/.codex 2>/dev/null
       done
     ) &
-    echo "[entrypoint] inotify perm guard active (copilot + claude + codex)"
+    (
+      # agy writes antigravity-oauth-token 0600 owned by the agent that signed
+      # in, which locks every other agent UID out of a credential the shared
+      # CLI home exists to share — the same shape as copilot's config.json
+      # above. Re-opening it to the node group is what makes ONE agy login
+      # serve the fleet instead of one login per agent.
+      while inotifywait -qq -e close_write,moved_to,create /data/home/.gemini/ 2>/dev/null; do
+        chmod -R g+rwX /data/home/.gemini 2>/dev/null
+        find /data/home/.gemini -type d -exec chmod g+s {} + 2>/dev/null
+        chown -R dev:node /data/home/.gemini 2>/dev/null
+      done
+    ) &
+    echo "[entrypoint] inotify perm guard active (copilot + claude + codex + gemini)"
   fi
   (
     CYCLE=0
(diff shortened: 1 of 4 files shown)


clubanderson (approved):
Reviewed. **Approve.**

The diagnosis is measured on a live hive rather than inferred, which is what makes this trustworthy: `2750 dev:node /data/home/.gemini` against `2775` for `.claude` and `2770` for `.copilot`, plus a `touch` as the agent uid returning `Permission denied`, plus the confirmation that one `chmod g+w` produced the token file and it survived a restart. That is cause established, not suspected.

## Verified against the diff

**`.gemini` lands in all five sites** — `mkdir -p`, `chmod`/`chown`, the `NEED_PERM_FIX` sample, the inotify guard, and the polling slow-cycle. Missing any one of the last two would have made this work until the first token rewrite and then silently regress.

**`2770`, not `.codex`'s `2775`.** Correct: it holds an OAuth token, so it follows `.copilot`/`.bob` in keeping world off it. Matching `.codex` would have been the easy, wrong consistency.

**The ongoing guard is the load-bearing half.** agy writes the token `0600` owned by whichever agent signed in, so create-time mode alone would not survive the first real sign-in. The inotify + slow-cycle coverage is what makes this durable rather than a one-shot fix.

**The heal gate is genuinely narrower.** `configHasTokens()` being backend-blind is a real bug in its own right — an agy agent parked at a Google OAuth prompt restarted on the strength of an unrelated claude credential, rotating the PKCE challenge the operator was actively using. The replacement is positive-evidence-only and scoped to the agent's own backend, so a backend whose credential this process cannot verify does not trigger a restart. That is the safe direction: a missed heal costs a stuck agent, a spurious one costs the operator's in-flight login.

Worth stating plainly since "we changed the restart gate" sounds alarming: this **removes** restart triggers, it does not add any. The case the gate was built for is unchanged and pinned by test.

## The compounding is the interesting part

Neither defect alone reads as "login is broken." The credential silently failed to persist, so the agent looked authenticated in-process and demanded a fresh login on every restart — while the restart gate was simultaneously cutting the operator off mid-flow on unrelated evidence. Two separate causes presenting as one vague symptom, with nothing anywhere naming either. That is why it needed measurement on a live hive to untangle.

## Not blocking

`chown -R dev:node /data/home/.gemini` in the slow cycle will reclaim ownership of a token an agent uid wrote. Group is `node` and the mode is `2770`, so the signing agent retains access and this is correct — but it is the same shape as the ownership walk #5534 just moved off the startup path. `.gemini` is small today; if it ever grows a cache, that walk is on the same PVC.

PR #5560, opened 2026-09-01
🐛 fix(ci): backend-smoke.yml rejected by workflow parser — matrix context in job-level if

### LINKED ISSUE #5550: [ci-maintainer] backend-smoke.yml invalid: matrix context in job-level if fails every push with 0s 'workflow file issue' run
## CI Issue

`.github/workflows/backend-smoke.yml` (merged ~2026-09-01 14:03 UTC, commits 3890c4d/ae92788) is rejected by GitHub's workflow parser. Since then **every push to every branch** (including v4) generates a 0s failed run named `.github/workflows/backend-smoke.yml` — 15+ failures in the last ~30 min — and the scheduled smoke lanes (cron 45 2,8,14,20) will never run.

## Evidence

- Runs 33540275463, 33540275048, 33540100983, 33538381599 (v4 merge push), all: `failure`, 0s, "This run likely failed because of a workflow file issue."
- actionlint on the file:
  ```
  backend-smoke.yml:74:50: context "matrix" is not allowed here. available contexts are "github", "inputs", "needs", "vars"
  ```
- Offending job-level `if` on the `smoke` job:
  ```yaml
  if: >-
    github.repository == 'kubestellar/hive' &&
    (github.event_name != 'workflow_dispatch' ||
     inputs.lane == 'both' || inputs.lane == matrix.lane)
  ```
  The `matrix` context is not available in `jobs.<id>.if`.

## Recommendation

Drop the `matrix.lane` term from the job-level `if` and enforce the lane filter inside the existing `scope` step (which already gates by backend), where `matrix` is available. Hold-gated PR to follow.

---
---

clubanderson:
## Problem

`.github/workflows/backend-smoke.yml` gated its `smoke` job with a job-level `if` that referenced `matrix.lane`:

```yaml
if: >-
  github.repository == 'kubestellar/hive' &&
  (github.event_name != 'workflow_dispatch' ||
   inputs.lane == 'both' || inputs.lane == matrix.lane)
```

`jobs.<id>.if` is evaluated **before** the matrix expands, so the `matrix` context does not exist at that point. GitHub rejects the entire workflow file. Consequences since it merged (~2026-09-01 14:03 UTC, 3890c4d/ae92788):

- every push to every branch produces a 0-second failed run named `.github/workflows/backend-smoke.yml`
- the scheduled smoke lanes (`cron: 45 2,8,14,20`) have **never run**

## Fix chosen, and why

Two options were on the table. I took the **matrix-computation** one.

The lane filter now lives in `strategy.matrix.lane`, which selects the lane list up front via `fromJSON`. The `inputs` context **is** in scope inside `strategy`, so a dispatch naming a single lane expands the matrix to just that lane:

```yaml
lane: >-
  ${{ (github.event_name == 'workflow_dispatch' && inputs.lane != 'both')
      && fromJSON(format('["{0}"]', inputs.lane))
      || fromJSON('["latest","pinned"]') }}
```

The alternative — moving the `inputs.lane` term to a step-level `if` — also parses, since `matrix` is available at step level. I rejected it because the whole point of the condition is to *avoid running* a lane the operator did not ask for. A step-level guard still spins up a runner for every unwanted arm and then no-ops, reporting green-but-empty jobs. Computing the matrix means the unwanted lane never schedules at all, which is what the original author was reaching for.

The `github.repository == 'kubestellar/hive'` fork guard **stays** on the job-level `if`, where it is perfectly valid — forks have no model credentials and would go permanently red without it.

Injection note: `inputs.lane` is a `choice` constrained to `both|latest|pinned`, so the `format()` interpolation cannot introduce arbitrary JSON.

## Verification — actionlint

**Before** (on `origin/v4`):

```
$ actionlint .github/workflows/backend-smoke.yml
.github/workflows/backend-smoke.yml:74:50: context "matrix" is not allowed here. available contexts are "github", "inputs", "needs", "vars". see https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability for more details [expression]
EXIT=1
```

**After** (this branch):

```
$ actionlint .github/workflows/backend-smoke.yml
EXIT=0
```

A full-repo `actionlint` run also exits 0; the only remaining output is pre-existing shellcheck `SC2016` *info* in the unrelated `podman-arm64-lane.yml`.

Since a file can parse locally and still be rejected by GitHub, registration on the pushed branch was checked separately — see the comment below.

## Not verified

- **The scheduled cron lanes actually work.** Cron cannot be triggered on demand, and this workflow is `schedule` + `workflow_dispatch` only, so pushing does not exercise the smoke itself. This PR proves the file is *accepted*; the first real evidence the lanes run will be the next scheduled tick.
- **The smoke suite passing.** No credentials are exercised here and no lane was dispatched; this change is confined to workflow scheduling semantics.
- `bin/test_backend_smoke.sh` is untouched.

Fixes #5550

### DIFF
--- a/.github/workflows/backend-smoke.yml
+++ b/.github/workflows/backend-smoke.yml
@@ -70,17 +70,22 @@ jobs:
   smoke:
     # Forks have no model credentials; a scheduled run there would only
     # produce a permanently red workflow.
-    if: >-
-      github.repository == 'kubestellar/hive' &&
-      (github.event_name != 'workflow_dispatch' ||
-       inputs.lane == 'both' || inputs.lane == matrix.lane)
+    if: github.repository == 'kubestellar/hive'
     runs-on: ubuntu-latest
     timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:
         backend: [claude, codex]
-        lane: [latest, pinned]
+        # The dispatch lane filter lives HERE, not in the job-level `if`: that
+        # `if` is evaluated before the matrix expands, so `matrix` is not in
+        # scope there and GitHub rejects the whole file (#5550). Selecting the
+        # lane list up front also means an unwanted lane never schedules a
+        # runner at all, rather than starting one that immediately no-ops.
+        lane: >-
+          ${{ (github.event_name == 'workflow_dispatch' && inputs.lane != 'both')
+              && fromJSON(format('["{0}"]', inputs.lane))
+              || fromJSON('["latest","pinned"]') }}
     env:
       # Per-arm credential scoping: each backend arm gets only its own
       # vendor's credential, in whichever form the repo provisioned (see the


clubanderson:
## Registration evidence — GitHub actually accepts the file now

A workflow can parse locally and still be rejected by GitHub, so this was checked against the live API rather than inferred from actionlint.

**`v4` (unfixed) still emits the broken-file run:**

```
$ gh run list --repo kubestellar/hive --branch v4 \
    --json name,conclusion,createdAt,databaseId \
    -q '.[] | select(.name | test("backend-smoke"))'
failure   .github/workflows/backend-smoke.yml   2026-09-01T18:58:33Z   33546786923
```

Run `33546786923` detail — note `run_started_at` and `updated_at` are the **same second**, the signature of a parse rejection rather than a real execution:

```
conclusion=failure  started=2026-09-01T18:58:33Z  updated=2026-09-01T18:58:33Z
```

**This branch emits none.** Same query, same push event, zero rows:

```
$ gh run list --repo kubestellar/hive --branch fix/5550-backend-smoke-matrix-if \
    --json name,conclusion \
    -q '.[] | select(.name | test("backend-smoke"))'
(no output)
```

The push registered 15 normal workflow runs (Podman lanes, Go Security Analysis, PR Verifier, Release Line Guard, …) and **no** run named after the raw workflow path. GitHub is parsing `backend-smoke.yml` successfully and correctly declining to start it, because this workflow is `schedule` + `workflow_dispatch` only and a `push` matches neither trigger.

That absence is the fix: a rejected file produces a failed run on *every* event, which is precisely what #5550 reported.

### Still not verified

The **scheduled cron lanes actually executing**. `cron` cannot be triggered on demand and `push` does not match this workflow's triggers, so nothing here exercises `bin/test_backend_smoke.sh` or the credentials. This PR establishes that the file is accepted and that the lane matrix is well-formed; the first genuine evidence of the lanes running will be the next scheduled tick at `45 2,8,14,20`.

clubanderson:
## Definitive proof: GitHub's own parser, both branches, same API call

`workflow_dispatch` via the REST API is an exact parser test — GitHub must parse the workflow at the given `ref` before it will accept the dispatch. Same workflow id (`347596454`), same inputs, only the `ref` differs.

**Against `v4` (unfixed) — HTTP 422, rejected:**

```
$ gh api -X POST repos/kubestellar/hive/actions/workflows/347596454/dispatches \
    -f ref=v4 -f 'inputs[lane]=latest' -f 'inputs[backends]=claude'

HTTP 422
Invalid Argument - failed to parse workflow: (Line: 73, Col: 9):
Unrecognized named-value: 'matrix'. Located at position 130 within expression:
github.repository == 'kubestellar/hive' && (github.event_name != 'workflow_dispatch' ||
 inputs.lane == 'both' || inputs.lane == matrix.lane)
```

GitHub names the exact defect, independently of actionlint.

**Against this branch — HTTP 204, accepted:**

```
$ gh api -X POST repos/kubestellar/hive/actions/workflows/347596454/dispatches \
    -f ref=fix/5550-backend-smoke-matrix-if -f 'inputs[lane]=latest' -f 'inputs[backends]=claude'

HTTP/2.0 204 No Content
```

**The dispatched run ([33549274632](https://github.com/kubestellar/hive/actions/runs/33549274632)) is real, and the lane filter works:**

```
name=Backend smoke   event=workflow_dispatch   status=queued

=== expanded matrix arms ===
queued   smoke (codex, latest)
queued   smoke (claude, latest)
```

Three things this establishes at once:

1. **`name=Backend smoke`** — GitHub is reading the `name:` key again. While the file was unparseable it fell back to the raw path, which is why the red runs were literally titled `.github/workflows/backend-smoke.yml`.
2. **The run queued and expanded jobs** rather than dying at 0 seconds.
3. **Exactly two arms, both `latest`.** The `pinned` lane did not schedule a runner at all — which is exactly what the original job-level `if` was reaching for, and what a step-level `if` would *not* have achieved (it would have started 4 arms and no-opped 2).

`backends: claude` still expands both backend arms by design; that axis is intersected inside the existing `scope` step, which is unchanged by this PR.

### Correction to the earlier comment

The absence of a `backend-smoke`-named run on a **push** is expected and correct — this workflow is `schedule` + `workflow_dispatch` only, so a push matches no trigger. The meaningful signal is the 422→204 flip above.

### Still not verified

The **cron schedule firing on its own**. `schedule` cannot be triggered on demand; the dispatch above exercises the identical job and matrix definition, but the first real evidence of the timer is the next tick at `45 2,8,14,20`. I also did not assess whether the smoke suite itself passes — that depends on repo credentials and vendor CLI state, and is out of scope for a parser fix.

clubanderson:
## The dispatched run finished — green, end to end

Run [33549274632](https://github.com/kubestellar/hive/actions/runs/33549274632) completed:

```
name=Backend smoke   event=workflow_dispatch   status=completed   conclusion=success
started=2026-09-01T19:24:02Z   updated=2026-09-01T19:34:16Z

success   smoke (codex, latest)
success   smoke (claude, latest)
skipped   create-issue-on-failure
```

This is stronger than the parse check alone. The workflow does not merely register — it **executes and passes**:

- ~10 minutes wall clock, versus the 0-second `run_started_at == updated_at` rejection still occurring on `v4`.
- Both `latest` arms ran `bin/test_backend_smoke.sh` against the real vendor CLIs and went green, so the suite and its credentials are healthy on this branch.
- Only the two `latest` arms ever existed — the `pinned` lane was never scheduled, confirming the `fromJSON` matrix filter does what the broken job-level `if` was reaching for.
- `create-issue-on-failure` correctly **skipped**: its gate is `failure() && github.event_name == 'schedule'`, and this was a successful dispatch. The issue-filing path is undisturbed by this change.

### Verified

| Claim | Evidence |
|---|---|
| actionlint clean | exit 0 (was exit 1, `matrix` context error) |
| GitHub parses the file | dispatch 422 on `v4` → 204 on this branch |
| Workflow registers properly | `name=Backend smoke`, not the raw path fallback |
| Job actually runs | 10-min run, `conclusion=success` |
| Lane filter narrows the matrix | `lane=latest` produced exactly 2 arms, no `pinned` |
| Issue-filing gate intact | `create-issue-on-failure` skipped as designed |

### Still not verified

The **cron timer firing on its own.** `schedule` cannot be triggered on demand. This dispatch exercised the identical job, matrix, steps, and credentials, so everything downstream of the trigger is now proven; the only untested element is the `45 2,8,14,20` timer itself, whose first real evidence will be the next tick.

Issue #5561, opened 2026-09-01
🔌 [v5] escalation reviewer lane is unreachable — buildReviewerWorkList has no caller


clubanderson:
`src/pkg/scheduler/reviewer_lane.go` exists on `v5` (merged via #5485, extended by #5515) but its work-list builder has **no caller outside its own file**.

```
$ git grep -l 'buildReviewerWorkList' origin/v5 -- 'src/**/*.go'
origin/v5:src/pkg/scheduler/reviewer_lane.go
```

`buildReviewerWorkList` and `buildReviewerMessage` are defined and tested, and nothing invokes them. The escalation-adjudication lane's Go half is present and unreachable.

## Why this matters beyond a dead function

This is the exact shape #5424 (tui T33) was written to catch, and which #5504 found in the snapshot guard: **"both halves merged" mistaken for "the feature is wired."** Package tests pass because the halves are individually correct. Nothing fails, because nothing calls it.

It is also invisible from the merge history — #5485 and #5515 are both real, well-tested, correct changes. The gap is in what did *not* land.

## Verified state on `v5`

Present and reachable:
- `escalation.Sweep` reviewer-verdict reconciliation (#5515) — reached via the sweep
- `ReviewerPassedLabel`, `NeedsHumanLabel` constants
- `GovernorConfig.ResolveReviewer()` / `ReviewerReady()`
- the `reviewer` agent in `level-5.yaml` with `reviewer-advisory.md` in both policy trees

Present and **not** reachable:
- `Scheduler.buildReviewerWorkList` (`reviewer_lane.go:170`)
- `Scheduler.buildReviewerMessage` (`reviewer_lane.go:95`)

Note the `reviewer` agent that IS wired on `v5` is a different capability — a repo-grounded PR reviewer at `mode: ADVISORY` that judges a proposed change and cites `file:line`. The escalation lane is post-fix-loop adjudication. They share the name `reviewer` and nothing else, which is part of why this went unnoticed.

## What to decide before fixing

Do not simply add a call site. The unresolved question from #5480 is whether escalated adjudication is:

1. a **separate agent** with its own name (`adjudicator`, `escalation-reviewer`), its own pack entry and policy template; or
2. an **authority tier** on the existing `reviewer` — L5 stays `ADVISORY`, L6 gains `ISSUES_AND_PRS` with a `reviewer-holdgated.md` template.

[...]

Issue #5529, opened 2026-09-01
🔧 no guard that every src/deploy/ test suite is wired into CI — test_entrypoint_symlinks.sh runs nowhere today


clubanderson:
The `#4363` guard asserts **"Every `bin/` test suite is wired into something"**. There is no equivalent for `src/deploy/`, and the gap is not hypothetical.

## Measured now

Of the **30** `src/deploy/test_*.sh` suites, five are absent from `v2-ci.yml`. Four of those run in a different workflow, which is fine:

| Suite | Runs in |
|---|---|
| `test_ambient_cap_runtime.sh` | `suid-contract.yml` |
| `test_image_suid_inventory.sh` | `docker.yml`, `suid-contract.yml` |
| `test_manifest_caps_runtime.sh` | `suid-contract.yml` |
| `test_quadlet_generator_gate.sh` | `quadlet-gate.yml` |
| **`test_entrypoint_symlinks.sh`** | **nowhere — no workflow references it** |

I checked every file under `.github/workflows/` for each name, so "nowhere" means nowhere, not "not in the obvious place."

## Why this keeps happening

`src/deploy/` suites are wired **individually**, by name. A new suite is therefore never picked up automatically, and an omission produces no failure — nothing fails when a test file is simply never named. That is the property `#4363` calls out for `bin/`: *"Each omission it catches is individually invisible."*

Two live instances already found this way:

- `test_entrypoint_system_gitconfig.sh` was wired into nothing and had never run in CI. Found incidentally during #5388 and registered there; it passes 10/0, so nothing was broken — but nothing was being checked either.
- `test_hive_snapshot_unit_contract.sh` merged in #5499 without registration and would never have run. Caught only because a duplicate PR (#5503) happened to include the registration; see #5504.

Both are the same failure as the guards #5388 catalogues: something that reads as coverage while asserting nothing. A test file present in the tree looks like a test that runs.

## Proposed

Extend the `#4363` guard, or add a sibling, asserting every `src/deploy/test_*.sh` is referenced by at least one workflow under `.github/workflows/`. Referenced *anywhere* is the right bar — four of the five above are correctly placed in specialised lanes, and requiring `v2-ci.yml` specifically would produce false failures.

## Acceptance

- [ ] The guard fails against the current tree, because `test_entrypoint_symlinks.sh` is genuinely unwired — **demonstrate that**, then wire the suite and show it pass. Per #5388's standard, a guard that has only ever been observed passing is not evidence.
[...]

Issue #5424, opened 2026-09-01
🧪 tui T33: end-to-end v1 acceptance coverage


Danathar:
Part of #4907. This is the final regression and closure gate. It depends on T17 (#5416), T19 (#5417), T21 (#5414), T26 (#5422), T27 (#5423), and the live-data/cadence chain through T32 (#5421).

## Why this exists

The tracker was decomposed intentionally into client, pane, and app tasks small enough for short contributor runs. That kept individual PRs reviewable, but it also allowed “both halves merged” to be mistaken for “the feature is wired”: Tokens and Events had complete clients and panes while the root app never delivered their messages. A final test must exercise the assembled operator loop rather than trusting isolated package tests or checklist state.

## Files to create or modify

- `src/pkg/tui/integration_test.go` (new)
- `src/pkg/tui/testdata/` fixtures only where shared deterministic payloads are needed
- existing TUI tests only to extract a reusable fixture helper without weakening assertions

Production behavior changes are out of scope. If the integrated test exposes a product bug, stop and file a narrowly scoped follow-up linked to #4907 rather than hiding a broad fix in the acceptance PR.

## Build this

Create a deterministic `httptest` dashboard that serves every endpoint the v1 TUI consumes, including a controllable SSE stream. Drive the real root model with teatest at a pinned supported terminal size. The test suite must prove the assembled behavior:

1. startup loads all four panes and live header fields without waiting production intervals;
2. an SSE status event updates agent/governor state immediately;
3. tokens/cost and audit continue refreshing at the fixed activity cadence while SSE is healthy;
4. stream drop changes connection state, activates poll fallback, preserves last-good data, and reconnects without duplicate loops;
5. focus and Agents/Events navigation target the displayed selection;
6. pause/resume confirmation, model selection/apply, `K` kick, and typed ACMM confirmation hit the correct escaped endpoint exactly once and render the authoritative response semantics;
7. help/footer list the same available bindings and modal keys cannot leak into global actions;
8. resize below/above the minimum swaps the warning/full frame cleanly;
9. `q`/`ctrl+c` quit and cancel the stream without a goroutine/reconnect leak.

Use channels, injected messages, and short test-only durations rather than real sleeps. Record requests in the fixture so assertions cover method, path, body, auth, and call count. Keep expected output focused on stable user-visible facts; pane-specific golden files already own exact layout bytes.

[...]

PR #5573, opened 2026-09-01
🧪 test(tui): end-to-end v1 acceptance coverage for the assembled operator loop

clubanderson:
## What this is

The #4907 epic was decomposed into client, pane and app tasks small enough for short contributor runs. That kept individual PRs reviewable, but it also let **"both halves merged" be mistaken for "the feature is wired"**: Tokens and Events had complete clients *and* complete panes while the root app never delivered their messages. Every isolated package test passed. The panes said `waiting for data` forever.

That exact shape has recurred three more times in this repo this week — #5504 (a 184-line security contract test merged and registered in no workflow), #5529 (`test_entrypoint_symlinks.sh` referenced by no workflow at all), #5561 (`buildReviewerWorkList` with no caller outside its own file). So the acceptance test has to prove the loop is **invoked**, not that its parts return correct values.

One deterministic `httptest` dashboard serves every endpoint the v1 TUI consumes including a controllable SSE stream, and drives the **real root model** with `teatest` at a pinned terminal size.

## What it asserts

- startup fills all four panes and both live header fields **without waiting a production interval**, and sends the bearer token on every request
- an SSE status event updates agent and governor state immediately, and does **not** blank the configured governor interval — that is the specific #5418 regression shape
- tokens/cost and audit keep refreshing **while the stream is healthy** — T32's property, and the one most likely to regress silently, since a healthy stream is exactly when nobody looks
- a stream drop activates poll fallback, preserves last-good data, and reconnects **exactly once**
- pause, resume, kick, model apply and typed ACMM confirmation each hit the correct escaped endpoint **exactly once** and render the authoritative response
- agent names are escaped into path segments
- help and footer advertise the same bindings, and modal keys cannot leak into global actions

Attach is asserted at its binding/selection path only — spawning tmux cannot run safely in CI, and the command construction already has its own tests. That boundary is called out in a test comment.

## Provenance and verification status — read this before merging

This is **salvaged work**. The agent that wrote it produced 2,039 lines and then died before committing or opening a PR. I recovered the file from its worktree, verified it compiles clean under `go vet ./pkg/tui/`, and committed it unchanged.

**I have not seen the full suite pass.** A local `go test ./pkg/tui/` exceeded my timeout and was still running when I opened this. So:

- I am **not** claiming this is green. CI is the gate and its result is the first real evidence.
- The per-property mutation evidence the task asked for — break each assertion in turn, confirm the specific test fails, revert — was **not performed**. Without it, an assertion here could be vacuous in exactly the way this PR exists to prevent, and I would rather say so than let the irony pass unremarked.

If CI is green, this is still worth landing on its own merits and the mutation pass can follow. If any assertion is red, treat it as unproven rather than as a product regression until someone has looked.

`Refs #5424` deliberately, not `Fixes` — the issue should close on a run that has been mutation-verified, not on this.

### DIFF
--- a/src/internal/testutil/eventually.go
+++ b/src/internal/testutil/eventually.go
@@ -45,6 +45,22 @@ const pollInterval = 10 * time.Millisecond
 //     strictly worse.
 func Eventually(t testing.TB, timeout time.Duration, cond func() bool, msg string, args ...any) {
 	t.Helper()
+	EventuallyEvery(t, timeout, pollInterval, cond, msg, args...)
+}
+
+// EventuallyEvery is Eventually with a caller-chosen sampling interval.
+//
+// The default 10ms suits a condition that becomes true after real work — a
+// server responding, a file appearing. It is far too coarse for an in-process
+// harness whose state changes in microseconds: sampling ten times slower than
+// the thing you are watching turns a fast suite into a slow one, and a suite
+// that takes minutes stops being run under -count=N, which is where genuine
+// flakes surface.
+//
+// Prefer plain Eventually. Reach for this only when you have measured that
+// the default interval dominates the wait, and say so at the call site.
+func EventuallyEvery(t testing.TB, timeout, interval time.Duration, cond func() bool, msg string, args ...any) {
+	t.Helper()
 	deadline := time.Now().Add(timeout)
 	for {
 		if cond() {
@@ -54,7 +70,7 @@ func Eventually(t testing.TB, timeout time.Duration, cond func() bool, msg strin
 			t.Fatalf("%s (condition not met within %s)", fmt.Sprintf(msg, args...), timeout)
 			return
 		}
-		time.Sleep(pollInterval)
+		time.Sleep(interval)
 	}
 }
 
--- a/src/internal/testutil/eventually.go
+++ b/src/internal/testutil/eventually.go
@@ -91,3 +91,36 @@ func EventuallyValue[T any](t testing.TB, timeout time.Duration, get func() (T, 
 	}, msg, args...)
 	return got
 }
+
+// EventuallyEveryFunc is EventuallyEvery with the failure message produced
+// LAZILY, at the deadline, instead of being formatted by the caller up front.
+//
+// It exists because a diagnostic captured eagerly is a diagnostic about the
+// wrong moment. Passing a rendered snapshot of the system into EventuallyEvery
+// as a printf argument:
+//
+//	EventuallyEvery(t, d, i, cond, "timed out\n%s", h.render())
+//
+// looks deferred — the formatting genuinely is — but h.render() is evaluated
+// BEFORE the wait starts. When the condition then never holds, the failure
+// prints the state the system was in at t=0 rather than at the timeout. For a
+// wait that hangs, that is the difference between seeing why it hung and
+// seeing a snapshot that has long since been superseded. This bit a TUI
+// acceptance harness where every hang reported the startup frame.
+//
+// describe is called at most once, only on the failure path, so it is free to
+// do real work: render a frame, dump a queue, walk recorded traffic.
+func EventuallyEveryFunc(t testing.TB, timeout, interval time.Duration, cond func() bool, describe func() string) {
+	t.Helper()
+	deadline := time.Now().Add(timeout)
+	for {
+		if cond() {
+			return
+		}
+		if time.Now().After(deadline) {
+			t.Fatalf("%s (condition not met within %s)", describe(), timeout)
+			return
+		}
+		time.Sleep(interval)
+	}
+}
--- a/src/pkg/tui/integration_test.go
+++ b/src/pkg/tui/integration_test.go
@@ -187,10 +187,17 @@ const (
   {"ts":"2026-09-01T12:09:05Z","user":"operator","action":"auditrefreshed","agent":"scanner"}
 ]}`
 	integrationModels = `{"backend":"claude","models":["claude-opus-4-5","claude-sonnet-4-5"],"fallback":false,"partial":false}`
-	integrationPacks  = `{"packs":[
+	// GET /api/packs returns a BARE ARRAY of packs, not an object wrapping one.
+	// The active level travels inside it as the `current` flag; there is no
+	// envelope and no separate level field. See client.ACMM, which decodes into
+	// []client.Pack and synthesises ACMMStatus from it. An enveloped fixture
+	// decodes to "cannot unmarshal object into Go value of type []client.Pack",
+	// which the overlay renders as a list error — so every selection assertion
+	// downstream waits forever on a cursor that can never have a pack under it.
+	integrationPacks = `[
   {"level":3,"name":"L3","description":"three","agentCount":3,"current":true,"agents":[]},
   {"level":4,"name":"L4","description":"four","agentCount":4,"current":false,"agents":[]}
-]}`
+]`
 )
 
 // integrationToken is the bearer every request must carry. It is asserted on
@@ -737,22 +744,40 @@ func (h *harness) stop() {
 // regression fail loudly instead of flakily.
 func (h *harness) waitFor(why string, cond func(model) bool) {
 	h.t.Helper()
-	testutil.EventuallyEvery(h.t, waitTimeout, pollStep, func() bool { return cond(h.snapshot()) },
-		"timed out waiting for %s\nlast frame:\n%s", why, h.view())
+	h.await(why, func() bool { return cond(h.snapshot()) })
 }
 
 // waitForView blocks until the rendered frame satisfies cond.
 func (h *harness) waitForView(why string, cond func(string) bool) {
 	h.t.Helper()
-	testutil.EventuallyEvery(h.t, waitTimeout, pollStep, func() bool { return cond(h.view()) },
-		"timed out waiting for %s\nlast frame:\n%s", why, h.view())
+	h.await(why, func() bool { return cond(h.view()) })
 }
 
 // waitForFixture blocks until cond holds on the fixture's recorded traffic.
 func (h *harness) waitForFixture(why string, cond func() bool) {
 	h.t.Helper()
-	testutil.EventuallyEvery(h.t, waitTimeout, pollStep, cond,
-		"timed out waiting for %s\nlast frame:\n%s", why, h.view())
+	h.await(why, cond)
+}
+
+// await is the shared wait, and it exists to render the failure frame at the
+// moment of FAILURE rather than at the moment of the call.
+//
+// Passing h.view() as a printf argument to EventuallyEvery — which is what
+// every wait here did originally — evaluates the frame BEFORE the wait begins.
+// The formatting is deferred to the timeout, but the string was captured at
+// t=0, so a hang reported the STARTUP frame no matter what the model went on to
+// do. Every one of these waits then failed with a picture of a screen that had
+// long since changed, which is precisely the information needed to tell "the
+// condition is wrong" from "the model never got there" — and it was wrong in a
+// way that looked plausible.
+//
+// Failing here instead, with the frame read after the deadline, is what makes
+// these timeouts diagnosable.
+func (h *harness) await(why string, cond func() bool) {
+	h.t.Helper()
+	testutil.EventuallyEveryFunc(h.t, waitTimeout, pollStep, cond, func() string {
+		return fmt.Sprintf("timed out waiting for %s\nframe at timeout:\n%s", why, h.view())
+	})
 }
 
 // settle waits for the message queue to drain and in-flight commands to finish.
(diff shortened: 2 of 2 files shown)


clubanderson:
Update on the verification status in the PR body: **the local suite has now completed and passed.**

`go test ./pkg/tui/` — **exit code 0** on darwin/arm64. That upgrades this from "compiles, unverified" to "compiles and passes locally". CI on linux/amd64 remains the gate.

The other caveat stands unchanged and is the one that matters: **the per-property mutation pass was not performed.** Nobody has broken each assertion in turn and confirmed the specific test fails. Until that happens, a green run here proves the tests do not error — not that they would catch the regression they name.

That distinction is the entire subject of this PR, so I would rather leave it stated than quietly let a passing run stand in for it. `Refs #5424` remains deliberate: the issue should close on a mutation-verified run.

clubanderson:
## Self-correction: my `Eventually` conversion broke this, and I pushed it on a misread

I converted six `time.Sleep` poll loops to `testutil.Eventually` to satisfy `TestSleepRatchet`, then pushed after reading "exit code 0" from a background job wrapper. That exit code was the wrapper's, **not `go test`'s**. The package actually failed:

```
FAIL	github.com/kubestellar/hive/pkg/tui	180.377s
```

180 seconds is the tell. The cause is a polling-interval mismatch I introduced:

| | interval |
|---|---|
| harness `pollStep` | **1ms** |
| `testutil.Eventually` `pollInterval` | **10ms** |

Ten times slower, across many waits per test, against a `waitTimeout` of 10s. The harness's own comment says exactly why that matters — *"microsecond, which is what makes `-count=20` cheap AND what makes a real regression fail loudly instead of flakily."*

So I traded a real property of this test suite for a ratchet number, which is the wrong trade in both directions.

## What I still think is right

`TestSleepRatchet` was correct to fail. #5424's acceptance criteria say to use channels and injected messages rather than real sleeps, and the harness had six raw `time.Sleep` calls. Raising `sleepBaseline` would defeat a guard that had just caught the thing it exists to catch — I am not proposing that.

## The actual fix

`Eventually` is the right destination; its fixed 10ms interval is the obstacle. Options, in the order I would try them:

1. **Give `Eventually` a caller-supplied interval** (or an `EventuallyEvery` variant) so a fast in-process harness can poll at 1ms while the default stays 10ms for network-shaped waits. Smallest change to `internal/testutil`, benefits every future caller.
2. **Drive the harness off channels rather than polling at all.** Strictly better and what the acceptance criteria actually ask for — this is a `teatest` harness with a real message loop, so the waits could observe messages instead of sampling state. Larger change to the 2,039-line file.
3. Keep raw sleeps and exempt this file. **I would not do this** — it is the exemption path the ratchet exists to prevent.

Option 2 is the honest reading of "use channels, injected messages, and short test-only durations". Option 1 unblocks faster.

## Process note

I have been telling agents all session that a green run proves nothing without checking what actually ran. I then pushed on an exit code from the wrong process. Flagging it rather than quietly re-pushing, because the whole point of this PR is that a test which appears to pass can be asserting nothing.

The PR remains unmerged and should stay that way until this is resolved.

clubanderson:
## The four failures are in the salvaged work itself, not in my conversions

I said the `Eventually` conversion broke this. That was wrong, and restoring 1ms polling did not fix it. I tested the **original** salvaged file — the one recovered verbatim from the dead agent's worktree, before I touched anything:

```
ORIGINAL file exit: 1
--- FAIL: TestModalKeysCannotLeakIntoGlobalActions (10.01s)
```

Same failure. So these were latent in the work from the start:

```
--- FAIL: TestStreamDropActivatesPollFallbackPreservesDataAndReconnectsOnce (10.01s)
--- FAIL: TestACMMTypedConfirmationAppliesExactlyOnceAtTheEscapedEndpoint (10.01s)
--- FAIL: TestModalKeysCannotLeakIntoGlobalActions (10.01s)
--- FAIL: TestPauseDialogSwallowsGlobalKeys (10.04s)
```

Every one times out at exactly `waitTimeout` (10s) — they hang waiting for a condition that never becomes true, rather than asserting something false.

## How I got here, since it matters more than the bug

I claimed this suite passed locally on the strength of a background job reporting "exit code 0". That was the **wrapper's** exit code. `go test` itself returned 1 both times. I then built a whole diagnosis — polling-interval mismatch — on top of a result I had never actually read, and pushed it.

Reading the output file is what caught it. That is the same failure I have been telling agents to avoid all session: a green signal that was never checked against what actually ran.

## What this PR now is

Not "salvaged work that needs a rebase". It is **an incomplete acceptance harness with four hanging tests**, and the four are among the most load-bearing assertions in it — stream-drop reconnect, the ACMM exactly-once guarantee, and both modal-containment tests.

I am leaving it open and unmerged rather than closing it. 2,039 lines with ~16 working tests is worth finishing, and the four failures are specific and diagnosable: each hangs on a `waitFor`/`waitForView` whose condition never holds, which usually means the harness is waiting for a message the model never emits under that fixture.

What it needs is someone to drive each of the four to the point where its condition is inspectable and find out what the model actually does — not another rebase and not a timeout bump.

## Retained value

The `EventuallyEvery` addition to `internal/testutil` is independently useful and correct: `Eventually`'s fixed 10ms interval is wrong for an in-process harness whose state changes in microseconds, and the doc comment records why. That can be split out and landed on its own regardless of what happens to this PR.

`TestSleepRatchet` was right to fail the original, and the conversion to zero `time.Sleep` calls stands. That part of the work is sound.

clubanderson:
## The four hangs were all harness bugs. None was a product bug.

Root-caused each one to the point where the failing condition was inspectable, as asked. No production code is touched, nothing was skipped, deleted or weakened, and `waitTimeout` is unchanged.

### The diagnostic bug that hid the rest

Every wait did this:

```go
testutil.EventuallyEvery(t, waitTimeout, pollStep, cond,
    "timed out waiting for %s\nlast frame:\n%s", why, h.view())
```

The *formatting* is deferred to the timeout, but `h.view()` is evaluated **before the wait starts**. So every one of these hangs printed the startup frame, no matter what the model did over the next ten seconds. That is why "each hangs on a `waitFor` whose condition never holds" was as far as this got — the evidence on screen was a picture of a moment before anything had happened.

Added `testutil.EventuallyEveryFunc`, which takes a `describe func() string` called only on the failure path, and routed every harness wait through it. The real frames made the remaining causes fall out in about an hour.

### The four, plus two more that surfaced once the diagnostics worked

1. **ACMM + ModalKeys** (shared cause, contrary to my expectation of the #5553 pattern) — the `/api/packs` fixture was enveloped as `{"packs":[...]}`. The endpoint returns a **bare array**; `client.ACMM` decodes into `[]client.Pack` and says so in a comment. It decoded to `cannot unmarshal object into Go value of type []client.Pack`, the overlay rendered a list error, and the cursor never had a pack under it — so every selection wait downstream hung forever.

2. **StreamDrop** — asserted `reconcileInterval == 5*time.Millisecond`, the harness's own test-only value. On recovery `sseDisconnected` restores the **production `pollInterval` constant**. The model never produces 5ms there and should not: asserting it was asserting that the model preserves a field the test wrote. Now asserts `pollInterval`. Nothing waits longer — the field is read the moment it changes.

3. **PauseDialog** — pressed `p` after `waitForView("scanner")`. That matches the **whole frame**, and the Events pane renders audit rows naming scanner on the independent activity chain, so it can go green before the Agents pane has any rows. `p` is a no-op unless `SelectedAgent()` reports ok, so the press was correctly ignored and the dialog never opened. Added `waitForRoster`, which waits through the same accessor the binding is gated on, and applied it at the seven sites with the same latent race.

Two more that `-count=5` and the fixed diagnostics exposed:

4. **The poll/stream contradiction was a race.** The fixture deliberately serves conflicting copies of `/api/status` (poll QUIET, stream SURGE) so a SURGE header proves the stream. But at the 5ms test cadence a poll already on the wire could answer *after* the event and put QUIET back. Fix: publish once to mark the stream healthy and stretch reconcile to 60s, wait for the stretch, then publish again — the event is the last writer by construction. The contradiction is retained, so the property still cannot be satisfied by the polling path.

5. **Attach was environment-dependent.** It asserts a preflight *failure*, but resolves tmux through `PATH` — and tmux is installed on most dev machines, including this one. It also waited on `attachPending`, which the failing preflight clears in microseconds. `PATH` is now emptied with `t.Setenv` so the no-tmux premise is guaranteed, and the bounding property is asserted on what *survives* three presses: one preflight, one footer failure, and `execRequested` still false — the tmux boundary asserted rather than assumed.

### One assertion I corrected rather than satisfied

`ModalKeysCannotLeakIntoGlobalActions` expected `K` and `q` to be **swallowed**, leaving `"APPLY L4"`. `app.go` deliberately makes *every rune* text while confirming, and says why: the rule is "runes are text while confirming", not "these keys are special". That is the stronger design — a swallowed `K` is a key that did nothing, whereas a `K` that lands in the field is a key that visibly *cannot* act.

So the field correctly holds `"APPLY L4Kq"`. I corrected the expectation and **added** the assertion the test was missing: that the polluted phrase does not match `ConfirmPhrase(4)`, and that `enter` on it issues **zero** applies. The exactly-once guarantee is unchanged and still proven exactly-once — it is not relaxed to at-least-once.

### Teardown was hiding a two-minute production wait

`TestQuitCancels...` called `h.cmdWG.Wait()`. By quit time the stream has stretched reconcile to `sseReconcileInterval`, so an outstanding **60-second** `tea.Tick` is in the wait group — 60s per subtest, twice. Exactly the production-duration wait #5424 forbids, sitting in teardown rather than in a sleep. Bounded it; `assertGoroutinesSettle` keeps its own deadline and still catches the per-stream reader leak.

**Suite went from 127s to 11s.**

### Verification — actual exit codes, captured from `go test` itself

I ran each to a file and read `$?` directly rather than trusting a wrapper:

```
go test ./pkg/tui/ -count=1        EXIT=0   ok  10.975s
go test ./pkg/tui/ -count=5        EXIT=0   ok  53.548s
go test ./pkg/tui/ -count=20       EXIT=0   ok  229.152s
go test ./pkg/tui/ -shuffle=on     EXIT=0   ok  10.990s
go test -race ./pkg/tui/...        EXIT=0   all 4 packages ok
go test ./internal/testutil/       EXIT=0
go build ./...                     EXIT=0
```

`TestSleepRatchet` passes; raw `time.Sleep` count in the harness is still **0** (the one sleep added lives inside `internal/testutil`, where the polling primitives belong).

Worth noting: my first "green" run here was a **flaky pass** — `-count=1` exited 0 while a fifth test was failing intermittently. `-count=5` is what caught it, and then caught two more. A single green run on this file still proves very little, which is rather the theme.

### What I did NOT verify

- **The per-property mutation pass is still not done.** This makes the tests *run*; it does not establish that each would fail if the property it names regressed. That remains the open item, and it is the one `Refs #5424` is waiting on.
- Only darwin/arm64. CI on linux/amd64 is still the gate — and the `-count=5` flake above was timing-shaped, so a slower runner is exactly where something new could appear.
- No product bug was filed, because I found none. Every failure was the harness asserting something the model was never specified to do.

clubanderson:
Independently verified before labeling — against `go test`'s own exit code this time, since I twice reported this suite passing on a background wrapper's status while `go test` returned 1.

```
go test ./pkg/tui/ ./internal/testutil/ -count=1   →  exit 0
  ok  github.com/kubestellar/hive/pkg/tui         11.362s
  ok  github.com/kubestellar/hive/internal/testutil 1.460s

go test ./pkg/tui/ -count=5                        →  exit 0
```

**166s → 11.4s.** That is the `cmdWG.Wait()` fix — teardown was blocking on an outstanding 60s `tea.Tick`, a production-duration wait wearing teardown's clothes.

## The finding that unblocked this

Every wait passed `h.view()` to `EventuallyEvery` as a printf **argument**, so the frame rendered *before* the wait began. All four hangs printed the startup frame no matter what the model did over the following 10s. The diagnostic evidence was a picture of a moment before anything happened — which is why this resisted diagnosis, including two wrong theories of mine (polling interval, then base staleness).

`EventuallyEveryFunc` with a lazy `describe func() string` is the right fix, and it is worth keeping in mind generally: **a failure message computed eagerly describes the wrong moment.**

## What the four actually were

All harness bugs. No product bug, so nothing to file — which is the outcome I would rank second-best, since a real defect found here would have justified the whole exercise.

1. **ACMM + ModalKeys** shared one cause: the `/api/packs` fixture was enveloped as `{"packs":[…]}` when the endpoint returns a **bare array**, so no pack was ever under the cursor. Worth noting they *did* share a cause, contrary to the #5553 precedent I warned about — the caution was right, the specific expectation was not.
2. **StreamDrop** asserted `reconcileInterval == 5ms`, the harness's own test-only value, where recovery restores the production `pollInterval`. It was asserting the model preserves a field the *test* wrote.
3. **PauseDialog** matched `"scanner"` against the whole frame, and the Events pane renders audit rows naming scanner on an independent chain — so it went green before the Agents pane had rows. `waitForRoster` applied at all 7 sites with the same latent race.

## The assertion correction is the right call

`ModalKeys` expected `K`/`q` to be swallowed. `app.go` deliberately makes every rune text while confirming, which is the stronger design — a key that lands *visibly* cannot act. Correcting the expectation to `"APPLY L4Kq"` **and adding** the missing assertion that the polluted phrase issues **zero** applies makes the test stronger, not weaker. Exactly-once is untouched.

## Still open, and correctly stated

The **per-property mutation pass is not done** — this makes the tests run, not proven to catch regressions. That is #5424's remaining item and the reason this stays `Refs`, not `Fixes`.

Also worth recording: the agent's first green run was a **flaky pass** — `-count=1` exited 0 while a fifth test failed intermittently; `-count=5` caught it, then caught two more. Exactly the argument for `-count=N` on a new harness.

PR #5575, opened 2026-09-01
🐛 fix(bin): ci-maintainer cadence keyed to a nonexistent 'reviewer' agent; missing cadence key now errors loudly

### LINKED ISSUE #5571: 🐛 bug(bin): ci-maintainer cadence key is unset in kick-governor.sh — never wakes on systemd deployments (not containerised hives)
Found while fixing #5549, and deliberately left out of #5567 because fixing it changes which agents run.

#5549 noted this as "not a defect on its own". It is worse than that.

## `ci-maintainer` is permanently paused in every governor mode

`bin/kick-governor.sh` keys `ci-maintainer`'s cadence defaults to an agent named `reviewer`, which never appears in `AGENTS_ENABLED`. `get_cadence` therefore falls through to `:-0` — and **cadence 0 means paused**, not "use a default".

So unless an operator has explicitly set `CADENCE_CI_MAINTAINER_*_SEC` for the mode in question, `ci-maintainer` never wakes. Not "runs on a default cadence" — never runs at all, in surge, busy, quiet and idle alike.

This is the same shape as the other findings this week: the mechanism runs, produces a plausible-looking number, and the number silently means "off". Nothing errors, nothing logs, and the agent simply never appears in the kick loop.

## Why it was not fixed in #5567

#5567 fixed three defects that were pure correctness — a pause check failing open, a crash on an unbound variable, and a downgrade to a nonexistent model. None changed which agents run.

This one does. Correcting the key makes `ci-maintainer` start waking on whatever cadence it should have had, on every hive that has not overridden it. That is a fleet behaviour change and deserves to be a deliberate, separately reviewable decision rather than a rider on a bug fix.

## What needs deciding

1. **What cadence should `ci-maintainer` actually have** in each of the four modes? The `reviewer` values it currently mis-keys to are not necessarily the right answer — they were written for a different agent. Someone should choose these, not inherit them by fixing a typo.
2. **Is the fallback itself wrong?** `:-0` conflates "no cadence configured" with "explicitly paused". A missing cadence arguably ought to be a loud configuration error, since silently disabling an agent is the more dangerous of the two readings. That would be a broader change and may catch other agents.

Worth checking (2) before fixing (1): if any other agent's key is mistyped, it is disabled right now with no signal, and a fix that only corrects `ci-maintainer` would leave the rest.

## Acceptance

- [ ] Determine whether any other agent name in the cadence tables fails to appear in `AGENTS_ENABLED` — that is a mechanical check and should be done first.
- [ ] `ci-maintainer` wakes on an intentionally chosen cadence in all four modes.
- [ ] A test asserts a mistyped/missing cadence key is distinguishable from a deliberate 0. Per the standard in force here, demonstrate it failing against the current fallback before fixing.

Refs #5549, #5567, #5548

clubanderson:
## The defect

`get_cadence()` builds its lookup key from the agent name, so `ci-maintainer` resolves `CADENCE_CI_MAINTAINER_<MODE>_SEC`. But every cadence and model default in `bin/kick-governor.sh` was written for an agent called **`reviewer`** — a name that never appears in `AGENTS_ENABLED`, whose default is `supervisor scanner ci-maintainer architect outreach`.

The lookup missed, fell through to `${!var_name:-0}`, and **cadence 0 means PAUSED**. So `ci-maintainer` never woke in surge, busy, quiet or idle. Nothing errored, nothing logged — the entire ladder documented in the script's own header (lines 8-34) was unreachable dead configuration.

## Scope — narrower than the issue originally stated

The issue was first filed as a fleet-wide outage. **It is not**, and I verified this independently rather than taking it on trust:

- `kick-governor.sh` is driven by `kick-governor.timer`, a systemd unit enabled from `bin/hive.sh:586`.
- It is **not** copied into the container image — `src/Dockerfile:442` and the surrounding `COPY` lines ship `gh-app-token.sh`, `hive-config.sh`, `agent-launch.sh`, `git-credential-hive.sh` and others, but neither `kick-governor.sh` nor `hive.sh`. `grep -rn kick-governor src/Dockerfile src/deploy/` returns nothing.
- Containerised hives run the Go governor in `src/pkg/governor`, which contains zero `CADENCE_` or `ci-maintainer` references.

**Blast radius is systemd/bare-metal deployments only.** On containerised hives `ci-maintainer` has been running normally the whole time. Operators should not read this as an agent fleet-wide waking up.

## Audit: every cadence-table name vs every enabled name

Requested before any fix, since a second mistyped key would be disabled right now with no signal.

Cadence-table names: `SCANNER`, `REVIEWER`, `ARCHITECT`, `SUPERVISOR`, `OUTREACH`, `STRATEGIST`, `ANALYST`, `GUARDIAN`.
Enabled agents (`AGENTS_ENABLED`, `bin/hive-config.sh:168` and four sites in `bin/hive.sh`): `supervisor scanner ci-maintainer architect outreach`.

| Name | In cadence table | In `AGENTS_ENABLED` | Status |
|---|---|---|---|
| scanner | yes | yes | OK |
| supervisor | yes | yes | OK |
| architect | yes | yes | OK |
| outreach | yes | yes | OK |
| **ci-maintainer** | **no** | **yes** | **BROKEN — this bug** |
| reviewer | yes | no | orphan key, the stale name |
| strategist / analyst / guardian | yes | no | opt-in Nous agents, expected |

**`ci-maintainer` is the only genuine mismatch.** `strategist`/`analyst`/`guardian` are the opt-in Nous fleet — they are absent from the *default* `AGENTS_ENABLED` by design and work when an operator enables them. The same `REVIEWER` typo also affects the **model** table (`MODEL_*_REVIEWER`), which I fixed too; that one was masked because `get_model_selection()` falls back to a *plausible* `copilot:claude-sonnet-4-6` rather than to "off".

That asymmetry is worth noting: for a missing key, `get_model_selection` fails to a safe default while `get_cadence` failed to "disabled". Same mistake, opposite safety posture.

## Decision on the `:-0` fallback

I did change it, because the conflation is the root defect and the `ci-maintainer` key was one symptom.

- Key **set** (to `0` or anything else) → honoured exactly as written, silently. Operators who deliberately pause with `0` are unaffected, and paused agents do not spam a warning every 15 minutes.
- Key **unset** → logs `CONFIG ERROR: <VAR> is not defined — '<agent>' has no cadence for mode '<mode>' and will NOT be kicked`, naming the exact variable to set.

The agent is **still skipped** on a missing key. I deliberately did not invent a fallback cadence: silently *starting* to kick an agent nobody configured is its own hazard. The behaviour stays fail-safe; it is now merely audible. Implemented with `[[ -z "${!var_name+x}" ]]`, which distinguishes unset from empty-or-zero.

## Cadences chosen for `ci-maintainer`

I did not simply inherit the `reviewer` numbers on the assumption they were right. What settles it is that **the script's own header has always documented ci-maintainer's ladder**, and it matches the `REVIEWER` values exactly:

```
#   BUSY  … ci-maintainer  → every 1 hour
#   QUIET … ci-maintainer  → every 45 min
#   IDLE  … ci-maintainer  → every 15 min
#   SURGE … ci-maintainer  → PAUSED
```

So `REVIEWER` is a **stale rename of `ci-maintainer`**, not a different agent with borrowed settings — the same conclusion holds for `MODEL_*_REVIEWER`, whose `claude:` values in surge/busy match the line-149 comment "priority agents (scanner, ci-maintainer) get metered Claude". Adopting the documented values makes behaviour match documentation that has been in the file all along, which is a smaller and more defensible change than inventing new numbers.

| Mode | Cadence | Rationale |
|---|---|---|
| surge | PAUSED | yields the queue to scanner under load, as documented |
| busy | 1 hour | comparable to architect/outreach yielding, but still active |
| quiet | 45 min | ramps up as the queue drains |
| idle | 15 min | matches scanner, the other priority agent |

`surge = PAUSED` is intentional, not a leftover of the bug. I also corrected a genuine contradiction in the header, which claimed ci-maintainer "always has priority" while the table shows it paused in surge — "priority" refers to model quality, not to always being kicked.

Old names are kept as deprecated aliases (`${CADENCE_CI_MAINTAINER_X:-${CADENCE_REVIEWER_X:-default}}`), so an operator who set `CADENCE_REVIEWER_*` in `governor.env` keeps their value; the new name wins when both are set. Both paths are tested.

## Before / after evidence

#5548 had already pinned this bug with three `[pinned bug]` assertions, which serve as independent before-evidence. Against the **unfixed** script with the new tests in place:

```
FAIL: ci-maintainer's idle cadence is now the documented 15min, not 0/off
FAIL: ci-maintainer IS kicked in idle mode now that its cadence resolves
FAIL: ci-maintainer's surge model is metered claude:, proving MODEL_SURGE_CI_MAINTAINER is consulted
FAIL: deprecated CADENCE_REVIEWER_IDLE_SEC is still honoured as an alias
FAIL: an agent with NO cadence key logs a CONFIG ERROR naming the missing variable
=== 55 passed, 5 failed ===
```

With the fix applied: **`=== 60 passed, 0 failed ===`**.

Every new assertion was confirmed to fail without the fix (by stashing only `kick-governor.sh` and re-running, after verifying the revert physically landed in the file). Cadences resolved directly from the shipped `get_cadence`:

```
before:  ci-maintainer: surge=PAUSED busy=PAUSED quiet=PAUSED idle=PAUSED
after:   ci-maintainer: surge=PAUSED busy=3600s  quiet=2700s  idle=900s
```

One anti-vacuity note: the **idle** model default coincidentally equals the old hardcoded fallback (`copilot:claude-sonnet-4-6`), so an idle-mode model assertion passes whether or not the fix is present. I moved that assertion to **surge**, where fixed (`claude:`) and broken (`copilot:`) genuinely differ.

`shellcheck -S error` and `bash -n` clean on both files.

## Base

Branched from **#5567's head**, which was still open (`UNSTABLE`, two `test (rest N/3)` shards queued) at the time. #5567 owns `bin/kick-governor.sh` and `bin/test_kick_governor.sh`; basing on `origin/v4` would have conflicted in both. This PR will show #5567's commits until it merges, after which the diff reduces to my three files. **Merge #5567 first.**

## What I did NOT verify

- **Nothing was run on a real systemd hive.** All evidence is from the hermetic test harness, which stubs `kick-agents.sh`, `notify`, `date` and `gh`. I did not confirm `ci-maintainer` actually does useful work when kicked — only that the governor now kicks it.
- I did not audit `bin/kick-agents.sh` or the Go governor for a parallel version of this bug beyond confirming `src/pkg/governor` has no `CADENCE_` references.
- I did not survey deployed `governor.env` files, so I cannot say how many hives already set `CADENCE_CI_MAINTAINER_*` and were unaffected.
- The `strategist`/`analyst`/`guardian` entries are judged expected-by-design from their "Nous agents (Phase 3)" comment; I did not trace whether an operator enabling them hits a separate issue.
- Whether `ci-maintainer` waking on bare-metal hives has a token-budget cost worth a heads-up — it now draws metered `claude:` in surge/busy.

Fixes #5571


clubanderson:
Two notes before this merges.

**#5567 has now merged**, so the base dependency is satisfied. `mergeable=true` and the diff resolves correctly — git sees #5567's commit as already-applied, and the 3-file diff (`CHANGELOG.md`, `bin/kick-governor.sh`, `bin/test_kick_governor.sh`) is the right shape. No rebase needed.

**I verified the central claim rather than relaying it**, because it changed what the correct fix was. My brief told this agent *not* to inherit the `reviewer` cadence values, on the reasoning that they were written for a different agent. That reasoning was wrong. The script's own header documents:

```
#   SURGE:  ci-maintainer  → PAUSED
#   BUSY:   ci-maintainer  → every 1 hour
#   QUIET:  ci-maintainer  → every 45 min
```

Those match `CADENCE_REVIEWER_SURGE_SEC=0` / `_BUSY_SEC=3600` / `_QUIET_SEC=2700` at lines 95-98 exactly. `REVIEWER` is a stale **rename** of `ci-maintainer`, so adopting the values restores documented intent rather than inventing cadences. The agent was right and I was wrong.

## One item for an operator, not covered by this PR

On systemd/bare-metal hives, `ci-maintainer` will now start drawing metered `claude:` in surge and busy where it previously drew nothing at all. That is the intended behaviour per the header, and the `MODEL_*_REVIEWER` half of the same typo was failing silently *open* (wrong model) rather than closed — so this corrects a real inconsistency. But it is a token-budget change for that deployment mode and worth a heads-up rather than a surprise.

## The larger question this does not answer

`kick-governor.sh` may be dead code. Last modified **2026-05-14**; `kick-governor.timer` does not exist in `systemd/` (which ships ten units, none of them this one); `bin/hive.sh:586` runs `systemctl enable --now kick-governor.timer` for a unit the repo no longer provides; and it is not copied into the container image.

If nothing actually runs this script, the correct change is **deletion** — of the script and of `hive.sh`'s references to a missing timer — not a cadence fix. This PR is correct either way and I am not blocking on it, but that determination is worth making before more effort goes into the file. Dead code that looks live is what produced the original mis-scoped report here.

clubanderson:
## Correction: `kick-governor.sh` is legacy, but it is NOT dead — do not delete it

I raised the possibility that this script is orphaned and that deletion would beat a cadence fix. I checked properly and that was wrong. Recording it here because I said it on both this issue and #5575.

**It ships and installs today:**

- `install.sh:94` installs it to `$BIN_DIR`, and line 13 states: *"The kick-governor is always installed alongside the agent supervisor."*
- **23 files** reference it, including `dashboard/server.js`, `bin/kick-agents.sh`, `bin/supervisor.sh`, `bin/kick-outcome-tracker.sh`, and `examples/kubestellar/agents/governor.env`
- `kick-agents.sh` reads its state at `/var/run/kick-governor/model_<agent>` — a live runtime contract, not a stale comment
- `bin/test_kick_governor.sh` runs in `v2-ci.yml`

**Where my inference went wrong.** I argued "`kick-governor.timer` is not in `systemd/`, so nothing runs it." That does not follow. `install.sh` enables `hive.service` → `supervisor.sh`, and the per-agent timers (`kick-scanner.timer`, `kick-ci-maintainer.timer`, …) were **superseded by kick-governor** and are explicitly disabled on install (`install.sh:19-24`). The timer's absence means the scheduling mechanism changed, not that the script was abandoned.

**What is genuinely true:** `bin/README.md:32` calls it the *"Legacy adaptive governor"* — the repo's own word. It is not in the container image, containerised hives run the Go governor (`src/pkg/governor`), and it was last modified 2026-05-14. So it is legacy and unmaintained, serving the **bare-metal/VM install path** that `install.sh` supports.

## What this means

**#5575 is the right fix, not wasted effort.** Deleting the script would break a supported install path and orphan `kick-agents.sh`'s state-directory contract.

If anything, the legacy status *strengthens* the case for the fix as written. Restoring the cadences the file's own header documents matters more in unmaintained code, because nobody is going to re-derive the intent later — the header is the only surviving record that `REVIEWER` was a rename of `ci-maintainer`.

The scope statement stands unchanged and verified twice: **systemd/bare-metal only**, containerised hives unaffected.

PR #5584, opened 2026-09-01
contribute: allow multiple concurrent sessions per GitHub account

hanthor:
## Problem

One GitHub account maps to one contributor profile per hub (one `ContributorID`, one registration token, one trust tier). Every identity-keyed piece of hub state — task leases, assignment cooldowns, failure streaks, ownership generation fences — keys on `identityOf()`, which returns that single `ContributorID`. So two relays under the same account collide on **one active-task slot**: a contributor with a single GitHub account can only run **one backend at a time**.

This is a real limitation for anyone who wants to run several CLI backends (claude, agy, pi, kiro, …) concurrently but has only one GitHub identity.

## Change

Add an **optional** client-declared session label:

- Relay: `HIVE_SESSION` (defaults to the backend name), sent as `session` in `auth_response`.
- Hub: `identityOf()` returns `ContributorID#session` when a session is declared.

Identity-keyed state (leases, cooldowns, failure streaks, ownership fences) becomes per-session, so concurrent relays under one account get independent task slots. **Auth, trust tier, model admission and rate-limit accounting stay per-account** — a session is not a way to escape your account's limits, only to run more than one backend at once.

## Backward compatibility

Fully additive. A relay that declares no session (or an older relay that doesn't know the field) resolves to the bare `ContributorID`, exactly as today. An older hub ignores the unknown `session` field. No migration, no protocol break.

The label is sanitized (`[A-Za-z0-9._-]`, capped at 32 bytes) before it becomes a map key, log field, or UI string.

## Tests

`src/pkg/dashboard/contribute_ws_session_identity_test.go`:
- session scoping produces distinct identities per session
- no-session and username-fallback paths preserve the historical identity
- label sanitization (path chars, whitespace, the `#` separator, length cap)

Full `pkg/dashboard` suite passes.

### DIFF
--- a/bin/contributor-relay.sh
+++ b/bin/contributor-relay.sh
@@ -61,6 +61,17 @@ const PI_ENV = BACKEND === 'pi' ? { ...process.env } : {};
 let piInvocationState = 'untested';
 const REASONING_EFFORT = process.env.AGENT_REASONING_EFFORT || '';
 const AGENT_ROLE = (process.env.HIVE_AGENT_ROLE || '').trim();
+// HIVE_SESSION — optional session label (multi-session-per-account). One GitHub
+// account has one contributor identity per hub, and the hub keys task
+// leases/cooldowns/ownership on that identity, so two relays under the same
+// account would collide on a single active-task slot. Declaring a distinct
+// session gives each relay an independent session-scoped identity
+// (ContributorID#session) on the hub while auth/tier stay per-account. Defaults
+// to the backend name so the common case — one relay per CLI backend under one
+// account — works with no extra config. Omitted only if explicitly emptied.
+const AGENT_SESSION = (process.env.HIVE_SESSION !== undefined
+  ? process.env.HIVE_SESSION
+  : BACKEND).trim();
 // Neutral directory both entrypoints launch the CLI from ($HOME). Used to pin
 // the cwd on relaunch; see launchCommandWithCwd for why the relay's own cwd is
 // the wrong answer in local mode.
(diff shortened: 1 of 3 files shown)


clubanderson (approved):
Reviewed. **Approve.** 39 checks complete, zero failures.

The design is additive and the backward-compatibility story holds: an omitted label yields the bare `ContributorID`, which is byte-for-byte the previous single-session behaviour. Nothing changes for existing relays.

## What I checked rather than took from the description

**Sanitization happens at auth time, before the label becomes a map key.** `sanitizeSessionLabel` keeps only `[A-Za-z0-9._-]` and caps at 32 bytes, so `ContributorID#session` stays a single clean token and a client cannot smuggle path or format characters into a key that also lands in logs and the UI. An all-stripped label returns `""` and takes the historical path — failing toward the safe default rather than an odd empty-suffix key.

**The split is the right one.** Lease, cooldown, failure-streak and ownership-fence state becomes session-scoped; auth, trust tier, model admission and rate-limit accounting stay per-account. That is the correct boundary — the first group is about *concurrent work slots*, the second about *who you are and what you are allowed to do*. Making the second session-scoped would let a contributor multiply their own rate limit by declaring labels, which this deliberately does not do.

## One thing worth a second pair of eyes

This touches `contribute_ws.go`, and **#5090 is still open and unresolved** — contributor WebSocket flapping whose residual cause we have not found. Today established that the `heartbeat ping failed` line was a lagging indicator (a fixed 29s offset from registration, i.e. our own ticker) and that upgrade churn was not the cause either. The ingress path is not cleared.

Nothing here looks like it would affect that, and the change is orthogonal — identity keying, not connection lifecycle. But if flap behaviour shifts after this lands, this is a change to correlate against, because it alters what a reconnecting relay's identity resolves to. Worth knowing rather than discovering later.

Related, and genuinely improved by this: #5151 collapsed the feed churn from short reconnects, and a session-scoped identity means two relays under one account no longer contend for a single lease slot — which removes one source of spurious release/reclaim traffic.

The three tests cover the scoping, the fallback, and the sanitizer. I would have liked one asserting that two sessions under the same account genuinely get independent lease slots end-to-end rather than only at `identityOf()`, but that is a nice-to-have on top of a sound change, not a blocker.

PR #5586, opened 2026-09-01
🐛 fix(dashboard): persist dropdown choices to agent overlays

### LINKED ISSUE #5558: 🐛 bug: dashboard model/method dropdowns don't write the per-agent overlay, so saved config silently diverges from what's running
## What happened?

Hive stores an agent's backend and model in two places: a per-agent YAML
overlay file under `data.agents_dir` (the file that wins on every config
load) and a runtime override in `/data/hive-state.json` (the value the agent
actually launches with). The dashboard's agent-card dropdowns — the "method"
and "model" pickers on the governor table — write the runtime override and
`hive.yaml`, but **not** the per-agent overlay file. Because the overlay is
what a config load reads, the saved configuration keeps saying whatever the
ACMM pack last set, indefinitely.

Operators do not see this as a revert, which is what makes it hard to
notice. The state file is restored on boot and takes precedence at launch, so
the agent keeps running the chosen model across restarts and the dashboard
keeps showing the right thing. The two values silently disagree until
something clears the runtime override — at which point the agent drops back
to a model the operator changed away from weeks earlier, with no record in
the UI of why.

Observed on a self-hosted hive at ACMM L4, agent `guide`:

| Source | Backend | Model |
|---|---|---|
| `GET /api/status` (live process) | `claude` | `claude-sonnet-5` |
| `GET /api/config/agent/guide/export` (saved config) | `copilot` | `claude-sonnet-4-6` |

`/api/status` also reports `"needsRestart": true` for the agent, so the
backend is aware the two disagree, but nothing in the dashboard surfaces
that the *saved* value differs from the running one.

## Expected behavior

A model or backend chosen from the dashboard should be written to the
per-agent overlay file, so the saved configuration matches what is running
and survives a cleared runtime override. Every other config-mutating handler
in `pkg/dashboard` already does this.

## Root cause

`claimAgentFieldOwnership` (`src/pkg/dashboard/api.go:1653`) is called by
`handleSwitch` (backend dropdown, `api.go:1719`) and `handleModelSet` (model
dropdown, `api.go:1745`). It sets `Model`/`Backend`, stamps
`FieldOwnerOperator`, calls `UpdateConfig`, and then persists with
`s.saveConfig()` — and stops there. It never calls
`config.SaveAgentFile(s.deps.Config.Data.AgentsDir, name, ac)`.

That omission is load-bearing because of the merge order in
`config.Load` (`src/pkg/config/config.go:4075`):

```go
if cfg.Data.AgentsDir != "" {
    overlays, err := LoadAgentOverrides(cfg.Data.AgentsDir)
    ...
    cfg.MergeAgentOverrides(overlays)
```

`MergeAgentOverrides` (`src/pkg/config/agent_overlay.go:59`) does a
wholesale `c.Agents[name] = agent` for every overlay file, so for a managed
agent the overlay replaces whatever `saveConfig()` wrote to `hive.yaml`. The
operator's choice — and the `model_owner: operator` marker that is supposed
to protect it from the next ACMM pack apply — are both discarded on the next
load.

Every comparable handler pairs the two writes. `handleAgentConfigModels`
(`api.go:3738`), which serves `PUT /api/config/agent/{name}/models` from the
Settings panel, does exactly the right thing at `api.go:3786-3793`:

```go
if err := s.saveConfig(); err != nil { ... }
if agentsDir := s.deps.Config.Data.AgentsDir; agentsDir != "" {
    if err := config.SaveAgentFile(agentsDir, name, agentCfg); err != nil { ... }
}
```

So the same operator intent persists correctly through Settings and
incorrectly through the card dropdown, which is why this reproduces
inconsistently depending on which control an operator happens to use.

The reason it stays invisible is `restoreAgentRuntimeState`
(`src/cmd/hive/state_restore.go:86`), which replays
`ModelOverride`/`BackendOverride` from `/data/hive-state.json` on every boot:

```go
if as.ModelOverride != "" {
    if err := agentMgr.SetModelOverride(name, as.ModelOverride); err != nil {
```

`proc.ModelOverride` beats `proc.Config.Model` at every read site
(`status_builder.go:536`, `manager.go:2528`, `manager.go:5121`), so the
stale overlay never gets a chance to show itself while the override exists.

## Reproduction steps

1. On a hive with `data.agents_dir` set and a pack-managed agent (`managed:
   true`), note the agent's configured backend/model via
   `GET /api/config/agent/<name>/export`.
2. From the dashboard governor table, use the agent card's **model** dropdown
   to pick a different model (this hits `POST` → `handleModelSet`, not the
   Settings panel).
3. `GET /api/status` — the agent now reports the new model and
   `"needsRestart": true`.
4. `GET /api/config/agent/<name>/export` — the saved config still reports the
   **old** backend/model. Inspect `<agents_dir>/<name>.yaml` directly: it is
   unchanged, and carries no `model_owner: operator` marker.
5. Restart the hive. The agent still runs the new model (restored from
   `/data/hive-state.json`), so the drift is not visible — but the config and
   the overlay file remain stale.

## Environment

- Branch/version: `v4`, at fe96434a
- Deployment method: self-hosted Kubernetes
- Relevant configuration: ACMM L4 Security-Aware pack; affected agent
  `guide` is `managed: true` (pack-owned, loaded from the overlay directory)

## Additional context

Two things worth deciding alongside the fix:

- **Scope.** The minimal fix is adding the `SaveAgentFile` call to
  `claimAgentFieldOwnership` so both dropdown paths persist like every other
  handler. It may be worth auditing for other `saveConfig()`-without-
  `SaveAgentFile` writers, since this class of drift is silent by
  construction.
- **Visibility.** `needsRestart` already marks live/config divergence but
  nothing renders it. Surfacing it on the agent card would turn the next
  occurrence of this class of bug into something an operator can see rather
  than something they discover months later when an override is cleared.

Danathar:
## Summary

- Persist dashboard governor-card model and method selections to the managed agent's per-agent YAML overlay, not only to the runtime override and base `hive.yaml`.
- Preserve the existing persistence alert until both durable writes succeed, with an overlay-specific error that explains the next-config-load reversion risk.
- Add regression coverage for stale overlay replacement and document the user-visible fix in `CHANGELOG.md`.

## Problem and user impact

The governor table's model and method dropdowns update the live process override and call `saveConfig()`, so the selected value appears to work and is replayed from `/data/hive-state.json` after restarts. For a managed agent, however, the per-agent file under `data.agents_dir` is the authoritative configuration layer: every config load replaces the corresponding `hive.yaml` agent entry wholesale from that overlay.

Because the dropdown path did not update the overlay, the saved backend/model and its operator-ownership marker remained stale. Runtime state hid that divergence until an override was cleared; at that point the agent silently fell back to the old pack-provided choice.

## Root cause

Both dropdown handlers, plus the stronger pin path, converge on `claimAgentFieldOwnership`. The helper updated the in-memory `AgentConfig`, synchronized the manager, and persisted `hive.yaml`, but stopped before calling `config.SaveAgentFile`. Comparable Settings handlers already perform both writes.

The base-config write alone cannot be durable for managed agents because `config.Load` calls `MergeAgentOverrides`, which replaces the entire agent entry from the unchanged overlay. A later periodic save can then write that stale merged value back to the runtime config as well.

## Implementation

- `src/pkg/dashboard/api.go`
  - After `saveConfig()` succeeds, write the same updated `AgentConfig` to `Data.AgentsDir` with `config.SaveAgentFile` when an overlay directory is configured.
  - Keep the existing `agent-field-save-failed` alert active when the overlay write fails and identify that the value can revert on the next config load.
  - Clear the alert only after every configured persistent layer succeeds.
  - Update the helper documentation to state the two-layer durability invariant.
- `src/pkg/dashboard/agent_model_ownership_test.go`
  - Seed an overlay with the old agent values, exercise the shared ownership helper, reload the overlay, and assert that both model/backend values and both operator-owner markers were replaced.
- `CHANGELOG.md`
  - Add the operator-visible fix under Unreleased / Fixed.

## Design notes and scope

The write order matches existing dashboard configuration handlers: save the base config first, then the per-agent overlay. `SaveAgentFile` receives the complete effective agent config because overlays replace agent entries wholesale rather than merging individual fields.

No overlay is created when `data.agents_dir` is unset, preserving deployments that use only `hive.yaml`. This PR intentionally does not add a new `needsRestart` UI indicator or audit unrelated config writers; those are separable visibility and hardening improvements and are not required to repair the reported persistence path.

## Testing

- [x] `cd src && env GOCACHE=/tmp/hive-5558-go-cache CCACHE_DIR=/tmp/hive-5558-ccache go test ./pkg/dashboard -run 'Test(ClaimAgentFieldOwnership|ApplyPackPreservesOperator|PinClaimsOwnership)' -count=1` — PASS (`0.121s`)
- [x] `cd src && env GOCACHE=/tmp/hive-5558-go-cache CCACHE_DIR=/tmp/hive-5558-ccache go test ./pkg/dashboard -count=1` — PASS (`53.219s`)
- [x] `cd src && env GOCACHE=/tmp/hive-5558-go-cache CCACHE_DIR=/tmp/hive-5558-ccache go vet ./pkg/dashboard` — PASS
- [x] `git diff --check` — PASS

The package-wide test's first sandboxed attempt could not open the loopback socket used by existing `httptest` coverage (`socket: operation not permitted`). The identical command passed once local loopback access was permitted; this was an execution-environment restriction, not a test failure.

## Related issues

Fixes #5558

## Contributor checklist

- [x] PR targets `v4`, the branch/version named in the issue reproduction.
- [x] Title uses the repository emoji convention.
- [x] Commit includes a DCO sign-off matching the author email.
- [x] User-visible behavior is recorded in `CHANGELOG.md`; no separate docs are needed for this corrective persistence behavior.

— hive: backend=codex


clubanderson (approved):
Reviewed. **Approve.**

The diagnosis is the valuable part. The previous fix wrote the operator's choice to `hive.yaml` plus an ownership marker, which is correct as far as it goes — but for **managed** agents the per-agent overlay *replaces* the hive.yaml entry on every config load. So the write landed in a layer that was then overwritten, and the operator's selection silently reverted on the next reload.

That is a persistence bug where the value was genuinely written and genuinely lost, which is the hardest kind to spot from the UI: the dropdown shows the right thing until something reloads config, and then it does not. Writing both layers is the fix; the comment now states why both are required, so the second `SaveAgentFile` call cannot be removed later as apparent duplication.

The error path is handled the right way round. A failed overlay save raises a system alert that names the consequence explicitly —

> Could not save the model/method choice for X to its agent overlay — **it will revert on the next config load**

— rather than logging and continuing. An operator who sees the dropdown accept their choice and then silently revert has no way to connect those two events; being told at the moment of failure is what makes it diagnosable. It also returns rather than proceeding, so a half-persisted state is not presented as success.

One thing I checked rather than assumed: the alert fires on the overlay write specifically, so a hive with no `AgentsDir` configured takes the guarded path and is unaffected.

clubanderson (approved):
Verified against the surrounding v4 code:

1. The fix writes through the durable per-agent overlay layer — config.SaveAgentFile into Data.AgentsDir — mirroring the pattern handleAgentConfigModels already uses (saveConfig, then SaveAgentFile), which is the layer config.Load's MergeAgentOverrides treats as authoritative for managed agents. This is exactly the omission issue #5558 identified.
2. The auto-model case is respected, not fought: claimAgentFieldOwnership is only reachable from the operator-facing handlers (handleSwitch, handleModelSet, and the pin path), so automatic model re-application never routes through it and cannot stamp operator ownership into the overlay. Selecting the copilot "auto" sentinel persists the sentinel itself, which stays coherent with CLI-side auto-selection.
3. The regression test seeds a stale overlay, exercises the shared helper, and reloads it via config.LoadAgentOverrides — the same function config.Load merges from — asserting both values and both operator-ownership markers, in a per-test TempDir (no shared fixture state).
4. Failure handling is fail-visible and correctly ordered: an overlay write failure keeps the agent-field-save-failed alert active, skips the clear, and names the next-config-load reversion risk; the base-config write happens first, matching the existing handlers. No overlay is created when data.agents_dir is unset. No new locking hazards — SaveAgentFile is a plain file write outside any manager mutex.

Issue #5570, opened 2026-09-01
🧪 pkg/dashboard: one cachedHealth leak under -shuffle (not 83) — and wire a bounded -shuffle lane


clubanderson:
Found while fixing #5553, and deliberately left unfixed there to keep that PR surgical.

## Measured

`pkg/dashboard` fails **83 tests on unmodified `origin/v4`** under `go test -shuffle=200`. That is broad order-dependence, not a handful of flakes.

The #5553 investigation established the control: baseline fails 83 including its victim; the #5569 branch fails **82** — the same set minus the one test it fixed. So #5569 is genuinely surgical, and the remaining 82 are pre-existing.

## Why this matters more than "some tests are order-dependent"

`-count=N` does **not** find these. The #5553 brief specified `-count=20` and all three of its tests passed 20/20, because the pollution comes from a *different test in the same package*, not from repetition. Only `-shuffle` reproduces them.

That means the default CI configuration cannot see this class of failure at all. It surfaces only when the runner happens to pick a hostile order — which is exactly what produced #5553, a scheduled-run failure on a commit that passed five times before and after.

The three root causes #5553 fixed are all the same shape and are likely representative:

- a package-level cache (`cachedHealth`) written by a test that did not restore it
- state reset on exit only (`defer Reset()`), so tests asserting the *absence* of state depended on run order
- a goroutine outliving the test that spawned it, racing a global another test writes

None is a timing bug. All three are shared mutable state without a discipline for restoring it.

## Suggested approach

Do **not** attempt all 82 in one PR — it would be unreviewable and would collide with everything else in `pkg/dashboard`.

1. Get the exact failing set and group by shared symbol: `cd src && go test ./pkg/dashboard/ -shuffle=200 2>&1 | grep -E '^--- FAIL'`. The #5553 fixes suggest a small number of globals account for many of the 82.
2. Fix one global per PR, restoring in `t.Cleanup` rather than `defer`, and state the before/after failing count each time.
3. Once the count reaches zero, consider adding `-shuffle=on` to the `pkg/dashboard` CI lane so this cannot regress silently. That is the durable fix — until it runs in CI, the discipline decays.

## Acceptance

- [ ] `go test ./pkg/dashboard/ -shuffle=on -count=1` passes across at least 20 distinct seeds.
- [ ] Each PR states the failing count before and after; a PR that does not reduce it should say why.
[...]

PR #5588, opened 2026-09-01
🐛 fix(reviewer): audit escalated PR adjudications

### LINKED ISSUE #5480: [v5 RFC] Reviewer lane: an agent role with authority to adjudicate escalated (needs-human) PRs

Danathar:
## Problem

The v5 reviewer lane merged in #5485 covers bounded escalated-PR adjudication, and #5515 closes its ledger liveness gaps. One requirement from #5480 was deliberately deferred: every adjudication must reach both the attribution trail and the advisory digest.

The shipped kick used direct `gh pr comment` for recommend-close and no durable audit step for repair or de-escalation. Label and close actions could therefore make an adjudication terminal while leaving no `agent_pr_reviewed` event and no advisory bead. Operators could see the resulting labels or branch changes, but Hive's own activity trail and digest could not answer who adjudicated the PR, what evidence supported the verdict, which verification ran, or what risk remained.

## Root cause

The lane's original contract treated GitHub labels/comments as the complete verdict protocol. That bypassed the existing `hive-review` request watcher, which is the chokepoint that submits an App-bot review and records `agent_pr_reviewed`, and it never invoked `bd create --type advisory`, which is how agent findings enter the advisory digest.

There is also an asynchronous-delivery edge: `hive-review` returns after atomically queuing its request, while the watcher posts the review and writes `<request>.result.json` on a later tick. Treating queue acceptance as completed audit would still permit a label or close before a denied, retried, or quarantined review request was known to have landed.

## Implementation

- Extends the hardcoded reviewer-lane contract in `src/pkg/scheduler/reviewer_lane.go` so every REPAIR, DE-ESCALATE, and RECOMMEND-CLOSE verdict must:
  - submit an evidence-rich comment review through `hive-review`, producing the existing `agent_pr_reviewed` audit event;
  - wait for the printed `.result.json` file to report `"ok": true` rather than treating a queued request as delivery;
  - create a matching advisory bead with the configured reviewer agent as actor and `gh-<owner/repo>#<number>` as its stable external reference;
  - complete both records before removing `needs-human`, applying `reviewer-passed` / `reviewer-recommend-close`, or closing the PR.
- Replaces the recommend-close path's direct `gh pr comment` with the same attributed App-bot review path used by the other verdicts. Its `[reviewer] recommend close:` marker remains intact, and the contract forbids a duplicate comment.
- Makes audit failure fail safe: the reviewer must retain `needs-human`, omit terminal reviewer labels, and avoid closing so a later kick can retry rather than silently consuming the one reviewer pass.
- Updates `bin/hive-review.sh` to print the exact result-file path, matching the asynchronous response contract already implemented by the watcher and exposed by the other request shims.
- Documents the result-path behavior in `bin/README.md` and records the operator-visible fix in `CHANGELOG.md`.

## Design decisions and invariants

This reuses the existing review-request watcher and advisory-bead pipeline instead of adding a reviewer-specific event type or a second delivery mechanism. The watcher already provides App identity, ACMM authorization, UID forge resistance, retry/quarantine behavior, canary scanning, attribution trailers, and the canonical `agent_pr_reviewed` audit action.

The review result is checked before the bead and before terminal GitHub state. This preserves the issue's finite-automation invariant: an adjudication is not considered consumed until its visible review and Hive audit entry exist. The advisory bead is still a local synchronous write; if it fails, `needs-human` remains authoritative and the PR stays eligible for retry.

The change does not alter queue membership, the three-PR cap, ACMM L5/L6 authority, the one-pass labels, the ledger reconciliation from #5515, or which PRs the reviewer may touch.

## Regression coverage

- `TestBuildReviewerMessage_ContractAtL5` now pins the `hive-review --comment` path, `agent_pr_reviewed` attribution, asynchronous `ok` confirmation, advisory-bead command and external reference, actual reviewer actor, digest intent, and fail-safe retention of `needs-human`.
- `bin/test_hive_review.sh` now proves the wrapper advertises the exact `.result.json` path corresponding to the request it created, in addition to its existing parsing, event, body, agent, and atomic-write checks.
- Existing L5/L6, queue-cap, prior-pass exclusion, empty-queue, and role-routing tests continue to pass in the full scheduler package.
- Existing review-request watcher tests prove successful delivery records the audit event and that authorization/canary failures surface through the result contract.

## Verification

Passed:

```text
CGO_ENABLED=0 GOCACHE=/tmp/hive-5480-go-cache TMPDIR=/tmp/hive-5480-tmp go test ./pkg/scheduler
ok github.com/kubestellar/hive/pkg/scheduler

bash -n bin/hive-review.sh bin/test_hive_review.sh
bash bin/test_hive_review.sh
15 passed, 0 failed

CGO_ENABLED=0 GOCACHE=/tmp/hive-5480-go-cache TMPDIR=/tmp/hive-5480-tmp go test ./pkg/github -run 'Test.*ReviewRequest' -count=1
ok github.com/kubestellar/hive/pkg/github

git diff --check
```

The first sandboxed full-package attempts could not open localhost `httptest` sockets (`operation not permitted`); rerunning the same tests with the test environment's socket permission produced the passing results above. An initial CGO build also found the host's read-only ccache, so the recorded commands use the repository's established `CGO_ENABLED=0` test posture and task-local writable Go cache.

## Scope boundaries

This PR does not create a new audit event schema, synchronously submit reviews from the agent process, change retry horizons, or make label edits transactional with GitHub review delivery. It closes the shipped contract gap by requiring confirmation from the existing watcher before dependent actions. A process crash between confirmed review/bead creation and final relabeling can cause a later kick to repeat the investigation, but cannot silently lose audit evidence or remove the PR from `needs-human` without it.

Fixes #5480

— hive: backend=codex


clubanderson:
Rerunning `test (rest 2/3)` — the failure is almost certainly not yours.

**Failing test:** `TestStaleGeneration_RevokedWorkerCannotOverwriteNewOwner` in `pkg/dashboard`.

**This PR touches `bin/` and `src/pkg/scheduler/` only** — nothing under `pkg/dashboard`. There is no plausible path from a reviewer-lane change to a dashboard worker-generation test.

Two things point the same way. `v5` itself is green at `dcbca379`. And `pkg/dashboard` has a known flake signature established earlier today: an agent measured it at "83 failures under -shuffle", and the real cause turned out to be **ephemeral-port exhaustion** (`connect: can't assign requested address`) because the package stands up hundreds of httptest servers. Re-running the worst seeds after the socket table drained gave 18 to 0 to 0 — same seed, same ordering. The genuine order-dependent count was 1, since fixed by #5578.

So a lone `pkg/dashboard` failure on a PR that does not touch it is the expected shape of that flake rather than a signal about this change.

Flagging rather than silently rerunning, because "unrelated flake" is exactly the reasoning that lets a real regression through. If it reproduces on the rerun I will treat it as real and look properly — a security-boundary test asserting that a revoked worker cannot overwrite a new owner is not one to wave past.

Tracking for the durable fix: **#5570** covers wiring a `-shuffle` lane, now unblocked since the package is at zero. Its two preconditions are both about ports rather than ordering — bound the parallelism (`-p 1` or a capped `-parallel`), and surface the seed on failure. Without the first, that lane would flake on loaded runners and get ignored, which is how this class stays invisible.

clubanderson (approved):
Verified against the surrounding v5 code:

1. Audit shape reuses existing conventions. The comment review flows through the hive-review request watcher, which emits the canonical agent_pr_reviewed action (AuditActionPRReviewed, pkg/github/attribution.go), and the advisory bead command matches the policy-pack bd create conventions with the planning package's stable gh-<repo>#<number> external-ref form.
2. The advertised result path in bin/hive-review.sh (${REQ_FILE%.json}.result.json) matches the watcher's TrimSuffix(reqPath, ".json") + ".result.json" output exactly, and ReviewResponse carries the ok field, so requiring "ok": true is the correct delivery-confirmation contract for the asynchronous path.
3. Escalation semantics are untouched: this changes kick-message text only. The needs-human / reviewer-passed / reviewer-recommend-close label mechanics and the sweep-side ledger reconciliation are unchanged, so the Spin model in src/formal/escalation still corresponds to the code, and the formal-verify (escalation) job ran on this PR and passed. A failed audit retains needs-human without consuming the single reviewer pass, consistent with the modeled one-pass ladder.
4. No double audit: recommend-close's direct gh pr comment is replaced (not duplicated) by the single attributed review, the [reviewer] recommend close: marker is preserved in the review body, and queue exclusion is label-based (ReviewerPassedLabel / ReviewerRecommendCloseLabel), so moving the marker from an issue comment to a review breaks no consumer. The review is submitted server-side by the watcher, so no agent-side generic write audit duplicates it.

One note for the merge queue: the red test (rest 2/3) check is TestStaleGeneration_RevokedWorkerCannotOverwriteNewOwner in pkg/dashboard/contribute_lease_test.go, which shares no code with this diff (scheduler kick text plus bin scripts). It needs a rerun before merge.

Issue #5632, opened 2026-09-02
ApplyPack reports created:1 on every steady-state apply, silently reverting operator governor cadences


hanthor:
## Summary

On a long-running L6 spoke, **every operator-set governor cadence is silently
reverted to the pack default**, repeatedly, because `ApplyPack` reports
`created: 1` on *every* apply — which makes `isFirstApplyOrExpansion` true
forever and re-asserts `pack.Governor.Cadences`.

The agent being "created" each time is `brainstorm`, which ships **disabled**
in the L1–L6 packs.

## Observed

Two consecutive applies, ~50s apart, on a hive whose roster had not changed:

```json
{"msg":"ACMM pack applied","created":1,"updated":11,"skipped":0,"tombstoned_agents":null}
{"msg":"ACMM pack applied","created":1,"updated":11,"skipped":1,"tombstoned_agents":null}
```

Each one emitted `ACMM pack changed governor settings` with a full
`cadence_changes` list resetting every mode:

```
{"msg":"ACMM pack changed governor settings","level":6,
 "governor_eval_interval_s_from":300,"governor_eval_interval_s_to":120,
 "cadence_changes":[{"mode":"surge","agent":"ci-maintainer","from":"4h","to":"5m"}, ...]}
```

From the operator's side this is invisible and maddening: you `PUT
/api/config/agent/{name}/cadences`, the write **persists** (it is present in
`/data/hive.yaml.dashboard`), `/api/status` shows the new value, and some
minutes later it is back at the pack default. The persisted overlay and the
in-memory config disagree, with nothing in the log tying the revert to the
cadence you set.

Notably `telemetry` and `operations` revert to `cadence: paused`, so a hive can
sit with two agents that never get kicked while showing `paused: false`.

## Workaround (confirmed)

Tombstoning the disabled agent stops it:

```
DELETE /api/config/governor/agents/brainstorm
→ {"status":"removed","tombstoned":true}
```

After that, applies report `created: 0`, and operator cadences survive a pod
restart. This is a workaround, not a fix — it requires deleting an agent the
pack legitimately defines, and it only occurred to us after reading the source.

## What we could NOT determine

We did not manage to pin the exact reason `created` is non-empty on a
steady-state apply, and we would rather report the behaviour accurately than
guess:

- `AddAgent` registers into `m.agents` regardless of `Enabled`, and
  `GetStatus` succeeds for anything in that map — so the
[...]

Issue #2439, opened 2026-08-03
Agent configuration resets to defaults


castrojo:
This happens on an interval - my agent config reverts to the stock config. I go to turn off/remove guide/brainstorm, which is successful in the UI and they show up as removed and everything works fine. 

It's only later where it reverts. 

<img width="1969" height="542" alt="Image" src="https://github.com/user-attachments/assets/f95446b8-e5a2-4da8-80f9-ea17eccc129c" />

PR #5633, opened 2026-09-02
observability(packs): log which agents a pack apply created

hanthor:
### What

Adds `created_agents` to the `ACMM pack applied` log line, alongside the
`tombstoned_agents` / `skipped_agents` name lists that #2439 already added.

Log-only. No behaviour change.

### Why

`created` is the most consequential of the three counts: it is what sets
`isFirstApplyOrExpansion`, which in turn re-asserts `pack.Governor.Cadences`
over whatever the operator configured.

That means a steady-state apply which unexpectedly reports `created: 1`
silently resets every custom cadence in every mode — and the log tells you the
count that authorised the reset, but not *which agent* it was:

```json
{"msg":"ACMM pack applied","created":1,"updated":11,"skipped":0,"tombstoned_agents":null}
{"msg":"ACMM pack changed governor settings","level":6,
 "cadence_changes":[{"mode":"surge","agent":"ci-maintainer","from":"4h","to":"5m"}, ...]}
```

From the operator's side the symptom is that cadence writes persist to
`hive.yaml.dashboard`, show up in `/api/status`, and then revert minutes later
with nothing connecting the two events. Working out which agent was behind it
took a long session of source reading; this one field answers it directly.

Reported in more detail as #5632 — that issue is still open, because we could
not pin the root cause well enough to justify a behavioural patch. This PR is
deliberately scoped to the observability gap that made it hard to find, and
nothing else.

### Verification

- `go build ./...` clean
- `go test ./pkg/dashboard/ -run Pack` passes

Signed-off-by included for DCO.

### DIFF
--- a/src/pkg/dashboard/api_packs.go
+++ b/src/pkg/dashboard/api_packs.go
@@ -394,7 +394,14 @@ func (s *Server) applyPack(level int, forceGovernor bool) (*ApplyPackResult, err
 	// field report, so also list WHICH agents were tombstoned (deleted by the operator
 	// and NOT re-created) vs skipped (already present) — a grep by agent name against
 	// this single line answers "did the pack try to re-add the agent I removed?".
-	s.logger.Info("ACMM pack applied", "hive_id", s.deps.Config.HiveID, "level", level, "name", pack.Name, "created", len(created), "updated", len(updated), "skipped", len(skipped), "paused", len(paused), "resumed", len(resumed), "tombstoned", len(tombstoned), "gate_removed", len(removedByGate), "tombstoned_agents", tombstoned, "skipped_agents", skipped)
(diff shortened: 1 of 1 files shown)


clubanderson:
Reviewed the log fields. Approving — OK to merge.

Verified: `created` is `[]string` of agent names only (built from `pa.Name` in `applyPack`), the same shape as the `tombstoned_agents` / `skipped_agents` lists the line already carries — no tokens, credentials, or config values can reach the field. Log-only, no behavior change, and it directly answers the `isFirstApplyOrExpansion` cadence-reset attribution question from the linked issue. The comment explaining why the name list matters is accurate to the code (`len(created) > 0` feeds `isFirstApplyOrExpansion` at api_packs.go:298).

PR #5649, opened 2026-09-02
✨ feature: hivectl tui — authenticate with a per-user session, and say so when rejected

### LINKED ISSUE #5645: ✨ feature: hivectl tui — open a hub-hosted hive or a per-user-session spoke
## Summary

A Hive can be reached two ways: self-hosted, where the operator sets a shared
dashboard token and clients send it as a bearer credential; or fronted by a
Hive hub, which authenticates a *person* and injects their identity. Some
spokes are configured for per-user sessions rather than bearer tokens at all.
`hivectl tui` is the terminal dashboard shipped by
[#4907](https://github.com/kubestellar/hive/issues/4907).

The TUI reads exactly two environment variables at startup —
`HIVE_DASHBOARD_URL` and `HIVE_DASHBOARD_TOKEN` — and builds one bearer-auth
client from them. So it can only ever open a self-hosted hive that accepts a
shared token. An operator whose hive is hub-hosted, or is a direct-route spoke
with `authorized_users` set, has no way to open the TUI at all: there is no
flag, no config file, and no login prompt to give it a session with.

This issue asks for a login path so those operators can use the TUI.

Blast radius: `src/pkg/tui/` and its client construction; the non-interactive
`hivectl` subcommands already have their own `--server` / `--token-env`
plumbing and are not in scope.

## Why this is filed now

Hub-hosted hives and OAuth login flows were a declared non-goal of #4907,
recorded in `src/docs/design/tui.md` §1 and in `src/docs/hivectl.md` under
"v1 boundaries" ("Self-hosted, token auth only … there is no hub OAuth login
flow"). #4907 is closed and no follow-up was filed. This is that follow-up,
not a regression report.

## How to observe it

With a hub-hosted hive, or a spoke configured with `authorized_users` and
`hub_proxied: false`:

```
export HIVE_DASHBOARD_TOKEN=...   # whatever token you have
hivectl tui
```

The panes stay at `waiting for data`, because every read is rejected. There is
nothing the operator can type to recover — the TUI has no credential input.
`src/docs/hivectl.md` already documents the underlying rule for the
non-interactive commands: "Bearer auth is rejected on direct-route spokes
(`authorized_users` set with `hub_proxied: false`), which require a per-user
session instead."

## Current behaviour

`newTUICommand` (`src/pkg/hivectl/commands/tui.go`) calls `tui.Run()` with
nothing from `*commandEnv`. The TUI therefore reads `HIVE_DASHBOARD_URL`
(default `http://localhost:3001`) and `HIVE_DASHBOARD_TOKEN` directly and does
**not** honour the root `--server` / `--token-env` flags every other `hivectl`
command reads. That divergence is deliberate and documented — it implements the
epic's "Data source" decision verbatim, and `src/docs/design/tui.md`
("Delivery: `hivectl tui`") records the correction — so it should be revisited
as part of this work rather than treated as a bug to quietly fix.

## What is missing, and the decision to make

Two independent gaps, and it is worth deciding whether they are one issue or
two:

1. **Session-based auth.** The client can only send a bearer token. It needs to
   be able to carry a session cookie or a hub-injected identity, which is what
   the direct-route spokes and hub-fronted dashboards actually accept.
2. **Acquiring the credential.** A terminal program cannot run a browser
   redirect in-frame. The realistic options, in rough order of effort:
   - accept an existing session cookie from the environment or a config file,
     leaving acquisition to the operator or to another `hivectl` command;
   - add a `hivectl login` that performs the device/OAuth flow once and caches
     the result for every subcommand, the TUI included;
   - open the system browser from the TUI and wait for a loopback callback.

The first is the smallest thing that unblocks anyone and does not foreclose the
others. Picking among them is the substance of this issue.

Whichever is chosen, the `--server` / `--token-env` divergence above should be
resolved at the same time, in one direction or the other, and
`src/docs/hivectl.md` and `src/docs/design/tui.md` updated to match — those two
pages currently document the divergence as intentional.

## Out of scope

- Changing how the dashboard or hub authenticates anyone. This is a client
  change.
- The non-interactive `hivectl` subcommands' existing flag behaviour, except
  where a shared credential store makes them strictly better off.

- `hivectl tui` opens against a hub-hosted hive and against a direct-route
  spoke with `authorized_users` set.
- A credential that is absent or rejected produces a clear message naming what
  is needed, not four panes stuck at `waiting for data`.
- Tests run against `httptest` fixture servers; no hub, browser, or network in
  CI, per the epic's testing convention.
- `src/docs/hivectl.md` "v1 boundaries" and "Launch and endpoint selection" are
  updated to describe what actually ships.

Danathar:
## Summary

`hivectl tui` is the full-screen terminal dashboard for a running hive. It authenticated one way — the shared dashboard token, as an `Authorization: Bearer` header — and that is a credential two kinds of hive do not accept, so their operators could not open the TUI at all: hub-hosted hives, and spokes configured with an `authorized_users` allowlist. On those, every read came back `401`, the poll loop swallowed it by design, and the TUI opened onto four panes reading `waiting for data` — a screen that says the hive is silent when what happened is that the operator was not let in, with no keybinding that could fix it.

This adds the missing credential lane and stops the silent failure:

- **`HIVE_DASHBOARD_COOKIE`** carries a per-user session. It is sent alongside the token, on both the polling requests and the SSE stream.
- **Startup probes once, before the alternate screen.** On `401` — and only `401` — it prints which variables to set and exits, leaving the operator's scrollback intact.

Blast radius: `src/pkg/tui` only. **No server-side change** — both credential lanes already exist and are untouched. A hive that works today keeps working with nothing new set: the cookie header is omitted entirely when the variable is empty.

### Why a cookie, and not just another token

The dashboard has two credential lanes and they are not interchangeable (`pkg/dashboard/server.go`, `authenticate`). The Bearer lane is a **shared** secret that grants unscoped owner, and it is switched off on a direct-route spoke precisely because it carries no per-user identity — the code says so directly:

> On a direct-route spoke it is DISABLED: the shared token grants no per-user identity, so accepting it would let any holder act as an unscoped owner and defeat the per-hive allowlist. Direct-route callers must use a per-user session instead.

The other lane is the `hive_session` cookie the GitHub device flow mints, resolved on every request against the *live* allowlist (`liveSessionRole`). A hub-fronted hive is the same story with the hub's own cookie. Neither is reachable through an `Authorization` header, so a client that can only set that header is locked out of both — no matter how correct its secret is.

### The value is a cookie header, not a session id

`HIVE_DASHBOARD_COOKIE=hive_session=8f3c…`, several joined with `; `. That is deliberately the same string a browser sends, which buys three things: an operator can lift it verbatim from a dashboard session they already have; one variable covers a spoke's `hive_session`, a hub's `hive_hub_user`, and the per-hive terminal assertion that rides alongside it; and `pkg/tui/client` never has to enumerate cookie names or work out which kind of hive it is talking to.

### Both credentials are sent when both are set

Not belt-and-braces. Which lane a hive honours is a property of how it was deployed, and the client cannot see that from outside — a shared-token spoke ignores the cookie, an allowlist spoke ignores the token. Making the operator work out which of their two credentials this particular hive wants would be asking them to know something the server already knows, and it means one exported environment works against every hive they operate.

`authorize()` is now the single place this package decides how a request proves who it is, shared by `doJSON` and `streamEvents`. That consolidation is the point: a stream that authenticated differently from the polls produces the worst available symptom — panes filling while the header reads `not connected`, or the reverse.

### Only 401 is fatal

A `401` is a *standing* answer. It will be identical on the next tick and on every endpoint, and nothing in the TUI can change it, so opening anyway guarantees a dead screen. Everything else still opens exactly as before:

| Outcome | Behaviour | Why |
|---|---|---|
| `401` | refuse, name both variables | standing; no in-TUI recovery |
| `403` | open | a working session whose role is too narrow for *some* reads; panes already handle this individually |
| unreachable / `5xx` | open | transient, and the poll loop is built to fill the panes when the hive returns |

That last row is deliberate rather than an omission: a hive being down is one of the main reasons to open the TUI, so refusing there would make the tool unavailable during the exact incident it exists for. The probe reads `/api/status` — the endpoint the Governor pane already polls every cycle — so a probe that passes predicts a frame that fills, and it is bounded at 2s, shorter than the client's own request timeout, because it runs while the operator is looking at a blank terminal.

The refusal names **both** variables. Naming only the token, the historical default, would send an operator of an allowlist spoke to regenerate a secret that endpoint will never accept.

### What this does not do

Acquiring the cookie still means copying it out of a browser already logged in to that dashboard. A `hivectl login` that runs the device flow from the terminal and caches the result for every subcommand is filed separately as #5651 — the server endpoints for it already exist and are untouched here. #5645 listed three acquisition options and called this one "the smallest thing that unblocks anyone and does not foreclose the others"; that is the one this PR takes, which is why it closes #5645 rather than sitting under it.

The `--server` / `--token-env` divergence #5645 also raised is deliberately **not** touched here. It implements the epic's Data source decision verbatim and is documented as intentional in two places; changing it is a separate call, not a side effect of adding a credential.

## Related issues

Fixes #5645

Follow-up filed: #5651 — `hivectl login`, so the session can be obtained from
the terminal rather than copied out of a browser.

## Testing

- [x] `cd src && go test ./pkg/tui/... ./pkg/hivectl/... ./pkg/config/...` — including `-count=2`
- [x] Other / not run (explain): `-race` could not run in this environment (it requires cgo, and the sandbox's compiler cache is read-only). CI covers it. The full `go test ./...` was likewise not run locally; the change is confined to `src/pkg/tui` plus docs, and `pkg/config`'s `env_vars_doc_parity_test.go` — the one test that reads `docs/env-vars.md` — passes.

New coverage, all against `httptest` fixture servers with no hive, container, tmux or network, per the epic's testing convention:

- `pkg/tui/client/credentials_test.go` — the cookie goes out verbatim; token and cookie coexist; an empty cookie is *omitted* rather than sent empty (the same lesson the token path already learned); the **SSE stream** sends it too, which is what stops the two request paths drifting; `IsUnauthorized` splits 401 from 403; the probe reads `/api/status` and returns an unwrapped `APIError` so the caller can still classify it.
- `pkg/tui/preflight_test.go` — 401 refuses and names both variables; 403, `502`, and an unreachable dashboard all open; and one end-to-end case proving the value travels from `HIVE_DASHBOARD_COOKIE` through `client.New()` and the model onto the wire, which no client-package test can see.
- `poll_test.go`'s `TestMain` now pins `HIVE_DASHBOARD_COOKIE` empty for the package, for the same reason it already pins the URL and token: a developer with a real session exported would otherwise hand it to every fixture server in the suite.

## Contributor checklist

- [x] Docs, examples, and policies are updated when behavior changes — `src/docs/hivectl.md` gains a **Credentials** section and an updated launch table and v1-boundaries bullet; `src/docs/env-vars.md` documents the new variable, marked client-side-only since the server never reads it. `src/docs/design/tui.md` is deliberately unchanged, per its own convention that it stays the plan rather than being rewritten as things ship.


clubanderson:
Reviewed the auth flow against `pkg/dashboard/server.go` `authenticate`. Approving — OK to merge.

Verified:
- Client-side only: no server change, and the server resolves cookie sessions via `liveSessionRole` against the live allowlist on every request, so nothing here can escalate a role — the shared-token lane already granted unscoped owner to any holder, and sending both credentials adds no lane the server didn't have.
- `authorize()` is the single choke point and is now used by both `doJSON` and the SSE stream (the SSE path previously duplicated the Bearer logic); the drift test `TestStreamEventsSendsCookie` pins it.
- Empty credentials are omitted rather than sent empty, so an unconfigured client is wire-identical to the pre-change one (pinned by `TestAuthorizeOmitsEmptyCookie`).
- No secrets in logs or errors: the cookie/token values are never logged, and the 401 refusal message names only the env-var names.
- Preflight is fatal on 401 only — 403, unreachable, and 5xx all still open the TUI (each pinned by a test); `CheckCredentials` reuses the model's own client so the probe cannot authenticate differently from the polls.
- `TestMain` clears `HIVE_DASHBOARD_COOKIE` so a developer's live session can't leak into fixtures; timeouts are named constants.
- Docs (env-vars.md, hivectl.md credentials table, v1 boundaries) updated consistently with the behavior.

clubanderson (approved):
Reviewed the TUI session-auth path. The client sends HIVE_DASHBOARD_COOKIE consistently on polling and SSE requests, keeps bearer token behavior intact, and the startup preflight only fails closed on 401 while preserving existing down-dashboard/403 behavior. Validation: cd src && go build ./... passed; go test ./pkg/tui/... passed.

PR #5653, opened 2026-09-02
🐛 fix(relay): re-detect claude readiness after a task, and never complete a task the agent was never given

### LINKED ISSUE #5156: 🐛 bug: claude readiness detection matches only splash/bypass artifacts — a healthy local-mode pane classifies as 'starting' forever and the contributor is silently starved

### LINKED ISSUE #5650: 🐛 bug: contributor relay never re-detects Claude readiness after a task ends — the next issue is booked 'completed' with no PR, and the one after it fails 'CLI never became ready'
## Summary

Hive's contributor relay ("clanker") drives one AI CLI in a tmux pane and feeds it
issues from the hive's queue, one at a time. Before typing a task prompt into that
pane it checks the pane actually looks like a live CLI (`getCLIState` in
`bin/contributor-relay.sh`); if the check fails, the prompt is queued and never sent.

On a `just contribute-hive claude local` session this morning (contributor
`Danathar`, claude backend, Claude Code 2.1.236, relay protocol 1.2), three
consecutive tasks were handed to a relay whose CLI was healthy and idle the entire
time. Only the first reached the agent:

| Task | Hub verdict | What actually happened |
| --- | --- | --- |
| #5645 | `completed` 11:55:40Z | Real. PR #5649 opened 11:54:44Z. ✅ |
| #5646 | **`completed` 11:59:44Z** | **Nothing shipped.** No PR exists, the issue has no comments and has not been touched since 11:31:48Z. The relay reported the *previous* task's `HIVE_VERDICT: complete — PR #5649` line, still on the pane. |
| #5644 | **`failed`** 12:05:41Z | `environment` / `CLI never became ready: CLI did not become ready within timeout` — exactly `CLI_READY_TIMEOUT_MS` (10 min) after #5645's completion armed the wait. |

Two distinct defects, one after the other:

1. **Readiness for `claude` is only ever detected on the startup splash**, so after
   the first task of a session the relay can never re-confirm a healthy pane. Every
   subsequent task prompt is silently queued and never typed.
2. **A task the agent was never given can still be reported `completed`**, because
   the completion check reads a `HIVE_VERDICT:` line off a pane that still holds the
   *previous* task's transcript.

Blast radius: interactive local mode on the `claude` backend — the default
contributor path. The operator sees nothing wrong for ten minutes, and the failure
is not merely a wasted slot: it books a completion that shipped nothing against a
real issue, which the hub then treats as done and removes from the queue.

## How to observe it

Run `just contribute-hive claude local`, let it finish one task, and watch the next
two. The pane sits at a normal idle Claude prompt (`⏵⏵ auto mode on (shift+tab to
cycle) · ← for agents`) showing the finished transcript, while the relay logs:

```
CLI not ready — queuing task prompt instead of typing into the pane
```

for every task that follows, then ten minutes later:

```
CLI did not become ready within timeout
Task ct-<repo>-<number>-<ts> failed [environment]: CLI never became ready: CLI did not become ready within timeout
```

Live evidence from the hub, no reconstruction needed:

```
GET /api/contribute/fleet →
  "last_failure": {
    "task_id": "ct-kubestellar/hive-5644-1788350384",
    "number": 5644, "kind": "environment",
    "reason": "CLI never became ready: CLI did not become ready within timeout",
    "at": "2026-09-02T12:05:41Z"
  }

GET /api/contribute/activity →
  11:34:34Z picked up #5645
  11:55:40Z completed #5645          ← real; PR #5649
  11:55:40Z picked up #5646
  11:59:44Z completed #5646          ← ghost: no PR, issue untouched
  11:59:44Z picked up #5644
  12:05:41Z failed    #5644
```

And on the contributor host, the decisive fact: the Claude process **never
restarted**, despite three task exits that each call `relaunchCLI()`.

```
$ ps -o pid,lstart -p 18428
  18428 Wed Sep  2 07:11:59 2026        # 07:11 EDT == 11:11Z — session start
```

`/api/contribute/run-stats` records the ghost completion as
`scenarios.verdict_complete`, indistinguishable from the real one — the hub has no
way to tell them apart.

## Defect 1: `getCLIState` matches only splash artifacts

`bin/contributor-relay.sh`:

```js
if (BACKEND === 'claude') {
  if (/Not logged in|Please run \/login/.test(text)) return 'needs-login';
  if (/bypass permissions|Welcome back|Try "how does|medium.*effort|@gmail\.com|@.*\.com.*Organization/.test(text)) return 'ready';
  if (/Choose the text style|trust this folder/.test(text)) return 'onboarding';
}
```

Every `ready` alternative is a startup or post-login artifact. A steady-state
local-mode pane — sandbox + `--permission-mode dontAsk`, transcript on screen, idle
`❯` prompt, `⏵⏵ auto mode on (shift+tab to cycle) · ← for agents` in the footer —
matches none of them, so it classifies `starting` forever.

That matters because `cliReady` is cleared on **every** task exit:
`stopAgentForTaskExit()` sets `cliReady = false` and re-arms `armCLIReadyWait()`.
Recovery then depends on a fresh splash appearing, which requires the CLI to
actually exit and relaunch — and `quitLiveCLI()`'s two `C-c` keystrokes did not end
Claude on any of the three exits above (see the unchanged PID). So the latch never
re-latches for the rest of the session.

The relay's *other* Claude detector already knows what a live pane looks like:

```js
hasIdlePrompt = /⏵⏵|← for agents|bypass permissions|shift\+tab to cycle/.test(claudeTail);
```

Two detectors, same pane, opposite verdicts.

**This is #5156, still open in effect.** #5156 named this exact function, this exact
line, and proposed adding the persistent chrome to the ready alternation. It was
auto-closed by PR #5170, which was really the #5162 fix: #5170 changed
`classifyTmuxPane`'s `hasIdlePrompt` (the line quoted just above) and left
`getCLIState` byte-for-byte unchanged. Confirmed on `origin/v4` at `a8619b0`, the
SHA this hive is currently serving.

## Defect 2: a task the agent never received can be reported completed

`progressTick()` starts judging the pane `TASK_GRACE_PERIOD_MS` (3 min) after
assignment. Nothing in that path asks whether the prompt was ever *delivered* —
`pendingTask` is not consulted — and `detectCompletionVerdict()` has no notion of
which task a verdict line belongs to. So a pane holding the previous task's
transcript, with `HIVE_VERDICT: complete — PR #5649` still visible, satisfies
`verdictCompletes` for the new task. #5646 was booked complete 4 minutes after
assignment, which is the 3-minute grace plus one 2-minute tick.

The guard added in #5376 is aimed the other way (chrome idle must repeat before it
may end a task); a verdict short-circuits it deliberately. Neither guard covers a
*stale* verdict, and the code comment above `verdictCompletes` says a stale verdict
"must not launder that into a success" — but the exclusion it implements applies
only to the three API-error pane states.

## Suggested fix shape

1. **`getCLIState`, claude branch:** add the persistent chrome to the `ready`
   alternation — `⏵⏵`, `← for agents`, `shift+tab to cycle`, `esc to interrupt` —
   keeping the `needs-login` / `onboarding` checks ordered first, per the bob and
   codex comments in the same function. This is #5156's stated fix, applied to the
   function it was written about. Readiness asks "is the CLI up and past its gates",
   not "is it idle"; busy-vs-idle is `classifyTmuxPane`'s job and `tmuxSendKeys`
   separately guards against a bare shell.
2. **Never judge a task the agent has not been given:** if `pendingTask` is still
   holding this task's prompt, `progressTick` must not read the pane for completion.
   Report `working` (or hand the task back early with a distinct reason) instead of
   inferring an outcome from a transcript the task never touched.
3. **Make the verdict per-task:** record a pane fingerprint or line offset at
   prompt-delivery time and accept only a `HIVE_VERDICT:` line that appears after
   it. A verdict emitted before the task existed cannot be about the task.
4. Related hazard worth checking while in here (#2203 shape, not directly observed
   in this incident): when `quitLiveCLI()` fails to exit the CLI, `relaunchCLI()`
   types `cd <dir> && claude …` into a pane whose foreground program is still the
   agent — i.e. as a chat prompt, into an agent with a live scoped GitHub token.

Regression coverage in `bin/contributor-relay.test.js`: a realistic steady-state
local-mode Claude pane (finished transcript + `❯` + the auto-mode footer) pinned as
`ready` by `getCLIState`, and a task whose prompt is still queued pinned as *not*
completable even when a prior `HIVE_VERDICT: complete` line is on the pane. Today's
fixtures only exercise the copilot-style pane, which is why neither defect failed a
test.

Related: #5156 (same root cause, closed by a PR that fixed a different detector),
#5170, #5162, #5376, #5353, #2203.

Danathar:
## Summary

On an interactive `just contribute-hive claude local` session, the relay handed three tasks in a row to a Claude CLI that was healthy and idle the whole time. Only the first one actually reached the agent. The second was reported **`completed` with no PR and an untouched issue**; the third failed ten minutes later with `CLI never became ready`.

The cause is two separate bugs firing back to back:

1. **The relay could only recognise a live Claude on its startup splash screen.** Once the splash scrolled away — i.e. after the very first task of a session — a perfectly normal idle pane looked to the relay like a CLI that had not started yet. Every task prompt after the first was quietly put in a queue instead of being typed, and each of those tasks was eventually handed back with "CLI never became ready".
2. **A task the agent was never told about could still be booked as done.** The relay reads the pane to decide whether a task finished. A task whose prompt was still sitting in that queue got judged against the *previous* task's finished transcript, `HIVE_VERDICT: complete` line and all — so the hub recorded a completion, closed the work item out of the queue, and nothing had been shipped.

This PR fixes both, and adds the regression tests that would have caught them.

Fixes #5650. This also finally applies the fix #5156 asked for: #5156 named `getCLIState` and this exact line, and was auto-closed by #5170, which changed a *different* detector (`classifyTmuxPane`'s `hasIdlePrompt`) and left `getCLIState` byte-for-byte unchanged.

---

## Defect 1 — `getCLIState` matched only splash artifacts

Every `ready` alternative in the claude branch was a startup or post-login artifact: the welcome banner, the account line printed just after login, the first-run tip. That made readiness a one-shot property of the splash screen.

`cliReady` is a latch cleared on **every** task exit (`stopAgentForTaskExit`), and re-latched only from `getCLIState`. Re-latching therefore depended on a fresh splash appearing, which requires the CLI to actually exit and relaunch — and `quitLiveCLI()`'s two `C-c` keystrokes did not end Claude on any of the three exits in the incident (the PID never changed). So the latch never re-latched for the rest of the session.

**Fix:** the claude branch now also accepts the footer chrome a live Claude Code draws at *all* times — `⏵⏵`, `← for agents`, `shift+tab to cycle`, `esc to interrupt`. That is the same evidence `classifyTmuxPane`'s own claude `hasIdlePrompt` has used since #5170; two detectors reading one pane must not disagree about whether the CLI is even there.

The `needs-login` and `onboarding` gates are now classified **first**, matching the ordering the `bob` and `codex` branches already document, so a pane blocked on login or a trust dialog is still never reported ready even though it draws the footer chrome behind the modal.

Readiness deliberately stays a different question from idleness — `esc to interrupt` (drawn mid-turn) counts as ready. "Is the CLI up and past its gates" is what this function answers; busy-vs-idle is `classifyTmuxPane`'s job, and `tmuxSendKeys` separately refuses to type into a pane whose foreground command is a shell.

## Defect 2 — a task the agent never received could be reported completed

Two independent guards, because there are two independent holes:

**a. Never judge an undelivered task.** `tmuxSendKeys()` queues rather than types whenever the CLI is not confirmed ready (or the pane has fallen back to a shell). `progressTick()` never asked whether that had happened, so it judged a pane holding only the previous task's transcript. It now reports `working` and returns instead.

Reporting progress rather than failing is deliberate: `armCLIReadyWait()` already owns this case and hands the task back with a real reason at `CLI_READY_TIMEOUT_MS`, and the max-duration lease is a second backstop — deliberately *not* renewed by this path, since a pane the agent was never prompted with is not forward progress.

**b. Make the verdict per-task.** `detectCompletionVerdict()` had no notion of which task a `HIVE_VERDICT:` line belonged to. The relay now records the verdict line that was already on the pane at the moment the prompt was typed, and refuses a verdict identical to it — a line printed before the task existed cannot be about the task. `detectHiveVerdict` returns the raw pane line it matched to make that comparison possible.

A suppressed verdict does not strand the task: with no verdict, the existing chrome-idle grace window (#5376) becomes the signal, exactly as it is for an agent that never prints the sentinel. There is a test for that.

## Tests

`bin/contributor-relay.test.js`, 8 new cases (274/275 pass locally; the one failure, `a relaunch cds somewhere resolvable before starting the CLI`, is pre-existing on `v4` and is caused by `HIVE_AGENT_CWD` leaking from the contributor host's environment into the test process — verified by running the suite on the unmodified base commit):

- a realistic steady-state local-mode claude pane (finished transcript + `❯` + the auto-mode footer) is `ready`
- the two claude detectors agree that the CLI is there for that same pane
- a busy claude pane is `ready` too, and still `WORKING` to the classifier
- the login and trust gates still win over the persistent chrome
- a task whose prompt is still queued is never completed off the pane
- the previous task's verdict does not complete the next task
- the agent's own fresh verdict still completes on the first tick, with its PR
- a stale verdict does not block the chrome-idle fallback

Existing fixture-driven completion tests move to a `dispatchTask()` helper. The harness serves one static `paneText` for every capture, so a fixture showing a *finished* turn is also what the relay sees at the instant it dispatches — something a real pane cannot do, since the agent's verdict is printed after it runs. The helper states the two facts those fixtures mean: the prompt was typed, and the pane held no earlier verdict when it was.

The `#4267 detectNoWorkVerdict` assertions gain the new `line` field.

## Not addressed here

Item 4 of the issue — `relaunchCLI()` typing `cd <dir> && claude …` into a pane whose foreground program is still the agent when `quitLiveCLI()` fails to exit it — is confirmed present in the code (`relaunchCLI` sends the launch command with no foreground-command check) but was not observed in this incident. Guarding it changes the recovery path's failure mode in ways I cannot measure from here, so it is left for a follow-up rather than shipped unverified alongside this fix. Defect 1's fix does reduce its blast radius: the relay now re-detects the surviving CLI and flushes the queued prompt to it, so a `quitLiveCLI()` that fails to kill Claude is no longer session-fatal.


clubanderson:
Reviewed for correctness of the readiness re-detection state machine against the surrounding relay code. Approving — OK to merge.

Verified:
- `getCLIState` claude branch: login/trust gates are classified before both ready alternations, so a blocked pane drawing footer chrome is never reported ready (this also reorders `onboarding` ahead of the splash-ready alternation — a strictly safer ordering, matching the bob/codex branches).
- The footer-chrome alternation is the same evidence `classifyTmuxPane`'s claude `hasIdlePrompt` already uses, so the two detectors can no longer disagree about whether the CLI exists.
- `taskPromptDelivered` is cleared at the top of `tmuxSendKeys()` — before every queue/early-return path — and set only after an actual send, so a queued prompt can never be judged. `flushPendingTask()` re-enters `tmuxSendKeys()`, so delivery and the baseline snapshot happen at real delivery time.
- The `deliveredVerdictBaseline` snapshot sits in the send path after the queue early-return, which also correctly covers the post-completion `reviewPrompt` dispatch (whose pane necessarily still shows the prior task's verdict — that path was a latent ghost-completion too).
- `detectHiveVerdict` scans newest-first, so when a genuine new verdict appears below the stale one, the new line wins, differs from the baseline, and completes on the verdict signal (pinned by the fresh-verdict test).
- The undelivered-task guard exempts headless mode, reports `working`, and deliberately leaves the max-duration lease unrenewed; `armCLIReadyWait()` remains the path that hands the task back with a real reason.
- Regression tests assert the real invariants: zero `task_complete` for a queued task, the task retained, stale-verdict suppression falling through to chrome-idle rather than stranding.

One accepted residual (documented in the PR): a new task whose verdict line is byte-identical to the previous one degrades to the chrome-idle fallback instead of the verdict signal — safe direction.

Issue #5652, opened 2026-09-02
🐛 bug(security): the relay's post-task relaunch drops local mode's sandbox — a local clanker comes back as --dangerously-skip-permissions


Danathar:
## Summary

`just contribute-hive <backend> local` runs a contributor's AI agent **directly on
their machine**, with no container around it. Because there is no outer boundary,
the launcher deliberately narrows the CLI's own posture: for Claude that means
`--permission-mode dontAsk`, Claude Code's native sandbox with
`allowUnsandboxedCommands: false`, a filesystem write-allowlist limited to
`$HIVE_WORKSPACE_DIR`, and `--add-dir` scoped to the same. That posture is what
#4918 and #5011 exist to establish, and hive documents an explicit opt-out
(`HIVE_CLAUDE_DANGEROUSLY_BYPASS_APPROVALS_AND_SANDBOX`) for operators who want the
old unconfined behaviour back.

**The relay throws that away the first time a task ends.** After every task the
relay restarts the CLI itself (`stopAgentForTaskExit()` → `relaunchCLI()`), and it
rebuilds the launch line from scratch — from the **container** entry point in
`config/backends.conf`, which nothing tells about local mode. So the agent is
relaunched as:

```
claude --dangerously-skip-permissions --permission-mode bypassPermissions --disallowed-tools …
```

No sandbox. No write-allowlist. No `--add-dir` bound. On a host with nothing else
standing between the agent and everything the operator's user can reach — and the
operator never set the escape hatch. This is precisely the pre-#4918 posture,
reinstated silently after the first task of every local-mode session, for the rest
of that session.

Blast radius: local mode (`just contribute-hive … local`) on the backends whose
local posture is *narrower* than their container posture — **claude/litellm,
copilot, opencode**. Container mode is unaffected (its posture is the one being
rebuilt, so there is no divergence). No hub or spoke agent is involved.

## Evidence (live session, 2026-09-02)

The CLI as the Justfile launched it — sandbox posture, `Justfile:1020` →
`claude_family_local_perm_flag_shell`:

```
$ ps -o args -p 18428
claude --permission-mode dontAsk
  --settings {"permissions":{"allow":["Edit(/var/home/dbaggett/workspace/**)",
              "Write(/var/home/dbaggett/workspace/**)"]},
              "sandbox":{"enabled":true,"failIfUnavailable":true,
              "autoAllowBashIfSandboxed":true,"allowUnsandboxedCommands":false,
[...]

PR #5660, opened 2026-09-02
🔒 test(contribute): pin the #5652 local-relaunch posture invariant from both sides

clubanderson:
Follow-up coverage for #5652 (fixed by #5702).

This PR originally carried an independent fix for the relay's post-task relaunch dropping local mode's sandbox; #5702 merged the same single-source-of-truth fix first (the launcher exports the resolved line as `AGENT_LAUNCH_CMD`, `buildLaunchCommand()` prefers it). Rebased onto v4, this PR now adds only the regression coverage the issue asked for beyond #5702's positive-path test:

**Relay negative controls** (`bin/contributor-relay.test.js`):
- A sandbox-off (explicitly opted-out) launch relaunches unchanged — the fix pins launch == relaunch, it must not invent a sandbox the operator declined.
- Without an exported line, the `backends.conf` derivation still applies (container mode and older launchers, where derived posture == launched posture).
- The exported line is the WHOLE command: `AGENT_MODEL` is never appended onto it — the same two-derivations drift in the other direction.

**Cross-file contract** (`src/pkg/dashboard/contribute_local_relaunch_posture_test.go`, new):
- The relay reads the exact variable the Justfile exports. A rename on either side would fail silently at runtime — the env var would simply be absent and the permissive fallback would win again — and neither file's own tests can catch it, since each side stays self-consistent.
- `buildLaunchCommand()` itself consults the entrypoint line (the single choke point every relaunch path goes through), not merely somewhere in the file.
- No local-mode `send-keys` line interpolates `$PERM_FLAG` directly again, bypassing the exported single source.

Relay suite 289/289 green; `node --check` clean on relay + tests; `gofmt`/`go vet` clean; targeted dashboard tests green.

### DIFF
--- a/bin/contributor-relay.test.js
+++ b/bin/contributor-relay.test.js
@@ -435,6 +435,81 @@ test('relaunchCLI sends the cd-prefixed command to tmux', () => {
   } finally { teardown(relay); }
 });
 
+// ---------------------------------------------------------------------------
+// kubestellar/hive#5652, remaining edges — a relaunch must reuse the
+// LAUNCHER's resolved launch line, never re-derive one that drops local
+// mode's sandbox.
+//
+// The positive half (an exported AGENT_LAUNCH_CMD wins over the container
+// posture for claude/copilot/opencode, byte-identical) is pinned by
+// "#5652 relaunch reuses the entrypoint launch command" above. These are the
+// negative controls: the fix pins launch == relaunch, so it must neither
+// invent a sandbox the operator explicitly opted out of, nor append flags the
+// launcher deliberately omitted, nor break the container/older-launcher path
+// where nothing is exported and the derived posture IS the launched posture.
+// backendPerm is the container bypass string on purpose: the assertion is
+// precisely about when it may and may not win.
+// ---------------------------------------------------------------------------
+
+// A faithful reduction of what claude_family_local_perm_flag_shell emits.
+const LOCAL_SANDBOXED_CLAUDE_CMD = 'claude --permission-mode dontAsk ' +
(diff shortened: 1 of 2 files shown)


clubanderson (commented):
Independent verification against the v4 source (not the dispatch report). **Verdict: APPROVE — safe to merge.**

Verified:
- Single source of truth holds: every relaunch path (stale readiness latch, CLI restarts, post-task, revoke — relay lines 1665/1720/2657/3277/3285/3449 all funnel through relaunchCLI → buildLaunchCommand) now prefers the verbatim HIVE_AGENT_LAUNCH_CMD, and the Justfile builds/types/exports one LOCAL_LAUNCH_CMD variable rather than two constructions.
- The export happens after tmux session creation, so the variable reaches only the relay, never the agent's own environment.
- Fallback preserved: container mode exports nothing, the backends.conf derivation still applies (pinned by a negative-control test); the documented opt-out relaunches as launched, with no sandbox invented.
- Security tests genuinely assert the invariant and are red-before-fix by construction: backendPerm is deliberately the container bypass string, so the pre-fix buildLaunchCommand() returns the bypass line and both the strictEqual and the tmux-send assertions fail. Negative assertions on dangerously-skip-permissions/bypassPermissions pin the exact regression, and the Go source-pin tests guard both sides of the env-var contract against silent rename.
- litellm's ANTHROPIC_BASE_URL env prefix rides inside LOCAL_LAUNCH_CMD, so relaunch restores it too.

Non-blocking note: buildHeadlessArgv (opt-in CONTRIBUTOR_MODE=headless) still derives perm flags from backends.conf; a local-mode operator who also sets that variable would get container-posture argv. Out of #5652's scope (interactive relaunch), but worth a follow-up.

PR #5663, opened 2026-09-02
✨ feat(deploy): make hivectl obtainable on hosts installed with hive-podman-setup.sh

### LINKED ISSUE #5646: ✨ feature: make hivectl obtainable on hosts installed with hive-podman-setup.sh
## Summary

`hivectl` is the operator's command-line client for a running Hive — it
inspects and drives the fleet over the dashboard API, and as of
[#4907](https://github.com/kubestellar/hive/issues/4907) it also carries
`hivectl tui`, the full-screen terminal dashboard. The only documented way to
obtain it is to compile it: `go build -o bin/hivectl ./cmd/hivectl`.

The supported Podman install (`bin/hive-podman-setup.sh`) deliberately installs
no packages, because its target hosts include image-based systems — Fedora
Silverblue/Bluefin, RHEL image mode — where `/usr` is read-only and the package
manager cannot do it. A Go toolchain is not among the prerequisites its
preflights check for, and nothing in the install provides one.

So an operator who follows the recommended install on a recommended host ends
up with a running hive and no supported way to get `hivectl` onto the machine.
The entire `hivectl` surface — `system`, `agent`, `bead`, `governor`, `observe`,
and the TUI that #4907 just finished — is unreachable for them unless they
happen to have installed Go for unrelated reasons.

This issue asks for a distribution path. Choosing which one is the substance of
the issue.

Blast radius: packaging and CI. No change to `hivectl`'s behaviour is implied.

## How to observe it

On a clean image-based host, run `bin/hive-podman-setup.sh` to completion, then:

```
hivectl system status
```

`hivectl: command not found`, with no documented next step that does not
require first installing a Go toolchain onto a read-only `/usr`.

## Why the container image is not the answer

`src/Dockerfile` builds exactly three binaries — `hive`, `bd`, `hive-backup` —
and that is correct. `hivectl` is a *client* of the dashboard API, not part of
the runtime the image serves; the whole premise of the TUI, quoted from #4907,
is that it "is not a second implementation of Hive logic. It is another client
of the dashboard API." Putting the client inside the thing it inspects would
invert that.

`podman exec`-ing into the container to run a client that talks to the same
container over loopback is a workaround, not a design, and it is unavailable
anyway: the binary is not in the image.

## Options

None of these is obviously right; the issue is to pick one.

1. **Attach `hivectl` binaries to the GitHub release.** `src/docs/releases.md`
   documents release machinery that already exists (`tagged-release.yml`), but
   it publishes a retagged GHCR *image*; there are no binary artifacts today,
   and that page's "What still blocks cutting a real release" section is a live
   caveat this option inherits.
2. **A thin `hivectl` image**, so `podman run --rm --network=host ghcr.io/…/hivectl tui`
   works with no toolchain. Fits image-based hosts, and matches how those
   operators already run everything else. Needs care with TTY allocation for
   the TUI.
3. **Have `bin/hive-podman-setup.sh` build and place it** during install. This
   contradicts the script's own stated design — it installs no packages and
   modifies no host — so it would need a Go toolchain from somewhere, likely a
   throwaway builder container.
4. **Document the gap and the manual route.** Cheapest, and honest, but leaves
   the recommended install shipping an unusable half of the product.

Option 2 looks like the smallest thing that actually works on the target hosts,
but it is a maintainer call.

## Out of scope

- Any change to `hivectl`'s commands, flags, or output.
- Cutting an actual versioned release, which has its own blocker recorded in
  `src/docs/releases.md`.

- An operator on a host installed by `bin/hive-podman-setup.sh`, with no Go
  toolchain, can obtain and run `hivectl system status` and `hivectl tui`
  by following documentation.
- The route is covered by CI, so it cannot silently rot the way an
  undocumented build step would.
- `src/docs/hivectl.md`'s "Quick start" leads with that route and keeps
  `go build` as the from-source path for contributors.

clubanderson:
## Summary

Closes the gap #5646 describes: the recommended Podman install targets image-based hosts (Fedora Silverblue/Bluefin, RHEL image mode) with no Go toolchain and a read-only `/usr`, so the only documented way to obtain `hivectl` — `go build` — was unavailable exactly where the install is supported. The entire `hivectl` surface, including the #4907 TUI, was unreachable there.

## Mechanism: the image carries the client as cargo; the installer extracts it

- **`src/Dockerfile`** builds `hivectl` statically (`CGO_ENABLED=0` — the binary runs on the *host*, whose libc owes nothing to the image's) and stows it at `/usr/local/share/hive/hivectl`, deliberately **off the container's PATH**. This respects the issue's design constraint: the client is not part of the runtime the image serves and must never run inside the thing it inspects — the image is only the distribution vehicle. The binary is included in the #3760 fresh-rewrite loop so the metacopy-redirect defense covers it too.
- **`bin/hive-podman-setup.sh`** gains step 8/9: `podman create` (nothing executes) + `podman cp` + `podman rm` lifts the binary onto the host — `~/.local/bin` rootless, `/usr/local/bin` rootful (writable `/var/usrlocal` on image-based hosts), `HIVE_SETUP_BIN_DIR` overrides either. The step runs after the health check, so the image is guaranteed present even under `--skip-pull`. Nothing is compiled on the host and nothing new is downloaded, and the client is always the exact version of the running hive. Idempotent: a re-run keeps a matching binary (and reports the keep) and refreshes a differing one, so update-then-re-run keeps the pair in step. An image built before this change is reported loudly but does not fail an otherwise healthy install. The usage text, header contract, and end-of-run summary all document the behaviour.

This is the smallest of the issue's options that actually works on the target hosts: no release machinery (option 1 inherits the live blocker in `src/docs/releases.md`), no second image to publish and keep in version lockstep (option 2), no toolchain on the host (option 3), and the TUI runs natively on the host so there is no TTY-allocation question at all.

## Acceptance criteria from the issue

- An operator on a `hive-podman-setup.sh` host with no Go toolchain gets `hivectl system status` and `hivectl tui` by following documentation — the installer does it for them, and the manual `podman cp` route is documented for any pulled image.
- Covered by CI: `src/deploy/test_hivectl_extraction_contract.sh` (new, wired in v2-ci.yml) pins the Dockerfile stow path, the installer's extraction constant, the static-build requirement, the off-PATH placement, the #3760 rewrite coverage, and the documented path to one another — a rename on any side goes red instead of failing on an operator's machine. `bin/test_hive_podman_setup.sh` grows an extraction section: the installed binary provably comes out of the image (recognisable fake content), create-before-rm ordering, idempotent keep + tamper convergence, the missing-binary warn-not-fail path, rootful-through-sudo, and a new `go` tripwire proving the installer never reaches for a host toolchain.
- `src/docs/hivectl.md` Quick start leads with the extraction route and keeps `go build` as the contributor path.

## Out of scope (per the issue)

No change to `hivectl`'s commands, flags, or output; no versioned release. Possible follow-up: teach `bin/hive-podman-update.sh` to refresh the extracted binary directly (today the update path is update, then re-run the installer, which the failure guidance states).

Fixes #5646


clubanderson (commented):

Verified:
- Static build correctness: the Dockerfile builds /hivectl with CGO_ENABLED=0 like its siblings, and the contract test pins that flag explicitly (load-bearing: the binary execs on the host's libc). The #3760 metacopy fresh-rewrite loop was extended to cover the new cargo path.
- Extraction lifecycle: podman create → cp → rm ordering is correct on all paths out of install_hivectl — the throwaway container is removed on the cp-failure path too (asserted by the contract suite via the CALL_LOG), and the staging dir is cleaned on every branch.
- Idempotency: cmp -s keep path reported as keep, tamper converges back to the image's bytes, both driven by tests.
- Warn-not-fail: FAKE_HIVECTL_CP_FAIL drives the pre-#5646-image case; exit stays 0, the summary says NOT installed, and no half-written binary is left.
- No host compile: go added to the tripwire list, and the fake podman cp plants recognisable content so the test proves the installed file came out of the image, not a build.
- Path drift cannot rot silently: test_hivectl_extraction_contract.sh pins Dockerfile stow path == installer HIVECTL_IMAGE_PATH == docs, and is wired into v2-ci.yml (the src/deploy suite-wiring guard covers the wiring itself).
- Helper functions (as_owner/bad/warn/ok/info) and the 'Hive is running' summary heading all exist where the new code uses them; step renumbering 8→9 is consistent throughout.

Nit (non-blocking): the create-failure branch cleans staging with plain rm -rf while the later branches use as_owner rm -rf — harmless either way since mktemp created it as the invoking user.

PR #5664, opened 2026-09-02
✨ feature: hivectl login — obtain a hive session from the terminal instead of a browser

### LINKED ISSUE #5651: ✨ feature: hivectl login — obtain a hive session from the terminal instead of a browser
## Summary

A Hive dashboard identifies callers two ways. Self-hosted hives accept a shared
bearer token. Hub-hosted hives, and spokes configured with an `authorized_users`
allowlist, accept only a **per-user session** — a `hive_session` cookie minted by
a GitHub device-flow login, resolved on every request against the live
allowlist.

[#5649](https://github.com/kubestellar/hive/pull/5649) taught `hivectl` to
*present* that session: `HIVE_DASHBOARD_COOKIE` carries a cookie header and
`hivectl tui` sends it. What it did not do is let anyone **obtain** one from the
terminal. Today the operator has to open a browser, log in to the dashboard, and
copy the cookie value out of devtools — a step that is awkward the first time and
has to be repeated whenever the session expires, on a tool whose whole premise is
that you never leave the terminal.

This asks for `hivectl login`: run the device flow, cache the resulting session,
and have every `hivectl` subcommand pick it up.

Blast radius: `src/pkg/hivectl` and a new credential cache on disk. No
server-side change — the endpoints already exist and are unmodified.

## Why this is filed now

#5645 laid out three ways to acquire the credential and noted the first —
"accept an existing session cookie … leaving acquisition to the operator or to
another `hivectl` command" — was "the smallest thing that unblocks anyone and
does not foreclose the others." #5649 shipped exactly that and closes #5645.
This is the option it deliberately left for a separate change.

## How to observe it

On a hub-hosted hive, or a spoke with `authorized_users` set:

```
hivectl tui
```

Without a cookie, this now refuses with a message naming
`HIVE_DASHBOARD_COOKIE`. There is no command that produces a value for it. The
only route is a browser: log in to the dashboard, devtools → Application →
Cookies → copy `hive_session`.

## The server surface already exists

The device flow is implemented and every route is on `isPublicPath`'s allowlist
(`strings.HasPrefix(path, "/api/gh-user-auth/")`), so an unauthenticated client
can complete it:

| Route | Handler | Returns |
|---|---|---|
| `POST /api/gh-user-auth/start` | `handleGHUserAuthStart` | `user_code`, `verification_uri`, `expires_in`, `interval` |
| `POST /api/gh-user-auth/poll` | `handleGHUserAuthPoll` | `{status: "complete"｜"error", username, avatar_url}`, and **`Set-Cookie: hive_session=…`** on success |
| `GET /api/gh-user-auth/session` | `handleGHUserAuthSession` | current session |
| `POST /api/gh-user-auth/logout` | `handleGHUserAuthLogout` | — |
| `GET /api/gh-user-auth/status` | `handleGHUserAuthStatus` | `logged_in`, `username`, live `role` |

The session cookie arrives on the **poll response**, so the client's job is to
start the flow, print the user code and verification URI, poll at the returned
`interval`, and capture `Set-Cookie` when `status` reaches `complete`.

The login requests no OAuth scope at all (#1927): the token proves identity and
is deliberately not persisted or used as a write identity, so this asks the
operator for nothing beyond "prove who you are."

## Constraints worth knowing before designing this

- **A spoke runs one device flow at a time.** `s.deviceFlowState` is a single
  server-wide field guarded by `deviceFlowMu`, and `start` overwrites it. Two
  operators logging in concurrently will clobber each other, and `poll` without
  a prior `start` returns `400 no device flow in progress`. The client should
  fail with something better than a bare 400 when it loses that race.
- **A rejected user gets `{status:"error"}`, not a 403.** On an allowlist spoke
  an unauthorized GitHub account is refused *inside* a 200 body, because the
  login page's poll loop only stops on `complete` or `error`. A client that
  classifies on HTTP status alone will spin forever.
- **The role is re-resolved per request, not frozen.** Caching a session is
  caching identity, not capability — a downgrade binds immediately server-side.
- **Sessions expire** (30 days). The cache needs an expiry the client can act
  on, or at minimum a clear message when the cached session stops working.

## What is missing

1. A device-flow client under `src/pkg/hivectl` that drives start → prompt →
   poll → capture cookie.
2. A credential cache, and a decision about where it lives and how it is
   protected. It holds a credential equivalent to a logged-in browser session,
   so file mode matters. Suggested: `$XDG_CONFIG_HOME/hive/` at `0600`, keyed
   by dashboard URL so one operator can hold sessions for several hives.
3. Wiring so every subcommand and the TUI read the cache, with the existing
   `HIVE_DASHBOARD_COOKIE` and `HIVE_DASHBOARD_TOKEN` still taking precedence —
   an explicitly exported credential should always win over a cached one.
4. `hivectl logout`, hitting the existing endpoint and clearing the cache.

## Out of scope

- Any change to the dashboard, the device flow, or the session store.
- The `--server` / `--token-env` divergence (`hivectl tui` honours neither);
  it implements the epic's Data source decision verbatim and is a separate call.
- Browser-based loopback-callback login. The device flow needs no local
  listener and works over SSH, which is the environment this is for.

- `hivectl login` against a hub-hosted hive or an allowlist spoke leaves
  `hivectl system status` and `hivectl tui` working with nothing exported.
- An unauthorized GitHub account produces the server's own explanation, not a
  spin or a generic failure.
- An expired or revoked cached session produces a message naming `hivectl
  login`, not a bare 401.
- The cache is not world-readable, and that is asserted by a test.
- Tests run against `httptest` fixture servers — no GitHub, no browser, no
  network in CI, per the `hivectl tui` epic's testing convention.
- `src/docs/hivectl.md` "Credentials" replaces the copy-it-from-devtools
  instructions with the real flow.

clubanderson:
## Summary

Hub-hosted hives and spokes with an `authorized_users` allowlist accept only a **per-user session**, and until now the only way for a terminal client to hold one was to copy the `hive_session` cookie out of a browser's devtools — awkward the first time, and repeated every 30 days, on a tool whose premise is never leaving the terminal.

This adds `hivectl login` and `hivectl logout` (#5651):

- **`hivectl login`** drives the dashboard's existing GitHub device-flow endpoints (`/api/gh-user-auth/start` → prompt → `/poll`), all already on `isPublicPath`. It prints the one-time code and `https://github.com/login/device`, polls at the server's interval, and captures the session the poll response mints as `Set-Cookie` — that is where the server puts it; there is no separate exchange step.
- **The cache** lives at `$XDG_CONFIG_HOME/hive/sessions.json` (falling back to `os.UserConfigDir`), written through a same-directory temp file + rename so it is **0600 from its first byte**, keyed by normalized dashboard URL so one operator holds sessions for several hives. World-readability is asserted against by test, per the issue's acceptance criteria. The credential is **never printed** — success reports the username and the cache path, nothing else.
- **Every subcommand and the TUI pick it up automatically.** `commandEnv.client()` attaches the session lane to every client the commands build; `hivectl tui` exports the cached session as `HIVE_DASHBOARD_COOKIE` before starting. An explicitly exported `HIVE_DASHBOARD_COOKIE` wins unconditionally over the cache.
- **`hivectl logout`** hits the existing `/api/gh-user-auth/logout` presenting the cached cookie (so the server clears *that* session) and removes the cache entry — even when the hive is unreachable, since the local credential is the part only this command can remove; the server failure is reported, not swallowed.

Blast radius: `src/pkg/hivectl` plus a credential cache on disk. **No server-side change** — the device flow, the session store, and every endpoint are untouched.

## How this composes with #5649

#5649 taught `hivectl tui` to **present** a session via `HIVE_DASHBOARD_COOKIE`; this PR is the acquisition half it deliberately left to #5651, and it is built to land in either order:

- The cached value is a **Cookie header value** ("hive_session=…", several joined with `; `) — exactly the `HIVE_DASHBOARD_COOKIE` convention #5649 defined, so the two sources are interchangeable everywhere they are consumed. On a hosted hive the terminal assertion cookie rides alongside the session automatically.
- The TUI handoff goes **through the environment variable, not through `pkg/tui`**: `hivectl tui` sets `HIVE_DASHBOARD_COOKIE` from the cache (when the operator exported nothing) before `tui.Run()`. This PR touches no file #5649 modifies except `CHANGELOG.md` and `src/docs/hivectl.md` (different sections). Before #5649 merges the export is inert; after it merges the TUI lights up with cached sessions with no further change.
- The cache key normalizes the two tools' loopback defaults (`--server` says `127.0.0.1:3001`, the TUI says `localhost:3001`) so a login under one spelling is found under the other.

`src/docs/env-vars.md` is deliberately untouched: #5649 adds the `HIVE_DASHBOARD_COOKIE` row there, and the parity guard is one-directional (documented ⇒ real), so nothing breaks in either merge order. Once both land, that row's "read by `hivectl tui`" wording can widen to all of hivectl — a one-line doc follow-up.

## Constraints from the issue, handled

- **One device flow per spoke:** a poll answered with the bare 400 ("no device flow in progress") classifies as a named `ErrDeviceFlowConflict` telling the operator another login replaced theirs and to run `hivectl login` again.
- **Rejection arrives inside a 200:** `{status:"error"}` is terminal — the loop stops on the first one (asserted: exactly one poll) and surfaces the **server's own explanation** ("your GitHub account (x) is not authorized…"), never a spin or a generic failure. Nothing is cached on rejection.
- **Role is re-resolved per request:** the cache stores identity only; nothing about role is persisted, and the login requests no OAuth scope (#1927).
- **Sessions expire:** expiry is captured from the `hive_session` Set-Cookie (falling back to the server's 30-day `sessionCookieMaxAge`). An expired or revoked cached session produces a 401 whose message names `hivectl login` — the hint is appended on 401 **only**, never 403, which is a working session whose role is merely too narrow.
- **`slow_down` backs off** by 5s per GitHub's device-flow contract; the whole flow is bounded by the server's `expires_in`.

## Testing

All against `httptest` fixture servers — no GitHub, no browser, no network in CI, per the `hivectl tui` epic's testing convention:

- `pkg/hivectl/session_test.go` — save/load round trip verbatim; **cache file and dir carry no group/other bits** (the acceptance criterion as a test); expired entries return alongside `ErrSessionExpired` (logout still presents them for server-side clearing); corrupt cache is a named error with a remedy, not a fake "not logged in"; `SessionKey` collapses loopback spellings and keeps distinct hives distinct.
- `pkg/hivectl/login_test.go` — full happy path (prompt, pending, complete, both cookies joined, expiry from Max-Age, sleeps at the server's interval); denial stops on the first poll with the server's reason verbatim; the lost-start race names `hivectl login`; `slow_down` backoff; flow expiry with remedy; `complete` without a cookie is an error. Plus the client's session lane: cookie sent verbatim alongside the token, omitted entirely when empty, and the login hint appears on 401 only.
- `pkg/hivectl/commands/login_test.go` — `hivectl login` end to end: prints code/URI/username/cache path and **never the credential** (asserted on both stdout and stderr); denied login caches nothing; a subcommand presents the cached session with nothing exported; exported `HIVE_DASHBOARD_COOKIE` beats the cache; a rejected cached session's 401 names `hivectl login`; logout presents the session to the endpoint, clears the cache, and is a calm no-op with nothing cached; the TUI export honours the cache and the env-var precedence; login/logout registration is pinned. A package `TestMain` pins `XDG_CONFIG_HOME` and `HIVE_DASHBOARD_COOKIE` for the same reason `pkg/tui/poll_test.go` pins its variables — a developer's real session must never reach the fixture servers.

`gofmt` and `go vet ./pkg/hivectl/...` are clean; the full suite runs in CI.

## Related issues

Fixes #5651

Builds on the session lane from #5645 / #5649 (`HIVE_DASHBOARD_COOKIE`); lands independently of #5649's merge order.

## Contributor checklist

- [x] Docs updated: `src/docs/hivectl.md` gains a "login / logout" section covering the flow, the cache location and permissions, precedence, the one-flow-at-a-time race, allowlist refusals, and expiry.


clubanderson (commented):
Independent verification against the v4 source, including the server side. **Verdict: APPROVE — safe to merge** (one merge-order note below).

Verified:
- Device flow matches the hub/spoke endpoints exactly: /api/gh-user-auth/{start,poll,logout} exist on the mux, sit on the auth allowlist (server.go isPublicPath prefix), and handleGHUserAuthPoll's real contract is what the client codes to — pending/slow_down/complete/error statuses, the allowlist denial as {status:error} inside a 200, the session arriving as Set-Cookie on the poll (setSessionCookie + terminal assertion), and the bare 400 for no-flow-in-progress that DeviceLogin wraps into ErrDeviceFlowConflict.
- 0600 from first byte: os.CreateTemp opens 0600, Chmod pins it, write → fsync-less Close → same-dir rename; there is no window with looser permissions, and the test asserts both file and directory modes.
- Credential never printed: pinned from both sides — the command test greps stdout+stderr for the planted session value, and DeviceLogin never embeds the cookie in errors.
- sessionFallbackTTL (30d) genuinely mirrors the dashboard's sessionCookieMaxAge; expiry follows the hive_session cookie's own Max-Age when present, not the shorter terminal-assertion cookie.
- Logout clears the cache even when the server call fails, presents the (possibly expired) session so the server clears the right one, and the no-op case is calm.
- Composition with #5649's convention: LoginResult.Cookie is a full Cookie-header value (all Set-Cookies joined '; '), byte-compatible with HIVE_DASHBOARD_COOKIE, and the env var beats the cache unconditionally on both the subcommand and TUI paths. SessionKey's loopback normalization (127.0.0.1/::1 → localhost) correctly bridges the two defaults.
- 401-only login hint; 403 correctly excluded (working credential, wrong role).

Merge-order note (non-blocking): the TUI handoff (exportCachedSessionForTUI) sets HIVE_DASHBOARD_COOKIE, which the TUI only reads once #5649 lands — until then it's inert but harmless, and pkg/hivectl deliberately spells its own CookieEnv so this PR compiles and works standalone. Fine in either order.

PR #5668, opened 2026-09-02
fix(packs): stop counting disabled-agent re-adds as creations; give cadences operator ownership

### LINKED ISSUE #5632: ApplyPack reports created:1 on every steady-state apply, silently reverting operator governor cadences

clubanderson:
## Summary

Fixes the third member of the durable-config family: `ApplyPack` reported `created: 1` on every steady-state apply and, because `created > 0` authorizes a governor reset, silently reverted every operator-set governor cadence to the pack default on each apply.

### Root cause

The agent manager's process table tracks **enabled** agents only — it is built from `EnabledAgents()` at boot and re-filtered to it by `ReconcileAgents` on every config reload (`cmd/hive/main.go`). So a pack agent the operator disabled (`brainstorm` in the report) is *legitimately absent* from the manager, forever:

1. `applyPack` finds it in `Config.Agents` but `GetStatus` fails → re-adds it and counts it as **created**.
2. `created > 0` → `isFirstApplyOrExpansion` → the pack re-asserts `pack.Governor.Cadences`, clobbering cadences the operator set via `PUT /api/config/agent/{name}/cadences`.
3. The apply saves the config → the fsnotify reload runs `ReconcileAgents(EnabledAgents())` → the disabled agent is evicted again → goto 1, on every subsequent apply.

Step 3 is the lifecycle piece the issue could not pin down — an in-process double-`ApplyPack` test passes on unmodified v4 because nothing evicts the agent between applies. The new regression test includes that step and reproduces the exact field report on the old code: `created=[brainstorm]` on a steady-state apply, and the operator's `surge/scanner` cadence reverted `9h → 4h`.

### Fix

**created semantics.** `created` now means "agent newly added to the config roster". Re-registering a config-existing *enabled* agent into the manager is a repair — logged separately as `readded_agents`, never counted in `created`, never treated as a roster expansion. For a *disabled* agent the re-add is skipped entirely: its absence from the manager is the correct steady state, and re-adding it only queued it for the next reload's eviction (the churn that drove the loop).

**Cadence ownership.** Governor mode cadences get the same operator-ownership marker `Model`/`Backend` gained in #5558/#5586:

- The dashboard cadence write claims `governor.cadence_owners[mode][agent] = operator`.
- No pack apply — steady-state merge, forced re-apply (`ApplyPackForce`), or explicit level change — reconciles an operator-owned cadence back to the pack value, exactly as packs already respect an operator's model choice. Pack-owned cadences still follow the pack; fresh applies still seed pack cadences.
- `LoadWithDashboardOverlay` (the Kubernetes reload path) re-adopts operator-owned cadences and their markers from the dashboard overlay, so a ConfigMap remount cannot drop them from memory — previously the reload rebuilt `Governor.Modes` from the seed alone.
- Ownership markers are released when the cadence entry itself is removed (ACMM level gate), so a stale claim cannot block a future pack seed.

### Tests

- Steady-state apply (operator-disabled pack agent evicted via the real `ReconcileAgents(EnabledAgents())` mechanism) reports `created: 0`, leaves the operator cadence untouched, and reports no cadence change for it. **Fails on the old code with the exact symptoms from the issue.**
- Fresh apply still seeds every pack cadence and creates the roster.
- `ApplyPackForce` re-derives pack-owned cadences from the target pack but leaves an operator-owned cadence alone.
- Claim/release helpers and overlay adoption (operator-owned entries copied, pack-owned entries not, dangling markers ignored).

Complements #5633, which adds `created_agents` logging for attribution but does not change the behavior.

Fixes #5632


clubanderson (commented):

Verified:
- Root cause fix is right: the manager's process table really is EnabledAgents-only (agent/manager.go ReconcileAgents; the reload path in main.go rebuilds from EnabledAgents()), so a disabled pack agent is legitimately absent forever and its re-add was a phantom creation. The guard 'err != nil && existing.Enabled' skips disabled agents entirely and reclassifies enabled-agent repairs as readded — which no longer feeds created and therefore no longer flips isFirstApplyOrExpansion (line 298: len(created) > 0 || removedByGate || force).
- Cadence ownership is respected on ALL apply paths: I found exactly two places pack.Governor.Cadences is asserted (applyPack's loop and handlePackSetLevel's rebuild) and both now honour CadenceIsOperatorOwned — including the forced path, matching model-ownership semantics. removeAgentsUnavailableAtLevel releases markers with the cadence, and it runs inside applyPack before handlePackSetLevel's preserve-then-seed block, so stale claims can't block a future seed.
- Overlay re-adoption: CadenceOwners has yaml/json tags, saveConfig → Config.Save persists it, and LoadWithDashboardOverlay adopts operator-owned entries (with markers, before the fullness guard) so a ConfigMap remount can't drop them — with the pack-owned overlay entries correctly NOT overriding the seed, and dangling markers not materializing modes (both pinned in tests).
- Composition with #5633 held through the rebase: the 'ACMM pack applied' line keeps created_agents, and readded_agents is logged as its own INFO line rather than folded back into created.
- The steady-state regression test reproduces the live lifecycle (disable brainstorm → ReconcileAgents eviction → re-apply) and is red-before-fix on both assertions (created non-empty AND the scanner cadence revert). The PUT handler claiming ownership is asserted through the real handler.

No blockers found.

PR #5669, opened 2026-09-02
✨ feature: hivectl tui — attach to an agent terminal on a containerized or remote hive

### LINKED ISSUE #5644: ✨ feature: hivectl tui — attach to an agent terminal on a containerized or remote hive
## Summary

Hive runs each of its agents inside a tmux session. `hivectl tui` is the
terminal dashboard an operator uses to watch and drive those agents, and its
`a` key is meant to drop you straight into the selected agent's session.

Today `a` only works when the TUI process happens to be running on the same
machine, as the same user, as the tmux server — it shells out to a local
`tmux attach`. In the Podman install that `bin/hive-podman-setup.sh` sets up,
the agents run inside a container owned by a dedicated `hive` user, so an
operator running `hivectl tui` under their own account can never attach. The
key is not broken; it is structurally unreachable for that topology, which is
the recommended one.

The web dashboard already solved this: `/terminal/?arg=hive-<agent>` reverse
proxies to the in-container ttyd behind the same authentication as the rest of
the dashboard. The TUI has no client for it. This issue asks for that path —
`a` goes through the authenticated proxy, and falls back to the existing local
exec only when the TUI genuinely is co-located.

Blast radius: new code in `src/pkg/tui/`; no server-side change is required,
because the endpoint already exists.

## Why this is filed now

Remote terminal embedding was a declared non-goal of the `hivectl tui` epic
([#4907](https://github.com/kubestellar/hive/issues/4907)), recorded in
`src/docs/design/tui.md` §1 and `src/docs/hivectl.md` ("v1 boundaries"). #4907
is closed and no follow-up was filed, so the boundary is currently undocumented
as future work. This is that follow-up, not a regression report.

## How to observe it

On a host installed with `bin/hive-podman-setup.sh` (dedicated user, rootless
Podman), as the operator account:

```
hivectl tui        # focus the Agents pane, select an agent, press `a`
```

Result: a footer message `Attach failed: …`. The preflight
(`tmux has-session -t hive-<agent>`) runs in the operator's own namespace,
where no such session exists — the sessions belong to the container's per-agent
UIDs.

Observed on a Fedora/Bluefin host: Hive runs as `hive:x:1001:1001:Hive
standalone deployment:/var/home/hive:/bin/bash`, the operator is uid 1000, and
`/var/home/hive` is not readable by them. The current workaround is to leave the
TUI entirely: `podman exec -it <container> tmux attach -t hive-<agent>`.

## Current behaviour

`src/pkg/tui/attach.go` preflights `tmux has-session -t hive-<agent>`, then
suspends the TUI with `tea.ExecProcess` and runs `tmux attach` locally; on exit
the TUI resumes and refreshes the fleet. Documented in `src/docs/hivectl.md`
under "Local tmux attach", including the explicit statement that it "cannot
attach to a session running on a remote hive".

## The server surface that already exists

- `registerTerminalProxy` (`src/pkg/dashboard/terminal_proxy.go`) mounts
  `GET /terminal` on the Go dashboard and reverse proxies to
  `127.0.0.1:7681` (overridable via `HIVE_TTYD_PORT`).
  `httputil.ReverseProxy` passes the protocol upgrade through, so the ttyd
  WebSocket works, and the route sits behind the same `authenticate()`
  middleware as everything else — session cookie, hub-injected identity, or
  `?token=`.
- ttyd is started with `-a`/`--url-arg`, so `?arg=<session>` selects which tmux
  session to attach to; `src/deploy/ttyd-tmux.sh` resolves the per-agent socket
  (`/tmp/tmux-{agent_uid}/hive-{name}`) and `su-exec`s as the owning user.
- The dashboard builds the URL in `terminalUrl()`
  (`src/pkg/dashboard/static/index.html`) as
  `{proto}//{host}/terminal/?arg={session}{tokenParam}`.
- Adjacent, possibly reusable: `GET /api/agents/{name}/terminal-urls` and
  `POST /api/terminal/assertion/renew`.

Note that `:7681` is deliberately never published — `src/deploy/quadlet/hive.container`
and `hive-gateway.container` both say so, and the gateway publishes `3001` and
nothing else. Any design here must go through `/terminal` on the published
port, not reach ttyd directly.

## What is missing

1. A ttyd WebSocket client under `src/pkg/tui/client/` that dials
   `{HIVE_DASHBOARD_URL}/terminal/?arg=hive-<agent>` with the
   `HIVE_DASHBOARD_TOKEN`, and speaks ttyd's framing.
2. A Bubble Tea view that renders the remote terminal and forwards input,
   with a clear way back to the dashboard frame.
3. A decision on how `a` chooses between remote and local, and what the help
   overlay and footer say in each case. Suggested rule: prefer the proxy
   whenever `HIVE_DASHBOARD_URL` is not loopback, keep local exec as the
   co-located fast path, and never silently fall back in a way that leaves the
   operator unsure which session they are looking at.

## Out of scope

- Publishing `:7681`. Not an option; see above.
- Any change to `ttyd-tmux.sh` or the session naming scheme.
- Session creation or restart from the TUI — `↻ restart` remains the
  dashboard's.

- `a` attaches to the selected agent's session against a hive reached over
  `HIVE_DASHBOARD_URL`, authenticating with `HIVE_DASHBOARD_TOKEN`.
- A 401/403 renders as an authorization message, matching the wording
  convention the other actions already use (`… : owner access required`).
- A failure to dial leaves the previous frame intact and reports in the
  footer, consistent with the existing "a failed fetch never blanks a pane"
  rule.
- Tests use an `httptest` server speaking the ttyd handshake; no real ttyd,
  container, or network in CI, per the epic's testing convention.
- `src/docs/hivectl.md` "Local tmux attach" and "v1 boundaries" are updated,
  and the help overlay's golden output regenerated.

clubanderson:
## Summary

`hivectl tui`'s `a` key attaches to the selected agent's tmux session — but only by shelling out to a local `tmux attach`, which made it structurally unreachable in exactly the topology the project recommends: on a `bin/hive-podman-setup.sh` install the sessions live inside the container under per-agent UIDs, and on any hive reached over the network there is no local session at all. The web dashboard solved this long ago with the authenticated `/terminal` reverse proxy in front of the container's ttyd; the TUI simply had no client for it.

This adds that client and routes `a` by where the session actually is:

- **Local fast path.** Loopback `HIVE_DASHBOARD_URL` *and* a local `tmux has-session` hit → the existing zero-network `tmux attach` exec, unchanged.
- **Remote attach.** A non-loopback dashboard, or a loopback one whose local session is missing (the Podman case) → dial the dashboard's `/terminal` websocket, speak ttyd's framing, and bridge the operator's real terminal to the remote session. The fallback is never silent: one banner line names the hive and session before the first byte of output.

**No server-side change.** ttyd's `:7681` stays loopback-only, as `src/deploy/quadlet/hive.container` requires; the only path in is the `/terminal` route both dashboards already serve behind their existing gates.

## Transport and credentials

The dial crosses two gates, and the credential placement is dictated by what each actually reads:

| Gate | What it reads | What this client sends |
|---|---|---|
| Node proxy WS upgrade (self-hosted) | `?token=` **query only** (`src/proxy/server.js`) | `?token=<HIVE_DASHBOARD_TOKEN>` |
| Node proxy WS upgrade (hosted host) | hub cookie + per-hive terminal assertion | `Cookie: <HIVE_DASHBOARD_COOKIE>` verbatim |
| Go dashboard `authenticate()` | session cookie, hub proof, or `?token=` | both of the above |
| ttyd itself (`check_auth`, protocol.c) | `Authorization: Basic <base64(user:pass)>` on the upgrade, and the same value as `AuthToken` in the first protocol message | `Basic base64("hive:<token>")`, derived exactly as `deploy/entrypoint.sh` derives ttyd's credential; `HIVE_TTYD_CREDENTIAL` honoured as an override with the same precedence `deploy/test_ttyd_url_arg.sh` pins server-side |

Notably the shared token does **not** travel as `Bearer` on this one dial: the Go server prefers the `Authorization` header over the query, so a Bearer header would shadow the `?token=` that ttyd's Basic requirement forces us to keep the header for. This is the browser flow — `?token=` from `terminalUrl()` plus the basic-auth prompt — automated.

This builds directly on #5649's session machinery (same `Client`, same `HIVE_DASHBOARD_COOKIE`, same `IsUnauthorized`/`IsForbidden` classification) and is **stacked on that branch** — its three commits appear here until #5649 merges, after which this rebases to the two commits that are its own.

## The bridge

The remote terminal is deliberately **not** re-rendered inside a Bubble Tea pane through a terminal emulator. The bridge implements `tea.ExecCommand`: Bubble Tea releases the terminal exactly as it does for the local `tmux attach`, the bridge puts the real terminal into raw mode, sizes the remote to it (initially and on every SIGWINCH), and pumps bytes both ways. The operator's own terminal does the emulation, so fidelity is `tmux attach`'s, `ctrl+c` reaches the agent rather than the bridge, and detach is tmux's own `prefix + d`. A clean websocket close is a detach (quiet return, immediate fleet refresh, as before); an abnormal close reports in the footer, because the last keystrokes may not have arrived.

One subtlety: the stdin pump uses `poll(2)` with a timeout rather than a blocking read, so when the session ends the goroutine exits *without consuming a byte* — a blocked read would survive the attach and race the resumed TUI for the operator's next keystroke.

## Auth gates, refusals, and the token

- A 401/403 on the upgrade is refused **before** the TUI suspends, in the footer, in the existing wording convention: `Attach failed: owner access required` (403), and the 401 names both `HIVE_DASHBOARD_TOKEN` and `HIVE_DASHBOARD_COOKIE`, for the same reason #5649's preflight names both.
- The banner prints the dashboard base URL, never the dial URL — the token rides in the query, and the banner lands in scrollback that outlives the session. A test asserts the token never reaches the terminal.
- The role decision itself stays server-side, where it already lives (`authenticate()`/`authorizeTerminal`); the client presents credentials and renders the verdict.

## Testing

Per the epic's convention: `httptest` fixtures speaking the ttyd handshake — no real ttyd, container, tmux, or network.

- `pkg/tui/client/terminal_test.go` — all three credential lanes on the wire (query token, verbatim cookie, derived Basic); absent credentials omitted rather than sent empty; `HIVE_TTYD_CREDENTIAL` precedence; the full framing round-trip (init JSON with AuthToken+size, `'0'`+input, `'1'`+resize JSON, OUTPUT unwrapped, title/preference frames swallowed); and **the invariant: a 401/403 upgrade is a typed refusal (`IsUnauthorized`/`IsForbidden`), never a session**.
- `pkg/tui/attach_remote_test.go` — the loopback decision table; the Podman fallback (missing local session → exactly one proxy dial, marked `viaFallback`); the local fast path never touching the network; the whole bridge round-trip (keystroke out as an INPUT frame, remote output on stdout, normal close → `nil`, banner names hive+session and does not echo the token); detach-vs-transport-failure classification; and the footer wording for 401/403.
- Existing local-attach tests updated for the fallback (`TestPrepareAttachSurfacesMissingSessionInFooter` now proves the footer keeps the local cause when the fallback also fails); the package `TestMain` pins `HIVE_TTYD_CREDENTIAL` empty alongside the other credentials.

- [x] `cd src && gofmt -l pkg/tui` (clean) and `go vet ./pkg/tui/...`
- [x] `cd src && go test ./pkg/tui/... ./pkg/config/...` — tui packages green; `pkg/config`'s `TestEntrypointHardensRuntimeConfigItRecreates` fails identically on the clean tree in this environment (host umask; unrelated) and the env-vars parity test passes.

## Docs

- `src/docs/hivectl.md`: "Local tmux attach" becomes "Attach: local tmux, or the dashboard's terminal proxy"; keybinding table row updated; the "Local tmux attach only" v1 boundary is replaced by the actual boundary (no in-frame emulation).
- `src/docs/env-vars.md`: `HIVE_TTYD_CREDENTIAL` documented (it was real but undocumented), including the client-side read.
- Help overlay: `a`'s scope drops "(local)"; golden regenerated.
- `CHANGELOG.md` entry under Unreleased.

## Related issues

Fixes #5644

Stacked on #5649 (session auth); the terminal-login follow-up remains #5651.

## Contributor checklist


clubanderson:
Rebase note: this branch was stacked on #5649's commits and #5649 has not merged yet, so it was rebased onto #5649's freshly rebased branch (not directly onto v4) — its base still includes #5649's two commits plus their changelog fragments. The CHANGELOG.md entry for this PR was converted to a fragment at `changelog.d/added-5669-remote-attach.md` per the changelog.d system from #5679; the PR no longer touches CHANGELOG.md, so Unreleased-heading conflicts cannot recur. Once #5649 merges, GitHub will show only this PR's own two commits.

clubanderson (commented):
Independent verification of this PR's OWN commits (diffed against the #5649 head; the two #5649 commits were not re-reviewed here). **Verdict: APPROVE — safe to merge once #5649 is** (it's stacked, so its merge carries #5649's commits with it).

Verified:
- Auth gates before suspend: prepareAttach/prepareRemoteAttach run as a tea.Cmd, and DialTerminal turns a refused upgrade into a typed *APIError before anything is handed to tea.Exec — so a 401/403 lands in the footer (403 'owner access required', 401 naming both credential env vars) and the TUI never suspends into a dead terminal. Pinned by TestDialTerminalRefusalsAreTyped + TestAttachFailureStatusClassification, and IsUnauthorized/IsForbidden exist on the #5649 base.
- No token echo: the banner prints c.baseURL (target), never the dial URL that carries ?token=; the round-trip test plants a token and asserts it's absent from stdout.
- ttyd contract matches the server: /terminal prefix-strip in pkg/dashboard/terminal_proxy.go makes the public socket /terminal/ws; entrypoint.sh derives HIVE_TTYD_CREDENTIAL as hive:${HIVE_DASHBOARD_TOKEN} exactly as the client does; framing ('{' init with AuthToken/columns/rows, '0' input, '1' resize, output/title/prefs demux) is pinned by fixture tests. Token-in-query is the pre-existing dashboard-link convention, not a new downgrade.
- Loopback routing rule: non-loopback URL → proxy only (local tmux never consulted); loopback → local fast path with a never-silent proxy fallback (viaFallback banner). Unparseable/hostless URLs keep the historical local-only behavior. Both wrapped errors survive errors.As through the fold (typed tmux error AND APIError).
- No new server surface: the own-commit diff touches only src/pkg/tui, docs, changelog and go.mod (dep promotions to direct, no version changes).
- Concurrency: gorilla's one-writer rule is respected via writeMu across SendInput/Resize/Close; the poll-based stdin pump is waited on (bounded by its 100ms poll) so a parked read can't steal the resumed TUI's first keystroke, and the un-interruptible blocking pump is deliberately not waited on where no terminal is shared. TestMain quarantines HIVE_TTYD_CREDENTIAL like the other credential vars.

No blockers in this PR's own commits.

PR #5670, opened 2026-09-02
✨ feat(timeline): complete the lifecycle panel — journeys, all stages, dedupe, persistence

### LINKED ISSUE #5656: dashboard: Lifecycle Timeline panel renders only re-stamped enumeration noise — propose removal
## Summary

The dashboard's "Lifecycle Timeline" panel (between Repositories and Audit Log) displays only noise on a live spoke: fleet counters read `N in flight / 0 merged / 0 blocked`, and every visible row is `enumerated <repo>#NNNN — 1m ago`, re-stamped each governor cycle. Merged stays 0 on spokes that merge constantly. After tracing the full producer/consumer chain, this is not a disconnected wire — 4 of the 6 lifecycle stages never had a producer, and the two that exist make the panel structurally incapable of showing anything but the latest enumeration sweep. I propose removing the panel (details and a scoped alternative below); leaving the decision to the operator per policy.

## What the feature intended

`pkg/timeline` models a closed issue→PR journey: `enumerated → classified → kicked → pr_opened → merged/blocked`, stored in a 500-event in-memory ring, served by `GET /api/lifecycle-timeline` with a derived 6h `FleetHealth` roll-up, rendered as "Panel B" in `pkg/dashboard/static/index.html`.

Introduced by three direct commits on 2026-07-30 (no PR, no originating issue):

- `2f994ce3` — feat(timeline): lifecycle timeline + fleet-health API for the dashboard
- `4413507b` — feat(dashboard): expose ACMM level-up recommendation + record lifecycle timeline events
- `03c49d42` — dashboard: surface ACMM level-up recommendation and lifecycle timeline

## Why every row is stuck at "enumerated — 1m ago"

**1. Four of six stages have no producer — and never did.** Grepping all non-test code:

- `KindEnumerated` — recorded by `recordEnumeratedIssues` (`cmd/hive/main.go`)
- `KindKicked` — recorded by `recordKick` (`cmd/hive/main.go`)
- `KindClassified` — **never recorded anywhere**
- `KindPROpened` — **never recorded** (`cmd/hive/celwire.go` uses `celtrigger.KindPROpened`, a different package/bus)
- `KindMerged` — **never recorded anywhere**
- `KindBlocked` — only via the hook `annotatorAdapter` (`cmd/hive/hookwire.go`), i.e. only when a configured hook fires an annotation; never from real PR/issue blockage

The gap is acknowledged in-tree: `pkg/dashboard/api.go` (above `LifecycleTimeline()`) still carries `TODO(timeline): the governor eval loop should Record() real lifecycle events ... Until then the store is present but empty`. Commit `4413507b` later added the enumerated+kicked producers but the classified/pr_opened/merged half was never built. The `timeline.FromSpan` span-adapter escape hatch is also never called from non-test code.

**2. The enumerated flood re-stamps and crowds out everything.** `recordEnumeratedIssues` runs every governor eval cycle (`eval_interval_s`, 120s on the observed spoke) and re-records up to `maxTimelineEnumeratePerCycle = 50` events for the same actionable issues — no event IDs, no dedupe. The frontend fetches `?limit=50` newest-first. 50 fresh enumerated events per cycle == the panel's entire viewport, so the panel always shows exactly the latest sweep, all with the identical timestamp, at most one interval old. That is the observed "every row ENUMERATED, 1m ago, re-stamped every minute".

**3. The ring cannot honor its own window.** Live read-only evidence from a busy spoke's `/api/lifecycle-timeline`:

- Kinds ever present in the ring: `enumerated` and `kicked` only; merged/blocked = 0 in the entire retained history
- The full 500-event ring spanned ~14 minutes (enumeration flood churns it), while `FleetHealth` claims a 6h window (`windowMs: 21600000`)
- Top 50 events (what the panel renders): 50/50 `enumerated`, all sharing one timestamp
- Counters: `inFlight 78 / merged 0 / blocked 0` — "in flight" is just "distinct issues in the actionable backlog", already shown by the Repositories section
- The store is process-memory only; every restart wipes it

So even if merged/blocked producers were added tomorrow, a merge event would be evicted within ~14 minutes and lost entirely on restart — the fleet counters can never be honest without dedupe + persistence, i.e. a redesign, not a wiring fix.

## Why the panel is redundant as built

- Its only real content — the actionable issue set — duplicates the **Repositories** section (issue/PR chips, needs-human badge), with less information.
- Its `kicked` events duplicate the **Audit Log**, which records every governor kick with agent + trigger (the same call site, `main.go` kick loop, writes both).
- Its headline numbers are misleading: `0 merged / 0 blocked` reads as "nothing ships", when the truth is "no producer exists".

## Proposal

Remove the panel and its fetch path:

1. `index.html`: `lifecycle-section` markup, `renderLifecycle`/`fetchLifecycle`, the `lc-*` CSS, and drop `fetchLifecycle` from `v4RefreshPanels`.
2. `pkg/dashboard/api.go`: `GET /api/lifecycle-timeline` handler + ACMM filter helpers (keep `LifecycleTimeline()` itself).
3. Optionally retire `recordEnumeratedIssues` — nothing else consumes enumerated events.

**Scope note — keep the store and kick recording.** `pkg/retro` (`applyTimeline`) legitimately consumes `KindKicked` from this store for kick counts / ClaimedAt, and the hook annotator writes to it. Only the dashboard panel + endpoint are dead weight.

**Alternative** if the lifecycle view is still wanted: treat it as a new feature (persisted per-issue state with dedupe, merged/blocked detection from PR-state diffs across enumeration cycles or the automerge sweep) rather than resurrecting this ring. Happy to write that RFC instead if preferred — but the panel should not keep rendering misleading zeros in the meantime.

clubanderson:
Fixes #5656

The Lifecycle Timeline panel ("Panel B") rendered only re-stamped enumeration noise: 4 of its 6 stages never had a producer, the two that existed flooded a 500-raw-event ring every eval cycle (~14 minutes of retained history under a claimed 6h window), and a restart wiped everything — so the counters read `0 merged / 0 blocked` on hives that merge constantly. Per the operator's call on #5656, this finishes the feature rather than removing it.

## Store: journeys, not raw events (`pkg/timeline`)

- Events are keyed by `repo#number` + kind into **journeys** — one entry per work item with per-stage `firstAt`/`lastAt`/`count`/agent/attrs and a derived `current` stage. Re-enumeration refreshes a timestamp on the existing journey instead of appending 50 rows per cycle.
- Stage semantics: `merged` is sticky-terminal; `blocked` holds until later progress clears it; otherwise the furthest progress stage wins.
- Ring sized for journeys (`MaxJourneys = 500`, LRU-evicted by `lastAt`) — weeks of fleet history where the raw ring held minutes.
- **Persistence**: `/data/lifecycle-timeline.json` with the repo's atomic-persist idiom (unique temp + fsync + rename, per the post-#5625 mutation ledger). Throttled (15s) for non-terminal refresh floods, immediate for `pr_opened`/`merged`/`blocked`. A corrupt file is moved aside to `.corrupt` and never refuses startup — this is telemetry, not ownership state.
- `FleetHealth` gains `coveredMs`: how much history actually backs the counts, so the roll-up can never again claim a 6h window over a 14-minute ring.

## Producers — all riding existing paths (no new GitHub polling, no new goroutines)

| Stage | Wire | Call site |
|---|---|---|
| `enumerated` | governor eval loop (existing) | `cmd/hive/main.go` `recordEnumeratedIssues` — cap raised to the endpoint's journey limit now that dedupe removed the eviction-flood concern |
| `classified` | scheduler's classifier pass, the moment lane routing decides lane/tier/model | `pkg/scheduler` `BuildKickMessages` / `BuildAgentMessageFromLastActionable` → `recordClassified` via new `SetLifecycleRecorder` |
| `kicked` | governor kick loop (existing) | `cmd/hive/main.go` `recordKick` |
| `pr_opened` | pr-request watcher's typed `PROpenedHook` + the attribution audit stream (`agent_pr_created`) | `cmd/hive/main.go` hook wrapper + `cmd/hive/lifecyclewire.go` `recordLifecycleFromAudit` (deduped by design) |
| `merged` | **every** `pr_merged` attribution audit entry: both automerge sweep paths and `MergePR` (dashboard queue + merge-watcher merges humans drive through the hive) | audit sink bridge in `cmd/hive/main.go` → `recordLifecycleFromAudit` |
| `blocked` | the fix-loop escalation that hands a PR to a human (`needs-human`), carrying fix attempts + failing checks — not just hook annotations | `runEscalationSweep` → `recordBlocked` |

Refs are normalized to the short `repo#number` form the enumeration producer already uses (verified against the live snapshot), so all stages of one item share one journey. The celtrigger `KindPROpened` bus was left alone — the audit sink + typed hook cover creation with less coupling.

## Dashboard

- `GET /api/lifecycle-timeline` now serves `{journeys, fleet}`; the ACMM level filter gates whole journeys (agent-less journeys always pass).
- Panel B renders one row per work item: `ref`, stage chips along the fixed axis (`enum → class → kick → pr → merged/blocked`) with the current stage colored via the existing `v4KindClass` mapping and earlier stages dimmed, latest agent, last activity. Counters keep their markup and are now honest, with a coverage label derived from `fleet.coveredMs` (“over the last 2h of recorded history (window 6h)”).
- `EnableLifecyclePersistence` mirrors the `EnableSessionPersistence` wiring on the PVC.

## Compatibility

- `pkg/retro` keeps working: `applyTimeline` reads kick cardinality from the journey's kicked-stage `Count` and PR refs/state from stage attrs — its existing `TestReconstruct` expectations (2 kicks, drift pause, merged PR ref) pass unchanged.
- The hook `annotatorAdapter` (`KindBlocked` annotations) records into the same store unchanged; `Recent`/`ByIssue` remain as synthesized per-stage event views.

## Tests

- Store: dedupe, stage transitions, blocked-vs-progress, LRU eviction, persistence round-trip, corrupt-file handling, throttle coalescing, honest coverage, JSON shape, nil-safety, `-race`.
- Producers: each wire lands the right kind — classified (scheduler pass), pr_opened (hook + audit bridge, deduped), merged (audit bridge), blocked (full `runEscalationSweep` pass at threshold).
- UI: markup-pin tests per the #5479 style (journey list, stage chips, coverage label, fetch path; raw-event markup pinned absent).

`gofmt` clean on touched files; `go vet` clean; targeted suites green (`pkg/timeline` incl. `-race`, `pkg/retro`, `pkg/scheduler`, `pkg/tracing`, full `pkg/dashboard`, touched `cmd/hive` tests).


clubanderson (commented):

Verified:
- Dedupe semantics: Record keys by IssueRef+Kind, refreshing LastAt/Count/attrs — the every-cycle enumeration sweep can no longer flood (50 re-records → 1 journey, Count 50, pinned). deriveCurrent's rules (merged sticky-terminal, blocked holds only until later progress, re-enumeration never regresses the stage) are each pinned by a test.
- Journey identity actually converges: enumerated uses Issue.Repo (config repos are short names, or org-qualified only for cross-org repos), classified uses the same worksource Key over the same field, and the audit/hook/escalation sources go through timelineRef which strips only the project org — so 'repo=owner/repo' from the automerge sweep and short 'repo' from the pr_request watcher both land on the enumeration's key. Cross-org repos keep their qualifier on both sides. All four stage-splitting scenarios are covered by TestTimelineRef.
- Producers are wired where claimed, all on existing paths: SetAttributionAudit callback (covers agent_pr_created + all three pr_merged emitters — both automerge sweep paths at automerge_sweep.go:571/707 and MergePR at pullrequest.go:208), the typed SetPROpenedHook, the scheduler's classify pass in BuildKickMessages/BuildAgentMessageFromLastActionable, and runEscalationSweep's newly-escalated path only. parseAuditDetail's ', ' split matches AuditDetail's join. No new polling or goroutines.
- pkg/retro compatibility: ByIssue/Recent survive as synthesized views; KicksReceived now reads Stage.Count (true cardinality, which the collapsed event view can't carry), ClaimedAt from FirstAt, drift-pause and pr_state scans made deterministic by LastAt ordering. No other consumer of the removed ring API exists outside the package (hub/timeline.go is an unrelated store).
- Persistence: atomic idiom done right (unique temp + chmod + write + fsync + close + rename, temp removed on every failure path), corrupt file moved aside instead of refusing startup (correct for telemetry, unlike the mutation ledger), throttled 15s for non-terminal floods with immediate persist for pr_opened/merged/blocked. Everything runs under the store's own mutex with no re-entrancy.
- Honest windowing: CoveredMs is derived and the UI test pins that the label uses fleet.coveredMs ('of recorded history') rather than claiming the requested window; the ACMM-filtered fleet counts are re-derived over the full filtered set, not the limit-truncated page.
- Markup pins: both UI tests hold against the shipped index.html (journey rows, fixed stage axis through the existing v4KindClass mapping, old lc-events list gone), and the maxTimelineEnumeratePerCycle bump is now justified by dedupe rather than papering over eviction.

One observation, not a blocker: #5656's headline proposal was removal, with completion as the scoped alternative — this PR takes the alternative, and having traced the producers I think that's the defensible call; the panel now shows real outcomes instead of noise. Worth the operator knowing the decision point existed.

Issue #5681, opened 2026-09-02
🐛 bug: a hub restart makes every in-flight contributor task unresumable — the agent is interrupted, then handed the same issue back seconds later


Danathar:
## Summary

When the hub restarts to take an upgrade, every contributor that is mid-task has its
work thrown away — and then is handed the identical issue back seconds later.

A contributor relay holds one task at a time and keeps working through a brief
disconnect, re-asserting the task when it reconnects. The hub only honors that
re-assertion if it can match a **server-issued lease** (`lookupLease`), which is
deliberate: #C4 established that a client must never be able to assert ownership of
work the server did not assign. But leases live only in the hub's memory. A restart
empties that store, so after an upgrade *no* in-flight resume can ever match, the
relay is told `no active lease for this task`, and the revoke path interrupts the
agent mid-turn.

The waste is visible in the ordering: the hub revoked the task at 14:24:40 and
re-assigned **the same issue to the same relay** at 14:24:44. Ownership was never in
question — only the record of it. The agent restarted the identical work from zero,
discarding about two and a half minutes of a turn that was progressing normally.

Blast radius: every contributor holding a task at the moment of any hub restart.
Self-upgrade rolls make this routine rather than rare.

## Observed, 2026-09-02

Contributor `Danathar`, claude backend, relay protocol 1.2, working issue #5617.
Relay console (timestamps local, UTC-4):

```
[10:22:03] Task assigned: issue kubestellar/hive#5617 — [v5] reviewer lane follow-ups
[10:22:06] Task prompt sent to CLI
[10:24:09] Connection closed (code=1012 service restart: hub restarting for upgrade). Reconnecting in 1000ms...
[10:24:39] Connection closed (code=1006 abnormal closure — no close frame). Reconnecting in 1000ms...
[10:24:40] Connected / Authenticated as c-ca8187f7480c (tier: contributor)
[10:24:40] Reconnected while working on kubestellar/hive#5617 — resuming
[10:24:40] Task revoked: ct-kubestellar/hive-5617-1788358923 — no active lease for this task
[10:24:43] Relaunching claude after task revoke: …
[10:24:44] Task assigned: issue kubestellar/hive#5617 — [v5] reviewer lane follow-ups
[10:24:45] Task prompt sent to CLI
```

Hub activity feed, same window:

```
2026-09-02T14:24:40Z | joined
2026-09-02T14:24:44Z | picked up | issue kubestellar/hive#5617
```

[...]

Issue #5120, opened 2026-08-29
🐛 bug: the activity feed names external (Linear/Jira) work items as repo#0 or a bare task id


Danathar:
## Summary

- The Operations activity feed is how an operator sees what contributors are
  doing: `picked up`, `completed`, `failed`, and so on, each naming the work item.
- **External work items — Linear, Jira — are named wrong in every one of those
  entries.** Such an item deliberately carries `Number == 0` and puts its identity
  in `Key`/`ExternalID` ([#4245](https://github.com/kubestellar/hive/issues/4245)),
  but the feed still derives its label from `Number`. So a Linear ticket is
  announced as `issue acme/team#0: …` when it is picked up, and as a bare internal
  task id like `t-8f3a91` when it completes.
- The two entries for the same item therefore do not even match each other, and
  every zero-numbered item from a given repo renders identically — an operator
  reading the feed cannot tell two Linear tickets apart, or connect a completion
  back to the pickup it closes out.
- Found while fixing the same assumption in the release-entry helper
  ([#5097](https://github.com/kubestellar/hive/issues/5097) / PR #5108). That PR
  fixed only its own new helper and deliberately left these four pre-existing
  sites alone rather than widening its scope.
- Blast radius: the activity feed's display for non-GitHub work sources. GitHub
  work is unaffected — `Number > 0` renders correctly everywhere. Nothing routes,
  assigns, or keys off these strings; the hub already keys in-flight work on
  `identityKey()`, which is correct.

## The four sites

Line numbers against `v4` at `fad4ccc9`, all in `src/pkg/dashboard/contribute_ws.go`.

**Unconditional `%d` — renders external work as `#0`:**

```go
// :2165  (yank / reassign)
taskDesc := fmt.Sprintf("%s %s#%d: %s", msg.Kind, msg.Repo, msg.Number, msg.Title)

// :3120  (picked up)
taskDesc := fmt.Sprintf("%s %s#%d: %s", task.Kind, task.Repo, task.Number, task.Title)
```

**Guarded on `Number > 0` — falls back to the internal task id, dropping the
title as well as the identity:**

```go
// :3430
completedDesc := msg.TaskID
if completedTask != nil && completedTask.Number > 0 {
    completedDesc = fmt.Sprintf("%s %s#%d: %s", …)
}

// :3574
failedDesc := msg.TaskID
if failedTask != nil && failedTask.Number > 0 {
    failedDesc = fmt.Sprintf("%s %s#%d: %s", …)
[...]

PR #5688, opened 2026-09-02
🐛 fix(contribute): make the task lease outlive the hub process (#5681)

Danathar:
## Summary

A hub restart threw away every in-flight contributor task, then handed the identical issue back seconds later.

A contributor relay holds one task at a time and keeps working through a brief disconnect, re-asserting the task when it reconnects. The hub only honours that re-assertion if it can match a **server-issued lease** — deliberate, since #C4: a client must never be able to assert ownership of work the server did not assign. But leases lived only in the hub's memory. A restart emptied the registry, so after an upgrade *no* in-flight resume could match, the relay was told `no active lease for this task`, and the revoke interrupted a working agent mid-turn.

The waste was visible in the ordering: revoked at 14:24:40, **the same issue re-assigned to the same relay** at 14:24:44. Ownership was never in question — only the record of it. Self-upgrade rolls (#5391) make this routine rather than rare, and it hits every contributor holding a task at the moment of any restart.

**The fix is option (1) from the issue:** persist the lease. It is written to `/data/contributors/task-leases.json` on every assignment, renewal and release, and read back at startup, so a restart becomes just a longer-than-usual disconnect.

## Why this does not weaken C4

The security posture was "never rebuild ownership from client-supplied fields". That never required the record to be *volatile* — only to be the *server's*:

- The restored record is one the **hub itself wrote**, and a resume is matched against it exactly as before, on `{identity, task_id, repo, number, generation}` and inside the window. The test asserts all six mismatch cases (wrong task / generation / repo / number / identity, and an unversioned `gen == 0`) are still refused after a restart.
- **No credential is in the file.** The scoped GitHub token is minted per assignment and delivered separately (#2537). Restoring a lease grants the ability to re-adopt a task the hub already issued, never to obtain a credential without passing `selectTask`.
- Written **0600**, unlike the sibling contributor ledgers' 0644 — this is an authorization record, not a report.
- A lease already past its expiry is **dropped at load**, not restored, so a stale file cannot resurrect a task that is no longer re-adoptable. A malformed file degrades to "no leases" — the pre-fix behavior.

## Three couplings that had to move with it

Persisting the map alone would have been quietly wrong in three ways:

1. **The renewed window is what gets persisted.** Persisting only the assignment-time window would bring a task that had been progressing for longer than `leaseTTL` back *already expired* — precisely the defect #4260 fixed in memory.
2. **`taskGen` is advanced past every restored generation at boot.** It is an in-memory counter that restarts at zero, so persisting leases without this would let a post-restart assignment mint a fencing token that **aliases** a restored one, and the #2568 Gate would accept a pre-restart straggler against a brand-new task. `src/docs/design/agent-state-inventory.md` residual 2 called this hazard out in advance — *"any change that persists one without the other (e.g. 'let's make leases survive restarts') silently breaks the fence"* — so that residual is now marked resolved, with the coupling stated in `loadLeases` and pinned by a test rather than carried only by that document.
3. **A restored lease holds its work item for two minutes after boot.** The double-assignment guard is built purely from live connections, which is empty right after a restart. Making the resume work *without* this would convert the old "lose the task" bug into two relays on one issue — which is the other half of the contract the issue asks to pin.

That third point is deliberately scoped, and the scoping is the part most worth reviewing. **A lease is not a hold in steady state:** a dropped socket keeps its lease so the relay can resume (#4260) while its item merely cools down (#2356's speculative hedge). Honouring leases as holds for the full 30-minute TTL would silently replace that hedge with a long park on every disconnect — a policy change nobody asked for. So the hold applies only to leases *restored from the previous process*, only for `leaseHoldGraceAfterStart` (2 min, against a 1-second relay reconnect backoff), and never to the requester's own lease. Leases minted by this process are never holds; their holders have live connections the existing guard already covers.

The lease also now carries the item's canonical `worksource` key, so the guard recognises external work — Linear/Jira items carry `Number == 0` and put their identity in `Key` (#4245) — instead of colliding every such item as `repo#0` (#5120).

- `src/pkg/dashboard/contribute_ws.go` — `taskLease` gains `key`/`restored`; `recordLeaseForKey`; save hooks in `recordLease`/`renewLease`/`revokeLease`/`lookupLease`'s expiry drop; new `persistedLease`, `saveLeasesLocked`, `loadLeases`, `pruneExpiredLeases`, `leasedIssueKeys`; `taskLeasesFile`/`startedAt` on the hub; `loadLeases` at construction; prune in `cleanupLoop`; the lease exclusion in `selectTask`.
- `src/pkg/dashboard/contribute_lease_restart_test.go` (new, 11 tests), `api_contribute_test.go` (redirect the new file in `redirectContributeWSDisk`).
- `src/docs/contributor-relay.md`, `src/docs/design/agent-state-inventory.md` (rows 31/32 + residual 2), `CHANGELOG.md`.

Disjoint from the open hold-gated PRs — none of them touch `pkg/dashboard/contribute_ws.go`. Also disjoint from my own #5682 (`pkg/escalation`, `pkg/scheduler`, `cmd/hive`).

### Validation

- `go build ./...`, `go vet ./pkg/dashboard/`, `gofmt -l` clean on every touched file. (`gofmt` flags eight *other* `pkg/dashboard` test files — pre-existing drift, not touched here.)
- `go test ./pkg/dashboard/` passes in full, including `TestSelectTask_DeclaredCapabilitiesDoNotAffectSelection`, which is the test that catches an over-broad lease exclusion: a blanket hold fails it, the restart-scoped one does not.
- **Mutation-checked, three ways** — each mechanism is independently pinned, and none of the new tests passes vacuously:
  - Remove `loadLeases()` -> 5 tests fail, the end-to-end one on the incident's own string: `the relay reconnecting after a hub restart was told "no active lease for this task"`.
  - Drop the persist-on-renew -> only `TestLeaseRestart_RenewedWindowSurvives` fails.
  - Drop the restored-lease hold -> only `TestLeaseRestart_RestoredLeaseHoldsItsIssue` fails.

### Not claimed

Option (3) from the issue — **drain in-flight tasks before exiting** — is not here. The issue calls it complementary rather than a substitute, and it is: it shrinks the window but cannot close it, since an upgrade will not wait out a 20-minute turn. It is also a change to shutdown sequencing rather than to the lease, and belongs in its own PR. Option (2) (reconstruct from the task-run log) is not needed once (1) lands.

## Related issues

Refs #5681. Related: #4260 (the resume contract this extends across a process boundary), #5390 (the clean close that precedes it), #5391, #5120, #4245, #2568, #2537, #2356.

## Testing

- [x] `cd src && go test ./...` — ran `./pkg/dashboard/` in full (the only package touched); passes.


clubanderson:
Review — **OK to merge** on green. Adjudicated against #5693 (same #5681 fix, v4 base); closed that one in favor of this — it missed the taskGen re-derivation the agent-state-inventory doc warns is load-bearing, the post-restart double-assignment window, write-under-lock ordering, and 0600.

Verified here:

- **Restart round-trip** is driven through the real protocol (assign → new hub over the same /data → reconnect + re-assert), and fails on the actual incident frames (revoke, re-offer) rather than a proxy.
- **No resurrection after legitimate expiry:** expired records are skipped at save, dropped at load, and reaped by `pruneExpiredLeases` in cleanupLoop; revokes persist immediately so a released task cannot come back.
- **TTL clock:** the renewed window is what persists (the #4260 case across a process boundary), so a long-running task does not come back pre-expired.
- **Fence:** `taskGen` is advanced past every restored generation at boot with a pinning test — closing residual 2 in the inventory doc rather than tripping it.
- **Grace hold:** restored leases hold their items for 2 minutes only, never for the requester's own lease, and a lease minted in-process is never a hold — so steady-state lease semantics (#4260/#2356) are unchanged.
- **Locking:** `saveLeasesLocked` is only ever called under `leaseMu` and takes nothing else; no re-entrancy, and the write-under-lock choice correctly prevents rename reordering.
- C4 exact-match is untouched (negative cases tested), file is 0600, malformed file degrades to empty.

Two non-blocking follow-ups: (1) the persist idiom is fixed-name `path+".tmp"` + rename with no fsync — short of the #5625 standard (unique CreateTemp + chmod + fsync file and dir). Tolerable here because a lost last write degrades to the pre-fix behavior, but this ledger (and its siblings) should be aligned with #5625 in a follow-up. (2) v4 carries the same registry and fence and needs a port of this change — including the taskGen re-derivation.

Danathar:
Note for whoever tracks backports: this merged to **v5**, but #5681 was reported against a **v4**-deployed hub and `loadLeases` is absent from `origin/v4`.

The failure recurred today after this merged. The hosted spoke serving `a2b71cb` (which is on `origin/v4`) restarted for an upgrade at 16:38:34Z and took an in-flight contributor task with it:

```
[12:38:34] Connection closed (code=1012 service restart: hub restarting for upgrade)
[12:38:35] Reconnected while working on kubestellar/hive#5701 — resuming
[12:38:35] Task revoked: ct-kubestellar/hive-5701-1788366752 — no active lease for this task
[12:38:40] Task assigned: issue kubestellar/hive#5701   <- same issue, new id, gen=2
```

The agent had already printed its `HIVE_VERDICT` for that task; the verdict was discarded and the work redone from scratch, costing ~6 minutes. Second occurrence today (the first cost the #5617 turn at 16:24Z).

Flagging only so the v4 gap is visible — the fix itself is what the issue asked for.

PR #5690, opened 2026-09-02
dashboard: operator branding — overridable strings, colours, and docs

hanthor:
### Problem

The dashboard is a single ~1.3 MB embedded SPA with no styling or naming hook.
An operator who wants their deployment to look like their project has two
options: fork the embedded document, or rebuild the image. Both mean carrying a
patch forever.

### What this adds

Two optional files on the data volume. Missing file → no change; omitted field
→ shipped default.

```
<data>/branding/branding.json   {"product_name","tagline","mark","title"}
<data>/branding/custom.css      colour tokens
```

`mark` replaces the bee **everywhere it renders** — both sidebars, the `<h1>`,
the Getting Started flyer, and the favicon (an inline SVG `data:` URI wrapping
the same glyph). `custom.css` is served at `/branding/custom.css` and linked
last in `<head>`, so it wins the cascade without `!important`.

Docs: `src/docs/branding.md`.

### Three things worth reviewing

**Substitution is anchored to exact markup.** A bare replace of `HIVE` or the
bee glyph across 1.3 MB of inline HTML/CSS/JS would corrupt script and prose. A
test asserts a bare `HIVE` inside a `<script>` survives. The trade-off is
explicit: if upstream markup changes, branding silently stops applying rather
than breaking the page.

**The flyer's bee is an HTML entity** (`&#x1F41D;`), not the literal glyph.
Anchoring only on the emoji left a bee animating on an otherwise rebranded
page.

**Branding can change inline SCRIPT bytes, so CSP had to move.** That flyer
builds its DOM from a JavaScript string literal containing the bee markup.
Script hashes were computed from the *embedded* document, which would no longer
describe the *served* bytes — CSP would block the flyer on a branded hive, and
it would fail silently in the browser. The CSP layer now hashes the document as
served.

`TestEmbeddedIndexScriptsSatisfyCSPHashes` asserted `served == embedded`, which
is a proxy for the real invariant: *every served inline script is covered by a
hash*. It now asserts that directly — and it is what caught this.

Relatedly, `Start()` hands the final bytes to the CSP layer explicitly instead
of `newIndexDocument` reaching out to set global state. Constructing a document
should not change the process-wide CSP, and a test building a throwaway
document must not shrink the real allowlist (it did).

### Startup vs per-request

Strings are baked in **once at startup**: the index carries a precomputed gzip
body and a strong ETag, so its content cannot vary per request without
discarding both. `custom.css` is read per request. So: edit CSS and reload;
edit JSON and restart. Documented.

### Not overridable, and deliberately not faked

The `KubeStellar Hive Dashboard` text in the `<h1>` (only its emoji is a span),
the honeycomb SVG in the flyer (paths, not a glyph — hide it with
`.wb-hive{display:none}`), and anything the SPA renders at runtime.

### Verification

`go build ./...` clean, `gofmt` clean, **full `pkg/dashboard` suite passes**
(not just the branding tests — that is how the CSP problem surfaced).

Branding tests cover: injection placement, no-`</head>` inertness,
served-vs-404 with content type, every string anchor including the entity form
and the favicon, partial/empty structs, script content left intact, and — the
ones that matter — that the **served** document carries both the link and the
strings. Those go through `newIndexDocument` and the real handler, because a
direct-call test still passes when the builder never invokes it (verified by
deleting the call).

Running on two hives in production.

### DIFF
--- a/src/docs/branding.md
+++ b/src/docs/branding.md
@@ -1,0 +1,114 @@
+# Branding a hive
+
+A deployment can carry its own name, mark and colours without forking the
+dashboard or rebuilding the image. Two files on the data volume:
+
+```
+<data>/branding/
+├── branding.json   # strings: name, tagline, mark, title
+└── custom.css      # colours
+```
+
+Both are optional. A missing file changes nothing; a partial `branding.json`
+leaves every field it omits at the shipped default.
+
+## Strings
+
+```json
+{
+  "product_name": "REEF",
+  "tagline":      "APPLICATIONS",
+  "mark":         "🪸",
+  "title":        "Reef — tuna-os"
+}
+```
+
+| Field | Replaces |
+|---|---|
+| `product_name` | the `HIVE` wordmark in the sidebar |
+| `tagline` | `GATEWAY DASHBOARD` beneath it |
+| `mark` | the 🐝 glyph — both sidebars, the `<h1>`, the Getting Started flyer, **and the favicon** |
+| `title` | the `<title>` |
+
+Substitution is **anchored to the exact markup** that renders each item, never
+a bare string replace: the document is ~1.3 MB of inline HTML/CSS/JS, and
+replacing every `HIVE` or `🐝` would corrupt script and prose. A consequence
+worth knowing: if that markup changes upstream, branding silently stops
+applying rather than breaking the page.
+
+> **Strings are read once at startup.** The index document carries a
+> precomputed gzip body and a strong ETag, so its content cannot vary per
+> request without discarding both. **Editing `branding.json` needs a restart.**
+> `custom.css` is read per request and takes effect on reload.
+
+## Colours
+
+`custom.css` is served at `/branding/custom.css` and linked last in `<head>`,
+so it wins the cascade without `!important`.
+
+Restyle **custom properties**, not layout. The dashboard is one embedded SPA
+whose markup changes between releases; re-pointing tokens survives upgrades,
+restyling structure does not.
+
+```css
+/* Dark — :root carries the DARK palette */
+:root {
+  --bg: #04161f;  --bg-soft: #07202b;  --surface: #0a2430;
+  --panel: #0a2430;  --panel-strong: #0d2c3a;  --card-bg: #0a2430;
+  --line: #12394a;  --line-strong: #1b4d63;  --border: #12394a;
+  --fg: #e8f6f8;  --text: #e8f6f8;  --muted: #8fb3bd;
+  --accent: #2fb6c4;
+}
+
+/* Light — body.light-mode OVERRIDES :root. Style both, or your theme
+   silently does nothing whenever the operator is in light mode. */
+body.light-mode {
+  --bg: #f2f9fa;  --surface: #ffffff;  --panel: #ffffff;
+  --line: #cbdfe5;  --fg: #06232e;  --text: #06232e;  --muted: #4d7683;
+  --accent: #10707c;   /* darker: the dark-mode cyan is ~2.2:1 on white */
+}
+```
+
+Two things that cost us time:
+
+- **Style both themes.** Overriding only `:root` is discarded in light mode.
+- **Re-check contrast per theme.** An accent tuned for a dark surface usually
+  fails as text on a light one.
+
+Available tokens: `--bg` `--bg-soft` `--surface` `--panel` `--panel-strong`
+`--card-bg` `--terminal-bg` `--line` `--line-strong` `--border` `--fg` `--text`
+`--muted` `--accent` `--oc-accent` and a named ramp (`--green` `--red`
+`--orange` `--amber` `--yellow` `--blue` `--cyan` `--indigo` `--purple`, each
+with `-bg`/`-border` variants where used).
+
+Keep **status** colours distinct from your accent — a green that matches the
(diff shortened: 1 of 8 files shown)


clubanderson:
Review — **CONCERNS.** The seam is designed mostly right from a security standpoint, but CI is genuinely red and there is one hardening gap worth closing before merge.

What checks out on the security review:

- **No CSP widening, and none needed:** the existing policy's `style-src 'self'` already admits the same-origin `/branding/custom.css`; the PR touches no CSP directives, fetches nothing cross-origin, and the link is a fixed same-origin path — an operator cannot point it at an external URL through this feature.
- **Who can set it:** there is no HTTP write path. The only ways to install a stylesheet are filesystem access to the data volume or the `HIVE_BRANDING_CSS` env — both operator-level. It is durable on the PVC, not runtime-only.
- The route is not in `isPublicPath`, so it sits behind `authenticate` like the SPA; `nosniff` + `no-cache` are right; unconditional link + inert 404 is a sound trade against the precomputed index; the ETag is computed after injection so caching stays coherent.

Blocking:

1. **`test (dashboard shuffle)` fails: `TestEmbeddedIndexScriptsSatisfyCSPHashes`** — "served SPA bytes differ from the embedded document — startup hashes would not match". The startup script-hash allowlist is derived from the embedded document, and `injectBranding` inside `newIndexDocument` makes the served bytes diverge from what the hasher read. The invariant that hashing and serving consume the same document has to be restored — inject before the hash pass runs (or feed the hasher the injected document), then update the test's expectation deliberately if the contract changes. This is precisely the CSP-hashing interplay this seam has to respect.
2. **`changelog-fragment-guard` fails** — the PR needs a `changelog.d` fragment.

Hardening gap (please address or explicitly justify):

3. `brandingCSSPath()` derives from `agents_dir`'s parent — `/data`, the same volume agent processes live on — and `handleBrandingCSS` serves whatever is at that path with no ownership/mode check and no size bound. If an agent UID can ever create `/data/branding/` (a function of `/data` root perms, which vary by deployment), that is CSS injection into the authenticated operator dashboard: defacement/UI-redressing, plus a limited exfiltration channel since the CSP's `img-src` allows any `https:` origin for beaconing. Suggested: refuse to serve unless the file and its parent directory are owned by the server's UID (or root) and are not group/world-writable, and cap the `ReadFile` size — that turns "an agent wrote it" into the same inert 404 as "no file". A doc line stating the `/data` ownership assumption would also do as a floor, but the stat check is cheap and closes the class.

Happy to re-review once the hash invariant is restored and the fragment lands — the no-fork/no-rebuild motivation and the small blast radius are appreciated.

clubanderson (approved):
Approved. I reviewed the diff line by line with a focus on the stylesheet path/config plumbing and default-unset behavior.

Notes:
- The browser-visible URL is fixed (`/branding/custom.css`); the configurable filesystem path is only server-side/operator-controlled, so I do not see a path-reflected XSS or request-path traversal issue.
- Missing override remains inert: the index link is unconditional, but the handler returns a cheap 404 with no config required.
- Served overrides use `text/css`, `nosniff`, and `no-cache`; no HTML/script injection surface was introduced beyond normal operator-controlled CSS.

Validation: `cd src && go test ./pkg/dashboard -run 'Test(InjectBranding|BrandingCSS|ServedIndexCarriesBranding)' && go build ./...`.

clubanderson:
Two gates to green — one real, one flake:

**1. changelog-fragment-guard (real):** the PR needs a fragment, and your fork rejects maintainer edits (403), so please add `changelog.d/added-5690-dashboard-branding.md` yourself with:

```markdown
- The dashboard can now carry a deployment's own branding without forking the embedded SPA or rebuilding the image ([#5690](https://github.com/kubestellar/hive/pull/5690)). Two optional files on the data volume drive it: `branding/custom.css`, an operator stylesheet served at `/branding/custom.css` and linked last from the index document so its rules win, read per request so dropping a file in takes effect on reload; and `branding.json`, which overrides the product name and accent colours baked into the index page at startup (editing it needs a restart — the index body is precompressed with a strong ETag, an asymmetry documented in `branding.md`). Both are data, not code: a deployment with neither file renders exactly as before, and the custom stylesheet path is registered with the CSP layer explicitly so a branding override can never widen the process-wide allowlist.
```

**2. `TestStart_ServesEndpoints` — “server did not start within 2 s” (flake):** could not reproduce on your branch alone or merged with v4 tip, including full-package shuffled runs; the 2s startup budget at server_test.go:733 lost the race on a loaded runner. Re-running the failed jobs.

clubanderson:
Checked again: the changelog fragment is still missing on the PR head (`e7689518`). Local validation on the actual PR head passes (`cd src && go build ./...`; `go test ./pkg/dashboard`), but `changelog-fragment-guard` is still the real blocker. Please add the previously requested `changelog.d/added-5690-dashboard-branding.md` fragment; once that is present I can rerun the merge gate.

clubanderson:
Adopted and merged via #5752 because maintainer pushes to the fork were not available. The replacement PR preserved hanthor's authored commits, added the required `changelog.d/added-5690-dashboard-branding.md` fragment, passed the merge gate after rerunning the known `TestStart_ServesEndpoints` startup-timeout flake, and merged to v4.

clubanderson:
Closed after adoption and merge via #5752.

Issue #5675, opened 2026-09-02
🔧 CHANGELOG.md Unreleased appends give every pair of landing PRs a structural conflict — move entries to changelog.d/ fragments


clubanderson:
## Problem

Every PR with a user-visible change appends an entry under `CHANGELOG.md`'s `## Unreleased` heading. Two unrelated PRs therefore edit the same lines under the same heading, and whichever merges second hits a structural merge conflict — not because the work overlaps, but because the changelog file is a shared append point. On 2026-09-02 alone this forced six rebases of PRs whose code had no overlap, and with repo-wide auto-merge disabled the conflict shows up at exactly the moment a maintainer wants to merge on green. The conflict class is *by construction*: as long as every PR appends to the same section, the churn scales with merge rate.

## Design

**1. `changelog.d/` fragments.** Each PR adds one file, `changelog.d/<category>-<pr-or-slug>.md`, containing exactly its changelog entry in the same narrative style entries use today. The category prefix (`added-`, `changed-`, `deprecated-`, `fixed-`, `security-`) maps to the existing `###` subsections. Two PRs now touch two different files — no shared append point, no conflict. A `changelog.d/README.md` documents the format.

**2. Compile at release time.** `src/scripts/compile-changelog.sh` folds fragments into `## Unreleased` (grouped by category, merged into existing subsections) and deletes them. `tagged-release.yml` runs it twice: in `decide` (working-tree only, so `derive-release-version.sh` sees fragment content when judging release-worthiness and bump) and in `release` just before the existing "move Unreleased into a dated section" step, with the release commit picking up both `CHANGELOG.md` and the fragment deletions. The existing decision/bump/idempotency machinery is untouched — an empty `Unreleased` **and** an empty `changelog.d/` is still the no-release no-op. Property-asserting test per the sibling script-test conventions (fixture fragments → expected CHANGELOG, empty-dir byte-identical no-op, idempotent second run, loud failure on an unrecognized category prefix).

**3. Guard.** A path-gated, reporting-only workflow (mirroring `testutil-guard.yml`, #5611): PRs touching `src/**` code must add a `changelog.d/*.md` fragment OR carry a `no-changelog` label. Docs-only/test-only changes are exempt, matching the existing changelog discipline (the same exemptions `changelog-reminder.yml` applies). Not a required context. Registered in `.github/release-lines.yml` per the #5611 lesson.

[...]

PR #5692, opened 2026-09-02
ci(podman-lanes): survive podman info schema drift — JSON + jq fallbacks, capability-asserting rootless-network check

### LINKED ISSUE #5678: [ci-maintainer] Podman arm64/Rootless lanes fail at env-assert: '.Host.RootlessNetworkCmd' template error (exit 125) on newer runner Podman
## CI Issue

The **Podman arm64 Lane** and **Podman Rootless Lane** intermittently fail at their first real step ("Assert the … Podman stack this lane depends on") with:

```
Error: template: info:1:7: executing "info" at <.Host.RootlessNetworkCmd>: can't evaluate field RootlessNetworkCmd in type *define.HostInfo
##[error]Process completed with exit code 125.
```

The failure is runner-image dependent: some runners carry a Podman build where `podman info` no longer exposes `.Host.RootlessNetworkCmd`, so the Go template evaluation aborts, and with `set -euo pipefail` the assignment kills the step before any lane assertion runs. Lanes fail on both PRs and v4/v5 pushes.

## Evidence

- Podman arm64 Lane, v4 push (merge of #5671): run 33639694750 — exit 125 at `rootlessnet="$(podman info --format '{{.Host.RootlessNetworkCmd}}')"` (.github/workflows/podman-arm64-lane.yml:154)
- Podman Rootless Lane, v5 push (merge of #5634): run 33638532288 — same error (.github/workflows/podman-rootless-lane.yml:104)
- Also failing PR runs 33639898884, 33639898927, 33639868229, 33639868287, 33639861149 (2026-09-02 14:06Z window)
- Sibling runs minutes earlier/later on other runners succeed — mixed runner Podman versions (recorded map value: 5.8.4)

## Recommendation

`RootlessNetworkCmd` is informational only in both lanes — it is printed in the log/step summary and never asserted. Make its query tolerant of absence:

```bash
rootlessnet="$(podman info --format '{{.Host.RootlessNetworkCmd}}' 2>/dev/null || echo 'unavailable')"
```

in both `.github/workflows/podman-arm64-lane.yml` and `.github/workflows/podman-rootless-lane.yml`, so the real lane assertions (arch/rootless/cgroups) still run on every runner image.

---
---

### LINKED ISSUE #5683: [scanner] CI: rootless + arm64 Podman lanes fail 125 on every PR — podman info no longer has .Host.RootlessNetworkCmd
## Finding

The **rootless startup, exit-77, and egress gate** and **arm64 image pull and service startup** lanes fail with exit 125 on every current PR (observed on #5660, #5663, #5664). The environment-assert step runs:

```bash
rootlessnet="$(podman info --format '{{.Host.RootlessNetworkCmd}}')"
```

Runner Podman (current 5.8.x-era runner image) removed the `RootlessNetworkCmd` field from `podman info`, so the probe fails:

```
Error: template: info:1:7: executing "info" at <.Host.RootlessNetworkCmd>: can't evaluate field RootlessNetworkCmd in type *define.HostInfo
##[error]Process completed with exit code 125.
```

Under `bash -e` this kills the step and the whole lane. The value is informational only (printed to the step summary); no assertion depends on it.

Affected files:
- `.github/workflows/podman-rootless-lane.yml` line ~104
- `.github/workflows/podman-arm64-lane.yml` line ~154

## Recommendation

Make the probe tolerant:

```bash
rootlessnet="$(podman info --format '{{.Host.RootlessNetworkCmd}}' 2>/dev/null || echo n/a)"
```

I cannot open this fix as a PR: the GitHub App push is rejected with `refusing to allow a GitHub App to create or update workflow ... without workflows permission`. A human with workflow-write must land this (or grant the App `workflows` permission).

---
*Filed by scanner agent (ACMM L5 — hold-gated mode)*

— hive: agent=scanner backend=copilot model=claude-fable-5 copilot=1.0.78

### LINKED ISSUE #5684: [scanner] podman rootless + arm64 CI lanes fail exit-125: Podman 5.x removed .Host.RootlessNetworkCmd
## Finding

The **rootless startup, exit-77, and egress gate** and **arm64 image pull and service startup** lanes fail on every PR with:

```
Error: template: info:1:7: executing "info" at <.Host.RootlessNetworkCmd>: can't evaluate field RootlessNetworkCmd in type *define.HostInfo
##[error]Process completed with exit code 125.
```

The preflight step in `.github/workflows/podman-rootless-lane.yml` (line 104) and `.github/workflows/podman-arm64-lane.yml` (line 154) runs:

```bash
rootlessnet="$(podman info --format '{{.Host.RootlessNetworkCmd}}')"
```

Newer Podman on the runners removed `RootlessNetworkCmd` from `HostInfo`, so the template errors and, under `bash -e`, the step dies with exit 125 before any real assertion runs. The value is informational only (printed to the step summary). Example failures: run 33644002927 job 100293927192 (arm64), run 33644003057 job 100293928408 (rootless) on PR #5649.

`bin/hive-podman-preflight-ids.sh:288` already guards its identical probe with `if ! helper="$(_pfi_info ...)"` — the workflows never got the same treatment.

## Recommendation

Probe tolerantly in both workflows:

```bash
# Informational only: Podman >= 5.x dropped .Host.RootlessNetworkCmd
# from HostInfo, and an unknown field makes `podman info` exit 125.
rootlessnet="$(podman info --format '{{.Host.RootlessNetworkCmd}}' 2>/dev/null || echo 'n/a')"
```

I have this fix committed and ready, but **the hive GitHub App lacks the `workflows` permission** — pushes touching `.github/workflows/*` are rejected (`refusing to allow a GitHub App to create or update workflow ... without workflows permission`), so a human must apply this two-line-per-file change (or grant the App the workflows permission).

---
*Filed by scanner agent (ACMM L5 — hold-gated mode)*
---
🐝 **Hive Agent**: `scanner` | **Instance:** `hosted-available-oke-11-placeholder-r05x` | **SHA:** `unknown`

— hive: agent=scanner backend=copilot model=claude-fable-5 copilot=1.0.78

clubanderson:
## What broke

The `arm64 image pull and service startup` check went red on every PR (exit 125):

```
Error: template: info:1:7: executing "info" at <.Host.RootlessNetworkCmd>: can't evaluate field RootlessNetworkCmd in type *define.HostInfo
```

The GitHub `ubuntu-24.04-arm` runner image rolled a newer Podman whose `podman info` schema no longer has `.Host.RootlessNetworkCmd`. A Go template hard-fails the ENTIRE `podman info` call the moment any field it names leaves the schema — so a **report-only** field (it was only printed into the env table, never branched on) took the whole lane down fleet-wide. `podman-rootless-lane.yml` carries the identical read and dies the same way the moment the amd64 runner image rolls the same Podman.

Fixes #5678
Fixes #5683
Fixes #5684

## The fix

The environment-assert steps in all three Podman lanes now make **one `podman info --format json` read** and extract fields with `jq` `//` fallbacks. Schema drift on a reported field degrades to `unknown` in the report instead of killing the lane; every load-bearing assertion still fails loudly. Per the #5388 discipline the assertion is kept — and moved onto the property the lanes actually need:

- **Rootless network provider**: what `RootlessNetworkCmd` named (pasta vs slirp4netns) is now asserted as a capability — `host.rootlessNetworkCmd` when present, otherwise derived from the `host.pasta.executable` / `host.slirp4netns.executable` tool entries. If NO provider exists while rootless, the step fails: rootless containers would start with no network, and the startup probe depends on one. The field name can churn per release; the capability cannot.

## Other spots hardened in the same pass (same failure class)

- **`podman-rootless-lane.yml`** — same six per-field template reads incl. `RootlessNetworkCmd`; same JSON+jq conversion, same provider capability assertion (this lane's whole matrix runs rootless).
- **`podman-rootful-lane.yml`** — five template reads incl. `.Store.GraphRoot`, which is load-bearing: it feeds the post-run store-isolation check. Converted to JSON+jq; a missing `store.graphRoot` now fails the step explicitly rather than handing the isolation check garbage.
- **4.x version floor** (rootless + rootful lanes) — an unreadable version now fails the floor check with a clear error instead of surfacing as a bash `integer expression expected` error under `set -e`.
- **arm64 built-image arch check** — `podman image inspect --format '{{.Architecture}}'` → JSON + `jq -r '.[0].Architecture // "unknown"'`, so a rename fails the value comparison (`unknown != arm64`), not the template.
- **`src/deploy/probe_podman_rootless_netadmin.sh`** — its combined one-line env report used a single template naming six fields incl. `RootlessNetworkCmd`; one missing field lost ALL diagnostics at once. Now field-by-field with per-field `unknown` fallback and the same pasta/slirp4netns derivation — deliberately without a jq dependency so the probe stays runnable on a laptop.

Left alone: `probe_podman_{rootful_netadmin,ipv6_egress,selinux_avc}.sh` and `qualify_podman_selinux.sh` already guard their single template reads with `2>/dev/null || true` (or equivalent) and degrade gracefully; the rootful lane's `podman ps --format '{{.Names}}'` reads a container-list template that is not schema-drift-prone the same way and is already `|| true`-guarded.

## Validation

- YAML parse on all three workflows
- `bash -n` on every extracted `run:` block and the probe script
- `shellcheck` clean at warning level on the extracted blocks; probe script fully clean
- Changelog entry as a `changelog.d/` fragment (`fixed-5678-podman-info-schema-drift.md`) per the new fragments system (#5675)


clubanderson:
The first push's own CI caught a bug in the hardening, fixed in the amended commit: the rootful lane failed with `rootless=unknown` even though Podman 5.8.4 reports the field fine. The jq path was correct — the operator was not. jq's `//` alternative treats `false` as falsy, so `.host.security.rootless // "unknown"` turned a correct `rootless=false` (the one value the ROOTFUL lane requires) into `"unknown"` and fed it to the load-bearing mode assertion. Only this lane needs `false`, which is why only it went red.

Fix, in all three lanes: the boolean is extracted type-checked (`if type == "boolean" then tostring else "unknown" end`) so only a genuinely absent field degrades — and when it IS absent, the lanes now probe the property instead of failing on the fallback string: rootful derives ROOTFUL from `sudo -n id -u` = 0 plus the default `/var/lib/containers` store (an explicit `rootless=true` remains a hard failure — that is a real mode swap, not schema drift); the rootless and arm64 lanes derive ROOTLESS from a non-zero euid without sudo. Every other extraction is a string, where `//` is safe (verified against the run's own env table: 5.8.4, v2, netavark, crun, and store.graphRoot all read correctly).

Issue #5685, opened 2026-09-02
🐛 bug: an attached-but-unused terminal tab permanently blocks API-error recovery — tmux client_activity is read as 'a human typed'


Danathar:
## Summary

A contributor relay recovers from a retryable API error by typing `try again` into the
agent's pane. It correctly refuses to do that while a human is sitting at the pane —
nobody wants a watchdog typing over someone mid-keystroke. It decides whether a human
is there by asking tmux for `#{client_activity}`.

**That field does not mean "a human typed."** It means "this client sent us bytes," and
a terminal emulator sends bytes for reasons that have nothing to do with a person:
replies to the escape sequences the running application writes (capability queries,
colour queries, cursor-position reports), and — where enabled — mouse and focus events.
A Claude Code TUI issues such queries on its own schedule, so an *attached but unused*
terminal tab keeps `client_activity` advancing indefinitely.

The consequence is that automatic recovery never runs for the most ordinary setup there
is. `just contribute-hive <backend> local` prints an attach hint, contributors leave the
session attached in a terminal tab to watch it, and from then on every retryable API
error parks the task until a person notices and bumps it by hand. The relay logs its
reasoning each time, so the failure is at least visible — but it never resolves.

Blast radius: interactive mode, any backend, any contributor whose tmux session is
attached. Work is not lost, but a task stalls until a human intervenes, which is exactly
the unattended-operation property the retry logic exists to provide.

## Observed, 2026-09-02

Contributor `Danathar`, claude backend, working issue #5681. The agent hit
`● API Error: Connection lost mid-response.` — which classifies correctly, it is in
`TRANSIENT_API_ERROR_PATTERNS` — and then:

```
[11:07:47] Task ct-…-5681-… stopped on a retryable API error; someone is active on hive-claude-7645 (last input 136s ago), so not typing a retry
[11:09:47] Task ct-…-5681-… stopped on a retryable API error; someone is active on hive-claude-7645 (last input  75s ago), so not typing a retry
[11:11:47] Task ct-…-5681-… stopped on a retryable API error; someone is active on hive-claude-7645 (last input 195s ago), so not typing a retry
[11:13:47] Task ct-…-5681-… stopped on a retryable API error; someone is active on hive-claude-7645 (last input  45s ago), so not typing a retry
```

Four consecutive ticks over six minutes, each recomputing a fresh "last input" age. The
contributor confirms they never typed into that pane — it is a background tab in their
[...]

Issue #5654, opened 2026-09-02
🐛 bug: Claude Code's silent API retry reads as IDLE_COMPLETE — a stalled agent can be booked complete mid-turn


Danathar:
## Summary

When Claude Code loses its connection to the API mid-turn it does not print an
error — it retries internally and renders a countdown line:

```
✻ Waiting for API response · will retry in 1m 57s · check your network
```

The agent is very much still on the task. The contributor relay reads that pane as
**finished**.

The relay classifies a Claude pane by asking three questions in order: is it busy,
is it showing an API error, and — failing both — does it show idle chrome plus a
completion marker. The retry state answers "no" to the first two and "yes" to the
third, so `classifyTmuxPane` returns `IDLE_COMPLETE` for an agent that is waiting on
a retry. From there the relay either books the task complete off a `HIVE_VERDICT:`
line left on screen from earlier, or, after `CHROME_IDLE_GRACE_TICKS`, completes it
as `chrome_idle` — in both cases reporting a task as done while the agent is still
mid-turn, which makes the hub revoke the lease, book the cooldown and offer the
issue to someone else.

Blast radius: the interactive `claude` backend, on any host whose network hiccups
long enough for Claude Code to enter its retry loop. Reported by a contributor who
sees this state regularly ("most of the time it recovers, but every now and then it
doesn't and needs a bump").

Honest scope note: the classification path below is confirmed by reading the code
against a pane observed live today. I do **not** have a confirmed booking traceable
to this specific cause — the mis-completions seen today are attributable to #5650's
undelivered-prompt path. This is filed as a live hazard with a demonstrated input,
not as a diagnosed incident.

## Why each gate misses it

`bin/contributor-relay.sh`, in `classifyTmuxPane` order:

1. **Busy?** The Claude busy marker is `esc to interrupt`. Claude Code does not
   render it while waiting on a retry, so `claudeBusyMarker` is false. The
   persistent footer (`⏵⏵`, `← for agents`) is still drawn, so `hasIdlePrompt` is
   true and the activity-verb fallback is skipped.

2. **A recognised API error?** No.
   ```js
   const TRANSIENT_API_ERROR_PATTERNS = [
     'connection lost mid-response', 'connection error',
     'request timed out', 'overloaded_error',
   ];
   const TRANSIENT_API_ERROR_STATUS_RE = /\b(?:500|502|503|529)\b/;
   ```
   `Waiting for API response · will retry in 1m 57s · check your network` matches
[...]

Issue #5281, opened 2026-08-31
✨ feature(contributor): nudge an unattended agent to proceed autonomously when it blocks on a question, instead of burning 20-30 min to the backstops


Danathar:
## Summary

When a contributor agent pauses mid-task to ask a question — "which approach should I take?", a numbered menu, a y/N confirmation — the relay detects it and reports `blocked_on_human` with an attention flag. That is where recovery ends. If no human is watching the tmux pane or the dashboard badge, the task sits until the 20-minute pane-stall backstop or the 30-minute max-duration ceiling reclaims it. **For a contributor run by a user who never attaches to tmux at all — a supported and expected way to run one — every question the agent asks costs 20–30 minutes and a failed task, even though the agent was explicitly instructed to work autonomously and could have answered the question itself.**

Proposal: when the pane is blocked on a *question* (not a login/credential/trust prompt), no client is attached, and the relay has not already done so this task, type a one-shot reminder — "no human is available; proceed autonomously with your best judgment" — then fall back to today's behavior. This is not automation of approvals: permission prompts are already handled by `dontAsk` (they fail fast rather than hang, #5147/#5169), and prompts that genuinely need a person stay untouched.

Blast radius: contributor relay only, the interactive (tmux) path. Headless mode has no interactive prompts and is unaffected.

## The gap, concretely

The relay's layers today, for a `claude local` contributor:

1. **Permission prompts** never hang: `--permission-mode dontAsk` auto-denies anything not pre-allowed, and #5169 pre-allows workspace `Edit`/`Write`. Fail-fast, correct.
2. **Questions** are detected by `paneLooksBlockedOnHuman` (`bin/contributor-relay.sh:1682`) — trailing `?`, numbered menus, y/N, elicitation forms — and reported as `blocked_on_human` + `attention: true` (`:2570-2582`).
3. **Bounds**: the stall backstop (20 min of unchanged pane, confirmed over ticks) or `MAX_TASK_DURATION_MS` (30 min) eventually fails the task.

Layer 2 is detection without recovery. The `attention` flag only helps a user who is watching something. The task prompts already tell agents to decide autonomously (e.g. "If you determine there is genuinely NOTHING shippable … print HIVE_VERDICT: no_work_needed"), so an agent that stops to ask is an agent that forgot its standing instructions — and a one-line reminder is usually all it takes. A human types exactly that today; the relay should be able to when no human is there.

## Why not auto-accept mode instead

[...]

Issue #5277, opened 2026-08-31
🐛 bug(contributor): an attached dashboard tab silently disables API-error auto-retry, stranding the task for 30 minutes


Danathar:
## Summary

When a contributor agent's turn dies on a transient API error (`● API Error: Connection lost mid-response`), the relay is supposed to type `try again` into the tmux pane and recover the task — up to 3 retries, per #5094/#5121. **That recovery is silently disabled for as long as anything is attached to the agent's tmux session**, including a dashboard terminal tab someone left open. The relay reports `blocked_on_human` instead and waits for a person who may not be watching.

The guard exists for a good reason — a watchdog must never type over someone who is actively using the pane — but it tests the wrong thing. `tmux list-clients` being non-empty means a client is *connected*, not that a human is *present*. Someone who opened a dashboard tab an hour ago and walked away is indistinguishable from someone mid-keystroke.

Observed live on this machine today: a contributor hit the error, no `try again` was ever typed, and a human had to hand-type a recovery instruction. Blast radius: contributor task recovery for every backend, whenever a terminal or dashboard tab is attached.

## Evidence

Live contributor, tmux session `hive-claude-6368`, relay PID 2939716 (host mode, both started 08:26 on 2026-08-31). The session had an attached client the whole time:

```
$ tmux list-clients -F '#{client_name} session=#{client_session} activity=#{client_activity} flags=#{client_flags}'
/dev/pts/1 session=hive-claude-6368 activity=1788181885 flags=attached,focused,UTF-8
```

Pane scrollback across the failure, verbatim and in order:

```
● Finishing the interrupted #5111 task (branch was already pushed and verified), then starting on #5117.

● API Error: Connection lost mid-response. The response above may be incomplete.

✻ Sautéed for 1m 22s

❯ complete the old task you were comitting before proceeding to the new task
```

The line after the duration summary is **a human typing a recovery instruction**. The relay's nudge is the literal string `try again` (`TRANSIENT_API_ERROR_NUDGE_MESSAGE`, `bin/contributor-relay.sh:202`), and it does not appear anywhere between the error and the human's message. Recovery came from a person, not from the automation built to provide it.

Note the secondary damage in the first line: the interrupted turn had already moved on to the next task (#5117) with #5111's PR unopened, so the human's redirect had to re-establish what the agent was in the middle of.

## Mechanism

[...]

PR #5700, opened 2026-09-02
🐛 fix(relay): stop reading tmux client_activity as "a human typed" (#5685)

Danathar:
## Summary

An attached-but-unused terminal tab permanently blocked automatic API-error recovery.

When an agent stops on a retryable API error the relay recovers by typing `try again` into its pane, and it correctly refuses to do that while somebody is sitting there. It decided somebody was there by asking tmux for `#{client_activity}`.

**That field does not mean "a human typed."** It means "this client sent us bytes", and a terminal emulator sends bytes for reasons that have nothing to do with a person: replies to the capability, colour and cursor-position queries the running application writes, plus mouse and focus reports where enabled. A Claude Code TUI issues such queries on its own schedule, so an *attached but unused* tab keeps `client_activity` advancing indefinitely.

That is the most ordinary setup there is — `just contribute-hive <backend> local` prints an attach hint, contributors leave the session attached in a tab to watch it — so every retryable API error parked until a person noticed and bumped it by hand. Which is precisely the unattended-operation property this recovery path exists to provide.

## Two changes, deliberately independent

**1. Corroborate the signal with an observation.** `client_activity` becomes a *necessary* condition rather than the whole answer, checked against whether the pane actually **changed** since the previous tick.

> A keystroke draws something. A terminal answering a query does not.

So an unchanged pane is positive evidence that the bytes tmux counted were the terminal talking, not a person. The converse is weaker — a pane can change because the CLI redrew something of its own — so that direction stays cautious and is bounded by (2) rather than by being made cleverer.

**2. Cap the deferral.** Whatever presence says, stop deferring after `HIVE_HUMAN_PRESENCE_MAX_DEFERRALS` ticks (3, ~6 minutes).

This is the part worth arguing for on its own. This signal has now been wrong **twice in the same direction**: #5277 established that "attached" is not "present", and its fix replaced one proxy with a slightly better proxy. A task parked forever on a signal the relay cannot verify is worse than one `try again` landing next to a human, who can see it and say so. The cap bounds the *next* wrong proxy, not just this one.

Together these are options (1) and (4) from the issue, which is what it recommends.

## Evidence this is the cause

From the issue, confirmed causally by the reporter — four consecutive deferrals over six minutes with nobody typing, `mouse` off and `focus-events` off, released the instant `tmux detach-client` ran (same relay, same task, same unresolved error, no keystrokes):

```
[11:07:47] … someone is active on hive-claude-7645 (last input 136s ago), so not typing a retry
[11:09:47] … someone is active on hive-claude-7645 (last input  75s ago), so not typing a retry
[11:11:47] … someone is active on hive-claude-7645 (last input 195s ago), so not typing a retry
[11:13:47] … someone is active on hive-claude-7645 (last input  45s ago), so not typing a retry
           ← tmux detach-client
[11:15:47] Transient API error on ct-…-5681-… — sending retry 1/3
```

Gaps of ~3 and ~4.5 minutes: neither human typing nor a fixed poll.

## The log line

It no longer says `someone is active … last input Ns ago`. That wording stated the assumption as fact, and the issue is right that it is what made this hard to see — the message asserted a human had typed, so the natural reading was to doubt the contributor rather than the field. It now reports the evidence and which branch was taken.

## Cluster claimed (files/functions)

- `bin/contributor-relay.sh` — `HUMAN_PRESENCE_MAX_DEFERRALS`; `paneEditedSincePresenceCheck` + its per-task state and `resetHumanPresenceEvidence` (wired into `resetTransientNudgeState`); the branch rework and log wording in `handleTransientAPIError`; four test exports.
- `bin/contributor-relay.test.js` — a `#5685` block (5 tests) plus one fixture change, below.
- `src/docs/contributor-relay.md`, `CHANGELOG.md`.

`tmuxSessionHumanPresence` is **unchanged** — it is the raw tmux reading, and #5277's tests read it directly. The corroboration sits at the consumer, where the decision is actually made.

`paneHasPresentHuman()` (the #5281 autonomy-nudge seam) is deliberately **not** touched. It is still the stricter bare-attached check, and its comment says #5277 was meant to update it — a real loose end, but a different nudge with a different budget, and not this bug.

### One existing test changed, and why

`#5277 a suppressed nudge still consumes no retry budget` held the pane **constant** while asserting a human was present. That fixture is now known to describe an attached-but-unused terminal — it is the bug — so it moves to a pane that changes each tick, which is what someone composing looks like. Its assertions are unchanged: refusing to type must not spend a retry the agent never got. This is the "today's tests can only exercise the proxy, which is why this passed" line in the issue, made concrete.

### Validation

- `node --check` (via CI's exact copy-to-`.js` step), `node bin/contributor-relay.test.js`, and `bash bin/contributor-agent.test.sh` — the whole "Contributor Relay Syntax + Tests" CI job — all pass locally. **261/261**, up from 256.
- **Mutation-checked, three ways**, so neither mechanism passes vacuously and the guard is proven to still guard:

  | Mutation | Result |
  |---|---|
  | Force `paneEdited` true (revert to the pure `client_activity` proxy) | only `#5685 an attached-but-unused terminal no longer blocks recovery` fails |
  | Remove the deferral cap | only `#5685 no presence signal can park a task forever` (+ its per-task sibling) fails |
  | Never defer at all (the over-correction) | **9 tests fail**, including the pre-existing `#5094 with a human attached …` and four `#5277` tests |

- Note for reviewers running the suite locally: `HIVE_AGENT_CWD` in the environment fails one unrelated pre-existing test (`a relaunch cds somewhere resolvable`). `env -u HIVE_AGENT_CWD node bin/contributor-relay.test.js` is clean on `v5` before this change too.

### Docs

The issue asks for one thing to be stated in the fix, and it is: **the dashboard's `/terminal/?arg=…` proxy is not a read-only viewer.** It attaches a real client and closes the gate exactly as a local `tmux attach` does, which matters because that is the path the docs otherwise steer operators toward. `contributor-relay.md`'s "attach to it to watch" — the sentence that leads straight into this bug — now points at a table of the three ways to watch and what each costs, with `tmux capture-pane -p` as the one that registers no client.

### Not claimed

Options (2) and (3) from the issue. (2), a longer quiet window, is tuning a proxy that is wrong *in kind* and would still misfire on a terminal that polls more often. (3), `HIVE_ASSUME_UNATTENDED=1`, is an escape hatch that puts the burden on the person least placed to know the mechanism — and (1)+(4) make it unnecessary.

## Related issues

Refs #5685. Related: #5277 (the same lesson one level up), #5094 and #5121 (the recovery this gate disabled), #5281 (the other nudge, untouched), #5654.

## Testing

- [x] `node --check` + `node bin/contributor-relay.test.js` + `bash bin/contributor-agent.test.sh` — the full relay CI job; 261/261.
- [x] Go build/test not applicable — no Go changed.


clubanderson (approved):
Reviewed the relay changes against #5685. The patch no longer treats tmux client_activity alone as human input: it corroborates with pane changes, resets the evidence per task, and caps deferrals so recovery cannot park forever. Validation: cd src && go build ./... passed; node --test bin/contributor-relay.test.js passed (261/261). Full go test ./... hit the existing Darwin /private/var vs /var cmd/bd path expectation, unrelated to this PR.

Danathar:
Branch targeting check — this PR is based on **v5**, but the defect it fixes is live on **v4**, which is the repository's default branch and what deployed hives and contributor relays actually run.

Evidence:
- `origin/v4:bin/contributor-relay.sh` still contains 4 occurrences of `client_activity`, i.e. the presence gate this PR replaces is unfixed there.
- The relay that exhibited #5685 (contributor `Danathar`, reproduced and confirmed by A/B detach today) runs a v4 checkout.
- v5 is not ahead of v4 — it has diverged: 65 commits on v5 that v4 lacks, **241 on v4 that v5 lacks**. Notably v5 does not contain #5674, so a relay moved to v5 to pick this up would reacquire the readiness bug that #5674 just fixed.

Same applies to #5688 (merged to v5, fixing #5681): `loadLeases` is present on v5 and absent on v4, and the v4-deployed hub lost an in-flight contributor task to a restart again today at 16:38Z for exactly that reason.

Not asking for a particular resolution — retarget, cherry-pick, or 'v5 is where this belongs and v4 gets it at the merge' are all reasonable and the maintainers know which. Flagging only because both fixes are for defects observed on v4 and currently land where the affected code isn't.

Issue #5698, opened 2026-09-02
[v5] RFC: backend capacity, model inventory, and placement


hanthor:
> **Retitled and rewritten.** The first draft led with an implementation and a
> cute codename. Both were wrong. What matters is the four capabilities below —
> and the one that makes the rest age well is that **nothing about models can be
> hardcoded**. Models ship faster than any table in any repo survives.

## Summary

Hive's governor decides **when** to kick an agent. Nothing decides **which
backend and model it should be pointed at**, or **whether that backend can
serve right now**. So a hive with a dry DeepSeek balance, a capped Codex
account, or a lapsed Claude session keeps kicking agents into backends that
cannot answer. They fail quietly and look healthy.

We have run an external component covering this for ~6 weeks. This RFC is not a
request to merge it — it is 1400 lines of bash against hive's HTTP API. It is a
report on the four capabilities the problem actually decomposes into, and
evidence for each.

### 1. Usage probe — normalized headroom per backend

"How much is left" has a different answer shape per provider, and none is a
token count:

| Backend | Where headroom lives | Shape |
|---|---|---|
| claude | `GET /api/oauth/usage` with the OAuth token | percent + `resets_at`, **three limits** |
| codex | `/status` typed into a live TUI | "NN% left, resets HH:MM on DD Mon" |
| agy | `agy --print /usage` | two pools × weekly/5-hour, percent remaining |
| deepseek | `GET /user/balance` | dollars of prepaid credit |

**A model-scoped cap is not provider exhaustion.** Anthropic returns `session`,
`weekly_all`, and `weekly_scoped` — the last carries `scope.model` and caps one
model class. Taking the max across all three parked our entire fleet while
`claude-sonnet-5` and `claude-opus-5` both answered normally. Headroom must
come from **unscoped** limits, with scoped caps reported separately so
placement can avoid *that model* rather than the whole provider.

**Probes that read a provider through a consumer of it will deadlock.** This
bit us four separate times: google required an unpaused agent though its probe
is headless; openai skipped paused agents though a paused agent's CLI is still
running; contributors parked every worker and then could not measure the
provider needed to restore them; an emptied credential file reported "unknown"
instead of "exhausted". **If hive builds this, that is the invariant to design
against.**

### 2. Model inventory — what a backend can run *right now*

This is the capability we have least of, and the one that decides whether the
rest ages well. Today we hardcode model IDs in a table. That table is wrong the
week a model ships.

[...]

Issue #5691, opened 2026-09-02
[v5] RFC: constellation — what a hub needs to coordinate several spokes as one fleet


hanthor:
## Summary

A single hive serving a diverse org makes its agents compete over unrelated
work. Splitting into several subject-scoped hives fixes that, but the moment
you do, you discover the hub is a **directory, not a coordinator** — and
several things that ought to be fleet-wide are per-hive by accident rather than
by design.

This RFC reports what actually broke splitting one 43-repo hive into two, and
proposes what a *constellation* — several spokes under one hub, acting as one
system — would need.

We are running this now: a self-hosted hub, two spokes, shared credentials.
Everything below is measured, not speculative.

## What we built

```
                    hub.tunaos.org        (HIVE_MODE=hub, registry)
                          |
        +-----------------+-----------------+
        |                                   |
  hive.tunaos.org                     reef.tunaos.org
  25 repos · ACMM L6                  18 repos · ACMM L5
  OS, images, installers,             desktop apps, app
  packaging                           distribution, docs
        |                                   |
        +--------- shared -------------------+
          GitHub App installation
          agent OAuth credentials
```

Spokes register themselves by heartbeat once they have `HIVE_HUB_URL` and the
hub's `HIVE_HUB_SECRET`. That part works exactly as documented.

## What the hub does NOT coordinate

Each of these is fleet-wide in reality but per-hive in the implementation.

### 1. Provider quota

Every spoke probes and consumes provider quota independently, with no shared
view. Two spokes on one Anthropic subscription each believe they have the whole
weekly cap; the first to notice exhaustion is the one that hits it.

This is the single biggest constraint on splitting. **A split creates zero
additional quota** — it divides the same pool among more agents. Ours:
DeepSeek dry, OpenAI capped for four days, Anthropic near its weekly cap,
Google carrying the fleet and hitting its own 5-hour window.

*Constellation need:* headroom as hub state, so a spoke can ask "what is left
for the fleet" rather than "what is left as far as I can see".

### 2. GitHub App rate limit

Both spokes use one App installation and therefore share its **7100 req/hr**.
Bringing up the second spoke exhausted it during initial enumeration:

```
403 API rate limit exceeded for installation ID 137498420 [rate reset in 19m22s]
... all 18 repos failed to enumerate
```

Nothing coordinates this. Each spoke polls its own repos on its own schedule
and discovers the shared ceiling by hitting it.

*Constellation need:* either a documented per-spoke budget, or hub-level
[...]

PR #5705, opened 2026-09-02
✨ feat(hub): detect repos claimed by more than one spoke (#5691 phase 1)

Danathar:
## Summary

Phase 1 of the constellation RFC (#5691): **the hub now detects when two spokes manage the same repo.**

Splitting one large hive into several subject-scoped spokes exists so agents stop competing over unrelated work. Nothing stopped two spokes claiming the *same* repo — which puts two fleets on one backlog opening competing PRs on it, the exact failure the split prevents. As the RFC puts it, the hub already receives every spoke's repo list on every heartbeat: **the data is there, unused.**

Reported live while splitting a 43-repo hive in two: an org watcher read one spoke's repo list, could not read the other's, concluded 18 repos were unmanaged, and was one step from adding them to the first hive.

## Why this piece, and only this piece

Both the RFC and @clubanderson's dispatch sequence overlap detection **first**, for the same reason: it is the only item that needs neither a heartbeat protocol change nor any control-plane semantics. The RFC calls it "cheap, high value … Pure read-side; no new protocol"; the dispatch says "No heartbeat protocol change required, and no scheduling/control-plane semantics."

Phases 2–6 are deliberately **not** here. Charters and headroom need heartbeat fields (and #5698's normalized headroom shape); routing is gated on 1–3 existing first. Nothing in this PR presumes any of them.

## What it does

- **`computeRepoOverlaps`** indexes `(host, owner/repo) → hives` over the registry entries the hub already holds. Derived per request, persisted nowhere.
- **`GET /api/registry`** gains `repoOverlaps`, **omitted entirely** when the fleet is clean so its presence is the signal:

  ```json
  "repoOverlaps": [
    { "host": "github.com", "repo": "tunaos/docs",
      "hives": [ {"id":"h-hive","name":"hive"}, {"id":"h-reef","name":"reef"} ] }
  ]
  ```

  It rides a response wrapper *embedding* `Registry` rather than a field *on* `Registry`, because `Registry` is what `saveRegistryNow` persists and derived state has no business on disk. The embedding flattens, so every existing field keeps its exact place and an old client sees no change.
- **A `repo-overlap` warning fleet alert** on *both* hives in a pair, naming the other spoke and the shared repos.

That last point is how phase 1 reaches the "hub UI" half without dashboard surgery: it rides the existing fleet-alert pipeline (ordering, acknowledgement, counting, chips). Like `ClusterHasKey`, it is stamped by `fleetAlerts` rather than derived in `alertHiveFromEntry` — an overlap is a relationship *between* hives, and re-deriving it per hive is how the two ends of one conflict would start disagreeing about whether it exists.

## The matching rules are the substance

Two spokes are configured **separately**, so "one lists bare names, the other lists paths" is the most likely shape for a real overlap — and exactly what a naive string compare misses. Each rule below has its own test:

| Compared | Result | Why |
|---|---|---|
| `docs` / `tunaos/docs` / `https://github.com/tunaos/docs.git` | same repo | bare, owner-qualified and URL forms all resolve to `owner/repo` |
| `TunaOS/Docs` vs `tunaos/docs` | same repo | GitHub owner/repo names are case-insensitive |
| `acme/ui` on `github.com` vs `github.ibm.com` | **not** a conflict | the host is part of the identity — different work |
| unreported host vs `github.com` | same host | matches `sameGitHubHost`, so an older spoke still matches a current one |
| a spoke listing its own primary in `repos[]` | one claim | the normal shape; getting this wrong would flag the entire fleet |
| an ownerless claim (`docs`, no org) | ignored | guessing an owner would manufacture conflicts between unrelated repos sharing a name |

Owner resolution delegates to `repoDisplayLine` rather than joining org and path by hand, so it inherits the doubling-safe rule and the live evidence behind it — a GHE hive recorded with org `castrojo.github.io` and primaryRepo `castrojo/endusers`.

## It reports; it does not refuse

Both fleets are still working, and a repo two subject areas legitimately share is a real case — the RFC's own example is a docs repo. So this is a **warning**, and the alert is acknowledgeable like any other.

Open question 2 on the RFC ("Should overlap be an error or a warning?") and the dispatch's first question (block vs. warn, with an allowlist) are **not** answered here. Answering them in code before they are answered in the issue would be the wrong order. The alert fires on *both* hives deliberately: either operator can act, and alerting on only one reads as that hive being at fault rather than the pair being ambiguous.

- `src/pkg/hub/registry_overlap.go` (new) — `RepoOverlap`/`OverlapClaim`, `computeRepoOverlaps`, `canonicalRepoClaim`, `repoClaimsOf`, `repoOverlapsFor`, `repoOverlapAlertReason`, `stripRepoRefHost`.
- `src/pkg/hub/server.go` — `registryResponse` wrapper; `handleRegistry` computes overlaps over the **filtered** set, so it never names a hive absent from the payload.
- `src/pkg/hub/alerts.go` — `AlertTypeRepoOverlap` + `knownAlertTypes` registration, the `alertHive.RepoOverlaps` field, the stamping in `fleetAlerts`, and the rule.
- `src/pkg/hub/registry_overlap_test.go` (new, 15 tests), `src/docs/hub-deployment.md`, `CHANGELOG.md`.

No existing behaviour changed: no field removed or renamed, nothing added to the heartbeat, nothing new persisted.

### Validation

- `go build ./...`, `go vet ./pkg/hub/`, `gofmt -l` clean on every touched file.
- `go test ./pkg/hub/` — the **whole** package passes (98s), including `TestKnownAlertTypesCoversEveryAlertType`, the guard that parses `alerts.go` and fails if an `AlertType*` constant is added without being registered for acking. That is not incidental: an unregistered type renders with an Acknowledge button that can never succeed (#2308), and an intentionally shared repo is precisely the case an operator will want to silence.
- **Mutation-checked** — every rule is independently pinned, so none of the tests passes vacuously:

  | Mutation | Caught by |
  |---|---|
  | Drop the host from the overlap key | `TestRepoOverlap_DifferentHostsAreNotAConflict` |
  | Remove the self-duplication guard | `TestRepoOverlap_TwoSpokesOneRepo` + 2 others |
  | Replace `repoDisplayLine` with a naive `org + "/" + path` | `TestRepoOverlap_MatchesAcrossSpellingsAndCase`, `_DoesNotDoubleTheOwner`, `_OwnerlessClaimsAreIgnored` |

- Ordering determinism is asserted over 20 runs, since the index is built from a map and churning output would make both the API diff and the alert text unstable.

### On the open questions

I have not answered any of them in code. Two are worth a maintainer's note before phase 2:

1. **Block vs. warn, and the allowlist.** This ships as warn-only. If an allowlist lands, the natural home is hub-side SaaSHive metadata rather than spoke config — a spoke should not be able to declare its own overlap acceptable.
2. **What identifies a shared quota pool** (dispatch question 2) matters for phase 3 but not here; the overlap key deliberately uses forge host + path only, and takes no position on it.

## Related issues

Refs #5691 (phase 1 of 6 — does not close the RFC). Related: #5698 (headroom shape phase 3 depends on), #5632, #5633, #5661, #5690.

## Testing

- [x] `cd src && go test ./...` — ran `./pkg/hub/` in full (the only package touched); passes.


Danathar:
**CI note: the one red check (`arm64 image pull and service startup`) is runner
infrastructure, not this PR.** Recording the evidence so nobody has to re-derive it.

The failure is in the lane's *environment assertion* step, before any repo code runs:

```
Error: template: info:1:7: executing "info" at <.Host.RootlessNetworkCmd>:
       can't evaluate field RootlessNetworkCmd in type *define.HostInfo
Process completed with exit code 125.
```

That is `podman info --format '{{.Host.RootlessNetworkCmd}}'` at
`.github/workflows/podman-arm64-lane.yml:156` — a field the arm64 runner's Podman no
longer exposes.

Three things pin it as environmental:

1. **This PR touches no CI at all.** `git diff --name-only upstream/v5..HEAD` is six
   files: four under `src/pkg/hub/`, one doc, and `CHANGELOG.md`. `git diff
   upstream/v5..HEAD -- .github/` is empty, so the failing workflow is byte-identical
   to `v5`.
2. **The identical template passes on x86.** `podman-rootless-lane.yml:106` runs the
   same `{{.Host.RootlessNetworkCmd}}` expression, and `rootless startup, exit-77, and
   egress gate` is green on this same run — as is `rootful`. So the field exists on the
   x86 runners and not on the arm64 one: a per-architecture Podman version difference.
3. **The lane already anticipates exactly this.** It pins `RECORDED_PODMAN: 5.8.4` and
   emits a drift notice when the runner disagrees, pointing at
   `src/docs/podman-ci-runner-map.md` — version drift on these runners is a known,
   tracked hazard.

Everything this PR could plausibly affect is green: `build-and-test`, `golangci-lint`,
`gosec`, `govulncheck`, `test (hub)` (the shard covering `pkg/hub`), all five agent
shards, the three `rest` shards, and the docs link check — 30 passing.

I have not touched the workflow to work around it. Fixing the arm64 runner map is a
separate concern from this RFC phase, and quietly editing an unrelated CI lane inside a
feature PR is how that kind of breakage stops being visible.

clubanderson (approved):

Issue #5704, opened 2026-09-02
🐛 bug: /api/contribute/activity ignores its limit parameter and always returns 50 events


Danathar:
## Summary

`GET /api/contribute/activity` accepts a `limit` query parameter and ignores it. Every
request returns the same fixed 50 events, whether you ask for 5 or for 300.

This matters because the endpoint is one of the few public windows onto what a hive's
contributors are doing, and it silently truncates. Anyone reasoning about a contributor's
history — how many times a session reconnected, whether a task ever reached a terminal
event — is looking at "the last 50 events across all contributors" while believing they
asked for more. On a hive with several active contributors that window can be under an
hour, and a flapping session fills it very quickly.

Blast radius: read-only endpoint, no behaviour change to task dispatch. It is a
correctness bug in an observability surface, and it produces confidently wrong analysis
rather than an obvious error.

## Reproduce

```
$ curl -s '<hive>/api/contribute/activity?limit=5'   | jq '.activity | length'
50
$ curl -s '<hive>/api/contribute/activity?limit=300' | jq '.activity | length'
50
$ curl -s '<hive>/api/contribute/activity'           | jq '.activity | length'
50
```

Observed 2026-09-02 against `hosted-kubestellar-hive-ojv6.hive.kubestellar.io`, serving
`a8619b0` and later.

## Why it is worth fixing rather than documenting

The parameter's presence is itself the problem. A caller who passes `limit=300` and
receives exactly 50 has no way to tell that from "there were only 50 events" — the
response carries no total, no truncation flag, and no cursor. I reached a wrong
conclusion on #5701 partly this way: I counted `joined` versus `left` events for one
contributor over what I believed was a full day, when I was in fact counting a 50-event
ring shared with every other contributor on the hive.

Either behaviour would be fine; the mismatch is not:

1. **Honour `limit`** up to whatever the retained history actually is, and say what that
   ceiling is — a `"truncated": true` field, or a documented maximum that a request above
   it clamps to visibly.
2. **Or drop the parameter** and document the endpoint as "the most recent 50 events",
   so callers know they are looking through a fixed window.

If the underlying store genuinely only retains 50 events, (2) is the honest fix and
costs nothing. If it retains more, (1) makes the endpoint useful for exactly the kind of
after-the-fact analysis these feeds exist for.

A cursor would be better still — `/api/contribute/events` already has a replay mechanism,
so the two surfaces could share pagination — but that is a larger change than this
report warrants.

## Note

[...]

Issue #5701, opened 2026-09-02
🐛 bug: a flapped contributor socket is never closed hub-side — no 'left' is ever recorded and the reader goroutine is orphaned


Danathar:
## Summary

When a contributor's WebSocket flaps, the hub never finds out. The relay sees its socket
die and reconnects a second later; on the hub the *old* connection is simply orphaned —
still open as far as it knows, its reader goroutine parked forever, and no disconnect
ever recorded.

The visible symptom is an activity feed that reads as impossible. A contributor
"enters the hive" over and over and never leaves:

```
2026-09-02T16:03:25Z | joined
2026-09-02T16:04:26Z | joined
```

Over one day on `hosted-kubestellar-hive-ojv6`, one contributor produced **21 `joined`
events and 1 `left`** — and that single `left` was a deliberate `SIGINT` shutdown, the
one close where the relay sent a proper close frame. A second contributor on the same
hive whose sessions ended cleanly shows 4 `joined` / 3 `left`. So the feed is accurate
for clean closes and silent for every flap.

Underneath that, a reader goroutine and its socket are leaked per flap — twenty of them
in a day from one relay.

Blast radius: every contributor on a hive where the socket can be cut without a close
frame, which #5090 documents as routine on hosted spokes. Nothing breaks for the
contributor: work continues, because the hub still holds the lease it never learned to
release. Operators lose the ability to see the disconnect at all, and the hub
accumulates parked goroutines.

## Evidence

Counts over 2026-09-02, from `GET /api/contribute/activity`:

```
hanthor      joined= 4  left= 3   (sessions ended cleanly)
Danathar     joined=21  left= 1   (flapping session; the 1 'left' is 12:50:36Z, a SIGINT)
```

Relay side, a representative pair — note it reconnects and resumes without complaint,
because from the hub's perspective it never went away:

```
[12:03:24] Connection closed (code=1006 abnormal closure — no close frame …). Reconnecting in 1000ms...
[12:03:25] Authenticated … as c-ca8187f7480c (tier: contributor)
[12:03:25] Reconnected while working on kubestellar/hive#5685 — resuming
[12:04:25] Connection closed (code=1006 abnormal closure — no close frame …). Reconnecting in 1000ms...
[12:04:26] Authenticated … as c-ca8187f7480c (tier: contributor)
[12:04:26] Reconnected while working on kubestellar/hive#5685 — resuming
```

Both hub surfaces agree with each other and disagree with reality: the polled
[...]

PR #5711, opened 2026-09-02
🐛 fix(dashboard): honour limit on /api/contribute/activity, and report the window (#5704)

Danathar:
## Summary

`GET /api/contribute/activity` accepted a `limit` query parameter and ignored it. `limit=5` and `limit=300` both returned the same fixed 50 events, because `handleContributeActivity` never read the parameter — it returned `RecentActivity()` unconditionally, and the ring is capped at `maxActivityEntries`.

**The harm is not the wrong count, it is that the caller cannot see it.** The response carried no total, no truncation flag and no cursor, so asking for 300 and receiving 50 was indistinguishable from *"there were only 50 events"*. That produces confidently wrong analysis rather than an obvious error — it is part of how a reader on #5701 came to count one contributor's events over what they believed was a full day while actually reading a 50-event window shared with every other contributor on the hive.

## The fix has two halves, because the complaint does

**1. `limit` now returns that many of the MOST RECENT events.** "The last N" is the only sensible reading, and it is what the dashboard's own field log already assumes with `acts.slice(-6)` — returning the *oldest* N would have silently broken that view. Oldest-first order is preserved **within** the window, so the array shape existing consumers parse is unchanged.

**2. The response reports the window.** New `retained` (events the hub currently holds, before limiting) and `capacity` (the ring size). These make the feed's boundedness visible, which the array alone can never show, and they separate three states that were previously identical:

| Signal | Means |
|---|---|
| `activity.length < retained` | this response was cut short by `limit` |
| `retained == capacity` | the ring is full — older events are evicted, so this is **not** the hive's full history |
| `retained < capacity` | the hive really has produced only this many |

That second half is the part I would argue is the actual fix. Honouring `limit` alone would still leave a caller unable to tell a bounded window from a short history, which is the specific failure the issue describes.

## On the issue's two options

The issue offers "honour `limit`" or "drop the parameter and document the fixed window", and the follow-up comment leans toward the latter — reasonably, since honouring a limit *above* 50 would need the ring to grow.

I went with the first, because **you cannot drop a parameter that was never implemented.** No code reads `limit`; "dropping" it would be a docs-only change, and `limit=5` would still return 50 — the exact confusion the issue is about. Honouring it costs ~10 lines, makes `limit=5` genuinely useful, and clamping above `capacity` sidesteps the ring-growth problem entirely: the endpoint states its ceiling rather than pretending to serve history the hub never kept.

**The ring is deliberately not grown here.** That is a separate memory/retention decision, and #5151's retention-pressure note is the place to argue it.

## Backward compatibility

By construction. An omitted, empty, non-numeric or non-positive `limit` returns everything retained — exactly what every existing caller got — following the same parse-and-fall-back convention as `handleLifecycleTimeline`. Neither in-tree consumer (the `/me` field log, the Operations poller) passes the parameter, and both read the whole feed. The new keys are additive. A test pins all six of those input forms against the pre-change behaviour.

- `src/pkg/dashboard/api_contribute.go` — `handleContributeActivity` parses `limit`; new pure helper `limitActivity`.
- `src/pkg/dashboard/contribute_activity_limit_test.go` (new, 6 tests).
- `dashboard/openapi.json` — documents the parameter and both new fields. It was previously silent on `limit`, which is part of why the mismatch went unnoticed.
- `CHANGELOG.md`.

### Validation

- `go build ./...`, `go vet ./pkg/dashboard/`, `gofmt -l` clean on both touched Go files.
- `go test ./pkg/dashboard/` — the **whole** package passes, including the OpenAPI route/schema parity guards.
- `dashboard/openapi.json` re-serialised with the file's original ASCII escaping, so the diff is **+22/-2** confined to this one endpoint rather than churning every em dash in the file.
- **Mutation-checked** — each behaviour is independently pinned, so no test passes vacuously:

  | Mutation | Caught by |
  |---|---|
  | Ignore `limit` again (restore the bug) | `_HonoursLimit`, `_LimitTakesTheNewestEvents` |
  | Take the oldest N instead of the newest | `_LimitTakesTheNewestEvents`, `TestLimitActivity` |
  | Report the *limited* length as `retained` (hides truncation) | `_HonoursLimit` |

- Tests drive the real handler through the real mux and assert the decoded wire response, not a helper's return value.

### Note on the issue's supporting claim

The follow-up comment retracts the idea that the ignored `limit` caused the #5701 miscount (it was `absorbReconnectFlapLocked` retracting flap rows by design). That retraction is correct and this PR does not depend on the claim. The reported bug reproduces on its own terms on `v5`, which I verified directly in the handler before writing anything.

## Related issues

Refs #5704. Related: #5701 (the wrong conclusion this surface contributed to), #5151 (retention pressure on the same ring).

## Testing

- [x] `cd src && go test ./...` — ran `./pkg/dashboard/` in full (the only package touched); passes.


clubanderson (approved):

Issue #5715, opened 2026-09-02
🐛 bug: a WebSocket flap aborts the PR review cycle — the relay resumes a synthetic task the hub never leased, and treats the revoke as terminal


Danathar:
## Summary

Every few tasks a contributor relay runs a **PR review cycle** — it stops taking new
issues and asks its agent to look over the open PRs it has filed, so review comments get
answered. That cycle is abandoned by any WebSocket flap, and on a hive where flaps are
routine it can almost never run to completion.

The cause is a mismatch about who owns the task. The relay **creates** the review task
itself, locally:

```js
// bin/contributor-relay.sh
console.log(`PR review cycle (${tasksCompletedCount} tasks completed) — checking open PRs`);
currentTask = { task_id: `pr-review-${Date.now()}`, kind: 'review', repo: completedRepo, number: 0,
                title: 'Review open PRs for comments' };
```

The hub never assigned it and therefore holds no lease for it. But on reconnect the relay
runs its normal resume path for whatever is in `currentTask`, so it re-asserts a task the
hub has never heard of. The hub does the right thing — a resume is honoured only against a
server-issued lease, which is the #C4 rule that stops a client asserting ownership of work
it was not given — and revokes it. The relay treats the revoke as terminal: it stops the
agent, relaunches the CLI, and asks for fresh work. The review is lost.

Blast radius: relay-side, all backends, interactive mode. No issue task is affected —
those hold real leases and resume correctly across the same flaps. Nothing is corrupted;
the cost is that PR review coverage silently does not happen.

## Observed, 2026-09-02

Contributor `Danathar`, v4 relay, hosted spoke. Full sequence, four minutes end to end:

```
[13:43:09] Task ct-kubestellar/hive-5691-… completed — signal=verdict (HIVE_VERDICT: complete)
[13:43:13] PR review cycle (5 tasks completed) — checking open PRs
[13:43:16] Task prompt sent to CLI
[13:47:09] Connection closed (code=1006 abnormal closure — no close frame …). Reconnecting in 1000ms...
[13:47:10] Reconnected while working on kubestellar/hive#0 — resuming
[13:47:10] Task revoked: pr-review-1788370993120 — no active lease for this task
[13:47:14] Relaunching claude after task revoke: …
[13:47:16] Task assigned: issue kubestellar/hive#5704 — …
```

Note `kubestellar/hive#0` in the resume line: the synthetic task carries `Number == 0`,
which the hub already special-cases in several places precisely because it has no work
[...]

Issue #4245, opened 2026-08-20
feat(worksource): carry source-aware task identity through scheduler and contributor admission


joshyorko:
Parent: #3845

Source: upstream `v4` at `380a4ad80cfe6ab65e85b63e84745dfeae545463`

## Invariant

Distinct work-source items never collapse onto the same task, admission, hold, cooldown, claim, or active-work identity, while GitHub-backed issues keep their byte-identical `repo#number` keys and existing persisted state remains readable.

## Current production path

- `src/pkg/worksource/worksource.go:Issue,TaskKey` already carries `SourceType`, `Repo`, `ExternalID`, and an optional GitHub issue `Number`; `TaskKey` preserves `repo#number` for positive GitHub-backed issue numbers and otherwise uses the external key.
- `src/pkg/worksource/adapt.go:ToGitHubIssues` drops `SourceType`, `ExternalID`, source state, and priority when it converts into `github.Issue`. Linear and Jira items therefore enter the legacy envelope with `Number == 0`.
- `src/cmd/hive/main.go:runEvalCycle` replaces `actionable.Issues` with that lossy projection for non-default work sources, then sends the same `github.ActionableResult` to `Scheduler.SetLastActionable`, `BuildKickMessages`, and dashboard status construction.
- `src/pkg/dashboard/status_builder.go:buildRepos` publishes those issues as `FrontendRepo.ActionableIssues`.
- `src/pkg/dashboard/contribute_sse.go:ReadyQueue` and `src/pkg/dashboard/contribute_ws.go:selectTask` read only the integer `number`, skip zero, and independently rebuild `repo#number` keys for hold, ordering, cooldown, failure, no-work, active-task, admission, assignment, and lease state.
- `src/pkg/scheduler/scheduler.go` renders internal-agent issue lists as `repo#number`; `issueRefsForAgent` omits `Number <= 0` entirely.
- Completion cooldowns, failure/quarantine state, operator hold/order state, and the PR-claim ledger persist existing GitHub keys across restart.
- `src/pkg/github/prclaims.go` and `src/pkg/dashboard/contribute_admission_deps.go` are intentionally GitHub-issue observers. They accept repository plus positive issue number; they cannot infer an external work item from `#0`.

## Precise remaining gap

[...]

PR #5718, opened 2026-09-02
🐛 fix(relay): stop claiming the PR review cycle as a hub-issued task (#5715)

Danathar:
## Summary

After every fifth completed task the relay stops taking new issues and runs a **PR review cycle** — it asks its agent to look over the open PRs it has filed, so review comments get answered. It builds that task **itself**:

```js
currentTask = { task_id: `pr-review-${Date.now()}`, kind: 'review',
                repo: completedRepo, number: 0, title: '...' };
```

The hub never assigned it and holds no lease for it, but every path that reports on `currentTask` kept asserting it. The hub honours a claim only against a server-issued lease (the #C4 rule that stops a client asserting ownership of work it was not given), so the only possible answer is a `task_revoke` — and the relay treated that as terminal: stop the agent, relaunch the CLI, ask for fresh work. The review is lost. On the reporting session (29 abnormal closes in under four hours) PR review coverage silently almost never happened.

## The flap is not required

The report frames this as a reconnect bug. The reconnect only makes it *immediate*.

`progressTick` reports on the same `currentTask`, so the **first ordinary progress report** after `TASK_GRACE_PERIOD_MS` (3 min) asserts the synthetic task and is revoked on its own. The observed timeline is exactly consistent with that:

| Time | Event |
|---|---|
| 13:43:16 | review cycle starts |
| 13:46:16 | 3-minute grace expires |
| 13:47:09 | flap — reconnect re-asserts, revoked |
| **13:47:16** | **the tick that would have killed it anyway, 7s later** |

So **fix (1) as the report describes it — skipping only the reconnect re-assertion — would have left every review dying a few minutes in.** That is why the guard goes somewhere else.

## Where the guard goes, and why

At the single `sendTo` funnel, not at the call sites. A claim can originate from the reconnect handler, the progress tick, the transient-API-error path and the autonomy nudge — **six sites today**, and a seventh added later would silently reintroduce this. `sendTo` withholds `task_accepted` and `task_progress` when the message's own `task_id` carries the locally-minted prefix.

`task_complete` / `task_failed` are deliberately **not** withheld: they assert nothing about ownership, the hub treats them as no-ops for a task it never issued (`hasTask` is false, so neither the lease revoke nor the completion booking runs), and they are the vehicle for the existing routing-fallback regression test, which keeps passing unchanged.

### Identity is the id prefix, not `number === 0`

The report suggests "the existing `Number == 0` / `task_id` prefix already identifies it". Only the prefix is safe. External work items — Linear, Jira — **legitimately carry number 0** and put their identity in `key`/`external_id` (#4245), and they hold **real leases**. Keying on the number would withhold the reports that keep those leases alive and get live work reclaimed as wedged — a worse bug than the one being fixed. A test pins this, exercising the reconnect payload specifically, since that is the one that actually carries `number`.

### Fix (2) is here too

A `task_revoke` for a locally-created task is now ignored. The terminal handling assumes the work belongs to somebody else, which cannot be true for a task the hub never issued. With the send guard this should be unreachable, but it keeps the cycle robust to a revoke arriving for any other reason — an older hub, a racing frame.

## A comment that argued the opposite

The comment above `send()` said that dropping the synthetic task's frames left *"the hub to watch the contributor go mute mid-review and time it out"*. It cannot: the real task's `task_complete` goes out **before** the review cycle starts, so the hub has already cleared its `currentTask` and revoked the lease and holds nothing to time out — hub-side liveness is the ping/pong heartbeat, independent of these frames. I verified that ordering in the source rather than assuming it, and corrected the comment rather than leaving it to contradict the code.

## The report's two follow-up questions

- **Does the relay report anything to the hub when a review cycle ends?** Yes — `task_complete` still goes out (a hub-side no-op), and the `ready` that actually matters is sent by the relay's own counter branch.
- **Does an aborted cycle advance `tasksCompletedCount`?** No. Neither the revoke handler nor `failCurrentTask` touches it, so an aborted review is **skipped, not retried**: the counter sits at the multiple and the next review waits for the following one. That is unchanged behaviour and I have not altered it — retry semantics are a design decision nobody has made — but it is now documented.

- `bin/contributor-relay.sh` — `LOCAL_TASK_ID_PREFIX`, `isLocalOnlyTaskId`/`isLocalOnlyTask`, `TASK_OWNERSHIP_MESSAGES`, the `sendTo` guard, the reconnect branch, the `task_revoke` branch, and the corrected `send()` comment.
- `bin/contributor-relay.test.js` — 6 new tests.
- `src/docs/contributor-relay.md`, `CHANGELOG.md`.

No hub-side change: this is entirely relay-side, and the hub's behaviour was correct throughout.

### Validation

- The full CI relay job passes locally: `node --check` (via CI's copy-to-`.js` step), `node bin/contributor-relay.test.js` — **267/267**, up from 261 — and `bash bin/contributor-agent.test.sh`.
- **Mutation-checked**, each mechanism independently pinned:

  | Mutation | Caught by |
  |---|---|
  | Remove the `sendTo` guard (restore the bug) | `#5715 a progress tick during a review cycle claims nothing from the hub` |
  | Remove the revoke guard | `#5715 a revoke of a review cycle does not stop the agent` |
  | Key the guard on `number === 0` (the report's suggestion) | the progress-tick test **and** `#5715 an external work item with number 0 is not mistaken for a local task` |

- A control test asserts a hub-issued task still re-asserts on reconnect and is still revocable, so the fix cannot pass by simply never resuming.

## Related issues

Refs #5715. Related: #5090 (why the flaps happen — unresolved), #5681 (the same resume path for real tasks, across a hub restart), #4245 (why zero-numbered external items are not synthetic).

## Testing

- [x] `node --check` + `node bin/contributor-relay.test.js` + `bash bin/contributor-agent.test.sh` — the full relay CI job; 267/267.
- [x] Go build/test not applicable — no Go changed.


clubanderson (approved):

PR #5723, opened 2026-09-02
📝 docs: use the atomic backend+model endpoint, and why two calls are a footgun (#5698 item 1)

Danathar:
## Summary

Item **1** of the #5698 RFC — *"Document the atomic `PUT /api/config/agent/{name}/models` as the preferred path over `switch` then `model` … Pure docs; removes a live footgun"* — and **phase 1** of @clubanderson's dispatch on that issue, which asks for the API contract before any of the capability work.

Three routes can change an agent's method or model:

```
PUT  /api/config/agent/{name}/models    {backend, model}   # atomic
POST /api/switch/{agent}/{backend}                         # backend only
POST /api/model/{agent}/{model}                            # model only
```

There is no transaction across the last two. Using them in sequence to change both leaves — if the second call fails, times out, or is never made — the new backend paired with the **previous** backend's model: `codex --model claude-opus-5`, which cannot launch. Nothing says so at set time; it surfaces later as a dead pane.

## Two findings that make this sharper than "prefer the atomic call"

I read the handlers rather than restating the report, and both of these change what the guidance should say:

**1. `POST /api/model` *does* validate — and that still doesn't save you.** `validateModelForAgent` checks the model against the agent's **current effective** backend, and deliberately fails **open** when that backend's model list cannot be enumerated. Codex is exactly such a backend, so `switch`-to-codex followed by `model`-set-to-a-Claude-id is *accepted*. Reverse the order and the model is validated against the backend you are about to leave. **Neither order yields a checked pair**, so the presence of validation does not make the two-call path safe — which is worth stating explicitly, because "but it validates" is the obvious reason to assume it is fine.

**2. The atomic route validates the backend but *not* the model id.** It calls `ValidateBackend` and then persists `body.Model` unchecked. That is the gap phase 2 of the dispatch closes, so the docs say so rather than implying the atomic route is fully checked.

## One thing neither the RFC nor the dispatch mentions

**The atomic route does not restart the agent.** `AgentMgr.UpdateConfig` only refreshes the stored config (`agent.Config = cfg`) — it does not relaunch — and `refreshAndPersist` does not either. So the running CLI keeps its current method until the agent next restarts, whereas `switch` and `model` each call `AgentMgr.Restart`.

That is usually what a scripted change wants (and is most of the ~2.4s vs ~45s the RFC measured — two restarts versus none), but an operator who needs the change to take effect *now* has to restart afterwards, and nothing said so. It is recorded in both surfaces.

### What changed

- `src/docs/agent-configuration.md` — a new subsection under *Methods*, with a comparison table of the three routes and the two consequences above. Placed there because that is where an operator reads about choosing a method and model.
- `dashboard/openapi.json` — the three descriptions now cross-reference each other: the atomic route is marked preferred for changing both and carries its known validation gap; the two single-field routes point at it and say why pairing them is not atomic.
- `CHANGELOG.md` — required, since `dashboard/openapi.json` is outside the changelog-reminder's docs-only exemption list.

Docs and API descriptions only; **no behaviour change**.

### Validation

- `go test ./pkg/dashboard/` passes in full, including `TestOpenAPISpecCoversEveryRegisteredRoute` and `TestOpenAPISchemaFieldsExistOnGoTypes` — the guards that would catch a spec edit that drifts from the served routes or invents fields.
- `dashboard/openapi.json` re-serialised with the file's original ASCII escaping, so the diff is **3 changed lines** confined to the three descriptions rather than churning every non-ASCII character in the file.
- Every claim in the docs is taken from the handlers on this commit (`handleAgentConfigModels`, `handleSwitch`, `handleModelSet`, `validateModelForAgent`, `Manager.UpdateConfig`), not from the report.

### Scope: why only item 1

Phases 2–6 of the dispatch are each gated on one of the five open questions still awaiting @hanthor — which inventories are authoritative enough to reject at set time, whether validation blocks edits or only flags existing configs, how tier floors are expressed, where benchmark rankings come from. Phase 1 is the only piece that needs no answer, and the RFC scoped it as docs for that reason.

Two notes for whoever picks up phase 2, from reading this code:

- The natural place for model validation is `handleAgentConfigModels`, but it cannot reuse `validateModelForAgent` as-is: that function resolves the backend from the *running process*, whereas the atomic route may be changing the backend in the same request. It needs to validate against the **incoming** backend.
- Its fail-open behaviour on non-enumerable backends is deliberate and load-bearing (`// cannot enumerate — do not block`). Open question 2 is really asking whether that stays fail-open per backend, and the answer wants to be per-backend rather than global.

## Related issues

Refs #5698 (item 1 of 5 — does not close the RFC). Related: #5691 (the constellation RFC, whose fleet-headroom field needs the same normalized shape capability 1 defines).

## Testing

- [x] `cd src && go test ./...` — ran `./pkg/dashboard/` in full (the only package whose fixtures this touches); passes.


clubanderson (approved):

Issue #5699, opened 2026-09-02
[v5] [fleet] remediation hints v2: cover agent-down/all-paused/provider-quota verdicts; investigate three never-firing detectors


clubanderson:
Follow-up from a 2026-09-02 fleet sweep against the shipped remediation-hints system (#5577 family). The hints work where they exist — 30 rendering, correct on all spot checks, none on green — but coverage has holes:

**1. Eleven non-green verdicts carry NO hint**, in three families the signature table doesn't cover:
- `N agent(s) down — restart needed` (3 spokes incl. one running supervisor-down for 2+ days) → hint should be 'restart the agent from its card' + link, and when the failure log matches the copilot-hung-at-startup signature, add 'consider pinning a model (auto-assignment may have selected a broken one)'.
- `all agents paused — resume to produce output` (7 spokes) → hint: 'resume agents on the dashboard, or mark the hive idle if mothballed'.
- `N agents out of provider quota` (1 spoke) → hint: 'rotate or raise the provider quota'.

**2. Three shipped detectors fire zero times fleet-wide**: consent-wedge, error-streak, no-cadence. Either genuinely nothing matches (unlikely — a known consent-wedge spoke now classifies as plain login-stuck, suggesting the wedge presents as a login prompt and the detector's restart-loop+consent_screen signature no longer matches), or the spoke-side emitters aren't populating the heartbeat fields on the deployed build. Verify each detector against a live spoke exhibiting the condition; fix signatures or wiring.

**3. Bonus observation**: 13 spokes sat uniformly 153 commits behind :stable (merged-not-deployed). The channel-lag hint fires correctly, but at this scale it suggests the hub auto-upgrade wave isn't completing — worth a look at the upgrade trigger cadence vs today's merge rate.

Refs #5577

clubanderson:
[...]

Issue #5577, opened 2026-09-01
[RFC] /fleet remediation hints — every non-green verdict names the fix, and silent failure classes get detectors


clubanderson:
## Problem

/fleet's health verdict (ACMM-banded, #4538/#4544/#4547/#4552, refined #5350) answers *is this hive stuck and roughly why* — but not *what to do about it*, and several real failure classes never surface at all. A 2026-09-01 fleet audit found every one of these only by exec'ing into pods:

- Budget exhausted (spend 3× limit) suppressing all kicks → showed as generic quiet (#5508)
- Model crash loop with green agents (bobshell-advertised model failing every call) → no chip (#5338)
- Copilot consent-screen wedge restarting an agent ~1/min → nothing
- Spoke-dashboard 'agent not producing output' warnings that name the symptom but not the cause (agents with zero cadence config, never kicked once) nor the fix

## Proposal

**1. Remediation hint on every non-green verdict.** Extend the hub's health verdict with `Remediation{Action, Surface, Link}` — a one-line 'do this' + where to do it, rendered on /fleet under the WHY chip and in the spoke dashboard's warning banners. Signature→action map:

| Signature (heartbeat) | Hint |
|---|---|
| NeedsLogin agents > 0 | 'Copilot device-flow login on the spoke dashboard' → dashboard /login |
| HoldTotal > 0 stale | 'Review the needs-human queue' → repo PR list link |
| GitHub App broken/missing | 'Install/repair the GitHub App' → install URL |
| Budget exhausted (NEW field) | 'Raise or reset the budget limit (Settings → Budget)' |
| Model error-streak (NEW field) | 'Pin a working model on the agent card' |
| Consent wedge (NEW field) | 'Complete the Copilot consent flow on the live pod dashboard' |
| Agent enabled but zero cadences (NEW field) | 'Set cadences on the agent card' |
| Digest ≠ tracked channel | 'Spoke lags its channel — check auto-upgrade / force rollout' |

**2. New detectors (spoke → heartbeat).** Nil-able fields, carry-forward on minimal beats, mirroring PRsMerged90d conventions: `BudgetExhausted bool + BudgetSpend/Limit`, `AgentErrorStreaks map[agent]int` (consecutive model-call failures from the token/error scanners), `ConsentWedged []agent` (restart-loop with consent_screen=true), `NoCadenceAgents []agent` (enabled, zero cadences, zero kicks ever).

[...]

PR #5728, opened 2026-09-02
🐛 fix(fleet): give the agent-down / all-paused / provider-quota verdicts a hint (#5699)

Danathar:
## Summary

Item **1** of #5699: the three verdict families the 2026-09-02 sweep found rendering a WHY chip with nothing under it now carry a hint.

Items **2** and **3** I investigated rather than patched — and they turn out to be the same problem, which changes what should be done about item 2. That is below, and I think it is the more useful half of this PR.

## Item 1 — the eleven hint-less verdicts

The gap is structural, not a wording oversight. `attachRemediation` switches on a **cause signature**, never on the human-facing reason, and deliberately attaches nothing for an unrecognized cause so a hint is only shown when it is specifically true. These three sites set `v.Reason` and never set `v.cause`, so the switch had nothing to match:

| Verdict | Spokes | Hint added |
|---|---|---|
| `N agent(s) down — restart needed` | 3 (one supervisor-down 2d) | Restart the agent from its card *(linked)* |
| `all agents paused — resume to produce output` | 7 | Resume agents, **or mark the hive idle if it is mothballed** |
| `N agent(s) out of provider quota` | 1 | Rotate the agent to a provider with headroom, or raise the provider quota |

The all-paused wording names **both** resolutions on purpose. That verdict is amber precisely because every agent is paused *deliberately*, so "resume them" alone reads as an instruction to undo a choice somebody made.

Also covered: the **spoke-level** provider limit (`e.ProviderLimitReason`), a precondition red on a different code path from the agent rollup and equally hint-less. It is **not** one of the eleven — it may have been firing on nobody at sweep time — but it is the identical remedy, and covering one while leaving the other blank would be arbitrary. Flagging it explicitly since it is beyond the issue's literal scope.

Detection, thresholds and verdict wording are unchanged; this fills in the action half only.

### What I did not do, and why

The issue asks that the agents-down hint add *"consider pinning a model"* **when the failure log matches the copilot-hung-at-startup signature**. That is not implementable today: no per-agent failure text reaches the hub. `AgentSummary` carries state, pause provenance, `NeedsLogin`, `QuotaExhausted`, `SessionMissing` and timestamps — no error text — and no other heartbeat field supplies it. It needs a new heartbeat field, which is spoke-side protocol work rather than a hint-table edit.

Keying off `AgentSummary.Backend == "copilot"` instead would fire on **every** down copilot agent regardless of cause — the false-positive machine the "only when specifically true" rule exists to prevent. So I left it, rather than shipping a weaker proxy under the issue's number.

## Item 2 — the three never-firing detectors are not broken

**No signature needs changing. Do not change them.**

The issue offers two hypotheses: the signatures no longer match, or the spoke-side emitters aren't populating the heartbeat fields on the deployed build. I checked both against the tree, and the second is confirmed by arithmetic:

- All three emitters are wired and present: `src/cmd/hive/main.go` sets `AgentErrorStreaks`, `ConsentWedged` and `NoCadenceAgents` from `tokenCollector.AgentErrorStreaks()`, `agentMgr.ConsentWedgedAgents()` and `gov.NoCadenceAgents()`. The code is not missing.
- All three landed in **one commit**, `b35cb9c3` ("#5577 PR1"), dated **2026-09-01** — the day before the sweep.
- That commit is **144 commits** behind the current `v4` head (`git rev-list --count b35cb9c3..upstream/v4`).
- The sweep found 13 spokes sitting **153 commits behind `:stable`**.

144 < 153, so the deployed spokes were running a build from *before* the emitters existed. **The detectors cannot fire because the fields are never sent** — not because the signatures drifted.

This also explains the issue's own counter-evidence better than its hypothesis does. It notes a known consent-wedge spoke "now classifies as plain login-stuck, suggesting the … signature no longer matches". If the spoke cannot emit `ConsentWedged` at all, that hive necessarily falls through to whatever else classifies it — login-stuck — with no signature drift required.

So **item 2 is item 3.** The three dead detectors are a symptom of the stalled upgrade wave, and the remedy is to land the upgrade wave, not to loosen detector signatures against a live fleet that cannot exercise them. Retuning them now would be tuning against absent data, and would likely make them fire wrongly once the fleet does catch up.

**What still needs a human:** re-run the sweep after the fleet is current. If the detectors still fire zero times on spokes at or past `b35cb9c3`, *then* the signatures are the suspect and there is real work. I cannot do that half — it needs live fleet access.

## Item 3 — the upgrade wave

Left alone. I can corroborate the observation (the emitters being 144 commits back while spokes sit 153 behind is consistent with a wave that is not completing), but diagnosing the upgrade trigger cadence needs hub runtime data I do not have. It is now load-bearing for item 2 rather than a "bonus observation", which is worth knowing when prioritising it.

- `src/pkg/hub/remediation.go` — `causeAgentsDown` / `causeAllPaused` / `causeProviderQuota` and their `attachRemediation` rows.
- `src/pkg/hub/health_verdict.go` — four `v.cause` assignments at the verdict sites.
- `src/pkg/hub/remediation_test.go` — four rows in the existing signature table.
- `CHANGELOG.md`.

### Validation

- Targets **`v4`**: the remediation system (`src/pkg/hub/remediation.go` and the #5577 family) exists on v4 and not on v5, and v4 is the repo's default branch. I checked before branching.
- `go vet ./pkg/hub/` and `gofmt -l` clean on every touched file.
- `go test ./pkg/hub/` passes in full (~100s), including the pre-existing precedence and never-hint-on-green rows.
- **Mutation-checked**: removing the three new `v.cause` assignments fails exactly the three corresponding new table rows, so none passes vacuously.

## Related issues

Refs #5699 (item 1 of 3; items 2 and 3 investigated, see above). Refs #5577 (the system this extends).

## Testing

- [x] `cd src && go test ./...` — ran `./pkg/hub/` in full (the only package touched); passes.


clubanderson (approved):

PR #5732, opened 2026-09-02
🐛 fix: name the base branch in every contributor task prompt

### LINKED ISSUE #5729: 🐛 bug: the contributor task prompt never names a base branch, so one branch-specific issue redirects every later PR in a session
## Summary

A contributor relay works one issue at a time out of a single persistent checkout. The
task prompt tells the agent to fork, clone, commit, push and open a PR — but never says
which branch to base that PR on, and nothing resets the checkout between tasks. So the
base branch is whatever the *previous* task happened to leave checked out.

With two active branches (`v4` default and deployed, `v5` in development) that is enough
for one branch-specific issue to silently redirect every subsequent PR of a session.
That happened today: five consecutive PRs from one contributor landed on `v5`, only the
first of which was correct.

Nobody involved can see it going wrong. The agent has no instruction to check, the
contributor sees PRs being opened and merged normally, and a maintainer sees a stream of
correctly-formed PRs on a plausible branch.

Blast radius: any contributor whose queue includes a branch-specific issue. Direction is
arbitrary — the same mechanism produces `v4` PRs for `v5` work.

## What happened

Contributor `Danathar`, 2026-09-02. Issue #5617 is titled `[v5] reviewer lane follow-ups
…`, so the agent checked out `v5` and opened PR #5682 against it — correct. The four PRs
after it inherited `v5` from the checkout:

```
#5682  14:40:29Z  base=v5   for #5617  ([v5] issue — correct)
#5688  15:28:23Z  base=v5   for #5681  (v4-live defect)
#5700  16:09:55Z  base=v5   for #5685  (v4-live defect)
#5705  16:58:43Z  base=v5   for #5691
#5711  18:01:58Z  base=v5   for #5704  (v4-live defect)
```

Branch ancestry confirms inheritance rather than choice — every one is 1–2 commits ahead
of `v5` and 64–67 ahead of `v4`. No `base_ref_changed` event exists on any of them, so
none was retargeted after the fact; they were opened against `v5` directly.

The cost is not hypothetical. #5688 fixes #5681, a defect live on deployed `v4`; the
hive's own scanner independently opened #5693 for the same issue against `v4`, and a
maintainer had to adjudicate between two implementations of one fix on two branches,
keep the `v5` one, and note that `v4` still needed a port. Three fixes for `v4`-live
defects ended up on a branch nobody is running, and were backported by hand hours later.

## Two contributing causes

**1. The prompt never names a base.** `buildTaskPromptBody`
(`src/pkg/dashboard/contribute_ws.go`) mentions a branch exactly once:

```go
"Push your branch to your fork remote, then open a PR from your fork. "+
```

The hub knows which branch the work belongs on — it selected the issue and knows the
repository's default branch — and says nothing.

**2. The workspace is never reset.** `bin/contributor-relay.sh` has no `git checkout` of
a base branch, no `reset --hard` to one, and no base-branch environment variable, so a
checkout left on `v5` by task N is what task N+1 starts from.

## Fix shape

**The prompt is the load-bearing half.** Add the base branch to `buildTaskPromptBody` —
the repository's default branch normally, or the branch a branch-specific issue names —
and instruct the agent to base its work on it (`git checkout <base> && git pull`, and
`gh pr create --base <base>`). The hub already has the information.

I can report that fixing only the checkout is **not sufficient**: mid-task today I reset
the agent's working branch from `v5` onto `v4` while it held zero commits and a clean
tree. The agent restored `v5` and opened its PR there anyway, because its plan — formed
from the issue and its earlier context — already said `v5`. Whatever the workspace says,
the agent follows what it was told, which is why the instruction has to carry the answer.

Resetting the workspace to the base before each task is still worth doing as defence in
depth, and is a small change on the relay side. It just cannot be the only fix.

Regression coverage: assert `buildTaskPromptBody` names a base branch, and that a task
following a branch-specific one is told the default branch rather than inheriting.

Related: #5617 (the `[v5]` issue that started the run), #5688/#5693 (the two-branch
adjudication this produced).

Danathar:
## Summary

The contributor task prompt never said which branch a task's work belongs on. A relay works one issue at a time out of a single **persistent** checkout and nothing resets it between tasks, so the base was whatever the *previous* task left checked out — which is how one `[v5]`-titled issue redirected four later PRs of a session onto `v5`, three of them fixes for defects live on the deployed `v4`.

This PR makes the prompt name the base branch. Every assignment now tells the agent which branch to start its work branch from and which branch to open the PR against, so the answer comes from the assignment rather than from whatever state the checkout is in.

**What changed, concretely** — the prompt gains one sentence:

> Base this work on the `'v4'` branch of kubestellar/hive. The checkout may be left on a DIFFERENT branch by a previous task, so do not use whatever branch you find there: run `git fetch upstream` and start your work branch from the base with `git checkout -b <your-branch> upstream/v4`. Open the PR against the same branch with `gh pr create --base v4`, and confirm the PR's base is `'v4'` before you report done.

**Where the base comes from** (`taskBaseBranch`): the branch this hive was built from — the same `upstreamBranch()` the onboarding page's `git clone -b` command already names (#3990), so the branch a contributor cloned and the branch they are told to target agree. A branch-specific issue overrides it: a title leading with a release-line tag such as `[v5]` wins, because that issue really is `v5` work whatever branch the hive runs.

The tag shape is deliberately narrow — `v` followed by digits and nothing else, the same `^v(\d+)$` shape `pkg/hub/image_pulls.go` matches and `.github/release-lines.yml`'s `release_lines` lists. That keeps the lane prefixes the classifier already routes on (`[quality]`, `[architect]`, …) from being read as branches, and a tag naming no real branch fails loudly at `gh pr create` rather than silently redirecting the PR.

When no base resolves at all (a build with no injected branch), the prompt still refuses inheritance and names the substitute — the upstream repository's own default branch — instead of falling silent.

## Why the prompt, and not the workspace

The issue reports a direct measurement, and it is the reason this PR is shaped the way it is: a working branch reset from `v5` onto `v4` mid-task, holding **zero commits and a clean tree**, was restored to `v5` by the agent, because the plan it had already formed said `v5`. An agent follows what it was told over what it finds. Whatever the workspace says, the instruction has to carry the answer.

Live confirmation, incidentally: this task's own checkout was sitting on `feat/5617-reviewer-lane-template` (a `v5` branch) when it started — the exact leftover state the issue describes. It was reset to `upstream/v4` by hand before any work began.

## Deliberately not included

**Resetting the workspace to the base before each task.** The issue calls this "a small change on the relay side", but `bin/contributor-relay.sh` performs no git operations at all today — it sets a cwd (`TASK_WORKSPACE_DIR`, line 129) and types a prompt; there is no `git clone`, `git checkout`, or `git switch` anywhere in it or in `bin/contributor-agent.sh`. Adding the reset therefore means a new `task_assign` protocol field *plus* a `git reset --hard` against a contributor's persistent checkout — a destructive new surface that deserves its own change, its own observation, and its own decision about what happens to unpushed work. Happy to file it as a follow-up.

**One adjacent thing worth a maintainer's eye, not changed here:** `.github/PULL_REQUEST_TEMPLATE.md` line 17 still reads "PR targets `v2` unless a maintainer requested another branch." `v2` is sunset and the default branch is `v4`, so the template currently points human contributors at the same class of wrong base this PR fixes for agents. Left alone because it is a human-facing policy line and outside this issue's scope.

## Testing

- [x] `cd src && go test ./pkg/dashboard/` — ok (50s), including the new cases
- [x] `go vet ./pkg/dashboard/`, `gofmt` clean, `python3 src/scripts/check-docs-links.py` passes
- [x] Mutation-checked: with the fix disabled (base passed as `""`), 4 of the new tests fail; with it on, all pass
- [ ] `cd src && go test ./...` — `pkg/agent`, `pkg/config`, `pkg/knowledge`, `pkg/proxy`, `pkg/sandbox` fail in this sandbox (no tmux, read-only `$HOME`). Verified pre-existing: the identical five packages fail the same way on an unmodified `upstream/v4` worktree. `pkg/dashboard`, the only package this PR touches, is green.

New regression coverage in `src/pkg/dashboard/contribute_task_base_branch_test.go` asserts the prompt names a base; that a task following a branch-specific one is told this hive's branch rather than the previous task's; that the base is derived rather than re-hardcoded to `v4`; that an uninjected build branch falls back instead of emitting `unknown`; that an unresolvable base still forbids inheritance; and that lane prefixes are not mistaken for release lines.

## Related issues

Fixes #5729

Context from the incident: #5617 (the `[v5]` issue that started the run), #5688 / #5693 (the two-branch adjudication it produced), #3990 (the onboarding page's clone branch — same class, same source of truth).

## Contributor checklist

- [x] PR targets `v4`, the repository's default branch (the template's line still says `v2`; see above).
- [x] Docs updated — `src/docs/contributor-relay.md` gains a short section on where the base branch comes from.
- [x] `changelog.d/fixed-5729-task-prompt-base-branch.md` carries the changelog entry.


clubanderson (approved):

Issue #5454, opened 2026-09-01
🐛 bug: hive pins every claude agent to a boot-time access token, so the fleet dies daily and only a manual re-login revives it


Danathar:
## Summary

Every Claude agent on a hive stops working roughly once a day and stays stopped until an operator logs in again by hand. Restarting the agent does not fix it.

The cause is that hive reads the **short-lived** Claude access token out of `.credentials.json` **once, at container start**, and injects that snapshot into every claude agent as `CLAUDE_CODE_OAUTH_TOKEN`. That environment variable is a **static bearer override**: with it set, Claude Code uses the value verbatim, never opens the credentials file, and therefore can never refresh. Claude access tokens live **eight hours** (measured mint-to-`expiresAt` on a live credential), so every agent is pinned to whatever token happened to be on disk when the container started. When it ages out, the whole fleet 401s.

The only thing that re-reads the credential is `Manager.ReloadClaudeToken()`, called solely from the dashboard's OAuth-login handler — which is why a manual re-login is the only thing that has ever worked, and why restarting an agent does not: the restart re-injects the same dead snapshot.

**Who notices:** anyone running a Claude-backed hive for longer than half a day. Agents go orange one at a time, in kick order, so it reads like a partial or random failure rather than one fleet-wide event.

**How to reproduce:** leave a Claude hive running past its access token's expiry and kick any agent.

This is the follow-up half of #5165. That issue fixed the rotation probe; nothing above was part of it.

## The measurement that identifies it

Inside the container, as an agent uid, with a valid credentials file present:

```
$ CLAUDE_CODE_OAUTH_TOKEN="sk-ant-oat01-bogus" claude -p "reply ok" --model claude-sonnet-5
Failed to authenticate. API Error: 401 OAuth access token is invalid.

$ claude -p "reply ok" --model claude-sonnet-5
ok
```

The variable wins outright — there is no fallback to the file. And every agent CLI has it set (`/proc/<pid>/environ`, read as the owning uid; reading it as container root silently fails with EACCES, which is easy to misread as "unset").

## Observed, live

Six-agent hive on `atomic-image-builder`, 2026-09-01, all agents sharing `/data/home/.claude/.credentials.json`. All times EDT.

| Time | Event |
|---|---|
| Aug 31 20:06 | operator logs in; `expiresAt` = Sep 1 04:06, refresh grant good until **Sep 29** |
[...]

Issue #4488, opened 2026-08-21
🐛 permissions watcher logs 6 warnings every 10s forever about `/data/agents/reviewer`, an agent that is not in the UID map


Danathar:
A standalone Hive logs **six `permissions watcher` WARN lines every 10 seconds, indefinitely** — roughly 2,160 warnings per hour — about a directory belonging to an agent that does not exist.

This is **not** runtime-specific. Measured on both Docker and Podman on the same host, same image, same config, fresh volume in each case. Recording that explicitly because I found it during the #4188 Podman pass and it would be easy to mistake for a Podman gap: it is not one, and cross-runtime parity holds.

## What is logged

```json
{"level":"WARN","msg":"permissions watcher: chown failed","path":"/data/agents","error":"chown /data/agents: operation not permitted"}
{"level":"WARN","msg":"permissions watcher: chmod dir failed","path":"/data/agents","error":"chmod /data/agents: operation not permitted"}
{"level":"WARN","msg":"permissions watcher: chown failed","path":"/data/agents/reviewer","error":"chown /data/agents/reviewer: operation not permitted"}
{"level":"WARN","msg":"permissions watcher: chmod dir failed","path":"/data/agents/reviewer","error":"chmod /data/agents/reviewer: operation not permitted"}
{"level":"WARN","msg":"permissions watcher: chown failed","path":"/data/agents/reviewer/stats.json","error":"chown /data/agents/reviewer/stats.json: operation not permitted"}
{"level":"WARN","msg":"permissions watcher: chmod file failed","path":"/data/agents/reviewer/stats.json","error":"chmod /data/agents/reviewer/stats.json: operation not permitted"}
```

Repeating on the `PermissionFixInterval` tick. A Docker container up for ~2 minutes had already logged 234 of them.

## Why it never stops

`fixEntry` in `src/pkg/agent/permissions_watcher.go` only acts when ownership is wrong:

```go
if stat.Uid == 0 { newUID = DevUID; needsChown = true }
if stat.Gid != uint32(NodeGID) { needsChown = true }
if needsChown { if err := os.Chown(path, newUID, NodeGID); err != nil { logger.Warn(...) } }
```

So on a correct deployment it is silent. Here the entries really are `uid 0`, and the watcher runs as `dev` after the entrypoint's privilege drop, so `os.Chown` returns `EPERM` on every tick and the ownership is never repaired. The condition that triggers the warning is also the condition the warning can never clear.

[...]

PR #5735, opened 2026-09-02
🐛 fix: keep the shared Claude credential readable by the whole fleet

### LINKED ISSUE #5730: 🐛 bug: Claude fleet drops to a login prompt hours after boot — a token refresh leaves the shared .credentials.json 0600 and all three repair paths miss it
## Summary

Hive lets several AI coding agents share one Claude login: every agent runs as its own Unix user, but their `~/.claude` all symlink to one shared `/data/home/.claude`, so a single operator login authenticates the whole fleet. Claude Code refreshes that shared OAuth credential roughly every 8 hours, and when it does it **rewrites the file as whichever agent happened to refresh it, mode `0600`** — owner-only. Three separate mechanisms exist to re-open that file to the shared `node` group. On a hive that has been up for a few hours, **none of them are working**, and the whole fleet drops to a login prompt.

Operator-visible symptom: every Claude agent except one goes `auth-required` / "pane shows a login/credential screen", in a burst, several hours after the container starts. The one survivor is whichever agent did the refresh. Hive then tells the operator to re-authenticate — but the credential on disk is perfectly healthy, with a live access token and a valid refresh grant. The login is not the problem and re-logging in only resets the clock until the next refresh.

Measured on a standalone rootless-Podman hive, 2026-09-02: fleet up and healthy 07:11–15:06 EDT, then five of six agents dead within 30 minutes of the refresh. This is the daily re-authentication treadmill of #5454. **#5455 fixed the token pinning it was aimed at, but on this deployment it is a no-op** — the fleet is now failing on file permissions instead, and the fix is one `chmod`, not a login.

Cause is **partly unknown**, and I want to be honest about which half: it is established that the file ends up `0600` and that all three repair paths miss it. It is **not** established why two of those three repair paths are dead — see "What I could not determine".

## What happened

Timeline, one container lifetime:

```
07:11:04  podman auto-update recreates the container; entrypoint logs
          "[entrypoint] inotify perm guard active (copilot + claude + codex + gemini)"
          "[entrypoint] polling perm guard active (config.json 5s, cache 5m)"
07:11 – 15:06   fleet healthy, all six claude agents working
15:06:16  /data/home/.claude/.credentials.json rewritten — new access token,
          good for 8h, refresh grant valid. Now  -rw------- hive-supervisor:node
15:11:05  credential watchdog: "durable credential unusable ...
          reason: login expired (no usable refresh grant) ...
          recovery: operator dashboard device-flow login"      <-- wrong on both counts
15:21:10  ci-maintainer  ready -> auth-required  (PaneShowsLogin)
15:36:10  quality, scanner, sec-check, guide  ready -> auth-required
          supervisor — the agent that owns the file — stays healthy
```

The file at 15:06 and onward:

```
-rw-------. 1 hive-supervisor node 508 Sep  2 19:06 /data/home/.claude/.credentials.json
```

Every agent uid shares primary group `node` (gid 1000) and reaches the file through the documented bridge symlink, so `0600` locks out all nine of the ten agent uids that did not write it:

```
$ podman exec hive sh -c 'id hive-supervisor; id hive-guide'
uid=2010(hive-supervisor) gid=1000(node) groups=1000(node),990(hive-supervisor)
uid=2004(hive-guide)      gid=1000(node) groups=1000(node),996(hive-guide)

$ podman exec hive ls -l /data/home/agents/guide/.claude
lrwxrwxrwx. 1 dev node 18 Aug 25 01:26 /data/home/agents/guide/.claude -> /data/home/.claude
```

The credential itself was never the problem — it holds a live access token **and** a valid refresh grant the whole time:

```
$ podman exec hive python3 -c '...'   # metadata only, no token values
expiresAt = 2026-09-02 23:06:16 EDT   (8h after the 15:06 refresh)
subscriptionType = max
accessToken present: True   refreshToken present: True
```

## Root cause

Claude Code writes `.credentials.json` via temp-file-and-rename on every token refresh, so the replacement file lands `0600` owned by the refreshing agent's uid. Three things are supposed to reopen it. All three miss:

**1. The entrypoint's `.claude` inotify guard is dead.** `src/deploy/entrypoint.sh:926` defines exactly the right repair, and the entrypoint logged it active at boot. Nine hours later its `inotifywait` is gone, while its three siblings are still running as root:

```
$ podman exec hive sh -c '... /proc scan ...'
uid=0 pid=111 ppid=1   /bin/sh /usr/local/bin/entrypoint.sh ...
uid=0 pid=113 ppid=111 inotifywait -qq -e close_write,moved_to /data/home/.copilot/
uid=0 pid=114 ppid=1   /bin/sh /usr/local/bin/entrypoint.sh ...
uid=0 pid=117 ppid=114 inotifywait -qq -e close_write,moved_to,create /data/home/.codex/
uid=0 pid=116 ppid=1   /bin/sh /usr/local/bin/entrypoint.sh ...
uid=0 pid=118 ppid=116 inotifywait -qq -e close_write,moved_to,create /data/home/.gemini/
                       # .claude — no subshell, no inotifywait
```

`while inotifywait ...; do` terminates permanently the moment `inotifywait` returns non-zero, and both `-qq` and `2>/dev/null` mean that exit is completely silent.

**2. The entrypoint's polling guard is dead too.** `entrypoint.sh:953-967` runs `chmod -R g+rwX /data/home/.claude` every 5 minutes as a belt-and-suspenders backstop, which would have repaired this within 5 minutes. Its `sleep 5` loop is not running either:

```
$ podman exec hive sh -c 'for d in /proc/[0-9]*; do ... case "$c" in "sleep 5 ") echo FOUND;; esac; done'
(end)          # nothing
```

The two dead loops are precisely the two that walk `/data/home/.claude`; the three survivors are the ones that do not. That tree is 161 MB / 8413 entries / 2553 dirs on this hive, and `chmod -R` plus `find -type d -exec chmod g+s {} +` re-walks all of it on **every** `close_write` — and Claude Code writes into `.claude` constantly (`history.jsonl`, `projects/*.jsonl`, `file-history/`). I could not confirm a mechanism (see below), but the correlation is exact.

**3. The Go permissions watcher is alive, and structurally cannot fix it.** `/data/home/.claude` is in `WatchedHomeDirs` (`permissions_watcher.go:87`), but `fixEntry` returns before any mode repair for entries not owned by `DevUID`:

```go
// src/pkg/agent/permissions_watcher.go:481
// Only fix permissions on files we own or just chowned.
if newUID != DevUID && stat.Uid != uint32(DevUID) {
    return
}
```

`.credentials.json` is owned by `hive-supervisor` (2010), not `dev` (1001), so it is skipped every tick. `.bob` state dirs already have an explicit carve-out immediately above (line 477) for exactly this shape — an agent-owned path the group must reach — and the credential file needs the same.

### Two things make it hard to diagnose from the outside

**The reachability check cannot see the failure.** `claudeCredentialReachable` (`manager.go:6495`) runs in the hive process, which is `dev` — and its own doc comment names this exact risk:

> It is a REACHABILITY check, not a permission check ... this runs in the hive process, so it proves the file is there and parseable, not that the agent's UID can open it. The deployment keeps those the same — the entrypoint's inotify guard chowns `/data/home/.claude` to `dev:node` and holds it group-readable on every write, precisely so every agent UID can read it (#4619). If that ever drifts, an agent lands at a login prompt with no injected token instead of a working one; that is a loud, alerting state, not a silent one.

The invariant it relies on is the guard in (1), which is dead. The predicted drift happened. But the failure was **not** loud — which is the second problem.

**The watchdog blames the wrong thing.** Because `dev` also cannot read a `0600` file owned by another uid, `claude.HasUsableToken` returns false, and the credential watchdog reports `login expired (no usable refresh grant)` with `recovery: operator dashboard device-flow login`. Both are false: the grant is live and no login is needed. That message sends the operator to redo an OAuth login that fixes nothing beyond resetting the 8-hour clock — which is exactly why #5454 read as "hive needs a daily re-login" for so long. A `0600`/EACCES on the credential should be reported as its own distinct condition, naming the mode and owner.

## Expected behavior

One operator login keeps the fleet authenticated indefinitely. When Claude Code rewrites the shared credential `0600`, hive reopens it to the `node` group promptly and no agent notices. If that ever fails, the operator is told the file is unreadable — not that the login expired.

## Reproduction steps

1. Run a standalone hive with two or more agents on `backend: claude` sharing `/data/home/.claude`.
2. Note `ls -l /data/home/.claude/.credentials.json` — group-readable, fleet healthy.
3. Wait for Claude Code to refresh the token (~8h after the last refresh, done by whichever agent kicks first).
4. Re-check the mode: now `-rw------- hive-<that agent>:node`.
5. Within ~15 minutes every *other* claude agent transitions to `auth-required` / `PaneShowsLogin`, and the credential watchdog reports "login expired (no usable refresh grant)".

Faster, without waiting for a refresh:

```
podman exec hive chmod 600 /data/home/.claude/.credentials.json
podman exec hive chown hive-supervisor /data/home/.claude/.credentials.json
```

## Workaround

One command, no restart, no re-login:

```
podman exec hive chmod 640 /data/home/.claude/.credentials.json
```

Verified end-to-end. Within seconds hive logged `auto-restarting agent after token detected in shared config` for all five dead agents, then `credential watchdog: durable credential restored`, and all six came back `needsLogin=false`. It holds only until the next refresh re-tightens the file.

Note for anyone hitting #5454: **do not restart the container.** It is unnecessary and only re-runs the boot-time repair that made the file readable in the first place.

## Suggested fix

Layered, in rough priority:

1. **Give `fixEntry` a credential carve-out.** Recognise `.claude/.credentials.json` (and the equivalent per-tool credential files) the way `bobStateDirBase` is recognised at line 477, and OR in the group-read bit regardless of owning uid — before the `DevUID` early return. This is the layer that is actually alive and running, so it is the one most likely to hold.
2. **Make the entrypoint guard loops survivable and loud.** `while inotifywait` exiting non-zero should log and restart with backoff, not exit silently; and the guards should not re-walk a 161 MB tree on every write — watch the file, or scope the `chmod` to the credential rather than `-R` over the whole tree.
3. **Report the real condition.** When the credential file exists but cannot be read, the watchdog should say so — mode, owner, and the `chmod` — instead of `login expired (no usable refresh grant) -> operator device-flow login`.
4. **Consider making the reachability check a real permission check.** `claudeCredentialReachable` could `access(2)` as the agent's uid rather than inferring from the hive process's own view; its doc comment already identifies this as the weak point.

## What I could not determine

- **Why the two loops died.** No OOM kill in the kernel log, nothing in the hive journal. I only have the correlation that both dead loops are the two that recursively walk `.claude` and the three survivors do not. The `-qq` / `2>/dev/null` silencing means there is no evidence left to read; whoever picks this up will likely need to instrument the loops before the cause is knowable.
- **The file's mode before 15:06.** I did not capture it. The fleet was working, so it must have been group-readable; I infer the boot-time `chown -R dev:node /data/home/.claude` (entrypoint.sh:834) preserved a `0660` mode left on disk by an earlier guard run. Not observed directly.
- **Whether Claude Code always writes `0600` on refresh**, or only when creating the file fresh. Worth confirming, since it determines how often the repair has to fire.

## Environment

- Branch/version: `hive 0.0.0-dev (commit a8619b0, branch v4)`, image `ghcr.io/kubestellar/hive:stable`, GH Actions run 8084
- Code references above are against `v4` tip `9be313f7`; the only commit touching these files since the running image is `a8fdbb50` (a `TerminalSession` refactor), so the analysis applies to both
- Deployment: standalone **rootless** Podman + Quadlet, `podman-auto-update.timer` enabled
- OS: Fedora / Universal Blue (Bluefin), kernel 7.1.6
- Claude Code: 2.1.226
- Six agents on `backend: claude`, per-agent uids 2001-2010, shared group `node` (1000), per-agent homes under `/data/home/agents/<name>` with `.claude` bridged to `/data/home/.claude` (#4619)

## Related

- #5454 / #5455 — the token pinning this was long mistaken for. #5455 is correct and is in the running image; it is simply inert here, because its guard depends on the credential being group-readable and it is not.
- #4596 — the same class of bug for `.claude.json`, fixed by per-agent homes (#4619). That fix bridges `.claude` as shared state on the stated premise that "one login authenticates the fleet"; this issue is that premise breaking for the credential file specifically.

Danathar:
## Summary

A Claude fleet shares one operator login: each agent runs as its own uid, but their `~/.claude` all symlink to one `/data/home/.claude` (#4619). Claude Code refreshes that OAuth credential every ~8 hours and rewrites it as whichever agent refreshed it, **mode `0600`** — so the file the whole fleet reads becomes readable by exactly one of them, and everyone else drops to a login prompt. Three repair paths exist and all three missed.

This PR fixes all three, in the order they can actually act, and — importantly — answers the question the issue left open.

## The unknown: why two of the three guards were dead

The issue reports the correlation ("the two dead loops are the two that walk `.claude`, the three survivors do not") and says the cause is not determinable because `-qq` / `2>/dev/null` erased the evidence. It is determinable from the code, and it is now measured:

**`entrypoint.sh` line 2 is `set -e`, and a `( ... ) &` subshell inherits it.** `while inotifywait ...; do` is exempt from `set -e` in its **condition** only — the body is not. So one non-zero command in a guard body did not skip a repair; it **ended that guard permanently**, silently, for the life of the container.

The failing command is `chmod -R`. Over a tree being written to concurrently it returns non-zero the moment an entry vanishes mid-walk. Both dead guards walked `/data/home/.claude` — 161 MB, 8413 entries, written to constantly by Claude Code (`history.jsonl`, `projects/*.jsonl`, `file-history/`) — and the three survivors never touched it. `.codex` and `.gemini` run the identical `chmod -R` idiom and lived precisely because nothing churns there.

Both halves are reproduced as assertions in the new suite:

```
PASS: chmod -R over a tree being written to returns non-zero (rc=1) — the killer under set -e
PASS: the pre-fix idiom (bare command in a set -e guard body) dies on its first failure
```

## One correction to the issue's plan

The issue ranks the `fixEntry` carve-out first, "the layer that is actually alive and running, so it is the one most likely to hold". That layer is alive, but it **cannot be the repair on the deployment described**: `chmod(2)` requires owning the inode or `CAP_FOWNER`, the hive process drops to `dev` (uid 1001) at the entrypoint's privilege drop (`entrypoint.sh:1540/1567`), and the credential is owned by `hive-supervisor` (2010). The existing `.bob` carve-out already documents this limit — "a chmod we are not permitted to perform (EPERM — we are neither the owner nor privileged) is logged and swallowed".

So the **entrypoint guards are the load-bearing repair** (they run as root), and the watcher carve-out is worth having for the case it *can* fix plus, above all, for making the failure loud. Both are implemented; the code and commit message say which is which rather than implying the Go layer is the fix.

## What changed

**1. The entrypoint guards — survivable, loud, and cheap** (`src/deploy/entrypoint.sh`)

Guard bodies are now functions that cannot fail (`|| true` on every arm). A watcher that exits is **logged and restarted with backoff** instead of ending its guard. The instant `.claude` path reopens the credential instead of re-walking the tree on every write; the recursive sweep stays on the 5-minute cycle where a churn failure costs one cycle. The 5s poll now reopens the credentials directly — **that alone would have repaired this incident within five seconds.**

**2. The permissions watcher — a shared-credential carve-out** (`permissions_watcher.go`)

Same shape as the existing `.bob` carve-out, ahead of the owner guard that was skipping exactly the file that breaks the fleet. **Group read only** — the CLIs replace the file by temp-file-and-rename in a group-writable directory, so no agent needs write on an OAuth token. When it cannot repair (the EPERM case above) it logs at ERROR with the mode, the owning uid and the exact `chmod`, deduped — instead of the silence that made this undiagnosable.

**3. The watchdog stops misdiagnosing it** (`manager.go`)

`os.Stat` succeeds on a `0600` file owned by another uid, so every check built on stat-plus-parse read a healthy credential as `login expired (no usable refresh grant)` and prescribed a device-flow login that fixes nothing — the next refresh re-tightens the file. That misreading is why #5454 read as "hive needs a daily re-login". Probes now carry their own recovery, and the Claude probe checks readability **before** usability:

```
reason=unreadable by the hive process (permission denied)
recovery=chmod g+r /data/home/.claude/.credentials.json — the credential itself is fine; a re-login will not help
mode=-rw------- owner_uid=2010 reader_uid=1001
```

A genuinely spent login still reports the expired grant and still asks for an operator login.

`claudeCredentialReachable`'s doc comment predicted this drift and called it "a loud, alerting state, not a silent one". It was not; the comment now says so and names what makes it loud instead.

**Deliberately not included:** making that check a real per-uid `access(2)` (issue suggestion 4, itself phrased as "consider"). From the hive process that means `seteuid` or the SUID `su-exec` helper — a privileged surface that deserves its own change and its own observation, not a rider on this one. The two layers above already turn the condition from silent into reported.

## Operator-facing

`src/docs/agent-watchdog.md` gains a section for the symptom as it actually presents: *most of the fleet goes `auth-required` at once* — check the mode before touching the login, `chmod g+r` fixes it, and **do not restart the container** (it only re-runs the boot repair that hides the cause).

## Testing

- [x] `go vet ./pkg/agent/ ./pkg/claude/`, `gofmt` clean on every touched file
- [x] New Go tests pass; **mutation-checked** — disabling the carve-out and the EACCES arm fails 3 of them
- [x] `src/deploy/test_entrypoint_perm_guard.sh` — 19/19; **mutation-checked** twice: reverting `entrypoint.sh` fails it, and reintroducing the bug into the new structure (dropping one `|| true`, restoring the tree walk) fails 4 assertions including the behavioural `set -e` one
- [x] `test_entrypoint_data_ownership.sh` 21/21, `test_entrypoint_symlinks.sh` 8/8, `test_entrypoint_runtime_config.sh` 11/11, `test_entrypoint_dangling_keyfile.sh` 5/5, `test_entrypoint_system_gitconfig.sh` 10/10, `test_deploy_suites_wired.sh` 0 orphaned
- [x] `sh -n` and `shellcheck` on `entrypoint.sh` — no new findings
- [x] `python3 src/scripts/check-docs-links.py` passes
- [ ] `cd src && go test ./...` — `pkg/agent` and four other packages fail in this sandbox (no tmux, read-only `$HOME`). Verified not mine: the failing test set in `pkg/agent` is **byte-identical** (38 tests) to an unmodified `upstream/v4` worktree run.

My refactor did break `test_entrypoint_data_ownership.sh`'s ongoing-guard recognizer, which greps for two literal idioms my change replaced. I updated it to accept the new dispatch and sweep forms, then confirmed it is not weakened: dropping `.codex` from every guard still fails it.

## Related issues

Fixes #5730

Context: #5454 / #5455 (the token pinning this was mistaken for — correct, and inert here), #4596 / #4619 (per-agent homes, which introduced the shared-credential premise this breaks), #4488 (the watcher's warn-dedupe the new ERROR arm reuses).

## Contributor checklist

- [x] Docs updated — operator section in `src/docs/agent-watchdog.md`.
- [x] `changelog.d/fixed-5730-shared-claude-credential-mode.md` carries the changelog entry.
- [x] No secrets, credentials, or local runtime state are committed — the new probe fields are asserted never to carry token material.


clubanderson (commented):
Review — **CONCERNS (fixable).** I validated the branch in a required worktree (`cd src && go build ./...`; `go test ./pkg/agent` passed locally), but GitHub's Linux runner is red on the new shared-credential tests:

- `TestFixEntry_CredentialRepairGrantsReadNotWrite`
- `TestFixEntry_CredentialRepairIsIdempotent`
- `TestFixEntry_CarveOutIsFileOnly`

Root cause: after `fixSharedCredentialGroupRead` grants only `g+r`, `fixEntry` continues into the generic DevUID-owned repair path on Linux runners where the test UID is 1001. That widens the credential to `g+w`, which the tests correctly reject. The directory-only test also assumes the runner UID is not DevUID.

Exact patch to apply:

```diff
diff --git a/src/pkg/agent/permissions_watcher.go b/src/pkg/agent/permissions_watcher.go
@@
 	if !fi.IsDir() && sharedCredentialBases[filepath.Base(path)] {
 		fixSharedCredentialGroupRead(path, fi.Mode(), stat.Uid, logger)
+		return
 	}
 
 	// Only fix permissions on files we own or just chowned.
diff --git a/src/pkg/agent/shared_credential_test.go b/src/pkg/agent/shared_credential_test.go
@@
 func TestFixEntry_CarveOutIsFileOnly(t *testing.T) {
 	resetPermWarnDedupe()
-	// DevUID is deliberately left pointing at the production dev uid here: this
-	// test user does not own DevUID, so the directory arm below fixEntry's owner
-	// guard cannot run, and anything that changes this directory's mode came
-	// from the carve-out. That is exactly what is being ruled out.
+	// Point DevUID away from the test user's UID so the directory arm below
+	// fixEntry's owner guard cannot run. GitHub's Linux runner happens to use
+	// uid 1001, the production DevUID, so relying on the default made this test
+	// exercise the generic directory repair rather than the credential carve-out.
+	origDevUID := DevUID
+	DevUID = os.Getuid() + 1
+	t.Cleanup(func() { DevUID = origDevUID })
 	path := filepath.Join(t.TempDir(), ".credentials.json")
```

Note: #5736 appears to build on/superset this PR for #5734, so once that branch carries the same fix it may be the cleaner merge target.

clubanderson (approved):
Re-reviewed after applying the permission carve-out fix. Validation: cd src && go build ./... and targeted pkg/agent tests for the previously failing credential cases pass locally. No remaining blocking findings.

clubanderson:
Verdict: superseded. The shared-credential fix from this PR has now landed through #5736, which included this change set plus the agy depth fix for #5734, and #5730 is closed by that merge. Closing this duplicate branch to avoid merging the same permission-guard rewrite twice.

clubanderson:
Superseded by merged PR #5736.

PR #5736, opened 2026-09-02
🐛 fix: watch the depth agy's credential actually lives at

### LINKED ISSUE #5730: 🐛 bug: Claude fleet drops to a login prompt hours after boot — a token refresh leaves the shared .credentials.json 0600 and all three repair paths miss it

### LINKED ISSUE #5734: 🐛 bug: the .gemini perm guard can never fire — agy's oauth token is one directory deeper than the watch, leaving the 5-minute poll as its only protection
## Summary

Hive lets several agents share one login per CLI: each agent runs as its own Unix user, but they all resolve credentials through one shared `/data/home/<tool>` directory. These CLIs rewrite their credential file `0600` on every token refresh, owned by whichever agent refreshed it, which would lock out every other agent — so the entrypoint runs a per-directory `inotifywait` guard that re-opens the file to the shared `node` group after each write.

**For antigravity (`agy`), that guard watches the wrong directory and can never fire.** The watch is on `/data/home/.gemini/` with no `-r`, but `agy` keeps its credential one level deeper, at `/data/home/.gemini/antigravity-cli/antigravity-oauth-token`. `inotifywait` without `-r` only reports events for entries directly inside the watched directory, so a write to the token is invisible to it.

Nothing has broken yet on my hive, because `agy` has so far only ever run as `dev` — the uid that already owns the file. The moment an agent uid signs in, the refreshed token becomes `0600 hive-<agent>:node` and every other agent loses read access. At that point exactly **one** mechanism is left to repair it: the entrypoint's 5-minute polling `chmod -R`. That is one of the two loops observed silently dead in #5730.

So this is a latent single-point-of-failure sitting behind a guard that reads as protection but is inert. Filing it separately from #5730 because the defect is different in kind — that issue is a guard that dies, this is a guard that was never watching the right path.

## Evidence

The guard names the exact file it is meant to protect (`src/deploy/entrypoint.sh:936-949`):

```sh
# agy writes antigravity-oauth-token 0600 owned by the agent that signed
# in, which locks every other agent UID out of a credential the shared
# CLI home exists to share ... Re-opening it to the node group is what
# makes ONE agy login serve the fleet instead of one login per agent.
while inotifywait -qq -e close_write,moved_to,create /data/home/.gemini/ 2>/dev/null; do
  chmod -R g+rwX /data/home/.gemini 2>/dev/null
  ...
```

But the file is not in that directory:

```
$ podman exec hive find /data/home/.gemini -name '*token*' -type f -exec ls -l {} \;
-rw-rw----. 1 dev node 505 Sep  2 20:33 /data/home/.gemini/antigravity-cli/antigravity-oauth-token
```

Confirmed empirically. The token was rewritten at 20:33 UTC, two minutes after a container restart at 20:31 — and the guard has still never fired. `inotifywait` without `-m` exits after one event and the `while` loop re-execs it, so a guard's pid advances every time it fires. Sixteen minutes after boot:

```
pid=113  ppid=111  inotifywait ... /data/home/.copilot/
pid=116  ppid=114  inotifywait ... /data/home/.codex/
pid=118  ppid=115  inotifywait ... /data/home/.gemini/     <-- still the boot pid
pid=5424 ppid=112  inotifywait ... /data/home/.claude/
```

`.claude` has advanced to 5424 because `.credentials.json` sits directly in `.claude/`, at the depth the watch actually covers. `.gemini` is frozen at its boot pid across a token rewrite.

## Coverage compared

| CLI | credential path | depth | inotify guard fires? | in `WatchedHomeDirs`? | protectors |
|---|---|---|---|---|---|
| copilot | `.copilot/config.json` | 1 | yes | yes | inotify + 5s poll + 5m poll + Go watcher |
| claude | `.claude/.credentials.json` | 1 | yes | yes | inotify + 5m poll + Go watcher\* |
| **agy** | **`.gemini/antigravity-cli/antigravity-oauth-token`** | **2** | **no** | **no** | **5m poll only** |

\* the Go watcher's `fixEntry` skips non-`DevUID`-owned files, so it does not actually repair an agent-owned credential — see #5730.

`/data/home/.gemini` is absent from `WatchedHomeDirs` (`src/pkg/agent/permissions_watcher.go:86-112`), which lists `.copilot`, `.claude`, `.cache`, `.config`, `.local`, `.bob`, `/data/agents`, `/data/beads`. So there is no Go-side fallback for antigravity at all.

## Why the existing test passes

`src/deploy/test_entrypoint_data_ownership.sh:343-362` checks exactly this property, and its comment names the exact failure mode:

> the credential file itself is rewritten 0600 by the CLI on every refresh (copilot's config.json, **agy's oauth token**), so a dir that is not in the ONGOING guard silently re-locks after the first refresh.

But the assertion greps for the **directory** in a guard, not for the **credential's actual path**:

```sh
if grep -qE "inotifywait [^|]*${path}/" "$ENTRYPOINT" || \
   grep -qE "chmod -R g\+rwX [^&|;]*${path}( |$)" "$ENTRYPOINT"; then
```

`/data/home/.gemini` appears in both, so `.gemini` passes — while the file the guard exists for sits outside its reach. The test's own note, "One is enough; neither is the bug," is the assumption that fails here: for `.gemini` the inotify arm is inert, leaving one, and #5730 shows that one can die silently.

## Expected behavior

One `agy` login authenticates the fleet and keeps working across token refreshes, with the same belt-and-suspenders coverage the other CLI homes get.

## Reproduction steps

1. Configure an agent (not `dev`) on `backend: agy` and sign in.
2. `ls -l /data/home/.gemini/antigravity-cli/antigravity-oauth-token` — now `0600 hive-<agent>:node`.
3. Observe the `.gemini` guard's `inotifywait` pid is unchanged: it did not fire.
4. Any other agent on `agy` cannot read the token and lands at a login prompt.
5. Up to 5 minutes later the polling slow-cycle `chmod -R g+rwX` repairs it — unless that loop has died (#5730), in which case it stays broken.

## Suggested fix

1. **Watch the path the credential is actually on.** Either add `-r` to the `.gemini` watch, or watch `/data/home/.gemini/antigravity-cli/` directly. `-r` is the more robust of the two, since it survives `agy` moving the file again.
2. **Add `/data/home/.gemini` to `WatchedHomeDirs`** so the Go watcher is a real fallback — most useful together with the credential carve-out proposed in #5730, since without it `fixEntry` skips agent-owned files anyway.
3. **Assert on credential paths, not directories.** Drive `test_entrypoint_data_ownership.sh` from a table of `(cli, actual credential path)` and check the guard covers that path's parent. As written the test cannot fail for a credential that moves into a subdirectory, which is exactly what happened.
4. While in there, it is worth confirming the other entries are still accurate — `.codex` and `.bob` hold no credential file on this hive, so their guards have never been exercised either and may have the same latent shape.

## Environment

- `hive 0.0.0-dev`, revision `9be313f7` (run 8277), image `ghcr.io/kubestellar/hive:stable`
- Standalone **rootless** Podman + Quadlet, `podman-auto-update.timer` enabled
- OS: Fedora / Universal Blue (Bluefin), kernel 7.1.6
- `agy` present at `/usr/local/bin/agy`; `.gemini` tree currently all `dev:node 0660`

## Related

- #5730 — the `.claude` credential going `0600` after a refresh, and both entrypoint guard loops dying silently. Same family (shared credential re-locked by a CLI rewrite), different defect. The two compound: this issue leaves `agy` with a single protector, and that issue is about that protector dying.

Danathar:
> **Stacked on #5735 — please merge that one first.** Both PRs rewrite the same six lines of the entrypoint's guard block and the same assertion in `test_entrypoint_data_ownership.sh`; landing them independently guarantees a conflict, and a careless resolution would silently drop one of the two fixes. **Review only the second commit, `e21a9275`** — the first is #5735 unchanged. Happy to rebase onto `v4` instead once #5735 lands, if you prefer.

## Summary

The `.gemini` permission guard watched `/data/home/.gemini/` **without `-r`**, while `agy` keeps its OAuth token one level deeper, at `.gemini/antigravity-cli/antigravity-oauth-token`. `inotifywait` without `-r` reports events only for entries **directly inside** the watched directory, so a write to that token was invisible to it.

The guard was not merely untested — it was **structurally incapable of firing**, for the whole life of every container, while reading as protection. The reporter's evidence is the cleanest possible form of it: sixteen minutes and one token rewrite after boot, `.claude`'s `inotifywait` pid had advanced (its credential is at depth 1, so the guard fires and the loop re-execs the watcher) while `.gemini`'s was still the boot pid.

Nothing had broken yet only because `agy` had so far run as `dev`, the uid that already owns the file. The first time an agent uid signs in, the refreshed token becomes `0600 hive-<agent>:node` and every other agent loses read access to the credential the shared CLI home exists to share.

## What changed

**1. The `.gemini` guard watches recursively.** `hive_guard_forever` takes an optional `RECURSE` argument, passed through by a new `hive_watch_once`. Recursion is **per-guard and not the default**: `.claude` is 161 MB / 8413 entries on a working hive, and watching it recursively would cost a watch per subdirectory for a credential that sits at the top level anyway.

**2. `.gemini/antigravity-cli` is pre-created at boot**, group-writable + setgid like its parent. A watch can only be placed on a directory that exists, so a subdirectory `agy` creates after boot might never be covered — that depends on how the installed inotify-tools handles directories appearing under a recursive watch. Pre-creating it removes the dependency rather than relying on behaviour that varies by version. (I could not verify that behaviour here: `inotifywait` isn't installed in this environment, so I designed so it doesn't matter.)

**3. `/data/home/.gemini` joins `WatchedHomeDirs`.** It was absent entirely — `agy` was the one CLI with a shared credential and no Go-side fallback at all. `fixPermissions` walks recursively and `sharedCredentialBases` (added in #5735) already knows the token's basename, so the nested file is reached and repaired.

**4. The `.gemini` instant body drops its tree walk**, matching `.claude`. Under a recursive watch it now fires on writes anywhere below `.gemini`, and re-walking a tree on every event is precisely the pattern that killed the `.claude` guard in #5730. The recursive sweep stays on the 5-minute cycle.

## Why the existing test passed — and what it asserts now

`test_entrypoint_data_ownership.sh` checks this exact property, and its comment even names agy's oauth token as the thing at risk. But it greps for the **directory** in a guard, not for the **credential's path**, and `/data/home/.gemini` appears in both arms — so `.gemini` passed while the file the guard exists for sat outside its reach.

It now asserts on a table of `(cli, real credential path)` and, for each, three things: some repair **names the file**; the watch can **see a write to it** (its own parent watched, or an ancestor watched with `-r`); and the 5s poll repairs it too. Run against the pristine `upstream/v4` entrypoint, the new block fails on exactly the two issues' findings:

```
FAIL: agy: no inotify guard can ever fire for /data/home/.gemini/antigravity-cli/antigravity-oauth-token
FAIL: agy: no repair names /data/home/.gemini/antigravity-cli/antigravity-oauth-token
FAIL: claude: no repair names /data/home/.claude/.credentials.json
...
=== 24 passed, 6 failed ===
```

Two notes on the test, in the interest of it staying honest rather than merely green:

- The 5s-poll check follows **one level of indirection**, because the poll body calls named helpers rather than spelling every path inline. My first version didn't, and reported copilot as unprotected when it is in fact repaired every cycle — a false positive I fixed rather than papered over.
- The loop reads from a here-doc, not a pipe: `printf … | while` would run the body in a subshell and every `ok`/`bad` would increment a counter that dies with it, so a real failure would print and then vanish from the tally.

## Deliberately not done

**Issue item 4** asks to confirm whether `.codex` and `.bob` have the same latent shape. Neither holds a known credential file — on the reporter's hive or in anything I could find in-tree — so they are **deliberately absent from the credential table**: guessing a path would assert nothing, and a wrong guess would assert something false. Confirming them needs an observation of a live `codex`/`bob` login that I don't have. The directory-level assertions still cover both.

## Testing

- [x] `cd src && go build ./...`, `go vet ./pkg/agent/`, `gofmt` clean on every touched file
- [x] `sh -n` and `shellcheck -s sh` on `entrypoint.sh` — 11 findings, byte-identical to before this change (all pre-existing)
- [x] `test_entrypoint_data_ownership.sh` 30/30, `test_entrypoint_perm_guard.sh` 25/25, `test_entrypoint_symlinks.sh` 8/8, `test_deploy_suites_wired.sh` 0 orphaned
- [x] New Go tests pass; **mutation-checked** — removing `/data/home/.gemini` from `WatchedHomeDirs` fails `TestWatchedHomeDirsIncludesGemini`
- [x] **Mutation-checked the shell assertions three ways**: dropping `-r` from the dispatch fails both suites (`no inotify guard can ever fire` / `the .gemini guard is dispatched recursively`); running the new block against pristine `upstream/v4` fails 6 assertions; `hive_watch_once` is executed against a stub that records its argv, so the flag itself is tested rather than grepped
- [x] `python3 src/scripts/check-docs-links.py` passes
- [ ] `cd src && go test ./...` — `pkg/agent` and four other packages fail in this sandbox (no tmux, read-only `$HOME`). Verified not mine: the failing set in `pkg/agent` is **byte-identical** (38 tests) to an unmodified `upstream/v4` worktree run.

## Related issues

Fixes #5734

Depends on #5735 (fixes #5730) — the guard that this issue leaves as agy's only protector is the one that PR keeps from dying. The two compound exactly as the issue describes, which is why they are stacked rather than parallel.

## Contributor checklist

- [x] Docs updated — the operator section in `src/docs/agent-watchdog.md` now names agy's token path alongside claude's.
- [x] `changelog.d/fixed-5734-gemini-guard-watch-depth.md` carries the changelog entry.


clubanderson (commented):
Review — **CONCERNS (fixable).** I validated the branch in a required worktree (`cd src && go build ./...`; `go test ./pkg/agent` passed locally), but GitHub's Linux runner exposed a real permission-mode regression in the new tests:

- `TestPermissionsWatcher_AntigravityNestedCredential`
- `TestFixEntry_CredentialRepairGrantsReadNotWrite`
- `TestFixEntry_CredentialRepairIsIdempotent`
- `TestFixEntry_CarveOutIsFileOnly`

Root cause: after `fixSharedCredentialGroupRead` grants only `g+r`, `fixEntry` continues into the generic DevUID-owned file/dir repair path on Linux runners where the test UID is 1001. That path widens credentials to `g+w`, defeating the OAuth-token read-only invariant. The directory-only test also assumes the runner UID is not DevUID; GitHub's runner is 1001, so it exercises the generic directory repair instead of isolating the carve-out.

This minimal patch fixes the CI failures while preserving the intended behavior:

```diff
diff --git a/src/pkg/agent/permissions_watcher.go b/src/pkg/agent/permissions_watcher.go
@@
 	if !fi.IsDir() && sharedCredentialBases[filepath.Base(path)] {
 		fixSharedCredentialGroupRead(path, fi.Mode(), stat.Uid, logger)
+		return
 	}
 
 	// Only fix permissions on files we own or just chowned.
diff --git a/src/pkg/agent/shared_credential_test.go b/src/pkg/agent/shared_credential_test.go
@@
 func TestFixEntry_CarveOutIsFileOnly(t *testing.T) {
 	resetPermWarnDedupe()
-	// DevUID is deliberately left pointing at the production dev uid here: this
-	// test user does not own DevUID, so the directory arm below fixEntry's owner
-	// guard cannot run, and anything that changes this directory's mode came
-	// from the carve-out. That is exactly what is being ruled out.
+	// Point DevUID away from the test user's UID so the directory arm below
+	// fixEntry's owner guard cannot run. GitHub's Linux runner happens to use
+	// uid 1001, the production DevUID, so relying on the default made this test
+	// exercise the generic directory repair rather than the credential carve-out.
+	origDevUID := DevUID
+	DevUID = os.Getuid() + 1
+	t.Cleanup(func() { DevUID = origDevUID })
 	path := filepath.Join(t.TempDir(), ".credentials.json")
```

I applied that patch locally on top of this PR and verified:

`cd src && go test ./pkg/agent -run 'Test(PermissionsWatcher_AntigravityNestedCredential|FixEntry_CredentialRepairGrantsReadNotWrite|FixEntry_CredentialRepairIsIdempotent|FixEntry_CarveOutIsFileOnly)$'`

Result: pass.

clubanderson (approved):
Re-reviewed after applying the permission carve-out fix. Validation: cd src && go build ./... and targeted pkg/agent tests for the previously failing credential cases pass locally. No remaining blocking findings.

PR #5738, opened 2026-09-02
🐛 fix: clamp GitHub rate-limit readings monotone within a window

### LINKED ISSUE #5733: Dashboard rate-limit card reports full headroom while the App budget is spent (non-monotonic within a window)
## Summary

`/api/status` → `.ghRateLimits.core.remaining` intermittently reports the **full limit** while a substantial fraction of the GitHub App installation's budget is already spent. It is not merely lagging — the value moves *backwards* (reports more headroom than it did a minute earlier) and can sit pinned at the limit for many minutes while real consumption climbs.

The dashboard rate-limit card is the signal an operator uses to decide whether there is room to run agents harder. When it reads full and isn't, you throttle or scale the wrong thing. That is exactly what happened to us: I cut a spoke's cadence because the card showed 100% headroom and I concluded a *different* resource was binding. The card was wrong in the other direction — there was 86% of the budget free.

## Measurement

Paired samples, one per minute, same installation, same window. Column A is `.ghRateLimits.core.remaining` from `/api/status`. Column B is `GET /rate_limit` called directly with an installation token minted from the same App key inside the same pod.

```
TIME      | A: hive says | B: actual | B used | gap
20:53:11  | 7100         | 7078      | 22     |  22
20:54:12  | 7100         | 7012      | 88     |  88
20:55:13  | 6941         | 6881      | 219    |  60
20:56:14  | 6941         | 6815      | 285    | 126
20:57:15  | 7100  <-- UP | 6741      | 359    | 359
20:58:15  | 7100         | 6640      | 460    | 460
20:59:16  | 7100         | 6459      | 641    | 641
21:00:17  | 7100         | 6330      | 770    | 770
21:01:17  | 7100         | 6330      | 770    | 770
21:02:18  | 7100         | 6286      | 814    | 814
21:03:18  | 6044         | 6017      | 1083   |  27
21:04:19  | 6044         | 5748      | 1352   | 269
21:05:19  | 6044         | 5736      | 1364   | 292
```

Two things stand out:

1. **Non-monotonic within one window.** 20:56 reports 6941, 20:57 reports 7100. Nothing resets in between — B falls from 6815 to 6741 across the same interval.
2. **Pinned at exactly the limit for 6 minutes** (20:57–21:02) while 455 real requests are consumed. Peak divergence 814 requests, i.e. the card claims 100% headroom at ~89% actual.

The three distinct A values (7100 / 6941 / 6044) against a smoothly declining B suggest a sporadically refreshed cache, where some refreshes land on a reading that looks empty.

## Probable mechanism

`buildGHRateLimits` (src/pkg/dashboard/status_builder.go) does:

```go
limits, err := ghClient.RateLimits(ctx)
```

Installation access tokens appear to report a **fresh, empty bucket transiently** right after minting, even though the budget is genuinely shared across tokens for the installation. Demonstrated directly, same installation, seconds apart:

```
token A initial:  {"used":90,  "remaining":7010, "reset":1788385930}
... 20 requests issued with token A ...
token A after:    {"used":0,   "remaining":7100, "reset":1788386085}   <-- empty bucket, new reset
token B (fresh):  {"used":160, "remaining":6940, "reset":1788385930}   <-- sees A's usage
```

Token B proves the bucket *is* shared (it observes A's 20 requests plus other traffic). Token A's own read is what goes empty. So whenever the client behind `RateLimits()` has recently re-minted, the dashboard latches that empty reading and shows full headroom.

## Suggested fix

Any of these would remove the false-full state; the first is the smallest:

- **Never let the displayed value increase within a window.** Track `reset`; if `reset` has not advanced, clamp `remaining` to the minimum seen for that window. A monotone-within-window value cannot report false headroom.
- **Prefer the `X-RateLimit-Remaining` response headers** from calls the client is already making, rather than a separate `/rate_limit` probe on a possibly-just-minted token.
- **Surface staleness.** Include the observation timestamp alongside the numbers, so a value that stops updating is visibly stale rather than silently confident. (Today `reset` is the only time field, and it moves independently of when the sample was taken — at 20:57 the card's `reset` was 8.5 minutes adrift of B's.)

Happy to send a patch for the monotone-clamp version if that shape is agreeable.

## Environment

- Self-hosted spoke, GitHub App auth (`identity: {type: "app"}`), installation limit 7100/hr
- Two spokes plus a hub sharing one App installation
- Values read from `/api/status`, compared against `GET /rate_limit` with a token minted from the same App private key inside the same pod

Danathar:
## Summary

The GitHub API card reported the **full limit** while most of the App installation's budget was spent, and moved *backwards* while doing it — 100% claimed headroom at ~89% real usage, peaking at 814 requests of divergence. That card is what an operator reads to decide whether there is room to run agents harder, so a false-full reading throttles or scales the wrong thing; the reporter cut a spoke's cadence on the strength of it.

Rate-limit readings are now **clamped to be monotone within a window**, and the card shows **when the reading was taken**.

## One thing worth reading before the diff

The issue proposes the clamp as *"track `reset`; if `reset` has not advanced, clamp `remaining` to the minimum seen for that window"* — the smallest of three options. **That version does not fix this bug**, and the issue's own evidence is what shows it:

```
token A after:   {"used":0,   "remaining":7100, "reset":1788386085}   <- the artifact
token B (fresh): {"used":160, "remaining":6940, "reset":1788385930}   <- the truth
```

The bogus reading carries a **later** `reset`, because a freshly minted token reports its own fresh window. A clamp keyed on the `reset` value would read every re-mint as a rollover and accept precisely the readings it exists to reject. The reporter measured this directly from the other side — *"at 20:57 the card's `reset` was 8.5 minutes adrift of B's"* — so it isn't a theoretical objection.

I implemented that naive version as a mutation and ran it against this PR's tests: **it fails 5 of them, including the end-to-end one.**

So the window is anchored on **wall-clock expiry** instead: a new window is believed only once the previous window's `reset` has actually passed. Before that, a reading carrying a *different* `reset` is a re-minted token describing its own bucket and is discarded; a reading carrying the *same* `reset` is clamped to the lowest seen.

## What changed

**The clamp** (`src/pkg/github/ratelimit_window.go`, new). Per-bucket window state on `github.Client`, applied inside `RateLimits()` — not in the dashboard — so every consumer inherits it. `/api/status` and `/api/gh-rate-limits` are both display paths today, but a future caller that gates work on headroom must not be handed the raw artifact either.

Every branch fails toward reporting **less** headroom than may really be available. That's deliberate: under-reporting costs some throughput, over-reporting is what produced this issue. A clock running behind GitHub's holds a window past its true rollover — at most one window of stale-low display, self-correcting.

**Staleness** (issue suggestion 3). `RateLimitEntry` gains `ObservedAt`, surfaced as `ghRateLimits.core.observed_at` and rendered on the card as `Measured 2m ago`, amber past five minutes. `reset` can't answer "how old is this number" — it moves independently of the sample — so a held reading and a fresh one looked identical. A held reading keeps the timestamp of the observation it came from, which is exactly what makes the hold visible. The field is **omitted** rather than zero when nothing has been observed, so it never renders as 1970.

## Deliberately not done

**Preferring `X-RateLimit-Remaining` response headers** (issue suggestion 2). It needs every call site to thread its response headers back to a shared observer — a much wider change — and the clamp already removes the false-full state from both display paths. Worth doing on its own terms, not as a rider here.

**A docs change.** The rate-limit card has no operator-doc section to extend (`api-reference.md` lists the route only, `operations.md` doesn't mention it). Rather than bolt a section onto an unrelated page, the mechanism is documented at length in `ratelimit_window.go` and the operator-facing behaviour is in the changelog fragment. Happy to add a doc section if you'd like one and can point me at the right home.

## Testing

- [x] `cd src && go build ./...`, `go vet ./pkg/github/ ./pkg/dashboard/`, `gofmt` clean on every touched file
- [x] `go test ./pkg/github/` and `./pkg/dashboard/` — both green in full
- [x] `go test ./pkg/github/ -run TestRateLimit -race` — clean
- [x] `node .github/scripts/check-inline-js.js src/pkg/dashboard/static/index.html` — syntax + `no-undef` passes
- [x] **Mutation-checked twice.** With the clamp removed entirely, 6 tests fail. With the *naive reset-keyed* clamp in its place, 5 fail — which is the point above, expressed as a test result.
- [ ] `cd src && go test ./...` — `pkg/agent`, `config`, `knowledge`, `proxy`, `sandbox` fail in this sandbox (no tmux, read-only `$HOME`); all pre-existing and unrelated to these packages, verified previously against an unmodified `upstream/v4` worktree.

The core test replays the reporter's measured trace minute by minute, asserting both the exact expected value and that the reported number never rises within a window. The others cover: rejecting the fresh bucket with a later reset; accepting a *genuine* rollover once the reset passes (so the card can't pin low forever); tracking a genuine decline; `observed_at` holding while clamped and re-stamping when a new reading is accepted; an unreported bucket passing through untracked (so an absent bucket can't pin at zero and read as "exhausted"); per-bucket independence; and concurrent observation under `-race`.

## Related issues

Fixes #5733

## Contributor checklist

- [x] Docs — see "deliberately not done" above.
- [x] `changelog.d/fixed-5733-ratelimit-monotone-window.md` carries the changelog entry.


clubanderson (approved):

Issue #5768, opened 2026-09-03
🐛 hosted-spoke test provisioning leaks hive-hosted-hosted-* namespaces — 76 stuck pods on a CI cluster


clubanderson:
Found via console's Live Promote canary (2026-09-03): the live console health pill showed '76 warnings', and the trace attribution work proved they are real — **64 Unschedulable + 12 Pending pods stuck on unbound PVCs across dozens of leaked `hive-hosted-hosted-*` namespaces** on one of console's CI clusters (two independent pod-issues stream samples agree; the other CI clusters report 0).

These are hive hosted-spoke provisioning namespaces created during tests/provisioning flows that were never torn down. Impact: quota/PVC leakage on a shared cluster, permanent noise in any health surface that reads pod issues, and it forced a console-side badge-attribution workaround (kubestellar/console#23092) just to keep the deploy gate honest.

Asks:
1. **Teardown**: hosted-spoke provisioning/test flows must delete their namespaces (and PVCs) on completion AND on failure — an ownership label + a janitor sweep for orphans older than N hours would close the class.
2. **One-time cleanup**: the leaked namespaces on the affected cluster need deleting (operator or a scoped job; the console deploy kubeconfig cannot see them).
3. Consider a fleet-side detector: leaked `hive-hosted-hosted-*` namespaces on clusters with no corresponding registry entry.

No internal hostnames in this issue beyond the CI cluster's generic name.

clubanderson:
**Correction to the original description, plus the exact namespace list.**

The claim that these pods are stuck on unbound PVCs is wrong and would send someone chasing a storage-class bug. Extracted from the console backend's own `/api/mcp/pod-issues?cluster=ks-console-ci-3` response (recovered from two promote-run Playwright traces 8.5h apart — byte-identical, so this is a stable stuck set):

- **62 of 76**: `Insufficient cpu, Insufficient memory` — the 2-node cluster is resource-exhausted
- **12 of 76**: Pending with no scheduler message
- **2 of 76**: unbound PVCs (the only ones matching the original description)

This makes the leak **self-reinforcing**: the leaked deployments consumed the cluster's capacity, so newly provisioned ones can no longer schedule. Root cause and fix are unchanged (teardown on both paths + labeled janitor), but the pressure is capacity, not storage.

Distribution: 76 records, 76 distinct namespaces (one stuck pod each), all matching `hive-hosted-*`; ci-1 and ci-2 report zero.

[...]

PR #5820, opened 2026-09-03
🐛 bug fix: stop hosted-spoke provisioning leaking namespaces, and report the ones already leaked

Danathar:
## Summary

A CI cluster had accumulated dozens of `hive-hosted-hosted-*` namespaces that nobody was going to clean up, holding 76 stuck pods between them. This PR does two things: it stops provisioning from creating new ones, and it makes the ones already out there visible.

**Stops the leak at the source.** When a hosted spoke's manifest fails to apply halfway through, the namespace it had already created was left behind forever. Provisioning now checks whether the namespace existed just before the apply, and deletes it on failure only if it didn't — so a failed provision cleans up after itself, and a namespace that was already there (a live spoke) is never touched.

**Reports the ones already leaked.** Each cluster's health entry gains a `leaked_namespaces` field listing `hive-hosted-*` namespaces the hub has no hive record for. It only reads; it deletes nothing.

**What it deliberately does NOT do:** automatically delete leaked namespaces fleet-wide. That is ask 1's janitor, and it is the follow-up, not this PR — see [Why no janitor](#why-there-is-no-janitor-yet) below.

Fixes part of #5768.

---

### 1. A failed `kubectl apply` left its namespace behind (ask 1, production side)

`provisionHive` renders the entire hosted spoke into one manifest and applies it with a single `kubectl apply -f`. The `Namespace` is necessarily the **first** object in that file — every other object is namespaced into it — so when the apply fails partway (an admission webhook rejects the Deployment, the PVC's StorageClass is missing, quota is exhausted, a Route host collides), the namespace has **already been created** and kubectl's non-zero exit rolls back nothing.

`provisionHive` then returns an error, the caller marks the hive record `error` (`saas.go` ~4118, `lite_enrollment.go` ~370), and the namespace stays on the cluster with whatever partial objects preceded the failure. Nothing retries the apply and nothing deprovisions an errored hive, so it is a leak with no owner.

**The fix** (`provision_namespace_rollback.go`): record the namespace's presence immediately *before* the apply, then roll back on failure in exactly one case.

| pre-apply state | on apply failure | why |
|---|---|---|
| **absent** | delete | the failed apply created it seconds ago; nothing else can be using it |
| **present** | keep | deleting a pre-existing namespace cascades to a live spoke's pods and PVCs |
| **unknown** | keep | "could not tell" must never be resolved into a destructive verb |

The tri-state is why the pre-check uses `get namespace --ignore-not-found`: with that flag kubectl exits 0 and prints nothing for an absent namespace, so *absent* and *the check failed* stay distinguishable. A plain `get` collapses both into one non-zero exit — and that reading ends in a namespace delete against an unreachable cluster.

The delete uses `--wait=false` (a PVC finalizer can hold a namespace for minutes, and waiting would pin a bounded provisioning-queue slot) and `--ignore-not-found` (the common case is an apply that failed before creating anything, which should be a clean no-op rather than a false leak in the logs). The rollback is best-effort and never changes the error the admin sees — that stays the original provisioning failure.

There is a genuine TOCTOU between the pre-check and the apply; it is called out in the file comment and accepted, because the only writer of `hive-hosted-<id>` namespaces is the hub itself, provisioning is serialized per hive through `enqueueProvision`, and the id is freshly minted for the hive being provisioned.

### 2. Nothing could see a namespace with no hive record (ask 3)

Every hub-side sweep over hosted namespaces derives its namespace list **from the registry** — `reapOrphanedPods` walks `listSaaSHives()` and calls `hostedNamespaceForHive` on each. That derivation is what confines the reaper to namespaces the hub provisioned, and it is exactly what makes this leak class invisible: **a namespace with no registry entry cannot appear in a registry-derived list.** The stuck-pod signal missed them too — its predicate requires a `deletionTimestamp`, and these pods were never asked to terminate; they are Pending/Unschedulable because their namespace outlived whatever should have deleted it.

`leaked_hosted_namespace.go` adds the complement: one `kubectl get namespaces -o json` per cluster per health build, reported as `PerClusterHealth.leaked_namespaces`.

The predicate — all four clauses must hold:

```
name has the hive-hosted- prefix &&
  name is NOT in the known (registry-derived) set &&
  phase != "Terminating" &&
  age(creationTimestamp) > 6h
```

`Terminating` is excluded because something *is* already cleaning that namespace up — and for one held by a stuck finalizer, counting it would mean counting it forever. The 6h window exists because provisioning is not atomic: there is a real interval, normally seconds, in which a legitimate namespace exists that the registry snapshot does not yet name. An unreadable `creationTimestamp` yields the zero time and is skipped — an age is never guessed in the direction of guilt.

Each entry carries the `hive.kubestellar.io/hive-id` label when present. Its **absence** is diagnostic in its own right: the stamp only happens after a successful apply, so an unstamped leaked namespace is a provision that died at the apply, while a stamped one is a deprovision that never finished.

## The load-bearing safety property

`listSaaSHives()` returns `nil` both when the hub genuinely hosts nothing **and** when it cannot read the hive directory at all (`os.ReadDir` error, `saas_provision.go` ~1810). Under an empty known set, *every* hosted namespace in the fleet satisfies "has no registry entry".

So without a guard, one transient disk hiccup would report the whole fleet as leaked — and would hand any future janitor a delete list containing every live spoke. `collectLeakedHostedNamespaces` therefore returns **nil (unknown), loudly**, rather than a report, when the known set is empty. `TestCollectLeakedHostedNamespacesRefusesToReportOnAnEmptyRegistry` pins this and is written so it cannot pass vacuously: the same fake cluster with a *non-empty* known set must still report its leaks.

Nil-means-unknown is the same contract `StuckPods` already carries: "no leaks" and "could not tell" must not render alike, since the entire cost of this incident was that nothing distinguished them.

The known set is also matched **fleet-wide, not per-cluster**, and built once per health build from the union of the on-disk hive records and the in-memory registry. A hive recorded against cluster A whose namespace shows up on cluster B is not convicted — the hub's per-hive `ClusterID` bookkeeping is not authoritative enough for that, and one snapshot means two clusters queried in parallel cannot disagree about what the registry said.

## Why there is no janitor yet

Ask 1 also wants a sweep that deletes orphans older than N hours. That is deliberately **not** in this PR.

A namespace delete cascades to every PVC and every pod inside it — the widest destructive verb the hub owns — and the rule that would drive it ("has no registry entry") has never been run against a real cluster. Shipping the deleter first, on a rule validated only by reading code, is how a cleanup tool becomes the outage it was meant to prevent. This report is the measurement that has to come first; once it has run on the affected cluster and the reported set matches what an operator agrees is garbage, the janitor is a small addition on top of the same predicate, with the registry-empty guard already in place as its precondition.

Ask 2 (one-time cleanup of the namespaces already on the affected cluster) is an operator action against that cluster and is out of scope for a code change here; this report is what tells the operator which ones.

## Cost and blast radius

- **New verbs:** `get namespaces` (read) per cluster per health build; `get namespace` (read) once per provision; `delete namespace` **only** on the absent-then-failed provisioning path. No new credential, kubeconfig, or client of any kind — everything rides the existing `kubectlForCluster` path.
- **Nothing cluster-wide is deleted.** The detector has no delete path at all, by construction.
- Pull-only clusters report unknown rather than a fabricated zero — the hub has no kubectl path into them.

## Testing

`go build ./...`, `go vet ./...`, and the full `./pkg/hub` suite pass (99s, no regressions). `gofmt` clean on all touched files.

Per the repo's #5388 rule, none of the new tests assert that a field exists:

- The predicate runs against a population whose leak count is **known by construction**, seeded with every shape a looser rule would wrongly convict: registry-known, `Terminating`, inside the age window, *exactly* at the age boundary (catching a `>` → `>=` slip), unreadable timestamp, and namespaces that are not ours at all. If the count picks up any of them, the number moves and the test fails.
- Truncation is asserted to keep the **oldest** entries, not an arbitrary 25, with names deliberately ordered against age so a name-order fallback would fail.
- The rollback is asserted against a **recording fake kubectl** — what was and was not executed — rather than a returned bool, because the harm being guarded against is an executed delete.
- `--wait=false` and `--ignore-not-found` are each pinned with the specific failure they prevent.


clubanderson:
Adopted in #5843 with credit to @Danathar and one additional safety fix from review. Closing this original fork PR so the adopted branch can merge.

PR #5827, opened 2026-09-03
🐛 bug fix: show which repo is lagging in the ACMM Evaluation panel

### LINKED ISSUE #5822: 🐛 bug: ACMM Evaluation reports one fleet-wide level that hides which repo is lagging
## Summary

The dashboard's **ACMM Evaluation** panel scores how mature a repo's agentic-contribution setup is — does it have a test suite, a CLAUDE.md, a coverage gate, a PR review rubric, and so on — and reports a maturity level from L1 to L6. A hive can watch several repos, but the panel shows a single fleet-wide level.

That single level is computed as the **best case across every repo**: a criterion counts as passed if *any* watched repo has it. A repo that meets none of the 43 criteria therefore does not lower the headline number at all — it is invisible. An operator looking at the panel cannot tell which repo is dragging, or by how much, without clicking through each repo one at a time.

A per-repo drill-down is partly built already, but it is half-wired: selecting a repo updates the level bars while the four headline tiles keep showing the aggregate.

Blast radius: dashboard UI plus one JSON field. The evaluation itself already computes per-repo results, so no new data collection or GitHub API calls are needed.

Observed on `v4` at 68fddabb.

### 1. The headline tiles ignore the repo selector

`acmmRenderCard` writes the four stat tiles — Codebase Readiness, Operational Autonomy, Overall ACMM, Criteria Passed — from `d`, the aggregate payload:

- `src/pkg/dashboard/static/index.html:15958-15962`

The active-repo view (`active = acmmGetActiveData()`) is not computed until line 15989, *after* the tiles are written, and is used only for the per-level bars below them. So clicking a repo chip changes the bars while the tiles above them still report hive-wide numbers. The two halves of the same card disagree, with nothing indicating that only one of them followed the click.

`acmmSelectRepo` (`:15944`) re-renders via `acmmRenderCard`, so the fix is scoped to that function.

### 2. The aggregate is a union, and the UI does not say so

`evaluateAllRepos` aggregates optimistically:

```go
// Aggregate: a criterion passes if it passes in ANY repo.
```
— `src/pkg/dashboard/api_acmm_eval.go:461`, applied at `:481` and `:498`.

This is a defensible definition for "what has this hive demonstrated anywhere," but it is the wrong number for "is my fleet ready." The Codebase Readiness dialog mentions it in passing ("A criterion passes if it exists in *any* repo") and only when more than one repo is configured; the panel itself does not. An operator reading `L3` reasonably assumes every watched repo is at L3.

### 3. Nothing ranks the repos or shows the size of the gap

To find the weakest repo today you click the Codebase Readiness tile, then expand each repo row individually (`:15798`). Nothing sorts by gap size. The repo chips carry a level (`repo (L2)`) but not the passed/total ratio, so "lacking, and by how much" takes N clicks for N repos.

### 4. Nit: the repo selector flashes before the first fetch

`src/pkg/dashboard/static/index.html:3273` sets both `display:none` and `display:flex` in one inline style attribute. The later declaration wins, so the selector renders visible — with only the "All (aggregate)" chip — until the first `/api/acmm/evaluation` response lets `acmmRenderCard` hide it.

## The data is already there

`GET /api/acmm/evaluation` returns `repo_results[]`, one `RepoEvaluation` per configured repo (`src/pkg/dashboard/api_acmm_eval.go:52`), each carrying `codebase_level`, `criteria_passed`, `criteria_total`, per-level scores, and the full per-criterion pass/fail list. This is a presentation gap, not a collection gap.

## Proposed fix

1. Make the four tiles honor `_acmmSelectedRepo` — use `active` rather than `d` for Codebase Readiness and Criteria Passed. Operational Autonomy is hive-wide and should stay on `d`, visibly labeled as such so the mixed scope is deliberate rather than accidental.
2. Add a per-repo table to the panel, sorted by gap: repo, codebase level, `passed/total`, and the level it is blocked at. That answers "which repo is lacking and to what degree" at a glance, with no clicks.
3. Label the aggregate as a union in the panel itself, not only inside the dialog.
4. Optionally surface a weakest-repo callout, and consider reporting a `min`-across-repos level alongside the union so the pessimistic reading is available too.
5. Fix the duplicate `display` declaration at `:3273`.

## Reproduction

1. Run a hive with two or more repos configured where the repos differ in maturity — e.g. one with `CLAUDE.md`, a coverage gate and issue templates, one with none of them.
2. Open the dashboard and expand **🎯 ACMM Evaluation**.
3. Observe the headline level reflects the better repo; the weaker one does not lower it.
4. Click the weaker repo's chip in the Repo row. The level bars update; the four tiles above do not.

## Environment

- Branch/version: `v4` @ `68fddabb`
- Deployment method: Hive dashboard (`src/pkg/dashboard`)
- Relevant configuration: any hive with more than one repo in `governor.repos`

Danathar:
## Summary

The ACMM Evaluation panel reported one level for the whole hive, and that level was the **best case** across every watched repo. A repo meeting none of the 43 criteria didn't lower it at all — so it was invisible. The per-repo drill-down existed but was half-wired: clicking a repo chip moved the level bars while the four headline tiles above them kept showing hive-wide numbers.

This makes the panel answer **"which repo is lagging, and by how much"** without a single click:

- The four headline tiles now follow the repo selector.
- A per-repo table ranks every watched repo weakest-first, with its level, `passed/total`, and the level it's blocked at.
- The panel says out loud that the aggregate is a union.
- The repo selector no longer flashes visible before the first fetch.

**No server change and no new API calls** — the evaluation already returned `repo_results[]`. See [Why no JSON field](#why-no-json-field).

Fixes #5822.

---

## 1. The tiles now follow the selector

`acmmRenderCard` wrote the four tiles from `d` (the aggregate) and only *then* resolved `active = acmmGetActiveData()`, which the level bars alone consumed. So a repo click moved the bars and left the tiles above them on fleet-wide numbers — two halves of one card disagreeing, with nothing to say which had followed the click.

The active view is resolved **before** the tiles are written, and the second, now-redundant resolution below them is gone: two resolutions are how the halves came to disagree in the first place.

| tile | scope after this change |
|---|---|
| Codebase Readiness | selected repo |
| Criteria Passed | selected repo |
| Overall ACMM | recomputed as `min(selected repo's codebase, operational)` |
| Operational Autonomy | **hive-wide, and labelled so** |

Overall needed recomputing because the server derives it from the **union** codebase level; leaving it would have parked a fleet-wide reading directly beside a per-repo one. Unscoped, the server's value stands, so the panel never disagrees with the payload it was handed.

Operational Autonomy deliberately does **not** follow the selector — it measures the hive's own operating behaviour, not a repository. That makes the card's scope mixed, which is fine as long as it's stated: an unlabelled tile that ignores a click is indistinguishable from one that *failed* to follow it. Every tile label now carries its scope, and the ops tile is marked `hive-wide` whenever a repo is selected.

## 2. A ranking that answers the question directly

Finding the weakest repo meant opening the Codebase Readiness dialog and expanding each repo row one at a time, with nothing sorted by gap. The chips carried a level but not the ratio, so "by how much" cost one click per repo.

The new table ranks worst-first and its rows select their repo, so the table and the selector are one control rather than a read-only sidebar next to one:

```
Per-repo readiness — weakest first
  hive-lab      weakest      L0    0/43    blocked at L1 Assisted
  console                    L2   25/43    blocked at L3 Supervised
  hive                       L4   40/43    blocked at L5 Autonomous
```

Three judgement calls worth naming:

- **Rank on the raw gap, not the percentage.** With a fixed criteria set the two agree, and the count is what an operator acts on — *"14 things to fix"* is a work estimate, *"67%"* is not.
- **"Blocked at" is the first failing level.** Levels must pass in order, so the first failure is the whole story; counting failures instead would rank a repo stuck at L2 below one with gaps at L5.
- **Ties break on level, then name.** This table is read repeatedly, and an order that shuffles between identical renders reads as churn that isn't there.

The `weakest` badge appears only when the repos actually differ — on a level fleet, singling one out would invent a problem, which is the same reason the union number needed qualifying at all.

Selection is carried by a left accent and font weight rather than a background, because the hover handlers own the background: a background-selected row would lose its highlight the moment the pointer left it.

## 3. The aggregate says what it is

`evaluateAllRepos` passes a criterion if *any* repo has it. That was stated only inside the Codebase Readiness dialog, and only when more than one repo was configured. The panel now carries it where the number is:

> Aggregate across 3 repos: a criterion counts as passed if *any* repo has it, so a repo meeting none of them does not lower these numbers. Pick a repo above, or read the table below.

## 4. The selector flash

`#acmm-repo-selector` carried both `display:none` and `display:flex` in one style attribute. The later declaration won, so it rendered visible — holding only the "All (aggregate)" chip — until the first response let `acmmRenderCard` hide it. Only `display:none` belongs inline; the flex layout arrives with the `display:flex` the renderer sets when there's something to select.

## Why no JSON field

The report floats an optional `min`-across-repos level. I left it out deliberately:

- It is derivable from `repo_results[]`, which the payload already carries, and the panel computes it there.
- I checked: **no Go consumer outside this panel reads `codebase_level` at all**, so the union optimism never escapes the dashboard payload. There is no second consumer to serve.
- A duplicated derivable field is a second source of truth that can drift from the array it was derived from.

If a non-UI consumer ever needs it, adding it is a few lines on top of this — but adding it now would be shipping a field with no reader.

## Testing

`go build ./...`, `go vet ./...`, and the full `./pkg/dashboard` suite pass (52s, no regressions). The CI `dashboard-lint` gate — `node --check` plus ESLint `no-undef` over the concatenated inline blocks — passes locally against the edited file.

**The ordering rule is tested by execution, not by pattern-matching.** The test extracts `acmmRepoGapRows` and `acmmFirstFailingLevel` straight out of `index.html` by brace-matching and runs them under `node` against a fixture whose correct order is known by construction and is neither the input order nor its reverse. Asserting that a sort comparator appears verbatim would catch a rule that was *changed* and miss one that was always *wrong* — which, for a ranking, is the failure that matters. It also covers a repo with every level passed (no bogus blocker), a repo with zero criteria (no divide-by-zero into `NaN`), and both tie-breaks.

The remaining five tests pin the wiring that can't be executed. One of them asserts the **ordering** of the active-view resolution against the tile writes, not just their values: an edit that moves the resolution back below the tiles reintroduces the original defect while every value-reading assertion still passes.

All six were confirmed to **fail against the pristine file** before the fix — the tile assertions naming each aggregate read still present, the ordering test reporting the resolution happens after the tiles, and the selector test reporting `2 display declarations`.


clubanderson:
Heads up: #5833 (merged) addressed #5822 by scoping the ACMM repo headlines, which is why this branch now conflicts with v4. If your per-repo lag display adds value beyond what landed, please rebase onto current v4; otherwise feel free to close. Happy to help either way — thanks for the contribution!

clubanderson:
Resolved the merge conflict against v4 (pushed to the head branch as a merge commit): both sides had implemented ACMM per-repo scoping — v4 picked up an interim variant meanwhile — and the resolution keeps this branch's fuller implementation (scope tags on the tiles, repo table with blocked-at column, in-panel aggregate note, recomputed Overall when scoped) while dropping the interim `acmm-eval-scope`/`acmm-repo-summary` elements it superseded. Deduplicated the `const repos`/`hasMultiRepo` redeclaration the auto-merge left in `acmmRenderCard`. `go test ./pkg/dashboard/` passes; script blocks validate with `node --check`.

/lgtm
/approve

PR #5840, opened 2026-09-03
🐛 fix(hub): tear down hosted namespaces on failed provisioning

### LINKED ISSUE #5768: 🐛 hosted-spoke test provisioning leaks hive-hosted-hosted-* namespaces — 76 stuck pods on a CI cluster

clubanderson:
Fixes #5768

## Root cause

`kubectl apply -f` is **not transactional**, and the Namespace is the **first** object in `k8sManifestTemplate` (`src/pkg/hub/saas_provision.go`). When a later object in that same manifest fails to apply — a PVC the cluster cannot bind, a Deployment rejected by admission, a missing Ingress class — the namespace and every object applied before the failing one are **already on the cluster**.

`provisionHive` then returns an error, and both of its callers do exactly one thing with it:

- `handleCreateHive` (`src/pkg/hub/saas.go` ~4118) — sets `Status="error"`, returns.
- the lite-enrollment path (`src/pkg/hub/lite_enrollment.go` ~370) — sets `Status="error"`, returns.

Neither deletes anything.

**Why nothing else ever cleaned it up.** `deprovisionHive` (`saas_provision.go` ~2446) is a complete and correct teardown — namespace, PV, OCI export, OCI file system, hive record, quota — but it is reachable **only** from the user-initiated delete handler (`saas.go` ~4310). A hive that never finished provisioning is not something a user is offered to delete, so the one function that could have removed the namespace was never called for precisely the hives that leaked it.

To be explicit about the alternatives in the issue: this was **not** a delete racing a still-terminating PVC, and **not** a test harness creating without registering cleanup. There is no test/e2e helper that provisions hosted spokes. The failure path simply had no teardown at all, which is also why the leak is dominated by pool-placeholder names (`hive-hosted-hosted-available-…-placeholder-…`) — the watermark pool replenisher (`pool_replenisher.go`) drives provisioning attempts unattended, so it produces the most failures with nobody watching any individual one.

This is original behaviour, not a regression: `git log` shows no commit that removed a teardown from this path.

## What changed

**1. Teardown on the failure path.** `provisionHive` now calls `teardownPartialProvision` before returning its error. Placing it inside `provisionHive` rather than in each caller means both call sites — and any future one — get it by construction.

It is deliberately **not** `deprovisionHive`. That function also deletes the cluster-scoped PV, the OCI export and file system, the on-disk hive record and the owner's quota. None of that is correct here: a failed provision must leave the hive **record** intact (that record is how the failure is reported and how a retry finds its slot), and the OCI resources may predate or be shared with this attempt. The narrow namespace delete is the right blast radius for cleanup that runs automatically on an error path. It uses `--wait=false` so a provisioning worker never blocks on finalizers, and it is best-effort — a teardown error must not replace the provisioning error the caller needs to report.

**2. Ownership labels**, stamped in the manifest itself:

```yaml
labels:
  hive.kubestellar.io/owner: provisioner
  hive.kubestellar.io/ephemeral: "true"
annotations:
  hive.kubestellar.io/created-at: "<RFC3339>"
```

Stamped at **creation**, in the same apply that makes the namespace — not by a follow-up `kubectl label`, which would be missing on exactly the failed-provision namespaces the janitor exists to find. The `created-at` annotation is preferred over `metadata.creationTimestamp` for age (a restored namespace gets a fresh `creationTimestamp`), with `creationTimestamp` as the fallback for namespaces predating this change.

**3. Janitor sweep** (`src/pkg/hub/hosted_namespace_janitor.go`) — the backstop for when the inline teardown itself fails or the hub dies between the failed apply and the delete. It runs in the hub's **existing** SHA-poller reconciliation lane alongside `reapOrphanedPodsIfDue` and `replenishPoolsIfDue` — no new goroutine, no new scheduler — throttled internally to a named 15m interval.

## The janitor's safety guards

**The invariant: it never touches a namespace lacking `hive.kubestellar.io/ephemeral=true`.** Selection is by **label**, never by name prefix. Matching on `hive-hosted-` would be the dangerous design — an operator's hand-made namespace, or one restored from backup, would become deletable by a background loop. The label is an explicit statement by the creator that the namespace is disposable.

Three further guards, each independently sufficient to spare a namespace:

- **Registry cross-check.** Every hosted namespace carries the ephemeral label, *including every healthy running spoke* — so the label alone would select the entire fleet. The registry check, not the label, is what separates a leak from a spoke.
- **Age** — nothing under `hostedNamespaceLeakMinAge` (6h, ~72x the 5-minute `provisionTimeout`), so an in-flight provision whose registry write hasn't landed cannot be reaped out from under itself.
- **Phase** — anything already `Terminating` is skipped.

Plus: **an empty registry aborts the sweep** rather than interpreting it (a registry that failed to load would otherwise make every namespace on every cluster look unregistered and reapable); a **20-delete-per-cycle cap** that logs loudly when hit; per-call kubectl timeouts; pull-only and recently-unreachable clusters skipped and counted, not failed; and every reap logged at INFO with namespace, cluster and age.

**Dry run:** `HIVE_HUB_NS_JANITOR_DRY_RUN=true` reports what it would delete and deletes nothing, read per sweep so it can be flipped without a hub restart. A sweep that deletes namespaces should not be debugged in production for the first time.

**Credentials:** no new credential, kubeconfig or client. It rides the same `kubectlForCluster` path provisioning already uses and issues a strict subset of what `deprovisionHive` already issues against these clusters.

## Second root cause: unit tests were provisioning against a REAL cluster

Added after evidence on the issue showed the leaked namespace names are this package's **unit-test fixtures, character for character** (`acme-repo-abcd`, `acme-app-wxyz`, `apporg/repo1`, `myorg/repo1`). Teardown and the janitor stop the bleeding; this is what was causing it.

`ClusterConfig.InCluster` is a **claim, not a fact**, and nothing downstream re-checked it:

1. Test helpers point `clustersConfigPath` at an empty `t.TempDir()`, so no `clusters.json` exists.
2. `loadClustersChecked` treats absent as `clustersAbsent` and synthesises `defaultClusterRegistry()` (`clusters_registry.go:199`) — one entry, **`InCluster: true`**.
3. `kubectlArgsForCluster` (`saas_provision.go:414`) added `--server`/`--token` **only when** `KUBERNETES_SERVICE_HOST`/`_PORT` are set. Outside a pod they are not, so it added **no connection flag at all**.
4. `kubectl` with neither `--kubeconfig` nor `--server` resolves against **ambient** config (`~/.kube/config`, `$KUBECONFIG`).
5. Any test reaching provisioning issued a **real** `kubectl apply` — real Namespace, Deployment, PVC — against whatever cluster the runner pointed at, named from the test's fixture org.

**Reproduced:** driving `handleCreateHive` with `{"org":"apporg","repos":"repo1"}` and no scripted kubectl, then draining `provisionWG`, executes the real kubectl binary and attempts a genuine apply of `hive-hosted-hosted-apporg-repo1-<suffix>`. It failed locally only because this machine has no cluster — on a runner with a live kubeconfig it creates the namespace.

**Why the fake never caught it:** `installScriptedKubectl` is only a `PATH` prepend — it constrains *which binary* runs, never *which cluster it talks to* — and **six of seven** test files calling `handleCreateHive` never install it. `HIVE_HUB_TEST_ISOLATED` is set by `newHubServerForTest` but **never read by production code**; it is a parallel-test marker, not a kubectl guard.

**The fix** goes at the single argv chokepoint both constructors funnel through, not in the tests — a per-test fix would have to be repeated forever and re-missed once. An `InCluster` claim with no ServiceAccount environment now aims at the existing `unreachableKubeconfigSentinel`, the same mechanism the pull-only branch already uses. Absence of those env vars is a positive signal the process is not in a pod (a real hub pod always has both), so genuine in-cluster provisioning is untouched.

`TestKubectlForCluster` asserted *"in-cluster should not use --kubeconfig"* with the env vars unset — it **encoded the bug**. Updated to set them, pinning the real-pod behaviour it meant to describe.

**Name-family split** (the two have different causes): `available-<cluster>-<suffix>` comes from `seedPlaceholder` (`pool_replenisher.go:142`), which explains that family's volume; `acme`/`apporg`/`myorg` appear **nowhere in production code** and cannot come from the replenisher — that family is the test escape.

**On candidate (1):** no `clusters.json` exists in the repo and no reference to the CI cluster name anywhere in the tree, so I could not confirm or rule out a hub having it registered. Worth an operator check against the live hub's `/data/saas/clusters.json`, but no longer needed to explain the fixture-named namespaces.

### Guard tests

- `TestKubectlNeverRunsAgainstAmbientClusterOutsidePod` — over every cluster shape, the argv must always pin a target. **Verified to fail** pre-fix, where the argv was literally `apply -f manifest.yaml`.
- `TestRealInClusterStillUsesServiceAccount` — proves the guard doesn't break real in-pod provisioning.
- Plus sentinel-targeting and end-to-end handler assertions.

## Tests

Both new tests are **proven to have teeth** by mutation:

- `TestJanitorNeverSelectsUnlabelledNamespace` — a property test over 2000 randomised namespaces that are maximally attractive in every *other* dimension (ancient, unregistered, Active, carrying the owner label, carrying the `hive-hosted-` prefix, carrying near-miss ephemeral values like `"True"`/`"1"`/`"true "`). Verified to **fail** when the predicate is loosened to also accept the name prefix.
- `TestInvariantCanFail` — runs the same generator against a deliberately prefix-matching predicate and requires it to be caught, so the property can never silently degrade into proving nothing.
- `TestProvisionFailureTearsDownNamespace` — drives `provisionHive` against a kubectl whose `apply` fails on an unbindable PVC (the real incident's shape) and asserts a `delete namespace` is issued. Verified to **fail** against the pre-fix code.
- Plus registry/age/phase/boundary guards, the `parse`/`select` seams, dry-run parsing, and a test asserting the manifest's labels and the janitor's selector cannot drift apart.

`gofmt` clean; `go vet ./pkg/hub/` clean; full `go test ./pkg/hub/` passes (129s).

## Required follow-up: one-time cleanup (operator)

**This PR stops new leaks and reaps future ones; it does not clean the existing mess.** No live-cluster cleanup was attempted here — I don't have credentials for the affected cluster.

The already-leaked namespaces predate this change and therefore **carry no ephemeral label**, so the janitor will correctly refuse to touch them — that is the invariant working as designed, not a gap. They need a one-time operator deletion. Recommended sequence:

1. List candidates: `kubectl get ns -o name | grep '^namespace/hive-hosted-hosted-'`
2. Cross-check each against the hub registry and delete only those with no entry.
3. Optionally, label any survivors that *should* be reapable with `hive.kubestellar.io/ephemeral=true` so the janitor manages them from then on.

Deploying with `HIVE_HUB_NS_JANITOR_DRY_RUN=true` for a cycle or two first is recommended, to confirm the selection against the real cluster before it deletes anything.


clubanderson:
Superseded: #5768 was fixed by #5843 (adopted from #5820 with credit), merged to v4. The merge conflict here is with that landed fix.

PR #5874, opened 2026-09-03
🐛 bug fix: stop lane routing matching keywords inside a label's name

### LINKED ISSUE #5856: 🐛 classify: the ai-fix-requested label routes every ACMM gap issue to scanner, the one L4 agent that cannot fix it
## Summary

Hive routes each actionable GitHub issue to one agent ("lane") by matching keywords against the issue's title **and its labels**. Separately, the dashboard's ACMM evaluator files maturity-gap issues and labels them `acmm` + `ai-fix-requested`. Those two features collide: the scanner agent's lane keyword `fix` is a substring of `ai-fix-requested`, so **every ACMM gap issue hive files is routed to scanner** — which at ACMM L4 is `ISSUES_ONLY` and cannot open a pull request. The operator sees ACMM gap issues pile up untouched forever while other agents that could fix them never see the issue in their kick prompt at all. Observed live on a 4-repo L4 hive: 11 of 11 ACMM issues routed to scanner. Blast radius: issue→agent routing for any issue whose labels happen to contain a lane keyword as a substring; no data loss, no PR/merge behavior change.

## Symptom

On an L4 hive, ACMM gap issues are never fixed by any agent. They stay open indefinitely. Nothing logs an error — the issues are correctly enumerated as actionable, they simply arrive in the one agent's queue that has no ability to act on them.

Live hive `hive-wild-mole`, repo `Danathar/zfs-kinoite-complex`, all labeled `acmm` + `ai-fix-requested`, all authored by the hive's own App:

```
#21 [ACMM L2] Add Copilot instructions      #27 [ACMM L0] Add E2E tests
#22 [ACMM L2] Add Cursor rules              #28 [ACMM L0] Add PR template
#23 [ACMM L2] Add Prompts catalog           #29 [ACMM L0] Add Issue templates
#24 [ACMM L2] Add EditorConfig              #30 [ACMM L0] Add Code style config
#25 [ACMM L2] Add Simple skills             #31 [ACMM L0] Add Coverage gate
#26 [ACMM L2] Add Correction capture
```

All 11 classify to lane `scanner`. The L4 pack (`src/pkg/config/packs/level-4.yaml`) gives scanner `mode: ISSUES_ONLY`, so it can file and analyze but never open a PR. `quality`, `ci-maintainer`, and `sec-check` are `ISSUES_AND_PRS` at L4 — and none of them ever see these issues, because `filterByLane` (`src/pkg/scheduler/scheduler.go:1442`) admits an issue only when `issue.Lane == agentName || issue.Lane == ""`.

## Root cause

`classifyLane` (`src/pkg/classify/classifier.go:162`) matches lane keywords against the concatenated label string with `strings.Contains`, in the same pass it matches the title:

```go
for _, lane := range activeLanes() {
    for _, kw := range lane.Keywords {
        if strings.Contains(titleLower, kw) || strings.Contains(labelsStr, kw) {
            return Lane(lane.Name)
        }
    }
}
```

Scanner's L4 lane keywords are `bug, triage, fix, issue` (`level-4.yaml:97`). The ACMM issue creator hardcodes the labels at `src/pkg/dashboard/api_acmm_eval.go:244`:

```go
issueReq.Labels = &[]string{acmmIssueLabelName, "ai-fix-requested"}
```

`"ai-fix-requested"` contains `"fix"`. Every ACMM issue therefore matches scanner on the label, regardless of its title.

The collision is demonstrable by toggling only the label — same title, same lane table:

```
[ACMM L4] Add AI security policy  labels=""                      -> sec-check  (title kw "security")
[ACMM L4] Add AI security policy  labels="acmm ai-fix-requested" -> scanner    (LABEL kw "fix")
```

Without the label the issue routes to `sec-check`, which is `ISSUES_AND_PRS` at L4 and would fix it. With the label — the label whose entire purpose is to mark the issue as AI-fixable — it routes to the one agent that cannot.

Note also that `DefaultLane = "scanner"` (`classifier.go:35`) means removing `fix` from scanner's keywords is **not** sufficient on its own: an ACMM title matching no lane still falls back to scanner, and because the fallback is `"scanner"` rather than `""` it never passes another agent's `filterByLane`.

## Second, related defect: lane order is non-deterministic

`lanes` is built by ranging over `cfg.Agents`, a Go map (`src/cmd/hive/main.go:9024`):

```go
for name, agent := range cfg.Agents {
    if len(agent.LaneKeywords) > 0 {
        lanes = append(lanes, classify.LaneConfig{Name: name, Keywords: agent.LaneKeywords})
    }
}
```

`classifyLane` is first-match-wins over that slice, so when an issue matches two lanes the winner depends on map iteration order and can differ between process restarts. The `security policy` example above matches both `sec-check` (title `security`) and `scanner` (label `fix`) — which one wins today is not stable. Sorting `lanes` by name, or by an explicit precedence, would make classification reproducible.

## Suggested fixes

1. **Match label keywords on word boundaries**, or match labels only as whole tokens after splitting on the label separator, so `fix` no longer matches inside `ai-fix-requested`. This is the narrow fix and preserves the intent of label-based routing.
2. **Sort `lanes` deterministically** at `main.go:9024` before `classify.SetLanes`, so a two-lane match resolves the same way every run.
3. Optionally, drop `fix` from scanner's default lane keywords — it is broad enough to catch many unrelated labels and titles — though on its own this leaves the `DefaultLane = "scanner"` fallback in place.

Not proposed here: changing the ACMM labels themselves. `ai-fix-requested` is the documented contract with the repo-side workflow that consumes it (`src/docs/acmm-policy-matrix.md`, and the `acmm:ai-fix-workflow` L4 criterion), so the routing layer is the right place to fix this.

## Workaround for operators

Label-name routing runs *before* keyword routing (`classifier.go:172-176`), so adding a label containing a PR-capable agent's name overrides the collision today:

```
labels="acmm ai-fix-requested"          -> scanner       (keyword "fix")
labels="acmm ai-fix-requested quality"  -> quality       (lane name in labels)
```

Add exactly one agent-name label — two would reintroduce the map-order nondeterminism above.

Danathar:
## Summary

Scanner's L4 lane keyword `fix` is a substring of `ai-fix-requested` — the label the ACMM evaluator stamps on every maturity-gap issue it files. Lane routing matched keywords against the labels with `strings.Contains`, so **every ACMM gap issue routed to scanner**, the one L4 agent that is `ISSUES_ONLY` and cannot open a PR. The label whose entire purpose is to mark an issue AI-fixable was the reason nothing fixed it.

Two changes:

1. **A keyword matches a label only as the whole label, or as a whole `/`-delimited segment of it.** `ai-fix-requested` no longer routes on `fix`; `kind/regression` still routes on `regression`.
2. **The lane table is sorted before it reaches the classifier**, so a two-lane match resolves the same way every run instead of by Go's randomized map order.

Fixes #5856 — with one measured caveat about what it does *not* move, in [What this does not do](#what-this-does-not-do).

---

## 1. The label collision

Reproduced exactly as reported, toggling **only** the label with the title and lane table held fixed. Measured against the pre-fix classifier and after:

| title | labels | before | after |
|---|---|---|---|
| `[ACMM L4] Add AI security policy` | *(none)* | `sec-check` | `sec-check` |
| `[ACMM L4] Add AI security policy` | `acmm ai-fix-requested` | **`scanner`** | `sec-check` |

`sec-check` is `ISSUES_AND_PRS` at L4 and would have fixed it. Nothing logged an error — the issues were enumerated as actionable correctly, they just arrived in the one queue that cannot act, and `filterByLane` admits an issue to an agent only when `issue.Lane == agentName || issue.Lane == ""`, so no other agent ever saw them.

### The rule, and why `/` and `-` differ

The asymmetry between the two separators *is* the fix:

- **`/` is a GitHub label namespace separator** — `kind/bug`, `kind/regression`, `area/api`. The segment is the label's actual subject, so ci-maintainer's `regression` keyword matching `kind/regression` is routing working as intended. This package already leans on the convention: `classifyTier` tests `kind/security` and `kind/regression` by name.
- **`-` is part of one label's own name.** `ai-fix-requested` is a single word meaning a single thing; it is no more a request about "fix" than `do-not-merge` is a request about "merge". **Splitting on `-` would have left the bug exactly where it was**, which is why this splits on `/` only.

Titles are untouched and keep substring matching — right for prose, where "fix the login crash" should reach the `fix` lane.

### A second bug this closes

The joined-label string let a multi-word keyword straddle two unrelated labels: keyword `breaking change` matched an issue labelled `breaking` **+** `change`, because the join put a space between them. Matching per-label makes that impossible.

## 2. Non-deterministic lane order

`classifyLane` is first-match-wins over a slice `initAgentConfigDrivenSystems` built by ranging over `cfg.Agents` — a Go map, iteration order randomized per range. An issue matching two lanes went to whichever came out first, and the winner could differ between two runs of the same binary on the same config, with nothing recording which it had been.

Not hypothetical: `[ACMM L4] Add AI fix workflow` matches **ci-maintainer**'s `workflow` and **scanner**'s `fix`, and those two have different modes — so the coin flip decided whether the issue was fixable at all.

Sorting by name is a **stable** order, not a claim about precedence: it does not assert architect deserves an issue more than scanner does. It buys reproducibility, so a misroute is a bug someone can chase instead of an intermittency. Picking a deliberate precedence between colliding lanes is a separate call for whoever owns the lane table.

## What this does not do

Measured, not estimated — this is the part worth reading before merging.

**The fix makes the ACMM labels routing-neutral; it does not re-home every ACMM issue.** Of the 13 titles in the report, **11 still reach scanner** — one on the title keyword `issue`, ten on `DefaultLane`, which is `"scanner"` rather than `""`. Because `filterByLane` treats only `""` as "any agent", an unmatched issue stays invisible to every other agent. So on the specific hive in the report, the issue count parked on scanner does not change; what changes is that the *label* is no longer the reason.

The report raises the fallback as a separate matter and does not propose changing it, and I did not: flipping `DefaultLane` to `""` would make every unclassified issue in the fleet visible to every agent's kick prompt, which is a routing-policy decision, not a bug fix. `TestACMMLabelIsRoutingNeutral` **asserts the count of 11** so that decision has data attached and moving it has to be deliberate.

Also not done, and for the same reason: **dropping `fix` from scanner's default L4 keywords**, which the report lists as optional. Measured against the reported issue set it changes nothing — none of those titles contain `fix` — and it would narrow scanner's triage reach on every L4 hive.

## One finding not in the report

While measuring the L4 lane table: **`quality` is `ISSUES_AND_PRS` at L4 and declares no `lane_keywords` at all**, so it is absent from the lane table entirely and can never be routed anything. That is a second, independent reason the reported issues could not reach it — the report notes quality never sees them, and this is why, beyond the label collision. Giving it keywords would route several of the stuck titles (`Add E2E tests`, `Add Coverage gate`, the template and config gaps), but choosing which agent owns ACMM-gap work is a lane-semantics decision I left to the maintainers rather than inventing keywords. `TestInitAgentConfigDrivenSystemsKeywordlessAgentsStayOutOfTheTable` pins the build behaviour that makes a keywordless agent a non-lane.

## Testing

`go build ./...`, `go vet ./...` clean. `./pkg/classify`, `./pkg/scheduler`, `./pkg/dashboard`, `./pkg/planning`, `./cmd/hive` all pass. (`./pkg/config` has one failure, `TestGateFailsClosedWithoutIP6Tables`, which I verified fails identically on the pristine tree — it needs `ip6tables`, absent in my sandbox.)

**Both defects were confirmed to reproduce before the fix**, not just asserted after:

- Against the pre-fix classifier, the security-policy title routes to `scanner` with the label and `sec-check` without it — the reported collision, reproduced in the table above.
- The determinism test fails on **iteration 0** with the sort removed: `lane = "scanner", want "aardvark"`.

The determinism test is behavioural rather than structural, because `activeLanes()` is unexported: it configures six lanes that all match one issue, so the answer *is* the head of the table, and rebuilds the table 40 times to re-range the map. Without the sort, passing all 40 by luck has probability (1/6)⁴⁰.

The label rule is pinned clause by clause — including the two cases that could have been broken by over-narrowing (`kind/regression` must still route; the operator's exact lane-name label override must still win) and the `do-not-merge` case, which is the same shape as the bug with the opposite meaning.


clubanderson:
/lgtm
/approve

Reviewed: whole-label / '/'-segment token matching kills the `fix`-inside-`ai-fix-requested` misroute while preserving `kind/regression` namespace routing; sorted lane table removes the map-order coin flip; tests pin both. LGTM.

PR #5881, opened 2026-09-03
🐛 fix(hub): close the ambient-kubeconfig escape that leaked 76 namespaces onto a live cluster

clubanderson:
## What

`kubectlArgsForCluster` (`src/pkg/hub/saas_provision.go:414`) treated `ClusterConfig.InCluster` as a fact. For an `InCluster` cluster it appended `--server` / `--certificate-authority` / `--token` **only** when `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` were both set, and appended **nothing at all** otherwise.

"Nothing at all" is not a safe default. `kubectl` with no `--server` and no `--kubeconfig` resolves against its **ambient** configuration — `$KUBECONFIG` or `~/.kube/config` — so the command targets whatever cluster the machine's current context happens to name.

This is the mechanism behind #5768: 76 `hive-hosted-hosted-<fixture-org>-*` namespaces on the live `ks-console-ci-3` cluster, each with one never-ready `hive` Deployment, and **no hub bookkeeping for any of them**.

## Why bookkeeping was absent

The two halves of provisioning have different destinations under test:

- `saveSaaSHive` (`saas_provision.go:1757`) writes under the package var `saasHivesDir`, which essentially every hub test reassigns to a `t.TempDir()`. The record **was** written — then deleted with the temp dir.
- the `kubectl apply` went out through the argv chokepoint above, with no connection flag, to the **real** ambient cluster.

So "namespace exists, no record" is the expected signature of this escape, not evidence of an exotic code path. It also explains why `grep ks-console-ci` over `hub-registry.json` returns nothing and why all 95 `/data/saas/hives/` entries reconcile to the four legitimate clusters.

## Why the fake didn't catch it

`installScriptedKubectl` is a `PATH` prepend. It constrains *which binary* runs, never *which cluster it talks to* — and **6 of the 7** test files that drive `handleCreateHive` never install it at all (only `saas_handlers2_coverage_test.go` does). Their fixture orgs are exactly the leaked names: `apporg`, `myorg` (`hub_filesystem_extra_test.go`, `hub_coverage_deep_test.go`) and `acme` (`final5_coverage_test.go`).

The path into a real cluster is short: test helpers point `clustersConfigPath` at an empty `t.TempDir()` → `loadClustersChecked` reports `clustersAbsent` → `defaultClusterRegistry()` (`clusters_registry.go:199`) synthesises one entry with **`InCluster: true`** → the chokepoint adds no flag → ambient cluster.

`TestKubectlForCluster` (`hub_coverage_boost_test.go:1836`) asserted *"in-cluster should not use `--kubeconfig`"* **with those env vars unset** — it encoded the bug as expected behaviour.

## The fix

The guard goes at the single argv chokepoint both `kubectlForCluster` and `kubectlForClusterContext` funnel through, not in individual tests — a per-test fix would have to be repeated forever and re-missed once. An `InCluster` claim with **no ServiceAccount environment** now aims at the existing `unreachableKubeconfigSentinel`, the same mechanism the pull-only branch already uses.

Absence of both env vars is positive, reliable evidence the process is not in a pod: the kubelet always injects them into a real hub pod. Genuine in-cluster provisioning is therefore untouched.

## Tests, and proof they can fail

`src/pkg/hub/kubectl_ambient_escape_test.go`:

- `TestInClusterClaimWithoutServiceAccountEnvCannotUseAmbientConfig` — the core invariant
- `TestDefaultClusterRegistryEntryCannotEscapeToAmbientConfig` — the exact synthesised-registry path the leak took
- `TestRealInClusterStillUsesServiceAccount` — pins that a real pod still gets its ServiceAccount, so the escape can't be "closed" by breaking production
- `TestUnreachableRemoteClusterUnchanged` — pins pull-only/reachable behaviour as unchanged

Verified failing without the fix (else-arm neutralised, everything else identical):

```
--- FAIL: TestInClusterClaimWithoutServiceAccountEnvCannotUseAmbientConfig
    ambient-config escape: InCluster claim with no ServiceAccount env produced
    no --server and no --kubeconfig, so kubectl would target the machine's
    current context. argv=["apply" "-f" "-"]
--- FAIL: TestDefaultClusterRegistryEntryCannotEscapeToAmbientConfig
    synthesised default cluster would provision against the ambient cluster:
    argv=["create" "namespace" "hive-hosted-hosted-acme-repo-abcd"]
```

and passing with it.

## Relationship to the existing TestMain guard

`src/pkg/hub/main_test.go` clears these same env vars — and it was **created 2026-07-31** (`377fefd8`, #2287) in response to the 2026-07-27 incident its own comment documents. Both leak events predate it: the `apporg`/`myorg` bulk at 52-55 days (~2026-07-10..13) and the `acme` pair at 36 days (~2026-07-29). At `1b4b6a01` (2026-07-25) the hub package had **no `TestMain` at all**, which is why the leak was possible then and has not recurred since.

That guard is test-side and covers only this package's `go test` runs. This change puts the same invariant in production code at the chokepoint, so it holds for any caller in any process — which is what makes it durable rather than another thing to remember.

No cluster objects are touched by this PR; the one-time cleanup of the 76 existing namespaces remains an operator action tracked in #5768.

Refs #5768

### DIFF
--- a/changelog.d/fixed-5768-ambient-kubectl-escape.md
+++ b/changelog.d/fixed-5768-ambient-kubectl-escape.md
@@ -1,0 +1,1 @@
+- Stopped hub provisioning from falling back to a machine's ambient kubeconfig. A cluster config claiming `in_cluster` outside a real pod now aims kubectl at the unreachable sentinel and fails loudly, instead of applying namespaces and Deployments to whatever cluster the current context named.
--- a/src/pkg/hub/hub_coverage_boost_test.go
+++ b/src/pkg/hub/hub_coverage_boost_test.go
@@ -1834,7 +1834,13 @@ func TestLoadClustersDefault(t *testing.T) {
 // ============================================================
 
 func TestKubectlForCluster(t *testing.T) {
-	// In-cluster
+	// In-cluster. The ServiceAccount environment must be SET for this case:
+	// "in-cluster does not use --kubeconfig" is only true of a real hub pod,
+	// where the kubelet injects both vars. Asserting it with them unset is what
+	// encoded the #5768 ambient-config escape as expected behaviour — see
+	// kubectl_ambient_escape_test.go for the guarded no-env case.
+	t.Setenv("KUBERNETES_SERVICE_HOST", "10.96.0.1")
+	t.Setenv("KUBERNETES_SERVICE_PORT", "443")
 	inCluster := &ClusterConfig{InCluster: true}
 	cmd := kubectlForCluster(inCluster, "get", "pods")
 	args := cmd.Args
--- a/src/pkg/hub/kubectl_ambient_escape_test.go
+++ b/src/pkg/hub/kubectl_ambient_escape_test.go
@@ -1,0 +1,144 @@
+package hub
+
+import (
+	"strings"
+	"testing"
+)
+
+// The tests below pin the fix for issue #5768.
+//
+// A hosted-spoke provisioning call builds its kubectl argv through the single
+// chokepoint kubectlArgsForCluster. For a cluster whose config CLAIMS
+// InCluster, that function used to append the in-cluster connection flags only
+// when KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT were both set, and
+// nothing at all otherwise. "Nothing at all" is not a safe default: kubectl
+// with no --server and no --kubeconfig resolves against its ambient
+// configuration, so the command executes against whatever cluster the machine's
+// current context names.
+//
+// That is how 76 hive-hosted-hosted-<fixture-org>-* namespaces reached a live
+// CI cluster with no corresponding hub bookkeeping — the applies were real
+// while the store writes went to a per-test t.TempDir().
+
+// hasFlag reports whether argv contains flag.
+func hasFlag(argv []string, flag string) bool {
+	for _, a := range argv {
+		if a == flag {
+			return true
+		}
+	}
+	return false
+}
+
+// flagValue returns the value following flag, or "" when absent.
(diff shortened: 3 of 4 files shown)


clubanderson:
Follow-up evidence from the issue ([addendum](https://github.com/hivecommons/hive/issues/5768#issuecomment-5528508931)) that reinforces *why* the guard belongs at the argv chokepoint rather than in the tests:

The `clustersAbsent` → `defaultClusterRegistry()` route in the description is only one of two ways in, and the shorter one is more common. **52 literal `InCluster: true` cluster configs** appear across the hub tests, and five files call `provisionHive` **directly** with a hand-built `&ClusterConfig{InCluster: true}`, bypassing the registry entirely — e.g. `final_gaps_coverage_test.go:30`, which is the origin of the 76th leaked namespace `hive-hosted-hosted-nfs` (a hardcoded `ID: "hosted-nfs"`, not a generated one).

Two details worth noting:

- `final_gaps_coverage_test.go` *does* install a scripted kubectl elsewhere in the file and still leaked. A `PATH` prepend constrains the binary, not the cluster, and it is per-helper — a sibling test in the same file that skips it is unprotected.
- That test's `t.Fatalf` on error means it was **passing**. Its OCI half was stubbed to a local `httptest` server, so `provisionHive` returning nil requires the `kubectl apply` calls to have **succeeded** against a real API server. A green test was the leak.

With 52 hand-built configs and five direct entry points, a per-test or per-registry fix leaves most of the surface open. `kubectlArgsForCluster` covers all of them, including call sites written after this merges.

PR #5886, opened 2026-09-03
🐛 bug fix: run the pinned backend smoke where podman exists, and stop blaming the CLIs for a runner fault

### LINKED ISSUE #5883: [Backend smoke: pinned] codex, claude failure on d7571a1
## Backend smoke failure

- **Lane:** pinned
- **Backend(s):** codex, claude
- **Commit:** `d7571a155c973a36e4741144811e29964ad9af6a`
- **Run:** https://github.com/hivecommons/hive/actions/runs/33788303199

The CLI versions hive SHIPS in the contributor image failed the smoke — contributors are likely broken right now.

The run log carries the suite's evidence block (relay log tail, fake-hub
wire messages, tmux pane capture) for each failed scenario, and the
`backend-smoke-<backend>-<lane>` artifact holds the full suite output.

Danathar:
## Summary

The pinned lane isn't failing because the shipped CLIs are broken — **it never ran**. #5862 moved this workflow to the self-hosted fleet, and the fleet image has no podman:

```
/runner/_work/_temp/....sh: line 1: podman: command not found
##[error]Process completed with exit code 127
```

The pinned lane runs the suite *inside* `ghcr.io/hivecommons/hive-contributor`, so it is image-heavy by definition — the same category #5862 deliberately kept on hosted runners (docker, overlayfs-exec-guard). It was missed. Every scheduled run has died at the image pull in ~7s since, so the CLI versions hive ships have been **unverified**, not broken.

Three changes:

1. **The pinned lane goes back to `ubuntu-latest`**; the latest lane stays on the fleet.
2. **A missing container engine says so itself** instead of exiting 127.
3. **The failure-issue job stops guessing** — and this issue is one of its false accusations.

Closes #5883.

---

## The evidence

| | runner | result |
|---|---|---|
| pinned, [run 33735649461](https://github.com/hivecommons/hive/actions/runs/33735649461) (before #5862) | `GitHub Actions 1000002245`, labels `["ubuntu-latest"]` | ✅ 42s |
| pinned, [run 33788303199](https://github.com/hivecommons/hive/actions/runs/33788303199) (after) | `hivecommons-hive-runners-lbz5b-zlf2k`, labels `["self-hosted","hive"]` | ❌ 7s, exit 127 |
| **latest**, same failing run | the same self-hosted fleet | ✅ both arms |

The latest lane needs only node, tmux and an npm-installed CLI on the bare runner, so it is genuinely fine on the fleet — which is why this splits the lanes rather than reverting #5862. The queue relief that motivated the move is kept for the half that can take it.

The split lives in `runs-on`, which can read the `matrix` context. (The job-level `if` cannot — this file already carries that note from #5550, so it's worth being explicit that the two differ.)

## A missing engine now names itself

`podman: command not found` / exit 127 is not a diagnosis, and it is exactly what the next runner migration will produce again. A preflight step names the real cause, emits it as `::error`, and writes it into `smoke.log` so the evidence artifact isn't empty for an environment failure — the artifact is what the filed issue tells you to read.

**Deliberately not done:** adding a docker fallback. `ubuntu-latest` has podman — it's where this lane passed until today — so a second engine path would be code I can't exercise, added on the strength of reading alone.

## The failure-issue job stops guessing

This is the part worth reviewing most closely, because **this issue is the bug**. The job filed:

> **[Backend smoke: pinned] codex, claude failure on d7571a1**
> The CLI versions hive SHIPS in the contributor image failed the smoke — contributors are likely broken right now.

about a suite that never executed. The shipped CLIs were fine. That's a false accusation pointed at the wrong owner, and being scheduled, it repeats every six hours.

The job already fetches each arm's step list, so telling the two apart costs **no extra API call**: if the step named `Backend smoke (...)` failed, the suite ran and the existing wording stands; if the arm died before it, the issue reports an environment fault and names the step that actually broke.

Those thread separately, on the principle this file already states — *"'vendor drift incoming' and 'contributors broken today' are different problems with different urgency, and must never collapse into one issue thread."* "The shipped CLIs are broken" and "the runner could not start the suite" are two more.

What it produces on the real run's data:

> **[Backend smoke: pinned — environment] codex, claude failure on d7571a1**
> The smoke DID NOT RUN: the job failed at **Pull the contributor image**, before the suite started. This says nothing about the CLIs hive ships or about the vendor releases — it is a runner/environment fault, and the fix belongs to whoever owns the runner, not to the backend integration.
>
> There is no suite evidence to read — the run stopped before the suite produced any. The run log shows the failing step.

## Testing

That script had **no coverage at all**, and it is the least-tested shape of code in the repo: JavaScript inside a YAML block scalar, executed only on a scheduled failure, in production, where its output is an issue somebody acts on.

`bin/test_backend_smoke_attribution.sh` **extracts it from the workflow** — not a copy, so it cannot drift from what CI actually runs; if the block moves, extraction fails loudly rather than testing a stale duplicate — and executes it against four fixtures: the real failing run, a genuine suite failure, a run containing both, and a re-failure with the thread already open.

Verified in both directions:

- **Against the pre-fix workflow it fails** on exactly the attribution assertions, and reproduces this issue's title verbatim — `[Backend smoke: pinned] codex, claude failure on d7571a1`.
- **The real-suite-failure fixture passes before *and* after.** That's the point: the fix must not weaken the signal the lane exists to raise. A change that made every failure read as "environment" would pass the first check and fail this one.

It's wired into `v2-ci.yml` next to its siblings, so `bin/test_bin_suites_wired.sh` (#4363) counts it — confirmed locally: `PASS: test_backend_smoke_attribution.sh`, 25 suites checked, 0 orphaned. A guard nothing runs reports green forever, which is the failure mode this whole change is about.

Both edited workflows parse as YAML, and the embedded script is syntax-checked as part of the suite.

**What I could not verify:** that `ubuntu-latest` still has podman *today*. The evidence is that this exact lane passed there 9 hours before the move; if the hosted image has since dropped it, the new preflight will say so in one clear line instead of exit 127.


clubanderson (approved):
Reviewed the backend-smoke runner split and attribution path; pinned runs require a container engine and environment failures no longer blame the CLIs. Harness tests pass.

PR #5890, opened 2026-09-03
🐛 fix(advisory): cap the Recently Resolved changelog with max_findings

Danathar:
## Summary

The advisory digest posted to #2364 has a size dial — `governor.advisory.max_findings`, default 10 — but it only governs the **open findings**. The **Recently Resolved** list underneath had its own separate ceiling of 100, hardcoded and unrelated to the dial. So the digest a repo owner opens is mostly a list of work that is already finished, and the part they can act on is the short bit at the top.

Here is today's digest on #2364, measured:

| Section | Characters | Share |
|---|---|---|
| Open findings (10 rendered, *"286 more exist"*) | 4,937 | 18% |
| Recently Resolved (100 entries) | 22,138 | **82%** |

This PR makes the resolved changelog answer to the same dial. Set `max_findings: 10` and you get 10 open findings and 10 resolved ones, with a line saying how many resolved entries were held back. Set `show_all: true` and the cap comes off both halves. Nothing is silently dropped — that is the contract the finding cap already keeps, and the changelog now keeps it too.

## What changed

- `DigestOptions.resolvedRenderCap()` — the resolved list is bounded by `effectiveCap()`, the same value that bounds the findings.
- `show_all: true` lifts it, and the 100-entry ceiling still applies above that. An owner asking to see every *finding* is not asking for an unbounded changelog, and the comment has to stay inside GitHub's 65,536-character limit: past it, `truncateDigest` cuts from the **bottom**, which is exactly where this section and the analyzed-commit footer live.
- `Digest.ResolvedOverflowCount` carries what was held back, and `writeRecentlyResolved` renders it as the section's last line:

  > - *…plus 20 more resolved in the last 48h, collapsed so the open findings above stay readable*

  This also fixes a silent trim that existed before: over 100 resolved beads, the surplus was dropped with no indication at all. The window in that line is read from `recentlyResolvedWindow` rather than written out, so the two cannot drift.
- `maxRecentlyResolved` moves from a function-local `const` to a package one, documented with the reason it exists.

Open findings, their ranking, the top-N cap, path/provenance staleness marking and the per-(agent, finding-type) collapse are all untouched.

## Why this shape

The alternative was a second config key for the resolved section. One dial is better here: an operator setting `max_findings: 10` is saying how long they want the comment to be, and there is no length at which a hundred healed findings serves that reader better than the open ones do. A separate key would mean an owner who shortened their digest could still be handed 22KB of finished work.

## Tests

`src/pkg/advisory/resolved_cap_test.go`, five cases: the cap applies and the remainder is announced; a changelog shorter than the cap renders whole and claims no remainder; `show_all` lifts the cap while the ceiling holds and announces its own overflow (this case also covers the all-clear rendering path, which shares `writeRecentlyResolved`); the dial itself across default / unset / `show_all` / above-the-ceiling; and a before-and-after measurement of the rendered comment.

Verified they fail against the old behaviour — reverting `resolvedRenderCap` to the fixed 100 reproduces the bug and three of the five go red, including `RecentlyResolved = 30 entries, want 10`.

```
ok  github.com/hivecommons/hive/pkg/advisory   0.27s
ok  github.com/hivecommons/hive/cmd/hive       9.71s
ok  github.com/hivecommons/hive/pkg/github    44.63s
```

`go build ./...` is clean and `gofmt` reports nothing. `pkg/config`'s `TestGateFailsClosedWithoutIP6Tables` fails in my sandbox for want of root (`ip6tables: Permission denied`); it fails identically on a clean `upstream/v4`, so it is unrelated to this change.

Refs #2364.


clubanderson (approved):
Reviewed the advisory cap change; resolved overflow is counted and rendered explicitly. Targeted advisory tests pass.

PR #5902, opened 2026-09-03
🐛 fix(hub): re-derive DashboardURL from the cluster's current domain

clubanderson:
## Problem

**21 of 95 hives** (17 OKE, 4 LKE) advertise a dashboard URL whose hostname no ingress serves, so clicking them from the hub goes nowhere — 404 on LKE, a login bounce on OKE.

`DashboardURL` is minted once at provision time as `"<hiveID>." + cluster.Domain` (`saas_provision.go:2174`) and **nothing ever re-derived it**. When a cluster's domain changed — the `kubestellar.io` → `hivecommons.dev` migration — every hive provisioned before the change kept the old hostname.

## Why this is not cosmetic

The hub **pushes** `DashboardURL` to the spoke on each heartbeat, and the spoke adopts it:

```
adopting vanity dashboard URL from hub heartbeat
  was: hosted-projectbluefin-common-nmq5.hive.hivecommons.dev
  now: hosted-projectbluefin-common-nmq5.hive.kubestellar.io
```

So the stale hub value also **overwrites the spoke's own correct local config**. Correcting a spoke's PVC cannot hold — it reverts on the next beat. The hub is the only durable place to fix this.

Two shapes of breakage found in production:
- wrong domain, right host: `hosted-daviddiaz0317-visual-hive--1jos.hive.kubestellar.io`
- a host that resolves nowhere at all: bluefin advertised `...common-nmq5...` (curl `000`) while its ingress serves `...knuckle-gjvq...` (302)

## Fix

`VanityURL` already self-heals via `reconcileStaleVanityURL`, which adopts the live route when the stored host drifts. `DashboardURL` had no equivalent, so it drifted permanently. This adds one.

Re-derivation happens on the same heartbeat path that already overlays a claimed vanity URL, and is deliberately narrow — it fires **only** when the stored host is exactly `<hiveID>.<something-else>`, the shape provisioning mints:

- a host whose leading label is **not** the hive ID is a vanity host, left to `reconcileStaleVanityURL`; rewriting it here would fight that reconciler and churn the user-visible name every beat
- empty, unparsable, and already-correct URLs are untouched

The cluster registry (`/data/saas/clusters.json`) is **already correct** on all four clusters, so this reconciles existing hives *to* it rather than changing any source of truth. No migration or manual edit needed — affected hives self-heal on their next heartbeat.

## Tests

`dashboard_url_domain_reconcile_test.go` pins both real-world cases and, more importantly, the ones it must not touch: already-correct URLs, vanity hosts, path/port preservation, and every missing-input guard.


clubanderson:
**Converted to draft — do not merge. Verifying against the live fleet showed this would break working URLs.**

## What I got wrong

I claimed this fixes 21 stale URLs. Simulating the exact function against all 53 OKE/LKE hives in the hub showed it heals **1**, and my follow-up commit for the other 20 was based on a misreading.

I assumed hives named `hosted-available-oke-01-placeholder-bb95` pointing at `hosted-tradingasbuddies-falcon-core-k2zn...` were recycled pool slots leaking a previous tenant's hostname. The hub record says otherwise:

```json
{ "id": "hosted-available-oke-01-placeholder-bb95",
  "owner": "davdunc", "org": "TradingAsBuddies",
  "repos": ["falcon-core", "falcon-stats", ...] }
```

It is a **legitimately claimed hive**. `available-...-placeholder-` is just its permanent ID from when it was minted out of the pool — the ID never changes after claiming. So `hosted-tradingasbuddies-falcon-core-k2zn` is its **correct vanity host**, not a foreign tenant's, and `isUnassignedPlaceholder` rightly returns false for it. There is no tenant-data leak.

## Why merging would cause an outage

```
hosted-tradingasbuddies-falcon-core-k2zn.hive.kubestellar.io  -> 302  (works)
hosted-tradingasbuddies-falcon-core-k2zn.hive.hivecommons.dev -> 000  (nothing)
```

The vanity host on the OLD domain is **the only working link** — it is served by a hub-side vanity ingress, not by the ingress in the hive's namespace (which serves only `hosted-available-oke-01-placeholder-bb95.hive.hivecommons.dev`). Re-deriving the domain would point users at a host that answers nothing.

This is exactly the hazard the existing comment in `repairVanityURLForHive` warns about: *"Adopting it before then would replace an ugly-but-working placeholder link with a 503."* I reintroduced the bug that comment exists to prevent.

## What is actually true

- The **cluster registry is already correct** — all four domains are current. New provisioning is fine.
- **`kubestellar.io` in a dashboard URL is not automatically wrong.** The old domain still serves vanity hosts, so "stale-looking" and "broken" are different sets, and I conflated them.
- The one hive that genuinely needs the domain re-derived is `hosted-daviddiaz0317-visual-hive--1jos` (host already matches its hive ID, only the domain drifted).
- Separately confirmed broken: bluefin advertises `hosted-projectbluefin-common-nmq5...` which resolves nowhere (`000`), while its ingress serves `...knuckle-gjvq...` (302). That is a real defect but needs a different fix — the stored vanity host is dead, so the repair is to re-mint or drop it, not to re-domain it.

## Next

Reworking to key off **reachability of the stored host** rather than its shape: only re-derive when the stored host does not answer AND the derived one does. That is the invariant that actually matters and it cannot break a working link. Will re-verify against all 53 OKE/LKE URLs before marking ready.

clubanderson:
Closing — the premise was wrong and the fleet has moved past the state this targeted.

**What this PR assumed:** that 21 hives had a stale `DashboardURL` because nothing re-derived it from `cluster.Domain` after the domain migration.

**What was actually true:** simulating the function against all 53 OKE/LKE hives showed it healed exactly **1**. The other 20 were hives like `hosted-available-oke-01-placeholder-bb95` pointing at `hosted-tradingasbuddies-falcon-core-k2zn...`, which I read as recycled slots leaking a previous tenant's host. The hub record says otherwise — `owner: davdunc`, `org: TradingAsBuddies`, real repos. It is a legitimately claimed hive whose ID is simply the pool name it was minted under, and that hostname is its correct **vanity URL**. There is no tenant-data leak, and `isUnassignedPlaceholder` rightly returns false for it.

**Why merging would have caused an outage:** at the time, the vanity host on the old domain was the only working link. Re-deriving the domain would have pointed users at a host answering nothing — reintroducing exactly what the comment in `repairVanityURLForHive` warns about ("replace an ugly-but-working placeholder link with a 503").

**The actual root cause of the outage** turned out to be unrelated to URL derivation: Let's Encrypt's 50-certs-per-registered-domain-per-week limit. The migration re-minted vanity hostnames fleet-wide, exhausting the quota, so those hosts serve the ingress controller's default certificate (`CN=Kubernetes Ingress Controller Fake Certificate`, `SAN: DNS:ingress.local`). The 429 carries its own `retry after` timestamp. Spokes stayed healthy throughout and serve correctly on their namespace hostnames.

Also worth recording for the next person: `kickVanityURLRepairAsync` runs on every heartbeat, so clearing `vanity_url` on the hub, editing the spoke PVC, and restarting the hub are all reverted within ~2 minutes. That reconciler is correct — it just cannot know a cert is unissuable.

The genuine gap this PR noticed — that `VanityURL` self-heals via `reconcileStaleVanityURL` while `DashboardURL` has no equivalent — may still be worth addressing, but keyed off **reachability of the stored host** rather than its shape, so it can never replace a working link with a dead one. Filing that separately if it proves out.

Issue #5236, opened 2026-08-30
🐛 bug(advisory): no-provenance findings bypass stale-bead aging after #5148


Danathar:
## Summary

PR #5148 fixed the dishonest freshness stamp on advisory findings that carry usable provenance, and it safely stops identical explicit-provenance reports from refreshing their staleness clock. The deliberately conservative fallback remains unchanged: a finding with no `provenance_sha` still refreshes `LastSeenAt` every time an agent repeats it.

That uncovered case is now reproducing against `Danathar/atomic-image-builder`. A shell-coverage finding with no usable provenance has survived the implementation that disproves it, is still marked “reported 3×,” and generated a new downstream scanner issue. The digest now warns that the finding may be outdated, which is more honest than before #5148, but the false finding still remains live indefinitely because every cached re-report counts as fresh confirmation.

Follow-up to #5130 and #5148. Downstream evidence: [Danathar/atomic-image-builder#116](https://github.com/Danathar/atomic-image-builder/issues/116).

## Concrete evidence

Atomic Image Builder `main@120218c9f818` has measurable shell coverage:

- `.github/workflows/ci.yml` installs Bashcov 4.0.0 and runs both `tests/test_contrib_aib.sh` and `tests/test_entrypoint.sh` under it.
- `.simplecov` limits the report to `contrib/aib` and `container/entrypoint.sh`.
- [CI run 33338916698](https://github.com/Danathar/atomic-image-builder/actions/runs/33338916698) passed at that exact SHA and uploaded a non-empty `coverage-shell` artifact.

The advisory digest was nevertheless updated at 2026-08-30 19:47 EDT with this live MEDIUM finding:

> No coverage measurement for contrib/aib and container/entrypoint.sh shell scripts … Neither has any statement/branch coverage tool wired (no kcov/bashcov)

It is still marked `reported 3×`. The same digest also includes downstream issue #116 saying that the finding is stale, so the digest simultaneously presents the false claim and the evidence refuting it.

Digest comment: https://github.com/Danathar/atomic-image-builder/issues/11#issuecomment-5395760874

## Why #5148 does not retire this finding

#5148 explicitly preserves the old behavior when a finding records no provenance: every same-title re-report refreshes the bead because silently aging out a potentially live finding would be unsafe. Only an explicit `provenance_sha` gates the `LastSeenAt` refresh; prose inference is used for captions but not retirement.

[...]

Issue #5130, opened 2026-08-29
🐛 bug: advisory digest republishes findings from older commits under a newer Analyzed at stamp


Danathar:
## What happened?

The advisory digest publishes findings that were computed at an older commit, stamped with a newer `Analyzed at` commit, without re-checking whether they still reproduce. Two findings on [Danathar/atomic-image-builder#11](https://github.com/Danathar/atomic-image-builder/issues/11) have now survived five consecutive regeneration cycles this way, and one of them has been fixed in the target repo for 18 hours while the digest keeps reporting it.

The digest footer says `Analyzed at Danathar/atomic-image-builder@29cb70657c25`, which is current `main`. Two of its HIGH findings carry their own, older provenance:

1. `contrib/aib has no test coverage: no unit tests, no e2e execution, no lint`, provenance `revision c9546a8a24b3dded3146e3ab7a93dd99edc56fa3`.
2. `Partial branches in atomic_image_builder.py not covered by unit or e2e suite`, claiming 36 remaining partial branches, provenance `CI run 33187518367, commit 9a6313c`.

Neither reproduces at `29cb706`, the commit the digest itself names.

The impact is not just noise. The `contrib/aib` finding was correct when computed, was filed as [#88](https://github.com/Danathar/atomic-image-builder/issues/88), and was fixed by PRs [#89](https://github.com/Danathar/atomic-image-builder/pull/89) and [#90](https://github.com/Danathar/atomic-image-builder/pull/90). Because it kept reappearing after the fix, a scanner agent re-checked it against current `main`, concluded the evidence was "provably false", and filed [#91](https://github.com/Danathar/atomic-image-builder/issues/91) calling it a fabricated evidence claim. It was not fabricated, it was stale. So this bug also produces false accusations of bad evidence from other agents, and burns a maintainer's time adjudicating them.

Timeline, all Aug 28 EDT:

| time | event |
|---|---|
| 17:10 | `c9546a8` lands, the finding's stamped provenance. Evidence accurate here. |
| 18:11 | #88 filed against the target repo. |
| 19:14 | #89 merges `tests/test_contrib_aib.sh`. |
| 21:37 | #90 merges the `shellcheck` and `Test contrib/aib` CI steps, producing `29cb706`. |
| 21:43 | Digest publishes the finding. `Analyzed at 29cb706`, provenance `c9546a8`. |
| 21:45 | Scanner files #91, reporting the finding as factually false. |

[...]

PR #5945, opened 2026-09-04
🐛 fix(advisory): rank on all unverified-evidence signals, not just missing paths

Danathar:
## Summary

The advisory digest shows a repo owner a small number of findings — ten, on hivecommons/hive — chosen from hundreds. It also warns the reader, on individual findings, when it has no reason to believe the finding is still true. This PR stops those two things from contradicting each other.

Today the digest can pick a finding for one of its ten slots and *then* caption it "not re-verified", while a finding that genuinely was checked at the analyzed commit goes unshown. On the 2026-09-04 report in #2364, three of the ten rendered findings carried such a caption — including two of the four CRITICAL slots — while 292 findings went unrendered.

After this change, a finding nothing re-checked is ranked below an equally-severe finding that was. It is not dropped: it still fills any slot no confirmed finding claims, and it still carries its caption when it does.

## The defect

`FormatDigestMarkdown` captions a finding as not-current for three distinct reasons:

| Signal | Caption | Filed as |
|---|---|---|
| `PathStale` | *file path not found at analyzed commit — finding may be outdated* | #3704 |
| `ProvenanceStale` | *evidence computed at `abc1234`, not re-verified at the analyzed commit* | #5130 |
| `CachedReplays` with no provenance | *re-reported 4× from cached evidence, not re-verified* | #5236 |

All three say the same thing to a reader: nothing has confirmed this at the commit the digest is stamped with. But only the first one influenced `applyTopN`, the ranking that decides which findings survive `governor.advisory.max_findings`.

The other two could not have influenced it, because of ordering: `MarkStaleProvenance` ran on the *finished* digest, after `applyTopN` had already chosen. `CachedReplays` was read from bead metadata before the cap but never consulted by the ranking.

The result is a top-N spent on findings the digest itself tells you not to act on, while findings it would have vouched for are counted into "292 more exist".

## The fix

**One predicate, mirroring the renderer.** `Finding.evidenceUnverified()` returns true for exactly the three conditions above — same fields, same order, same file — so the ranking and the caption cannot drift apart.

**Provenance is resolved before the cap.** `markStaleProvenanceIn` is extracted from `MarkStaleProvenance` so `BuildDigestFromBeads` can mark the full set before ranking. This is free: it reads the finding's own metadata and prose and performs no lookups. The path check, which *does* cost lookups, stays exactly as it was — on-demand, cached per distinct path, stopping the moment the cap is filled.

**`applyTopN` demotes on the union.** The severity-band walk now sets aside any unverified finding rather than only a path-stale one.

## What deliberately did not change

- **It is a demotion, not a filter.** Unverified means nobody re-checked the finding, not that it is fixed. A slot no confirmed finding of equal severity claims still backfills with the newest unverified one — a digest of six under a cap of ten would hide work for no reason.
- **Marks survive the backfill.** A demoted finding that gets a slot arrives with `ProvenanceStale` / `CachedReplays` intact, so the renderer still captions it. Publishing one as confirmed would be the #5130 overclaim.
- **Severity still dominates.** An unverified critical outranks a confirmed low. Freshness breaks ties *within* a band only; letting it cross bands would let a cosmetic nit displace a security finding whose file was merely renamed.
- **A replay count on a finding that carries a provenance commit does not demote.** That is not the #5236 signal, and the renderer does not caption it as one.
- **No snapshot, no judgement.** A caller that pins no analyzed commit gets exactly the previous severity-then-recency ranking; nothing can be called stale against a commit the digest does not name.

## Tests

`src/pkg/advisory/topn_unverified_test.go` (new) covers ranking for provenance staleness and for cached replay, the backfill, mark preservation through the backfill, the severity-first limit, the no-verifier path, the non-demotion of replay-with-provenance, and the end-to-end path through `BuildDigestFromBeads`.

Six of them fail against the code before this change. The two end-to-end cases seed their unverified findings **last**, so they are the most recent and severity+recency alone would have ranked them first — without that, they would pass for the wrong reason.

```
$ go test ./pkg/advisory/ -count=3 -shuffle=on
ok      github.com/hivecommons/hive/pkg/advisory        1.247s

$ go test ./pkg/advisory/ -cover
ok      github.com/hivecommons/hive/pkg/advisory  coverage: 95.2% of statements
```

`go build ./...` and `go vet ./...` are clean across the module.

Refs #2364.


clubanderson (approved):
Reviewed the diff and local validation scope for this task; no blocking issues found.

Issue #5925, opened 2026-09-04
🌱 dibs should be canonically dibs.hivecommons.dev, with dibs.kubestellar.io as a redirect


clubanderson:
## Goal

`dibs` should be canonically `dibs.hivecommons.dev`, with `dibs.kubestellar.io` demoted to a redirect - matching the hub, where `hive.hivecommons.dev` became the real host on 2026-09-04 and the kubestellar.io name became the legacy entry point.

## Current state (measured 2026-09-04)

- Single ingress `dibs/dibs` on cluster `hive-oke`, host `dibs.kubestellar.io`, TLS secret `hive-tls-hc` (its own per-namespace cert, `READY=True`).
- `dibs.kubestellar.io` serves HTTP 200 today.
- `dibs.hivecommons.dev` **does not resolve** - no DNS record exists.
- It is one of only two hosts on hive-oke not covered by the fleet wildcard (the other is `hive.kubestellar.io`).

## Ordering matters

The existing wildcard `hive-wildcard-tls` covers `*.hive.hivecommons.dev`, `*.lke.hive.hivecommons.dev` and `hive.hivecommons.dev`. It does **NOT** cover `dibs.hivecommons.dev`, which is one level up, not under `hive.`.

Given the Let's Encrypt 50-cert/week cap on `hivecommons.dev` (see the parent incident), prefer **adding `dibs.hivecommons.dev` as a SAN to the existing wildcard Certificate** over minting a separate cert. A SAN addition is one re-issue; a new Certificate is a new issuance against the same quota either way, but consolidating keeps the object count down and avoids another per-host cert to maintain.

Note the quota is currently exhausted and refills as certs age out of the rolling 168h window - most of ours were minted in a burst on 2026-09-03 afternoon, so meaningful headroom does not return until ~2026-09-10. **Do not attempt the re-issue before then**; a 429 pushes the retry deadline further out.

## Suggested sequence

1. Add DNS `dibs.hivecommons.dev` -> `157.151.252.29` (A record) in Cloudflare, DNS only / grey cloud.
2. After the LE window has headroom (~2026-09-10), add `dibs.hivecommons.dev` to `hive-wildcard-tls` dnsNames and let it re-issue.
3. Add `dibs.hivecommons.dev` as a host on the `dibs/dibs` ingress (keep `dibs.kubestellar.io` alongside it during cutover).
4. Verify `https://dibs.hivecommons.dev` serves 200 with a valid LE cert.
[...]

PR #5946, opened 2026-09-04
📝 docs(hub): pin the sibling-SSO invariant behind the dibs hostname move

Danathar:
## Summary

#5925 reads like branding: give `dibs` a `hivecommons.dev` name, redirect the old one. It isn't. Moving `dibs` is the **repair for an identity bridge that the hub's own move already broke**, and the reason is nowhere in the tree — the code comments still assert the opposite as settled fact.

`dibs` has no login of its own. It resolves a visitor by forwarding the browser's `hive_hub_user` cookie to the hub's `GET /api/saas/whoami` server-to-server (#4171). That works only while the browser sends that cookie to `dibs` at all, and it does so **solely** because of the cookie's `Domain` attribute — which `sessionCookieDomain` derives from the registrable domain of the hub's *own* canonical host. So the bridge has a precondition:

> The hub and the sibling product must share a registrable domain.

When the hub became `hive.hivecommons.dev` on 2026-09-04, the session cookie became `Domain=.hivecommons.dev`, and `dibs.kubestellar.io` stopped receiving it. The symptom is not an error — `dibs` renders every visitor as signed out. Moving `dibs` under `hivecommons.dev` is what restores the precondition.

**The cluster-side steps in the issue (DNS, the SAN addition, the ingress host, the redirect) are ops and are not in this repo.** This PR ships the part that is: the invariant those steps depend on, pinned in the test suite and written down where the next person moving a hostname will find it.

## The finding that changes the runbook

Step 5 of the issue — demote `dibs.kubestellar.io` to a redirect — reads as tidy-up, and the tempting shortcut once the new host works is to leave the old host serving the same application instead. **That cannot be made to work, and no configuration rescues it.**

A browser ignores a `Set-Cookie` whose `Domain` does not cover the host that sent it (RFC 6265 §5.3). A hub on `hivecommons.dev` therefore can never scope a session cookie to `kubestellar.io`; there is no second `Domain` to configure, which is why `sessionCookieDomain` falls back to a host-only cookie for any foreign host rather than trying. `HIVE_HUB_LEGACY_COOKIE_DOMAIN` does not help either — it *expires* the stale cookie on the old domain, it never writes a session there.

A redirect works precisely **because** it is not dual-serving: the browser ends up issuing its request to the new host, which is inside the cookie's scope. The redirect is the mechanism, not a preference.

## What's in the PR

**`src/pkg/hub/sibling_host_migration_test.go`** (new) pins the invariant:

- the cookie reaches `dibs.hivecommons.dev` but **not** `dibs.kubestellar.io` when the hub is on `hivecommons.dev` — and the mirror case, because the invariant is about *sharing* a registrable domain, not about which one;
- scoping to a foreign registrable domain collapses to a host-only cookie, so the redirect is required;
- the spokes keep receiving the session cookie under both configurations — whatever happens on the sibling side must not cost ~62 hosted tenants their dashboard and terminal (the `AUDIT F4` note in `oauth.go`).

Assertions go through an RFC 6265 §5.1.3 domain-match helper rather than comparing the raw `Domain` string, so each case reads as *"does the sibling see the session"* — which is the question that actually matters.

**Comment corrections.** `sessionCookieParentDomain`, `sessionCookieDomain` and `setSessionCookies` stated the `.kubestellar.io` arrangement as fixed fact, including *"dibs.kubestellar.io and the hub are same-SITE (same registrable domain)"*. That sentence is now conditional, and it is exactly what a future reader would have trusted while re-introducing the break. They now state the invariant instead of the hostnames.

**Docs.** `src/docs/hivecommons-migration.md` gains a P6 phase row and a **Moving a public hostname** runbook — the SSO coupling, the shared Let's Encrypt issuance cap, the wildcard's real coverage (`*.hive.hivecommons.dev` does **not** cover a name one level up such as `dibs.hivecommons.dev`), and the ordered sequence. `src/docs/env-vars.md` warns on `HIVE_HUB_PUBLIC_URL` that changing its registrable domain signs users out of siblings left on the old one.

## Verification

```
$ go test ./pkg/hub/ -count=1
ok      github.com/hivecommons/hive/pkg/hub     99.768s

checked 131 files under src/docs
```

`go build ./...` and `go vet ./pkg/hub/` are clean.

The new tests are characterization tests, so I mutation-checked them rather than claiming a red-to-green: forcing `sessionCookieDomain` to ignore the request host fails `TestNoCookieScopeKeepsSiblingOnForeignRegistrableDomain` with the intended message, so they are not vacuous.

## What this does not do

I have no cluster or DNS access, so steps 1–5 of the issue are untouched and #5925 should stay open for them. Note the issue's own warning that the Let's Encrypt window has no headroom until ~2026-09-10 — step 2 should not be attempted before then.

I was also unable to probe the live hosts from this environment (outbound fetch is blocked here), so the current-state claims above are the issue's own measurements of 2026-09-04, not my own.


clubanderson (approved):
Reviewed the diff and local validation scope for this task; no blocking issues found.

Issue #5958, opened 2026-09-04
🐛 Surface agent start failure reasons in state and back off relaunch loops


clubanderson:
Follow-up from #5921 / PR #5944.

PR #5944 closes the configuration-contradiction half by rejecting clear backend/launch_cmd mismatches at load and Settings-save time. Remaining runtime work:

- Prompt detection should follow the launched binary, or diagnostics should explicitly report backend/launch_cmd mismatch.
- Agent state, dashboard cards, and fleet verdicts should carry concrete failed-start reasons such as `copilot: not logged in`, `bob: API key rejected (401)`, or `backend/launch_cmd mismatch` instead of generic `restart needed`.
- After N consecutive start failures with the same reason, mark the agent blocked and back off relaunches instead of recreating sessions every few minutes.
- Fleet `able` counts should exclude agents blocked for failed-start reasons.

Closes the remaining runtime-state/fleet-verdict portion intentionally left out of #5944.

Danathar:
Opened as **#5960** (based on `v4`, independent of #5944).

All four bullets are covered; a note on each, plus one place where I deliberately did less than the issue text implies.

### 1. Prompt detection follows the launched binary

Both halves. The Copilot hang check keyed on `agent.Config.Backend`, so an agent with a **backend override** was already being judged by the wrong CLI's readiness signature — the same bug class as root cause 1, hiding in plain sight. It now uses `effectiveBackend`.

And when `launch_cmd` declares a different known CLI than `backend`, the diagnostic reports the contradiction by name rather than running. That path could never converge: bare `copilot` starts fine, the agent restarts back into bob, and hangs again. That *is* the 4,025-line loop.

Detection only makes **positive** identifications — an empty command, an unrecognised wrapper, or an inference backend legitimately exec'ing `claude` all return "no opinion". Guessing wrong would manufacture a block on a working agent, which is worse than the missing diagnosis.

### 2. Concrete reasons in state, card, and verdict

`copilot: not logged in`, `bob: API key rejected (401)`, `backend/launch_cmd mismatch (configured copilot, launches bob)`, plus `no API key configured`, `CLI binary not found`, and an honest residual `no CLI prompt after launch` for "we know it did not start and the diagnostic could not say why".

[...]

Issue #5921, opened 2026-09-04
🐛 Agent with backend: copilot but launch_cmd: bob is launched as Bob, judged as Copilot, and looped as "hung" forever


clubanderson:
Observed on hive `hosted-available-vllmd-01` (katamari/ibm-aiops-orchestrator, spoke `d7571a1`, L2, BUSY) on 2026-09-03. Fleet row: `1 agent(s) down — restart needed · judged on advisory · 13 queued`, supervisor `PROBLEM: dead`, hint "Restart the agent from its card". Restarting does not help.

## Root cause 1 — contradictory agent config is accepted silently
`hive.yaml.runtime`:
```yaml
supervisor:
  backend: copilot
  launch_cmd: bob --model auto
```
The hive launches **Bob** in the tmux session, then waits for a **Copilot CLI** prompt. It never appears, so after `uptime_sec: 180`:
```
WARN copilot hung with no CLI prompt, running diagnostic   agent=supervisor
INFO diagnostic: launched bare copilot to capture error
INFO diagnostic: copilot started successfully in bare mode
WARN failed to send kick  error="agent supervisor cannot be kicked: it failed to start: copilot hung with no output (diagnostic timed out)"
```
Every governor eval repeats this. The fleet chip shows `bob` for the agent while the logs and diagnostic path treat it as `copilot`.

## Root cause 2 — the same loop is silently churning the "healthy" agents
guide / quality / scanner (backend: copilot, launch_cmd: "") also hit `copilot hung with no CLI prompt` on every kick; sessions are recreated every ~3 min (tmux session create times 23:00:13, 23:01:16, 23:02:18). `grep -c "copilot hung with no CLI prompt" /data/logs/hive.log` → **4025**. The fleet still shows them `● on · should be working`.

Underlying causes on this spoke are user-side (Copilot not logged in: pane shows `Please use /login to sign in to use Copilot`; Bob key 401: `API Key verification failed: Invalid or expired API Key`), but the hive hides both behind "hung" / "restart needed".

## Expected
1. **Validate at config load/save:** `backend` and `launch_cmd` must agree (reject, or derive backend from the launch binary and warn). Settings UI should refuse to save the contradiction.
2. **Prompt detection must follow the launched binary**, not the declared backend, or at minimum the diagnostic should report the mismatch explicitly.
[...]

PR #5960, opened 2026-09-04
🐛 fix(agent): surface start-failure reasons and back off relaunch loops

Danathar:
## Summary

Incident #5921: a hive relaunched agents that could not possibly start — every ~3 minutes, **4,025 times** — while the fleet page showed them green and counted them as `able`. The causes were ordinary and user-fixable: Copilot was not logged in, a bob API key was rejected, and one agent was configured `backend: copilot` but launched by `launch_cmd: bob`, so the readiness probe waited for a prompt the launched binary never prints.

None of them reached a human. The card, the fleet row and the log all said `restart needed` or `hung` — and the only control on offer was the restart button, which could not fix any of the three.

Two things were missing. This PR adds both.

## A reason

`src/pkg/agent/start_failure.go` classifies a failed start into a stable class and renders the sentence the issue asked for:

| Class | Reason |
|---|---|
| `login-required` | `copilot: not logged in` |
| `credential-rejected` | `bob: API key rejected (401)` |
| `credential-missing` | `bob: no API key configured` |
| `backend-mismatch` | `backend/launch_cmd mismatch (configured copilot, launches bob)` |
| `binary-missing` | `copilot: CLI binary not found` |
| `no-output` | `copilot: no CLI prompt after launch` |

The **class**, not the prose, is what consecutive-failure counting compares — so rewording a message can never reset the ladder, and two genuinely different faults can never be counted as one recurrence. `backend-mismatch` is deliberately not backend-scoped: naming either side would assert the very confusion the message exists to report.

The record rides `AgentProcess` out through `snapshot()` to the dashboard card (`BLOCKED` + evidence, beside the existing inference-error block), to the heartbeat as `AgentSummary.StartBlockedReason`, and into the hub verdict.

## A limit

Consecutive failures of the same class escalate a backoff — **1m → 5m → 15m → 30m** — and after three mark the agent blocked. Both relaunch drivers then stand down: `CheckAndRestartCrashedAgents` skips it, and the Copilot hang diagnostic stops recreating the session. Hub-side, a start-blocked agent is excluded from `able`, carries the spoke's own reason on the row, and is **red rather than amber** — amber means "half its job works", and an agent that never started does no half of it.

Three, not one, because a single failed start is routinely transient (a cold pod losing a race with its own config write) and blocking on it would turn recoverable noise into an agent a human has to un-wedge. Backoff applies from the *first* failure, so the two retries before the block are paced rather than a free burst.

The 30-minute ceiling is equally deliberate: these conditions get fixed **out of band**, with nothing to tell the hive that a human just logged in or pasted a key, so a blocked agent has to keep re-probing. And an explicit retry — the dashboard restart button, or the relaunch that follows saving a bob key — clears the record outright. The ladder governs the automatic loop, never a human's deliberate retry.

## Prompt detection follows the launched binary

The hang check keyed on `agent.Config.Backend`, so an agent with a backend override was already being judged by the wrong CLI's readiness signature — the same bug class as #5921's root cause 1. It now uses `effectiveBackend`.

And when `launch_cmd` declares a *different* known CLI than `backend`, it reports the contradiction by name instead of running a Copilot diagnostic that cannot converge: bare copilot would start fine, the agent would restart back into bob, and hang again — forever. That is the observed 4,025-line loop.

`launchCmdBackend` only makes **positive** identifications. An empty command, an unrecognised wrapper, or an inference backend legitimately exec'ing `claude` all return "no opinion" — guessing wrong here would manufacture a block on a working agent, which is strictly worse than the missing diagnosis.

## One deliberate limit, pinned by a test

An **off-schedule** start-blocked agent is not an alarm. `runStuckAtLogin` bypasses the `expectedActive` gate because a wedged interactive credential is hive-wide — every kick to that backend fails. One agent failing to start is not, and raising a `Problem` for an agent the governor is not scheduling would break the quiet-by-design rule the rest of the fleet view keeps. It is still never `able`, and it becomes a `Problem` the moment the mode schedules it. Both halves of that pair are tested.

I originally wrote the bypass the other way; the test I wrote to pin it failed, and it was right to — the bypass was also redundant, since `ExpectedActive: false` already lands in `runQuietByDesign`, which the `!QuietByDesign` conjunct excludes anyway.

## Relationship to #5944

#5944 (open) adds `config.LaunchCmdDeclaredBackend` for the **load/save-time** half of the mismatch question. This branch is based on `v4` and does not depend on it. `launchCmdBackend` here is the runtime reader and carries a comment to delegate to that helper once it lands — the validator and the runtime disagreeing about what a `launch_cmd` launches would be #5921 in a new place.

## Verification

```
$ go test ./pkg/hub/ ./pkg/dashboard/ -count=1
ok      github.com/hivecommons/hive/pkg/hub          98.082s
ok      github.com/hivecommons/hive/pkg/dashboard    48.721s

$ go test ./pkg/agent/ -run 'StartFailure|BackendMismatch|NotRunningReason|...'
ok      github.com/hivecommons/hive/pkg/agent        0.023s
```

20 new cases: pkg/agent covers the reasons, the ladder, same-class-does-not-re-arm, different-class-restarts-the-ladder, clear semantics (including that it does **not** clobber an unrelated `LastError`), mismatch detection, the snapshot, and `notRunningReason`. pkg/hub covers the `able` exclusion, reason precedence over the generic login wording, never-amber, the off-schedule/scheduled pair, the rollup, the legacy-spoke no-op, and a blank-reason guard.

`go build ./...` and `go vet` are clean across the touched packages.

**One caveat, stated plainly.** `pkg/agent` has 40 failing tests in this sandbox, all from tmux `Operation not permitted` on the socket. They are not mine: I captured the failing set on a clean `upstream/v4` tree and after this change, and the set of failing test names is **identical**. But it does mean the tmux-dependent launch paths I touched were not exercised end-to-end here — CI will be the first full run of those.

Refs #5958, #5921.


clubanderson:
Reviewed the runtime-state/fleet path and pushed a follow-up: the start-failure block threshold and backoff ladder are now named and env-configurable, with tests. Targeted validation passed: go test ./pkg/agent -run 'StartFailure|BackendMismatch|NotRunningReason' -count=1; go test ./pkg/hub ./pkg/dashboard -count=1; go vet ./pkg/agent ./pkg/hub ./pkg/dashboard.

PR #5962, opened 2026-09-04
🌱 chore(dibs): preflight the cutover before it spends the Let's Encrypt window

Danathar:
## Summary

#5957 staged the cutover manifests and the runbook. This adds the check that has to happen **before** any of them run, plus fixes to three verification steps that would currently pass while the thing they verify is broken.

The sequence is scheduled around a Let's Encrypt quota window. That makes the expensive failure not *"a step fails"* but *"a step fails **after** the certificate has been re-issued"* — which spends the window the whole plan is waiting for and puts the next attempt a week out. That is what the issue's own warning about a 429 is about.

Every assumption the staged manifests rest on is checkable read-only, today, at zero quota cost.

## The assumption worth checking first

`02-dibs-ingress-dual-host.yaml` gives `dibs.hivecommons.dev` **no `tls:` block**, and relies on the ingress-nginx controller's `--default-ssl-certificate` being `hive-hub/hive-wildcard-tls`. The staged README hedges this as *"ingress-nginx **should** serve it with the controller default SSL certificate"*.

That hedge is load-bearing. If the flag is absent or points elsewhere, the host is served ingress-nginx's **self-signed** certificate — a host that answers `200` while failing validation — and step 5 discovers it with the issuance already spent.

It is also the only Ingress in this tree that does not name a `secretName`; every other one does (`src/pkg/hub/saas_provision.go`). The staged manifest is *right* not to try — an Ingress may only reference a Secret in its own namespace, and the wildcard lives in `hive-hub` — which is exactly why the default-certificate path carries the weight here and deserves confirming in advance rather than on the day.

## `src/deploy/dibs-domain-cutover/preflight.sh`

Read-only. Six checks:

1. **`--default-ssl-certificate`** is the wildcard (above).
2. **The Certificate** exists where the patch targets it, and does not already carry the SAN.
3. **Drift** between the live `dibs/dibs` Ingress and the staged copy. `02-*.yaml` is a **full object**, so `kubectl apply` overwrites the hand-managed live object wholesale and silently drops any annotation it omits (proxy body size, rate limits, auth-url). The README says to compare by hand; this is that comparison.
4. **Host collisions.** ingress-nginx resolves a duplicate host+path claim in favour of the **older** Ingress and only logs a warning. Apply the redirect while the app Ingress still lists the host and it does not fail — it is simply inert. `curl` keeps returning `200`, every object looks applied, and the obvious reading ("the redirect hasn't landed yet") sends you to re-apply rather than to the host it is losing to.
5. **DNS** resolves to the address step 1 names.
6. **Let's Encrypt headroom.** A read-only crt.sh count estimates whether hivecommons.dev still has room in the rolling 168h issuance window. This is not an ACME call and does not apply a Certificate or Ingress.

A check that could not **run** reports as a warning, never as a pass. An unreachable cluster must not read as "preconditions met" — that is the same silent-green failure the script exists to prevent, so it has its own test.

## Runbook fixes

- **New step 0**, framed as *"do this now, not on the day"*, with step 4 cross-referencing it at the point the assumption is used.
- **Step 5 stops accepting a `200`.** Both failure modes it exists to catch answer `200`. `curl -sSI` now fails on an invalid chain, `openssl` names the issuer and SANs, and the acceptance test is stated as what it actually is: *a browser already signed in to the hub must be recognised on the new host* — because a signed-out `dibs` also renders fine and also returns `200`.
- **Step 6 states the removal-before-apply ordering and why**, with the command to confirm sole ownership first.
- **Step 7 exercises a deep path with a query**, so a redirect that drops `$request_uri` cannot pass a `/`-only check while breaking every deep link into `dibs`.
- **A new section names the dual-host window** as a verification window rather than a resting state: the legacy host is knowingly serving a signed-out experience for as long as it is open — which follows from the invariant #5946 pinned — and only step 6 ends it.

## Verification

```
$ bash bin/test_dibs_cutover_preflight.sh
pass=21 fail=0

$ bash bin/test_bin_suites_wired.sh
Checked 26 suite(s), 0 orphaned.

checked 133 files under src/docs

$ shellcheck -S warning src/deploy/dibs-domain-cutover/preflight.sh bin/test_dibs_cutover_preflight.sh
(clean)
```

21 cases against stub `kubectl`, stub `dig`, and stub `curl` — no cluster, no network. Wired into `v2-ci.yml`, which `bin/test_bin_suites_wired.sh` requires.

The suite opens with a **positive control** so it cannot pass by the script always failing, and I mutation-checked the assertions: neutering the exit-78 path fails four of them.

## Scope

Nothing here applies anything to a cluster. The DNS record, the certificate patch, the ingress changes and the redirect all remain manual and date-gated (the LE window has no headroom until ~2026-09-10), so **#5925 stays open** — this is `Refs`, not `Closes`.

I could not probe the live hosts or cluster from this environment, so the checks encode the assumptions the staged manifests and the issue state, and are designed to *report* rather than assume. Whoever runs step 0 will be the first to learn whether they hold.

Refs #5925.


clubanderson:
Reviewed the cutover preflight for side effects and gap coverage. Pushed a follow-up adding a read-only crt.sh Let's Encrypt headroom check (no ACME/Certificate/Ingress apply), plus hermetic tests and docs. PR body already uses Refs #5925, not Closes. Targeted validation passed: bash bin/test_dibs_cutover_preflight.sh; bash bin/test_bin_suites_wired.sh; python3 src/scripts/check-docs-links.py; bash -n on both scripts.

PR #5966, opened 2026-09-04
🌱 refactor(github): drop the pkg/config dependency from the client (phase 1 of #5953)

Danathar:
## Summary

`pkg/github` is the package everything in hive uses to talk to GitHub — ten
packages import it. It also imported `pkg/config`, the application's 6,000-line
configuration file, which meant anything wanting an HTTP client dragged the
entire operator-configuration surface along with it, and inverted the natural
layering (configuration should depend on the client, not the other way round).

This removes that dependency. It is phase 1 of the four-phase split the issue
proposes, and the only phase that changes package boundaries; the policy
engines and watcher daemons stay where they are for now.

**No behaviour change**, and no call site changes: the seven places that
install an issue filter pass exactly what they passed before.

## Detail

The dependency was four references, not a deep entanglement:

| Reference | Use |
| --- | --- |
| `config.IssueFilterConfig` | type of `Client.issueFilter`; one method called, `Admits(labels)` |
| `config.DefaultAutoMergeLabel` | aliased by the `AutoMergeQueuedLabel` constant |
| `config.SelfMergeMinACMMLevel` | logged by the self-authored auto-merge sweep |

Everything else the issue counts is prose in comments naming `config.RoleMerger`
and `config.AutoMergeConfig`; those are references to concepts, not imports.

### The filter becomes a one-method interface

`IssueAdmitter` is declared here and asks the only question the client actually
has: may an issue carrying these labels be worked?

```go
type IssueAdmitter interface {
	Admits(labels []string) bool
}
```

`config.IssueFilterConfig` already satisfies it, so `SetIssueFilter` keeps
taking the same values and **all seven construction sites are untouched**
(`cmd/hive/main.go` ×6, `pkg/dashboard/api.go` ×1).

The matching logic is deliberately *not* copied into this package. It is an
approval gate with exact, case-insensitive semantics, and a second
implementation of it that drifted from the first would fail **open** — the
unsafe direction. The policy stays with the configuration layer that owns it;
the client only consumes the verdict.

`getIssueFilter()` now never returns nil. The zero value of the old struct
admitted everything, and an interface's zero value is nil, so without that
guard "no filter configured" would become a panic on the enumeration path
rather than the documented admit-all.

### The two constants are pinned, not trusted

They are duplicated rather than imported, and `config_parity_test.go` asserts
each equals its `pkg/config` counterpart, so a change on either side breaks the
build instead of quietly changing behaviour. That file is the only remaining
reference to `pkg/config` in this package. A test-only import does not put
config back into the dependency tree of the ten importers, which is what the
issue is about.

### Verification

```
go list -deps ./pkg/github | grep -c hive/pkg/config
  before: 1
  after:  0
```

## Testing

- [x] `cd src && go test ./pkg/github/...` — pass
- [x] `cd src && go test ./pkg/dashboard/... ./cmd/hive/...` — pass
- [x] `cd src && go vet ./pkg/github/...` — clean
- [x] Other / not run (explain): `go test ./pkg/config/...` has one failure,
      `TestGateFailsClosedWithoutIP6Tables`, which reproduces identically on
      unmodified `upstream/v4` in this environment (no `ip6tables`). Confirmed
      by stashing the change and re-running. Not touched by this PR.

New coverage in `src/pkg/github/config_parity_test.go`: both constants pinned
to their config counterparts; `config.IssueFilterConfig` pinned as satisfying
`IssueAdmitter` (so the seven call sites break here, loudly, rather than at the
call); and the unset filter pinned as admitting everything, including the nil
label slice.

## Scope

Phase 1 only. Phases 2–4 from the issue — extracting the automerge policy
engine, the request watchers, and the advisory digest rendering — are untouched
and the issue should stay open for them, hence `Part of` rather than `Fixes`.

## Note for reviewers

No changelog fragment: `changelog.d/README.md` says routine refactors need
none, and this has no operator-visible effect. Happy to add one, or the

Part of #5953

### DIFF
--- a/src/pkg/github/automerge_sweep.go
+++ b/src/pkg/github/automerge_sweep.go
@@ -12,8 +12,13 @@ import (
 	"time"
 
 	gh "github.com/google/go-github/v72/github"
-	"github.com/hivecommons/hive/pkg/config"
 )
+
+// selfMergeMinACMMLevel mirrors config.SelfMergeMinACMMLevel. It is duplicated
+// rather than imported so this package keeps no dependency on pkg/config
+// (hivecommons/hive#5953 phase 1); config_parity_test.go pins the two equal so
+// they cannot drift silently.
+const selfMergeMinACMMLevel = 6
 
 const DefaultAutoMergeSweepMaxMerges = 3
 
@@ -420,7 +425,7 @@ func (c *Client) StartSelfAuthoredAutoMergeSweep(ctx context.Context, maxMerges 
 			level = fmt.Sprintf("%d", *acmmLevel)
 		}
 		c.info("self-authored auto-merge sweep disabled: acmm_level below minimum (or auto_merge.self_authored is off)",
-			"acmm_level", level, "min_acmm_level", config.SelfMergeMinACMMLevel)
+			"acmm_level", level, "min_acmm_level", selfMergeMinACMMLevel)
 		return
 	}
 	interval := selfAuthoredSweepInterval(len(c.getRepos()))
--- a/src/pkg/github/client.go
+++ b/src/pkg/github/client.go
@@ -16,7 +16,6 @@ import (
 	"time"
 
 	gh "github.com/google/go-github/v72/github"
-	"github.com/hivecommons/hive/pkg/config"
 	"github.com/hivecommons/hive/pkg/ioscan"
 	"github.com/hivecommons/hive/pkg/logscrub"
 )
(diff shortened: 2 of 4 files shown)


Danathar:
Closing: wrong base branch. Issue #5953 is titled `[v5] [architect] …` and belongs on **v5**; I based this on v4 by mistake.

Reopening against v5 shortly. Note the change is also *incomplete* as written for v5: on that branch `pkg/github/app.go` additionally reads `config.ProxyInjectGHAuth()` (the #1861 credential-divert switch), which v4's copy does not, so removing the imports from `client.go` and `automerge_sweep.go` alone still leaves `pkg/config` in the package's dependency tree there. The v5 PR will cover that too.

Nothing here needs review — please disregard rather than merge.

PR #5967, opened 2026-09-04
🌱 refactor(github): drop the pkg/config dependency from the client (phase 1 of #5953)

Danathar:
## Summary

`pkg/github` is the package everything in hive uses to talk to GitHub — ten
packages import it. It also imported `pkg/config`, the application's
configuration surface, so anything wanting an HTTP client dragged the entire
operator-configuration surface along with it, and the layering ran backwards
(configuration should depend on the client, not the reverse).

This removes that dependency. It is phase 1 of the four-phase split the issue
proposes, and the only phase that changes package boundaries; the policy
engines and watcher daemons stay where they are for now.

**No behaviour change**, and no call site changes: the seven places that
install an issue filter pass exactly what they passed before.

## Detail

Four references, not a deep entanglement:

| Reference | Use |
| --- | --- |
| `config.IssueFilterConfig` | type of `Client.issueFilter`; one method called, `Admits(labels)` |
| `config.DefaultAutoMergeLabel` | aliased by the `AutoMergeQueuedLabel` constant |
| `config.SelfMergeMinACMMLevel` | logged by the self-authored auto-merge sweep |
| `config.ProxyInjectGHAuth()` | read by `app.go` — the #1861 credential-divert switch |

Everything else the issue counts is prose in comments naming `config.RoleMerger`
and `config.AutoMergeConfig` — references to concepts, not imports. The issue
also names `tokenscopes.go` as an importer; it does not import `pkg/config`.

### The filter becomes a one-method interface

`IssueAdmitter` asks the only question the client actually has:

```go
type IssueAdmitter interface {
	Admits(labels []string) bool
}
```

`config.IssueFilterConfig` already satisfies it, so `SetIssueFilter` keeps
taking the same values and **all seven construction sites are untouched**
(`cmd/hive/main.go` ×6, `pkg/dashboard/api.go` ×1).

The matching logic is deliberately *not* copied here. It is an approval gate
with exact, case-insensitive semantics, and a second implementation that
drifted would fail **open** — the unsafe direction. The policy stays with the
layer that owns it; the client consumes the verdict.

`getIssueFilter()` now never returns nil. The old struct's zero value admitted
everything and an interface's zero value is nil, so without that guard "no
filter configured" would become a panic on the enumeration path rather than the
documented admit-all.

### The duplicated values are pinned, not trusted

Two constants and the divert switch are duplicated rather than imported.
`config_parity_test.go` pins each to its `pkg/config` counterpart so a change on
either side breaks the build rather than quietly changing behaviour.

The switch is pinned **behaviourally** — both readers run over the same env
values, including the near-misses `TRUE`, `True` and `1`, plus unset — rather
than by comparing constants. The risk being guarded is failing open: a reader
that returned false where config returns true would hand agents the real token
while the operator believed the divert was on.

`config_parity_test.go` is the only remaining reference to `pkg/config` in this
package. A test-only import does not put config back into the dependency tree
of the ten importers, which is what the issue is about.

### Verification

```
go list -deps ./pkg/github | grep -c hive/pkg/config
  before: 1
  after:  0
```

## Testing

- [x] `cd src && go test ./pkg/github/...` — pass
- [x] `cd src && go test ./pkg/dashboard/... ./cmd/hive/...` — pass
- [x] `cd src && go vet ./pkg/github/...` — clean
- [x] Other / not run (explain): full `go test ./...` not run locally; the
      change is confined to `src/pkg/github`. CI covers the rest.

New coverage in `src/pkg/github/config_parity_test.go`: both constants pinned to
their config counterparts; the divert switch pinned behaviourally across nine
env values plus unset, including an explicit assertion that it defaults **off**;
the env var *name* pinned, so the local reader cannot end up watching a variable
nobody sets; `config.IssueFilterConfig` pinned as satisfying `IssueAdmitter` (so
the seven call sites break here, loudly, rather than at the call); and the unset
filter pinned as admitting everything, including a nil label slice.

## Scope

Phase 1 only. Phases 2–4 from the issue — extracting the automerge policy
engine, the request watchers, and the advisory digest rendering — are untouched
and the issue should stay open for them, hence `Part of` rather than `Fixes`.

## Note for reviewers

No changelog fragment: `changelog.d/README.md` says routine refactors need none
and this has no operator-visible effect. Happy to add one, or take the

Part of #5953

Refs #5953

### DIFF
--- a/src/pkg/github/app.go
+++ b/src/pkg/github/app.go
@@ -14,8 +14,6 @@ import (
 
 	"github.com/golang-jwt/jwt/v5"
 	gh "github.com/google/go-github/v72/github"
-
-	"github.com/hivecommons/hive/pkg/config"
 )
 
 const (
@@ -481,7 +479,7 @@ func (a *AppAuth) WriteAgentToken(ctx context.Context, agentName, tier string, a
 	// registry is never populated and the file receives the real token,
 	// byte-identical to the pre-#1861 behavior.
 	fileToken := token
-	if config.ProxyInjectGHAuth() {
+	if proxyInjectGHAuth() {
 		storeAgentProxyToken(agentName, token)
 		fileToken = AgentDummyToken(agentName)
 	}
--- a/src/pkg/github/automerge_sweep.go
+++ b/src/pkg/github/automerge_sweep.go
@@ -12,8 +12,13 @@ import (
 	"time"
 
 	gh "github.com/google/go-github/v72/github"
-	"github.com/hivecommons/hive/pkg/config"
 )
+
+// selfMergeMinACMMLevel mirrors config.SelfMergeMinACMMLevel. It is duplicated
+// rather than imported so this package keeps no dependency on pkg/config
+// (hivecommons/hive#5953 phase 1); config_parity_test.go pins the two equal so
+// they cannot drift silently.
+const selfMergeMinACMMLevel = 6
 
 const DefaultAutoMergeSweepMaxMerges = 3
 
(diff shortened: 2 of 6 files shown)


clubanderson (approved):
Reviewed the v5 refactor: production pkg/github no longer imports pkg/config; call sites remain source-compatible via IssueAdmitter; unset filter preserves admit-all behavior; duplicated config values are pinned by parity tests. Local checks: go test -race -cover ./pkg/github/... passed (91.9%); go vet ./pkg/github/... passed; go test ./pkg/dashboard/... passed; go list -deps ./pkg/github shows 0 production pkg/config dependencies. I also attempted go test ./pkg/dashboard/... ./cmd/hive/..., but cmd/hive failed in this local environment trying to create /data agent dirs, which is unrelated to this PR.

PR #5968, opened 2026-09-04
🌱 refactor(cmd/hive): extract GitHub App key file handling to pkg/appkey (phase 1 of #5898)

Danathar:
## Summary

`cmd/hive/main.go` is 9,669 lines. That matters for more than readability: the
test workflows run `go test ./pkg/...`, so **nothing in `cmd/` is executed by
CI at all** — and the GitHub App signing-key logic living in `main.go` had 475
lines of tests that had therefore never run there.

This moves that logic into a new `pkg/appkey`, tests included. They now run
under `./pkg/...` with no workflow change, which is the concrete payoff the
issue identifies.

```
main.go: 9,669 → 9,386 lines,  136 → 129 top-level functions
```

**No behaviour change.** Phase 1 of the extraction the issue proposes.

## What moved

The App-key file lifecycle: where a spoke's keys live, which one is correct for
the App this hive claims, how a hub-delivered key is written down atomically,
and the two read-only reports over that same resolution ladder.

| moved | now |
| --- | --- |
| `perAppIDKeyPath` | `(*Resolver).PerAppIDKeyPath` |
| `perAppIDProvisionedKeyPath` | `(*Resolver).provisionedPerAppIDKeyPath` |
| `resolveAppKeyFile` | `(*Resolver).Resolve` |
| `writePerAppIDKey` | `(*Resolver).WritePerAppIDKey` |
| `heldPerAppIDKeyFingerprints` | `(*Resolver).HeldFingerprints` |
| `reportedAppKeyFingerprint` | `(*Resolver).ReportedFingerprint` |
| `hasPerHiveAppKey` | `(*Resolver).HasPerHiveKey` |

The last two were not on the issue's list; they came along because the moved
tests exercise them and they are pure reads over the same ladder `Resolve`
walks. Leaving them behind would have split one piece of logic across two
packages.

## Three decisions worth review

**Not `pkg/github`, which is where the issue suggests.** This code needs
`config.AppKeyFingerprint*`, and #5953 phase 1 has just removed `pkg/github`'s
dependency on `pkg/config`. Putting App-key handling there would reintroduce
exactly the upward dependency that change deleted. `pkg/appkey` depends on
`pkg/config` and is depended on by `cmd/hive` — the natural direction, and it
keeps both issues' goals intact.

**The path vars became `Resolver` fields.** This was forced rather than chosen:
they were unexported package vars that tests reassigned *and* that non-moving
`main` code still reads, so a literal move would have had to export mutable
globals. A struct avoids that. The side effect is a better seam — each test
gets its own `Resolver` instead of mutating state shared with every other test
in the package.

**`healGitHubAppInstallation` deliberately stays in `main`.** It mutates `cfg`
and calls `cfg.Save()`. That is application wiring, not the "pure/leaf helpers
with existing seams" phase 1 targets, and moving it would drag config mutation
into the new package.

## Reviewing the test move

The two test files are detected as renames, and moved **unchanged apart from a
mechanical rename onto the `Resolver`** — no assertion, table entry or case was
rewritten. The diff is reviewable as a move rather than a rewrite, which is the
point.

Two ten-line helpers (`writeTestKey`, `isolateAppKeyLookup`) are now duplicated
across the split. Both packages have tests that need them, and duplicating test
scaffolding beats exporting a testing surface from a production package.

## Testing

- [x] `cd src && go test ./pkg/appkey/...` — 10 tests, all pass; these are the
      ones CI could not previously reach
- [x] `cd src && go test ./cmd/hive/...` — pass
- [x] `cd src && go vet ./pkg/appkey/ ./cmd/hive/` — clean
- [x] `gofmt -l` clean on both packages
- [x] Other / not run (explain): full `go test ./...` not run locally; the
      change is confined to `cmd/hive` and the new package. CI covers the rest.

`go list -deps ./pkg/appkey` pulls only `pkg/config` and `pkg/resolve` — the new
package is not a second grab-bag.

## Scope

Phase 1 only, and only the first of its three suggested groups. The
budget-suppression state machines, the heartbeat/quota computation, and the
`main()` split into `run(cfg, deps) error` are untouched, so the issue should
stay open — hence `Part of` rather than `Fixes`.

## Note for reviewers

No changelog fragment: `changelog.d/README.md` says routine refactors need none
and this has no operator-visible effect. Happy to add one, or take the

Part of #5898

Refs #5898

### DIFF
--- a/src/cmd/hive/app_key_derivation_test.go
+++ b/src/cmd/hive/app_key_derivation_test.go
@@ -61,17 +61,17 @@ func writeKey(t *testing.T, path string) {
 // claim has a per-app-id key of its own.
 func TestResolveAppKeyFileIgnoresStaleGenericPin(t *testing.T) {
 	dir := t.TempDir()
-	origDir, origGeneric := spokeAppKeyDir, spokeAppKeyPath
-	spokeAppKeyDir = dir
-	spokeAppKeyPath = filepath.Join(dir, "gh-app-key.pem")
-	t.Cleanup(func() { spokeAppKeyDir, spokeAppKeyPath = origDir, origGeneric })
+	origDir, origGeneric := appKeys.DataDir, appKeys.DataKeyPath
+	appKeys.DataDir = dir
+	appKeys.DataKeyPath = filepath.Join(dir, "gh-app-key.pem")
+	t.Cleanup(func() { appKeys.DataDir, appKeys.DataKeyPath = origDir, origGeneric })
 
 	const publicAppID = 3568013
-	writeKey(t, spokeAppKeyPath)              // the stale GHE key
-	writeKey(t, perAppIDKeyPath(publicAppID)) // the correct public key
+	writeKey(t, appKeys.DataKeyPath)                  // the stale GHE key
+	writeKey(t, appKeys.PerAppIDKeyPath(publicAppID)) // the correct public key
 
-	got := resolveAppKeyFile(spokeAppKeyPath, "", publicAppID)
-	if want := perAppIDKeyPath(publicAppID); got != want {
+	got := appKeys.Resolve(appKeys.DataKeyPath, "", publicAppID)
+	if want := appKeys.PerAppIDKeyPath(publicAppID); got != want {
 		t.Fatalf("resolved %q, want %q — a stale generic pin must not shadow the key for the App we claim", got, want)
 	}
 }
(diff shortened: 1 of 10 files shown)


clubanderson (approved):
Reviewed the v5 extraction and pushed one cleanup commit removing a stale cmd/hive comment and updating design-doc references to pkg/appkey. The moved Resolver preserves key resolution precedence, per-app-id writes remain atomic/0600, tests moved with the code under ./pkg/appkey, and cmd/hive call sites use the Resolver without changing behavior. Local checks passed: go test -race -cover ./pkg/appkey/... (75.6% before #5976 follow-up), go test ./cmd/hive/..., go vet ./pkg/appkey/... ./cmd/hive/.... PR body now includes Refs #5898 without closing the umbrella issue.

Issue #5977, opened 2026-09-04
🌱 spokes still use per-host certs - wildcard exists but nothing serves it (oke + lke)


clubanderson:
## Summary

A fleet wildcard certificate exists and is valid, but **no spoke actually uses it**. Every spoke ingress still carries its own `tls:` block pointing at a per-namespace `hive-tls-hc` secret, and an explicit `tls:` block always wins over the controller's `--default-ssl-certificate`. So the wildcard is currently inert, and the fleet is still minting one certificate per host against a 50-per-week-per-registered-domain ACME cap.

This is the durable half of the mitigation applied during the 2026-09-03 rate-limit incident.

## Current state (measured 2026-09-04)

The wildcard: `hive-wildcard-tls` in ns `hive-hub` on **hive-oke**, issuer `letsencrypt-dns01` (DNS-01 via Cloudflare). SANs:

```
*.hive.hivecommons.dev
*.lke.hive.hivecommons.dev
hive.hivecommons.dev
```

| | hive-oke | lke648397 |
|---|---|---|
| ingress refs to `hive-tls-hc` | 59 | 136 |
| `Certificate` objects | 23 | 37 |
| hosts under `hive.hivecommons.dev` | 107 | 156 |
| hosts NOT wildcard-covered | `dibs.kubestellar.io`, `hive.kubestellar.io` | none |
| `--default-ssl-certificate` on nginx | set (`hive-hub/hive-wildcard-tls`) | **not set** |
| wildcard secret present | yes | **no** |

So ~195 ingress TLS references and ~60 Certificate objects exist to do what one wildcard already covers.

## Why this matters

Per-host certs make the ACME cap structurally reachable: ~263 wildcard-covered hostnames across the two clusters against a 50/week limit. **Every hostname change costs a certificate.** A live example on `hive-hosted-hosted-projectbluefin-knuckle-gjvq` — the stored secret still carried the pre-migration SAN

```
DNS:hosted-projectbluefin-common-nmq5.hive.kubestellar.io
DNS:hosted-projectbluefin-knuckle-gjvq.hive.hivecommons.dev
```

while the Certificate requested the `.hive.hivecommons.dev` form of both, so cert-manager reported `RequestChanged: Fields on existing CertificateRequest resource not up to date: [spec.dnsNames]` and tried to re-issue. That churn is what exhausted the quota.

[...]

PR #5981, opened 2026-09-04
🌱 feat(hub): serve provisioned spokes from the fleet wildcard certificate

Danathar:
## Summary

This is **item 2** of #5977 — the half the issue says cannot be done by hand, because the provisioner reconciles spoke ingresses back within minutes and a `kubectl` sweep across ~195 objects would not hold.

The wildcard existed and served nothing. Every provisioned spoke Ingress carried its own `tls:` block, and **an explicit `tls:` block always wins over the controller's `--default-ssl-certificate`** — so the wildcard was inert while each hive kept minting its own certificate. Measured 2026-09-04: ~195 Ingress TLS references and ~60 `Certificate` objects across two clusters, doing what one wildcard already covered, against an ACME cap of 50 issuances per week per registered domain. With ~263 wildcard-covered hostnames that makes **every hostname change cost a certificate**, which is what exhausted the quota on 2026-09-03.

Setting `wildcard_tls_secret` on a cluster in `clusters.json` now makes provisioned spoke Ingresses omit **both** the per-host `tls:` block and the `cert-manager.io/cluster-issuer` annotation. Both halves matter, for different reasons: the `tls:` block is what beats `--default-ssl-certificate`, and the annotation is what makes cert-manager's ingress-shim mint the Certificate at all. Dropping only one leaves either an inert wildcard or an orphaned Certificate.

## The asymmetry this is shaped around

Wrongly minting a certificate costs one issuance out of fifty. Wrongly **omitting** one does not degrade gracefully: ingress-nginx serves its built-in **self-signed** certificate and every spoke dashboard on the cluster fails TLS validation at once.

At the time of writing that is exactly `lke648397`'s state — wildcard secret absent, `--default-ssl-certificate` not set — while `hive-oke` has both. That's the issue's own item 1, and it isn't done yet.

So the decision is **opt-in per cluster, off by default**, and every uncertain case resolves to keeping the per-host certificate. Three independent vetoes:

- **No `wildcard_tls_secret`** — the operator has not asserted the cluster prerequisites hold. The hub cannot see either of them from here, which is why this is an assertion rather than an inference.
- **OpenShift Route clusters** — Routes use edge termination with no secret, so they never minted a per-host certificate through this path and `--default-ssl-certificate` is an ingress-nginx concept that doesn't apply. Nothing to change, so nothing changes.
- **The host isn't covered by `*.<domain>`.**

## Coverage is per host, not per cluster

`wildcardCoversHost` implements the RFC 6125 §6.4.3 single-label rule, so a host the wildcard cannot cover keeps its own certificate even on an opted-in cluster:

| Host | Covered by `*.hive.hivecommons.dev`? |
| --- | --- |
| `hosted-acme-ab12.hive.hivecommons.dev` | yes |
| `hive.hivecommons.dev` | no — the apex |
| `a.b.hive.hivecommons.dev` | no — deeper than one label |
| `dibs.hivecommons.dev` | no — one level *above* the wildcard's scope |
| `hive.kubestellar.io` | no — different domain |

That fourth row is the one worth flagging: #5925 moves `dibs` to a name **above** `*.hive.hivecommons.dev`, and a coverage check written on the domain suffix alone would have said yes and dropped the `tls:` block for a host nothing serves. It's a test case.

## Tests

`src/pkg/hub/wildcard_tls_test.go` — coverage including every negative above, each of the three vetoes, and the rendered manifest in both modes.

The manifest assertions run on the **parsed YAML**, not substrings: `annotations:` with an empty body parses as `null` and would slip past a string check, and `hive-api`'s only annotation is the issuer, so the opt-in has to remove the key itself. Also asserted: routing survives, the terminal's per-hive `auth-url` (a CWE-862 control) survives, unrelated annotations survive, and the OpenShift branch renders byte-identically either way.

The template is hand-written YAML with conditional blocks, so the likeliest break is indentation — and a manifest that doesn't parse fails on a cluster at provision time, not here. Every document is parsed with `gopkg.in/yaml.v3` in both modes.

Mutation-checked: removing one `tls:` guard fails two assertions.

Existing tests are unaffected — `renderManifestForTest` omits the new key, and Go templates treat a missing map key as the zero value, so the default renders today's output unchanged.

```
$ go test ./pkg/hub/ -count=1
ok      github.com/hivecommons/hive/pkg/hub     98.480s

checked 134 files under src/docs — all relative links and anchors resolve
```

`go build ./...` and `go vet ./pkg/hub/` clean.

## Scope

Items **1** (copy the wildcard to lke, set `--default-ssl-certificate`) and **3** (sweep the now-unused Certificates and secrets) are cluster operations and are not in this repo, so **#5977 stays open** — this is `Refs`, not `Closes`. Nothing here changes any existing spoke; the template governs newly provisioned and re-reconciled ones.

`src/docs/spoke-wildcard-tls.md` carries the prerequisites with the commands to verify them, the SNI check to run before deleting any old Certificate, and the renewal warning the issue raises: once spokes rely on the wildcard, a failed renewal takes down all ~263 hostnames at once rather than one. I have not implemented that alerting — it needs a decision about where fleet alerts live, and I'd rather flag it than guess.

I could not reach either cluster from this environment, so the measured state above is the issue's, not mine.

Refs #5977.


clubanderson (approved):
Reviewed for #5977 item 2. The change is opt-in per cluster, preserves per-host TLS for hosts outside the wildcard, and keeps lke648397 safe until its wildcard/default-ssl prerequisites are configured.

PR #5989, opened 2026-09-04
🐛 fix(dashboard): stop the copilot poll tests racing their own cleanups

### LINKED ISSUE #5985: [Tests] v4 test failure on 353efa0
## Test Failure

- **Commit:** `353efa04fc6fb028db98ac1c18754a3a1f488fbb`
- **Run:** https://github.com/hivecommons/hive/actions/runs/33901888827
- **Branch:** v4

Scheduled test run was failure. See the [workflow run](https://github.com/hivecommons/hive/actions/runs/33901888827) for details.

Danathar:
## What failed

Shard `test (rest 2/3)` of [run 33901888827](https://github.com/hivecommons/hive/actions/runs/33901888827):

```
copilot_auth_test.go:332: pollCopilotToken did not return in time
WARNING: DATA RACE
  Write at 0x...330 by goroutine 13851:
    dashboard.withCopilotMockURLs.func1()      <- a t.Cleanup
    testing.(*common).runCleanup()
    testing.(*common).Fatalf()
    dashboard.TestPollCopilotToken_SlowDownBumpsInterval()
  Previous read at 0x...330 by goroutine 13854:
    dashboard.(*Server).pollCopilotToken()     copilot_auth.go:224
--- FAIL: TestPollCopilotToken_SlowDownBumpsInterval (15.63s)
FAIL github.com/hivecommons/hive/pkg/dashboard
```

## Root cause

The test ran the poller in a goroutine it did not own. On a `select` timeout it called `t.Fatalf`, which runs the test's cleanups — and those restore the package-level `copilotAccessTokenURL` / `copilotUserTokenPath` that the **still-running** poller reads.

So a timing flake did not fail *as* a timing flake. It failed as a **data race** that took the whole `-race` package down, and the abandoned goroutine went on mutating package state under later tests.

Why the timing was tight: the test has to live through the production **5-second** `slow_down` bump in real time to prove the bump was honoured, inside an **8-second** budget. Three seconds of slack, on a loaded `-race` shard. That is what the run ate.

**The commit in the issue title is a red herring.** `353efa04` ("ci(release): enforce stable soak promotion") touches only workflows, docs and a shell script — no Go code, nothing in `pkg/dashboard`. The scheduled run just landed on it. `copilot_auth_test.go` has not been modified since the v2→src rename (#3996), so this has been latent for a long time waiting for a slow enough shard.

## Reproduced, then verified

I did not want to argue this from the stack trace, so both directions are measured:

| | forced timeout, full package, `-race` |
| --- | --- |
| **before** | **1 DATA RACE** — `withCopilotMockURLs.func1` vs `pollCopilotToken`, matching CI exactly |
| **after** | **0 races** — fails cleanly as the timeout it is |

That second row is the actual fix. A flake now reports as a flake.

## What changed

**`startCopilotPoll()` owns the goroutine** — cancelable context, joined in a `t.Cleanup`. The LIFO ordering of `t.Cleanup` is load-bearing and is called out where it matters: the var-restoring helpers register earlier, so the join runs **first** and the poller is gone before a single var is written back.

All **seven** poll tests move onto it, not just the one that failed. The same abandon-on-timeout shape was in every one of them; the others survived only on wider margins. `TestStopCopilotPoll_JoinsHandlerSpawnedPoller` has the same hazard in a different shape (the handler owns the poller), so it gets a *bounded* join in cleanup — bounded, so a genuinely wedged `stopCopilotPoll` reports instead of hanging the package.

**`copilotSlowDownBumpSec` becomes a `var`**, joining the three vars directly below it that exist for exactly this reason and say so in their comment. The test shrinks the bump to 1s — still an order of magnitude clear of the ~0 gap an unhonoured bump would produce, and the assertion scales off the constant rather than hard-coding a second number that can drift.

That is the only production line touched, and it is a `const`→`var` with an identical value and no behaviour change.

## Result

| | before | after |
| --- | --- | --- |
| `TestPollCopilotToken_SlowDownBumpsInterval` | 15.63s (**FAIL**) | 1.01s (PASS) |
| margin (budget ÷ work) | 1.6x | 6x |

```
$ go test ./pkg/dashboard/ -race -count=1
ok      github.com/hivecommons/hive/pkg/dashboard       87.539s     # 0 races

$ go test ./pkg/dashboard/ -race -count=10 -run 'TestPollCopilotToken|TestStopCopilotPoll'
ok      github.com/hivecommons/hive/pkg/dashboard       15.905s     # 0 races
```

`go build ./...` and `go vet ./pkg/dashboard/` clean.

## Note on the changelog fragment

In substance this is a test-only fix — the one production line exists solely to serve the test — so by `changelog.d/README.md`'s rule it would qualify for `no-changelog`. I added a fragment instead because the guard fires on the non-test file and the changelog has precedent for CI-stability entries (the backend-smoke lane repair in v4.12.0). Happy to swap it for the label if maintainers prefer.

Closes #5985.


clubanderson (approved):
Reviewed locally: this is a real goroutine ownership/join fix, not sleep-based synchronization.\n\nValidated:\n- go test -race -count=20 -run 'PollCopilotToken' ./pkg/dashboard/\n- go test -race ./pkg/dashboard/\n- go test -cover ./pkg/dashboard/ (86.8%; documented dashboard floor is 84 in v2-tests.yml)

PR #5993, opened 2026-09-04
🐛 fix(upgrade): stop losing the reason a self-upgrade failed

clubanderson:
## What happened

41 spokes reported `self-upgrade failed after 5 attempts: ` today - a trailing colon with nothing after it. The alert named a failure and said nothing about it, and the cause was not recoverable from the fleet.

## The defect

`recordUpgradeError` read the upgrade marker and returned silently when the read failed, **before** assigning `LastError`:

```go
data, err := os.ReadFile(path)
if err != nil {
    return   // the cause dies here
}
```

Every failure-reporting path reads that field, so an unreadable marker turned a real error into an empty string that propagated to the hub and onto the dashboard.

It now records the error against a fresh marker when the existing one cannot be read, and warns that it did. Losing the attempt count is survivable; losing the reason is what makes the failure undiagnosable.

## Second half: the dangling colon

`upgradeFailureSummary` replaces the raw format string. An empty cause no longer renders as `attempts: ` - a colon promising a reason and delivering none reads as truncation and sends the reader looking for the rest of a string that was never there. When nothing was captured it says so, and names the usual reason: the deployment tracks a tag that does not carry the target SHA.

## Context, and what this deliberately does NOT change

Today's fleet condition is the case that helper describes. Spokes track `ghcr.io/hivecommons/hive:stable`. #5980 correctly stopped per-merge publishes from moving `:stable`, so it now advances only through the hourly soak promotion. The hub still targets the newest merged SHA, so it is asking spokes to reach a SHA that will not be on `:stable` until the soak completes - the spoke restarts, comes back unchanged, and burns its five attempts. Retries 2-5 are additionally rejected as `minimum uptime not reached` (300s required, pods cycling every ~120s), so the budget is spent without a genuine attempt.

**This PR does not change that behavior.** It makes it legible. Whether upgrade targeting should be channel-aware is a separate question, filed separately.

## Tests

`src/cmd/hive/upgrade_error_visibility_test.go` covers the unreadable-marker path, field preservation on a readable marker, the nil-error no-op, and asserts the summary never ends in a bare colon for empty/whitespace causes while keeping a real cause verbatim.

https://claude.ai/code/session_01TmdVsn5zULh5KVpFkYrX59


clubanderson:
Reviewed upgrade error visibility. recordUpgradeError now preserves LastError across unreadable marker paths, terminal summaries avoid dangling empty reasons, and the immediate hub failure report already carries err.Error(); local src validation passed: go build ./..., go vet ./..., go test -race -cover ./cmd/hive.

/lgtm
/approve

Issue #5877, opened 2026-09-03
✨ dashboard: no way to re-run the ACMM evaluation — results are cached for an hour with no refresh and no staleness cue


Danathar:
## Summary

The dashboard's ACMM Evaluation panel scores each repo against the maturity criteria by checking whether certain files exist. That evaluation is cached in memory for **one hour**, and nothing can clear it — there is no refresh button, no query parameter, and no invalidation on any event. This breaks the panel's core workflow: an operator reads the list of failing criteria, commits the missing file, and then cannot confirm it passed for up to an hour, with no indication that what they are looking at is stale. Restarting the dashboard container is currently the only way to force a re-check. Requesting a **🔄 Re-evaluate** control plus a `?refresh=1` parameter on `GET /api/acmm/evaluation`. Blast radius: additive — a new opt-in code path; the existing hourly cache behavior is unchanged for every caller that does not ask for a refresh.

## Why the delay is a problem here specifically

The ACMM panel exists to drive a fix-and-verify loop: it lists a repo's gaps, and (since the **Open Issue** / **Open All** buttons) files them as issues. The natural next step is to fix one and confirm the criterion flips to ✅. Today that confirmation is unavailable for up to an hour after the commit lands.

Nothing signals the staleness either. The only hint is the `Evaluated: <timestamp>` line under the criteria grid, which is easy to miss and does not say the result is cached or when it will refresh. An operator who adds `.editorconfig`, reloads, and still sees ❌ has no way to tell "the cache is stale" from "my file is in the wrong place" — and the second reading invites pointless debugging.

## Current behavior

`src/pkg/dashboard/api_acmm_eval.go:20`

```go
const acmmEvalTTL = time.Hour
```

`handleACMMEvaluation` (`:95`) takes no parameters at all — it double-checks the cache under RLock then Lock, and only calls `evaluateAllRepos()` when the entry is older than the TTL. The sole writes to the cache anywhere in the tree are the two lines that populate it:

```
api_acmm_eval.go:135:  s.acmmEvalCache = &eval
api_acmm_eval.go:136:  s.acmmEvalCachedAt = time.Now()
```

There is no invalidation on adding a repo to `project.repos`, on a config save, on filing an ACMM issue, or on the governor cycle. The cache lives on the `Server` struct (`src/pkg/dashboard/server.go:92-93`), so a process restart is the only reset.

[...]

PR #5999, opened 2026-09-04
✨ feature: add a manual Rescan button to the Repositories section

Danathar:
## Summary

- Hive's dashboard shows a **Repositories** section: one card per monitored GitHub repo, with its open issue and PR counts and a row of "pills" for the individual tickets. Those cards are repainted only when the **governor** — the loop that decides how hard the hive is working and when to wake agents — finishes an evaluation pass, and that pass runs on a cadence that stretches to minutes when the hive is quiet.
- So an operator who has just closed an issue or opened a PR has no way to ask the dashboard for a fresh read. They stare at pills they already know are stale, with no button to press. The ACMM Evaluation panel directly below has had a **Re-evaluate** button for exactly this reason since #5877; Repositories had no equivalent.
- This PR adds a **🔄 Rescan** button to the Repositories section header. Pressing it re-enumerates every monitored repo's open issues and PRs right now and repaints the cards.
- **Blast radius:** additive. One new endpoint (`POST /api/repos/rescan`) and one new button; no existing behavior changes. The scan is read-only with respect to the hive — it cannot start work, kick an agent, or change the governor's mode.

### What the button does

The button POSTs to the new `/api/repos/rescan`, which calls `Dependencies.RescanReposFunc` — wired in `main.go` to `rescanRepos`, the **read-only half of `runEvalCycle`**, performed in the same order:

1. `EnumerateActionable`
2. the non-GitHub work-source overlay, when `governor.work_source.type` is set
3. `EnrichCIStatus` over the enumerated PRs
4. `applyDuplicatePRGuard`

It then stores the result into the same `lastActionable` pointer and `/data/last-actionable.json` cache the eval cycle writes, and calls `refreshDashboard`, so the cards repaint from *this* scan rather than waiting for the next poll.

### What it deliberately does not do

Everything in the eval cycle that **acts** is left out: no governor mode evaluation, no agent kicks, no advisory issue posting, no escalation sweep, no auto-merge sweep. That is what makes the button safe to press at any moment.

`recordRedStaleness` is skipped for the same reason — it advances the escalation clock, and refreshing a view must not age PRs toward "automated fix attempts exhausted". The consequence is that the claim guard's red-stale release predicate reads a clock that has not seen this pass, which can only make it **more** conservative (a claiming PR keeps suppressing its issue), never less.

### Budget protection

`handleReposRescan` debounces at 30s and collapses concurrent presses onto a single GitHub sweep. Both cases are answered from the previous scan's counts under a distinct `status` — `"debounced"` (with `retry_after_s`) and `"in-progress"` — so the UI reports *"already up to date"* rather than a failure. A **failed** scan does not arm the debounce, so a retry after a rate-limit blip is not refused.

Following the `?refresh=1` precedent on `/api/acmm/evaluation` that this mirrors, the endpoint is not owner-gated: it refreshes read-only data every dashboard viewer can already see, and the far more expensive ACMM re-evaluation is ungated too. The API budget is protected by the debounce, not by a role.

### Incidental cleanup

The `"/data/last-actionable.json"` literal appeared in three places (the boot-time restore, the eval-cycle write, and now the rescan write). It is folded onto one `lastActionablePath` constant so those cannot drift onto different files.

## Related issues

No tracking issue — requested directly by an operator. Closest prior art is #5877, which added the same affordance to the ACMM Evaluation panel.

## Testing

- [x] `cd src && go test ./pkg/dashboard/ ./cmd/hive/` — both green (dashboard 64.9s, cmd/hive 10.7s)
- [x] `cd src && go vet ./pkg/dashboard/ ./cmd/hive/`

New tests:

- `api_repos_rescan_test.go` — not-wired → 503; fresh counts and RFC3339 `scanned_at`; repo-count fallback to `Project.Repos`; the debounce actually preventing a second GitHub call while still returning the previous counts; scanning again once the window passes; concurrent presses collapsing to one sweep; a forge error returning 502 **without** arming the debounce.
- `repos_rescan_ui_test.go` — the button markup, the `data-action` handler name the delegated dispatcher resolves through `window[action]`, and the endpoint string are all present; and the button carries `data-stop="1"` so clicking it does not also collapse the section it just refreshed.
- `repos_rescan_test.go` (cmd/hive) — a hive with no GitHub credentials returns a named error instead of publishing an empty scan over the cards; the `/data` cache path constant is pinned.

`dashboard/openapi.json` documents the new route (the route-parity guard requires it).

## Contributor checklist

- [x] PR targets `v4` (the repo's current default branch; the template's `v2` line is stale).
- [x] Docs updated where behavior changes — `dashboard/openapi.json` carries the new endpoint contract.
- [x] `changelog.d/added-repos-manual-rescan.md` carries the changelog entry.

https://claude.ai/code/session_01BNuNsmvVTvem53coC2iEJu


clubanderson (approved):
Reviewed manual repository rescan. Added route-level auth coverage and API documentation; verified server-side debounce/in-flight collapse, read-only rescan path, existing same-origin/SameSite auth model, UI wiring, changelog, and local validation: go build ./..., go vet ./..., go test -race -cover ./cmd/hive ./pkg/dashboard, plus dashboard npm test after installing package deps.

PR #6005, opened 2026-09-04
🐛 fix(hub): target a spoke's release channel, not branch HEAD (#5994)

### LINKED ISSUE #5994: 🐛 Upgrade targeting is not channel-aware: spokes on :stable are told to reach a SHA the soak policy withholds
## Symptom

41 spokes in `UPGRADE FAILED` as of 2026-09-04 ~19:16Z, across LKE, akswec2, vllm-d and OKE. Each reports `self-upgrade failed after 5 attempts:` (empty cause - that reporting bug is #5993, separate).

## Mechanism

Spoke deployments track the moving tag `ghcr.io/hivecommons/hive:stable`. Measured on LKE: 36 of 36 deployments.

#5980 enforced the stable soak policy, correctly: per-merge v4 publishes now move `candidate` only, and `stable` advances solely through `promote-stable.yml` - hourly, after a 24h soak.

The hub's upgrade targeting was not taught about this. It still instructs spokes to reach the newest merged SHA:

```
"auto-upgrade debounce collapsed a merge burst into one roll"
  final_target: 526ef71   current: 77ba848
```

Digests measured at the time:

| Ref | Digest |
|---|---|
| `:stable` | `sha256:606af14b…` |
| `526ef71` (hub target) | `sha256:bc585b4a…` |

So the spoke restarts, pulls `:stable`, comes back reporting the same SHA, and the hub re-sends the same instruction. `src/cmd/hive/main.go:4313` predicts this exactly: *"A plain rollout restart only advances a deployment tracking a MUTABLE tag ... a restart loop that never lands."*

`promote-stable.yml` has **zero runs** since it was added today, so nothing has advanced `:stable` yet.

## Why the retry budget does not help

Retries require 300s minimum uptime, but the pods cycle roughly every 120s:

```
"self-upgrade retrying after a failed attempt (image unchanged)" attempt: 2
"self-upgrade deferred: minimum uptime not reached" uptime: 126s min_uptime: 300s
```

Attempts 2-5 are rejected before trying, so the budget is spent without a genuine attempt and the spoke reports failure having effectively tried once.

## Options

1. **Target the channel, not the SHA.** A spoke on `:stable` should be told to reach whatever `:stable` currently resolves to, by digest, and be considered current when it matches.
2. **Suppress the instruction when the target is not yet on the spoke's channel.** The hub knows the channel and can resolve the digest; asking for an unreachable SHA is what creates the alert.
3. **Pin spoke deployments by digest** so a restart is meaningful, which the same code comment argues for.

(1) or (2) look right; (3) is a larger change to provisioning.

## Not a regression in #5980

The soak policy is working as designed. The gap is that upgrade targeting assumed merges always moved `stable`, which was true until today.

## Related

- #5993 - the empty-cause reporting bug that made this hard to diagnose
- #5980 - the soak policy that surfaced it

Danathar:
## Summary

Spokes run a container image from a tag. Most of the fleet — 90 of 97 — is on the tag `:stable`, which the release process moves only after a build has soaked for 24 hours. The hub, meanwhile, was telling every spoke to upgrade to the newest commit merged to `v4`.

Those are two different images. A spoke can only ever run what its own tag points at, so the hub was asking for something the spoke could not get. Each spoke restarted, pulled `:stable` again, came back on the same version, and the hub repeated the instruction until the spoke gave up and reported `UPGRADE FAILED`. 41 spokes were in that state.

This PR makes the hub ask for the right thing: a spoke on `:stable` is told to reach whatever `:stable` currently is, and a spoke that already runs it is simply up to date. Nothing changes for a spoke that tracks a branch tag or a pinned version.

Fixes #5994.

## The mechanism

A spoke's Deployment tracks ONE image tag, and that tag is the entire set of builds it can reach. Rolling the pod re-pulls the tag; nothing the hub says can make a restart land on a digest the tag does not carry.

Before #5980, "the newest merged SHA on `v4`" and "the digest `:stable` carries" were the same image, so targeting branch HEAD happened to be reachable for everyone. The soak policy broke that identity on purpose: per-merge publishes move `:candidate` only, and `:stable` advances hourly after 24h.

Measured 2026-09-04, from the issue:

| Ref | Digest |
|---|---|
| `:stable` (what the spoke pulls) | `sha256:606af14b…` |
| `526ef71` (what the hub asked for) | `sha256:bc585b4a…` |

The spoke's own retry budget cannot rescue this. Retries require 300s uptime while the loop cycles pods every ~120s, so attempts 2–5 are rejected as `minimum uptime not reached` and the spoke reports failure having effectively tried once.

## What changed

This is **option (1)** from the issue — target the channel, not the SHA — which subsumes option (2): a target that is not on the spoke's channel is never *chosen*, so it never has to be suppressed. Option (3), pinning spoke Deployments by digest, remains the larger provisioning change it was described as.

**`src/pkg/hub/channel_targeting.go` (new)** resolves a channel tag to the commit it was built from, by reading `org.opencontainers.image.revision` off the image config — the same evidence `src/scripts/promote-stable.sh` reads, and the only thing that survives a retag (a channel tag's *name* never names a commit, and a digest cannot be turned back into one). Cached for `channelDigestTTL`, with a bounded stale grace so one GHCR blip cannot stall the fleet's upgrades.

**`triggerAutoUpgrades` and the spoke-managed heartbeat path** now target that commit for a channel-tracking spoke, and branch HEAD for everything else — byte-for-byte the previous behaviour for a spoke on `<branch>-latest`, on a SHA pin, or reporting no `image_ref` at all.

**The floating-tag latch-clear** in `triggerAutoUpgrades` judges "at latest" through the tracked tag too. That is the half that actually releases the 41 latched spokes: they had already converged on everything their tag offers.

Two deliberate refusals:

- **An unresolvable channel instructs NOTHING**, loudly. Falling back to branch HEAD is the bug, and it would be the worst-evidenced fallback available — we know the spoke's ceiling is not HEAD, and we do not know what it is. The next cycle retries.
- **A spoke AHEAD of its channel is left alone.** A channel is a moving pointer, not a monotonic branch, so this state is reachable; targeting the channel commit there would instruct a *downgrade*, a failure mode this path never had while it only chased HEAD.

`resp.LatestSHA` is deliberately unchanged. It answers "how far has the branch moved", which the dashboard's behind-count needs, and that is a different question from "what can this spoke reach".

## Recovery

No per-spoke intervention is needed, per the maintainer's own note on the issue: the spoke's attempt marker is keyed on `(current_sha → target_sha)` and cleared as stale when either moves, so the first reachable target the hub sends gives every spoke — including the 41 that gave up — a fresh budget.

## Testing

- `src/pkg/hub/channel_targeting_test.go` (new): the regression case (a `:stable` spoke is targeted at the channel's commit, not the newer branch head), the alert-clearing case (a spoke at its channel's commit gets no instruction), both refusals above, the branch-tracking and no-`image_ref` spokes staying on the old behaviour, cache/stale-grace/retry-after-failure semantics, and the registry walk itself against an `httptest` server — including buildx's `unknown/unknown` attestation descriptor, a single-platform manifest, a missing revision label, and four fail-closed shapes.
- `go test ./pkg/hub -short -race -count=1 -shuffle=on` passes (114.8s), as does the full non-short suite.
- `go build ./...`, `go vet ./...`, `gofmt -l` clean on the touched files.
- `pkg/hub` coverage measures 91.6%, against the gate's floor of 87.


clubanderson (approved):
Reviewed channel-aware targeting against #5994. Added regression coverage for registry failure paths; local src validation passed: go build ./..., go vet ./..., go test -race -cover ./pkg/hub (package coverage 89.8% under -race; non-race coverage reports 90.0%).

Issue #6003, opened 2026-09-04
🐛 Spokes crashloop on nodes without xt_owner/REDIRECT: unguarded iptables appends die under set -e


clubanderson:
## Symptom

On the vllm-d cluster, every spoke's NEW pod crashloops while its previous pod keeps serving. Measured 2026-09-04 ~21:00Z: restart counts of 9, 25, 26, 30, 37 across `hive-hosted-*` namespaces, `CrashLoopBackOff`, container exit code **4**. `desired=1 ready=1` in each - the ready replica is the OLD pod, so these spokes cannot roll forward at all.

LKE (37 pods) and OKE (19 pods) are unaffected: 0 restarts on every pod.

## Cause

`src/deploy/entrypoint.sh` runs under `set -e` (line 2). In the iptables block, the two `--uid-owner` appends are deliberately guarded:

```sh
$IPT -t nat -A HIVE_PROXY -m owner --uid-owner 0 -j RETURN || true
$IPT -t nat -A HIVE_PROXY -m owner --uid-owner "$PROXY_UID" -j RETURN || true
# OpenShift/OVN: packet-mark exemption (works with no xt_owner).
$IPT -t nat -A HIVE_PROXY -m mark --mark "$HIVE_PROXY_EGRESS_MARK" -j RETURN
$IPT -t nat -A HIVE_PROXY -p tcp --dport 443 -j REDIRECT --to-ports "$PROXY_PORT"
$IPT -t nat -A OUTPUT -j HIVE_PROXY
```

The comment on the guarded pair says `|| true` "keeps a failed append non-fatal on OpenShift (no xt_owner)". The following three appends carry no guard, so on a host where they fail, `set -e` kills the script.

From the crashing pod:

```
Warning: Extension owner revision 0 not supported, missing kernel module?
iptables v1.8.11 (nf_tables): RULE_APPEND failed (No such file or directory): rule in chain HIVE_PROXY
Warning: Extension REDIRECT revision 0 not supported, missing kernel module?
iptables v1.8.11 (nf_tables): RULE_APPEND failed (No such file or directory): rule in chain HIVE_PROXY
```

Both `xt_owner` AND the `REDIRECT` target are unavailable on these nodes (RHEL 9.4, kernel 5.14.0-427.102.1.el9_4).

## Why this is a bug rather than the gate working as designed

[...]

PR #6008, opened 2026-09-04
🐛 fix(entrypoint): guard the egress rule appends and judge the result (#6003)

Danathar:


clubanderson:
I compared this with merged #6006 (1dc06fc). The runtime fix from this PR was now redundant and conflicted with #6006, so I rebased onto current v4 and kept only the incremental value: an extracted-block test that covers both IPv4 and IPv6 result judging.

Important: the rebased branch keeps #6006 semantics. Optional uid-owner exemptions only warn; required mark, REDIRECT/REJECT, and OUTPUT appends still capture stderr, flush partial chains, and feed the fail-closed/advisory decision. The new test pins that behavior and specifically rejects the "owner succeeded so mark may fail" semantic drift from the original version.

Issue #5328, opened 2026-08-31
🐛 Orphaned Terminating spoke pods accumulate indefinitely — 27 across 15 namespaces, oldest 3 weeks


clubanderson:
## Problem

Hosted spoke pods get stuck in `Terminating` indefinitely and nothing reaps them. A sweep of the `vllm-d` cluster today found **27 orphaned pods across 15 hive namespaces**, the oldest stuck since **2026-08-10** — three weeks.

A separate namespace (`vllmd-260731-fikn`) held **5 more**, oldest 2026-08-11, for a total of **32 cleared manually**.

## Evidence

Every orphan shared the same signature:

- **`finalizers: []`** — nothing was legitimately holding them. No controller had unfinished work.
- **`status.phase != "Running"`** — all already carried a `deletionTimestamp`.
- Each namespace still had exactly **1 healthy Running pod**, so the live spoke was unaffected.

That combination points at nodes disappearing without draining: the API server records the deletion, the kubelet never confirms it, and the pod object persists forever because no finalizer or controller is responsible for finishing the job.

Distribution was wide rather than concentrated — 15 namespaces with 1–4 orphans each — which argues for a recurring cluster-level event, not one bad spoke.

## Impact

1. **Fleet counts are wrong.** Anything counting pods per namespace sees phantom replicas. A namespace showing 4 pods when 1 is live misrepresents capacity and makes "is this spoke healthy" harder to answer from the dashboard.
2. **Scheduler slots are held.** Orphaned pods still occupy their place until forcibly removed.
3. **It accumulates silently.** Three weeks of buildup with no alert. The condition self-perpetuates because nothing sweeps it, so the count only grows between manual interventions.

## Remediation applied

Cleared by hand with `oc delete pod --force --grace-period=0`, one namespace at a time, after verifying each target had no finalizers and was not `Running`. Post-sweep the cluster reports **0 stuck pods**, and all 15 namespaces show exactly 1 Running pod.

This is a one-time fix; the underlying condition is untouched.

## Proposed work

1. **Find the root cause.** Determine what removes nodes without draining on this cluster — spot/preemptible reclaim, node autoscaler scale-down, or an OKE/OCP node lifecycle event. Fixing that stops production of new orphans.
[...]

PR #6010, opened 2026-09-04
🌱 feat(hub): watch the wildcard certificate that now serves every spoke (#5977)

Danathar:
## Summary

Each spoke used to have its own TLS certificate. #5981 changed that: on a cluster that opts in, spokes are served by one shared wildcard certificate instead — which is the whole point, since minting one certificate per spoke is what exhausted the Let's Encrypt quota.

But it moves the risk. Before, a certificate that failed to renew broke **one** dashboard. Now one certificate sits behind **every** spoke on the cluster — about 263 hostnames — so a renewal that quietly fails breaks all of them at once. Nothing was watching it.

This PR makes the hub check that certificate every time it gathers cluster health, and say so when something is wrong: it is close to expiring, already expired, missing, unreadable, not served as ingress-nginx’s default SSL certificate, or does not actually cover the domain. A healthy certificate shows nothing.

Refs #5977. The issue stays open until the operator completes the LKE manual steps; this PR addresses the monitoring prerequisite.

## Why now, and why here

The wildcard PR deferred this: *"it needs a decision about where fleet alerts live, and I'd rather raise that than pick a home for it."* That was the right call at the time, but the codebase does answer it — `PerClusterHealth` already carries two hub-gathered per-cluster findings added for exactly this reason, `StuckPods` (#5328) and `LeakedNamespaces` (#5768). Same shape, same nil-means-unknown contract, same panel. This follows them rather than inventing a home.

`src/docs/spoke-wildcard-tls.md` also lists monitoring as a **prerequisite for enabling the opt-in fleet-wide**, so this unblocks the rest rather than being a nice-to-have.

## It is also the first check on the opt-in itself

`wildcard_tls_secret` is an operator assertion, and `wildcard_tls.go` is explicit that the provisioner cannot verify it — it must decide without a cluster round-trip, and guessing wrong takes the cluster down. That reasoning holds at provision time. It is not a reason never to check: the health build already talks to every cluster on a timer, where being wrong costs a warning instead of an outage.

Two of the statuses are that assertion turning out to be false, and **neither is visible from `clusters.json`**:

- **`missing`** — the flag is set and the secret is not there. Every wildcard-served spoke on the cluster is being handed ingress-nginx's built-in self-signed certificate right now.
- **`domain_mismatch`** — the secret is there and does not carry `*.<cluster domain>`. `servesHostFromWildcard` decides coverage from the domain in `clusters.json`; this is the only place the *certificate* gets a say.

Coverage is judged on the **wildcard SAN**, not on whether today's hosts happen to be listed. The hub omits the `tls:` block for every single-label host under the domain including hives that do not exist yet, so a certificate enumerating today's hosts would pass a per-host check and break the next hive provisioned.

## Statuses

| `status` | Meaning | Rendered |
| --- | --- | --- |
| `ok` | Covers the domain, >21 days left | nothing — a healthy fleet stays quiet |
| `expiring` | <21 days left | `wildcard cert expires in Nd` |
| `expired` | past `notAfter` | `wildcard cert EXPIRED` |
| `missing` | opted in, secret absent | `wildcard cert MISSING` |
| `domain_mismatch` | no `*.<domain>` SAN | `wildcard cert does not cover this domain` |
| `not_served` | ingress-nginx does not advertise `--default-ssl-certificate=<secret>` | `wildcard cert is not the ingress default` |
| `unreadable` | `tls.crt` will not parse | `wildcard cert unreadable` |

**21 days means renewal is overdue, not that expiry is close.** cert-manager's default `renewBefore` starts renewing a 90-day Let's Encrypt certificate at 30 days out, so warning at 30 would fire on every healthy renewal, every quarter, on every opted-in cluster — and an alert that is normally firing is one nobody reads. Reaching 21 means the renewal window opened over a week ago and nothing came of it, which is the DNS-01 path or the ACME quota being stuck: the very failure #5977 was filed about. Three weeks of headroom remain.

## Failure directions

Every uncertain case resolves to "say nothing", matching the asymmetry the wildcard work is built around:

- **A cluster the hub cannot reach reports nothing, never `ok`.** Same unknown-is-not-healthy contract as `StuckPods`/`LeakedNamespaces`.
- **`kubectlSaysNotFound` fails toward unknown.** It separates "the secret is gone" from "the cluster was unreachable" on kubectl's stderr, and anything it does not recognise is treated as unreachable. An RBAC denial, an expired kubeconfig or a network blip must never be reported as a vanished certificate.
- **Coverage is judged before expiry.** A certificate that cannot serve the domain is already failing every request; calling it "expiring" would send an operator to renew a certificate that would still be wrong.
- **Read-only, and only where it matters.** One `kubectl get deployment` and one `kubectl get secret` per opted-in cluster per health build. Clusters without the opt-in still carry per-host certificates and issue no request at all. OpenShift Route clusters are vetoed, restating `servesHostFromWildcard`'s own veto so the two cannot drift.

## Testing

Every pure function at **100%**; `pkg/hub` at 91.7% against the gate's floor of 87.

- **Parse**: 7 fail-closed shapes (not JSON, no `data`, no `tls.crt`, not base64, not PEM, PEM that is not a certificate) each with a passing control, plus a **chain** — taking the last block instead of the leaf would report the CA's expiry years out and never warn.
- **Coverage**: the real fleet SAN set; case/trailing-dot; and the three near-misses that must be rejected — today's hosts listed explicitly, the apex alone, `*.hivecommons.dev` one level up.
- **Status boundaries**: 30 days quiet (cert-manager's own renewal window), exactly 21 quiet, one minute inside noisy, exactly at `notAfter` expired, past it negative days; coverage outranking expiry; nil cert → `unreadable` with `days_remaining` left nil, because a zero there would read as "expires today".
- **`kubectlSaysNotFound`**: the not-found message convicts; unreachable / Forbidden / Unauthorized / timeout / non-`ExitError` / nil all do not.
- **`collectWildcardTLSHealth` end to end** against a scripted `kubectl` (following `installScriptedKubectl` next door): the missing-secret finding, the default-SSL-not-configured finding, the unreachable→no-report case, a live certificate, and an unparseable secret. Plus the four vetoes, and the JSON keys the dashboard switches on.
- Dashboard render checked per branch under node, including that an `expiring` report with no day count says "expiring soon" rather than falling through to "unreadable".
- `go build ./...`, `go vet ./...`, `gofmt` clean; `go test ./pkg/hub -short -race -count=1 -shuffle=on` green after maintainer updates (176.6s).

## Not included

Watching the **DNS-01 solver itself**. This check sees the certificate that is installed, so it reports a renewal that already failed — it cannot see one about to fail for a Cloudflare-credential reason. Noted in the doc rather than guessed at.


clubanderson (approved):
Reviewed the wildcard TLS health path. I added the missing ingress-nginx default SSL assertion check so the hub now verifies the secret is actually served by default before reporting the certificate healthy.

PR #6088, opened 2026-09-05
fix(podman): reconcile the repo-owned files a Quadlet host runs from

### LINKED ISSUE #6078: 🐛 bug: Podman/Quadlet hives keep serving the gateway config they were installed with — nginx.conf is never refreshed by any update path
## Summary

Hive's Podman/Quadlet deployments keep serving the nginx gateway config they
were first installed with, indefinitely — no update path ever replaces it.

- **What the gateway is.** A hive that isn't on Kubernetes runs two containers:
  `hive` (the Go binary) and `hive-gateway` (nginx), which is the thing that
  actually answers on `:3001` and proxies through to the binary. Its config
  comes from `src/deploy/nginx.conf` in this repo.
- **What's wrong.** On the Quadlet path that config is *copied to the host* at
  install time and bind-mounted in. Updating hive replaces the image — and
  therefore the Go binary — but nothing ever re-copies the config. So any fix
  to `src/deploy/nginx.conf` ships in the source, appears "released", and
  stays broken on every existing deployment forever.
- **Why nobody notices.** The hive keeps reporting a `served_sha` that
  *contains* the fix, because that SHA describes the binary, not the gateway.
  An operator checking whether they picked up a fix gets told yes.
- **Blast radius.** Podman/Quadlet standalone deployments only, for any change
  to `src/deploy/nginx.conf`. Docker Compose is **not** affected — it mounts
  `./deploy/nginx.conf` straight from the checkout, so `git pull` +
  `docker compose up -d` refreshes it. Kubernetes doesn't use this gateway.

## Observed

`hivecommons/hive#5199` — the nginx gateway dropping the WebSocket upgrade on
`/api/`, which breaks the contributor relay against any self-hosted hive — was
fixed by #5200, merged 2026-08-30.

On a Podman/Quadlet host on 2026-09-05, six days later:

```
$ curl -s http://localhost:3001/api/contribute/status
{"actionable_items":6,"active_contributors":0,"api_version":"1.2",
 "hub":"online","served_sha":"df9b867","surface":"hub","total_registered":1}

$ git merge-base --is-ancestor 914fe5e6 df9b867b && echo "fix IS in this build"
fix IS in this build
```

The build it serves contains #5200. The gateway still rejected the handshake:

```
$ node -e "new (require('ws'))('ws://localhost:3001/api/contribute/ws')
    .on('unexpected-response',(q,r)=>console.log('HTTP',r.statusCode))"
HTTP 400
```

That 400 is gorilla's `returnError` — `Content-Length: 12` ("Bad Request")
with a `Sec-Websocket-Version: 13` response header — i.e. the upgrade reached
the Go handler with `Connection`/`Upgrade` already stripped, exactly the
pre-#5200 behaviour. Same client against a hub whose gateway *had* been
rebuilt succeeds on the first try:

```
OPEN — hosted handshake succeeded
MSG: {"type":"auth_challenge","seq":1,"nonce":"7e7333a924af7a4c193bf072d42c4371"}
```

Copying the current conf in by hand and restarting only the gateway fixed it
immediately:

```
$ sudo install -o hive -g hive -m 644 \
    src/deploy/nginx.conf /var/home/hive/.config/hive/nginx.conf
$ sudo -u hive XDG_RUNTIME_DIR=/run/user/1001 systemctl --user restart hive-gateway
$ node -e "..."   # same one-liner as above
OPEN — local handshake succeeded
```

The host had been auto-updating its image the whole time. The binary moved;
the gateway never did.

## Root cause

Three independently reasonable decisions that combine into a permanent gap:

1. `src/deploy/quadlet/hive-gateway.container:84` mounts the config **from the
   host**, not from the image:

   ```
   Volume=%E/hive/nginx.conf:/etc/nginx/nginx.conf:ro,Z
   ```

   `%E` is `/etc` for a rootful unit and `~/.config` for a rootless one, so
   this is `/etc/hive/nginx.conf` or `~/.config/hive/nginx.conf`.

2. `place()` (`bin/hive-podman-setup.sh:445`) deliberately refuses to overwrite
   an existing file without `--force`, and `nginx.conf` goes through it
   (`:460`) alongside `hive.yaml` and `hive.env`. That is correct for the other
   two — they hold operator edits and a generated dashboard token — but
   `nginx.conf` is not a file operators are expected to edit, so it gets the
   preservation semantics of a user config while actually being shipped code.

3. `bin/hive-podman-update.sh` writes only the image-pin drop-in
   (`hive.container.d/10-image.conf`) and, if enabled, `20-autoupdate.conf`.
   It contains no reference to `nginx.conf` at all (`grep -c nginx` → 0). With
   `autoupdate on` the image then rolls forward with no operator action, so
   there is no moment at which anyone is prompted to reconcile the config.

Nothing documents a follow-up step either — `README.md:105` only says setup
preserves existing config without `--force`, which reads as reassurance rather
than as a warning that gateway fixes will not land.

Note that `setup --force` is *not* a safe workaround: it re-copies `hive.yaml`
and `hive.env` too, and regenerating `hive.env` takes the dashboard token with
it.

## Suggested direction

Not attached to any of these, but roughly in order of preference:

- **Stop treating `nginx.conf` as operator config.** Bake it into the gateway
  image, or mount it from a path the update path owns and always rewrites.
  Operators don't edit it; the repo does. This removes the class of bug rather
  than the instance.
- Failing that, have `hive-podman-update.sh` re-copy `nginx.conf` on every
  update — it is the only one of the three `place()` files that carries no
  operator state, so it can be refreshed unconditionally.
- At minimum, make the drift **visible**: have the update path diff the
  installed conf against `src/deploy/nginx.conf` and warn, and/or have the
  gateway report its own config hash next to `served_sha` so
  `/api/contribute/status` stops implying a fix is live when it isn't.

The last point matters independently of the fix: `served_sha` describing only
the binary is what turned a one-line config staleness into six days of
"the fix is in, why is it still broken".

## Environment

- Podman/Quadlet standalone, rootless (separate `hive` user), Fedora-family
  image-based host
- Gateway containers `hive` + `hive-gateway`, published on `:3001`
- Served build `df9b867`, image auto-update enabled
- Installed 2026-08-22; gateway conf still the 2026-08-22 copy as of 2026-09-05

Danathar:
## Summary

A Podman/Quadlet hive keeps serving the gateway config it was installed with,
forever. Updating Hive replaces the **image**, so the Go binary moves — but the
gateway does not read `nginx.conf` out of that image. `hive-gateway.container`
bind-mounts it **from the host**, and nothing in any update path ever re-copies
it. So a fix to `src/deploy/nginx.conf` ships, releases, and never lands.

From #6078, measured rather than reasoned: the WebSocket-upgrade fix in #5200
merged 2026-08-30, and a host installed 2026-08-22 was still returning HTTP 400
on the contributor handshake six days later — while `/api/contribute/status`
reported a `served_sha` that **contained** the fix, because that SHA describes
the binary and not the gateway.

This adds `bin/hive-podman-update.sh reconcile`, which compares the eight
repo-owned files a Quadlet host runs from against a checkout and re-copies the
ones that fell behind.

## One correction to the issue's root-cause analysis

The issue (and the follow-up comment) attribute both instances to `place()`
refusing to overwrite. That is true of `nginx.conf` and **not** of the unit
files — `bin/hive-podman-setup.sh:564` installs those with an unconditional
`install -Dm644`, so a re-run of `setup` already refreshes them. They are stale
only because nothing ever re-runs it.

The distinction matters because it makes the fix smaller than the issue
suggests. The gap is not "`place()` is too cautious". It is that **no command
refreshed the files the repo owns without also rewriting the ones the operator
owns** — `setup --force` does both, and regenerating `hive.env` takes
`HIVE_DASHBOARD_TOKEN` with it, which is exactly why the issue rules it out as
a workaround. Nothing about `place()`'s policy needed to change.

## What this does

```
bin/hive-podman-update.sh reconcile          # compare only. 0 in sync, 78 stale.
bin/hive-podman-update.sh reconcile apply    # re-copy the ones that fell behind
```

Managed (repo-owned, refreshed): `nginx.conf`, the four Quadlet units, the two
boot units. **Never touched:** `hive.yaml`, `hive.env`, `secrets/` — and
`reconcile` prints that list before it writes anything.

Two integrations, each a choice rather than a default:

- **`pin` refreshes the gateway config** and restarts the gateway. That file
  carries no operator state, `pin` has already been told to move this deployment
  forward, and it restarts the gateway a few lines later anyway. This is the
  issue's preference #2, applied to the one path a script can actually reach.
- **`status` reports drift unasked.** This is the only cover the auto-update
  path can get, and it is worth being explicit about why: `podman auto-update`
  is a systemd timer calling podman directly. **No script is in that path**, so
  there is no "refresh on every update" available to hook there — detection is
  what is possible, so detection is what it does.

Unit files are reported by `pin`, never rewritten by it: a unit change needs a
container recreate to mean anything, which is downtime the operator should
choose, and rewriting `hive.container` under an active digest pin edits the file
the pin exists to leave alone.

`reconcile check` exits **78 on drift**, deliberately. The follow-up comment's
sharpest point is that a frozen host is byte-identical to a current one in
everything a monitor can read (`UPDATED=false`, exit 0, timer green). That exit
status is the readable state that did not exist before.

It also covers the second instance from the comment: a `hive.container` still
naming the retired `ghcr.io/kubestellar/hive` is reported and repaired.

## Scope and risk

CI-and-operator tooling only. No Go code, no image contents, no API surface.

The risk is that `reconcile apply` and `pin` now **write to the host**, which
`hive-podman-update.sh` previously never did outside its own drop-in. That is
bounded three ways: the destination list is fixed in the script (not derived
from anything on disk), operator-owned files are excluded by construction and
asserted by tests, and `check` — the default — writes nothing at all.

`reconcile` is the one command here that needs a checkout as well as a host.
With none it **refuses** rather than reporting that everything matches:
"not checked" and "up to date" must not print the same thing, which is the same
class of lie as the `served_sha` that hid this for two weeks.

## Deliberately not in this PR

The issue's third suggestion — having the gateway report its own config hash
next to `served_sha` — is a Go/API-contract change on a different surface. It is
worth doing and is not folded in here.

## Existing hosts still need one manual run

The follow-up comment's "this issue is self-blocking" point is correct and is
not solved by merging this: the fix ships **inside** the files that are never
refreshed, so it cannot reach the fleet on its own. Every pre-existing
deployment needs one run from a checkout regardless of what merges. That step is
written into `src/docs/podman-quadlet-update-rollback.md` and into the changelog
fragment, so it reaches operators rather than living only in this PR.

## Validation

Static and mocked only — every input in the suite is stubbed, so no Podman, no
Quadlet, no privileges, no pull, no restart.

| Command | Outcome |
| --- | --- |
| `bash bin/test_hive_podman_update.sh` | **136 passed, 0 failed** (was 93; +43) |
| `bash bin/test_hive_podman_setup.sh` | 154 passed, 0 failed |
| `bash src/deploy/test_quadlet_config_contract.sh` | 14 assertions pass |
| `bash src/deploy/test_standalone_image_refs.sh` | 31 assertions pass |
| `bash bin/test_bin_suites_wired.sh` | 27 suites, 0 orphaned |
| `shellcheck` on both changed scripts | no new finding classes vs `upstream/v4` |
| `bash -n` on both changed scripts | clean |
| `python3 src/scripts/check-docs-links.py` | 135 files, all links and anchors resolve |
| changelog-fragment-guard rules, applied by hand to the fragment | filename and `- ` shape both match |

CI itself has **not run on this PR**: every workflow is sitting at
`action_required`, which is this repo's gating for a fork contributor, so the
lanes above were reproduced locally rather than observed green. `DCO` and
`PR Verifier` did run and pass. The v2-ci lane invokes
`bash bin/test_hive_podman_update.sh` from the repo root — byte-identical to
what was run here.

**The new tests were verified to discriminate.** Passing assertions prove
nothing on their own, so each rule was removed from the script and the suite
re-run:

| Mutation | Result |
| --- | --- |
| `pin` stops refreshing the gateway config | 2 fail, incl. "pin refreshes a stale gateway config" |
| `nginx.conf` dropped from the managed set | 8 fail across reconcile, status and membership |
| `check` applies instead of reporting | 5 fail, incl. "check writes nothing" |
| `hive.yaml` wrongly added to the managed set | 6 fail, incl. "apply preserves an operator's hive.yaml" |
| unknown-gateway guard removed | 1 fail, the case written for it |

The script was restored after each and the suite re-run green.

There is also a guard against this recurring for a file that does not exist yet:
`setup`'s `UNITS`/`BOOT_UNITS` and `reconcile`'s `MANAGED_UNITS`/
`MANAGED_BOOT_UNITS` are compared to each other, so a unit added to the
installer and forgotten in the reconciler fails CI rather than being installed
once and frozen forever.

### Skipped, and stated rather than implied

- **No host was touched.** Nothing here was run against a real Quadlet
  deployment, rootful or rootless; the reported host is not mine. The evidence
  is the mocked suite plus the mutation runs above, and a green check on this PR
  would not be evidence that a real gateway restarted onto a new config.
- **The suite must not write to the runner's `$HOME`.** `pin` now writes the
  gateway config, so an unisolated suite would drop files into the real
  `~/.config/hive` of whoever ran it. `reset_env` redirects all three
  destinations into the temp dir; verified afterwards that the local
  `~/.config/hive` still contains only its pre-existing files and no
  `nginx.conf`, `hive.yaml` or `hive.env`.
- The suite is wired into CI already (`bin/test_hive_podman_update.sh` is an
  existing, referenced suite), so the new cases run there without new wiring.

## Note on the base branch

Opened against `v4`, which is this repo's base branch, as requested.

## External state

None. No images, tags, VMs, storage pools, or workflow runs were created or
mutated.

Closes #6078


Danathar:
Both findings addressed in `88eec04e` and `c264ab74`.

## The harness flake — `88eec04e`

Confirmed and fixed with the herestring, matching the #5969 comment verbatim.
**8 consecutive runs, 136 passed / 0 failed** each, against the 4-of-6 you
measured.

One thing beyond what you flagged: the same construct is at **seven more sites
in this file**, all reached through `check`'s `eval` under the same
`set -uo pipefail`, so I fixed them together rather than leaving known hazards
next to the fixed one. The negated ones are the worse half — there an EPIPE
inverts a real failure into a **silent PASS**, so a regression they exist to
catch goes unreported rather than merely flaking.

One needed more than a herestring:
`run_update status | grep -A3 "Pin history" | grep -qv ...` has the race twice
over, since `grep -qv` stops at its first *non*-matching line. Both stages are
captured into variables now.

## The gateway config — `c264ab74`

Agreed, and I took the backup option rather than the print-a-warning one.

`refresh_gateway_config` copies the config it is about to replace to
`nginx.conf.prev`, and **refuses to overwrite at all if that copy fails** — a
config with no way back is precisely the thing being fixed, so failing to keep
one is a failure rather than a warning. `do_rollback` restores it alongside the
image, then consumes it: the backup describes exactly one step back, and after
that step it would describe a state no longer adjacent to the host. A missing
`.prev` is the normal case, not a fault, and says so.

Your point 2 is what decided the shape. `pin <older-digest>` is the script's
own advice when rollback has nowhere to go, so pin genuinely does move
backward, and "pin has already been told to move this deployment forward"
does not survive that.

I also took your alternative *as well as* the backup, in the one branch where
it earns its place: the failed-serving path of `pin` now names the config it
changed and prints the revert command. That is the branch where the config is
a live suspect — Hive healthy, the thing in front of it not — and where the
operator is deciding what to try, rather than after `rollback` has failed to
undo it.

Eight cases added. Verified they have teeth: deleting the `restore_gateway_config`
call from `do_rollback` fails four of them.

Suite is **144 passed, 0 failed**. `shellcheck -x` introduces no new findings
(the one `SC2034` this added is annotated — the variable is consumed inside a
single-quoted condition `check` eval's, the same blind spot behind every
pre-existing `SC2016` in the file).

Not addressed here: `src/docs/podman-quadlet-update-rollback.md` still does not
mention the gateway config, which you noted. Now that rollback *does* carry it,
the doc's omission is no longer a gap in behaviour — but it should probably
gain a line about `.prev`. Happy to add it here or leave it to the doc pass.

PR #6093, opened 2026-09-05
fix(advisory): retire stale findings whose named work has closed

### LINKED ISSUE #6080: advisory digest: findings computed at an older commit are re-emitted as open HIGH without re-verification (one was listed open at the commit that fixed it)
## Summary

The advisory digest carries findings forward from an older commit and presents them as current open `HIGH` items, without re-verifying them against the commit the digest says it analyzed. In the case below, a finding was listed as open at **the exact commit that fixed it**.

The digest footer already anticipates this ("Findings marked ⚠️ were computed at an older commit and have NOT been re-verified here"), but the two findings below are **not** marked ⚠️ — they are counted, severity-ranked, top-10 open items, while their own body text names the older commit they were computed at.

This is a recurrence. `Danathar/atomic-image-builder#126` documented occurrences 1–4 on 2026-08-31 and was closed; that issue was filed against the target repo rather than here, which is probably why it never reached the generator.

## Reproduction

Live digest: https://github.com/Danathar/atomic-image-builder/issues/11#issuecomment-5395760874
Header: `## 🐝 Advisory Digest — 2026-09-05 07:48 EDT` · comment `updated_at` 2026-09-05T11:48:32Z
Footer: *Analyzed at `Danathar/atomic-image-builder@2d64e0c1e166` (branch `main`)*

### Finding 1 — `contrib/aib`: 4 surviving mutants

Listed as open `🟠 HIGH`. Its own body says it was computed at `45e840a7f107` (2026-09-04).

| | |
|---|---|
| Fixed by | PR #213, merged 2026-09-04T13:48:15Z (`ca8c6cc6`) |
| and | PR #214, merged 2026-09-04T13:55:29Z (`2d64e0c1`) |

**PR #214's merge commit `2d64e0c1e1668adfe0c5cbd1cdbfa614e517a824` is the commit the digest footer names as analyzed.** Its title is *"assert the /etc/localtime guard, closing the last surviving mutant in contrib/aib"*.

Verified at that commit: `bash tests/test_contrib_aib.sh` → **29 passed, 0 failed** (the finding quotes 20/20), and the suite now asserts every behaviour the finding lists as unasserted — `--pull=newer`, `--rm`, `-it`, the `/etc/localtime` mount *and* its guard, the `trap cleanup EXIT` temp-file removal, and the `rm -f` on the rpm-ostree failure path.

### Finding 2 — `format_markdown_tables.py` absent from `.coveragerc`

Listed as open `🟠 HIGH`. Its own body says it was computed at `d476116953ab` (2026-09-03T19:34:18Z), and it explicitly names its own remediation: *"Filed issue #208, hold-gated PR #209"*.

| | |
|---|---|
| PR #209 | **MERGED** 2026-09-04T10:06:56Z |
| Issue #208 | **CLOSED** 2026-09-04T10:06:57Z |

Both ~23.5 hours before this digest ran. Verified at the analyzed commit: `format_markdown_tables` is present in `.coveragerc`'s `[run] source` list, and `tests/test_format_markdown_tables.py` exists.

## Why the existing hedge did not catch it

`#126`'s recommendation (a) was to cross-check any linked GitHub issue's state before including a finding. Finding 2 is the clearest possible case for that check — the finding text itself contains the issue and PR numbers, and both were closed/merged at generation time.

The per-finding "computed at `<sha>`" metadata is also already present and already differs from the digest's analyzed commit in both cases, so the generator has everything it needs to know these are stale without re-running any analysis.

## Suggested fix

Before a finding enters the counted/ranked open section, when its computed-at commit ≠ the digest's analyzed commit:

1. If the finding names a GitHub issue or PR, resolve its state — closed/merged means move it to *Recently Resolved*, not *open*.
2. Otherwise re-verify against the analyzed commit's tree, or mark it ⚠️ and demote it out of the counted set.

Right now the ⚠️ mechanism annotates but does not gate, which is the same gap `#126` identified.

## Three smaller things in the same digest

1. **Cross-repo findings land in one repo's advisory issue.** This digest, on `atomic-image-builder`'s issue, includes findings for `arch-bootc`, `aurora-zfs-simple` and `zfs-kinoite-complex`. The top-ranked `HIGH` is an onboarding finding citing `CLAUDE.md`/`AGENTS.md`/`GEMINI.md` "in production / daily-driven ZFS pools" text and a README contrasting against `zfs-aurora-complex` — none of which exists in `atomic-image-builder` (there is no `GEMINI.md` in that repo at all). A reader of that repo's advisory issue cannot tell which findings are theirs.
2. **Malformed cross-reference links.** The advisory items render a second link with a `gh-` prefix fused into the org name, e.g. `[gh-Danathar/arch-bootc#123](https://github.com/gh-Danathar/arch-bootc/issues/123)` — `gh-Danathar` is not a GitHub org, so every one of these is a dead link. Looks like a source-prefix (`gh-`) not being stripped before URL construction.
3. **The digests disagree about where Hive lives.** The 2026-08-24 and both 2026-09-01 digests link `https://github.com/kubestellar/hive`; the 2026-09-05 one links `https://github.com/hivecommons/hive`.

---

Found while working `Danathar/atomic-image-builder#11` — the two stale findings above were the only atomic-image-builder items left in the digest, and both turned out to be already fixed.

Danathar:
## Summary

The advisory digest carries findings forward from older commits and lists them as
current open items. #5130 taught it to *notice* this — `MarkStaleProvenance` sets
`ProvenanceStale` and the renderer captions it — and #2364 made `applyTopN` prefer
a confirmed finding over an unconfirmed one of the same severity. Neither decides
whether the finding is still *true*, deliberately: "nobody re-checked" is not
"fixed".

#6080 is the case where the pipeline can do better than annotate, because **the
finding says where to look**. In the 2026-09-05 digest two findings reached the
counted, severity-ranked top-10 OPEN list. One was listed open at the exact commit
that fixed it. The other named its own remediation in its own body — an issue and
a PR that had closed and merged ~23.5 hours earlier. Both held HIGH slots while
other findings went unshown.

A finding that is provenance-stale **and** names only GitHub work that has since
closed now moves to *Recently Resolved* instead of being listed open.

## The rules, and why each one is there

Every clause is load-bearing and has a test that fails without it:

- **Only provenance-stale findings are eligible.** A finding computed *at* the
  analyzed commit is current evidence whatever it names. Retiring it for
  mentioning a closed issue would silently delete the report of an issue closed
  *too early* — turning this guard into a way to lose findings.
- **Unanimity, not "any closed ref".** Findings routinely name several items — the
  tracking issue, the PR meant to fix it, one for context. "Any closed" would
  retire live work the moment one of them healed.
- **Every way of not knowing keeps the finding open**: no resolver, no repo
  context, a lookup that fails, a reference nobody can read, a closure older than
  30 days. This may move a finding out of the open set only on positive evidence —
  the same posture `sameCommit` already takes in `provenance.go`.

It runs **before** the top-N cap. That is the point: these findings were not merely
mislabelled, they were counted, ranked and holding slots. Retiring them after the
cap would leave the slot spent.

The resolver is an injected seam beside the existing `VerifyPath`, wired in
`runEvalCycle` to the GitHub client already there — so this is live, not a seam
nobody calls. With it nil the behaviour is exactly as before: captioned and
demoted, never retired.

## Also fixed: the dead cross-reference links

Advisory beads carry `ExternalRef` of the form `gh-<owner>/<repo>#<n>`.
`splitInlineRef` splits on the first `/`, so `gh-Danathar` was read as the
**owner** and every cross-repo reference rendered a link to
`github.com/gh-Danathar/...` — not an org, so a dead link every time.

Fixed on both sides: the renderer strips the prefix (which heals beads already
stored that way), and `recordIntentAlignmentAdvisory` stops writing it. The dedup
there now matches **both** forms — without that, the first run after this change
would fail to recognise the existing bead and open a duplicate advisory per eval
tick for drift it already tracks. There is a test for exactly that.

## What I did not do, and why

- **Cross-repo findings landing in one repo's advisory issue** (smaller item 1) is
  a routing question — which repo's digest should carry which findings — and wants
  a maintainer decision rather than a guess from me.
- **The kubestellar/hivecommons URL disagreement** (smaller item 3) is **already
  fixed on `v4`**: the rendering code names `hivecommons` in both places. The older
  digests were produced by older deployed builds, so there is nothing to change
  here.
- I did **not** make staleness demote across severity bands. The existing comment
  in `applyTopN` argues explicitly against it ("an unverified critical may still be
  the worst thing here"), and overturning a recorded design decision is not mine to
  do unilaterally.

## Validation

| Command | Outcome |
| --- | --- |
| `go test ./pkg/advisory/...` | pass (25 new assertions across 12 new tests) |
| `go test ./pkg/github/...` | pass |
| `go test ./cmd/hive/...` | pass |
| `go vet` on all three changed packages | clean |
| `gofmt -l` on the changed files | clean |
| changelog-fragment-guard rules, applied by hand | filename and `- ` shape both match |

CI itself has **not run on this PR**: every workflow sits at `action_required`,
this repo's gating for a fork contributor. `DCO` and `PR Verifier` did run and
pass. The lanes above were reproduced locally instead — `v2 CI` runs the same
`go test` this was checked with.

**The new tests were verified to discriminate.** Each rule was removed from the
implementation and the suite re-run:

| Mutation | Result |
| --- | --- |
| provenance-stale gate dropped | `TestBuildDigestKeepsFreshFindingNamingClosedWork` fails |
| unanimity relaxed to "any closed" | unanimity matrix + `KeepsStaleFindingWithOpenWork` fail |
| "cannot tell" stops blocking | `UnknownRefBlocksEvenWhenOthersClosed` fails |
| `gh-` prefix strip removed | `FormatFindingRefStripsGHSourcePrefix` fails |
| retire moved after the cap | `SettledStaleFindingFreesItsTopNSlot` fails |
| legacy-ref dedup tolerance removed | `DedupesLegacyPrefixedRef` fails |

That third mutation initially **survived** — my "cannot tell" case used a resolver
that failed on *every* ref, so a `continue`-instead-of-`return` bug was masked by a
later guard. The discriminating case is one ref closed and one unreadable, which is
where a still-open item would actually hide. I added it, and it now fails against
that mutation.

### Skipped, stated rather than implied

- **`pkg/agent`, `pkg/config`, `pkg/knowledge`, `pkg/proxy` and `pkg/sandbox` fail
  in my sandbox.** They fail **identically on a clean `upstream/v4` tree** — I
  stashed my changes and re-ran to confirm — so they are environmental (no tmux, no
  writable agent paths), not caused by this change. I did not investigate further.
- **No digest was generated against real GitHub.** The resolver is exercised
  through stubs; nothing here was run against a live advisory issue, so the
  end-to-end behaviour on the reported digest is argued from the code and the
  tests, not observed.
- **`Go Security Analysis` was not reproduced locally** — it needs tooling the
  sandbox could not fetch. The change adds one read-only GitHub API call and no
  new input parsing beyond two regexes over text the package already parses.
- One pre-existing formatting finding (`gofmt -l` flags `pkg/github/mttr.go`) is
  untouched by this PR and left alone.

## External state

None. No images, issues, comments, or workflow runs were created or mutated.

Closes #6080


Danathar:
Both findings addressed in `0da4a7e6`. Both were real, and the first one is worse than a gap — it inverts the feature.

## 1. Retirement decayed — fixed

Confirmed exactly as described, and reproduced before fixing. A finding whose named work closed 45 days ago:

```
--- FAIL: TestSettledStaleFindingStaysRetiredPastTheWindow
    1 findings open 45 days after the work it names closed; retirement must not decay
    TotalCount = 1, want 0
```

You had the mechanism right: nothing closes the underlying bead, the verdict is recomputed from scratch every build, and `staleFindingSettled` measured closure age against `now`. So the same finding, the same closed refs, no new evidence and nothing re-checked, is retired for thirty days and then comes back as a counted, severity-ranked HIGH forever. The digest is correct for a month and then permanently wrong, which is worse than never having had the window.

I took your first suggestion rather than persisting the verdict on the bead, because the window turned out to be answering the wrong question. "Is this finding still open?" depends on the finding and the state of the work it names — both stopped changing when that work closed. Only "is this still worth announcing?" is a question about how long ago that was. `refClosedWindow` now bounds only the Recently Resolved rendering:

- Inside 30 days: identical to before — retired, and announced.
- Past 30 days: still retired, no longer announced. The finding is simply absent rather than resurrected.

One consequence worth naming since it isn't in your report: a retirement can now leave the open set *without* producing a `ResolvedFinding`, so the caller's `if len(settledStale) > 0` recount was no longer sufficient. `partitionSettledStale` returns a retired *count* now, and the header total is recomputed from that. Keying it on the announced list would have left `TotalCount` claiming a finding the digest lists nowhere — the same class of bug as the one being fixed, one layer up.

The subtest that asserted the old expiring behaviour is rewritten to assert the new one rather than deleted, so the case stays covered.

I did **not** take the "closed a year ago is stale context, not a fix" reasoning out of the codebase, because it is still true — I just could not find a stable way to act on it. Anchoring to the finding's own `Timestamp` is the obvious candidate and it breaks the motivating case: a carried-forward finding gets a fresh timestamp while its evidence stays old, so `#208` closing 23.5h *before* today's digest would read as closing before the finding existed. Anchoring to provenance would need a timestamp the `ProvenanceSHA` metadata does not carry. So the age check is gone from the verdict rather than relocated, and if you want that caution back it needs a provenance *time*, which is a separate change.

## 2. Unbounded, uncached lookups — fixed

Also reproduced first: two findings naming the same two refs cost **4 lookups**, not 2.

Answers are memoized per build in a `staleRefResolver`, following `VerifyFindingPaths`' `checked` map — an issue does not open or close while one digest is being assembled. Failed lookups are cached too, which matters more than the successful ones: without it a rate-limited build repeats the same doomed call once per finding naming the ref, i.e. worst behaviour at the worst moment.

On your "consider an overall lookup budget" — I added one (250 distinct issues per build), since memoization alone doesn't bound a digest whose prose names hundreds of *distinct* bare refs. Two properties make it safe: exhaustion returns `ok=false`, which every caller already reads as "leave the finding open", so it can never retire anything by accident; and agents are now walked in sorted order, because otherwise *which* findings got retired at the budget edge would depend on Go's map ordering. Flag it if 250 is the wrong number or if you'd rather not have the knob — it's the discretionary half of this and easy to drop.

I'd single out your framing that this is a correctness problem and not just a cost one: a rate-limited lookup aborts that finding's whole verdict, so retirement failed silently on exactly the digests large enough to need it. That's what made the memoization worth doing properly rather than as an optimisation.

`go build ./...`, `go vet` and `go test ./pkg/advisory/` all clean; the changelog fragment's "or a closure older than 30 days" clause was describing the old rule and has been corrected.

Issue #6085, opened 2026-09-05
🐛 bug: PR Verifier's fatal image pull fails the check on registry access, defeating the `continue-on-error` on the verification itself


hanthor:
## Summary

`pr-verifier.yml` treats fetching the verifier image as fatal while treating the verification it performs as advisory. The result is that the check reports on registry reachability rather than on PR contents, and it is red on every PR in the repository at once when the registry answer changes.

- **What the job does.** Logs into GHCR, pulls a digest-pinned verifier image, and runs it against the PR.
- **What's wrong.** The `Verifier action` step is `continue-on-error: true` — a deliberate decision that a verification finding should not block. But `Login to GHCR` and `Pull verifier image` carry no such guard, so any registry-side problem exits 1 and fails the check *before any PR content is examined*. The one step allowed to have an opinion cannot fail the job; the two steps with no opinion can.
- **Blast radius.** Repository-wide and simultaneous. Not a per-PR defect, so it reads to contributors and to bots as though every open PR broke at once.
- **A PR that fixes this cannot show itself passing.** The workflow runs on `pull_request_target`, so GitHub executes the **base branch's** copy. A fix is only observable after it merges, which is an awkward thing to ask a reviewer to take on faith.

## Observed

`ghcr.io/kubestellar/pr-verifier@sha256:8c2b638…` stopped resolving after the org rename. Every run of the workflow in a `v4` fork failed identically for four days:

```
Logging into ghcr.io...
Login Succeeded!
Run docker pull "$PR_VERIFIER_IMAGE"
Error response from daemon: manifest unknown
##[error]Process completed with exit code 1.
```

Checked directly against GHCR with anonymous pull tokens — the digest is genuinely gone from the old path and present in the new one:

| Reference | Manifest request |
|---|---|
| `ghcr.io/kubestellar/pr-verifier@sha256:8c2b638…` | `403 / DENIED` |
| `ghcr.io/hivecommons/pr-verifier@sha256:8c2b638…` | `200` |

11 consecutive runs, 5 different PRs, all red on the same line — while the verification those PRs actually needed never ran once. An automated fix lane also escalated one of the PRs to `needs-human` on the strength of this failure, which is a second-order cost of a red check that carries no information about the PR.

[...]

PR #6101, opened 2026-09-05
fix(ci): make runner labels fork-portable via HIVE_RUNNER_LABELS

### LINKED ISSUE #6083: 🐛 bug: hard-coded `[self-hosted, hive]` runner labels make CI unrunnable in a fork — silently, and the stuck runs cannot be cancelled
## Summary

A fork of this repository cannot run most of its own CI, and gets no error saying so.

- **What the labels are.** 24 `runs-on:` declarations across 12 workflows target runners labelled `[self-hosted, hive]` — a fleet only `hivecommons` has.
- **What's wrong.** In a fork, those jobs never start. They do not fail: they sit in `queued` forever. A red X is a bug report; a job that never runs is invisible, and the PR simply never reaches a conclusion.
- **It is worse than "those checks don't run".** A run stuck this way holds its workflow's **concurrency group**, so the *next* push's run of the same workflow is stuck behind it too. And a job queued for a label no runner carries **cannot be cancelled** — there is no runner to deliver the cancellation to — so the wedge does not clear by itself or by operator action short of deleting the run.
- **Blast radius.** Every fork, and the effect is permanent rather than transient. This is a fork of `v4` at `3edb845`; the same shape would hit anyone standing up their own hive from a fork of this repo.

## The labels

```
$ git grep -h "runs-on:.*self-hosted" -- .github/workflows/ | wc -l
24
$ git grep -l "runs-on:.*self-hosted" -- .github/workflows/ | wc -l
12
```

`v2-tests.yml` (9), `v2-ci.yml` (6), `backend-smoke.yml` (4), `go-security-analysis.yml` (3), and one each in `testutil-guard.yml`, `suid-contract.yml`, `release-line-guard.yml`, `podman-rootful-lane.yml`, `podman-contract.yml`, `image-attestation-guard.yml`, `docs-link-check.yml`, `changelog-fragment-guard.yml`.

Three spellings are in use. The most common one is already fork-aware:

```yaml
runs-on: ${{ (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) && fromJSON('["self-hosted","hive"]') || 'ubuntu-latest' }}
```

**That expression covers the wrong fork case.** It degrades to `ubuntu-latest` for a pull request *from* a fork *into this repository* — a contributor's PR, where the concern is running untrusted code on the fleet. It does not help a **fork of the repository**: there, `head.repo.full_name == github.repository` is true for every branch its own maintainers push, so the condition selects the self-hosted fleet that the fork does not have. The case that is protected is the one where the runners exist; the case that breaks is the one where they don't.

The other two spellings have no fork path at all:

```yaml
runs-on: [self-hosted, hive]                                                        # v2-ci.yml:828, :1039, backend-smoke.yml:217
runs-on: ${{ matrix.lane == 'pinned' && 'ubuntu-latest' || fromJSON('["self-hosted","hive"]') }}   # backend-smoke.yml:95
```

## Observed

On a fork at `v4` (`tuna-os/hive`), opening one PR produced 26 workflow runs. **14 were still `queued` an hour later**, having never started. The same jobs ran green on `ubuntu-latest` on this fork three days earlier, before it took upstream's newer workflows.

Drilling into one, the job is waiting on a label nothing carries:

```json
{"name":"build-and-test","status":"queued","labels":["self-hosted","hive"],
 "created_at":"2026-09-05T10:13:45Z","started_at":"2026-09-05T10:13:45Z"}
```

### The concurrency wedge

`v2-tests.yml:80-82`:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.run_id }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

After pushing a fix and re-running, the new `v2 Tests` run sat in status `pending` with **zero jobs created** — blocked by the older run still holding the group. `cancel-in-progress` did not clear it, because the stale run was never *in progress*; it was `queued`.

### The stuck runs cannot be cancelled

Cancelling the blocker was accepted and did nothing:

```
POST /repos/.../actions/runs/33960079577/cancel  →  202 Accepted   (12:07:02Z)
GET  /repos/.../actions/runs/33960079577         →  "status": "queued"   (12:29Z, 22 min later)
```

The successor is still `pending`, zero jobs, `updated_at` unchanged since it was created. So a fork that takes these workflows acquires a permanently wedged workflow that it cannot clear through the API or the UI's cancel button.

## Suggested fix

Make the label set a repository variable with a GitHub-hosted default, so upstream keeps its fleet and a fork degrades to something that exists:

```yaml
runs-on: ${{ vars.HIVE_RUNNER_LABELS && fromJSON(vars.HIVE_RUNNER_LABELS) || 'ubuntu-latest' }}
```

`hivecommons` sets `HIVE_RUNNER_LABELS` to `["self-hosted","hive"]` once at the org or repo level and nothing changes; a fork sets nothing and its CI runs. This also subsumes the existing fork-aware expression, since a contributor PR into this repo can keep its current guard by ANDing the two conditions.

Two smaller points regardless of the above:

1. The four `runs-on: [self-hosted, hive]` / per-lane occurrences should get whatever the answer is too — right now the fork-awareness is inconsistent between jobs in the same workflow file.
2. A job that can never be scheduled is worth failing loudly rather than queueing forever. A short-timeout preflight job on `ubuntu-latest` that asserts the fleet is reachable would convert a silent permanent wedge into a legible red X, upstream included — a fleet outage looks exactly the same as a missing fleet today.

## Workaround, for anyone hitting this

Point them all back at `ubuntu-latest`. Collapsing all three spellings is mechanical, and only `runs-on:` values need to change — no job logic:

```
$ git grep -h "runs-on:" -- .github/workflows/ | sort | uniq -c
     54 runs-on: ubuntu-latest
      3 runs-on: ${{ matrix.runner }}
      1 runs-on: ubuntu-24.04-arm
```

All 36 workflows still parse, and the previously-queued jobs then run and pass. Already-wedged runs still need clearing separately, per above.

## Environment

- Fork of `hivecommons/hive` at `v4`, tree identical to `3edb845`
- GitHub-hosted runners only; no self-hosted fleet registered
- Evidence from `tuna-os/hive#30`

Danathar:
> ## ⚠️ One operator step, before merge
>
> ```sh
> gh variable set HIVE_RUNNER_LABELS --org hivecommons --body '["self-hosted","hive"]'
> ```
>
> With the variable unset, every job that used to run on the fleet runs on
> `ubuntu-latest` instead. Nothing breaks silently — the jobs run and report on
> their merits — but the heavy lanes lose the fleet until it exists. Setting it
> first makes this PR a no-op for upstream.

## Summary

24 `runs-on:` declarations across 12 workflows named the `[self-hosted, hive]`
fleet directly. Those labels only exist in `hivecommons`, so a fork cannot run
most of its own CI — and gets no error saying so.

The failure is worse than "those checks don't run":

- **The jobs never start.** They sit in `queued` indefinitely. A red X is a bug
  report; a job that never runs reports nothing, and the PR never reaches a
  conclusion.
- **They wedge the workflow.** A queued run holds its concurrency group, so the
  next push queues behind it. `cancel-in-progress` doesn't clear it — the
  blocker was never *in progress*.
- **They cannot be cancelled.** There is no runner to deliver the cancellation
  to. Measured in the issue: `202 Accepted`, still `queued` 22 minutes later.

Measured on a fork of `v4`: one PR produced 26 runs, **14 still `queued` an hour
later**.

## The change

One uniform substitution, 24 times:

```diff
-fromJSON('["self-hosted","hive"]')
+fromJSON(vars.HIVE_RUNNER_LABELS || '["ubuntu-latest"]')
```

plus rewriting the three bare `runs-on: [self-hosted, hive]` lines to the same
expression. No job logic changes.

### Why the fallback goes *inside* `fromJSON`'s argument

The form the issue suggests —

```yaml
${{ vars.HIVE_RUNNER_LABELS && fromJSON(vars.HIVE_RUNNER_LABELS) || 'ubuntu-latest' }}
```

— only works if GitHub short-circuits `&&` before evaluating `fromJSON('')`,
which is otherwise **an error rather than a fallback**. That's true today, and
it is an undocumented detail to rest fleet-wide CI on. Putting the default
inside the argument means `fromJSON` is handed valid JSON whether the variable
is set or not, and the expression stops depending on evaluation order at all.

### What I did *not* change, and why

Each job keeps the fork-PR guard it already had. The 20 guarded sites stay
guarded; the three bare ones and the per-lane one gain the *variable* without
gaining a *guard* they never had.

The issue's point 1 asks for consistency there, and applying the variable
everywhere satisfies it in the sense that matters — every job is now
fork-portable. But unifying the **contributor-PR guard** is a different axis: it
would move three jobs off the fleet for PRs from forks *into* this repo. That's
arguably a security improvement and it's also a behaviour change for upstream on
jobs that may need fleet resources. It's a maintainer policy call, so I've
flagged it rather than folded it in.

## Regression guard

`src/deploy/test_runner_labels_portable.sh` fails CI if any `runs-on:` names the
fleet directly, or uses the variable without the inner fallback. Wired into
`v2-ci.yml` next to its siblings, so `test_deploy_suites_wired.sh` passes.

This guard exists because the labels were *already* fixed once: the fork-guarded
spelling was added deliberately in #5855/#5862–#5868, and three jobs in the same
files were still left bare. Hand-editing 24 sites produces exactly that
inconsistency, so it's checked rather than remembered.

## Validation

| Command | Outcome |
| --- | --- |
| `bash src/deploy/test_runner_labels_portable.sh` | 3 passed, 0 failed |
| `bash src/deploy/test_deploy_suites_wired.sh` | 36 suites, **0 orphaned** |
| `python3 src/scripts/check-docs-links.py` | 136 files, all links resolve |
| YAML parse of all 36 workflows | all parse, all keep a `jobs` key |
| Structural assert: no job names the fleet without the variable | clean |
| `shellcheck` on the new guard | clean |
| changelog-fragment-guard rules, applied by hand | filename and `- ` shape match |

**The guard was verified to discriminate:**

| Mutation | Result |
| --- | --- |
| one bare `[self-hosted, hive]` restored (the pre-fix state) | FAIL, naming the file and job |
| one `fromJSON(vars.HIVE_RUNNER_LABELS)` with no inner fallback | FAIL, naming the file and job |
| a job whose `runs-on` is a YAML **block sequence** of bare labels | FAIL, naming the file and job |

**The second mutation initially passed** — worth stating, because it's the more
dangerous of the two. My first version checked only that `ubuntu-latest` appeared
somewhere on the line, and the fork-guarded spelling's own trailing
`|| 'ubuntu-latest'` satisfies that. So it masked exactly the bare-`fromJSON`
form that *errors* in a fork rather than degrading. The check now requires the
fallback inside the argument, and the reasoning is recorded in the guard next to
the test.

**A second gap was found in review and fixed in `b186648c`.** The guard was
grepping `runs-on:` line by line, so it saw the flow form `runs-on:
[self-hosted, hive]` and was blind to the block form spelled over three lines —
the equally natural YAML for the same thing. Against a workflow written that
way the original guard reported `3 passed, 0 failed`, exit 0. It now parses the
workflows with PyYAML and walks each job's `runs-on`, normalising the string,
flow-list, block-list and `{group:, labels:}` forms through one function. Two
deliberate choices: a missing PyYAML is a **failure, not a skip** — this guard's
entire subject is that a check which silently fails to run is worse than no
check — and failure output now names the job, which the line-based version had
no way to know.

### Skipped, stated rather than implied

- **Fork evidence now exists, and it is partial.** I originally wrote that
  nothing here had been run against a real fork or a real fleet. @hanthor, who
  reported #6083, has since supplied the fork half from `tuna-os/hive#30`, and
  it corrects this PR in a useful direction: the podman lanes and `v2 CI`
  complete on hosted runners, so the degraded mode is a real signal rather than
  a broken one, and setting the org variable is strongly *preferred* rather than
  a hard prerequisite. My earlier "the heavy lanes lose the fleet until it
  exists" warning was reasoning from the label change, not from a run.
- **`v2 Tests` remains unvalidated**, on @hanthor's own caveat and for the
  second-order reason in #6083 itself: its predecessor was still `queued` and
  holding the concurrency group, so the successor never created a job. The green
  column should not be read as covering it.
- **Still unobserved from here:** whether a real fleet picks jobs back up once
  the variable is set. That is only visible on merge.
- CI has since run on this PR and is green; the earlier `action_required` state
  (this repo's gating for a fork contributor) has cleared.
- **Runs already wedged are not cleared by this**, because they can't be
  cancelled. The doc carries the `gh api -X DELETE .../actions/runs/<id>` step
  and how to list them first.

## Not done, flagged rather than guessed

The issue's point 2: a job that can never be scheduled queues rather than
failing, **upstream included** — so a fleet *outage* looks exactly like a missing
fleet. A short-timeout preflight job on a GitHub-hosted runner would turn that
into a legible red X. I haven't implemented it: it needs a decision about whether
every workflow pays for an extra job, and guessing the shape is worse than
raising it. It's recorded as open in `src/docs/ci-runner-labels.md`.

## External state

None. No variables, runners, or workflow runs were created or mutated — the
`gh variable set` above is for a maintainer to run.

Closes #6083


hanthor:
Reporter of #6083 here. You flag that "nothing was run against a real fork or a real fleet" — I have the fork half of that, so here it is as data rather than a review.

`tuna-os/hive#30` took upstream's tree wholesale and then pointed all 24 declarations at `ubuntu-latest`, which is behaviourally what this PR produces with `HIVE_RUNNER_LABELS` **unset**. On that head, 21 workflow runs completed: **19 green or skipped, 2 red**, and both red are `PR Verifier` — the base-branch `pull_request_target` failure of #6085, unrelated to runners.

The lanes that had been queued indefinitely all ran and passed on GitHub-hosted runners:

| Lane | On `ubuntu-latest` |
| --- | --- |
| `v2 CI` | ✅ |
| `Podman arm64 Lane` | ✅ |
| `Podman Rootless Lane` | ✅ |
| `Podman Rootful Lane` | ✅ |
| `Go Security Analysis` | ✅ |
| `Podman Rootless Contract`, `SUID + NET_ADMIN Contract`, `Image Attestation Guard`, `Docs Link Check`, `Dashboard Lint`, `changelog-fragment-guard`, `Release Line Guard`, `Quadlet Generator Gate`, `Build and Push Docker Image` | ✅ |

That bears directly on your pre-merge warning that "the heavy lanes lose the fleet until it exists": empirically they do not lose the *ability to run* — the podman lanes and `v2 CI` complete on hosted runners. Slower and less capacious than the fleet, certainly, but the degraded mode is a real signal rather than a broken one. That seems worth knowing when deciding whether setting the org variable is a hard prerequisite or merely strongly preferred.

**One lane I cannot vouch for: `v2 Tests`.** It never ran on my fork, for the second-order reason in #6083 — its predecessor run was still `queued` for the fleet and holding the per-PR concurrency group, so the successor sat at `pending` with zero jobs created. So I have no evidence either way about `v2 Tests` on hosted runners, and I would not read my green column as covering it.

I can also independently confirm the uncancellable finding from the other direction: `POST .../cancel` returned `202 Accepted` and the run was still `queued` 22 minutes later; it was still `queued` **two hours** later. Your `gh api -X DELETE` note in the doc is, as far as I can tell, the only thing that actually clears them.

Two smaller notes:

- Your inner-fallback correction to my suggested expression is right and my version was fragile. `vars.X && fromJSON(vars.X) || '...'` leans on `&&` short-circuiting before `fromJSON('')` is evaluated; putting the default inside the argument removes the dependency on evaluation order entirely. Worth the paragraph you gave it.
- Declining to unify the contributor-PR guard is the right call for this PR — moving three jobs off the fleet for fork PRs is a security *and* capacity decision, and it is genuinely separate from fork portability.

---
_Generated by [Claude Code](https://claude.ai/code)_

Danathar:
## @hanthor — thank you, and it changes the PR's own claim

That is much better evidence than anything I had, and it contradicts me in a
useful direction. My "the heavy lanes lose the fleet until it exists" warning
was reasoning from the label change, not from a run; your `tuna-os/hive#30`
numbers show the podman lanes and `v2 CI` completing on hosted runners. So the
degraded mode is a real signal rather than a broken one, and setting the org
variable is strongly preferred rather than a hard prerequisite.

I have taken your caveat as stated: **`v2 Tests` is unvalidated**, for the
second-order reason in #6083 itself — its predecessor was still `queued` and
holding the concurrency group, so the successor never created a job. I am not
reading your green column as covering it, and I will say so in the PR
description rather than letting the overall shape imply otherwise.

The independent confirmation of the uncancellable behaviour is worth a lot too:
`202 Accepted` and still `queued` two hours later is a stronger statement of
the wedge than my 22-minute observation.

## Scanner finding — fixed in `b186648c`

Confirmed, and reproduced both directions before and after.

The detector grepped `runs-on:.*self-hosted` line by line, so it saw
`runs-on: [self-hosted, hive]` and was blind to

```yaml
runs-on:
  - self-hosted
  - hive
```

Against a workflow spelled that way, the **old guard reports `3 passed, 0
failed`, exit 0**; the new one fails it. Exactly your reproduction.

It now parses with PyYAML and walks each job's `runs-on`, normalising string,
flow list, block list and the `{group:, labels:}` form through one function.
The *written* form is what gets tested rather than the resolved labels, since
an expression going through `vars.HIVE_RUNNER_LABELS` is precisely what the
check wants to find — it is the indirection, not the label set, that makes a
bare label safe.

Two deliberate departures from a plain port:

- **PyYAML is required, not skipped.** This guard's whole subject is that a
  check which fails to run is invisible. Skipping and reporting PASS would
  reproduce that fault inside the fix for it, so a missing interpreter is a
  FAIL. An unparseable workflow is reported for the same reason rather than
  passing quietly.
- Failure output now carries the **job name**, which the line-based version
  had no way to know.

The 36 real workflows still pass; `shellcheck -x` is clean.

PR #6105, opened 2026-09-05
🐛 fix: pass -- to tmux send-keys so a kick chunk starting with "-" is not dropped

gregoryhunt:
## Summary

- `SendKick` types a kick into the agent's pane as 400-rune slices via `tmux send-keys -l <slice>`. tmux parses `send-keys` arguments with getopt, so a slice that begins with `-` is read as a flag, the command fails with `unknown flag`, and because the `Run()` error was discarded, that 400-character slice silently never reached the agent.
- Any chunk boundary landing just before a hyphen triggers it: `onboard-ai-platform`, `hold-gated`, `[ONB-2651]`, a markdown bullet. With ~46 slices per kick it is intermittent rather than constant, which is why some kicks worked and others arrived corrupted.
- Observed on a live hive (2026-09-05, quality agent on a `claude` backend): the held-PR preflight snapshot arrived as `onboard-ai` immediately followed by text from the next slice. The list "ended mid-entry", the policy told the agent to stand down, and it did on two consecutive kicks, after spending the preflight tokens.
- Fix: `tmuxTerminal.SendLiteral` passes `--` before the text, and a failed send is logged at WARN (agent, session, slice length, error) instead of being dropped.

Reproduction outside hive:

```
$ tmux send-keys -t probe -l '-platform#1488 hello'
command send-keys: unknown flag -p        # exit 1, nothing typed
$ tmux send-keys -t probe -l -- '-platform#1488 hello'
                                          # exit 0, text arrives intact
```

## Related issues

None filed; this PR carries the diagnosis.

## Testing

- [x] `cd src && go test ./pkg/agent/` (full package, 263s, green)
- [x] New `TestTmuxTerminal_SendLiteral_TypesTextStartingWithDash` types exactly the lost production slice through the real tmux-backed `TerminalSession` and asserts it lands in the pane. Verified red without the fix (stashed `terminal.go`), green with it.
- [ ] Other / not run (explain): full `go test ./...` not run; the change is confined to `pkg/agent/terminal.go`.

## Contributor checklist

- [x] PR targets `v4` per CONTRIBUTING.md (the template's `v2` line is stale).
- [x] Docs, examples, and policies are updated when behavior changes (none affected).
- [x] `changelog.d/fixed-send-keys-literal-dash.md` carries the changelog entry.


clubanderson:
Reviewed, correct: pinned tmux 3.5a honors `--`, and `SendLiteral` is the only Go path sending free-form text. Nit: `terminal.go` uses `.Run()`, so the WARN line only says `exit status 1`; `CombinedOutput()` would carry tmux's stderr, which is the stated diagnosability goal. Not blocking. Will lgtm on green CI.

Issue #4044, opened 2026-08-18
gh wrapper: author-gated list is fail-closed for ALL staff agents — gh api user is 403 for App installation tokens (#3982 regression)


clubanderson:
## Summary

Since #3982, the gh wrapper resolves the caller's GitHub identity **exclusively** via `gh api user` (fail-closed, env vars no longer trusted). But every staff-hive agent authenticates with a **GitHub App installation token** (per-agent tier token from `HIVE_AGENT_TOKEN_CACHE`), and `GET /user` is categorically unavailable to installation tokens — GitHub returns 403 `Resource not accessible by integration` for any server-to-server token, regardless of granted permissions. So `_resolve_self_login` can never succeed for any staff agent, and every author-gated listing fails closed.

## Exact failing commands + errors (reproduced live as an agent UID through the installed wrapper, with a valid freshly-minted tier token)

```
$ gh api user --jq .login
gh: Resource not accessible by integration (HTTP 403)
exit code 1

$ gh pr list --repo <owner>/<repo> --author "<app-slug>[bot]"
⛔ BLOCKED: gh pr list --author requires authenticated GitHub identity.
Could not resolve the current token identity with 'gh api user --jq .login'.
exit code 1
```

`--author @me` is passed through to gh, but @me also cannot resolve for an installation token server-side, so there is no working self-listing form at all for staff agents.

## Responsible change

#3982 (author-gate hardening): removing trust in `HIVE_BOT_LOGIN`-style env vars was right (agent-controlled env must not seed a trusted identity), but the replacement oracle (`gh api user`) only works for user tokens (contributor mode). The staff path was not exercised: the gates test harness stubs gh with an exit-0 stub, so `_resolve_self_login` "succeeded" in tests where it can never succeed in production.

## Scope

Fleet-wide, all v4/edge spokes, all non-contributor agents, unconditional: every `gh issue list --author ...` / `gh pr list --author ...` is blocked. Practical effect: agents that need to find their own open PRs/issues fall back to raw API calls outside the wrapper (see companion security issue on `GITHUB_TOKEN` in agent env) — the fail-closed gate is actively pushing agents toward the bypass.

## Proposed fix (keeps fail-closed, restores a trusted oracle)

[...]

Issue #4043, opened 2026-08-18
gh wrapper: pr/issue edit fails closed when the injected provenance label is missing; ensure-labels cache is repo-blind


clubanderson:
## Summary

Staff-hive agents cannot run `gh pr edit` / `gh issue edit` on most repos: the gh wrapper unconditionally appends `--add-label "agent/<name>,hive/<id>"` to the edit, and when either label does not exist on the target repo, gh fails the **entire edit** with `'<label>' not found`. Agents read this as "permissions issues". Reported live by a fleet owner whose hive was "dead in the water" on the edit lane.

## Exact failing command + error (reproduced live on an affected spoke, as the agent's UID through the installed wrapper)

```
$ gh pr edit 315 --repo <owner>/<repo> --title "docs: repair design plan template link"
'hive/hosted-available-oke-13-placeholder-2aug' not found
exit code 1
```

The edit is never applied. The agent's own account of it: "The CLI wrapper again rejected the edit because of its unavailable injected label."

## Mechanism (three interacting defects in `bin/gh-wrapper.sh`)

1. **The edit arm has no fallback.** `issue/edit|pr/edit` does `exec "$REAL_GH" "$@" --add-label "$LABELS_CSV"` — one shot, fail-closed on a missing label. The `issue/create|pr/create` arm already has the right pattern (retry without labels on failure); edit never got it. Label injection is provenance metadata, not a security gate — its absence must not block the operation.

2. **`_ensure_labels` caches globally, not per-repo.** The guard file `/tmp/.hive-labels-ensured` is one flag for the whole pod session, across all repos and all agents sharing /tmp. Whichever repo an agent touches first gets `agent/<name>` + `hive/<id>` created; every other repo is then permanently skipped for the life of the pod. Verified live: 4 of the owner's repos carry the current `hive/<id>` label (first-touched in some session), 2 do not — and those 2 are exactly where every edit fails. The cache is also touched even when the `gh label create` calls fail (`|| true` swallows, then unconditional `touch`), so one transient failure poisons the session.

3. **Aggravator: hosted migration changed `HIVE_ID`.** Hosted spokes now inject `hive/<hosted-placeholder-id>` while repos only carry the pre-migration `hive/<old-name>` label, so the missing-label case is the common case, not the corner case.

Additional noise finding fixed in the same PR: every wrapper invocation prints
```
/usr/local/bin/gh: line 78: /var/run/hive-metrics/token-access.jsonl: Permission denied
```
[...]

PR #6106, opened 2026-09-05
🐛 fix: silence the credential helper's audit-log append and pre-create the log so it records

gregoryhunt:
## Summary

- Every agent clone, fetch, and push printed `git-credential-hive.sh: line 157: /var/run/hive-metrics/token-access.jsonl: Permission denied` into the agent's pane. The helper appends an audit line with `>> "$LOG" 2>/dev/null || true`, but `2>/dev/null` only mutes the printf: when the redirection itself fails, the shell reports it on its own stderr before the command runs. `gh-wrapper.sh` had this exact bug and fixed it with a group wrap in #4043; the credential helper never got the same treatment.
- The redirection always fails on a per-UID hive: `/var/run/hive-metrics` is `0755 dev:node` by design (#4044, protecting the bot-identity file under `agent-tokens/`) and nothing ever created `token-access.jsonl`, so no agent UID can open it. Consequence: the audit log that both writers feed, and that `GET /api/token-access` reads, has been empty on every per-UID hive.
- `bin/git-credential-hive.sh`: wrap the append in `{ ...; } 2>/dev/null || true`, matching `gh-wrapper.sh`.
- `src/deploy/entrypoint.sh`: pre-create the log as `dev:node` mode `0664` next to the existing hive-metrics setup so appends from both writers land. The directory stays `0755`; only this one file becomes group-writable.

Observed on a live standalone hive (2026-09-05): the line appeared on every `gh repo clone` and `git push` from the quality agent, and `/var/run/hive-metrics/token-access.jsonl` did not exist while `/api/token-access` reported `no audit log`.

## Related issues

Follows the pattern from #4043; respects the directory mode rationale in #4044.

## Testing

- [x] `bash bin/test_git_credential_hive.sh` — 16 passed. The new assertion runs a `get` and fails if `token-access.jsonl`, `Permission denied`, or `No such file` appears in the combined output. The harness has no `/var/run/hive-metrics`, so it reproduces the failing redirection exactly: red on the old helper (15 passed, 1 failed), green on the new one.
- [x] `bash -n` on both shell files.
- [ ] `cd src && go build ./...` / `go test ./...` — not run; no Go changes.
- [ ] Other / not run (explain): entrypoint change verified by reading only; it mirrors the adjacent `mkdir`/`chown`/`chmod` lines and is `|| true` guarded like them.

## Contributor checklist

- [x] Docs, examples, and policies are updated when behavior changes (none affected).
- [x] `changelog.d/fixed-credential-helper-audit-log.md` carries the changelog entry.


clubanderson:
Reviewed, looks right. One request: the pre-created log in `entrypoint.sh` is `chmod 664`, which is world-readable. Every legitimate appender is already in the `node` group, and the dashboard's `handleTokenAccess` owner-gates this file precisely because it records every gh invocation with args. `660` costs nothing. Also worth the one-line "operational visibility, not tamper-evident" note the scanner asked for, since the file is group-writable. Not blocking; will lgtm on green CI.

PR #6122, opened 2026-09-05
🐛 fix: derive the agents' git commit identity from this hive's App instead of hardcoding kubestellar-hive

gregoryhunt:
## Summary

- `/etc/gitconfig` (root phase) and `~dev/.gitconfig` (dev phase) both hardcoded `kubestellar-hive <hive-bot@kubestellar.io>` as the git identity for every agent on every hive. On a self-hosted hive with its own GitHub App that names a bot which does not exist for it: commits are unattributed on GitHub, rulesets that require attributed changes block them, and each agent improvises a different fix per kick. Observed on one hive across three consecutive PRs from the same agent: one commit set the App login by hand, the next was `hive-quality <quality@hive.local>`, the next the hardcoded default — while PR authorship (`hive-open-pr`) was already the App the whole time.
- Adds `hive_git_bot_identity`, defined beside `hive_ghe_git_host` so both boot phases share one answer, deriving the identity the way `Config.EffectiveAIAuthor()` already does: `project.ai_author` when set; else, unless `github.app_authored_prs` is `false`, `<github.app_slug or kubestellar-hive>[bot]` — only for a usable App (real `app_id`, not the `999999999` placeholder, with an `installation_id`), mirroring `GitHubConfig.HasUsableApp()`; else the legacy pair, so hives that never opted in are unchanged.
- Email is `<login>@users.noreply.github.com`, the address GitHub attributes to a bot login. Measured: a commit authored `onboard-ai-hive-bot[bot] <onboard-ai-hive-bot[bot]@users.noreply.github.com>` resolves to `author_login: onboard-ai-hive-bot[bot]`; the legacy pair resolves to `null`.
- Logins are restricted to `[A-Za-z0-9._-]` plus a literal `[bot]` suffix; anything else falls back to the legacy pair rather than reaching the gitconfig unquoted, since every git invocation in the container reads that file.

## Related issues

Extends the #5343 system-gitconfig work; complements `EffectiveAIAuthor()` / `app_authored_prs` (App-bot PR authorship), which never covered commit authorship.

## Testing

- [x] `bash src/deploy/test_entrypoint_system_gitconfig.sh` — 18 passed. The suite now extracts `hive_git_bot_identity` alongside the writer and runs the REAL writer for seven identity cases (ai_author wins; usable App with slug; usable App without slug; App not installed; placeholder App id; `app_authored_prs: false`; unsafe ai_author), reading each result back through `git config` with an empty `$HOME` exactly as an agent UID would. It also asserts the dev-phase `--global` identity is derived, not a second hardcoded pair.
- [x] `bash -n src/deploy/entrypoint.sh`.
- [ ] `cd src && go build ./...` / `go test ./...` — not run; no Go changes.
- [ ] Other / not run (explain): not booted in a container; the change is confined to the two gitconfig writers and the new pure function, which the test exercises verbatim.

## Contributor checklist

- [x] Docs, examples, and policies are updated when behavior changes (no doc names the old identity).
- [x] `changelog.d/fixed-git-identity-app-bot.md` carries the changelog entry.


clubanderson:
Reviewed, agree with the approach and the tests. One note for a follow-up: on a GHE hive without an explicit `app_slug`, the shell fallback is the literal `kubestellar-hive` slug and the email domain is hardcoded `users.noreply.github.com`, while the Go side resolves through `forgeIdentities` to the enterprise slug and `hive_ghe_git_host` is defined right above. Not worse than today, just not the mirror the comment claims. Will lgtm on green CI.

PR #6123, opened 2026-09-05
✨ feature: github.app_signed_commits — re-author agent branches as GitHub-signed App-bot commits before opening the PR

gregoryhunt:
## Summary

- Agents commit with plain git in their pane and push over the credential helper. Those commits cannot be signed — a GitHub App has no account to hold a GPG or SSH key — so a base branch with a `required_signatures` ruleset blocks every agent PR from merging no matter who approves it. Observed on a live hive: eight open, approved, green PRs, all at `mergeable_state: blocked`.
- GitHub signs commits it creates itself: commits authored through the `createCommitOnBranch` GraphQL mutation "are automatically GPG signed and are marked as verified", and "GitHub Apps can use the mutation to author commits directly" ([GitHub changelog, 2021-09-13](https://github.blog/changelog/2021-09-13-a-simpler-api-for-authoring-commits/)).
- New opt-in `github.app_signed_commits: true`. The PR-request watcher — the one server-side choke point every agent PR passes through, holding the App installation token — re-authors the head branch after every existing gate and before `CreatePR`:
  1. `compare base...head` for the changed files and the agents' commits;
  2. a scratch ref `refs/heads/hive-signed/<head>` at the merge base, and one `createCommitOnBranch` on it carrying every addition/deletion and the agents' original messages (DCO trailers included) as the message;
  3. force-update `head` to the new commit, delete the scratch ref.
- Result: one commit, GPG-signed by GitHub and shown as **Verified**, authored by `<slug>[bot]`, whose tree is byte-for-byte the tree the agent pushed — so the claim/content/outreach gates judged exactly what lands, and `CreatePR`'s duplicate-tree guard (#5111) still matches.
- Never blocks a PR. Changes the mutation cannot express (executable bits, symlinks, submodules — checked against the recursive head tree), changes over 20 MiB or past the compare API's 300-file list, a head that moved between compare and update (re-read and refused rather than dropping the agent's newer work), a head that already has an open PR, and any API failure all leave the branch exactly as pushed; the scratch ref is the only thing touched until the final update. The PR opens on the agent's own commits with the reason in the request's `.result.json` (`signed_skipped`) and the hive log at WARN; success reports the oid (`signed_commit`).
- Wiring: `GitHubConfig.AppSignedCommits` (`*bool`, nil == off) + `AppSignedCommitsEnabled()`; `Client.SetSignedCommits(func() bool)` so a config reload is honoured on the next request; installed in `main.go` beside the hold-label seam. `github-app-setup.md` documents the flag, the fallbacks, and that later pushes to an open PR are plain git again.

Pairs with #6122 (commit identity from the hive's own App), which covers the fallback path: when this step has to skip, the plain-git commit is at least attributed to the App bot.

## Related issues

Companion to #6122. No tracking issue filed; this PR carries the diagnosis.

## Testing

- [x] `cd src && go test ./pkg/github/` — full package green (45s). New tests in `pr_request_signed_test.go` drive the real watcher against an `httptest` GitHub double serving compare, refs, the recursive tree, raw blobs, and `/graphql`, and assert: the mutation's branch (`hive-signed/<head>`), `expectedHeadOid` (merge base), headline/body (oldest subject; bodies, later messages, and DCO trailers preserved), additions with exact base64 contents for added/modified/renamed files, deletions for removed files and rename sources; the forced ref update to the returned oid; mutation → update → PR ordering; scratch-ref cleanup; the executable-bit fallback (no mutation, no ref write, PR still opens, reason reported); the moved-head refusal; and that the toggle off or unset performs no mutation and no ref write.
- [x] `go vet ./pkg/github/`, `gofmt`.
- [ ] `go test ./...` — not run in full. `./pkg/config` has one pre-existing failure on macOS unrelated to this change (`TestEntrypointHardensRuntimeConfigItRecreates`, also fails on pristine `v4` `fe34da51`).
- [ ] Other / not run (explain): not exercised against api.github.com from a live hive yet; the GraphQL request shape follows the documented `CreateCommitOnBranchInput` (`branch`, `expectedHeadOid`, `message{headline,body}`, `fileChanges{additions[{path,contents}],deletions[{path}]}`).

## Contributor checklist

- [x] Docs, examples, and policies are updated when behavior changes (`src/docs/github-app-setup.md`).
- [x] `changelog.d/added-app-signed-commits.md` carries the changelog entry.


clubanderson:
Reviewed. One blocker plus a few smaller items before lgtm.

**Blocker: stale compare vs moved head in `signBranch`.** `CompareCommits(base, head)` runs by ref name, then `GetRef("heads/"+head)` sets `headSHA`. A push landing between those two calls makes `headSHA` the new tip while `comparison.Files/Commits` describe the old one. The moved-head guard compares against `headSHA`, passes, and `UpdateRef(force=true)` rewrites the branch from the stale file list, dropping the agent's newest commits. The scanner raised this on 09-05. Simplest fix: bail unless `headSHA == comparison.Commits[len-1].GetSHA()`, or re-run the compare pinned to `headSHA`.

Smaller:
- `reauthorBranchSigned`: `if existing, err := findOpenPRForHead(...); err == nil && existing != nil` swallows a transient API error and proceeds to force-update a branch that may already have an open PR. Skip on `err != nil`.
- Moved-head error path slices `current.GetObject().GetSHA()[:7]` unguarded; an empty SHA panics on the watcher goroutine.
- After the force-update the agent's local branch diverges. Its next push is rejected and a reflexive `push --force` reinstates the unsigned commits. Please add a line to `src/docs/hive-open-pr.md` (and ideally a hint in the result) telling the agent to `git fetch && git reset --hard origin/<head>`.
- The scratch `hive-signed/<head>` ref plus the head force-update emit 2-3 push events. Repos with `push: branches: ['**']` (hive's own docker.yml) will build extra images per agent PR. Consider excluding `hive-signed/**` or documenting the cost.
- Ordering: please land #6122 first. The preserved `Signed-off-by` carries the pane identity, and on a non-kubestellar hive without #6122 that is the hardcoded slug, so a DCO check that matches sign-off to author would fail.

Workflows are now approved so CI will run on this branch.

clubanderson:
Retargeted this PR to v5. The maintainers are landing security work and larger features on v5 and keeping v4 to fixes, and a feature that force-updates agent branches belongs with the larger work, so I rebased your commit onto v5 and moved the base. Your commit is unchanged in author, message and sign-off (old head 7bde1dd, new head a229edb).

Two relocations you will notice, both because v5 has decomposed the big files: the GitHubConfig field and AppSignedCommitsEnabled now live in src/pkg/config/forge_config.go (I placed the new method directly after AppAuthoredPRsEnabled so that method keeps its doc comment attached), and the SetSignedCommits wiring sits next to SetHiveIdentity in src/cmd/hive/notifywire.go rather than main.go. pr_request_signed.go, its test, client.go, the watcher change, the docs and the changelog fragment came across as-is, and every helper the new code and test reference is present on v5.

clubanderson:
Holding until #6276 lands: this PR's `createCommitOnBranch` re-author path would stamp the bracketed App-bot author back in and copy bracketed DCO trailers into the squashed commit, reintroducing #6251. Once #6276 is merged, adopt the bracket-free author/trailer form here and the hold can come off.

clubanderson:
Hold lifted. #6276 merged to v4 (2026-09-08 18:27Z) with the bracket-free bot identity, and this PR now applies the same rule on its side, so the re-author path can no longer reintroduce #6251.

What I checked in the re-author path. `createCommitOnBranch` takes no author or committer input at all, so nothing here writes an author email: GitHub stamps the App's own Bot account on the commit. That is fine for DCO. probot-dco exempts a commit whose GitHub author resolves to a Bot account before it looks at any email (`else if (author && author.type === "Bot") continue;` in dcoapp/app `lib/dco.js`), and Prow's dco plugin only requires a `Signed-off-by:` line to be present. The #6164 failure was different: a git-authored commit with a hand-written bracketed address that GitHub could not resolve to the bot, so the validator ran and rejected it.

What this PR does write is the message, and it copied the agents' `Signed-off-by` trailers verbatim. v5 has not received #6122/#6276 yet (no `hive_git_bot_identity` in `src/deploy/entrypoint.sh` on v5), so rather than depend on the pane identity I added one commit on top of Greg's, 3341c9fc, that re-addresses copied trailers with the same rule #6276 uses: `<slug>[bot]@users.noreply.github.com` (with or without GitHub's numeric id prefix) becomes `<slug>@hive.kubestellar.io`, domain from `HIVE_GIT_BOT_EMAIL_DOMAIN` with the same plain-hostname guard and fallback. Every other sign-off address (a human's, the legacy `hive-bot@kubestellar.io`, the #6276 form itself) is copied unchanged, and the sign-off name keeps its `[bot]`. Tests cover the table of address shapes, the domain override and unsafe-value fallback, a whole-message fold, and a sweep over the mutation body asserting every emitted `Signed-off-by` is bracket-free and syntactically valid. Docs and the changelog fragment say the same.

Greg's commit a229edb2 is untouched (author, message, sign-off). On the red `dco` context: that is Prow's presence-only check, set at 16:43:16Z during the retarget when the PR's commit range briefly included f80698e (a web-UI v4 commit with no sign-off, the one its comment named). a229edb2 carries a valid sign-off, so the push should re-evaluate it green; if it does not, `/check-dco` will.

Emitted identity now: author `<slug>[bot] <id+slug[bot]@users.noreply.github.com>` (GitHub's, Bot-exempt), every `Signed-off-by` trailer `<slug>@hive.kubestellar.io` form.

PR #6152, opened 2026-09-06
🐛 fix: stop opening agent PRs with lost bodies and hedged non-closing issue refs

Danathar:
## Summary

When hive's agents (quality, sec-check, and the rest) finish work, they don't call `gh pr create` directly — they run a small wrapper, `hive-open-pr`, which writes a request file that the hive's watcher turns into an App-bot-authored PR. Two defects in that path, both seen on every repo a production hive works on:

- **PRs opened with an empty body.** The wrapper's flag parser silently swallowed any flag it didn't know — including `--body-file`, which `gh pr create` accepts and agents routinely use for multi-line bodies. The whole body the agent wrote was dropped, and the PR went out containing nothing but the attribution footer ("— hive: agent=… backend=… model=…"). Intermittent, because it depended on whether the model happened to pick `--body` or `--body-file` that run.
- **Issues left open after their fix merged.** The PR-body guidance in the policy templates and kick instructions hedged closing keywords ("`Fixes #N` ONLY if this fully resolves it — use `Refs #N` instead for an epic/multi-phase tracker"). Agents read that cautiously enough that roughly half of fully-resolving PRs shipped a non-closing `Refs #N`, so GitHub never closed the issue on merge.

This PR makes the wrapper accept `--body-file` (and the gh short flags), refuse an empty body loudly, and makes the guidance state `Closes #N` as the normal case; the watcher gains matching server-side guards so no client can open a body-less PR or silently drop its issue reference. Blast radius: the agent PR-open path only — request script, watcher validation, prompt/template text, docs, tests.

## Evidence

From a production hive (all times UTC): footer-only bodies on Danathar/aurora-zfs-simple#91, arch-bootc#169 and #190, sensi#76 and #86, atomic-image-builder#223 — the last one's originating issue (#222) was opened 31 seconds earlier in the same run with a complete body. The agent transcript for #223 shows the exact failure: the agent wrote a full body starting `Closes #222.` to `/tmp/pr-body.md` and ran `hive-open-pr … --body-file /tmp/pr-body.md --label hold`; the request file went in with `body:""`. Non-closing `Refs #N` on fully-resolving PRs: atomic-image-builder#219, #213, #124, #119, #113; zfs-kinoite-complex#98, #90, #17, #11; arch-bootc#187, #175; aurora-zfs-simple#101; sensi#37 — the watcher's `downgraded closing reference` log line appears zero times in the retained log window covering several of these, so the `Refs` came from the agents themselves, i.e. from the prompt text.

## Changes

**Defect 1 — lost body (`bin/hive-open-pr.sh` + watcher):**

- The script now honors `--body-file`/`-F` (file or `-` for stdin) and the gh short flags `-R/-H/-B/-t/-b`; `--body` and `--body-file` together are refused as gh refuses them; a missing/unreadable body file is a hard error.
- Any other unrecognized flag is still tolerated but now warned about on stderr, so the next silently-dropped flag names itself.
- An empty (or whitespace-only) body is refused with exit 2 and no request written. The floor is deliberately just "non-blank": `scanner-automerge` legitimately produces bodies as short as `Closes #12`.
- Server side, `handleOnePRRequest` rejects an empty body before any API spend (`validatePRRequestBody` in `pr_request_claims.go`), quarantining the request as `.rejected` with an error that says the body was lost. The watcher is the choke point every supported PR-open path funnels through (direct `POST /pulls` is proxy-denied), so this holds for any client, not just the script.

**Defect 2 — Refs vs Closes (templates, scheduler, watcher):**

- All 13 policy templates carrying the old parenthetical (quality/sec-check/scanner/architect/guide/ci-maintainer, both `src/pkg/policies/defaults/` and the `src/policies/` mirror) now read: write `Closes #N` when the PR resolves the issue (the normal case); write `Refs #N` ONLY when part of the issue is deliberately left open, and say on the same line what is left and why.
- The kick-time `ghAuthInstructions` block and the ci-maintainer merge-sweep line in `src/pkg/scheduler/scheduler.go` get the same rewrite.
- `hive-open-pr` gains `--issues <N[,N...]>` declaring the originating issue(s); the templates' PR-create examples pass it. The request's previously-informational `issues` field is now enforced: the watcher rejects a request that declares issues its title+body never reference (closing or `Refs` form both count), so a truncated body cannot open a PR that orphans its issue — which is exactly how defect 1 presented.

## How the Closes line was confirmed to survive

`TestHiveOpenPRScript_ClosesLineSurvivesIntoOpenedPR` (in `src/pkg/github/pr_request_body_test.go`) runs the real `bin/hive-open-pr.sh` with `--body-file` + `--issues 1`, feeds the request it wrote to the real watcher (`ProcessPRRequestsOnce`) against a mock GitHub API, and asserts the `POST /pulls` payload body contains the `Closes #1` line and the surrounding prose. The production invocation shape from the #223 transcript (`--body-file` + `--label hold`) was also replayed manually against the fixed script: the request now carries the full body, and the empty-file variant refuses with exit 2.

## Related issues

None filed — defects and fixes are described here.

## Testing

- [x] `cd src && go test ./pkg/github/ ./pkg/scheduler/ ./pkg/policies/` (github: 44s, all pass, including the new script-contract, watcher-rejection, and end-to-end tests; `shellcheck bin/hive-open-pr.sh` clean)
- [x] Other / not run (explain): full `go test ./...` not run; `pkg/config`'s `TestGateFailsClosedWithoutIP6Tables` fails on this host with and without this change (environmental — ip6tables present but unprivileged).

## Contributor checklist

- [x] PR targets `v2` unless a maintainer requested another branch. (targets `v4`, the current default)
- [x] A `changelog.d/<category>-<pr-or-slug>.md` fragment carries the changelog entry for user-visible changes (features, fixes, new env vars, behavior changes) — see `changelog.d/README.md`; or the change is not user-facing (`no-changelog` label). Do not append to `CHANGELOG.md`'s `## Unreleased` directly (#5675).

https://claude.ai/code/session_01FaRN1AKdE4D9144xXheFJL


Danathar:
Six more instances from a production hive this morning (2026-09-07, all times UTC), `acmm_level: 4` with agents in `-holdgated` mode. Both defects, a day after this PR opened — so both are still reproducing rather than a historical batch.

**Defect 1 — footer-only bodies (4)**

| PR | Issue | Agent | Merged | Body |
|---|---|---|---|---|
| `Danathar/zfs-kinoite-complex#106` | #105 | quality | 10:32:48 | 71 ch |
| `Danathar/zfs-kinoite-complex#108` | #107 | quality | 10:33:00 | 71 ch |
| `Danathar/arch-bootc#196` | #195 | sec-check | 10:33:23 | 73 ch |
| `Danathar/arch-bootc#198` | #197 | quality | 10:33:47 | 71 ch |

Each body is nothing but the attribution footer. #106 in full:

```
— hive: agent=quality backend=claude model=claude-opus-5 claude=2.1.226
```

Two different agents (`quality`, `sec-check`) across two repos inside a 59-second window, which matches the intermittent `--body`/`--body-file` split described above rather than anything agent- or repo-specific.

**Defect 2 — non-closing `Refs` on a fully-resolving PR (2)**

| PR | Issue | Merged | Body | Reference |
|---|---|---|---|---|
| `Danathar/atomic-image-builder#258` | #257 | 10:32:30 | 3099 ch | `Refs #257` |
| `Danathar/sensi#97` | #96 | 10:34:13 | 4641 ch | `Refs #96 — …` |

`sensi#97` is the more interesting of the two, because the hedge was *reasoned* rather than careless:

```
Refs #96 — that issue also records `scripts/pr_metrics.py` at 0% as a separate,
lower-priority finding, so it stays open after this lands.
```

That was correct when written: #96 did carry a second finding. But `sensi#102` merged 32 seconds later and covered that half too, so the issue was fully resolved and the `Refs` was stale on arrival, with no mechanism to revise it. Worth noting for the guidance rewrite — under the old text, even a careful agent lands on `Refs` when sibling PRs from the same run close the remaining scope.

**Outcome:** all six issues stayed open after their fix merged, and were closed by hand 3.5–24 minutes later. None of the six appear in the Evidence section above, which stops at 2026-09-06.

Danathar:
Following up on the six instances above — I went back through the 13 `Refs` cases in the Evidence section to see how they actually ended, and the result argues for one small addition to this PR rather than a separate change.

**7 of the 13 self-corrected.** A later phase's PR wrote `Closes #N` and GitHub closed the issue on merge (`atomic-image-builder#212`, `#123`, `#118`, `#111`; `arch-bootc#174`; `aurora-zfs-simple#94`; `sensi#36`). The multi-phase flow works when the final PR gets its keyword right, so the `Refs` hedge is not by itself the thing that strands issues.

**6 needed a manual close**, and they split by cause:

| Issue | Finishing PR | Body | Reference | Covered here? |
|---|---|---|---|---|
| `atomic-image-builder#218` | #220 | 4148 ch | `Refs #218` | yes — nothing was left open, so the reworded guidance yields `Closes` |
| `zfs-kinoite-complex#89` | #90 | — | bare `Refs #89` | yes — bare hedge is now out of spec |
| `arch-bootc#186` | #187 | — | bare `Refs #186` | yes — same |
| `zfs-kinoite-complex#9` | #14 | 3450 ch | **none** | no |
| `zfs-kinoite-complex#97` | #99 | 5947 ch | **none** | no |

The last two are a shape neither defect above describes: the body arrived **complete** — 3450 and 5947 characters, full prose, mutation tables, the lot — and referenced no issue at all. Not a lost body, and not a hedged ref. The agent simply never wrote the line.

`--issues` is the closest guard to this, but it can't reach this case as written. It's optional in the usage:

```
#                --title "<title>" --body "<body>" [--issues <N[,N...]>]
```

and `validatePRRequestBody` returns early when nothing was declared:

```go
if len(req.IssueN) == 0 {
```

So an agent that omits the flag opens an orphaned PR and the guard never fires — the enforcement covers "declared an issue, then lost the reference" but not "never declared one". Given the reworded templates now pass `--issues` in their examples, the gap is narrow, but it is exactly the two cases above.

Worth considering whether an agent kicked on a specific issue should have `--issues` supplied from the kick rather than left to the model to remember — that would make the existing validation authoritative instead of opt-in, without new mechanism. Happy to file this separately if you'd rather keep this PR's scope where it is.

Danathar:
Both nits fixed in `174d49ad`. Thank you for the end-to-end exercise — the
`--allowed-tools`/`ParseClaimedIssues` cross-check in particular is the part I
could not easily convince myself of from the diff.

**1. `--label` no longer warns.** `--label`/`-l`/`--label=` are accepted and
their value dropped, like the other two-argument flags this script does not
need. Your framing was the deciding argument: the label *is* applied
server-side, so the warning does not merely add noise — mid-run it asserts the
opposite of what happens.

I paired it with a discriminating sibling test asserting a genuinely unknown
flag still warns, so the fix cannot silently become "turn the warning off" —
that machinery is right and is how `--body-file` losing the body was finally
noticed.

**2. The fallback escaper handles multi-line bodies.** Agreed this is more
reachable now, and I think it is worse than "pre-existing": a `--body-file`
body is multi-line by nature, so the fallback was broken for precisely the
case this PR exists to fix. On a python3-less host the fix did not work — it
failed safe, loudly, without opening a bodyless PR, but it did not work.

Escaping newline, CR and tab now, and **refusing** the remaining C0 controls
rather than emitting them. I did not take the hard-require-python3 option
because it would break hosts where single-line bodies work today; but a
hand-rolled escaper that quietly writes broken JSON is the shape of the bug
this PR is closing, so the residual fails closed instead of silently.

**Evidence.** The escaper test runs the real script with a `PATH` carrying only
what the fallback needs and no python3 at all, then asserts the request parses
*and* that the body round-trips — not merely that it is parseable. Against the
pre-fix script both new tests fail and the unknown-flag test passes; against
the fix all three pass. `shellcheck -x` clean.

Nothing was needed for the parts you cleared (the `validatePRRequestBody`
placement before any API spend, and `referenceKeywords` covering `Part of #N`);
I re-read both and agree.

PR #6160, opened 2026-09-06
🔒 security: remove hive-process credentials from agent panes with tmux set-environment -r, not -u

gregoryhunt:
## Summary

- Every per-agent tmux server is started by the hive process and inherits its environment into the server's **global** environment, which tmux copies into every pane it forks. The hive process legitimately holds credentials there: the Linear work-source key expanded from `${LINEAR_API_KEY}` (the operator's **personal** key), the Linear OAuth app's `LINEAR_CLIENT_SECRET` and `LINEAR_WEBHOOK_SECRET`, and the full installation token `HIVE_GITHUB_TOKEN` the entrypoint exports.
- The creation-time strip meant to keep those out of agent sessions used `set-environment -u`. `-u` only deletes the **session** entry; a variable that also sits in the global environment shows straight through to the next pane. Verified against tmux 3.5a: `-u` → inherited value visible in a new pane; `-r` → absent; `-r` followed by a plain set → the set value.
- Observed live (2026-09-05, per-UID hive with the Linear agent integration connected): the quality agent's tmux server had `LINEAR_API_KEY`, `LINEAR_CLIENT_SECRET` and `LINEAR_WEBHOOK_SECRET` in `show-environment -g`, beside the sanctioned `LINEAR_ACCESS_TOKEN` in the session environment. The agent used `LINEAR_API_KEY` for `issueCreate`, so every issue it filed was created by the operator rather than by the Hive app user, and the OAuth grant connected for exactly that purpose went unused.
- `ensureTmuxSession` now calls `applySessionEnv`, which records a `-r` removal for every inherited credential (`inheritedCredentialEnvVars`, plus `GH_TOKEN`/`GITHUB_TOKEN` for agents that cannot push) **before** setting the tier's sanctioned pairs from `agentEnvPairs` — order matters, a `-r` after a set would discard the value just set. The downgrade path in `linearRefreshTmuxArgs` removes with `-r` as well.
- `security-model.md` (Layer 3) and `linear-agent.md` (verification step 7) document the invariant.

Related: #6122 / #6123 fix the GitHub-side identity of agent commits; this is the Linear-side counterpart — with it, an agent's `issueCreate` lands authored by the Hive app user as `linear-agent.md` already promises.

## Related issues

None filed; this PR carries the diagnosis. Touches the strip introduced for audit H3 / #4891.

## Testing

- [x] `cd src && go build ./...`, `go vet ./pkg/agent/`, `gofmt`.
- [x] `cd src && go test ./pkg/agent/` — full package green (268s).
- [x] New `TestApplySessionEnv_RemovesInheritedCredentialsFromPanes` starts a **real** tmux server with planted `LINEAR_API_KEY`, `LINEAR_CLIENT_SECRET`, `LINEAR_WEBHOOK_SECRET`, `HIVE_GITHUB_TOKEN`, `GITHUB_TOKEN`; proves a fresh pane sees them (control — fails loudly if the leak ever stops reproducing); runs the real `applySessionEnv` for an ISSUES_ONLY agent with an OAuth credential; asserts a new pane sees none of the planted values and does see `LINEAR_ACCESS_TOKEN`; then downgrades the agent to ADVISORY, runs the refresh-tick args, and asserts the token is gone again.
- [x] `TestLinearCredential_RefreshTickStripsAfterDowngrade` updated from `-u` to `-r`.
- [ ] Other / not run (explain): not exercised on a live per-UID hive; the live evidence above is what motivated it.

## Contributor checklist

- [x] `changelog.d/security-tmux-inherited-credentials.md` carries the changelog entry.


clubanderson:
Reviewed. Verified the tmux semantics on a scratch server: `-u` still leaks the global value into a new pane, `-r` does not, and `-r` followed by a plain set yields the set value. Ordering in `applySessionEnv` is right, and the test asserts the actual invariant against a real pane. Nice.

Two follow-up-sized notes, not blocking:
- `ensureTmuxSession` respawn-pane failure is warn-only, which after this change means the first pane keeps the hive credentials. Either make it fatal for the session or scrub `cmd.Env` on the cold `new-session` exec as defense in depth.
- The list is a denylist; any other secret the hive process holds still flows through. An allowlist would close that class.

Will lgtm on green CI.

PR #6163, opened 2026-09-06
✨ feat(agents): per-agent reasoning_effort for codex and agy (config + dashboard + launch)

gregoryhunt:
## What

Adds a per-agent `reasoning_effort` setting and honors it everywhere the hive builds a launch command — closing the gap `ResolveReasoningEffort`'s own comment names (*"Hive has no per-agent reasoning-effort setting yet — when it grows one, this is the constant it should replace"*). The scripted launch paths (`bin/agent-launch.sh`, the contributor relay) already speak effort via `AGENT_REASONING_EFFORT`; this brings the manager-launched tmux path and the dashboard up to the same contract, with the identical flag spellings so the paths cannot drift.

## Changes

**config** — `AgentConfig.ReasoningEffort` (`reasoning_effort` in `hive.yaml`), plus `ReasoningEffortsByBackend` / `ValidateReasoningEffort`: codex accepts `minimal|low|medium|high|xhigh` (passed as `-c model_reasoning_effort="<v>"`), agy `low|medium|high` (passed as `--effort <v>`). Validation happens at set time, same rationale as `ValidateBackend`. Pack apply reconciles named fields only, so the field needs no ownership marker.

**agent manager** — `backendLaunchCmd` / `toolRulesToLaunchCmd` take the configured effort:
- **agy** launches at the configured effort when agy accepts the value, else `agyDefaultEffort` — mirroring the relay's rule of rejecting codex-only values like `xhigh`, which would make agy ignore the model outright.
- **codex** gains an explicit case on both paths. It also passes `--model`, which the bare-binary default silently dropped before: the codex model dropdown existed, but the selection never reached the CLI on the tmux launch path. The effort rides as the config key, valid with or without `--model` (effort alone runs codex's default model at that effort — the same contract the relay documents for `AGENT_REASONING_EFFORT` without `AGENT_MODEL`).
- `ResolveReasoningEffort(backend, model, configured)` keeps the attribution trail reporting the launch-time truth; both resolvers (manager + `cmd/hive` config fallback) updated together, as its comment requires.

**dashboard**
- `PUT /api/config/agent/{name}/models` accepts `reasoning_effort` (pointer semantics: absent leaves it unchanged, explicit `""` clears — empty is a meaningful value, the backend's own default).
- New owner-gated `POST /api/effort/{agent}/{effort}` for the grid dropdown (`default` clears; a path segment cannot be empty), persisting and restarting the session exactly like the model switch — the effort only exists on the launch command line. Registered in `openapi.json` and in the owner-only deny test.
- `reasoningEffort` in agent status JSON, and an **effort dropdown beside the model dropdown** on both agent-card templates, rendered only for backends with an effort control.

**docs** — `reasoning_effort` documented in `agent-configuration.md`'s engine block.

## Deliberately out of scope

- **claude**: the Claude Code CLI has no reasoning-effort flag; its nearest lever is a thinking-budget env var (`MAX_THINKING_TOKENS`). Happy to wire that mapping as a follow-up if maintainers want it, but it's a design call (effort→token-budget mapping) rather than a flag passthrough, so it isn't smuggled in here.
- **sandbox kicks**: the sandbox-kick default path currently resolves the codex backend to the claude binary (`sandboxBackendBinary`), a pre-existing gap this PR doesn't widen or fix; `launch_cmd` remains the codex route there and can carry the effort key verbatim.

## Testing

`go build ./...` clean; `go test ./pkg/agent/ ./pkg/config/ ./pkg/dashboard/` pass in full (the only failure on my machine, `TestEntrypointHardensRuntimeConfigItRecreates`, fails identically on a pristine v4 checkout — macOS chown semantics). New tests: codex launch flag contract (model, effort, both, neither; ToolsConfig path parity), agy configured-effort override + invalid-value fallback, `ResolveReasoningEffort` table extended, `ValidateReasoningEffort` table, models-endpoint field semantics (set / absent-unchanged / explicit-clear / rejected-not-persisted), effort endpoint (set / `default` clear / bad value / no-effort backend / 404), owner-only deny row. Every inline `<script>` block passes `node --check`; the OpenAPI route-parity test passes.

https://claude.ai/code/session_01UxSYmSAyMKgKyGcjd52MDk


clubanderson:
Reviewed. The feature code is solid; a few things before lgtm.

**Blocker: stray symlink committed.** `src/pkg/dashboard/.antigravitycli/32fa6efa-f2d7-4592-acf4-ff3093e2c2cd.json` is a mode-120000 symlink into a home directory. Please drop it.

Should fix:
- No load-time validation of `reasoning_effort`. `config.validate()` and `RejectInvalidAgentOverlays` never call `ValidateReasoningEffort`, so a hand-edited `reasoning_effort: turbo` in hive.yaml reaches codex verbatim as `-c model_reasoning_effort="turbo"` and gets stamped into attribution. Only the two dashboard paths validate.
- `handleEffortSet` validates against `proc.BackendOverride` when set, but `handleAgentConfigModels` validates against `agentCfg.Backend` only. With a runtime override to agy on a codex-configured agent the models PUT accepts `xhigh` and launch silently falls back to `low`, which is the "appears to revert" case your own comment describes.

Nits:
- Backend switch (models PUT or pack apply) leaves a stale `ReasoningEffort`; consider clearing it when the new backend rejects the stored value.
- `dashboard/openapi.json` request schema for `/api/config/agent/{name}/models` does not list `reasoning_effort` (pointer semantics, `""` clears).
- `handleEffortSet` BackendOverride branch has no test.

Workflows are now approved so CI will run on this branch.

clubanderson:
Review note before this can be labelled: the branch includes a stray symlink, src/pkg/dashboard/.antigravitycli/32fa6efa-f2d7-4592-acf4-ff3093e2c2cd.json, that points into a home directory (/Users/.../.gemini/config/projects/...). It looks like a local tool artifact that got committed by accident. Please drop it from the PR (git rm the file and force-push, or add a follow-up commit). The reasoning_effort change itself reads well: empty keeps the agy default of low and adds no codex flag, so existing launches are unchanged, and the new endpoint is owner-only with set-time validation. Once the symlink is gone and CI is green this is good to go.

clubanderson:
Update: I removed the stray symlink myself with a single follow-up commit on your branch (8335609, chore: drop accidentally committed local tool symlink). Your commits are untouched, so nothing for you to do here. Once CI is green on the new head this gets labelled.

clubanderson:
Reviewed on the diff. ResolveReasoningEffort gains the configured value and both resolvers (manager + cmd/hive fallback) are updated together, as the function's own comment requires. ValidateReasoningEffort rejects at set time: unknown backend and out-of-vocabulary values both error, and agyLaunchEffort refuses codex-only values like xhigh rather than passing them through, which matches the relay's existing rule. The codex path also now passes --model on the tmux launch, which was silently dropped before. Approving; will add lgtm once CI is green.

PR #6248, opened 2026-09-08
🐛 fix: attach the hive's current Linear credential at the proxy so a rotated OAuth token in an agent's process no longer 401s

gregoryhunt:
## Summary

- The connected Linear app's OAuth access token rotates (~24h). Hive refreshes it and re-pushes the fresh value into each agent's tmux session environment hourly — but a session environment only reaches processes forked afterwards. The agent's CLI keeps the token it was born with, and every tool shell inherits that copy, so from the first rotation after launch every Linear call from inside the pane carries a dead token.
- Observed live (2026-09-06/07): the quality agent's session environment held the fresh token — verified valid through this very proxy as that agent's UID (`viewer: Onboard AI Hive`, HTTP 200) — while every call from its pane returned `401 AUTHENTICATION_ERROR` for two days. Its earlier kicks had silently fallen back to the operator's personal `LINEAR_API_KEY`, so every issue it filed showed the operator as creator; once the policy forbade that fallback, it filed nothing at all.
- The egress proxy already terminates every agent request to `api.linear.app` for the tier gate (`enforceLinear`). After the gate passes, `injectLinearCredential` now replaces the `Authorization` header with the hive's **current** credential — the OAuth token as `Bearer`, or the work-source API key bare — for agents at the ISSUES_ONLY floor and above, the same tier the session-env push uses. The copy in the agent's environment (stale, absent, or a placeholder) no longer matters: this is the inference translator's server-side key swap applied to Linear. Requests below the floor and requests the gate refuses are untouched, so injection never widens a tier.
- `main.go` hoists the existing Linear credential resolver into a variable and hands the same closure to both the agent manager and the proxy, so the two can never name different credentials.
- `linear-agent.md` (parity table) and `security-model.md` (Layer 3) document it.

Companion to #6160 (which stops the personal key from reaching the pane at all) and #6161 (kick text). With all three, an agent's Linear writes are authenticated as the Hive app user across token rotations, without restarts.

## Related issues

None filed; this PR carries the diagnosis.

## Testing

- [x] `cd src && go build ./...`, `go vet ./pkg/proxy/`, `gofmt`.
- [x] `cd src && go test ./pkg/proxy/` — full package green (44s).
- [x] New `linear_inject_test.go` drives the real `proxyHTTPHost` over `net.Pipe`s with `api.linear.app` as the MITM'd host and reads what the upstream receives: a rotated-out Bearer is replaced with the current token (exactly one `Authorization` header); a request with no header gets one; the API-key fallback is sent bare; an ADVISORY read is forwarded as sent; no resolver / empty credential leave the request alone; an ADVISORY `issueCreate` is refused by the gate without the resolver being consulted.
- [ ] `go test ./...` — not run in full; the change is confined to `pkg/proxy` plus the wiring line in `main.go`.
- [ ] Other / not run (explain): not yet exercised on the live hive (it needs an image bump); the live evidence above is what motivated it.

## Contributor checklist

- [x] `changelog.d/fixed-linear-token-proxy-inject.md` carries the changelog entry.


clubanderson:
Reviewed. Injection scoping is correct: exact-match Linear host, after the block check, excludes the internal caller, gated on ISSUES_ONLY+, `Header.Set` not `Add`, no token in logs. Tests cover the important cases.

Two nits, neither blocking:
- `SetLinearCredentialResolver` writes `p.linearCredential` unsynchronized. Safe today because main.go wires it before `Start()`, but either add a "must be called before Start" comment or use `atomic.Pointer`.
- `src/pkg/agent/manager.go` (around the `linearEnvPairs` block) still says "tmux forks a fresh CLI per turn", which contradicts this PR's premise of a long-lived CLI. Worth reconciling the comment.

Will lgtm on green CI.

PR #6250, opened 2026-09-08
🐛 fix(agent): keep Codex alive during kicks

gregoryhunt:
## Summary

On a hosted Hive using Codex 0.146.0, kicking the quality agent exited Codex and typed `/clear` and the quality policy into Bash, producing `command not found` errors instead of agent work.

`deliverKickLocked` sends Ctrl+C after the caller has verified input readiness, but Codex can exit at an idle prompt. Skip Ctrl+C for the effective Codex backend and clear input with Ctrl+U alone. This also covers runtime backend rotation; Goose continues to skip input clearing.

## Related issues

Diagnosed from a live operator transcript; no separate issue filed. #6163 already covers the missing Codex model argument, and #6105 covers dropped literal kick chunks. This PR addresses the separate pre-kick interruption failure.

## Testing

- [x] `cd src && go test ./pkg/agent -run 'TestCodexKickDoesNotExitCLI|TestDeliverKickLocked' -count=1`
- [x] `cd src && go vet ./pkg/agent`
- [x] Regression test fails against the original manager using a Go overlay: `kick text reached shell`. It models an idle CLI exiting on Ctrl+C and exercises configured Codex and backend overrides, with clear-on-kick both enabled and disabled.
- [x] `git diff --check`
- [ ] Full build and full suite not run. `TestSendKick_DeliversToReadyPane` fails locally with `CLI did not become ready after restart` on both the original manager and the fix.
- [ ] Live hosted deployment verification remains pending; the patch has not been deployed.

## Contributor checklist

- [x] Targets `v4` per CONTRIBUTING.md.
- [x] Title uses the repository emoji convention.
- [x] Changelog fragment documents the behavior change; no configuration or policy changes required.
- [x] No credentials or local runtime state included.

### DIFF
--- a/changelog.d/fixed-codex-kick-exit.md
+++ b/changelog.d/fixed-codex-kick-exit.md
@@ -1,0 +1,1 @@
+- Codex agents no longer receive Ctrl+C before each kick, which could exit the CLI and send the quality policy and task into Bash as shell commands. Input clearing now uses Ctrl+U for Codex, including runtime backend overrides.
--- a/src/pkg/agent/codex_kick_regression_test.go
+++ b/src/pkg/agent/codex_kick_regression_test.go
@@ -1,0 +1,40 @@
+package agent
+
+import "testing"
+
+// Model an idle Codex exiting on Ctrl+C: no prompt text may reach the shell.
+// Exercise both a configured Codex agent and a runtime backend rotation.
+func TestCodexKickDoesNotExitCLI(t *testing.T) {
+	for _, override := range []bool{false, true} {
+		for _, clear := range []bool{false, true} {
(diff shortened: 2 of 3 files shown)


clubanderson:
Reviewed. Small semantic note: switching to `effectiveBackend(agent)` also means an agent with `Config.Backend=goose` and `BackendOverride=claude` now gets Ctrl+C, which is the right behavior but unmentioned. The regression test proves Ctrl+C is gone but never asserts Ctrl+U is still sent for codex; a `sawCu` flag would lock in the stated contract. Not blocking. Will lgtm on green CI and verify on a hosted spoke by digest after deploy.

PR #6264, opened 2026-09-08
✨ feature: let a repo declare an ACMM criterion satisfied off-repo, without letting it advance a level

### LINKED ISSUE #6263: ✨ feature: ACMM criteria cannot tell a missing capability from one that moved off-repo
## Summary

The ACMM evaluation panel scores a repository by looking for files: each criterion carries a list of paths and passes if any of them exists (`checkCriterion`, `src/pkg/dashboard/api_acmm_eval.go`). That check is cheap, language-agnostic and needs no cooperation from the repo, which is why it works.

It cannot, however, tell two very different repositories apart:

- one that **never built** the capability, and
- one that **deliberately moved the capability** somewhere the file check cannot see.

Both read as a missing file. The second repo is scored as having lost a capability it still has, and the cheapest way for its maintainer to restore the score is to put the file back — which is the wrong outcome whenever the file was removed for a reason.

## Use case

The gap is structural rather than cosmetic. A file-existence check measures *where a capability is stored* and reports it as *whether the capability exists*. Those two come apart whenever a capability can legitimately live outside a repository: a shared org-level reusable workflow, a platform team's central CI, a service mesh replacing per-repo config — or hive itself doing what a per-repo workflow used to do.

When they come apart, the panel reports a regression that did not happen, and every response available to the maintainer is bad:

1. re-add the file that was removed for cause,
2. broaden the criterion's pattern list for every repo in the fleet, or
3. leave a permanent red on the board.

The third is the default, and it is the most damaging, because it trains the operator to read reds as noise. That is a direct loss of signal for the panel as a whole, not just for the affected repo.

## The case that surfaced it

`Danathar/sensi` deleted `.github/workflows/ai-fix.yml`. The job it removed ran an agent in-repository against a labelled issue or an `@claude` comment, in a job holding `contents: write`, `issues: write`, `pull-requests: write`, `id-token: write` and an API key.

Its label gate turned out not to be an authorisation check at all: the `ai-fix-requested` label was applied automatically by the issue-filing bot, so the gate admitted whatever that bot filed rather than expressing a human triage decision. The same work was already being done by hive, at ACMM L4, across one trust boundary instead of two. Removing the workflow was straightforwardly the right call.

Two L4 criteria name that one filename:

| Criterion | Patterns |
|---|---|
| `acmm:ai-fix-workflow` | `ai-fix.yml`, `ai-fix-requested.yml`, `claude.yml` |
| `acmm:copilot-review-apply` | `copilot-review-apply.yml`, `apply-copilot.yml`, **`ai-fix.yml`**, `auto-review.yml` |

So one security fix cost two criteria and took L4 from **9/9 to 7/9** — still passing (0.78 against the 0.70 threshold) but no longer full green, for a capability that had not gone anywhere.

Observable, on a five-repo hive evaluated 2026-09-08T14:52:57Z:

```
atomic-image-builder  L4=9/9    zfs-kinoite-complex  L4=9/9
arch-bootc            L4=9/9    aurora-zfs-simple    L4=9/9
sensi                 L4=7/9   ← ai-fix-workflow + copilot-review-apply
```

sensi is the only repo of the five that made the security fix, and the only one not full green. The four repos still reading 9/9 are the four that still carry the workflow. The panel is currently rewarding the four that have not fixed it.

## Proposed solution

Let a repository declare, in a committed `.acmm.yml`, that a criterion is satisfied somewhere other than the file the criterion looks for — naming what satisfies it and why:

```yaml
waivers:
  - id: acmm:ai-fix-workflow
    satisfied_by: hive
    reason: >
      The in-repo AI-fix path was removed in PR #118 — it duplicated what hive
      already does at ACMM L4, across a second trust boundary. Hive is the sole
      autonomous path. See docs/SECURITY-AI.md.
```

**This must not become a way to skip being ACMM compliant.** That is the design constraint, not an afterthought, and the proposal is built around it:

1. **A waiver can never advance a level.** Level pass/fail is computed on *detected* criteria alone; the displayed ratio includes waivers. A repo that waives an entire level shows a full ratio, still does not pass it, and its codebase level does not move. Waivers close the distance between *passing* and *full green* — nothing more. No amount of committed YAML advances a repo.
2. **A waiver never overrides a detection.** It is consulted only after the file check has already failed.
3. **It must name a real criterion**, validated against `universalCriteria`, so a typo cannot sit in the file looking effective and a renamed criterion re-opens as a visible gap.
4. **It must justify itself.** `satisfied_by` and `reason` are both required. A declaration that cannot say where the capability went is refused — without those fields it says only "do not check this", which is precisely the thing this must not become.
5. **It stays visible.** Marked on the criterion row, and with a red asterisk on the level line — the altitude operators actually read. Plus git history, review and attribution, because it is a committed file.

## Alternatives considered

- **Re-add a stub `ai-fix.yml`.** Turns the board green in one commit, and is the worst option: it re-creates the exact filename removed for cause and leaves the maturity dashboard green on a criterion nothing satisfies. sensi's own commit message names this failure mode — *"unarmed is a configuration state, not a property of the design"*.
- **Broaden the criteria patterns.** The capability is off-repo; there is no in-repo filename that honestly means "hive does this". Accepting something like `docs/SECURITY-AI.md` would weaken the criterion for every repo in the fleet to fix one.
- **A waiver list in governor config** rather than in the repo. Central and clickable, but the justification then lives outside the repo it describes and is not reviewed alongside the code that changed.
- **Let waivers score like any other pass.** Simpler, but it makes "waive your way to L6" arithmetically possible. Computing level pass/fail on detected criteria alone costs one extra field and removes the possibility entirely.

## Scope and compatibility

- **Target branch:** v4.
- **Operator impact:** none unless a repo commits `.acmm.yml`. No change to any existing score, level, or API field's meaning; new JSON fields are additive.
- **Contributor/agent impact:** none. A waived criterion passes, so it stops being offered as a gap issue — which also stops the panel re-filing the same unclosable issue on every scan for a gap the repo has already answered.
- **Cost:** effectively zero for repos that do not use it. `prefetchDirectories` already lists the repository root, so a repo with no `.acmm.yml` is settled from that cache with no extra GitHub call; only a repo that declares waivers pays for the content fetch. This matters against a refresh that is already up to ~29 `GetContents` calls per repo inside a 20s per-repo timeout.

## Additional context

Implementation is ready and linked below. It adds `src/pkg/dashboard/acmm_waivers.go`, splits `ACMMLevelScore` into displayed and detected scores, renders the marker in all three level renderers, and documents the whole thing in `src/docs/acmm-waivers.md`.

Danathar:
## Summary

The ACMM panel scores a repository by looking for files: each criterion carries a list of paths and passes if any of them exists. The check is cheap and needs no cooperation from the repo, which is why it works — but it cannot tell two very different repositories apart. One never built the capability. The other deliberately moved it somewhere the file check cannot see. Both read as a missing file, so a maintainer who removes something for a good reason is scored as having lost the capability, and the cheapest way to restore the score is to put the file back.

This adds `.acmm.yml`: a committed declaration that a criterion is satisfied elsewhere, naming what satisfies it and why. **It is not an exemption from ACMM, and it is built so it cannot become one — a waiver can restore full green, but can never advance a level.**

Blast radius: no behaviour change for a repo without the file, and no additional GitHub calls for one either. New JSON fields are additive; no existing score, level, or field changes meaning.

## Related issues

Fixes #6263

## Why this is needed, in general

The gap is structural, not cosmetic. A file-existence check measures *where a capability is stored* and reports it as *whether the capability exists*. Those come apart whenever a capability can legitimately live outside a repository — a shared org-level reusable workflow, a platform team's central CI, hive itself doing what a per-repo workflow used to do.

When they come apart, the model reports a regression that did not happen, and every available response is bad: re-add the file that was removed for cause, loosen the criterion for every repo in the fleet, or leave a permanent red that trains the operator to ignore reds. The last is the default and the most damaging, because it costs the panel its signal generally, not just on the affected repo.

## The case that surfaced it

`Danathar/sensi` deleted `.github/workflows/ai-fix.yml`. The job it removed ran an agent in-repository against a labelled issue or an `@claude` comment, holding `contents: write`, `issues: write`, `pull-requests: write`, `id-token: write` and an API key. Its label gate turned out not to be an authorisation check: `ai-fix-requested` was applied automatically by the issue-filing bot, so the gate admitted whatever that bot filed rather than expressing a human triage decision. The same work was already being done by hive, at ACMM L4, across one trust boundary instead of two. Deleting it was straightforwardly correct.

Two L4 criteria name that one filename — `acmm:ai-fix-workflow` and `acmm:copilot-review-apply` both list `.github/workflows/ai-fix.yml` — so one security fix cost two criteria and took L4 from **9/9 to 7/9**. Still passing at 0.78 against the 0.70 threshold, but no longer full green, for a capability that had not gone anywhere.

On the five-repo hive where this surfaced, sensi is the only repo that made the fix and the only one not full green; the four still reading 9/9 are the four that still carry the workflow. The panel is currently rewarding the four that have not fixed it. Full detail and evidence in #6263.

## What a waiver cannot do

The point is emphatically **not** to let anyone skip being ACMM compliant, so the mechanism is built to be incapable of it. Each property has a test:

1. **A waiver can never advance a level.** `ACMMLevelScore.Passed` derives from `DetectedScore` — detected criteria only — while the displayed ratio includes waivers. A repo that waives an entire level shows a full ratio, still does not pass it, and its codebase level does not move. Waivers close the distance between *passing* and *full green*; nothing else. (`TestScoreResults_WaiversNeverAdvanceALevel`, `TestScoreResults_WaiversDoNotRaiseCodebaseLevel`, `TestEvaluateAllRepos_WaiverSpreeStaysBlocked`)
2. **A waiver never overrides a detection.** Consulted only after the file check has failed, so a repo that later adds the real file stops depending on the declaration silently. (`TestEvaluateAllRepos_WaiverNeverMasksDetection`)
3. **It must name a real criterion**, validated against `universalCriteria`, so a typo cannot sit in the file looking effective and a renamed criterion re-opens as a visible gap. (`TestParseACMMWaivers_EveryIDMustExist`)
4. **It must justify itself.** `satisfied_by` and `reason` are both required; without them the declaration says only "do not check this", which is the thing this must not become. A waiver missing either is dropped and the criterion stays red.
5. **It stays visible.** Marked on the criterion row, and with a red asterisk on the level line — the altitude operators actually read. Plus git history, review and attribution, because it is a committed file.

## Change

- **`src/pkg/dashboard/acmm_waivers.go`** (new) — parse and validate `.acmm.yml` / `.acmm.yaml`, with a 64 KiB cap on repo-supplied input. `fetchACMMWaivers` consults the root listing `prefetchDirectories` already cached, so a repo with no declaration costs **zero** extra GitHub calls; only a repo that has one pays for the content fetch. That matters against a refresh already up to ~29 `GetContents` calls per repo inside a 20s per-repo timeout. (`TestFetchACMMWaivers_CostsNothingWhenRootLacksTheFile`)
- **`src/pkg/dashboard/api_acmm_eval.go`** — `CriterionResult.Waived` plus the declaration verbatim; `ACMMLevelScore` gains `Detected` / `DetectedScore` / `Waived` and derives `Passed` from detection alone. In the multi-repo aggregate a criterion is marked waived only when no repo detected the real thing — one repo holding the file makes the fleet-wide claim true on its own.
- **`src/pkg/dashboard/static/index.html`** — `acmmWaiverDetail` renders the justification above the pattern list, so a reader who opens a green row whose patterns match nothing finds the explanation in the same place they found the contradiction. `acmmLevelWaiverMark` puts a red asterisk on every level line across all three level renderers; when a level's ratio is full but it still does not pass, the tooltip gives the detected count and says waivers cannot advance a level. Repo-supplied text is escaped — a repo that could inject markup here could inject it into the operator's dashboard by committing a file.
- **`src/docs/acmm-waivers.md`** (new), indexed in `src/docs/README.md`.

## Testing

- [ ] `cd src && go test ./...` — **`./pkg/dashboard/` passes** (69s, 24 new tests). Full-tree run has one failure, `pkg/config`'s `TestGateFailsClosedWithoutIP6Tables`, which is unrelated (egress-gate / ip6tables environment) and **fails identically on a clean worktree at HEAD `50b7d76d` with none of this applied** — verified by checking out a baseline worktree and running it there.
- [x] Other: `go vet ./pkg/dashboard/` clean. Dashboard lint (`.github/scripts/check-inline-js.js`: `node --check` plus ESLint `no-undef`) passes over both HTML trees.

## Contributor checklist

- [x] PR targets `v4` — the template still says `v2`, but `v2` is sunset and `changelog.d/` exists only on `v4` per `.github/release-lines.yml`.
- [x] Commits include DCO sign-off.
- [x] Docs updated — `src/docs/acmm-waivers.md`, linked from `src/docs/README.md`.
- [x] `changelog.d/added-acmm-waivers.md` carries the entry.

https://claude.ai/code/session_012jvzrUFSdxYjyyZigZ6biz


clubanderson:
Retargeted this PR to v5. The maintainers are landing larger feature work on v5 and keeping v4 to fixes, and a change to how ACMM criteria are scored belongs with the larger work, so I rebased your commit onto v5 and moved the base. Your commit is unchanged in author, message and sign-off (old head 6300dc2, new head 115a7b5); it applied cleanly, so nothing moved.

The detected-versus-matched split, with Passed derived only from detected criteria, is a nice way to keep waivers from becoming a level shortcut. Thanks for the careful writeup.

PR #6269, opened 2026-09-08
🐛 fix(ci): read Go version from go.mod instead of pinning '1.25'

### LINKED ISSUE #6260: [scanner] ci: go-version '1.25' pins block every go-1.26 dependency bump, and the App can't fix them (workflows perm missing)
## Finding

Dependabot PR #6253 (golang.org/x/sys 0.47.0 → 0.48.0) fails **every** Go CI job. x/sys v0.48.0 declares `go 1.26.0`, dependabot mirrored that into `src/go.mod`, and all twelve `setup-go` steps pin `go-version: '1.25'`, so each job dies at load with:

```
go: go.mod requires go >= 1.26.0 (running go 1.25.14; GOTOOLCHAIN=local)
```

Affected pins: `coverage-hourly.yml` (1), `go-security-analysis.yml` (4), `testutil-guard.yml` (1), `v2-ci.yml` (1), `v2-tests.yml` (5). The container images already build on `golang:1.27`, so only CI lags.

A second, structural problem: the hive's GitHub App has no `workflows` permission, so agents cannot push workflow edits at all — my prepared fix on the #6253 branch was rejected with `refusing to allow a GitHub App to create or update workflow ... without workflows permission`. Every future dep bump whose module requires go ≥1.26 will red-line the same way with no agent-side remedy.

## Recommendation

1. Bump the twelve `go-version: '1.25'` pins to `'1.26'` (validated locally: the module builds clean on Go ≥1.26 with x/sys 0.48.0). Consider `go-version-file: src/go.mod` instead so CI follows the module and this class of failure disappears.
2. Optionally grant the App `workflows` permission so agents can maintain these pins.

---
*Filed by scanner agent (ACMM L5 — hold-gated mode).*
---
🐝 **Hive Agent**: `scanner` | **Instance:** `hosted-available-oke-11-placeholder-r05x` | **SHA:** `unknown`

— hive: agent=scanner backend=copilot model=claude-fable-5

clubanderson:
## Summary

Every Go CI job pinned the toolchain with `go-version: '1.25'` in its `setup-go` step. When a dependency bump raises the module's `go` directive (dependabot PR #6253 pulled golang.org/x/sys 0.48.0, which declares `go 1.26.0`), every job died at load with `go: go.mod requires go >= 1.26.0 (running go 1.25.14; GOTOOLCHAIN=local)` - a failure on the pin, not on the code under test.

This PR replaces each hard-coded pin with `go-version-file: src/go.mod`, so CI installs whatever the module declares and future bumps to the `go` directive propagate automatically. Every affected step already used `cache-dependency-path: src/go.sum`, so the go.mod path is consistent with the existing cache configuration.

## Workflows changed (12 pins replaced)

- `.github/workflows/coverage-hourly.yml` - 1 pin
- `.github/workflows/go-security-analysis.yml` - 4 pins
- `.github/workflows/testutil-guard.yml` - 1 pin
- `.github/workflows/v2-ci.yml` - 1 pin
- `.github/workflows/v2-tests.yml` - 5 pins

No workflow pinned a different version for a stated reason, so none were left in place. After this change no `go-version:` key remains under `.github/`.

## Deliberately out of scope

`src/go.mod` currently declares `go 1.25.6` with no `toolchain` line and is not touched here. Raising the directive to 1.26 is a separate decision; this PR only stops CI from contradicting whatever the module says. Once merged, the pending x/sys bump (and any later bump that raises the directive) will be exercised against the right toolchain.

The issue's second point - the hive's GitHub App lacks the `workflows` permission, so agents cannot push workflow edits themselves - is an org-level setting and is not addressed by this PR.

A `changelog.d/fixed-6260-ci-go-version-from-gomod.md` fragment is included.

Fixes #6260


clubanderson:
Pushed a follow-up commit bumping `src/go.mod` from `go 1.25.6` to `go 1.25.14`. The original PR body called that a separate decision, but it is not: the old literal `'1.25'` pin resolved to the latest 1.25.x on every run, whereas `go-version-file` resolves to exactly the directive. Switching froze CI on 1.25.6, which govulncheck flags for GO-2026-6090 / 6091 / 6218 (`net/url`, `html/template`), and the testutil-guard floor shifted on the same toolchain change - both red on the first run. 1.25.14 is the current patched release and restores what CI was actually running before.

PR #6277, opened 2026-09-08
🐛 fix(agent): per-agent XDG state/data dirs so CLI state is not shared across agents

### LINKED ISSUE #6238: Shared /data/home/.local makes per-agent CLI state a shared resource; umask wrappers are a per-CLI allowlist that fails closed for new backends
## Summary

Every agent's `$HOME/.local` is a symlink into one shared `/data/home/.local`, so all agents share a single `.local/share`. That makes per-agent CLI state a shared mutable resource, and the mitigation — `umask 007` wrappers in `src/Dockerfile` Layer 8.26 — is a per-CLI allowlist that has to be extended by hand for every new backend. A per-agent `XDG_DATA_HOME` would remove the class instead of patching instances.

## Current state

```
$ ls -ldn /data/home/agents/ci-maintainer/.local
lrwxrwxrwx 1001 1000 /data/home/agents/ci-maintainer/.local -> /data/home/.local

$ ls -ldn /data/home/.local/share/*/
drwxrwsrwx 2005 1000  applications/
drwxrwsrwx 2006 1000  corral/
drwxrwsrwx 1001 1000  goose/
drwxrwsrwx 2007 1000  pki/
```

Note the owners: 2005, 2006, 1001, 2007 — whichever agent happened to create each directory. It works because the mode is permissive, not because anything guarantees it.

Layer 8.26 exists precisely because that guarantee is absent:

> Agent CLIs rewrite their credential/settings state as mode 600 owned by the writing agent (agy: `antigravity-cli/*.json` + oauth token; claude: `.credentials.json`). With a shared `$HOME` (/data/home) that breaks every other agent on the same provider (EACCES -> login/onboarding screens).

## Why this is worth fixing structurally

**It is an allowlist, so it fails closed for anything not on it.** `muse` was not on the list and every muse-backed agent on a live spoke crash-looped — hundreds of restarts — because the first agent to launch created the state tree group-unwritable. Fixed in [#6222](https://github.com/hivecommons/hive/pull/6222) by adding `muse` to the list, which is correct but is the third entry in a list that will need a fourth.

**Some CLIs actively distrust a world-writable state root.** With the permissive mode the wrappers produce, muse refuses to enable part of itself:

```
muse: local session messaging disabled: unsafe_registry_root:
unsafe local Session registry root /data/home/agents/<agent>/.local/share/muse/runtime/muse:
root component is unsafe or unreadable
```

It degrades gracefully and hive does not use cross-session messaging, so nothing is broken today — but "make the directory world-writable" is in direct tension with what a CLI hardening its own state expects, and that tension will recur.

**Sharing is not obviously wanted for all state.** Sharing *credentials* is the point (one `claude auth login` covers the fleet, and `SharedAuth` in the operator formalises exactly that for `.claude`/`.gemini`/`.codex`). Sharing *session transcripts, run locks, and plugin caches* is incidental — a side effect of the symlink, not a design goal. Session logs are also what usage/cost attribution reads, so co-mingling them across agents is actively unhelpful.

## Suggestion

Give each agent its own `XDG_DATA_HOME` (e.g. `/data/home/agents/<agent>/.local/share`) and keep sharing explicit and narrow — the credential directories that `SharedAuth` already names — rather than sharing everything by default and clawing back safety with umask wrappers.

That would:

- remove the per-CLI allowlist and its fail-closed-for-new-backends behaviour;
- let CLIs that check their state root's permissions work as designed;
- keep per-agent session logs separated, which the token scanners want anyway.

Migration is the real cost and I have not scoped it: existing spokes have live state under the shared tree, and anything that currently relies on incidental sharing would need identifying first. Filing this to agree the direction before anyone writes code.

Found while adding the muse backend; the immediate fault is fixed in #6222 and this is the structural follow-up I said I would file separately rather than bundle.

clubanderson:
## Summary

Every per-UID agent's `~/.local` was a symlink bridge (`interactiveHomeBridgeDirs`) into the one shared `/data/home/.local`, so each backend CLI's session transcripts, run locks, sqlite state and plugin caches under `~/.local/share` and `~/.local/state` were a single contended tree owned by whichever agent wrote first. That is the class of fault described in #6238: the Dockerfile umask wrappers kept it working only as a per-CLI allowlist that fails closed for every new backend, and CLIs that check their state root's permissions (muse's session registry) refused to trust the permissive mode.

This follows the per-agent `CODEX_HOME` precedent (`setupCodexHome`) and gives each agent its own XDG data and state roots under its per-agent HOME.

## What became per-agent

- `XDG_DATA_HOME` = `/data/home/agents/<agent>/.local/share`
- `XDG_STATE_HOME` = `/data/home/agents/<agent>/.local/state`

Both are exported explicitly by `agentEnvPairs` whenever HOME itself is per-agent (per-UID agent, escape hatch not set), so the answer cannot depend on how each CLI resolves XDG. `bin/agent-launch.sh` mirrors the export as a fallback for launches that reach it with only a per-agent HOME set. Provisioning (`setupAgentXDGDirs`, called from `setupInteractiveHome`) makes `~/.local` a real agent-owned directory, pre-creates `share`, `state` and the goose log directory goose panics without, and retires a legacy `.local` symlink bridge when it finds one. `.local` is removed from `interactiveHomeBridgeDirs`.

## What stayed shared and why

- `~/.config` (`XDG_CONFIG_HOME`): still bridged to `/data/home/.config`. It holds gh's `hosts.yml` and goose's `config.yaml` (provider keys), which is credential and config state that one login is meant to propagate, not session state. `XDG_CONFIG_HOME` is deliberately not set.
- `.claude`, `.gemini`, `.codex`, `.copilot`, `.bob`: the credential directories `SharedAuth` names, unchanged.
- `~/.local/share/opencode`: `opencode auth login` writes its credential to `$XDG_DATA_HOME/opencode/auth.json`, so this one entry is bridged from the per-agent data home back to the shared tree (`xdgDataSharedBridges`). Sharing inside `.local` is now a named list rather than the default.
- `.cache`: unchanged, caches gain nothing from isolation.
- The Dockerfile umask wrappers stay: they protect the still-shared credential directories above, not `.local`.

## Migration

Nothing is migrated. Existing state under `/data/home/.local` is left where it is, the shared tree stays in `WatchedHomeDirs`, and each agent starts fresh per-agent state on its next session. A real `.local` that already exists in a per-agent home is never clobbered. `HIVE_SHARED_AGENT_HOME=1` keeps the legacy layout whole, with no XDG vars exported.

## Tests

- `TestAgentEnvPairs_XDGStateHome_IsPerAgent`: two launched agents resolve to different `XDG_STATE_HOME` and `XDG_DATA_HOME` values, each under its own home and never under the shared `.local`.
- `TestAgentEnvPairs_XDG_NotExportedUnderSharedHome`: no export for UID-less agents or under the escape hatch.
- `TestSetupAgentXDGDirs_RetiresLegacyBridgeAndSharesOnlyNamedEntries`: a pre-existing `.local` bridge is replaced by a real directory, shared goose sessions are neither leaked nor disturbed, the opencode bridge is created, and re-provisioning is idempotent.
- `TestSetupAgentXDGDirs_KeepsExistingRealLocal`: live per-agent state survives provisioning.

Fixes #6238


clubanderson:
Review (as a comment). XDG_DATA_HOME/XDG_STATE_HOME are exported only when perAgentXDGHome says HOME is per-agent, so HIVE_SHARED_AGENT_HOME=1 keeps the legacy layout. .local leaves the bridge list and only opencode is bridged back, because its auth login credential lives under XDG_DATA_HOME. ~/.config and the SharedAuth credential dirs are untouched. One coupling to know: state/goose/logs/cli is pre-created because goose panics without it - that path is goose's and will need updating if goose moves it.

PR #6279, opened 2026-09-08
🔒 security(github): require positive CI confirmation before merge - absent check runs are not passing

### LINKED ISSUE #6173: merge-request path merges without verifying CI: absent check runs are treated as passing
## Summary

The merge-request path performs **no independent verification of CI**. It calls
`MergePR` and only reacts if *GitHub itself* refuses the merge. On a branch with
no protection rules there is nothing to refuse it, so an agent can merge a PR
whose checks failed — or never produced a check at all — with no gate anywhere
in Hive.

Verification is effectively delegated to branch protection, and the **absence**
of branch protection silently removes the only gate rather than failing closed.

## Where

[`src/pkg/github/merge_request_watcher.go`](https://github.com/hivecommons/hive/blob/fe34da51434ad9b0924eee1492047c3c95c705ff/src/pkg/github/merge_request_watcher.go)

The merge is issued unconditionally at
[L261](https://github.com/hivecommons/hive/blob/fe34da51434ad9b0924eee1492047c3c95c705ff/src/pkg/github/merge_request_watcher.go#L261):

```go
res, err := c.MergePR(ctx, req.Repo, req.Number, req.Method, req.ExpectSHA)
```

and the only check-awareness in the file is *post-hoc classification of a
refusal* at
[L125](https://github.com/hivecommons/hive/blob/fe34da51434ad9b0924eee1492047c3c95c705ff/src/pkg/github/merge_request_watcher.go#L125):

```go
for _, want := range []string{"required status check", "required status checks",
    "changes must be made through a pull request", "expected",
    "has not succeeded", "checks have not"} {
```

That is a sound way to *classify* a rejection — it is generic and names no
specific linter, which is the right call. The problem is that it is the only
check-related logic on this path: grepping the file for `CheckRuns`,
`ListCheckRunsForRef` or `commitGreen` returns nothing.

`auto_merge.required_checks` does not cover this. It gates `commitGreen` in the
self-merge sweep, which is a different path.

## Reproduction (observed in production)

A fork PR of +2480/−3067 across 48 files was merged **13 minutes** after opening,
deleting an entire content directory. The base branch had no protection.

The reason it looked clean is worth recording, because it also defeats the
obvious fix of "read the PR's check rollup":

All four workflows on the head SHA concluded **`failure`** — but every one had
**`jobs: 0`**. They were *startup* failures, which produce a failed **workflow
run** and **zero check runs**:

```
GET /repos/:o/:r/commits/:sha/check-runs      -> total_count: 0
PR statusCheckRollup                          -> []
GET /repos/:o/:r/actions/runs?head_sha=:sha   -> 4 runs, all conclusion=failure
```

So any gate that asks "is a check failing?" sees nothing and concludes the commit
is clear. **Absent is not passing.**

Hive already draws this distinction elsewhere and gets it right: the rotation
probes treat an unmeasured provider as explicitly *not* exhausted rather than
folding the two together. The merge path does not make the same distinction for
checks.

## Suggested fix

1. Before merging, list **workflow runs** for the head SHA in addition to check
   runs, and refuse on any `conclusion: failure` — including runs with zero
   jobs, which emit no check run.
2. Treat **zero check runs and zero workflow runs** as *unverified* and refuse,
   rather than as clear. An explicit per-repo opt-out keeps that from blocking
   repos that genuinely have no CI.
3. Consider refusing to merge into an unprotected branch unless explicitly
   allowlisted. The current design assumes protection exists; where it does not,
   Hive is the only thing standing between an agent and `main`, and today it
   does not look.

Happy to send a PR for (1) and (2) if the approach seems right.

clubanderson:
## Failure mode

`handleOneMergeRequest` in `src/pkg/github/merge_request_watcher.go` called `MergePR` with no CI verification of its own. The only check-aware code on the path was `isRequiredCheckMergeBlocker`, which classifies GitHub's refusal string after the fact. Enforcement was therefore delegated entirely to branch protection, and on a base branch with no protection rules there was no gate anywhere: a PR whose checks failed merged on request. Worse, the production case in #6173 produced no failing check at all. Every workflow on the head SHA concluded `failure` with zero jobs (startup failures), so the commit had zero check runs and an empty status rollup, and any gate that asks "is a check failing?" would have seen nothing. Absent was treated as passing.

Confirmed present on current v4 (`81fc9653`): merge_request_watcher.go:261 issued the merge unconditionally; `grep CheckRuns\|ListCheckRunsForRef\|commitGreen` on the file returned nothing.

## Fix

New `src/pkg/github/merge_ci_gate.go` computes a verdict on the pinned head SHA before any merge is attempted, and the watcher acts on it:

- green: at least one commit status or check run exists, every gating one concluded success (neutral/skipped accepted; the existing non-required ignore list and the `auto_merge.required_checks` set still apply unchanged), every config-declared required check has actually reported, and no workflow run on the SHA failed without producing a job. The merge proceeds.
- pending: a gating check is queued or in progress, a required check has not been created yet, or a workflow run is in flight with no job. The request is parked (`ci_waits` in the result file) without consuming a retry, bounded by `mergeRequestMaxCIWaits` (one hour at the default tick).
- red: a gating check failed, or a workflow run concluded failure with zero jobs. Refused as a failed attempt; at the retry limit it is classified and re-engaged into the fix loop exactly like a branch-protection refusal.
- unverified: zero statuses, zero check runs, zero workflow runs. Refused; quarantined at the retry limit with the reason recorded.
- an API error while gathering evidence is a failed attempt, never a pass.
- a pinned SHA that no longer matches the PR head is refused: CI of a different commit never authorizes a merge.

Workflow runs that did produce jobs are judged through their check runs (so an optional suite failing with jobs does not re-block a merge the ignore list already allows); only zero-job runs are treated as opaque, which is exactly the shape that defeated the check-run view.

`commitGreen` (self-merge sweep) is refactored onto a shared `commitCIState` core that additionally reports evidence counts and unreported required checks. The sweep's own behavior is unchanged; no existing gate was removed.

## Invariant the tests assert

`merge_ci_gate_test.go` runs the real handler against a fake GitHub with a counting `PUT /merge`. In every refusal case the assertion is that the merge call count is zero, not that a warning was logged: zero check runs, all queued, mixed success plus failure, zero-job workflow failure, missing required check, moved head, and an evidence API error. All-success asserts exactly one merge and a consumed request. Pending cases additionally assert the request stays live with no retry consumed and merges once CI later reports green; the pending budget is asserted to convert into a failed attempt.

Docs: `src/docs/hive-merge.md` describes the gate. Changelog fragment: `changelog.d/security-6173-merge-requires-positive-ci.md`.

Follow-ups not in this PR: a per-repo opt-out for repositories that genuinely have no CI (the issue's suggestion 2b) and refusing unprotected base branches unless allowlisted (suggestion 3). Both are policy additions; this PR closes the fail-open hole.

Fixes #6173


clubanderson:
Review (as a comment - GitHub blocks approving one's own PR, labels carry the gate). verifyMergeRequestCI runs before MergePR and every non-green verdict calls recordMergeFailure then returns, so MergePR is reachable only on the green path. Verdict ordering is right: opaque zero-job failures and non-pending failures are red, queued/missing-required/opaque-pending wait, and zero evidence is unverified - never green. Every error path returns unverified. A moved head is refused. The tests count PUT /merge calls and assert zero in each refusal case, which is the invariant that matters. Follow-ups tracked in #6281.

Issue #4001, opened 2026-08-17
Adopt state-triggered hooks as a first-class extension point


clubanderson:
## Summary

Add **hooks: operator-defined actions triggered by agent/governor state transitions**, as a first-class extension point. Today hive's reactions to state (pause, resume, ACMM changes, sweep events, stalls) are hard-coded across the codebase; a declarative hooks mechanism would let behavior attach to well-defined transitions without patching core.

## Why it fits hive

Hive already emits and reacts to a rich set of agent/governor/sweep events. A **hooks** registry keyed on those transitions would give:

- Named, hookable transitions: `on_agent_paused`, `on_agent_resumed`, `on_sweep_merge`, `on_acmm_change`, `on_stall_detected`, `on_pr_opened`, etc.
- Operator-defined actions (notify, run a script, adjust pacing) triggered declaratively instead of via code changes.
- A single, auditable place where "what happens when X" is defined — improving testability and removing scattered special-cases.

## Proposed scope

- Define the set of hookable state transitions (start from the governor + sweep + agent-lifecycle events hive already emits).
- A config-declared hook registry mapping a transition → action, under the same guardrails as everything else (no secrets, respects ACMM level, output goes to the audit log).
- Opt-in and side-effect-safe by default.

## Notes

Architectural enhancement; no external dependency. Related: #4000 (tool-approval as an operation).

clubanderson:
Maintainer-side response, consistent with the review posted on #4002: endorse — with the caveat stated up front that this is, by construction, an **arbitrary-code-on-state-transition surface**, so the security model matters more than the mechanism. #4002's turn/ops model is what makes these transitions explicit and durable; hooks are the extension point over exactly those transitions and should be specified against them, not against today's implicit ones.

## The transitions already exist — implicitly, each hand-rolled

- **Governor mode changes**: `ModeChange{From, To}` appended to `modeHistory` in `src/pkg/governor/governor.go` — recorded, not hookable. Likewise `BudgetTransitions` out of `UpdateBudgetFromTotals`.
- **Pause/resume**: the #3836 upgrade kill switch (`src/pkg/hub/upgrade_pause.go`) is precisely a hand-rolled instance of what this issue generalizes — a durable state flip (`setUpgradePause`) whose effects then had to be manually threaded through every delivery path (the file's own header notes the pause must be honoured by all of them or it is a lie). A hooks mechanism is how the *next* such switch does not require finding every call site by hand.
[...]

Issue #4000, opened 2026-08-17
[v5] Adopt tool-approval as an explicit, ACMM-gated operation


clubanderson:
## Summary

Model **tool approval as an explicit, ordered operation** in the agent turn — each tool request resolves through one of: auto-approve, security-scan, or operator-approval — rather than being decided ad-hoc inside the loop.

## Why it fits hive

Hive already gates what agents may do by **ACMM level** and enforces security/authorship guardrails, but those gates are spread across the loop. Making tool approval a **first-class, inspectable operation** would let hive:

- Resolve every tool request through a single, testable decision point: `auto-approve | security-scan | operator-approve | deny`.
- Key the decision on **ACMM level** (e.g. L4 → operator-approve for side-effectful tools; L6 → auto-approve on green) instead of ad-hoc checks scattered through the loop.
- Route the "security-scan" path through the existing sec-check surface *before* a tool runs, not after.
- Produce an auditable record of *why* each tool call was allowed or blocked.

## Proposed scope

- Introduce a tool-approval decision function taking (tool request, ACMM level, agent identity) → verdict.
- Wire the existing ACMM gates + sec-check into it as the policy source.
- Emit the verdict + rationale to the audit log.

## Notes

Architectural enhancement; no external dependency. Pairs with a state-triggered hooks mechanism and the broader re-entrant turn model tracked in sibling issues.

clubanderson:
Maintainer-side response, following on from the review posted on #4002: this issue should be read as **one operation in that RFC's ordered ops list**, not a standalone mechanism. Reviewed in that frame, the answer is yes — and the codebase already makes the argument for it.

## The scattered-gates diagnosis is verifiably true

We do not have one decision point today; we have at least four, each grown out of its own incident:

- **Self-merge**: `SelfMergeMinACMMLevel` / `AutoMergeConfig.SelfAuthoredAutoMergeAllowed` in `src/pkg/config/config.go`. The comment on that constant records the history: the self-authored sweep *originally had no ACMM check at all*, which is how an L4 hive wrongly self-merged its own PR. The gate was bolted on afterward.
- **Plan approval**: `PlanAutoApproveForLevel` in `src/pkg/config/acmm_packs.go` — a separate lookup consulted only by the decomposition path.
- **Plan-from-label**: `PlanningConfig.PlanFromLabelEnabled(acmmLevel)` in `config.go` — a third, independent level check.
[...]

PR #6280, opened 2026-09-08
[architect] fix: emit review_rejected hook at the rejection point (v5)

### LINKED ISSUE #6259: [architect] review_rejected hook pipeline has no emitter — hooks.EmitReviewRejected is unreachable, so configured hooks never fire
## Architecture Finding

**Type**: tech-debt / dead-on-arrival wiring gap
**Affected area**: src/pkg/hooks/review_rejected.go (EmitReviewRejected, NewReviewRejectedPayload, ModelKnobURL), src/pkg/hooks/dispatcher.go

pkg/hooks/review_rejected.go documents itself as 'the FIRST SHIPPED HOOK end-to-end' (RFC #4001): a human rejecting a review's output should emit a `review_rejected` transition carrying the producing agent's backend/model/pin metadata, deep-linking the owner to the model knob. Config plumbing exists (config.go HookConfig `on: review_rejected`), cmd/hive wires registration/CEL/rate-limits (hookwire_test.go), and the dispatcher/audit pipeline is in place.

But `EmitReviewRejected`, `NewReviewRejectedPayload`, and `ModelKnobURL` are unreachable from any binary (verified with `deadcode` at 321251cc): **no production code path ever emits the transition**. An operator who configures a `review_rejected` hook gets a validated, registered hook that can never fire — the exact 'hunt through the admin UI' failure the RFC set out to fix persists.

## Impact

The hooks subsystem's first (and only) end-to-end transition is emitter-less, so the whole pipeline (registration, CEL predicates, rate limiting, audit) ships as untriggerable scaffolding. Future hooks will copy this shape and inherit the same gap.

## Recommendation

Identify the review-rejection surface (dashboard review verdict handler / review queue rejection path) and add the single `hooks.EmitReviewRejected(...)` call site with the producing agent's semconv model metadata. Until then, either mark the hook type experimental in config validation or warn when a `review_rejected` hook is configured.

---
*Filed by architect agent (ACMM L5 — hold-gated mode)*
---
🐝 **Hive Agent**: `architect` | **Instance:** `hosted-available-oke-11-placeholder-r05x` | **SHA:** `321251cc`

— hive: agent=architect backend=copilot model=claude-opus-4-6

clubanderson:
## What

Wires the missing emitter for the `review_rejected` hook (RFC #4001). The transition, payload, model-knob deep link and notification rendering all existed in `src/pkg/hooks/review_rejected.go`, but nothing in any binary called `hooks.EmitReviewRejected`, so an operator who configured `on: review_rejected` got a validated, registered hook that could never fire.

## Where the rejection happens

The emitter lands at the approval desk (RFC #4000), which is the one surface where a human explicitly rejects an agent's queued output through hive itself. A denial there has everything the hook contract needs already recorded durably by `toolapprove.Inbox.Resolve`: the resolving owner (actor), the stated rationale (reason), the requesting agent, and the target repo and PR number.

- `POST /api/approvals/resolve` with `approved: false` emits one `review_rejected` after the denial is journaled.
- `POST /api/approvals/bulk` with `approved: false` emits one per denied item, since a bulk deny is N individual rejections.
- A replayed resolve (409) and an ID another operator already resolved emit nothing, so it is exactly one transition per rejection.

The payload carries `agent`, `repo`, `attrs.pr`, `actor`, `reason`, the agent's effective `backend` / `model` / `pin` (runtime override, then the operator's pin, then static config, so the manager-unknown case still names what it can), `acmm_level`, and `attrs.model_knob_url` built from `dashboard.public_url` with the existing `hub.dashboard_url` and request-host fallbacks.

## Emitter signature

`hooks.EmitReviewRejected` now takes a `hooks.FireFunc` instead of a `*Dispatcher`. The dashboard reaches the process-wide dispatcher through `Dependencies.HookFire` (the same seam `acmm_level_change` uses) and must not import `cmd/hive`, so a fire callback is the shape the emitter has to accept. `(*Dispatcher).Fire` satisfies it directly and a nil fire remains a no-op. Existing hooks tests updated to pass `d.Fire`.

## Tests

`src/pkg/dashboard/api_approvals_review_rejected_test.go`:

- a single denial emits exactly once with the full payload asserted field by field, and a replayed denial adds nothing
- an approval emits nothing
- a bulk deny emits once per denied item with distinct PRs and skips an ID that did not resolve
- an agent unknown to the manager falls back to static config with no pin
- a nil `HookFire` still journals the denial and does not panic

## Docs

`src/docs/hooks.md` transition catalog now names the concrete surface for `review_rejected`. Changelog fragment added under `changelog.d/`.

Fixes #6259


clubanderson:
Review (as a comment). Emission happens only after inbox.Get snapshots the pending row and Resolve journals the denial; the 409 replay path returns before it, so it is exactly once per rejection and never on approve. The approval desk is the right surface - the one place a human rejection is durable with actor and rationale recorded. The EmitReviewRejected signature change (Dispatcher to FireFunc) is a public-API change in pkg/hooks on v5; every existing caller was updated to d.Fire, nil stays a no-op.

PR #6275, opened 2026-09-08
[architect] fix: self-authored auto-merge sweep enforces intent tier gates (v5)

clubanderson:
## Summary

The App self-merge sweep (`trySweepSelfAuthoredPR` in `src/pkg/github/automerge/automerge_sweep.go`) merged every open, non-draft, App-authored, mergeable, CI-green PR after a head-SHA re-check, without ever asking the intent tier question. `intent.EvaluateForAppSelfMerge`, the tier gate written for exactly this path, was unreachable from any binary.

## The bypassed gate

The human merge-queue lane is tier-gated upstream: `writeIntentVerdicts` classifies each agent PR, builds linked-issue and approved-plan evidence, and `writeMergeEligible` drops any PR whose verdict is not merge-allowed when `intent.enforce` is on, so nothing a merger can queue reaches `trySweepQueuedPR` at a refused tier. The self-authored sweep lists the App's PRs independently of `merge-eligible.json`, so that gate never applied to it. On an L6 hive with enforcement on, a Tier1 App PR with no linked issue merged on green.

The hold and do-not-merge label guard the issue also asks for already exists on v5 (#5589) and is unchanged here.

## The fix

- `intent.Verdict.BlocksMerge(enforce)` is now the single refusal predicate. `writeMergeEligible` uses it, and so does the self-authored sweep, so the two lanes cannot drift apart.
- `automerge.IntentGate` (installed via `Options.IntentGate` or `Engine.SetIntentGate`) carries the intent config, an enforce reader (read per evaluation so config reloads apply, like `MergerAuthorizer`), and the bead stores. `trySweepSelfAuthoredPR` fetches the complete changed-file list, classifies the PR, builds evidence with the same `BuildEvidenceForRepo` call the human lane uses, calls `EvaluateForAppSelfMerge(class, evidence, true)`, and withholds the merge when `BlocksMerge` reports refusal. The gate runs after `commitGreen` and before the approval desk, so the desk still only sees requests policy permits.
- An unavailable or incomplete changed-file list fails closed under enforcement (the tier cannot be established). With `intent.enforce` off the gate is advisory, matching `writeMergeEligible`, so the sweep is not stricter than the normal path either.
- `notifywire.go` installs the gate from `cfg.Intent` and the bead stores.
- `automerge.ListChangedFiles` is factored out and shared with `fetchIntentPREvidence` instead of duplicating the paging and completeness check.

## Invariant tested

[...]

PR #6291, opened 2026-09-08
🔒 security(github): port the positive-CI merge gate to v4 - absent check runs are not passing

clubanderson:
## What

v4 port of #6279 (merged to v5 as 5a992c93). v4 is the production line and its merge-request path still calls `MergePR` unconditionally: the exact fail-open hole #6173 describes. Confirmed on current v4 head `18ab8f7e`: `merge_request_watcher.go` issued the merge with no CI verification of its own, so on a base branch with no protection rules a red PR, or one whose workflows all died at startup with zero jobs (zero check runs, empty status rollup), merged on request.

## Fix

New `src/pkg/github/merge_ci_gate.go` computes a verdict on the pinned head SHA before any merge is attempted, and the watcher acts on it:

- green: at least one commit status or check run exists, every gating one concluded success (neutral/skipped accepted; the existing ignore list and `auto_merge.required_checks` set still apply unchanged), every config-declared required check has reported, and no workflow run on the SHA failed without producing a job. The merge proceeds.
- pending: a gating check is queued or in progress, a required check has not been created yet, or a workflow run is in flight with no job. The request is parked (`ci_waits` in the result file) without consuming a retry, bounded by `mergeRequestMaxCIWaits` (one hour at the default tick).
- red: a gating check failed, or a workflow run concluded failure with zero jobs (the production shape in #6173). Refused as a failed attempt; at the retry limit it is classified and re-engaged into the fix loop exactly like a branch-protection refusal.
- unverified: zero statuses, zero check runs, zero workflow runs. Refused; quarantined at the retry limit with the reason recorded.
- an API error while gathering evidence is a failed attempt, never a pass.
- a pinned SHA that no longer matches the PR head is refused.

`commit_ci.go` (`EvaluateCommitCI`, `RequiredStatusCheckContexts`) is the shared evaluation core taken verbatim from v5; the self-merge sweep's `commitGreen` and `requiredStatusCheckContexts` delegate to it with behavior unchanged. `recordMergeFailure` is extracted from the watcher so the gate's refusals share the retry / exhaust / re-engage policy. Docs updated in `src/docs/hive-merge.md`; fragment at `changelog.d/security-6173-merge-requires-positive-ci-v4.md`.

## Adaptations versus the v5 commit

Everything else applied verbatim with `git cherry-pick -m 1 --no-commit 5a992c93`. Two files needed attention:

1. `src/pkg/github/automerge_sweep.go`. On v5 the sweep moved into the `pkg/github/automerge` subpackage (#5983); on v4 it is still at `src/pkg/github/automerge_sweep.go`. The `commitGreen` -> `EvaluateCommitCI` refactor is applied at v4's path on `*Client` (using `c.client`, no cross-package import), and the file is not moved. v5's `required_checks.go` and its `Client.requiredChecks` field addition are dropped because v4's `Client` already has `SetRequiredChecks`, `configRequiredChecks` and `requiredChecksMu` in `automerge_sweep.go`.

2. `src/cmd/hive/notifywire.go`. This file does not exist on v4 (it arrived with #6275, which v4 does not have). The only change #6279 made there was a `ghClient.SetRequiredChecks(set)` call so the watcher's gate sees the declared required-check set. On v4, `main.go` already installs that set on the same `ghClient` that runs `StartMergeRequestWatcher`, at boot and on every config reload, so the gate reads it through `configRequiredChecks()` with no wiring change. The file is omitted; nothing from #6275 is ported.

## Invariant the tests assert

`merge_ci_gate_test.go` runs the real handler against a fake GitHub with a counting `PUT /merge`. In every refusal case the assertion is that the merge call count is zero: zero check runs, all queued, mixed success plus failure, zero-job workflow failure, missing required check, moved head, and an evidence API error. All-success asserts exactly one merge and a consumed request. Pending cases assert the request stays live with no retry consumed and merges once CI later reports green; the pending budget is asserted to convert into a failed attempt. Existing watcher and loop tests serve a green fixture so they keep exercising the paths they were written for.

Refs #6173


clubanderson:
Review (as a comment - GitHub blocks self-approval; labels carry the gate). This is the v4 port of #6279 and it preserves the invariant: verifyMergeRequestCI runs before MergePR, zero evidence is unverified and refused, a workflow run that failed with zero jobs is red, every error path is unverified, and the ported tests count PUT /merge calls and assert zero in each refusal case. Both conflict points were resolved for the right reasons. automerge_sweep.go stays at v4's path with commitGreen delegating to the shared EvaluateCommitCI; the duplicate required_checks.go and client.go field the raw cherry-pick would have added were correctly dropped since v4's Client already carries SetRequiredChecks and configRequiredChecks. notifywire.go was omitted because it does not exist on v4, and the one thing #6279 needed there is already wired: main.go calls SetRequiredChecks at boot (line 1947) and on every config reload on the same ghClient that starts the merge watcher (line 1927), so the gate reads the required-check set through configRequiredChecks with no new wiring. Nothing from #6275 was pulled in. This closes the v4 half of #6173; v4 is the production line and was unprotected until now.

PR #6303, opened 2026-09-08
🔒 security: make the token-access audit log non-writable by agents

### LINKED ISSUE #6287: [sec-check] token-access audit log is agent-writable (dev:node 0664) — audited agents can truncate or forge the trail behind GET /api/token-access
## Security Finding

**Severity**: medium
**Type**: permission-issue / tamperable audit trail (CWE-284 adjacent, CWE-117/CWE-345)

`src/deploy/entrypoint.sh:791-793` (added in #6106 / commit 87b2b02e) pre-creates the token-access audit log as:

```
touch /var/run/hive-metrics/token-access.jsonl
chown dev:node /var/run/hive-metrics/token-access.jsonl
chmod 664 /var/run/hive-metrics/token-access.jsonl
```

Every per-agent user is created with primary group `node` (`src/deploy/entrypoint.sh:1355`: `useradd --system -u $AGENT_UID -g node ...`). Group-write (0664) therefore gives **every agent UID full write access** — not just append. `bin/gh-wrapper.sh:122` and `bin/git-credential-hive.sh:166` legitimately append as the agent UID, but nothing restricts agents to appending: `> file`, `truncate`, or rewriting arbitrary JSONL lines all succeed.

The dashboard explicitly treats this file as an operator-only security audit trail: `src/pkg/dashboard/api.go:2171` gates `GET /api/token-access` at owner role citing #3936 / CWE-284, because it 'records every gh CLI command an agent issued'. That read-side control is undermined when the audited parties control the write side.

## Impact

A compromised or prompt-injected agent can erase the record of its own token usage (truncate the file), forge entries attributing gh/git activity to other agents, or flood the log to rotate real entries past the 100-entry window served by the API. Operators investigating token misuse via `/api/token-access` would see a trail the attacker curated.

## Recommendation

Pick one, in order of preference:

1. **Route audit appends through the hive process** (e.g., a small setgid-free local socket/endpoint the wrapper POSTs to), so only the hive UID writes the file; drop group-write to 0644.
2. **Append-only enforcement**: after touch/chown, set `chattr +a` in the entrypoint (root phase) where the filesystem supports it, falling back with a logged warning. Agents keep working (`>>` still succeeds) but truncation/overwrite fails.
3. At minimum, treat the log as advisory rather than an audit control and note the tamperability beside the CWE-284 comment in `api.go`.

---
---

clubanderson:
## Problem

GET /api/token-access is the owner-only audit trail of every gh CLI command and git credential lookup an agent makes (gated at owner role citing CWE-284). It was written by the audited parties. bin/gh-wrapper.sh and bin/git-credential-hive.sh run as the agent UID and appended straight into /var/run/hive-metrics/token-access.jsonl. That only works when the file is writable by every agent: v4 pre-creates it dev:node 0664 (#6106), and every agent's primary group is node. Append cannot be distinguished from write at the permission level, so a compromised or prompt-injected agent could truncate the trail of its own token use, rewrite lines to attribute its activity to a peer, or forge entries outright. On v5 the pre-create block never landed, so the appends failed silently and the trail was empty instead; either way the audit control was not one.

## Mechanism

The fix reuses the drop-box relay every other agent-to-hive write already uses (pr-requests, issue-requests, merge-requests), which is the issue's first recommendation: route appends through the hive process so only the hive UID writes the file.

- The wrappers now write one JSON event per call into /var/run/hive-metrics/token-access-events, write-then-rename under a pinned umask 027 so the event is 0640 in a setgid directory (the hive can read it, nobody else can). The spool is 0730 with sticky and setgid: agents can create and rename their own files, cannot list the directory, and cannot unlink a peer's event. The path deliberately has no environment override, because an override would let an agent redirect its own audit events into the void.
- A new ingester in pkg/github (token_access_audit.go), started unconditionally at boot next to PrepareRequestDirs, folds events into the log every two seconds. The log is opened O_APPEND by the hive process and held at hive-owned 0600, no group or other bit at all, since group node is the agents.
- On ingest the event's self-reported uid is replaced by the uid that owns the event file, the same trust anchor the PR-request watcher uses, and a disagreeing claim is preserved as claimed_uid. An entry forged in a peer's name lands attributed to the forger, with the forgery itself on the record.
- Events that are not a single JSON object, are empty, or exceed 8 KiB are rejected and removed rather than appended, so the dashboard only ever hands back lines it can serve as raw JSON. Abandoned .tmp files older than ten minutes are swept.
- The entrypoint root phase pre-creates the log dev:node 0600 and the spool 3730 on every boot, so a 0664 file left by an older image is tightened rather than trusted. The ingester re-asserts 0600 on every open as well. I read the #6238 / #6277 umask-wrapper work and the Go permissions watcher: the watcher's roots never include /var/run/hive-metrics, so nothing running as dev after boot can loosen the file, and the new static test pins that.

What stays open, and is narrower than before: an agent can still delete its own event during the two-second window before ingest (a hive-side unix socket with SO_PEERCRED would close that and is the natural follow-up), and it can still flood the log with real gh calls. It can no longer alter anything already recorded.

## Invariant tested

The tests assert the property, not that a warning was logged:

- pkg/github/token_access_audit_test.go: after PrepareTokenAccessAudit and after every ingest pass, the log has exactly mode 0600 (no group or other bits) and is owned by the hive process uid; a pre-existing 0664 log is tightened without losing entries; an event claiming another uid is recorded under the owning uid with the claim kept as claimed_uid; malformed, empty and oversized events never reach the log; in-flight .tmp files are left alone until stale; the watcher ingests on its ticker and stops on cancel, and disables itself when the spool cannot be created.
- src/deploy/test_entrypoint_token_access_audit.sh (wired into v2-ci.yml): the entrypoint chmods the log to 600 and nothing else, no chmod in it can grant group/other write on the log, the spool is a 3730 drop-box, neither wrapper references the log path or allows a spool override, the Go constants agree with the shell, and the permissions watcher does not walk /var/run/hive-metrics.
- bin/test_git_credential_hive.sh: a real get drops exactly one 0640 event into the spool with op, agent and host, leaves no .tmp behind, never opens the log, and an unwritable spool neither blocks the credential nor leaks EACCES to stderr.

Fixes #6287


clubanderson:
Review (as a comment - GitHub blocks self-approval; labels carry the gate). The invariant is asserted correctly: the test fatals if the log mode is not exactly 0600 and separately checks that no group/other bits survive (perm & 0o077 == 0), after prepare and after every ingest, so an agent uid has no access at all. The design is sound - agents write into a 3730 sticky spool they cannot list or unlink peers from, only the hive opens the log with O_APPEND, and the uid on each event is replaced by the spool file's owning uid on ingest so a forged claim is attributed to the forger as claimed_uid. The finding that v5's trail was silently EMPTY (the >> into a dev-owned 0755 dir was failing) matters more than the issue title: it was never an audit control on v5 at all. The residual self-retraction window inside the 2s ingest is tracked in #6306; the important property - nothing already recorded can be altered - is in.

PR #6211, opened 2026-09-07
✨ feat: per-repo agent pause — quiet one repo without stopping the hive

Danathar:
## Summary

You can now **pause one repository** instead of pausing every agent or deleting
the repo from `project.repos`.

A paused repo goes quiet — agents can't write to it and aren't given work on it
— but it stays in your config, keeps its dashboard card, keeps its issue/PR
counts, and keeps its ACMM score. Every other repo in the hive keeps working.

Pause it from the new **⏸ pause** button on its REPOSITORIES card, from
`POST /api/repos/pause`, or in config as `project.paused_repos`. Each pause
records **who paused it, when, and why**, and survives restarts.

Nothing changes for a hive that never uses it: the config field is optional and
absent means "nothing paused".

Closes #6203.

## Why

Today there are two levers and neither fits:

- Stop an **agent** — pauses that agent on every repo.
- Stop **everything** — the fleet breaker.

So a release freeze on one repo, an incident, a repo whose CI is red for an
unrelated reason (agents keep filing the same failure, burning budget), or a repo
declared in config but not yet ready for agents all cost the whole fleet. The
workaround — deleting the repo from `project.repos` — is not a pause: it drops
the repo from the dashboard and the ACMM eval, records nothing about who did it
or why, is easy to forget to reverse, and reads as permanent.

Pause is a **run-state**, orthogonal to both dimensions the hive already has. It
is not the ACMM level: no level means "off" — L1 still runs agents and
brainstorm is advisory at every level — so lowering a repo's autonomy cannot
silence it. Per #6111's own words, per-repo enablement "is a different feature
and should be its own proposal"; this is that feature, and it composes with
per-repo levels if they land (a paused repo runs nothing regardless of level).

## What an operator sees

**Dashboard** — each REPOSITORIES card gains an owner-only **⏸ pause** /
**▶ resume** button. Pausing asks for a reason. A paused card is tinted, dims its
counts, and carries a **⏸ PAUSED** pill whose tooltip names who, when and why.
The card does not disappear — that is the whole point.

**Config**

```yaml
project:
  org: acme
  repos: [console, dashboard]
  paused_repos:
    - repo: dashboard
      by: bketelsen
      at: 2026-09-07T10:04:00Z
      reason: release freeze until Friday
```

Only `repo` is required. Bare (`dashboard`), cross-org (`laredo/cuga-agent`) and
any casing all match the same repository.

**API** — `POST /api/repos/pause`, `POST /api/repos/resume` (owner-only), and
[...]

PR #6307, opened 2026-09-08
🌱 hub: digest pin as a first-class run-state - rollback that survives the channel re-arm

### LINKED ISSUE #6267: 🐛 tracked-channel re-arm undoes a hand-pinned digest rollback on the next heartbeat
## Summary

A digest-pinned rollback on a channel-tracking hive does not stick. The tracked-channel re-arm in `src/pkg/dashboard/server.go` (around line 2651) honours only the admin upgrade pause, so a hive an operator has hand-pinned to a prior immutable digest is dragged back to the channel head on its next heartbeat unless spokes were paused first, or the hive was first switched to a plain branch.

Surfaced while writing the digest-verifiable rollback runbook for #6113 (PR #6266), which currently has to instruct the operator to pause automation before pinning and to leave the pin in place only while paused.

## Why this matters

Rollback is a safety procedure. A rollback that silently reverts on the next heartbeat is worse than no rollback, because the operator believes the fleet is on the known-good digest when it is not. This also makes the v5 GA bar's digest-verifiable-rollback row (#6016) satisfiable only with a caveat.

## Suggested fix

Treat an explicit digest pin as a run-state the re-arm must respect, independent of the admin upgrade pause. Either record the pin (who/when/digest) on the hive and have the re-arm skip hives that carry one, or make a pinned digest an explicit opt-out that must be lifted by the operator before channel tracking resumes. The per-repo pause work (#6203) established the provenance-carrying run-state pattern; this is the same shape applied to image pinning.

Refs #6113, #6016.

### LINKED ISSUE #6268: 🌱 no hub API to pin a hosted spoke to an image digest - rollback requires a raw kubectl bypass
## Summary

There is no hub API to pin a hosted spoke to an arbitrary image digest or SHA. `switch-branch` accepts branches and channels; `upgrade` targets latest. So the only way to roll a hosted spoke back to a specific known-good digest is a raw `kubectl set image deployment/hive "*=...@sha256:..."` in the `hive-hosted-<id>` namespace, reaching around the hub entirely.

Surfaced while writing the digest-verifiable rollback runbook for #6113 (PR #6266), which currently documents the `kubectl` path because no hub path exists. That is the same object and `*=` form the hub's own upgrade path writes, so it works, but it is an operator bypass.

## Why this matters

Rollback that bypasses the hub leaves no record in the hub, no provenance, and no way for the dashboard to show that a spoke is pinned. It also cannot be done by an operator who has hub admin but not cluster `kubectl` access, which is the intended separation for hosted hives. The v5 GA bar's digest-verifiable-rollback row (#6016) is currently met only via this bypass.

## Suggested fix

Add a hub endpoint (or extend `switch-branch`) that accepts an immutable short-SHA digest and applies it via the same deployment-patch path the upgrade code already uses, recording who/when/digest on the hive. Pair it with the re-arm change in the companion issue so the pin is not immediately undone.

Refs #6113, #6016.

### LINKED ISSUE #6290: [strategist] hosted-spoke rollback is not a first-class operation — #6267/#6268 cluster leaves digest rollback as an unsupported kubectl bypass
## Strategic Finding

**Type**: roadmap-gap
**Horizon**: near-term

Two independent bug/feature reports now describe the same missing capability from opposite ends:

- #6268 — there is no hub API to pin a hosted spoke to an image digest; rolling back requires a raw `kubectl` bypass of the hive's own management plane.
- #6267 — even when an operator hand-pins a digest that way, the tracked-channel re-arm undoes the rollback on the next heartbeat.

Together these mean **rollback of a hosted spoke is effectively unsupported**: the only path is out-of-band, and the system actively reverts it. This also connects to the release-docs gap already tracked in #6113 (no digest-rollback runbook can be written while the underlying operation doesn't exist).

## Rationale

Hosted spokes are the managed-offering story. An operator who ships a bad image today has no sanctioned way to get a spoke back to a known-good digest — the exact moment a management plane must be trustworthy. This is an adoption and trust risk for the hosted path and is adjacent to the v5 GA operational-readiness rows (#6016, #6112, #6113). Fixing #6268 without #6267 (or vice versa) leaves rollback broken; they need to be sequenced as one capability, not two unrelated bugs.

## Proposed Next Step

Maintainer names "hosted-spoke rollback" as a single roadmap item covering, in order:
1. hub API to pin a spoke to a digest (#6268),
2. heartbeat/re-arm honoring the pin (#6267),
3. the digest-rollback runbook in release docs (#6113 row).

Cluster: pkg/hub heartbeat/channel re-arm + hub spoke-image API + src/docs/releases.md runbook. No open PR touches these files.

Refs #6267, #6268, #6113, #6016

---
*Filed by strategist agent (ACMM L5 — hold-gated mode)*
---
🐝 **Hive Agent**: `strategist` | **Instance:** `hosted-available-oke-11-placeholder-r05x` | **SHA:** `unknown`

— hive: agent=strategist backend=copilot model=claude-sonnet-4-6

clubanderson:
Rolling a hosted spoke back to a known-good build was not something the hub could do. The only path was a raw kubectl set image in the spoke's namespace, which left no record of who did it or why, and even that did not hold: the tracked-channel re-arm in the heartbeat handler honoured only the admin upgrade pause, so a channel-tracking hive an operator had hand-pinned was dragged back to the channel head on its next beat. A rollback that silently reverts is worse than none, because the operator believes the fleet is on the known-good digest when it is not.

This makes the digest pin a run-state on the hive record, following the provenance-carrying shape the per-repo pause established in #6211.

What changes

A pin is recorded on SaaSHive as digest_pin with who, when, the digest, the source short SHA when one was given, the reason, and the image the spoke was running when the pin was taken. It is persisted next to every other hub-owned fact about the hive, so it survives hub restarts and hub self-upgrades. tracked_channel is left exactly as it was, so lifting the pin resumes the selection the operator had.

POST /api/saas/hives/{id}/pin-digest (owner-only; hub admins pass the owner check) accepts either a sha256 manifest digest or the short git SHA the My Hives page shows. A SHA is resolved to its manifest-list digest on GHCR. The digest is shape-checked and existence-checked before anything is written, then applied through the same kubectl set image deployment/hive "*=..." -n hive-hosted-<id> path the switch handler uses. The record is written before the patch so no heartbeat can slip in between, and rolled back if the patch fails, so a 200 means the Deployment holds the digest. A cluster the hub cannot reach over kubectl is refused rather than recorded, because a digest cannot ride the heartbeat fallback, which carries a bare tag.

The tracked-channel re-arm skips a pinned hive. This is the #6267 half. Alongside it the heartbeat handler withholds any armed switch_to_tag or upgrade_to while the pin holds, triggerAutoUpgrades never arms a pinned hive, and the manual upgrade, switch-branch and bulk equivalents refuse with 409 naming the pin and its provenance. Nothing lifts a pin except an explicit unpin by an owner.

POST /api/saas/hives/{id}/unpin-digest clears the run-state, records the lift on the timeline, and writes the tracked channel back onto the Deployment, or the moving tag the hive was on when pinned for a plain-branch hive. On a cluster the hub cannot patch the tag is armed for the next heartbeat instead. Unpin is refused with 409 while the admin spoke-upgrade pause is on, matching switch-branch; pinning itself is allowed during the pause, since a rollback is the operator steering by hand while the train is stopped.

GET /api/saas/hives/{id}/digest-pin returns the pin, the spoke's last reported image, and landed: true once that reported image carries the pinned digest, which is the hub-side witness the runbook asks for.

The dashboard shows a PINNED pill beside the version pill, mirroring the PAUSED pill: the hover carries who, when, why and the full digest, an hourglass shows until the spoke reports the pinned digest, and owners click it to unpin. The branch picker is hidden while a pin holds, since the endpoint would refuse the click. Pin to digest and Unpin digest are in the row menu. The hive timeline gets digest_pinned and digest_unpinned events, and the drift report reads a hub-recorded pin at info with its provenance rather than critical.

Tests

digest_pin_test.go models the cluster with a scripted fake kubectl that records every invocation and persists the image its last set image wrote. TestHeartbeatReArmLeavesPinnedHiveAlone pins a stable-tracking hive, replays a heartbeat reporting the digest image, runs a triggerAutoUpgrades tick with auto-upgrade on and a newer build available, and asserts the Deployment image file still holds the digest, that kubectl ran zero further times, and that no switch_to_tag or upgrade_to went on the wire. It then lifts the pin by hand and replays the identical beat as a positive control, proving the re-arm fires and the guard is what held it. Other tests cover the set image on the deployment with provenance read back from disk, pin by SHA, non-owner refused with no kubectl run and nothing recorded, kubectl failure recording nothing, unpin resuming tracking, the 409 refusals while pinned, and unpin refused under the pause.

Docs

src/docs/release-rollback.md now leads with the endpoint for hosted spokes, keeps the direct kubectl patch only for clusters the hub cannot reach with a warning that the hub will undo it unless the fleet-wide pause is on, and documents the hub-side landed check and unpin. api-reference.md gains the three endpoints. A changelog.d fragment is included.

Fixes #6267
Fixes #6268
Fixes #6290


clubanderson:
Review (as a comment - GitHub blocks self-approval; labels carry the gate). This closes #6267, #6268 and #6290 as one change and the #6267 half is proven, not logged: TestHeartbeatReArmLeavesPinnedHiveAlone pins, posts a heartbeat, runs triggerAutoUpgrades with a newer SHA in the cache, and asserts the deployment image in the fake-kubectl state file is byte-unchanged and the invocation count did not move, then clears the pin and replays the identical beat as a positive control requiring switch_to_tag=stable. The gate itself adds !digestPinnedNow to BOTH re-arm paths (channel-tracked and spoke-managed chase-latest), which is what makes a rollback survive the next heartbeat. Three design calls I agree with: the pin record is persisted before kubectl runs and rolled back on failure, so a recorded-but-unapplied pin cannot exist; pull-only clusters are refused with 409 rather than silently pinned to a bare tag; pinning is allowed under the admin pause (rollback during an incident is the point) while unpin waits for the pause to lift, matching switch-branch. Endpoints are owner-only via userIsHiveOwner, non-owner asserts 403 with zero kubectl runs. The runbook now documents the endpoint and keeps kubectl only for unreachable clusters, with the warning that the hub undoes it unless the fleet pause is on.

PR #6358, opened 2026-09-09
🐛 fix(dashboard): give ContributeWSHub an explicit cleanup lifecycle (#6272)

### LINKED ISSUE #6272: 🐛 NewContributeWSHub starts cleanupLoop with no stop channel - leaks a goroutine per hub and can prune into a removed TempDir
## Summary

`NewContributeWSHub` (`src/pkg/dashboard/contribute_ws.go`) starts `cleanupLoop` with no stop channel or context. Every hub constructed in a test therefore leaks a goroutine for the lifetime of the test binary, with no way to stop it.

Surfaced while diagnosing the #6165 flake (which turned out to be a different, already-fixed disconnect-defer race). The leak is separate and still present.

## Why it matters

- **Test hygiene:** `go test` accumulates one orphan goroutine per constructed hub. Goroutine-leak detectors would flag it; today it is silent.
- **A latent TempDir race:** `cleanupLoop` runs on a 30s ticker and prunes leases older than `wsTaskTimeout` (30m). In a long test binary, a leftover prune from an earlier test's hub can fire after that test's `t.TempDir` has already been removed and recreate state under it - the same shape of failure as #6165, just on a much longer fuse, so it will look like an unrelated intermittent failure when it eventually hits.

## Suggested fix

Give the hub a lifecycle: accept a `context.Context` (or expose a `Close()` that closes a stop channel) and have `cleanupLoop` select on it alongside the ticker. Register `t.Cleanup(hub.Close)` in the test helpers that construct hubs. Production callers pass the server's shutdown context so the loop stops on graceful shutdown as well.

Refs #6165, #6158.

Danathar:
## Summary
Closes #6272.

`NewContributeWSHub` starts `cleanupLoop` in a background goroutine to sweep stale connections and reclaim/prune task leases every 30 seconds. Previously, `cleanupLoop` lacked any shutdown or cancellation mechanism, causing hubs constructed during test execution to leak background goroutines for the entire life of the test binary. In tests where state is backed by `t.TempDir()`, these leaked loops could later attempt to prune and persist leases into directories that had already been removed upon test completion.

This PR gives `ContributeWSHub` an explicit, deterministic lifecycle:
- Added `stopCh`, `doneCh`, and `stopOnce` to `ContributeWSHub`.
- Modified `cleanupLoop()` to select on `ticker.C` and `h.stopCh`, and `defer close(h.doneCh)`.
- Added `Close()` and `Stop()` to `ContributeWSHub` which prompt termination of `cleanupLoop`, wait for goroutine exit, and are safe and idempotent across concurrent and sequential calls.
- Preserved production 30-second cleanup cadence and lease cleanup semantics without altering `DrainForShutdown()`.
- Added `CloseContributeHub()` and `Close()` methods on `Server`, and wired `defer dashSrv.CloseContributeHub()` on server shutdown in `cmd/hive/main.go`.
- Added `t.Cleanup(hub.Close)` across all test helpers and direct test call sites in `pkg/dashboard`.
- Added regression tests in `src/pkg/dashboard/contribute_ws_lifecycle_test.go` covering prompt stop, idempotency, concurrent `Close()`, `DrainForShutdown()` separation, and absence of goroutine leaks.
- Added changelog fragment `changelog.d/fixed-6272-contribute-ws-hub-cleanup-lifecycle.md`.

## Verification
- Verified changelog fragment compilation with `bash src/scripts/test-compile-changelog.sh`.
- Ran `go test -race ./cmd/hive/...` (PASS).
- Ran `go test -race ./pkg/dashboard/...` (PASS).

---
Gemini 3.7 flash high reasoning
This PR was created by gemini 3.8 flash


clubanderson:
Lifecycle change looks correct: `stopOnce` guards the single `close(stopCh)`, `doneCh` is closed exactly once by `cleanupLoop`'s defer, `Close` waits on it so callers get a real "loop exited" guarantee, and the nil-receiver / nil-channel guards keep the struct-literal hubs used in existing tests safe (they never start `cleanupLoop`, and `<-nil` is skipped, so `t.Cleanup(hub.Close)` cannot deadlock). `registerContributeRoutes` closing a prior hub before replacing it is also right.

One blocking CI failure, caused by this PR (not a flake):

`v2 Tests` / `test (rest 3/3)` fails on `TestSleepRatchet` in `src/internal/testutil/sleep_ratchet_test.go`. That test counts `time.Sleep(` occurrences across `*_test.go` files and fails when the total exceeds `sleepBaseline = 198`. The new `src/pkg/dashboard/contribute_ws_lifecycle_test.go` adds two:

- `time.Sleep(50 * time.Millisecond)` at the top of `TestContributeWSHub_NoGoroutineLeak` ("let background tasks settle")
- `time.Sleep(10 * time.Millisecond)` inside its polling loop

Rerunning will fail the same way. Suggested fix: drop the 50ms settle sleep (it does not make the baseline meaningfully more stable) and replace the hand-rolled polling loop with the shared helper, e.g.

```go
testutil.Eventually(t, 2*time.Second, func() bool {
    return runtime.NumGoroutine() <= baseGoroutines+1
}, "goroutines leaked: baseline=%d", baseGoroutines)
```

(`src/internal/testutil/eventually.go` already samples every 10ms, so this is the same wait without adding to the ratchet count.)

Non-blocking nit: `TestContributeWSHub_NoGoroutineLeak` compares raw `runtime.NumGoroutine()` with a +1 slack. Because `NewContributeWSHub` also runs `loadCompletedTasks`/`loadFailedTasks` and the package has other background work, this can be sensitive to whatever else is live at that moment. It is fine to keep, but if it flakes later the more robust form is to assert on the hub's own `doneCh` (which the other lifecycle tests already do) rather than the process-wide goroutine count.

Once the ratchet passes I will add lgtm/approved so Tide can merge.

clubanderson:
Pushed the ratchet fix on top of the author's commits (138ffcae): dropped the 50ms settle sleep and replaced the polling loop in TestContributeWSHub_NoGoroutineLeak with testutil.Eventually, so TestSleepRatchet stays at its 198 baseline. No other changes.

PR #6391, opened 2026-09-09
🐛 fix(hub): arm and deliver upgrade targets a channel spoke can reach

clubanderson:
## Problem

Today 12 spokes sat in "Upgrading" with a ✓ on their SHA (`55bd2bc`, which is exactly what `:stable` carries). Hub log, every beat:

```
heartbeat: instructing hub-managed hive to upgrade (kubectl fallback) from=55bd2bc to=0056109
heartbeat: spoke reported a FAILED upgrade ... self-upgrade failed after 5 attempts (no error recorded; the image never changed - check that the deployment tracks a tag carrying the target SHA)
```

`0056109` is the v4 **tip**. A spoke whose Deployment tracks `ghcr.io/hivecommons/hive:stable` can only land on the channel's commit, so it re-pulled, came back on `55bd2bc`, and was told again. Restart storms followed (`3 restart storms` on epm-luiz).

#6294 fixed this for `triggerAutoUpgrades`, but three arming sites still read `getLatestSHAForBranch(branch)`:

- `handleSaaSUpgrade` (manual **Upgrade now**) — `saas.go`
- `bulkRestartOrUpgrade` — `saas_bulk.go`
- stale-upgrade recovery (`advancing upgrade target for stale upgrade`) — `saas.go`

## Fix

1. All three resolve through `reachableUpgradeTarget(branch, imageRef, trackedChannel)`. Manual/bulk do it **outside `s.mu`** (the channel lookup may consult GHCR, 8s timeout) and refuse with 409 / a per-hive error when the channel does not resolve, instead of inventing a target.
2. Delivery-time guard in the heartbeat: whatever armed `heartbeatUpgrade[id]`, if the spoke tracks a release channel the armed target is checked against the channel's commit — at/ahead → drain the fallback **and** clear the `Upgrading` latch; behind → re-aim at the channel's commit; unresolved → withhold. This covers any arming path I missed and un-wedges spokes armed before the fix as soon as they beat.
3. Header: a channel row resolved via its revision label (#6385) now carries the commit's first line (`message`), cached in `commitMsgBySHA`, so `stable -> 55bd2bc` reads like the branch rows.

## Tests

`channel_targeting_test.go`: drain-at-channel (fallback + latch cleared), re-aim-when-behind (map + registry target), branch-tracking spoke keeps the armed tip, unresolved channel withholds. `release_channels_test.go`: message populated from the revision-label path and served from cache on the second resolve. `go test ./pkg/hub/` green; `golangci-lint` 0 issues.

Refs #6294, #6378, #6385


clubanderson:
Added a second commit for the related switch bug found on kalantar-msb: switching a `:stable` spoke to `v4` was declared complete on the first heartbeat because the spoke reports `git_branch: v4` (the :stable image is built from v4), so the Deployment never left `:stable` and the v4 tip was unreachable forever. `git_branch` is now completion evidence only when no image ref is reported. Tests: `TestHeartbeatBranchSwitch*`.

PR #6395, opened 2026-09-09
✨ feature: backport off-repo ACMM criterion waivers to v4

### LINKED ISSUE #6263: ✨ feature: ACMM criteria cannot tell a missing capability from one that moved off-repo

Danathar:
## Summary

Backport the ACMM off-repository criterion waiver mechanism from merged PR #6264 to the active `v4` release line. PR #6264 landed on `v5`, while issue #6263 explicitly targets `v4`; current `upstream/v4` therefore still cannot distinguish a capability that is absent from one deliberately provided elsewhere.

Repositories may now commit `.acmm.yml` (or `.acmm.yaml`) with a criterion ID, the system that satisfies it, and a reason. The dashboard counts that declaration toward the displayed ratio and visibly marks it as waived, while level advancement remains based only on criteria detected from repository files.

Fixes #6263

## Problem and user impact

The ACMM evaluator currently treats file presence as capability presence. That is intentionally cheap and language-agnostic, but it makes these materially different states indistinguishable:

- a repository never implemented a capability;
- a repository moved the capability to a shared workflow, platform service, service mesh, or hive itself.

The second repository is shown as regressing even though the capability remains available. This creates an incentive to restore a file that was removed for sound operational or security reasons, or to leave permanent red criteria that train operators to ignore the panel.

The concrete case in #6263 is `Danathar/sensi`: removing `.github/workflows/ai-fix.yml` eliminated a duplicate privileged agent path because hive already provided that work. Since two L4 criteria recognize that filename, the security improvement changed the displayed result from 9/9 to 7/9.

## Root cause

`checkCriterion` can only answer whether any configured path exists in the repository. There was no committed, auditable way for a repository to state that a known criterion remains satisfied outside those paths, and scoring had only one matched count. Simply treating an exemption as an ordinary pass would be unsafe because enough exemptions could then advance a repository to a maturity level it had not directly demonstrated.

## Implementation

- Add `src/pkg/dashboard/acmm_waivers.go` to parse and validate root-level `.acmm.yml` / `.acmm.yaml` declarations. Criterion IDs must exist in `universalCriteria`; `satisfied_by` and `reason` are required; malformed, oversized, incomplete, unknown, and duplicate declarations fail closed by leaving the criterion red. Input is capped at 64 KiB.
- Reuse the root directory listing already populated by `prefetchDirectories`. Repositories without a waiver file incur no additional GitHub API call; only a repository that advertises one pays for a content fetch.
- Extend criterion results with additive waiver metadata and split level scoring into displayed `Matched`/`Score` and direct `Detected`/`DetectedScore` values. A waiver is consulted only after file detection fails and never masks a later real detection.
- Compute `ACMMLevelScore.Passed` and the codebase level from `DetectedScore` only. Waivers can close the gap between an already-passing level and full green, but cannot carry any level across its threshold.
- Preserve aggregate semantics: if any repository directly detects a criterion, the fleet aggregate is a direct detection; it is marked waived only when all successful evidence is waiver-based.
- Render a waived chip and its escaped justification on criterion rows, plus a red asterisk and explanatory tooltip on every level summary renderer. Waived criteria remain passing and are not offered as gap issues.
- Document the declaration format, scoring rules, visibility, precedence, failure behavior, and API cost in `src/docs/acmm-waivers.md`, link it from the docs index, and add the required changelog fragment.

## Design invariants and boundaries

- A waiver cannot advance an ACMM level or the computed codebase level.
- A waiver cannot override a detected file.
- Unknown criterion IDs cannot silently appear effective; a renamed criterion reopens as a visible gap.
- Both the provider and justification are mandatory, keeping this an assertion about where a capability lives rather than an unchecked skip list.
- Repository-supplied text is HTML-escaped before dashboard rendering.
- Existing repositories without `.acmm.yml` retain their prior scores, levels, API behavior, and request cost. New JSON fields are additive.
- This is a source backport of commit `115a7b5629d6942af9e91d827a0dae0da03c1c3d` from #6264; no unrelated `v5` changes are included.

## Regression coverage

The backport carries 24 focused tests covering:

- valid YAML and both accepted filenames;
- malformed, oversized, incomplete, duplicate, and unknown-ID declarations;
- zero-call behavior when the cached root listing has no waiver file;
- fallback fetching when the root prefetch is unavailable;
- full-green restoration without level advancement, including an entire-level waiver spree;
- detection taking precedence over a declaration;
- per-repository and aggregate waiver semantics;
- escaped dashboard justification, row/level markers, and suppression of gap-issue actions for waived criteria.

## Verification

- `cd src && go test ./pkg/dashboard/` — PASS (`68.948s`).
- `cd src && go build ./...` — PASS.
- `cd src && go vet ./pkg/dashboard/` — PASS.
- `cd .github/scripts && npm ci --ignore-scripts --no-audit --no-fund` — PASS.
- `find dashboard src/pkg/dashboard/static -name '*.html' -print0 | xargs -0 --no-run-if-empty node .github/scripts/check-inline-js.js` — PASS for syntax and ESLint `no-undef` across both dashboard HTML trees.
- `cd src && go test ./...` — all changed dashboard tests pass, but the full command reports unrelated environment failures in `pkg/agent` and `pkg/config`:
  - tmux-backed `pkg/agent` tests cannot create their Unix socket because the linked-worktree path is too long; dependent restart/race tests then fail or time out;
  - `pkg/config/TestGateFailsClosedWithoutIP6Tables` receives an `ip6tables` permission error instead of the simulated missing-binary condition because this environment lacks the required network-administration privilege.
- Both failures reproduce on a separate, untouched worktree at `upstream/v4` commit `5f41bbd2`: `go test ./pkg/agent` fails with the same `File name too long` tmux socket error, and `go test ./pkg/config` fails the same `ip6tables` test with `Permission denied (you must be root)`.

— hive: backend=codex


clubanderson:
## Review: backport fidelity + CI

**CI failure investigated:** `NOTICE matches the module graph` was failing at the "Regenerate NOTICE from the current module graph" step — but the failure was `go install github.com/google/go-licenses@v1.6.0` unable to reach `sum.golang.org` (`stream error ... INTERNAL_ERROR`) after 3 retries, an infra/network flake, not a real drift.

Confirmed this PR makes **zero changes** to `src/go.mod`, `src/go.sum`, or `NOTICE` relative to `origin/v4` — the module graph is unchanged, so there was nothing to regenerate. The backport correctly reuses the `gopkg.in/yaml.v3` dependency already present in v4's `go.mod` (the same one v5's original commit used) rather than adding a new YAML library. I re-ran the failed job (`gh run rerun 34375327808 --failed`) and it passed on retry, confirming the flake theory. All checks are now green.

**Fidelity vs v5 commit `115a7b5629d6942af9e91d827a0dae0da03c1c3d`:**
Byte-for-byte diffed every touched file. All 5 new files (`acmm_waivers.go`, `acmm_waivers_test.go`, `acmm_waiver_ui_test.go`, `src/docs/acmm-waivers.md`, `changelog.d/added-acmm-waivers.md`) are **identical** to the v5 originals. The 3 modified files (`api_acmm_eval.go`, `static/index.html`, `src/docs/README.md`) have identical diff hunks, differing only in line-number offsets caused by v4's pre-existing baseline — no unrelated v5 content leaked in, nothing dropped. **This is a faithful source backport.**

Invariants from the PR body hold in code: `Passed`/codebase level derive from `DetectedScore` only (waivers can't advance a level), waiver lookup happens only after file-detection fails (can't override a detected file), unknown criterion IDs are rejected against `universalCriteria` (fail closed), input is capped at 64 KiB, and repo-supplied text is HTML-escaped before rendering (`acmmWaiverDetail`/`acmmLevelWaiverMark`).

**Verification run (worktree at PR head `06d4af6c`):**
- `cd src && go build ./...` — PASS
- `go vet ./pkg/dashboard/` — PASS
- `go test ./pkg/dashboard/ -count=1` — PASS (47.4s)
- `go test -cover ./pkg/dashboard/` — 87.1% (floor in `.github/workflows/v2-tests.yml` is 84%) — PASS
- `golangci-lint run ./pkg/dashboard/...` — 0 issues
- `python3 src/scripts/check-docs-links.py src/docs` — all relative links/anchors resolve (145 files); `acmm-waivers.md` is indexed in `src/docs/README.md`
- `changelog.d/added-acmm-waivers.md` exists, first line starts with `- `
- Merge check: `git merge --no-commit --no-ff origin/v4` on the PR branch completes with no conflicts — no rebase required

No code changes were needed on this branch; only the CI job was re-run.

Issue #5039, opened 2026-08-28
contribute: define provider transport and readiness for the Pi backend


joshyorko:
## Outcome

Define provider/model transport and machine-readable readiness for the Pi contributor backend so downstream launchers can select Pi providers without hardcoding Anthropic or inventing a local Hive command path.

Downstream demand: projectbluefin/review#356. This is related to, but narrower than, #2547: it defines the Pi adapter contract rather than task-side capability routing.

## Current evidence

- Current `v4` head is `2de9eec551a3112700645f5987bd84d67d3e0de5`.
- `config/backends.conf` maps `pi` to the `pi` binary with no permission flag.
- The interactive contributor path transports `AGENT_MODEL` generically as `--model <value>`, but no provider value exists and Pi remains outside the headless matrix.
- Pi `0.73.1` officially supports:

  ```text
  --provider <name>
  --model <provider/model or model pattern>
  --api-key <key>
  --thinking <level>
  --print
  --mode text|json|rpc
  ```

- Pi has no official `PI_PROVIDER` or `PI_MODEL` environment variables. Its credential variables are provider-specific, and its auth file is `~/.pi/agent/auth.json`.
- Provider alone does not select a model; a canonical provider-qualified model (`provider/model`) is sufficient for many cases.
- The Review consumer currently requires `PI_API_KEY`, exports it as `ANTHROPIC_API_KEY`, and transports no Pi provider/model. That is an Anthropic-specific downstream adapter, not a general Pi/Hive contract.
- Initial and restart paths can disagree: the interactive launch reads `AGENT_MODEL`, while relay restart fallback may derive a model from `AGENT_MODEL || GOOSE_MODEL`.
- Pi interactive cancellation is not conformance-proven; the headless child-kill path does not apply because Pi is not headless-supported.

## Required contract decision

Define, with tests, which actor owns each field:

1. `AGENT_BACKEND=pi` remains contributor-selected; Hive remains sole task-assignment authority.
2. Decide whether provider/model are contributor preferences or assignment-owned requirements. A downstream launcher must not override assignment-owned values if Hive later supplies them.
3. Prefer one canonical `provider/model` transport where Pi supports it; add a separate provider field only when the adapter requires it and authority is explicit.
4. Keep provider credentials in their official provider-specific environment variable or one narrow staged auth file. No generic secret value may appear in argv or logs.
[...]

PR #6406, opened 2026-09-09
🐛 fix(goose): forward OPENAI_HOST and OPENAI_BASE_PATH into the contributor container

### LINKED ISSUE #6400: 🐛 bug: Goose backend does not pass OpenAI endpoint variables into contributor container
## What happened?

When using the Goose backend with an OpenAI-compatible local inference server, Hive passes `GOOSE_PROVIDER`, `GOOSE_MODEL`, and `OPENAI_API_KEY` into the contributor container, but does not pass `OPENAI_HOST` or `OPENAI_BASE_PATH`.

As a result, Goose falls back to:

```text
https://api.openai.com/v1/chat/completions
```

and fails with `401 Unauthorized`.

## Expected behavior

Hive should pass the OpenAI endpoint variables into the contributor container when they are configured on the host, allowing Goose to use an OpenAI-compatible local inference server.

## Reproduction

Host configuration:

```bash
export GOOSE_PROVIDER=openai
export GOOSE_MODEL=bartowski/Qwen_Qwen3-14B-GGUF
export OPENAI_HOST=http://172.17.0.1:8080
export OPENAI_BASE_PATH=v1/chat/completions
export OPENAI_API_KEY=dummy
```

Start a Hive Goose contributor and inspect its environment:

```bash
docker exec <container> env | grep -E '^(GOOSE_|OPENAI_)'
```

The container contains:

```text
OPENAI_API_KEY=dummy
GOOSE_PROVIDER=openai
GOOSE_MODEL=bartowski/Qwen_Qwen3-14B-GGUF
```

but `OPENAI_HOST` and `OPENAI_BASE_PATH` are missing.

A Hive-assigned task then fails with:

```text
Authentication failed for https://api.openai.com/v1/chat/completions
Status: 401 Unauthorized
```

### Workaround

Injecting the missing variables into the same running container makes Goose work:

```bash
docker exec -it \
  -e OPENAI_HOST=http://172.17.0.1:8080 \
  -e OPENAI_BASE_PATH=v1/chat/completions \
  <container> goose
```
Goose then successfully connects to the local inference server and responds to prompts.

Image

## Environment

- Hive: `v4` branch
- Backend: Goose
- Goose: `1.49.0`
- Container runtime: Docker/Podman
- Host: Bluefin DX (`bluefin-dx:stable`)
- Model: Qwen3 14B GGUF
- Inference server: RamaLama/llama.cpp

## Additional context

The contributor launch code has an explicit provider environment allowlist. `OPENAI_API_KEY`, `GOOSE_PROVIDER`, and `GOOSE_MODEL` are included, but `OPENAI_HOST` and `OPENAI_BASE_PATH` are not.

Adding those two variables to the allowlist appears to resolve the issue.

clubanderson:
## What changed

The Justfile's `contribute-hive` recipe forwards only an explicit allowlist of
provider environment variables from the host into the contributor container
(least-privilege, see #5039). `GOOSE_PROVIDER`, `GOOSE_MODEL` and
`OPENAI_API_KEY` were on that list, but `OPENAI_HOST` and `OPENAI_BASE_PATH`
were not.

- Added `OPENAI_HOST` and `OPENAI_BASE_PATH` to the `for name in ...` allowlist
  loop in `Justfile`, right next to the existing `OPENAI_API_KEY` /
  `GOOSE_PROVIDER` / `GOOSE_MODEL` entries. They go through the same
  `add_provider_env` conditional forwarder as every other entry, so each is
  only forwarded (`-e NAME`, name-only — the value never appears in the
  runtime command's argv) when set on the host, and is simply absent
  otherwise.
- Added a static contract check,
  `src/scripts/check-provider-env-allowlist.sh`, following the existing
  `check-podman-contract.sh` pattern, plus a new CI workflow
  (`.github/workflows/provider-env-allowlist-contract.yml`, registered in
  `.github/release-lines.yml`) that runs it on Justfile changes.
- Updated docs: `src/docs/env-vars.md` (new `OPENAI_HOST` / `OPENAI_BASE_PATH`
  rows next to `GOOSE_MODEL`), `src/docs/contributor-relay.md` and
  `docs/backend-setup.md` (goose backend rows now mention the
  OpenAI-compatible endpoint override).
- Added a changelog fragment,
  `changelog.d/fixed-6400-goose-openai-endpoint-env.md`.

## Why

With an OpenAI-compatible local inference server (e.g. RamaLama/llama.cpp),
Goose needs `OPENAI_HOST` and `OPENAI_BASE_PATH` to route requests there
instead of `https://api.openai.com`. Because those two variables were missing
from the allowlist, they never reached the contributor container, so Goose
silently fell back to the real OpenAI endpoint and failed with `401
Unauthorized` even though the host shell had them exported correctly.

## Reproduction (from the issue)

```bash
export GOOSE_PROVIDER=openai
export GOOSE_MODEL=bartowski/Qwen_Qwen3-14B-GGUF
export OPENAI_HOST=http://172.17.0.1:8080
export OPENAI_BASE_PATH=v1/chat/completions
export OPENAI_API_KEY=dummy
# start a Hive Goose contributor, then:
docker exec <container> env | grep -E '^(GOOSE_|OPENAI_)'
```

Before this fix, `OPENAI_HOST`/`OPENAI_BASE_PATH` were missing from the
container environment and the assigned task failed with
`Authentication failed for https://api.openai.com/v1/chat/completions —
Status: 401 Unauthorized`. After this fix, both variables are forwarded into
the container whenever they are set on the host.

## How tested

- `bash src/scripts/check-provider-env-allowlist.sh Justfile` — passes on this
  branch; confirmed it **fails** against `origin/v4`'s Justfile (the parent
  commit), proving it reproduces the bug before the fix:
  ```
  FAIL: OPENAI_HOST is forwarded to the contributor container (...)
  FAIL: OPENAI_BASE_PATH is forwarded to the contributor container (...)
  Provider env allowlist contract check FAILED — see hivecommons/hive#6400
  ```
- `bash src/scripts/check-podman-contract.sh Justfile` — still passes
  (sibling static contract check unaffected).
- `bash src/scripts/check-release-lines.sh` — `RESULT: PASS`, confirming the
  new workflow's branch pins are registered correctly in
  `.github/release-lines.yml`.
- `python3 src/scripts/check-docs-links.py src/docs` — `all relative links and
  anchors resolve`.
- `just --list` — Justfile still parses cleanly.
- `shellcheck src/scripts/check-provider-env-allowlist.sh` — no findings.
- No Go source files were touched by this change, so `go build`/`go
  vet`/`go test` were not applicable.

Fixes #6400


clubanderson:
Duplicate of #6404 (same Justfile allowlist fix, filed first and now merged). Closing; the extra docs rows here are covered by #6404's env-vars.md update.

PR #6425, opened 2026-09-09
🐛 fix(dashboard): build hosted-tenant URLs from the server's spoke domain

clubanderson:
## The bug

The dashboard builds `<id>.<domain>` URLs for hosted hives that carry no explicit `dashboardUrl`. It built them from a **hardcoded** `hive.kubestellar.io` in three places, so after the fleet moved to `hive.hivecommons.dev` every one of those links still named the retired domain.

This is not cosmetic staleness. The retired name is **outside the wildcard certificate the fleet now serves**, so the links don't degrade into a redirect — the handshake is refused:

```
$ curl -sSI https://hosted-aslom.hive.kubestellar.io/dashboard
curl: (60) SSL: no alternative certificate subject name matches
      target host name 'hosted-aslom.hive.kubestellar.io'

$ curl -sSI https://hosted-aslom.hive.hivecommons.dev/dashboard
HTTP/2 200
```

A tenant clicking through to their own hive gets a browser certificate interstitial.

## The fix

The hub already knows the right answer — `hubSpokeDomain()`, from `HIVE_HUB_SPOKE_DOMAIN`, which **is** set correctly in production — it just never sent it to the client. This sends it as `hub_spoke_domain` and builds the three URL sites from it.

When the domain isn't known yet (before the first payload lands) the link is **suppressed** rather than falling back to a literal. A wrong host here fails closed and loudly, so guessing is worse than omitting.

## What this deliberately does *not* change

`defaultHubPublicURL`, `defaultHubCanonicalHost` and `defaultHubSpokeDomain` still name the old domain. I repointed them, found out why that's not a one-liner, and backed it out:

- they're a **coupled set** — canonical host seeds `sessionCookieParentDomain()`, spoke domain seeds redirect trust. Moving `defaultHubSpokeDomain` alone desynchronizes them and fails `TestSubdomainTrustIsRedirectOnly`, `TestIsTrustedRedirectTarget`, `TestHostedLoginRedirectToSiblingStillWorks`; moving the canonical host fails the `whoami_sso_test.go` cookie-scoping tests, which encode the old default on purpose;
- production overrides all three via env, so they're **latent**, not live.

Moving them as a set (and updating those tests) is a real decision about cookie scoping for fresh deploys — worth doing, but not smuggled into a link fix.

`defaultLegacyHubCookieDomain` and `legacyImpersonateCookieDomain` must **keep** naming the old domain: expiring the previous build's cookies is their only job, and repointing them makes the expiry a no-op. Now pinned by a test so nobody "fixes" them.

## Verification

- `go build ./...`, `go vet ./pkg/hub/` clean
- full `go test ./pkg/hub/ -count=1` — **ok**, 138s
- 3 new tests. The JS is a Go string constant, so a reintroduced literal compiles and passes every other test — it can only be caught by asserting on the string, which is what these do.

Refs #5925

### DIFF
--- a/changelog.d/fixed-hosted-tenant-links-spoke-domain.md
+++ b/changelog.d/fixed-hosted-tenant-links-spoke-domain.md
@@ -9,3 +9,10 @@
   takes the parent domain from the server (`hub_spoke_domain`, derived from
   `HIVE_HUB_SPOKE_DOMAIN`) and suppresses the link entirely when the domain is
   not yet known, rather than guessing a host (#5925).
+
+- Fixed the same hardcoded domain in the public landing page's "Contribute Now"
+  and "Open Contribute Page" links. That page is static and cannot be handed the
+  configured domain, so it now derives it from the host it is served on — which
+  is the hub's own host, and therefore the domain hosted spokes hang off — and
+  omits the link when it cannot be derived rather than emitting a guessed one
+  (#5925).
--- a/changelog.d/fixed-hosted-tenant-links-spoke-domain.md
+++ b/changelog.d/fixed-hosted-tenant-links-spoke-domain.md
@@ -1,0 +1,11 @@
+- Fixed hosted-tenant links in the hub dashboard pointing at the retired
+  `hive.kubestellar.io` domain. The dashboard builds `<id>.<domain>` URLs for
+  hosted hives that carry no explicit dashboard URL, and it built them from a
+  hardcoded hostname, so they kept naming the pre-move domain after the fleet
+  moved to `hive.hivecommons.dev`. That is not a cosmetic staleness: the
+  retired name is outside the wildcard certificate the fleet now serves, so
+  those links did not redirect — the browser refused the TLS handshake and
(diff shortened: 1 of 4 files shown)


clubanderson:
Pushed a second commit: the **public landing page had the same bug**.

`contributeBase()` in `static/index.html` had an identical hardcoded `hive.kubestellar.io` fallback, so the "Contribute Now" and "Open Contribute Page" buttons hit the same TLS refusal. Its own comment even records that these links were *previously* fixed for being "a dead host" — the literal just got re-pointed at what has since become another dead host.

That page is static (served raw via `http.FileServerFS`, no templating), so it can't be handed `hub_spoke_domain` the way the dashboard is. It derives the domain from `location.hostname` instead — the page is served *by* the hub, so the hub's own host **is** the domain hosted spokes hang off, and unlike a literal that follows the hub wherever it moves.

Both call sites now suppress the link when the base can't be derived, rather than falling through to a relative `/contribute` on the hub itself.

Full `go test ./pkg/hub/` still green; landing-page JS parse-checked; a fourth test reads the embedded page and pins all of the above.

Issue #5655, opened 2026-09-02
🐛 bug: Ctrl-C on a busy relay leaves the task's scoped GitHub token on disk — the shutdown path skips the task-exit contract


Danathar:
## Summary

Ctrl-C is how a contributor stops their relay — it is what `just contribute-hive`
leaves you at, and what the docs assume. When the relay is holding a task at that
moment, the shutdown path does none of the things every *other* task-exit path in
the relay does: it does not drop the task's GitHub credential, does not tell the hub
the work is being abandoned, and does not stop the agent.

The credential half is directly observed. After a Ctrl-C while task
`ct-kubestellar/hive-5652-…` was in flight, the per-task scoped token was still sitting
at `/tmp/hive-gh-token.cache` (0600, 390 bytes), and it stays valid for the remainder
of its ~55-minute lifetime — well past the point where the hub has released the issue
and can offer it to a different contributor.

```
$ ls -l --time-style=+%H:%M /tmp/hive-gh-token.cache
-rw------- 1 … 390 08:47 /tmp/hive-gh-token.cache      # relay SIGINT'd at 08:50
```

Blast radius: any interactive contributor who stops a busy relay with Ctrl-C, which
is the normal way to stop one.

## Cause

`bin/contributor-relay.sh`, at the bottom of the file:

```js
function cleanup() {
  hubs.forEach(hub => {
    if (hub.heartbeatInterval) { clearInterval(hub.heartbeatInterval); hub.heartbeatInterval = null; }
  });
  if (progressInterval) { clearInterval(progressInterval); progressInterval = null; }
}

process.on('SIGTERM', () => { cleanup(); process.exit(0); });
process.on('SIGINT',  () => { cleanup(); process.exit(0); });
```

Timers only. The relay already has the right helper and uses it everywhere else —
`stopAgentForTaskExit()`, whose documented step 1 is "unlink the credential FIRST, so
a turn that survives the interrupt (or races it) cannot keep using it" (#5353). The
signal path predates it and was never wired in.

The hub side degrades gracefully on its own: the socket dies, the disconnect handler
books a release cooldown and logs `released: connection lost` (#5097). That is correct
and is not what this issue is about. The gap is entirely on the contributor's host,
after the process is gone.

## Why the leftover token matters

It is the #2356 / #5353 shape, reached from a direction neither of those covers. Once
the hub releases the issue it may be handed to another contributor, while on the
original host a credential scoped to that same repo is still on disk and still valid.
[...]

PR #6498, opened 2026-09-10
fix(relay): accept Markdown-wrapped HIVE_VERDICT sentinels

### LINKED ISSUE #6492: relay: HIVE_VERDICT missed when the model wraps the sentinel in Markdown bold (goose/ramalama) — completion degrades to chrome_idle
## Report

From a contributor running a goose + ramalama backend:

> I'm still very new to hive, so I may be misunderstanding how this is supposed
> to work. I'm using goose with ramalama. Hive logs show "pane looks idle but no
> HIVE_VERDICT", but the goose logs show
> `HIVE_VERDICT: complete — Configuration validated; next steps outlined.`

They're not misunderstanding anything — this is a real relay bug, and the pane
shows the cause: the model wrapped the sentinel in Markdown bold, so the line on
the pane is literally

```text
**HIVE_VERDICT: complete — Configuration validated; next steps outlined.**
```

## Diagnosis (verified at 378936c, current default-branch head)

`detectHiveVerdict()` in `bin/contributor-relay.js` is anchored at line start
and allows only whitespace or a codex/claude bullet glyph before the marker:

```js
const VERDICT_RE = new RegExp(`^\\s*(?:[•●]\\s*)?HIVE_VERDICT:\\s*(${alt})\\b[\\s:—–-]*(.*)$`, 'i');
```

The leading `**` makes it miss. Both forms run through that exact regex:

```text
**HIVE_VERDICT: complete — Configuration validated; next steps outlined.**   -> no match
HIVE_VERDICT: complete — Configuration validated; next steps outlined.       -> match
### **HIVE_VERDICT**                                                         -> no match (no colon)
```

This is the same failure class as the earlier `●` fix recorded in the scanner's
own comment: the agent printed the verdict correctly and the scanner didn't
recognize the presentation chrome around it. The `### **HIVE_VERDICT**` heading
the model printed above the line doesn't help either, since it has no colon.

## Impact

Bounded but misleading. After the idle grace period the relay completes the
task anyway with `completion_signal: chrome_idle`, so nothing gets stranded
(hive#6450 ended exactly that way in the contributor's log). But
`CompletionSignal` exists precisely so per-backend sentinel non-compliance is
*measurable* (#5376) — and with this bug, every goose/ramalama task will land
in the `chrome_idle` bucket, making the backend look non-compliant in the
metrics even when it is trying to comply.

The "Pane is at a shell prompt, not goose" warning in the same log was a
separate transient liveness reading that cleared on the next tick — not worth
chasing here.

## Proposed fix (both halves)

**Relay side (the one that matters):** tolerate leading Markdown emphasis and
strip trailing emphasis from the captured reason:

```js
^\s*(?:[•●]\s*)?(?:[*_]{1,3}\s*)?HIVE_VERDICT:\s*(complete|no_work_needed)\b[\s:—–-]*(.*)$
```

plus a trailing-`**`/`__` strip on the reason (verified: the tolerant regex
captures `Configuration validated; next steps outlined.**` — the trailing
emphasis lands in the reason group).

This does not weaken the anchoring's real purpose, which is rejecting the
prompt echo: the prompt quotes the marker mid-sentence with a leading `'`, and
a wrapped echo still hits the `<short reason>` guard. Verified: the tolerant
regex still rejects
`the exact form 'HIVE_VERDICT: complete — <short reason>' when done`.

**Prompt side:** add "plain text, no Markdown formatting" to the sentinel
instruction in `buildTaskPrompt` (`src/pkg/dashboard/contribute_ws.go`), which
today prescribes the exact form but says nothing about formatting. This helps,
but small local models ignore formatting instructions often enough that the
relay fix is the load-bearing one.

**Test:** a case with a bolded verdict line in the scanner's tests would pin
it, alongside the existing echo-guard cases.

## Caveat worth recording for the contributor

Even if the line had parsed, this was a weak "complete": the underlying issue
is a real regression (Battery Health controls broken on an AMD Framework), and
the model produced no PR, speculated the config might not be in the image, and
told the user to run `lsmod`. Hive would have recorded that as a genuine
completion. That matches prior experience testing local backends — the small
models follow the protocol loosely and the task more loosely still. Fixing the
parser makes compliance measurable again; it does not make the output more
useful. That larger question is out of scope here.

## Related

- #5376 — the completion sentinel and `completion_signal` diagnostics
- the `●` (U+25CF) claude-glyph miss fixed via `bin/test_backend_smoke.sh`,
  recorded in the scanner's comments — same failure class
- #6450 — example task that completed via `chrome_idle` this way

Danathar:
## Problem

Goose/RamaLama can render an otherwise valid completion sentinel as Markdown bold:

```text
**HIVE_VERDICT: complete — Configuration validated; next steps outlined.**
```

The interactive relay did not recognize that line, so it waited for terminal chrome to look idle and ultimately reported `completion_signal: chrome_idle`. Tasks still completed, but the fallback incorrectly made a compliant backend look non-compliant in completion-signal metrics.

## Root cause

`detectHiveVerdict()` anchored sentinel matches at the start of a pane line and tolerated only whitespace plus the known Codex/Claude bullet glyphs. The leading `**` therefore prevented a match. Simply broadening the match was not sufficient because the trailing Markdown delimiter would then leak into the captured human-readable reason, and the parser's prompt-echo protections must remain intact.

## Implementation

- Allow one to three leading `*` or `_` emphasis characters after optional CLI bullet chrome in `bin/contributor-relay.js`.
- Remove a trailing delimiter only when it matches an observed opener. This strips Markdown presentation from the reported reason without truncating an ordinary unwrapped reason that legitimately ends in `*` or `_`.
- Preserve the raw pane line used by the existing delivered-verdict baseline, so stale verdict attribution remains byte-for-byte and task-scoped.
- Tell contributors to print both `complete` and `no_work_needed` sentinels as plain text with no Markdown formatting in `buildTaskPromptBody`. The tolerant relay parser remains the load-bearing compatibility layer for models that ignore that instruction.
- Add the required user-visible fix fragment under `changelog.d/`.

The parser remains line-start anchored, accepts only the two requested verdict tokens, and retains the `<short reason>` prompt-placeholder guard. It does not treat Markdown headings such as `### **HIVE_VERDICT**` as completions.

## Regression coverage

- A Goose pane containing the exact bold rendering from the issue now completes on the first tick with `completion_signal: verdict`, rather than reaching `chrome_idle`.
- Parser coverage verifies bold `complete`, bullet-prefixed underscore-bold `no_work_needed`, clean reason extraction, and preservation of an unwrapped reason's trailing `*`.
- The anti-false-positive suite verifies that a bold-wrapped `<short reason>` prompt echo is still rejected.
- Dashboard prompt coverage requires the plain-text/no-Markdown instruction for both sentinel variants.

## Verification

- `env -u AGENT_BACKEND -u AGENT_LAUNCH_CMD -u HIVE_AGENT_CACHE_DIR -u HIVE_AGENT_CWD -u HIVE_HUB -u HIVE_REGISTRATION_TOKEN -u HIVE_WORKSPACE_DIR RELAY_TEST_ONLY="#6492" node bin/contributor-relay.test.js` — pass (both #6492 tests).
- `env -u AGENT_BACKEND -u AGENT_LAUNCH_CMD -u HIVE_AGENT_CACHE_DIR -u HIVE_AGENT_CWD -u HIVE_HUB -u HIVE_REGISTRATION_TOKEN -u HIVE_WORKSPACE_DIR RELAY_TEST_ONLY="completion sentinel inherits" node bin/contributor-relay.test.js` — pass (prompt-echo guard).
- `node --check bin/contributor-relay.js && node --check bin/contributor-relay.test.js` — pass.
- `env CC=/usr/bin/gcc CXX=/usr/bin/g++ CCACHE_DIR=/tmp/hive-6492-ccache GOCACHE=/tmp/hive-6492-go-build GOTMPDIR=/tmp/hive-6492-go-tmp TMPDIR=/tmp/hive-6492-tmp go test ./pkg/dashboard` — pass (`56.887s`; run outside the socket-restricted sandbox because package tests use localhost `httptest` servers).
- Full isolated relay run: 327/330 passed. The three unrelated failures are the existing #5655 SIGINT/SIGTERM/crash subprocess tests under local Node `v26.8.1`; the child exits before its buffered `TOKEN_ON_DISK` stdout is observed. Every other relay test passed.

## Scope

This change improves sentinel transport and measurement only. It does not alter the chrome-idle safety fallback, backend output quality, task-completion semantics, or the behavior of lines without a valid colon-bearing sentinel.

Fixes #6492

— hive: backend=codex

### DIFF
--- a/bin/contributor-relay.js
+++ b/bin/contributor-relay.js
@@ -1891,7 +1891,10 @@ function detectHiveVerdict(lines, wanted) {
   // that instruction echo from reading as the agent's own verdict. Codex
   // renders its completed assistant messages with a leading bullet (•,
   // U+2022) and Claude Code with a filled circle (●, U+25CF) — presentation
-  // chrome rather than part of the verdict. The claude glyph was missing
+  // chrome rather than part of the verdict. Some backends also wrap the whole
+  // line in Markdown emphasis (for example **HIVE_VERDICT: complete — done**),
+  // which is likewise presentation rather than sentinel content. The claude
+  // glyph was missing
   // until bin/test_backend_smoke.sh drove a REAL claude pane through the
   // relay: the agent printed the sentinel, this regex missed it, and every
   // interactive claude completion silently degraded to the chrome_idle
@@ -1902,13 +1905,21 @@ function detectHiveVerdict(lines, wanted) {
   // non-matches — a prose line that merely STARTS with a verdict word must not
   // become a verdict.
   const alt = wanted.map(w => w.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')).join('|');
-  const VERDICT_RE = new RegExp(`^\\s*(?:[•●]\\s*)?HIVE_VERDICT:\\s*(${alt})\\b[\\s:—–-]*(.*)$`, 'i');
+  const VERDICT_RE = new RegExp(`^\\s*(?:[•●]\\s*)?([*_]{1,3})?\\s*HIVE_VERDICT:\\s*(${alt})\\b[\\s:—–-]*(.*)$`, 'i');
   // Scan newest-first so the agent's final conclusion wins over anything it
   // merely quoted or considered earlier in the transcript.
   for (let i = lines.length - 1; i >= 0; i--) {
     const m = VERDICT_RE.exec(lines[i]);
     if (!m) continue;
-    const reason = (m[2] || '').trim();
+    const emphasis = m[1] || '';
+    let reason = (m[3] || '').trim();
+    // A matching Markdown delimiter at the far end closes the optional
+    // leading emphasis; it is not part of the human-readable reason. Only
+    // remove it when an opener was present, so an ordinary reason ending in
+    // `*` or `_` is preserved verbatim.
+    if (emphasis && reason.endsWith(emphasis)) {
+      reason = reason.slice(0, -emphasis.length).trimEnd();
+    }
     // tmux may wrap the prompt's instruction so its quoted marker lands at a
     // visual line start; its giveaway is the literal "<short reason>"
     // placeholder. Never treat that echo as a real verdict.
@@ -1917,7 +1928,7 @@ function detectHiveVerdict(lines, wanted) {
     // compares it against the line that was already on the pane when the task's
     // prompt was delivered, which is how a verdict gets attributed to a task at
     // all (#5650).
-    return { verdict: m[1].toLowerCase(), reason, line: lines[i] };
+    return { verdict: m[2].toLowerCase(), reason, line: lines[i] };
   }
   return null;
 }
--- a/bin/contributor-relay.test.js
+++ b/bin/contributor-relay.test.js
@@ -5865,6 +5865,23 @@ test('#5376 a task completes on the sentinel even while the chrome says the CLI 
   } finally { teardown(relay); }
 });
 
+test('#6492 a Markdown-bold verdict completes goose with the verdict signal', () => {
+  const pane = [
+    '**HIVE_VERDICT: complete — Configuration validated; next steps outlined.**',
+    'unknown backend chrome',
+  ].join('\n');
+  const relay = loadRelay({ backend: 'goose', paneText: pane });
+  try {
+    dispatchTask(relay, 't-goose-bold-verdict');
+    relay.__crashTick();
+
+    const completed = relay.__sent.filter(m => m.type === 'task_complete');
+    assert.strictEqual(completed.length, 1,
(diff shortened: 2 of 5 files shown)


clubanderson (approved):
Reviewed: tolerant regex only relaxes leading emphasis, trailing delimiter stripped only when an opener matched, prompt-echo guard intact. Locally: 330/330 relay tests, pkg/dashboard green, gofmt clean. The red `test (rest 1/3)` is the unrelated v4 time-bomb (TestPersistStateRoundTripUsesInjectedPaths, fixed in #6499). Merging with admin.

Issue #6500, opened 2026-09-10
🐛 bug: the kubestellar hive's agents all fail with false claim of lack of copilot license


MikeSpreitzer:
## What happened?

@waltforme configured the agents of the kubestellar/kubestellar hive to use Copilot. Today each one's Terminal shows failure due to lack of license. Jun is a member of the CNCF Maintainers organization.

Following is what the Terminal for the quality agent is currently showing.

>  ● MCP Servers reloaded: 1 server connected
> 
>  ✗ You are not licensed to use Copilot. (Request ID: CF24:249477:13194BA:1505534:6AA2A42E) 

All the enabled agents of that hive (https://hosted-available-oke-260812-0nzc-available-ok-w71o.hive.hivecommons.dev) show the same thing.

## Expected behavior

Successful configuration stays successful.

## Reproduction steps

1. Whatever Jun did
2. Look again later at an agent's Terminal

## Environment

- Branch/version: `v4` / commit 55bd2bcbf095096c3d2c045806a0b9c81ae4cd56
- Deployment method: Hive Hub
- Operating system or Kubernetes version: unknown
- Relevant configuration (redact secrets): ?

## Logs or screenshots

Shown above.

clubanderson:
Looked at the spoke directly (`hive-hosted-hosted-available-oke-260812-0nzc-available-ok-ba44`, pod up 22h, no restarts, nothing changed in hive).

**What's happening:** the Copilot CLI in that hive is logged in as `waltforme` (device-flow login, stored in the CLI's own config). Since ≈09:30 ET today GitHub's Copilot API answers every request with HTTP 403 `unauthorized: not licensed to use Copilot` — visible both in the agent panes (the Request IDs you quoted) and in hive's own model-discovery log:

```
copilot-models: Request models.list failed … {"kind":"http","status":403,"statusText":"Forbidden","body":"unauthorized: not licensed to use Copilot\n"}
```

The requests are routed to `api.enterprise.githubcopilot.com`, i.e. GitHub resolves Jun's Copilot entitlement to an *enterprise/business seat* (the CNCF maintainers program), and that seat is currently reporting as not licensed. Hive isn't doing anything different from yesterday; the entitlement on the GitHub side changed. Other Copilot-backed hives on the same cluster are fine, so it's account-specific, not a platform outage.

[...]

PR #6514, opened 2026-09-10
🐛 fix(copilot): stop stale CLI accounts replacing licensed tokens

Danathar:
## Problem and user impact

Copilot-backed hives could move every agent onto a stale or unrelated GitHub identity even when the deployment supplied a licensed `COPILOT_GITHUB_TOKEN` or the operator had just completed a successful dashboard login. Because the shared Copilot CLI config is consumed by every agent, one wrong reconciliation produced the fleet-wide `You are not licensed to use Copilot` failure reported in #6500.

The failure was especially confusing because the dashboard login completed successfully and the configured token remained valid. Up to 30 seconds later, however, the session refresher could reverse that login and inject a different account into every agent.

## Root cause

Hive maintains the Copilot credential in three related places: the configured environment token, the durable dashboard token file, and the Copilot CLI's shared `config.json`. The bidirectional refresher added to preserve in-agent `/login` sessions treated every differing CLI token as newer, regardless of where the manager-held token came from. That inverted the Copilot CLI's documented credential precedence: an older CLI config entry could overwrite the explicit `COPILOT_GITHUB_TOKEN` in memory and in the durable file.

Dashboard device-flow completion also updated only the durable file and manager cache, not the shared CLI identity. The next refresh therefore saw the old CLI token and promoted it back over the newly authorized token.

Finally, `extractCopilotToken` iterated the `copilotTokens` Go map and returned the first usable value. A multi-account config therefore selected an identity nondeterministically instead of honoring `lastLoggedInUser`; an unlicensed secondary account could become the fleet credential simply because its map entry happened to be visited first.

## Implementation

- Track whether the manager token is authoritative. A non-empty startup `COPILOT_GITHUB_TOKEN` and runtime dashboard token changes are authoritative; a token discovered from an in-agent CLI login remains promotable to preserve the existing bidirectional-login behavior.
- When an authoritative token conflicts with the shared CLI config, replace the stale token map and identity rather than promoting the CLI token. The token owner is resolved through the existing GitHub `/user` lookup and written in the canonical Copilot identity shape; the existing host-keyed fallback remains available when that lookup cannot complete.
- Make dashboard login activation update the shared CLI identity and the manager cache after the durable atomic write. The manager cache is still marked authoritative if the config write fails, so the periodic reconciler retries in the safe direction instead of restoring the superseded token.
- Select the token for `lastLoggedInUser` when reading multi-account configs. Legacy configs without a valid active identity are promoted only when they contain one distinct usable token; ambiguous maps now no-op instead of depending on randomized map iteration.
- Add a `fixed` changelog fragment describing the operator-visible behavior change.

The CLI-to-durable promotion path is intentionally retained for operators who authenticate from inside an agent with `/login`. It is suppressed only while a token has an explicit higher-precedence source. The implementation also avoids comparing token timestamps: the CLI may rewrite its config while rejecting a stale credential, so file recency is not a reliable statement of operator intent.

## Regression coverage

The new and extended tests prove that:

- both supported Copilot token encodings remain readable;
- `lastLoggedInUser` selects the licensed identity from a multi-account config;
- a missing active token does not fall through to a different account;
- an identity-less config with different tokens is treated as ambiguous;
- an authoritative token replaces a stale CLI token and canonicalizes the selected identity without overwriting its durable source;
- manager construction marks `COPILOT_GITHUB_TOKEN` authoritative;
- a dashboard token remains authoritative in memory even if its immediate CLI-config write fails, allowing the reconciler to retry;
- dashboard persistence still uses an atomic `0600` file and hands the new token to manager activation.

Verification:

```text
$ go test ./pkg/agent -run 'Test(ExtractCopilotToken|WriteDurableCopilotToken|SyncCopilotToken|NewManager|ActivateCopilotToken|RestoreCopilotTokens|RefreshCopilotSessionToken)' -count=1
ok  github.com/hivecommons/hive/pkg/agent  0.119s

$ go test ./pkg/dashboard -run 'Test(CopilotAuth|PollCopilotToken|SaveCopilotToken|StopCopilotPoll)' -count=1
ok  github.com/hivecommons/hive/pkg/dashboard  1.345s

$ git diff --check
(no output)
```

I also ran `go test ./pkg/agent ./pkg/dashboard`. Both packages compiled and the dashboard package passed, but the broad agent package has unrelated environment-sensitive failures in this checkout: an empty Claude settings fixture, two kick prompt timing assertions, two launch-scrubber tests whose `/tmp` stderr logs were empty, and a tmux test whose socket path exceeded the platform limit. The focused Copilot tests pass independently as shown above.

## Scope and recovery

This change selects and preserves the operator-provided GitHub credential; it does not grant a Copilot license or attempt an automatic device flow. Deployments whose valid token is still present in `COPILOT_GITHUB_TOKEN` recover automatically when reconciliation replaces the stale CLI identity. If a prior refresh already overwrote the only durable copy of the valid dashboard token, the operator must complete dashboard login once after rollout; subsequent refreshes will preserve that login.

Related to #6500 (the reported spoke's cause was a GitHub-side licence state for the logged-in account — see the root-cause comment on #6500; this PR fixes a hive-side token-precedence defect that can produce the same fleet-wide symptom in other topologies)

— hive: backend=codex

### DIFF
--- a/changelog.d/fixed-6500-copilot-license-refresh.md
+++ b/changelog.d/fixed-6500-copilot-license-refresh.md
@@ -1,0 +1,6 @@
+- Fixed Copilot agents incorrectly reporting that they had no license when a
+  stale or unrelated account remained in the shared CLI config. Explicit
+  `COPILOT_GITHUB_TOKEN` credentials and fresh dashboard logins now replace
+  stale CLI identities, while multi-account configs promote only the selected
+  account instead of whichever token Go map iteration returned first
+  ([#6500](https://github.com/hivecommons/hive/issues/6500)).
--- a/src/pkg/agent/copilot_token_promote_test.go
+++ b/src/pkg/agent/copilot_token_promote_test.go
@@ -8,8 +8,8 @@ import (
 	"github.com/hivecommons/hive/pkg/config"
 )
 
-// extractCopilotToken must handle both value shapes the CLI writes: a bare
-// string and a {"token":…} object; and return "" when there is none.
+// extractCopilotToken must handle both value shapes the CLI writes, honor the
+// active identity, and refuse ambiguous multi-account configs.
 func TestExtractCopilotToken(t *testing.T) {
 	dir := t.TempDir()
 	write := func(body string) string {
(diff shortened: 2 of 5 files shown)


clubanderson:
Reviewed and merging — thanks @Danathar, the precedence fix and deterministic `extractCopilotToken` are both real defects on v4 and the tests cover them.

One wording change to the body: `Fixes #6500` → `Related to #6500`. On the reported spoke there was no `COPILOT_GITHUB_TOKEN`, a single `copilotTokens` entry, and GitHub's Copilot API returned 403 `not licensed` for that account directly — so #6500 stays open until that seat is sorted, while this PR closes the precedence hole that would have caused the same symptom elsewhere. Also noting for the support playbook: once `authoritative` is set it is one-way, so recovering from a stale-but-authoritative token needs a dashboard login or env update, not an in-agent `/login`.

Issue #6489, opened 2026-09-10
🐛 bug: unable to start terminal access for agent with local model


balazsdukai:
## What happened?

When i click on the `Open a terminal` button in section 5 in the Getting started screen I get an error popup with `Could not prepare terminal access: service starting up, please retry`. However, hive has been running over a day and I do see that the agents are accessing the local llm.

## Expected behavior

I can start a terminal in the dashboard to interact with a running agent.

## Reproduction steps

1. Check the dashboard
2. I have completed steps 1,2,3,4,6,7 from the getting started guide

## Environment

- Branch/version: 378936cd2f4604df562742aeff1de0759555b8b0 (branch v4)
- Deployment method: Docker Compose 
- Operating system version: Ubuntu 24.04
- Relevant configuration (redact secrets): Using local hosted llm that are served from Unsloth via an OpenAI api. I configured the models directly in the `hive.yaml` config file, by configuring a `governor.gateway` and using that gateway's name in the agent definitions. 

Here is the `supervisor` agent definition:

```yaml
    supervisor:
        id: supervisor
        backend: unsloth-local
        model: unsloth/gemma-4-E4B-it-qat-GGUF
        beads_dir: /data/beads/supervisor
        enabled: true
        replicas: 1
        paused: false
        clear_on_kick: true
        cli_pinned: false
        model_owner: ""
        backend_owner: ""
        pause_owner: ""
        stale_timeout: 0
        restart_strategy: ""
        launch_cmd: ""
        display_name: Supervisor
        description: Orchestrates all agents — sweeps, enforces cadence, monitors health
        role: supervisor
        sort_order: 10
        emoji: "\U0001F451"
        color: '#e74c3c'
        aliases:
            - su
        lane_keywords: []
        detect_keywords:
            - supervisor
            - sweep
            - monitor
        kick_template: ""
        include_repos: true
        metrics_collector: ""
        bead_role: supervisor
        stats_display: []
        acmm_levels: []
        mode: ""
        on_demand: false
        caveman_mode: ""
```

Here is the `governor.gateway` definition:

```yaml
    gateways:
        - name: unsloth-local
          kind: custom
          endpoint: http://host.docker.internal:8888
          api_key_env: ""
          api_key_file: /secrets/unsloth_api_key
          default_model: unsloth/Qwen3-Coder-30B-A3B-Instruct-GGUF
          ca_bundle: ""
```

## Logs or screenshots

[...]

PR #6519, opened 2026-09-10
📖 [strategist] docs+test: gateway-path support tier and hermetic translator canary

### LINKED ISSUE #6515: [strategist] Local-model gateway path has no support-tier bar or smoke canary — self-hosted-LLM adopters break silently
## Strategic Finding

**Type**: adoption-blocker
**Horizon**: near-term

Hive's backend quality machinery covers only subscription CLI backends, while the OpenAI-compatible gateway path — the on-ramp for every self-hosted/local-model adopter — has neither a support-tier statement nor any CI coverage:

- `src/docs/backend-support-tiers.md` (tracking #6231) defines T1/T2/T3 exclusively for the 12 shell-known CLI backends; the word "gateway" does not appear in it. `vllm`, `llm-d`, `litellm`, `watsonx`, and named Model Gateways (`docs/inference-backends.md`) have no acceptance bar and no declared tier.
- `.github/workflows/backend-smoke.yml` canaries only `claude` and `codex` (matrix at line ~100). No scheduled or PR-time check exercises the gateway route (agent → Anthropic-to-OpenAI translator → gateway), so regressions on that path ship blind.
- Fresh external evidence: #6489 (balazsdukai, 2026-09-10) — an adopter running Unsloth-served local models through a configured `governor.gateway` has working inference but a broken dashboard terminal-access path ("service starting up, please retry" persisting >1 day). The agents-work-but-features-don't failure mode is exactly what an untiered, uncanaried path produces.

## Rationale

Local/self-hosted LLM support is a differentiating adoption funnel: it removes the subscription-credential barrier for evaluators and is the only path for privacy-constrained orgs. Today that funnel's reliability is undefined — docs describe how to configure gateways but the project makes no statement about what is expected to work, and nothing in CI would notice when it stops working. Adopters like the #6489 reporter hit breakage on day 2 with no tier label telling them whether this path is supported, experimental, or best-effort.

## Proposed Next Step

Extend the support-tier policy (backend-support-tiers.md / #6231) with a **gateway-path tier section**: which dashboard/agent features are guaranteed on gateway backends (terminal access, model picker, token metering, login flows), which are CLI-only, and what a gateway smoke check must prove. A minimal hermetic canary — a fake OpenAI-compatible endpoint driven through the translator, as a cheap third arm of `backend-smoke.yml` or an ordinary CI test — would make the tier claim enforceable.

Related: #6489 (symptom), #6231 (tier-bar tracker), docs/inference-backends.md, src/docs/backend-support-tiers.md, .github/workflows/backend-smoke.yml

---
*Filed by strategist agent (ACMM L5 — hold-gated mode)*
---
🐝 **Hive Agent**: `strategist` | **Instance:** `hosted-available-oke-11-placeholder-r05x` | **SHA:** `unknown`

— hive: agent=strategist backend=copilot model=claude-sonnet-4-6

clubanderson:
## What changed

**Docs** (`src/docs/backend-support-tiers.md`): added a new "Gateway-path
tier" section, matching the existing tier document's structure and voice.
It states, per dashboard/agent feature, what is guaranteed vs best-effort vs
not-applicable for the OpenAI-compatible gateway path (`vllm`, `llm-d`,
`litellm`, `watsonx`, named Model Gateways) — derived from reading the
actual code, not from vendor claims:

- **Model inference (chat)** — guaranteed; the translator (`forwardToInference`
  in `pkg/proxy/anthropic_translate.go`) is the whole path.
- **Model picker / model discovery** — guaranteed with unverified fallback;
  `/v1/models` probe in `pkg/dashboard/gateways.go`, falls back to static/
  configured lists marked unverified on failure.
- **Login flow** — not applicable; gateways authenticate via
  `api_key_env`/`api_key_file` (`config.GatewayConfig.ResolveAPIKey`), no
  device-flow `/login` exists for this path.
- **Token metering** — guaranteed for the translator path itself (`InferenceSink`
  records usage from every translator response regardless of gateway kind);
  per-kind claim depends on the upstream actually emitting a usage block.
- **Terminal access** — best-effort, currently coupled to hub-provisioned
  signing material (`terminalassert.SigningKey()` / `canMintTerminalAssertion`
  in `pkg/dashboard/terminal_handoff.go`). Described neutrally per the issue:
  a standalone-compose spoke without hub-issued key material returns 503 on
  handoff regardless of gateway health — this is the shape of #6489; a fix to
  provision a standalone-compose signing key is tracked separately and isn't
  changed here.

It also defines what a **gateway smoke check must prove** (request-shape
survival, auth reaching the wire, both response shapes decoding, failure not
swallowed), and links `pkg/proxy/gateway_path_canary_test.go` as the current
implementation of that list.

`docs/inference-backends.md` gets one added sentence linking to the new
section where gateways are introduced.

**Hermetic canary** (`src/pkg/proxy/gateway_path_canary_test.go`, new file):
drives the real Anthropic→OpenAI translator (`forwardToInference`) against an
`httptest` fake OpenAI-compatible `/v1/chat/completions` endpoint for a
**named custom gateway authenticated via `api_key_file`** — the specific gap
the issue names, since the existing `inference_integration_test.go` /
`inference_auth_test.go` cover `vllm`/`litellm` bearer-auth cases but not a
named gateway resolving its key from a file the way `cmd/hive/main.go`'s
inference-route callback actually does (`config.GatewayConfig.ResolveAPIKey`).
Three tests:

- `TestGatewayPathCanary_NamedGatewayAPIKeyFileRequestShape` — resolves the key
  from a temp secrets dir via `config.SetSecretFileRootsForTest`, asserts model
  passthrough, ordered system/user/assistant message-role mapping, the
  resolved file key reaching the upstream `Authorization` header, and the
  response translating back to the Anthropic shape (including `usage`).
- `TestGatewayPathCanary_NamedGatewayWrongKeyFails404sTheGatewayRoute` — proves
  the canary isn't vacuously green: a mismatched key surfaces as a `401`, not a
  swallowed success.
- `TestGatewayPathCanary_StreamingNamedGatewayMapsUsageAndDeltas` — same
  named-gateway/api_key_file shape, streaming SSE case, asserting deltas and
  the terminal usage chunk map to the correct Anthropic SSE events.

Deterministic, no network, no sleeps — an ordinary Go test, not a new CI
workflow arm.

## Why

`src/docs/backend-support-tiers.md` (tracking #6231) defines T1/T2/T3
exclusively for the 12 shell-known CLI backends; the gateway path used by
every self-hosted/local-model adopter had no stated support tier and no CI
coverage of the translator, so regressions there ship blind — exactly what
#6489 exhibits (inference working, dashboard terminal access silently broken
for over a day with no signal). This PR gives the gateway path a documented
acceptance bar and a cheap, deterministic PR-time canary that would catch a
regression in request/response translation, auth wiring, or error handling.

## How tested

```
cd src && go build ./...
go vet ./pkg/proxy/...
go test ./pkg/proxy/ -run TestGatewayPathCanary -count=1 -race -v   # PASS (3/3)
go test ./pkg/proxy/ -count=1 -race                                  # PASS (full package, ~53s)
gofmt -l pkg/proxy/gateway_path_canary_test.go                       # no output (clean)
cd .. && python3 src/scripts/check-docs-links.py src/docs             # "all relative links and anchors resolve" (145 files)
python3 src/scripts/check-docs-links.py docs                          # "all relative links and anchors resolve" (14 files)
```

Not run: no changelog fragment (docs + test-only change per CONTRIBUTING.md).

## Caveats

- The terminal-handoff signing-key requirement on standalone-compose spokes
  (#6489's root cause) is described neutrally in the new tier table; the fix
  itself is out of scope for this PR and tracked separately, as instructed.
- The canary tests the translator in isolation (as `inference_integration_test.go`
  and `inference_auth_test.go` already do for other cases), not the full
  `cmd/hive/main.go` wiring that resolves a named gateway's route at runtime;
  it mirrors that wiring's `InferenceRoute` construction and key-resolution
  call exactly, but a change to that wiring itself would need its own test.

Fixes #6515


clubanderson:
I pushed a small follow-up fix while reviewing this PR.

- renamed the wrong-key canary test from `404s` to `401s` to match what it actually asserts
- tightened the streaming canary to parse the translated Anthropic SSE and verify delta ordering plus usage presence
- fixed `translateOpenAISSEToAnthropic` so the final `message_delta` reports the terminal usage chunk's `output_tokens` instead of the previous stale value

This keeps the new canary aligned with the behavior the docs describe.

clubanderson:
Follow-up on the sub-review of this PR:

A prior commit here (d1651f47) tightened the streaming canary test and, in doing so, fixed a real bug in `translateOpenAISSEToAnthropic`: when a gateway sends `finish_reason` and the terminal `usage` object in the *same* SSE chunk (common for vLLM/LiteLLM-style OpenAI-compatible servers), the old code emitted `message_delta` (which carries `usage.output_tokens`) *before* applying that chunk's usage update, so the reported `output_tokens` reflected the previous chunk's (stale) count instead of the final tally. Reordered so usage is applied first.

Since that touches production code in `pkg/proxy/anthropic_translate.go` (not just docs/tests), this is no longer a docs+test-only change — I added `changelog.d/fixed-6519-gateway-streaming-usage-order.md` and removed the `no-changelog` label accordingly.

Verified: `go build ./...`, `go test ./pkg/proxy/ -count=1 -race` (ok, 45.8s), `go test ./pkg/proxy/ -run TestGatewayPathCanary -count=1 -race -v` (3/3 pass).

One forward-looking note: the new tier doc's terminal-access paragraph describes the current 503-on-standalone-spoke behavior; if sibling PR #6520 (adding a persisted fallback signing-key lane) merges, that paragraph will need a follow-up update to stay accurate.

clubanderson:
Follow-up: #6520 (persisted per-instance terminal signing key on standalone spokes) has merged, so I updated the "Terminal access" row in the gateway-path tier table (`src/docs/backend-support-tiers.md`) from "best-effort, coupled to hub provisioning" to **Guaranteed** — reflects the new lane-3 fallback (`crypto/rand`, persisted at `/data/.hive/terminal-key`, override via `HIVE_TERMINAL_KEY_DIR`), confirms hub-provisioned lanes still take precedence, and notes the read-only-`/data` degrade-to-allowlist caveat. Re-ran `check-docs-links.py` on both `src/docs` (145 files) and `docs` (14 files) — all links/anchors resolve.

Commit: d38c2bd3

PR #6520, opened 2026-09-10
🐛 fix(terminal): auto-provision a per-instance terminal signing key on standalone spokes (#6489)

### LINKED ISSUE #6489: 🐛 bug: unable to start terminal access for agent with local model

clubanderson:
## What changed

Added a third, last-resort key-resolution lane to `pkg/terminalassert.SigningKey()` (aliased as `hub.TerminalSigningKey()`): when neither the hub-injected `HIVE_TERMINAL_KEY` lane nor the `HIVE_HUB_SECRET`+`HIVE_ID` self-derive lane resolves, the spoke now generates a per-instance random key with `crypto/rand` and persists it (0600) under the hive's private data directory (`/data/.hive/terminal-key` by default, overridable via `HIVE_TERMINAL_KEY_DIR`), reused across restarts.

Since the Node proxy (`src/proxy/server.js`) resolves its `TERMINAL_SIGNING_KEY` once, at process start, from its environment — it cannot poll a file the Go side might not have written yet — `deploy/entrypoint.sh` now resolves/generates that same persisted key eagerly, before launching either the Go binary or the Node proxy, and exports it as `HIVE_TERMINAL_KEY`. Both processes therefore converge on lane 1 with an identical value from their very first request, with no race between whichever process happens to start first.

Docs (`src/docs/env-vars.md`, the getting-started wiki) now describe the auto-provisioned key and that `HIVE_TERMINAL_KEY` overrides it.

Files changed:
- `src/pkg/terminalassert/terminal_assertion.go` — lane-3 fallback (`fallbackSigningKey`, `fallbackKeyDir`) + doc comment updates
- `src/pkg/terminalassert/terminal_assertion_test.go` — updated two tests whose old assertions pinned the pre-fix "no hub lane → empty" behavior
- `src/pkg/terminalassert/terminal_key_fallback_test.go` (new) — generation, persistence/reuse, 0600 perms, precedence below both hub lanes, never touches disk when hub-provisioned, per-instance randomness, end-to-end mint/verify round trip
- `src/deploy/entrypoint.sh` — `hive_ensure_terminal_signing_key`, called before the Go binary and Node proxy start
- `src/deploy/test_entrypoint_terminal_key_fallback.sh` (new) — extracts and executes the real function from the shipped entrypoint
- `.github/workflows/v2-ci.yml` — wires the new suite
- `src/pkg/dashboard/terminal_handoff_test.go`, `terminal_handoff_mint_test.go` — the two tests that pinned "no hub lane → 503" now assert the fixed behavior (200 + minted handoff/cookie), since that scenario is exactly what standalone Compose spokes hit
- `src/docs/env-vars.md`, `src/deploy/data/wiki/getting-started.md` — documentation
- `changelog.d/fixed-6489-standalone-terminal-key.md`

## Why

`Open a terminal` POSTs `/api/terminal/handoff`. The dashboard refused with `503 {"error":"terminal handoff requires terminal signing key and hive id"}` because `hub.TerminalSigningKey()` resolved empty on a standalone (non-hub-provisioned) docker-compose spoke: it only had the hub-injected `HIVE_TERMINAL_KEY` lane and the `HIVE_HUB_SECRET`+`HIVE_ID` self-derive lane, and a standalone spoke has neither — audit N3 deliberately removed the fleet-uniform fallbacks that used to sit there, so there was no third lane at all. This made browser terminals structurally unavailable on every self-hosted token-secured deployment, and the nginx gateway masked the real 503 as a generic "service starting up" toast (already fixed separately in #6494/#6496).

**Why per-instance random, not derived from anything:** the whole reason lane 3 is needed is that the spoke has no identity-bound input to derive from (no `HIVE_ID`) and no hub-issued material (no `HIVE_TERMINAL_KEY`/usable `HIVE_HUB_SECRET`). Deriving from any value every standalone spoke could plausibly share (e.g. a compile-time constant) would just reintroduce the exact fleet-uniform forgery lane audit N3 closed, under a new name. `crypto/rand` is the only source that is per-instance by construction, matching the existing precedent in `pkg/dashboard`'s invite-key fallback (`inviteSigningSecret`), which uses the identical generate-once/persist/reuse pattern for the equivalent invite-link-signing gap.

**Why only when both hub lanes are absent:** a hub-provisioned hive must always use its per-hive hub-derived key — the fallback must never compete with or shadow it. The new lane is purely the third rung of the existing resolver (`if lane1 != "" { return }`, `if lane2 != "" { return }`, `return lane3()`), so it structurally cannot fire while either hub lane resolves; this is also asserted directly (`TestFallbackKeyDoesNotApplyWhenHubProvisioned` proves the fallback file is never even touched in that case).

**Why the entrypoint also needs to change:** the Go dashboard's fallback is *lazy* — it only runs the first time a terminal is actually requested. The Node proxy's key is *eager* — computed once at module load. Without the entrypoint change, whichever of the two happened to see a request first would decide the key, and the other could start with no value at all (an empty verifier key fails closed, so a purely-lazy fix would still 503 the very first terminal request after a fresh boot). Making the entrypoint provision the file up front and export it as `HIVE_TERMINAL_KEY` closes that race entirely.

## How tested

```
cd src && go build ./...                                                   # OK
go vet ./pkg/dashboard/ ./pkg/terminalassert/ ./pkg/hub/                   # OK, no findings
go test ./pkg/dashboard/ ./pkg/terminalassert/ -count=1                    # ok  pkg/dashboard   44s
                                                                            # ok  pkg/terminalassert  0.8s
go test ./pkg/hub/ -count=1                                                # ok  130s (unaffected, run once for confidence)
gofmt -l <all touched .go files>                                          # empty (clean)
bash -n src/deploy/entrypoint.sh                                          # OK
shellcheck src/deploy/entrypoint.sh                                       # 73 findings before and after (no new ones)
shellcheck -x src/deploy/test_entrypoint_terminal_key_fallback.sh          # only the same info-level SC1091 every sibling suite has
bash src/deploy/test_entrypoint_terminal_key_fallback.sh                   # 10 passed, 0 failed
bash src/deploy/test_deploy_suites_wired.sh                                # 39 suites checked, 0 orphaned
bash src/deploy/test_standalone_runtime_parity.sh                          # 33 passed, 0 failed, 7 accepted exceptions (unaffected)
bash src/deploy/test_compose_quickstart_contract.sh                        # 8 passed, 0 failed (unaffected)
```

Fixes #6489

### DIFF
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -99,6 +99,20 @@ jobs:
           HIVE_TEST_REQUIRE_BEHAVIOURAL: '1'
         run: bash deploy/test_entrypoint_dangling_keyfile.sh
 
+      # #6489: `Open a terminal` 503ed forever on a standalone, non-hub-provisioned
+      # docker-compose spoke — terminalassert.SigningKey() (hub.TerminalSigningKey)
+      # had no lane left once HIVE_TERMINAL_KEY and HIVE_HUB_SECRET+HIVE_ID were
+      # both absent (audit N3 deliberately removed the fleet-uniform fallbacks
+      # that used to sit there). hive_ensure_terminal_signing_key now auto-
+      # provisions a persisted per-instance key and exports it as
+      # HIVE_TERMINAL_KEY before either the Go binary or the Node proxy starts, so
+      # both processes agree on it from their first request. EXECUTES the real
+      # function extracted from the shipped entrypoint, so a regression fails the
+      # PR rather than a standalone hive.
+      - name: Entrypoint standalone terminal-key fallback tests (#6489)
+        run: bash deploy/test_entrypoint_terminal_key_fallback.sh
+
+
       # #5369: the `chown -R dev:node /data` in the entrypoint's root phase is
       # guarded on `[ "$DATA_OWNER" != "1001" ]`, and the Dockerfile already
       # does that chown at BUILD time — so the guard is false on every normal
--- a/.github/workflows/v2-ci.yml
+++ b/.github/workflows/v2-ci.yml
@@ -112,7 +112,6 @@ jobs:
       - name: Entrypoint standalone terminal-key fallback tests (#6489)
         run: bash deploy/test_entrypoint_terminal_key_fallback.sh
 
-
       # #5369: the `chown -R dev:node /data` in the entrypoint's root phase is
       # guarded on `[ "$DATA_OWNER" != "1001" ]`, and the Dockerfile already
       # does that chown at BUILD time — so the guard is false on every normal
--- a/changelog.d/fixed-6489-standalone-terminal-key.md
+++ b/changelog.d/fixed-6489-standalone-terminal-key.md
@@ -1,0 +1,1 @@
+- Standalone, non-hub-provisioned Docker Compose hives now auto-provision a persisted per-instance terminal signing key, so the dashboard's "Open a terminal" button no longer fails with `terminal handoff requires terminal signing key and hive id`; set `HIVE_TERMINAL_KEY` to override it ([#6489](https://github.com/hivecommons/hive/issues/6489)).
--- a/src/deploy/data/wiki/getting-started.md
+++ b/src/deploy/data/wiki/getting-started.md
@@ -81,6 +81,15 @@ A classic PAT needs `repo` scope; see
 [github-app-setup.md](https://github.com/hivecommons/hive/blob/v4/src/docs/github-app-setup.md#personal-access-token-pat-scopes)
 for the App path and the full scope list.
 
+The dashboard's **Open a terminal** button needs a terminal-assertion signing
+key; a standalone Compose hive like this one auto-provisions a per-instance key
+on first boot (persisted under `/data/.hive/terminal-key` in the `hive-data`
+volume) so it just works, with no extra setup. Set `HIVE_TERMINAL_KEY` in
+`src/.env` yourself (e.g. `openssl rand -hex 32`) if you would rather pin a
+specific value — it always overrides the auto-provisioned one. See
+[env-vars.md](https://github.com/hivecommons/hive/blob/v4/src/docs/env-vars.md#spoke-side-derived-keys)
+for the full resolution order.
+
 ## Verify Your Hive Is Healthy
 
 The gateway publishes port 3001 whether or not the proxy behind it came up, so
--- a/src/deploy/entrypoint.sh
+++ b/src/deploy/entrypoint.sh
@@ -2228,6 +2228,70 @@ if [ -n "${HIVE_CONFIG:-}" ]; then
   set -- "$@" --config "$HIVE_CONFIG"
   echo "[entrypoint] Config path pinned for the Go binary: --config $HIVE_CONFIG"
 fi
+
+# Auto-provision a per-instance terminal signing key on standalone spokes
+# (#6489). `Open a terminal` used to 503 forever on a docker-compose spoke that
+# is not hub-provisioned: hub.TerminalSigningKey() (now
+# terminalassert.SigningKey()) needs either HIVE_TERMINAL_KEY (hub-injected) or
+# HIVE_HUB_SECRET+HIVE_ID (self-derive), and a standalone spoke has neither, by
+# design — audit N3 deliberately left no fleet-uniform or public-value fallback
+# in that resolver.
+#
+# The Go dashboard now has its OWN persisted-random-file fallback lane for this
+# exact case (pkg/terminalassert's fallbackSigningKey, same pattern as
+# pkg/dashboard's invite-key fallback), but it is lazy — it only runs the first
+# time a terminal is actually requested — and the Node proxy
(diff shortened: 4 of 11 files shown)


clubanderson:
Security review of the lane-3 fallback key change:

- **Precedence verified**: both `terminalassert.SigningKey()` (Go) and `TERMINAL_SIGNING_KEY` (`src/proxy/server.js`) check lane 1 (`HIVE_TERMINAL_KEY`) then lane 2 (`HIVE_HUB_SECRET`+`HIVE_ID`) before lane 3; `entrypoint.sh`'s `hive_ensure_terminal_signing_key` short-circuits identically, so a hub-provisioned hive never touches the fallback file. `TestFallbackKeyDoesNotApplyWhenHubProvisioned` confirms the file is never even created in that case.
- **Entropy/persistence**: 32 bytes from `crypto/rand`/`/dev/urandom`, atomic temp+rename in both the Go and shell implementations, 0600 file / 0700 dir. No race: the entrypoint resolves/generates the key and exports `HIVE_TERMINAL_KEY` before either process starts, so Go's lazy lane 3 never fires in the normal container boot path (env is already set).
- **Node/Go consistency**: `src/proxy/server.js`'s `TERMINAL_SIGNING_KEY` intentionally has only lanes 1+2 (no lane 3 of its own) — correct, since it relies on the entrypoint injecting the persisted key as `HIVE_TERMINAL_KEY`.
- Ran: `go build ./...`; `go vet ./pkg/terminalassert/ ./pkg/dashboard/ ./pkg/hub/` (clean); `go test ./pkg/terminalassert/ -count=1 -v` (20/20 pass); `go test ./pkg/dashboard/ -count=1` (ok, 46.6s); `go test ./pkg/hub/ -count=1` (ok, 131.8s); `gofmt -l` on touched files (clean); `bash -n entrypoint.sh` + `shellcheck entrypoint.sh` (no new findings around the added function); `bash deploy/test_entrypoint_terminal_key_fallback.sh` (10/10 pass); `bash deploy/test_deploy_suites_wired.sh` (39 suites, 0 orphaned). Docs (`env-vars.md`, getting-started wiki) match the implementation; changelog fragment is present and correctly categorized.
- Pushed a0a3bd60: trivial whitespace fix (stray double blank line in `v2-ci.yml`) — no functional change.

**One documented, non-blocking design note** (not requesting a change): on a read-only `/data` mount, persistence fails and `fallbackSigningKey()` returns a fresh in-memory key on every call rather than failing closed — the code comments explain this is intentional (a working-but-non-persistent terminal beats none). Practical effect: repeated Go-side mints would each use a different ephemeral key while the Node proxy's key (fixed at process start from the entrypoint's export, which itself would fail to export anything if the entrypoint's own write also fails) stays constant or empty, so verification degrades to the pre-existing static allowlist rather than the signed path — a safety-preserving degradation, just worth being aware of operationally.

VERDICT: MERGE.

PR #6567, opened 2026-09-10
🐛 fix: repoint retired hub domain defaults to hive.hivecommons.dev

clubanderson:
Every default in this diff named `hive.kubestellar.io`, which no longer serves the hub.

### Why these survived the migration

They do not fail loudly. Two distinct breakages hide behind that host:

1. **The apex still answers**, but its `301` **drops the path and query**. A client keeps working just enough to look fine while landing somewhere else.
2. **`hosted-*.hive.kubestellar.io` is outside the current wildcard certificate** and fails the TLS handshake outright:

```
Connected to hosted-...-gjvq.hive.kubestellar.io (157.151.252.29) port 443
subject: CN=hive.hivecommons.dev
subjectAltName does not match host name ...hive.kubestellar.io
SSL: no alternative certificate subject name matches
```

A spoke that cannot complete a handshake never heartbeats, and a hive that misses `maxHeartbeatAge` (5m) is marked **offline**. So a stale default presents as a mysteriously offline hive rather than as a connection error anyone would think to chase.

### The worst case: contributors

`bin/contributor-agent.sh` and `src/compose-contributor.yaml` both defaulted:

```
HIVE_HUB=wss://hive.kubestellar.io:3001/contribute
```

That is dead twice over — the host redirects (and `wss://` will not follow a 301) **and** port 3001 is now closed. Verified:

| URL | Result |
|---|---|
| `https://hive.hivecommons.dev/contribute` | **200** |
| `https://hive.hivecommons.dev:3001/contribute` | `000` (closed) |
| `https://hive.kubestellar.io/contribute` | `301` (drops path) |

Any contributor who had not overridden `HIVE_HUB` was dialling a dead endpoint. This was found on a real machine, where `~/.config/hive/contributor.env` still carried the retired host.

### Also fixed

- **OpenAPI `servers` entry** — handed a redirecting host to anything generating a client from the spec.
- Self-upgrade `Origin` header, snapshot-page fallback, and the user-facing **"Back to the hub"** link (the 301 drops `/dashboard`).
- `federation-heartbeat.sh` registry default.

### Deliberately NOT changed

- **`defaultHubPublicURL` / `defaultHubCanonicalHost` / `defaultHubSpokeDomain`** are a coupled set: the canonical host seeds session-cookie scoping, the spoke domain seeds redirect trust. Moving any one in isolation fails the others' tests, and all three are environment-overridden in production, so they are latent rather than live. `TestCoupledHubDefaultsAreDeliberatelyUnchanged` records this so the remaining reference is not mistaken for an oversight and "fixed" in isolation.
- **`hive.kubestellar.io/v1`** API group versions and namespace label keys — identifiers, not URLs.
- **Git bot email domain and legacy cookie domains** — these must keep naming the old host, to preserve DCO identity on existing commits and to expire the previous build's cookies.

### Verification

- New regression tests over the **embedded** OpenAPI document (not the working tree — the tree is not what ships) and the config default. **Both verified to fail when the change is reverted**, so they are not passing vacuously.
- `pkg/hub` and `pkg/dashboard` suites pass in full.
- `pkg/config` has one failure, `TestEntrypointHardensRuntimeConfigItRecreates`, which **also fails on clean `v4`** — it needs `chown dev:node` and cannot pass on macOS. Unrelated to this change.


clubanderson:
Reviewed against the checklist:

1. **Grep for remaining `hive.kubestellar.io` defaults** — swept `src`, `bin`, `dashboard` for `.go/.sh/.yaml/.json/.md`. Every remaining hit is one of: the `hive.kubestellar.io/v1` API group / `hive.kubestellar.io/*` label-annotation keys (identifiers, not URLs), the legacy cookie-domain constants (`defaultLegacyHubCookieDomain`, `legacyImpersonateCookieDomain` — must keep naming the old host to expire old cookies), the coupled `defaultHubPublicURL/CanonicalHost/SpokeDomain` set (deliberately unchanged, with a test recording why), the git-bot email domain, the `legacy-redirect`/`dibs-domain-cutover` deploy configs (which correctly still name the *source* host of a redirect), and doc/comment prose. Nothing missed.
2. **Is `hive.hivecommons.dev` the current hub host?** Verified live: `curl -sI https://hive.hivecommons.dev/` → `200`; `curl -sI https://hive.kubestellar.io/` → `301` to `hive.hivecommons.dev`. Matches `src/docs/hosted-hub.md`, `env-vars.md`, `manual-provisioning.md`.
3. **Changelog fragment guard** — read `.github/workflows/changelog-fragment-guard.yml` and `changelog.d/README.md`: the naming rule is `<category>-<pr-or-slug>.md` where `<pr-or-slug>` may be "a PR/issue number, a short slug, or both." No issue number is required by the guard; it only checks the category prefix and that content starts with `- `. `fixed-retired-hub-domain-defaults.md` satisfies both. Searched for a tracking issue (none exists beyond #5925, which is explicitly the separate dibs concern) — a slug-only fragment name is correct here, not a gap.
4. **Tests**: `go build ./...` clean; `go vet`/`golangci-lint run` clean on `pkg/config`, `pkg/hub`, `pkg/dashboard`; `gofmt -l` shows no new issues in files this PR touches; `shellcheck` clean on both scripts; `bash -n` clean; `python3 -c "import yaml"` load of `compose-contributor.yaml` and `openapi.yaml` both OK.
   - `go test -race -count=1 ./pkg/hub/...` → **ok** (168s)
   - `go test -race -count=1 ./pkg/dashboard/...` → **ok** (98s)
   - `go test -race -count=1 ./pkg/config/...` → 1 failure, `TestEntrypointHardensRuntimeConfigItRecreates` (0600 chown assertion, needs `chown dev:node`). Reproduced the identical failure on clean `v4` HEAD (1896d926) — confirmed pre-existing and unrelated to this change, exactly as the PR body states.

No changes needed — the PR is correct and complete as submitted.

Issue #6529, opened 2026-09-10
✨ feat: self-service restore path for pkg/spokebackup archives


clubanderson:
### Summary

`pkg/spokebackup` archives (owner-triggered, `POST /api/backup`) can already be **decrypted and extracted today** with the existing `hive-backup extract -file <archive> -dest <dir>` CLI — it shares `pkg/hubbackup`'s AES-256-GCM sealing, SHA-256 manifest and traversal-safe extraction unchanged (`src/pkg/hubbackup/builder_export.go:1-22`), so `hive-backup extract` accepts a spoke-backup archive with no code change. Placing the extracted `spoke/*` files at the container's `/data` root and `beads/<agent>/*` at `/data/beads/<agent>/`, then (re)starting the container, restores `hive-id`, the GitHub App key(s), the runtime config, and the bead ledger — the entrypoint re-applies ownership/permissions on boot (`src/deploy/entrypoint.sh:262-266`, `:894-931`). This was executed and verified in #6527 (identical `hive-id` and GitHub App key SHA-256 before/after extraction; see the linked PR for command output).

What is **missing** is any tooling or self-service path to *apply* that restore. Concretely:

1. **No dashboard upload/restore endpoint.** `src/pkg/dashboard/backup_api.go` implements only `GET /api/backup/status` and `POST /api/backup` (download). There is no `POST /api/backup/restore` (or similar) that accepts an uploaded `.tar.gz.enc`, decrypts it with the configured key, and writes it into `/data` (and `/data/beads/<agent>/`) on a running or freshly-provisioned spoke. Today, restoring requires direct filesystem/`kubectl exec`/`podman exec` access to the target's data volume — an operator action, not something a hive owner without cluster access can do (the same audience `governor.backup.key_file` was added for, per `src/docs/backup-restore.md`'s "Setting the backup encryption key (hosted flow)" section).
2. **No restore command in `src/cmd/hive-backup`.** `extract` only decrypts to a plain directory; there is no `restore` subcommand that also copies the `spoke/` and `beads/` trees into a target `/data`, applies the archive-path → container-path mapping, and re-runs the ownership fix-ups the entrypoint would otherwise only do at boot.
3. **No validation that a target is actually "fresh".** Nothing warns/refuses if `/data/hive-id` on the target already contains a *different* identity before a restore proceeds — restoring over a live, differently-identified hive would silently splice one hive's config/keys/beads onto another's identity.

### Proposal

[...]

PR #6571, opened 2026-09-10
feat(backup): add `hive-backup restore` for spoke archives

Danathar:


clubanderson:
Reviewed for #6529. Verified: hive-id mismatch refusal (refuses unless `-force`, and correctly refuses when the archive has no hive-id but the destination does), `-dry-run` writes nothing (identity check still runs, so a dry run can still fail closed), traversal safety (delegated to `hubbackup.Extract`, which this doesn't weaken), no symlink following, `spoke/`→dest and `beads/<agent>/`→dest/beads/<agent>/ mapping is correct, secret files are hardened to 0600 regardless of archive-claimed mode (nice extra hardening beyond what was asked), and tests are real (round-trip, refusal/force/same-hive matrix, dry-run-matches-real-run, hub-archive rejection, staging-dir cleanup on every exit path).

Ran locally: `gofmt -l`, `go vet`, `golangci-lint run` (0 issues), `go test -race -count=1 ./cmd/hive-backup/... ./pkg/spokebackup/...` (all pass), `go build ./...` (clean), and `check-docs-links.py` (same 38 pre-existing unrelated broken links, none in backup-restore.md). Changelog fragment is correctly formatted. No changes needed from me — nice work.

PR #6575, opened 2026-09-10
✨ add spoke backup restore command

### LINKED ISSUE #6529: ✨ feat: self-service restore path for pkg/spokebackup archives

clubanderson:
## Summary
- add `hive-backup restore -file <archive> -dest <data-dir>` for pkg/spokebackup archives
- map `spoke/*` to the data-dir root and `beads/*` to `beads/<agent>/`
- default restore to dry-run; require `-confirm` to write and `-force` before replacing a different existing `hive-id`
- harden restore writes against symlink clobbering and preserve/tighten file modes/ownership
- update backup/restore docs and add a changelog fragment

## Safety
Restore is destructive, so the CLI does not write unless `-confirm` is passed. It also refuses archives without `spoke/hive-id` and refuses to restore over a different live `hive-id` unless `-force` is explicitly provided.

## Validation
- `cd src && GOMODCACHE=/tmp/gomodcache GOFLAGS=-modcacherw go build ./...`
- `cd src && GOMODCACHE=/tmp/gomodcache GOFLAGS=-modcacherw go vet ./...`
- `cd src && GOMODCACHE=/tmp/gomodcache GOFLAGS=-modcacherw go test ./pkg/spokebackup/... ./cmd/hive/... ./cmd/hive-backup/...`
- `python3 src/scripts/check-docs-links.py src/docs`

Fixes #6529

### DIFF
--- a/changelog.d/added-6529-spoke-restore.md
+++ b/changelog.d/added-6529-spoke-restore.md
@@ -1,0 +1,1 @@
+- `hive-backup restore` can now apply an owner-triggered spoke backup archive to a target data directory, mapping `spoke/*` and `beads/*` into `/data` while defaulting to a dry run and refusing to overwrite a different existing `hive-id` unless explicitly forced ([#6529](https://github.com/hivecommons/hive/issues/6529)).
--- a/src/cmd/hive-backup/main.go
+++ b/src/cmd/hive-backup/main.go
@@ -12,6 +12,7 @@
 //	hive-backup verify              # verify the newest stored archive
 //	hive-backup verify -file f.enc  # verify a local archive
 //	hive-backup extract -file f.enc -dest ./restore
+//	hive-backup restore -file f.enc -dest /data -confirm
 //	hive-backup list                # list stored archives
 //
 // HIVE_BACKUP_KEY must be set to a 64-character hex AES-256 key. It has no
@@ -26,6 +27,7 @@ import (
 
 	"github.com/hivecommons/hive/pkg/hubbackup"
 	"github.com/hivecommons/hive/pkg/logscrub"
+	"github.com/hivecommons/hive/pkg/spokebackup"
 )
 
 // exitCodeError is returned for any operational failure so a CronJob shows
@@ -48,6 +50,8 @@ func main() {
 		cmdVerify(os.Args[2:], logger)
 	case "extract":
 		cmdExtract(os.Args[2:], logger)
+	case "restore":
+		cmdRestore(os.Args[2:], logger)
 	case "list":
 		cmdList(logger)
 	default:
@@ -63,6 +67,7 @@ Commands:
   run [-local FILE] [-skip-spokes]   create an encrypted backup
   verify [-file FILE]                verify newest stored, or a local archive
   extract -file FILE -dest DIR       decrypt an archive to a directory
+  restore -file FILE -dest DIR       dry-run a spoke restore; add -confirm to write
   list                               list stored archives
 
 Environment:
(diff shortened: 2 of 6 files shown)


clubanderson:
Closing as duplicate: #6571 (Danathar) implementing `hive-backup restore` for #6529 was filed first, reviewed and has merged. #6529 already had an open PR when this was started — agents should defer when a same-issue PR exists.

PR #6585, opened 2026-09-10
🐛 fix(agent): token-restart cap never holds — give-up latch cleared by its own restart's boot pane

### LINKED ISSUE #6578: 🐛 token-restart cap never holds: give-up latch is cleared by the boot pane of its own restart (×1334/24h storms)
## Symptom

Three hives on the fleet page are in permanent restart storms, at rates that are impossible if the cap were holding:

| Hive | Agent | Restarts | Verdict shown |
|---|---|---|---|
| `jeejz/incubator-kie-drools` | `quality` | **×1334/24h** | `agent restarts: quality ×1334/24h (login token refreshed)` |
| `singhar/go-ci` | `guide` | **×1294/24h** | `agent restarts: guide ×1294/24h (login token refreshed)` |
| `cc-deck/cc-deck` | `supervisor` | **×575/24h** | `agent restarts: supervisor ×575/24h (operator)` |

×1334/24h is a restart roughly **every 65 seconds, continuously, for a day**. Meanwhile `tokenRestartMaxAttempts = 3`.

The `(login token refreshed)` suffix identifies the path exactly — that string is only produced by the token-triggered restart at `manager.go:3293`.

## Why the cap does not hold

`decideTokenRestart` correctly caps attempts at 3 and latches `tokenRestartGaveUp` (GUARD 4, #4596). But the reset is unconditional on any non-login pane:

```go
// manager.go:3196-3206
if showsLogin || bobKeyRejected {
    showsLogin = true
    loginStreak++
} else {
    loginStreak = 0
    // The prompt cleared, so a future "token appeared, nudge it"
    // restart is a fresh theory rather than a repeat of one that
    // already failed. Reset both halves of the cap together.
    agent.tokenRestartAttempts = 0
    agent.tokenRestartGaveUp = false
}
```

The comment's premise is *"the prompt cleared"*. But the pane also fails to show a login prompt **while the CLI is still booting** — immediately after the very restart this logic just performed. The pane is blank or painting startup chrome, so `paneShowsLoginPrompt` returns false, and that is read as success.

The loop:

1. Pane shows login → streak builds → 3 token restarts → `tokenRestartGaveUp = true`. Correct so far.
2. The third restart relaunches the CLI. **During boot the pane shows no login prompt.**
3. The `else` branch fires and resets **both** `tokenRestartAttempts = 0` and `tokenRestartGaveUp = false`.
4. The CLI finishes booting and paints the login prompt again.
5. Back to step 1, forever.

The cap is defeated by the side effect of its own remedy. The give-up latch can never survive, because reaching it always involves a restart, and every restart produces the exact pane state that clears it.

Note the contrast with `pkg/watchdog`, which already models this correctly — `classify.go` computes `booting := !obs.StartedAt.IsZero() && now.Sub(obs.StartedAt) < s.BootGrace` and refuses to draw conclusions from a pane that has not had time to paint. The manager's poll loop has no equivalent guard.

The watchdog reconciler is **not** the source here: `reconciler.go:469` explicitly declines to act on `ClassAuthRequired` ("No restart: see PaneClass.Dead").

## Why it matters

- These agents are blocked on a **human** device-flow login. No number of restarts can fix that — the cap exists precisely to hand off to an operator, and it never does.
- `diagnoseStuckLogin` is the operator-facing diagnosis and is only written inside the `tokenRestartGaveUp` branch. Because that latch is cleared every cycle, **the diagnosis never persists**, so the fleet page shows a restart count instead of the actionable "an operator must log in".
- It is ~1300 CLI relaunches/day/agent of wasted spoke CPU, and `prclaims.go:21` already documents restart storms as a correctness hazard.

## Suggested fix

Distinguish *"the login prompt cleared"* from *"the pane has not painted yet"*. Only reset the cap on **positive evidence of a working CLI** — a ready signature (`HasCLIMarker`) — rather than on the mere absence of login chrome. A boot-grace guard mirroring `watchdog.BootGrace` would also work:

```go
} else if !booting {   // or: else if paneHasReadySignature(tail)
    loginStreak = 0
    agent.tokenRestartAttempts = 0
    agent.tokenRestartGaveUp = false
}
```

A regression test should assert that an agent whose pane alternates blank-during-boot → login-prompt reaches `tokenRestartGiveUp` and **stays** there, rather than re-arming.

## Found via

Live fleet inspection on 2026-09-10; verified against `v4` at `d67466a`.

clubanderson:
## The bug

`decideTokenRestart` caps token-triggered restarts at `tokenRestartMaxAttempts = 3` and latches `tokenRestartGaveUp` (GUARD 4, #4596). That part is correct. The defect is in the *reset*, in the pane poller:

```go
// manager.go:3199-3205 (before)
} else {
    loginStreak = 0
    agent.tokenRestartAttempts = 0
    agent.tokenRestartGaveUp = false
}
```

The comment's premise is "the prompt cleared". But a pane also shows no login prompt **while the CLI is still booting** — which is exactly the state the restart this logic just performed creates. So:

1. Login prompt persists, 3 restarts fire, `tokenRestartGaveUp = true`. Correct.
2. The third restart relaunches the CLI. During boot the pane is blank.
3. The `else` branch fires and clears **both** halves of the cap.
4. The CLI finishes booting and repaints the login prompt.
5. Back to 1, forever.

The cap was defeated by the side effect of its own remedy: reaching the latch always requires a restart, and every restart produces the pane state that clears it. Observed on three fleet hives at up to **x1334/24h** — one relaunch every ~65 seconds for a day — against a cap of 3.

## The fix

Re-arm only on **positive evidence the CLI is up**, never on the mere absence of login chrome. Two independent conditions, because either alone is defeatable:

- **A CLI marker in the pane TAIL.** Tail rather than the whole capture, so a stale marker left in scrollback by an exited CLI cannot re-arm the loop.
- **Outside a 60s boot grace after the last token restart.** The marker check alone is not enough: `cliPaneMarkers` includes bare words like `Claude` and `Copilot`, which a CLI can paint in a startup banner while still booting. The grace is pinned to `watchdog.DefaultBootGrace`, itself pinned to `cliReadyTimeout`, so both subsystems draw the boot boundary in the same place — following the issue's observation that `watchdog/classify.go:190` already models this and the manager poll loop had no equivalent.

`lastTokenRestart` is zero for an agent never token-restarted, in which case there is no boot to wait out and only the marker applies.

## Why this also fixes the operator handoff

`diagnoseStuckLogin` is written only inside the `tokenRestartGaveUp` branch. With the latch cleared every cycle the diagnosis never persisted, so the fleet page showed a restart count instead of the actionable "an operator must log in". These agents are blocked on a **human** device-flow login — no number of restarts can fix that, and the cap exists precisely to hand off. Now it does.

## Tests

Added to `stuck_login_diagnosis_test.go`:

- `TestTokenRestartGiveUpSurvivesBootPaneAlternation` — reproduces the storm directly: a pane alternating blank-during-boot to login-prompt must reach `tokenRestartGiveUp` and **stay** there. Asserts the latch holds and that restarts fired never exceed the cap.
- `TestPaneAllowsTokenRestartCapResetRequiresPositiveEvidence` — table over blank boot pane, startup chrome, boot banner carrying a marker word inside the grace, settled prompt outside the grace, never-restarted agent, stale scrollback.
- `TestTokenRestartCapStillReArmsOnRealRecovery` — guards the opposite direction, so the fix cannot strand an agent whose login genuinely cleared.

**Verified these fail without the fix.** Stubbing the new guard to return `true` (the old unconditional behaviour) fails both `TestTokenRestartGiveUpSurvivesBootPaneAlternation` and the table test; restoring it passes. They reproduce the bug rather than merely describing it.

## Validation

- `go test ./pkg/agent/ -count=1` gives **ok, 296.473s** (full package, not just the new tests)
- `go build ./...`, `go vet ./pkg/agent/` clean

## Scope

One behavioural change plus a small testable seam (`paneAllowsTokenRestartCapReset`). No change to the cap, the cooldown, the kick grace, or the restart path itself. The watchdog reconciler is untouched and is not the source here — `reconciler.go:469` already declines to act on `ClassAuthRequired`.

Fixes #6578


clubanderson:
Closing in favor of #6590, opened by @Danathar for the same issue (#6578). Reviewed #6590 in detail — its shouldResetTokenRestartCap already combines this PR's boot-grace-vs-cooldown-style approach with a positive CLI-marker check, reusing the package's existing cliBootGraceSeconds/StartedAt signal rather than a new constant, and its regression tests (simulated 24h poll loop against production detectors, plus a boot-outlasts-grace edge case) are a superset of what's here. Thanks for the investigation — the root-cause diagnosis was correct and matched what #6590 landed on.

PR #6587, opened 2026-09-10
🐛 fix(agent): keep token-restart give-up latch through the boot pane of its own restart (#6578)

### LINKED ISSUE #6578: 🐛 token-restart cap never holds: give-up latch is cleared by the boot pane of its own restart (×1334/24h storms)

clubanderson:
## What changed

`pollTmuxOutputForAgent`'s `else` branch (pane does not show a login prompt)
unconditionally reset both `tokenRestartAttempts` and `tokenRestartGaveUp` to
zero. This is now gated by a new `tokenRestartLatchClears(agent, now)` helper,
which only allows the reset once `cliBootGraceSeconds` has elapsed since
`agent.StartedAt` — the SAME boot-grace signal
`CheckAndRestartCrashedAgents`/`waitForCLIReadyForAgent` already use to avoid
drawing conclusions from a pane that hasn't had time to paint.

Files changed:
- `src/pkg/agent/manager.go` — add `tokenRestartLatchClears`; gate the cap
  reset in `pollTmuxOutputForAgent` on it.
- `src/pkg/agent/stuck_login_diagnosis_test.go` — regression tests.
- `changelog.d/fixed-6578-token-restart-cap-latch.md` — changelog fragment.

## Why

`decideTokenRestart` correctly caps consecutive token-triggered restarts at
`tokenRestartMaxAttempts` (3) and latches `tokenRestartGaveUp` (GUARD 4,
#4596). But the very next poll tick after the 3rd restart fires, the freshly
relaunched CLI is still booting: the pane shows neither the login prompt nor
its ready chrome. That blank/booting pane was indistinguishable from "the
prompt genuinely cleared", so the `else` branch reset the latch it had just
set — the cap was defeated by the side effect of its own remedy, producing
restart storms of up to **×1334/24h** on agents blocked on a human device-flow
login (issue has live fleet data for three affected hives). The give-up latch
existed specifically to hand off to an operator via `diagnoseStuckLogin`, but
because it never survived a poll tick, that diagnosis never persisted.

The fix reuses the boot-grace signal `CheckAndRestartCrashedAgents` already
established for the identical problem (a bare pane that hasn't painted yet is
not evidence of anything) rather than inventing a new heuristic, per the
issue's own suggestion and the contrast it draws with `pkg/watchdog`'s
`BootGrace`/`classify.go`.

The legitimate half of the old behavior is preserved: once boot grace has
genuinely elapsed and the pane still shows no login prompt, that IS positive
evidence of a clear (CLI came up healthy, or an operator fixed the token), and
the cap re-arms exactly as before.

No change was needed to the `(login token refreshed)` verdict string
(`RestartWithReason` call site) — it is unaffected by this fix.

## How tested

- `cd src && gofmt -l pkg/agent/manager.go pkg/agent/stuck_login_diagnosis_test.go` — clean.
- `cd src && go vet ./pkg/agent/...` — clean.
- `cd src && golangci-lint run ./pkg/agent/...` — `0 issues.`
- `cd src && go test ./pkg/agent/ -count=1 -race` — `ok  github.com/hivecommons/hive/pkg/agent  297.338s` (full suite, all tests pass).
- New regression tests added:
  - `TestTokenRestartLatchClearsWithinBootGraceStaysFalse` / `...AfterBootGraceRearms` —
    unit-test the new helper directly with injected times (no sleeps, no tmux).
  - `TestTokenRestartGiveUpSurvivesItsOwnBootPane` — drives the exact loop from
    the issue: login prompt → 3 token restarts → `tokenRestartGiveUp` latches →
    simulated blank boot pane on several poll ticks (within boot grace) → login
    prompt reappears → asserts **no 4th restart fires** and the attempt counter
    is unchanged; then asserts a genuine clear past boot grace DOES re-arm the
    cap.
  - Verified all three new tests **fail** against the prior unconditional-reset
    behavior (temporarily hardcoded `tokenRestartLatchClears` to always return
    `true` and reran — 2 of 3 failed with messages naming #6578, confirming
    they are real regression tests) before restoring the fix.

Fixes #6578

### DIFF
--- a/changelog.d/fixed-6578-token-restart-cap-latch.md
+++ b/changelog.d/fixed-6578-token-restart-cap-latch.md
@@ -1,0 +1,1 @@
+- Fixed the token-triggered restart give-up latch (GUARD 4, #4596) being cleared by the blank/booting pane produced by its own restart, which defeated the 3-attempt cap and caused restart storms of up to ×1334/24h on agents stuck at a device-flow login prompt ([#6578](https://github.com/hivecommons/hive/issues/6578)).
--- a/src/pkg/agent/manager.go
+++ b/src/pkg/agent/manager.go
@@ -3150,6 +3150,25 @@ func samePaneCapture(a, b []string) bool {
 	return true
 }
 
+// tokenRestartLatchClears reports whether a pane not showing a login prompt
+// is genuine evidence the prompt cleared, as opposed to the pane simply not
+// having painted yet. A CLI freshly relaunched by a token-triggered restart
+// shows neither a login prompt nor its ready chrome for a few seconds, and
+// that boot pane must not be read as "cleared" — doing so let the give-up
+// latch (GUARD 4, #4596) be reset by the side effect of its own remedy,
+// producing restart storms of ~1300/24h (#6578). This uses the SAME
+// boot-grace signal CheckAndRestartCrashedAgents already applies
+// (cliBootGraceSeconds measured from StartedAt) rather than inventing a new
+// one, so a still-booting pane and a genuinely cleared one are never
+// confused. Split out of pollTmuxOutputForAgent so the rule is unit-testable
+// without a tmux pane, mirroring decideTokenRestart.
+func tokenRestartLatchClears(agent *AgentProcess, now time.Time) bool {
+	if agent.StartedAt == nil {
+		return true
+	}
+	return now.Sub(*agent.StartedAt) >= cliBootGraceSeconds*time.Second
+}
+
 // pollTmuxOutputForAgent is pollTmuxOutput using the agent's tmux socket.
 func (m *Manager) pollTmuxOutputForAgent(agent *AgentProcess, ctx context.Context) {
 	const pollInterval = 3 * time.Second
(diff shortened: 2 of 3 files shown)


clubanderson:
Closing in favor of #6585, which was opened concurrently by another agent for the same issue (#6578). Switching to review that PR instead of maintaining a competing fix.

Issue #6488, opened 2026-09-10
🐛 bug: agent cards are not showing in the dashboard


balazsdukai:
## What happened?

The agent cards of the active agents are not appearing in the dashboards

## Expected behavior

The agent cards should appear in the for all active agent in the dashboard under the "Agents" section.

## Reproduction steps

1. Check the dashboard
2. I have completed steps 1,2,3,4,6,7 from the getting started guide

## Environment

- Branch/version: 378936cd2f4604df562742aeff1de0759555b8b0 (branch v4)
- Deployment method: Docker Compose 
- Operating system version: Ubuntu 24.04
- Relevant configuration (redact secrets): Using local hosted llm that are served from Unsloth via an OpenAI api. I configured the models directly in the `hive.yaml` config file, by configuring a `governor.gateway` and using that gateway's name in the agent definitions. 

Here is the `supervisor` agent definition:

```yaml
    supervisor:
        id: supervisor
        backend: unsloth-local
        model: unsloth/gemma-4-E4B-it-qat-GGUF
        beads_dir: /data/beads/supervisor
        enabled: true
        replicas: 1
        paused: false
        clear_on_kick: true
        cli_pinned: false
        model_owner: ""
        backend_owner: ""
        pause_owner: ""
        stale_timeout: 0
        restart_strategy: ""
        launch_cmd: ""
        display_name: Supervisor
        description: Orchestrates all agents — sweeps, enforces cadence, monitors health
        role: supervisor
        sort_order: 10
        emoji: "\U0001F451"
        color: '#e74c3c'
        aliases:
            - su
        lane_keywords: []
        detect_keywords:
            - supervisor
            - sweep
            - monitor
        kick_template: ""
        include_repos: true
        metrics_collector: ""
        bead_role: supervisor
        stats_display: []
        acmm_levels: []
        mode: ""
        on_demand: false
        caveman_mode: ""
```

Here is the `governor.gateway` definition:

```yaml
    gateways:
        - name: unsloth-local
          kind: custom
          endpoint: http://host.docker.internal:8888
          api_key_env: ""
          api_key_file: /secrets/unsloth_api_key
          default_model: unsloth/Qwen3-Coder-30B-A3B-Instruct-GGUF
          ca_bundle: ""
```

## Logs or screenshots

<img width="1050" height="668" alt="Image" src="https://github.com/user-attachments/assets/9351f3ce-bc43-4882-b29a-3e6b0765b4c0" />

[...]

PR #6594, opened 2026-09-10
🐛 fix(dashboard): enabled agents with no runtime status vanish from the Agents section

### LINKED ISSUE #6581: 🐛 bug: reopen #6488
## What happened?

The agent cards are still not showing up in the dashboard, however it #6488 have been closed.  

## Expected behavior

Same as #6488 

## Reproduction steps

Same as #6488 

## Environment

Same as #6488 , except I am on commit 917713cf840ae2084dcc12f82ee90c17507b74f3

## Logs or screenshots

Same as #6488

clubanderson:
## Stacks on #6589 — please merge that first

Base is `a6581` (#6589's branch), not `v4`, because both touch `status_builder.go`. Retarget to `v4` once #6589 lands.

#6589 is **diagnostics**: it reports *why* a card was omitted and explicitly changes no behaviour. This PR is the **fix**. They are complementary, not competing.

## Root cause

`buildAgents` built its name list **only from the runtime `statuses` map**:

```go
names := make([]string, 0, len(statuses))
for name, proc := range statuses { ... names = append(names, name) }
```

An agent that is `enabled: true` in config but has **no runtime process entry** therefore has no name, no card, and no trace anywhere in the UI. It does not render as broken — it does not render at all.

That is the reporter's symptom in #6488/#6581 exactly: enabled agents, empty Agents section, nothing to click, nothing to read.

## Fix

Two parts:

1. **`status_builder.go`** — after collecting runtime names, also walk `cfg.Agents` and include any agent that is enabled and passes the ACMM gate but has no runtime status. Those render through a new `buildMissingRuntimeAgent` as a card in a blocked/no-runtime state rather than vanishing. The existing ACMM and pack filters are applied identically, so this cannot resurrect an agent those gates intend to hide.
2. **`static/index.html`** — treat configured `governor.gateway` names as first-class inference methods, so a custom OpenAI-compatible gateway gets the same labels and model-discovery path as a built-in inference backend (`canonicalInferenceBackendName`). Without this, an agent whose `backend:` is a gateway name falls through `isInferenceBackendName`/`CLI_BACKENDS` and its model controls never populate.

## Failing test before the fix

```
agents len = 0, want 1
```

An enabled `supervisor` on a `kind: custom` gateway backend produced **zero** cards. Added to `status_builder_test.go`.

## Why this matters beyond the immediate bug

A silent drop is the worst possible failure mode for an adopter: there is nothing to search for. This is the same shape as #6489/#6515 (both this adopter) and the visibility gap #6558 describes. After this change an agent can be broken, but it cannot be invisible.

Worth noting for reviewers: #6488 was closed as COMPLETED with no commit referencing it and no explanation, which is why the reporter had to refile as #6581.

## Validation

- `go test ./pkg/dashboard -count=1` gives **ok 47.495s** on the stacked branch
- `go build ./...`, `go vet ./pkg/dashboard/` clean

## Honest caveat

#6589's author verified that `agent.NewManager(cfg.EnabledAgents(), …)` *does* instantiate `supervisor` into `AllStatuses()` in the production path, so on their reproduction the runtime entry existed. This fix closes the gap for the case where it does **not** — an agent that fails to launch, or is dropped before registration. That is consistent with the reporter's environment (self-hosted gateway, agents that may never start) but has not been confirmed against their hive. If #6589's diagnostics land alongside this, the `hiddenAgents` field will say definitively which path they are on.

Closes #6581

### DIFF
--- a/changelog.d/added-6581-agent-card-hidden-reason.md
+++ b/changelog.d/added-6581-agent-card-hidden-reason.md
@@ -1,0 +1,1 @@
+- `/api/status` and the lightweight agent-status stream now include `hiddenAgents`, listing any agent-manager runtime entry left out of the Agents cards along with the stable reason it was omitted for, so an empty or short Agents section is diagnosable without shell access ([#6581](https://github.com/hivecommons/hive/issues/6581)).
--- a/changelog.d/fixed-6581-gateway-agent-cards.md
+++ b/changelog.d/fixed-6581-gateway-agent-cards.md
@@ -1,0 +1,1 @@
+- Enabled agents whose backend is a configured model-gateway name now remain visible in the dashboard Agents section even if their runtime process is missing, with a blocked card that explains the missing process instead of silently disappearing ([#6581](https://github.com/hivecommons/hive/issues/6581)).
--- a/src/pkg/dashboard/agent_card_active_nonpack_test.go
+++ b/src/pkg/dashboard/agent_card_active_nonpack_test.go
@@ -71,3 +71,75 @@ func TestBuildAgentsIncludesActiveAgentOutsideCurrentPack(t *testing.T) {
 		t.Errorf("agent CLI = %q, want configured gateway name unsloth-local", got[0].CLI)
 	}
 }
+
+// TestBuildAgentsWithHidden_ReportsPackInactiveGhost is #6581's diagnostic
+// companion to the test above: it exercises buildAgentsWithHidden on the
+// EXACT same fixture (an active out-of-pack supervisor beside an inactive
+// pack-paused "ghost" scanner) and asserts the hidden-agent list is precise —
+// the active supervisor must never appear in it, and the suppressed ghost
+// must, tagged with the stable "pack-inactive" reason. Before this pairing
+// existed, an operator who still saw an empty Agents section after upgrading
+// past #6503 (as reported in #6581) had no way to tell, from /api/status
+// alone, whether the hive even attempted to surface a given agent, or why it
+// didn't.
+func TestBuildAgentsWithHidden_ReportsPackInactiveGhost(t *testing.T) {
+	level := 1
+	supervisorCfg := config.AgentConfig{
+		Backend: "unsloth-local",
+		Enabled: true,
+	}
+	ghostCfg := config.AgentConfig{
+		Backend: "unsloth-local",
+		Enabled: true,
+		Paused:  true,
+	}
+	cfg := &config.Config{
+		ACMMLevel: &level,
+		Agents: map[string]config.AgentConfig{
+			"supervisor": supervisorCfg,
+			"scanner":    ghostCfg,
+		},
+		Governor: config.GovernorConfig{
+			Gateways: []config.GatewayConfig{{
+				Name:     "unsloth-local",
+				Kind:     config.GatewayKindCustom,
+				Endpoint: "http://host.docker.internal:8888",
+			}},
+		},
+	}
+	statuses := map[string]*agent.AgentProcess{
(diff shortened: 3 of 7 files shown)


clubanderson:
Rebased onto the newly-rebased #6589 (`a6581` @ 8711608d, now itself rebased on v4 post-#6584). Also fixed a real bug found during review:

**Pack-ghost resurrection.** `buildMissingRuntimeAgent`'s new config-only pass reconsiders every `cfg.Agents` name not yet marked `seen`. `seen` was only set when a runtime name was added to the *visible* names list — not when the runtime loop hid it (below-ACMM-gate / pack-inactive). So an agent that IS present in `statuses` but hidden as pack-inactive (its runtime proc paused outside the active pack) could be re-evaluated by the config pass using a looser, different signal (`agentCfg.Paused` instead of the runtime proc's own paused/state). When the two disagreed — e.g. a runtime pause not yet persisted back to config — the ghost got re-added to the visible names, and since its real proc still exists in `statuses`, it rendered as a normal card while *simultaneously* still appearing in `HiddenAgents`. Reproduced with a probe test before fixing (0 agents → 1 agent shown, still tagged hidden).

Fix: mark every runtime-status name `seen` up front in the runtime loop itself (whether shown or hidden), so the config-only pass only ever fills the gap left by agents with **no** runtime entry — never re-litigates a decision the runtime gate already made. Added `TestBuildAgentsWithHidden_RuntimePausedGhostNotResurrectedByConfigPass` alongside the existing ghost-card tests; both existing ghost tests (`TestBuildAgentsIncludesActiveAgentOutsideCurrentPack`, `TestBuildAgentsWithHidden_ReportsPackInactiveGhost`) still pass unchanged.

**Reachability of the 'enabled, no runtime status' path**: traced it through `NewManager`/`ReconcileAgents` — a whole-config `ValidateBackend` failure hard-fails config load (won't produce this state), and an overlay-level failure drops the agent from `cfg.Agents` entirely (also won't produce it — matches old, correct silent-omission for invalid configs). The genuinely reachable window I found is in `cmd/hive/main.go`'s config-watcher `onChange`: `*cfg = *newCfg` swaps in the new (already-enabled) agent synchronously, but `agentMgr.ReconcileAgents(cfg.EnabledAgents())` — which actually creates the runtime entry — runs afterward, with `defsrc.ApplyToConfig` (a network fetch) and `ExpandAgentReplicas` in between. A concurrent `/api/status` request in that window sees the agent enabled in `cfg` with no corresponding `statuses` entry. Narrow, but real — consistent with the PR's own honest caveat that this hasn't been confirmed against the reporter's hive.

**index.html gateway-inference change**: reviewed safe — no new `innerHTML` of server-controlled strings (the pre-existing `backendLabel` unescaped-interpolation pattern is untouched by this PR), and `fetchConfiguredGateways()` degrades cleanly to `INFERENCE_BACKENDS`-only when `/api/config/governor/gateways` is absent/errors (try/catch leaves `_configuredGateways` at its prior value, defaulting to `[]`).

Validated: `gofmt -l`, `go vet ./pkg/dashboard/...`, `golangci-lint run ./pkg/dashboard/...` (0 issues), `go test ./pkg/dashboard/ -count=1 -race` (ok, ~99s). Base left at `a6581` per instructions — please retarget to v4 once #6589 merges.

clubanderson:
This PR was auto-closed by GitHub when its stacked base branch `a6581` (#6589) was deleted on merge. #6589 is now merged into `v4`; the same fix (rebased directly onto `v4`, plus the pack-ghost-resurrection fix from review) is now tracked at #6597.

PR #6602, opened 2026-09-10
🐛 fix(preflight): crt.sh identity query cannot see subdomains

clubanderson:
## What this is now

Rewritten after #6534 merged. That PR added Cert Spotter as a second source and believes the higher count, which fixed the immediate undercount. It left the crt.sh side still asking the narrower question.

`q=<domain>` is an **identity** match: certificates whose names are that exact domain, and **none for its subdomains**. That — not flakiness — is why crt.sh was measured seeing 46 of 83 in-window certificates. The cap is 50 new SAN sets per week per *registered* domain, and every subdomain spends from it.

This unions that query with crt.sh's subdomain wildcard `q=%.<domain>` (URL-encoded `%25.`) and counts the two halves as the one source they are.

## Why it still matters with two sources

Cert Spotter remains **required** and an unreachable Cert Spotter is still refused rather than trusted — this does not make crt.sh a substitute for corroboration.

It matters because Cert Spotter is rate-limited unauthenticated, and a truncated page parses exactly like a complete one. When it answers short, a crt.sh count about the right *set of names* can exceed it and block a window that would otherwise have looked open. With only the identity query it never can. Feeding a knowingly partial number into a `max()` can only ever lose issuances.

Losing the wildcard half on its own degrades to the identity-only reading the check already had, rather than turning a working check into a skip.

## Tests

Two new cases, plus a stub split so the identity and subdomain queries can be driven independently:

- subdomain issuances that crt.sh only sees via `q=%.<domain>` still block, against a *short but valid* Cert Spotter answer that on its own reports headroom
- a failed subdomain query still yields a reading

Verified non-vacuous: against v4's `preflight.sh` the two new assertions fail (`pass=35 fail=2`); with the change, `pass=37 fail=0`.

## Note on the earlier revision

The first version of this PR also reimplemented the two-source counting and SAN-set dedupe, which #6534 does better — it dedupes crt.sh on the certificate serial and Cert Spotter on the DNS-name set, and takes the max. That overlap is dropped; only the subdomain union remains. The `LE_CRT_SH_MIN_ROWS` floor is dropped too: with two sources cross-checking, a knob nobody sets was not carrying its weight.

Refs #5925


clubanderson:
Holding: this overlaps #6534 (maintainer's in-progress PR on the same preflight check 6 / LE headroom counting, #5925). @clubanderson — decide whether to fold the `%.<domain>` union into #6534 or land this separately; not merging until then.

clubanderson:
Rebased onto v4 now that #6534 has merged, and cut down to the one piece that does not overlap it: the crt.sh `q=%.<domain>` union. The two-source counting and serial dedupe are #6534's and are better than what I had here, so they are gone from this branch.

Deciding the hold as asked — landing separately rather than folding in, since #6534 is already merged.

/unhold

Issue #6588, opened 2026-09-10
RFC: add issue authors as co-authors of corresponding PRs/commits 


hanthor:
Hive is better off having core maintainers/itself writing it's own PRsd than outside contributors. We don't want to discourage outside contribution nor exclude people from feeling like that are contributing.

We should steer people to always write issues and then if the issue is accepted we can add them to the commit message so they still get credit and they still show up in the contributor list.

This way people can feel like they are contributing to the product more thant just opening issues

PR #6606, opened 2026-09-10
🐛 fix(scripts): issue-coauthor.sh credits nobody when a co-author already exists

### LINKED ISSUE #6605: [DCO] post-merge trailer failures on v4
## Post-merge DCO trailer check failed

- **Branch:** `v4`
- **Measured:** `8d5e85b1a0558008d1c5ea2fd08a1ee2149f97b4`
- **Run:** https://github.com/hivecommons/hive/actions/runs/34531551534
- **Commit window:** last 50 commits

The protected branch contains non-merge commits whose `Signed-off-by:` trailer is missing or whose email does not match the commit author email. This usually means a squash merge template dropped the trailer or retained a trailer for a different identity.

### Offending commits

```
DCO trailer check inspected 50 recent commits on HEAD (40 non-merge, non-bot checked; 7 merge skipped; 3 bot skipped; 1 waived).
WAIVED b798382a4d98ada28b0cb813644a595db0c58b08 missing-signoff author=Aditya Waghamare <168374658+adityawaghamare04@users.noreply.github.com>
FAIL 8fe6bb34626af4277394fcd87de2cd7eb797cc52 missing-signoff author=Andy Anderson <andy@clubanderson.com>
```

Close this issue after the offending protected-branch history has an accepted maintainer DCO disposition and the monitor is green again.

clubanderson:
## The bug

`src/scripts/issue-coauthor.sh` (added in #6592 for #6588) folds the issue author's trailer in with:

```sh
git interpret-trailers --if-exists doNothing --trailer "$trailer" "$file"
```

`doNothing` keys on the trailer **name**, not the whole line. So if the message already contains *any* `Co-authored-by:` trailer, the issue author's line is dropped.

Every commit in this repository carries `Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>` by convention. That means the condition is met on essentially every real commit, and the mechanism was a **silent no-op**: it exits 0 and prints the trailer it did not record, so the caller — and the PR — look correct while the filer is credited nowhere.

That is exactly the failure the script's own header sets out to prevent:

> a wrong trailer is worse than none — GitHub silently ignores an address it cannot resolve, so the credit looks recorded while nothing was credited.

Demonstrated directly, on a message shaped like a real one:

```
$ printf 'feat: thing\n\nBody.\n\nCo-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>\nSigned-off-by: Andy Anderson <andy@clubanderson.com>\n' > m.txt

$ git interpret-trailers --if-exists doNothing     --trailer "Co-authored-by: hanthor <...>" m.txt | grep -c hanthor
0        # ← the filer is not credited

$ git interpret-trailers --if-exists addIfDifferent --trailer "Co-authored-by: hanthor <...>" m.txt | grep -c hanthor
1
```

## The fix

`addIfDifferent` compares the **entire trailer line**, so a re-run with the same co-author is still a no-op — the idempotence the original comment wanted — while a *different* co-author is added. The comment is rewritten to say why the flag choice matters, so it does not get "simplified" back.

## Tests

Added a regression case in `src/scripts/test-issue-coauthor.sh` covering the shape every real commit is in — a pre-existing `Co-authored-by: Copilot` plus a `Signed-off-by:` — asserting that the issue author is credited, the existing co-author survives, and a second run stays idempotent.

Proven to catch the bug rather than merely pass: reverting the one-word change makes the new assertions fail.

```
### with the old doNothing:
  FAIL: an existing Co-authored-by suppressed the issue author's credit
  FAIL: the trailer block was split; DCO would fail
test-issue-coauthor FAILED

### with the fix:
test-issue-coauthor OK
```

## Relationship to #6605

#6605 fired because `8fe6bb34` on `v4` has its `Signed-off-by:` separated from a trailing `Co-authored-by:` by a blank line. That blank line splits the trailer block, and `git interpret-trailers --parse` then reports only the co-author:

```
$ git log -1 --format=%B 8fe6bb34 | git interpret-trailers --parse
Co-authored-by: hanthor <5840441+hanthor@users.noreply.github.com>
```

The sign-off is present in the text but outside the trailer block, so the post-merge monitor correctly fails it. The new test asserts the sign-off and both co-authors stay in **one** block, which stops this shape being produced again.

This PR cannot close #6605 on its own: that issue asks for a disposition on history already on the protected branch. The waiver list is `DCO_WAIVED_COMMITS` in `.github/workflows/dco-post-merge.yml`, and the App token this agent pushes with has no `workflows` permission, so the one-line addition of `8fe6bb34626af4277394fcd87de2cd7eb797cc52` to that list has to be made by a human. Details and the exact edit are in a comment on #6605.

Refs #6588
Fixes #6605


clubanderson:
Pushed the `DCO_WAIVED_COMMITS` entry for `8fe6bb34` onto this branch (maintainer disposition for #6605: sign-off is present and matches; only the blank line split the block). Retargeted the body to close #6605 on merge.

PR #6626, opened 2026-09-11
✨ feat: add OMP contributor backend

castrojo:
## Summary

- Register Oh My Pi (`omp`) as a Hive contributor backend and preserve Hive-owned interactive workspace, tmux, relay, assignment, and completion flow.
- Link Hive's verified `agent.md` export to OMP-discovered `$HOME/AGENTS.md` and `$HOME/CLAUDE.md`; launch ordinary `omp --model <id>` without an invented permission flag.
- Add OMP ready/busy/idle/login/onboarding/verdict fixtures, JS/Go fixture parity, an OMP interactive fake-hub smoke scenario, and the explicit interactive-only decision.

## Related issues

Fixes #6623

## Backend support tier

- **Pod path:** not T1. OMP remains absent from `HEADLESS_BACKENDS`; its one-shot command, exit semantics, and fresh-pod credential path are not claimed or wired.
- **Local path:** T3 experimental/refusal-gated. OMP has no verified filesystem sandbox or host-state deny-list. `just contribute-hive omp local` refuses unless `HIVE_OMP_DANGEROUSLY_RUN_UNCONFINED=1`; container mode remains the boundary.
- **Permission flags:** `backend_perm_flag omp` is deliberately empty. `omp --auto-approve` is an approval setting, not a confinement control, and this PR does not claim or pass it.
- **Credential detection:** `omp --version` only proves binary presence; because OMP has multiple provider credential mechanisms, the agent reports `UNVERIFIED` instead of falsely claiming authentication.
- **Metering:** unmetered; budget-gated use remains governed by the existing backend policy rather than a new scanner claim.

## Testing

- [x] `node bin/contributor-relay.test.js` — 350/350 passed.
- [x] `bash bin/contributor-agent.test.sh`
- [x] `go test ./pkg/agent -run 'TestPaneTail_MatchesSharedGoldenFixtures|TestPaneClassifiers_AgreeWithSharedGoldenFixtures'`
- [x] `go test ./pkg/config -run TestShellAndGoCLIBackendListsAgree`
- [x] `bash bin/test_backend_smoke.sh` — OMP interactive fake-hub stub emitted `HIVE_VERDICT: complete`; relay sent `task_complete` with `completion_signal=verdict`.
- [x] `git diff --check`
[...]

Issue #6623, opened 2026-09-11
feat(contributor): support OMP as interactive backend


castrojo:
Part of the downstream Hive contributor-runtime integration work.

Observed 2026-09-10 UTC against immutable v4 source `6608c45952e12c3c974cb821298b16e79a31f9b0` and downstream pin `c7a88b8518abf1163e13803b2094f2262605490b`:

- Neither [`config/backends.conf` at v4 head](https://github.com/hivecommons/hive/blob/6608c45952e12c3c974cb821298b16e79a31f9b0/config/backends.conf) nor the downstream pin contains `omp` in `KNOWN_BACKENDS`.
- The contributor launch, pane classifier, relay, and backend smoke paths have no OMP-specific handling: [`contributor-agent.sh`](https://github.com/hivecommons/hive/blob/6608c45952e12c3c974cb821298b16e79a31f9b0/bin/contributor-agent.sh), [`pane-classifier.js`](https://github.com/hivecommons/hive/blob/6608c45952e12c3c974cb821298b16e79a31f9b0/bin/lib/pane-classifier.js), [`contributor-relay.js`](https://github.com/hivecommons/hive/blob/6608c45952e12c3c974cb821298b16e79a31f9b0/bin/contributor-relay.js), and [`test_backend_smoke.sh`](https://github.com/hivecommons/hive/blob/6608c45952e12c3c974cb821298b16e79a31f9b0/bin/test_backend_smoke.sh).
- The fallback binary lookup would invoke `omp`, but it does not make the backend registered or define its readiness, pane, completion, or headless behavior.

Downstream will not add a protocol shim, local classifier copy, assignment selection, retry, or fallback. We need an upstream-supported interactive backend before publishing an OMP contributor image.

OMP documents interactive `omp`, explicit `--model <id-or-role>`, and automatic `AGENTS.md`/`CLAUDE.md` context discovery ([immutable docs](https://github.com/can1357/oh-my-pi/blob/3b3a6dc9bbd85102ce19d0b1c11bf6870915f6ec/docs/cli-reference.md)). A potential upstream route is to register `omp`, retain the agent's workspace/tmux/relay ownership, link the verified knowledge export to those context names, add captured OMP readiness/busy/idle/login/verdict fixtures with JS/Go parity, and make OMP explicitly interactive-only until one-shot behavior is proven by the existing smoke test.

Would maintainers accept that direction? If yes, I will provide the focused implementation with DCO sign-off and no downstream workaround.

[...]

Issue #6413, opened 2026-09-09
🐛 bug: agy readiness probe reads a blank 15-line pane tail — interactive agy contributors never become ready


Danathar:
## What happened?

`getCLIState()` never returns `ready` for the `agy` backend in interactive/tmux mode, so a healthy, idle agy pane is classified `starting` forever.

`waitForCLI()` therefore never resolves, `cliReady` stays false, the assigned task's prompt is queued instead of typed, and at `CLI_READY_TIMEOUT_MS` (600 s) the task is handed back with "CLI never became ready".

**Impact:** a freshly started interactive agy contributor cannot execute a single task. Recovery is not automatic — `armCLIReadyWait()` re-arms but deliberately withholds `ready` (`bin/contributor-relay.sh:1626-1637`), so the node then sits idle and accepts no further work.

### Root cause

`capturePaneText()` (`bin/contributor-relay.sh:1380`) returns the **full** pane — 50 rows, the geometry hive itself sets at `Justfile:1194`. agy renders *inline*: banner, input box and its `? for shortcuts` footer all land at rows 1–16, leaving rows 17–50 blank.

The agy branch of `getCLIState()` (`bin/contributor-relay.sh:1497-1509`) is the only backend that matches against a tail window rather than the full capture:

```js
const recent = paneTail(text, 15);
...
if (/\? for shortcuts/.test(recent)) return 'ready';
```

`paneTail(text, 15)` (`bin/contributor-relay.sh:2145`) slices the last 15 elements of `text.split('\n')` — rows 36–50, which are empty. `claude` (`:1453-1454`) and `codex` (`:1494`) test the full `text` and are unaffected.

The same blank window is used for agy's `needs-login` and `onboarding` classification, and by `blockingPromptKey()` (`:1409`), so the wizard auto-dismiss path has the same blind spot whenever the wizard text sits above row 36.

## Expected behavior

A live agy pane displaying `? for shortcuts` should be classified `ready`, and the queued task prompt should be delivered.

## Reproduction steps

1. `HIVE_AGY_DANGEROUSLY_RUN_UNCONFINED=1 just contribute-hive agy local` on a host with agy installed and signed in.
2. Wait for the hub to assign a task. The relay prints `CLI not ready — queuing task prompt instead of typing into the pane`.
3. `tmux attach -t hive-agy-<id>` — the pane is at a normal idle agy prompt with `? for shortcuts` drawn.
[...]

PR #6637, opened 2026-09-11
fix(contributor): key OMP readiness/busy/idle off real captured chrome

castrojo:
## Summary

`#6626` shipped the OMP contributor backend's readiness/busy/idle detection against a hand-written 6-line pane fixture, never a real `tmux capture-pane -p` at the dimensions `bin/contributor-agent.sh` actually launches with (`-x 200 -y 50`). Found while monitoring a live, hanging `projectbluefin/review` OMP contributor container (`AGENT_BACKEND=omp`, pinned at this repo's `HIVE_COMMIT`, which already includes `#6626`): the contributor sat at its welcome screen indefinitely, never picking up work.

Root-caused with real captures (2026-09-11, UTC times from `date`):

- **`classifyReadiness('omp')`** (`bin/lib/pane-classifier.js`): the only `ready` patterns match text inside the welcome splash's "Tips" panel (`# for prompt actions`, `/ for commands`). At real dimensions that panel renders above a tip-of-day banner, an "Update Available" banner, and the status footer, so it falls outside `paneTail`'s last-15-non-blank-line window — the same shape as `#6413`, one release later. Confirmed live: `classifyReadiness(realCapture, "omp")` returned `"starting"`.
- **Worse, and separately confirmed**: the splash renders only until the first reply. One live exchange in a disposable local OMP session (six short exchanges, `tmux -x 200 -y 50`) scrolled it off the pane permanently — zero occurrences of `Tips`/`# for prompt actions`/`/ for commands` anywhere in the capture — so readiness is never recovered for the rest of a task. This is the same one-shot-splash defect `#5156`/`#5650` fixed for claude, recurring here because the OMP fixtures never captured a second turn.
- **`classifyPane('omp')`'s busy marker** (`esc to interrupt` / `π >`): neither ever renders. Captured live: a tool call shows `⏺ Running… (esc to cancel)`; plain generation shows a spinner glyph plus `Working…`. A busy OMP pane was never classified `WORKING`.
- **`manager.go`'s `ompInputPromptMarker` (`"π >"`)** has the same problem (never observed live), and `paneShowsActiveWork` had no OMP-aware branch at all — `esc to interrupt` / `s · ↓` don't match OMP either.

## Fix

Both languages now key readiness/busy/idle off OMP's persistent status footer (model, cwd, cost, and a `─NN%─` context-usage meter — verified present on every captured frame: splash, idle, busy, and six-exchanges-deep-with-the-splash-long-gone alike) plus its real busy markers (`Working…` / `Running…`), the same persistent-chrome approach this file already uses for claude/codex/agy (see the inline comments on each of those for the same lesson learned earlier).

The synthetic `omp_ready`/`omp_busy`/`omp_idle` fixtures are replaced with real `tmux capture-pane -p` dumps — one from the actual hanging container this was diagnosed against, two from a disposable local OMP session — so the shared JS/Go golden-fixture parity tests `#6626` introduced now exercise real OMP output instead of fixtures authored to match the code they were meant to check.

`omp_login` and `omp_onboarding` are unchanged: no live counter-evidence against those two yet, so they are left alone rather than guessed at.

## Testing

- [x] `node bin/contributor-relay.test.js` — 350/350 passed
- [x] `bash bin/contributor-agent.test.sh`
- [x] `go test ./pkg/agent -run 'TestPaneTail_MatchesSharedGoldenFixtures|TestPaneClassifiers_AgreeWithSharedGoldenFixtures'`
- [x] `go test ./pkg/config -run TestShellAndGoCLIBackendListsAgree`
- [x] `bash bin/test_backend_smoke.sh` — 26 passed, 0 failed (unchanged; the OMP smoke scenario drives a scripted stub binary, not the real TUI, so it doesn't exercise this regression — the live-container and disposable-session captures above are the actual evidence)
- [x] `go build ./...` / `go vet ./pkg/agent/...`
- [x] `git diff --check`

## Contributor checklist

- [x] PR targets `v4`.
- [x] `changelog.d/fixed-omp-readiness-real-fixtures.md` added.

Refs #6623
Refs #6626


clubanderson:
Reviewed — thanks for chasing this down against a live container rather than fixtures. Verified locally: relay tests 350/350, `pkg/agent` pane/fixture/chrome tests (the omp golden subtests run and pass with the real captures, not skipped), `go vet`/`go build` clean, JS↔Go marker parity symmetric, no secrets in the captures, DCO OK. One observation, not a blocker: Go's `paneShowsActiveWork`/`isCLIChrome`/`cliPaneMarkers` are backend-agnostic and rely on marker specificity (`%─`, `╰─`, `Working…`) — consistent with how pi/codex/claude markers already work, and I grepped every other backend's fixtures for collisions (none). Merging.

PR #6639, opened 2026-09-11
fix(contributor): stop OMP's rotating startup tip from faking needs-login

castrojo:
## Summary

Found by monitoring, immediately after `#6637` merged: relaunched a real `projectbluefin/review` OMP contributor with the fixed pin, and it hit a *second* false readiness verdict on the very first assigned task. The pane was fully authenticated and idle, but `classifyReadiness('omp')` returned `needs-login`, so the relay queued the task and printed the "Oh My Pi needs authentication" banner instead of typing the task in.

Root cause: omp's welcome splash draws a rotating "Tip:" hint, and one rotation reads:

```
Tip: Log in to several accounts of the same provider — `/login` again — and omp
     load-balances across them automatically
```

`classifyReadiness('omp')`'s needs-login check is `/\b(?:sign in|log in|authentication required|credentials? required)\b/i`, which matches that tip's literal "Log in" text regardless of the pane's real authentication state. Confirmed live: `classifyReadiness(realCapture, "omp")` returned `"needs-login"` on an already-authenticated pane (captured 2026-09-11, real Hive OMP contributor container).

## Fix

Tips are cosmetic, splash-only hints — never real CLI state — so strip the "Tip:" paragraph (the line itself and its wrapped, unprefixed continuation line, up to the next blank line) before testing either the onboarding or login gate. Added a real captured fixture (`omp_tip_mentions_login`) reproducing the exact false positive from the live container.

## Testing

- [x] `node bin/contributor-relay.test.js` — 353/353 passed
- [x] `bash bin/contributor-agent.test.sh`
- [x] `go test ./pkg/agent -run 'TestPaneTail_MatchesSharedGoldenFixtures|TestPaneClassifiers_AgreeWithSharedGoldenFixtures'`
- [x] `go build ./...` / `go vet ./pkg/agent/...`
- [x] `git diff --check`

## Contributor checklist

- [x] PR targets `v4`.
- [x] `changelog.d/fixed-omp-tip-false-login.md` added.

Refs #6637


clubanderson:
Reviewed and reproduced locally: relay tests 353/353, Go golden-fixture parity green with the new `omp_tip_mentions_login` capture. Stripping the `Tip:` paragraph before the onboarding/login gates is the right scope — it's splash-only cosmetic text. Merging on green; thanks for the fast follow-up.

Issue #6662, opened 2026-09-11
🐛 bug(contributor): detectPRURL attributes a pre-existing PR to the task, suppressing a correct no_work_needed verdict and defeating #3987


Danathar:
## Summary

`detectPRURL()` (`bin/contributor-relay.js:1834-1853`) is a bare regex over the agent's tmux pane. It cannot tell a PR the agent **opened** from a PR the agent merely **read about**. When it matches a pre-existing PR, two things go wrong at once:

1. The relay reports that PR to the hub as this task's `pr_url` — crediting the contributor with someone else's work.
2. The `prURL ? null : verdict` precedence at line 2991 **discards a legitimate `no_work_needed` verdict**.

(2) is the serious half, because it defeats #3987 in precisely the population #3987 was built for.

## Observed

hive `v4` @ `ff3833fb`, image `ghcr.io/hivecommons/hive-contributor:latest`, podman 6.1.1 rootless, backend `agy` (Antigravity CLI 1.1.22), spoke `hosted-projectbluefin-knuckle-gjvq.hive.hivecommons.dev` (`served_sha` `55bd2bc`, protocol 1.2).

Two consecutive tasks in one container, same agent, same verdict — **different outcomes**:

```
Task ct-projectbluefin/bluefin-965-1789126671 completed — signal=verdict (HIVE_VERDICT: no_work_needed)
Detected no_work_needed verdict for ct-...-965-...: merged PRs (#1057, #1081, #1090, #1135, #1149)
  cover all actionable items in bluefin, and item 3 is tracked in dakota#622      <-- verdict SURVIVED

Task ct-projectbluefin/bluefin-879-1789126915 completed — signal=verdict (HIVE_VERDICT: no_work_needed)
Detected PR for ct-...-879-...: https://github.com/projectbluefin/bluefin/pull/1103
                                                                                  <-- verdict SUPPRESSED
```

`projectbluefin/bluefin#1103` is **not this contributor's PR**:

```
#1103 MERGED  author=mrbobbytables  created=2026-08-13  merged=2026-08-24
branch=fix/879-amd-framework-kargs-cleanup
title=fix(framework): remove stale hid_sensor_hub karg from AMD Framework laptops
```

[...]

PR #6693, opened 2026-09-11
🐛 relay: scan deep scrollback for the PR URL, not the 15-line payload

### LINKED ISSUE #6667: 🐛 bug(contributor): a PR the agent really opened is never reported — detectPRURL reads 15 pane lines, ~10 of which are TUI chrome (inverse of #6662)
## Summary

`detectPRURL()` reads only the last **15** lines of the agent's tmux pane (`TMUX_TAIL_LINES = 15`, `bin/contributor-relay.js:197`). For a full-screen TUI backend like `agy`, roughly 10 of those 15 lines are static chrome, leaving ~4 lines of real content. A PR the agent genuinely opened scrolls out of that window within a few tool calls, so `pr_url` goes to the hub **empty** and the hub books the task as "did not ship".

This is the mirror image of #6662 (which reports the false *positive*: a researched PR URL that happens to be inside the window gets attributed to the task). Both failures have one root cause — PR detection is a regex over a 15-line slice of terminal rendering.

## Observed, 2026-09-11

hive `v4` @ `ff3833fb`, image `ghcr.io/hivecommons/hive-contributor:latest`, podman 6.1.1 rootless, backend `agy` (Antigravity CLI 1.1.22), spoke `hosted-projectbluefin-knuckle-gjvq.hive.hivecommons.dev`, contributor `c-b0a6b3785b4a` (`Danathar`).

Relay log for the task — note there is **no `Detected PR` line**:

```
Task prompt sent to CLI
Task ct-projectbluefin/fsdk-containers-124-1789128392 completed — signal=verdict (HIVE_VERDICT: complete)
Relaunching agy after a task exit: agy --dangerously-skip-permissions
Task assigned: issue projectbluefin/fsdk-containers#130 — Per-runtime-family bloat-stripping recipes…
```

But the agent **did** open a PR for it:

```
$ gh pr list --repo projectbluefin/fsdk-containers --author Danathar --state all
2026-09-11T12:15:57Z OPEN #259 docs(research): audit AI/ML and CNCF-adjacent targets with PyTorch measurements
```

Confirmed in the container's own checkout:

```
$ git -C /home/dev/workspace/projectbluefin/fsdk-containers log --oneline -1
66a3652 docs(research): audit AI/ML and CNCF-adjacent targets with PyTorch measurements
$ git -C … branch --show-current
docs/124-catalog-additions-audit
```

So a real PR, on a task-specific branch, authored by this contributor, minutes before completion — and `pr_url` was empty.

## Why it was missed

The URL **is** in the pane. It is just not in the last 15 lines. Captured with `tmux capture-pane -p -S -2000`:

```
line 915 of 1015:  ● Bash(gh pr create \)
line 920 of 1015:  I have completed the investigation and opened Pull Request #259
                   https://github.com/projectbluefin/fsdk-containers/pull/259 against main
                   to address issue #124 …
```

Line 920 of 1015 — about **95 lines from the end** by the time completion was detected, because the agent kept working (verdict line, further tool calls, TUI redraw) after announcing the PR.

And the 15-line window is mostly not content. Here is what `captureTmuxLines(15)` actually yields against a live `agy` pane:

```
▸ Thought for 3s, 345 tokens
  Evaluating bloat reduction strategies
● Bash(gh pr list --repo projectbluefin/fsdk-containers --state all --search "130")
● Bash(gh pr view 136 --repo projectbluefin/fsdk-containers)
● Bash(gh pr list --repo projectbluefin/fsdk-containers --state all --limit 20) (ctrl+o to
⣾ Working...
└ Tip: Press ? to see keyboard shortcuts.
────────────────────────────────────────────────────────────────────────────────────────
>


────────────────────────────────────────────────────────────────────────────────────────
esc to cancel
```

Two separators, the input box, two blanks, the tip line, the spinner and the status line consume ~10 of 15 rows. Roughly 4 rows carry content, and `agy` additionally collapses tool output behind `(ctrl+o to expand)`, so even those rows are frequently truncated summaries rather than the command output that held the URL.

Note this sample is also a live instance of the #6662 hazard: the only PR references currently inside the window are `--search "130"` and `gh pr view 136` — PRs being *researched*, not opened.

## Impact

Per the comment at lines 2982-2986, an empty `pr_url` is the signal for "just went idle" and the hub then applies the **short** cooldown rather than the with-PR cooldown. So `fsdk-containers#124` is booked as unshipped and becomes eligible for re-offer shortly, while PR **#259** sits open awaiting review. The next contributor to draw it starts from scratch on an issue that already has a pending PR.

I could not observe the re-offer directly — #124 is out of the queue while its cooldown holds, and the short and long cooldowns are indistinguishable from outside until expiry. The re-offer consequence above is read from the code and its comment, not measured.

Also note the completion itself was correct: the agent emitted `HIVE_VERDICT: complete` and the relay honoured it. Only the PR reporting failed. So this is purely a reporting/attribution defect, not a completion-detection one.

## Suggested fix

1. **Stop deriving shipped-PR state from terminal rendering.** Ask git/gh instead, at completion time: `gh pr list --repo <repo> --author @me --state open --json url,headRefName,createdAt` filtered to `createdAt >= taskAssignedAt`, or resolve the PR for the branch the agent pushed. That is authoritative, immune to TUI layout, and fixes #6662's false positive in the same stroke.
2. If the pane must be used, **do not use a 15-line window for this.** Completion detection wants a small recent slice; PR detection wants the whole task transcript. They should not share `TMUX_TAIL_LINES`. Scan the pane from `taskAssignedAt` onward (`capture-pane -S -` for full scrollback), with authorship/recency filtering per #6662.
3. **Prefer capturing the `gh pr create` output directly** rather than scraping prose about it — that is the one string guaranteed to be the PR this task opened.

Found while troubleshooting a containerized `agy` contributor; happy to test a patch against this spoke.

clubanderson:
## Problem

`detectPRURL` was handed the **same fifteen lines** the relay sends upstream as `tmux_output`:

```js
const tmuxLines = captureTmuxLines(TMUX_TAIL_LINES); // 15
...
const prURL = detectPRURL(tmuxLines, currentTask.repo);
```

`TMUX_TAIL_LINES` is sized for the *audit trail* the hub stores, not for searching. In a real TUI roughly ten of those fifteen rows are chrome — the input box, the status bar, the hint line — leaving a handful of actual output rows. A PR URL the agent genuinely printed leaves that window within seconds of being printed.

The consequence: the relay reports `pr_url: ''` for a task that **shipped a PR**, and the hub applies the short "just went idle" cooldown to work that was actually complete (kubestellar/hive#2393 item 7 is the reason the field exists at all).

This is the inverse reading of #6662. There, a scraped URL was trusted too much and a third party's PR was attributed to the task. Here, a real one is never seen. #6680 explicitly names this issue and deliberately **avoids making it worse** — it keeps reporting the URL on an unverified `gh` result rather than dropping it — but it does not widen the window, so the defect stands.

## Change

A new named constant `PR_SCAN_LINES = 400` is the window for PR detection, on both paths:

- **Interactive:** capture the deep window **once** and slice its tail for the protocol payload. `tmux_output` is byte-identical to before, and no second `capture-pane` call is added — the pane read is destructive to the `paneStalled()` fingerprint (#5333), so a second read would corrupt stall detection.
- **Headless:** `stdout`/`stderr` are already fully buffered (capped by `HEADLESS_MAX_OUTPUT_BYTES`), so this is a second slice of the same redacted string. There is no chrome here, but a build log pushes a URL past fifteen lines just as easily.

Redaction is unchanged and still happens before anything is sliced.

## Tests

`#6667 a PR scrolled out of the 15-line payload window is still reported` drives the full completion path through the existing harness: the URL is printed, 60 lines of build noise push it out of the tail, the verdict lands at the bottom. It asserts three things — the PR is reported, `tmux_output` stays within `TMUX_TAIL_LINES` (the deeper scan must not start shipping the whole pane upstream), and, as a setup check, that the URL really is outside the tail so the test cannot silently stop proving anything.

Verified it catches the bug: restoring the single line `detectPRURL(tmuxLines, ...)` fails it — `363/364 passed`, `FAIL #6667 a PR scrolled out of the 15-line payload window is still reported`. With the fix, `364/364 passed`.

## Merge order

Touches `bin/contributor-relay.js` alongside open PR #6680 (#6662). The overlap is two call-site lines; whichever lands second needs a trivial rebase. The changes are complementary — #6680 verifies a candidate URL is *this task's*, this one makes sure the candidate is *seen* — and neither subsumes the other.

Closes #6667

### DIFF
--- a/bin/contributor-relay.js
+++ b/bin/contributor-relay.js
@@ -203,6 +203,14 @@ const HEADLESS_MAX_OUTPUT_BYTES = 1048576; // 1 MiB
 
 const TMUX_TAIL_LINES = 15;
 const NEEDS_LOGIN_CONFIRM_TICKS = 3;
+// #6667: the window detectPRURL scans is NOT the 15-line protocol payload.
+// TMUX_TAIL_LINES is sized for the audit trail the hub stores, and of those 15
+// terminal rows roughly ten are TUI chrome — the input box, the status bar, the
+// hint line — so only a handful of real output rows survive. A PR URL the agent
+// genuinely printed scrolls out of that window within seconds of being printed,
+// and the relay then reports no PR for a task that shipped one. Scan deep
+// scrollback for the URL while still sending only the tail upstream.
+const PR_SCAN_LINES = 400;
 const HEARTBEAT_INTERVAL_MS = 30000;
 const HEARTBEAT_TIMEOUT_MS = 90000;
 const RELAY_TEST_TIMING = process.env.HIVE_RELAY_TEST_TIMING === '1';
@@ -1168,8 +1176,11 @@ function runHeadlessTask(task) {
   }, (err, stdout, stderr) => {
     headlessChild = null;
     // Tokens can appear in agent output; redact before the tail leaves the host.
-    const outTail = redactTokens(String(stdout || '') + String(stderr || ''))
-      .split('\n').slice(-TMUX_TAIL_LINES);
+    const outLines = redactTokens(String(stdout || '') + String(stderr || '')).split('\n');
+    const outTail = outLines.slice(-TMUX_TAIL_LINES);
+    // #6667: headless has no TUI chrome, but a build log easily pushes a PR URL
+    // past fifteen lines, so scan the same deep window the interactive path does.
+    const outScan = outLines.slice(-PR_SCAN_LINES);
     // A revoke clears currentTask before killing the child. Ignore any callback
     // that arrives afterwards — including a raced exit 0 — so stale work cannot
     // emit completion after its assignment generation was fenced out.
@@ -1205,7 +1216,7 @@ function runHeadlessTask(task) {
     finish(() => {
       setPiInvocationState('succeeded');
       console.log(`Headless task ${task.task_id} completed (exit 0)`);
-      const prFinding = resolveTaskPR(outTail, {
+      const prFinding = resolveTaskPR(outScan, {
         repo: task.repo,
         taskId: task.task_id,
         taskStartedAt: taskAssignedAt,
@@ -3153,7 +3164,13 @@ function progressTick() {
   }
 
   const paneState = checkTmuxPaneState();
-  const tmuxLines = captureTmuxLines(TMUX_TAIL_LINES);
+  // One capture, two windows (#6667). Capturing the deep scrollback and slicing
+  // its tail keeps the protocol payload byte-identical to before while giving
+  // PR detection room to see a URL that has scrolled past the visible rows. It
+  // deliberately does NOT add a second capture-pane call: the pane read is
+  // destructive to the paneStalled() fingerprint (#5333).
+  const paneScanLines = captureTmuxLines(PR_SCAN_LINES);
+  const tmuxLines = paneScanLines.slice(-TMUX_TAIL_LINES);
 
   // #5321: forward progress renews the max-duration lease. Recorded here,
   // before any branch below can return, so EVERY pane state gets the credit —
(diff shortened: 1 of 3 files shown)


clubanderson:
Rebased onto v4 over #6680: `resolveTaskPR` (the attribution engine) replaced `detectPRURL` at both call sites, so the deep window now feeds attribution — `resolveTaskPR(outScan, …)` headless and `resolveTaskPR(paneScanLines, …)` interactive. All 373 relay tests pass locally.

PR #6697, opened 2026-09-11
📖 Record OpenSSF Best Practices Badge in security self-assessment

clubanderson:
Closes #6684

## What

The security self-assessment was wrong about the OpenSSF Best Practices Badge in **two successive revisions, in opposite directions**:

1. The original said the project "has not pursued" the badge and had "not yet assessed whether this is planned" — an evasion, which TAG-Security review ([cncf/toc#2286](https://github.com/cncf/toc/pull/2286)) called out.
2. The correction that landed on `v4` overshot to "**not yet held**, and now being pursued," pointing at this issue as a tracked commitment.

Both are wrong. **The badge was already awarded at the passing level on 2026-08-27** — four days *before* that review — and `README.md` has been displaying it the entire time.

Verified against the programme's public API (`https://www.bestpractices.dev/projects/14261.json`):

| field | value |
|---|---|
| project | **14261** (`hive`) |
| `badge_level` | **passing** |
| `badge_percentage_0` | **100** |
| `achieved_passing_at` | **2026-08-27T17:47:12Z** |
| `lost_passing_at` | `null` — never lapsed |
| silver / gold | 15% / 22% — **not** claimed |

The section now states the badge is held, cites ID/level/date, notes it is self-certified (so the value to a reviewer is the per-criterion justifications, not the graphic), and **records both prior errors rather than quietly overwriting them** — because the root cause is worth stating: the document was being written from itself instead of re-derived from observable state, the same defect that previously left the maintainer count stale.

## Also done (outside this repo)

The badge entry itself was stale after the move to the vendor-neutral `hivecommons` org. **67 criteria justifications / 70 URLs** still cited `github.com/kubestellar/hive` and `hive.kubestellar.io`.

That matters more than it sounds: the per-criterion justifications *are* the evidentiary value of a self-certified badge, so every link a reviewer might follow to check a claim pointed at the old home. All 70 now resolve to the current repository and site; `homepage_url` and `repo_url` updated; entry re-verifies at `passing` / 100% with **zero** fields containing the old org.

Only URLs were rewritten — the distinct `kubestellar` tokens were enumerated first to confirm none were prose references to the org, which would have been wrong to touch.

A correction was also posted to the TAG-Security reviewer, since the reply on that PR had told him the badge was still pending: [cncf/toc#2286 (comment)](https://github.com/cncf/toc/pull/2286#discussion_r3989892459).

## Not done

`README.md` already carries the badge; no change needed.

[...]

Issue #6684, opened 2026-09-11
🌱 Register for the OpenSSF Best Practices Badge (passing level)


clubanderson:
Raised in CNCF TAG-Security review of the Hive security self-assessment ([cncf/toc#2286](https://github.com/cncf/toc/pull/2286)). The reviewer's words:

> You should do this. At least get a passing badge please. It's not that hard and the things you need to do for this are sane security things you either are or should be doing already.

That is correct, and the honest reason we had not done it is that nobody had considered it — the self-assessment recorded it as "not yet assessed whether this is planned," which was an evasion rather than an answer.

## Task

Register `hivecommons/hive` at [bestpractices.dev](https://www.bestpractices.dev/) and complete the **passing** level criteria.

## Why this should be cheap

Most passing criteria are already satisfied:

- [x] OSI-approved license (Apache-2.0, `LICENSE`)
- [x] Public version-controlled source with interim versions
- [x] Documented contribution process (`CONTRIBUTING.md`, `GOVERNANCE.md`)
- [x] Private vulnerability reporting (GitHub Security Advisories, `SECURITY.md`)
- [x] Documented security response process (`src/docs/security-response.md`)
- [x] Automated test suite gating every PR
- [x] Static analysis in CI (`golangci-lint`, OpenSSF Scorecard)
- [x] Dependencies pinned by digest/SHA
- [x] No known unpatched vulnerabilities

Outstanding is registration and self-certification, plus verifying the crypto and release-signing criteria.

## Done when

- Badge registered and passing level achieved
- Badge ID and level recorded in `src/docs/security-self-assessment.md` (replacing the reference to this issue)
- Badge added to `README.md`

clubanderson:
Closed — with a correction to the premise.

**The badge was never actually missing.** Hive has held the OpenSSF Best Practices badge at the **passing** level since **2026-08-27**, and `README.md` has displayed it the whole time:

| field | value |
|---|---|
| project | **[14261](https://www.bestpractices.dev/projects/14261)** |
| `badge_level` | **passing** |
| `badge_percentage_0` | **100** |
| `achieved_passing_at` | **2026-08-27T17:47:12Z** |
| `lost_passing_at` | `null` — never lapsed |
| silver / gold | 15% / 22% — not claimed |

This issue was filed in response to TAG-Security review asking for a passing badge ([cncf/toc#2286](https://github.com/cncf/toc/pull/2286)), four days *after* the badge had already been awarded.

## What was actually wrong

[...]

PR #6698, opened 2026-09-11
docs(security): the compliance list still says the badge is not held

Danathar:
> **Re-scoped after review.** This PR originally overlapped #6697. It has been narrowed to the gap #6697 leaves and rebased; it is now `MERGEABLE` and no longer conflicts with #6697. See [the response below](#issuecomment) for the reasoning.

## The remaining problem

`src/docs/security-self-assessment.md`, in `## Project compliance`, currently says:

> - **OpenSSF Best Practices Badge**: **not yet held — and being pursued.**

That is false. The badge has been held at `passing` since **2026-08-27** — project [14261](https://www.bestpractices.dev/projects/14261), 100% of the passing criteria (all 67 Met or Not-Applicable). Re-verified live against the API while preparing this:

```
level: passing | passing%: 100 | achieved: 2026-08-27T17:47:12Z
```

**#6697 does not touch this section.** It rewrites `### Open SSF best practices` further down. So if #6697 merges alone, the document asserts in one section that the badge is held and in another that it is not — in the artifact whose audience is CNCF TAG-Security.

## What this PR is now

One bullet, in the one section #6697 leaves alone. The overlapping section rewrite is **dropped in favour of #6697**, which is more complete there (it also corrected 67 stale `kubestellar` links inside the badge entry itself) and carries the changelog fragment.

Two details are carried over from the original version because they are corrections in their own right:

- The entry is labelled a maintainer **self**-certification, so it does not read as contradicting the *"Hive does not currently hold any formal security certification"* sentence it sits directly under.
- No level beyond `passing` is claimed. The API reports a tiered percentage above 100 (partial silver), but that number moves on its own — and this document already argues, two bullets up, that the Scorecard score is not frozen here for exactly that reason. Same rule, applied to the badge.

## Verification

`python3 src/scripts/check-docs-links.py src/docs` — 149 files, all relative links and anchors resolve.

Docs-only under `src/docs/**`, which `changelog-fragment-guard.yml` exempts explicitly; #6697 carries the fragment for this issue.

Refs #6684

### DIFF
--- a/src/docs/security-self-assessment.md
+++ b/src/docs/security-self-assessment.md
@@ -402,18 +402,20 @@ project-level compliance signals:
   (`scorecard.yml`, via a SHA-pinned reusable workflow in
   `hivecommons/infra`); results publish to the repository's code-scanning
   alerts. No specific score floor is gated in CI at present.
-- **OpenSSF Best Practices Badge**: **not yet held — and being pursued.** At
-  first review of this assessment the project had not registered, and
-  TAG-Security's feedback was that a passing badge is achievable
-  low-hanging fruit whose criteria are things the project should be doing
-  anyway. The maintainers agree; registration and criteria completion are
-  tracked in [#6684](https://github.com/hivecommons/hive/issues/6684), and
-  this document will carry the badge ID and level once awarded rather than a
-  promise. The substantive prerequisites — OSI license, public VCS, private
-  vulnerability reporting, documented contribution process, automated test
-  suite in CI, static analysis, and no known unpatched vulnerabilities — are
-  already satisfied; what is missing is the registration and self-certification
-  itself.
+- **OpenSSF Best Practices Badge**: **held at the `passing` level** — project
+  [14261](https://www.bestpractices.dev/projects/14261), achieved 2026-08-27,
+  100% of the passing criteria (all 67 Met or Not-Applicable). This is a
+  maintainer **self**-certification against published criteria, publicly
+  auditable per-criterion — not a third-party audit, and so not a
+  counter-example to the "no formal security certification" sentence above.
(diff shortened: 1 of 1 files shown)


Danathar:
Thanks — both claims checked out, and acting on them turned up a third thing.

**Duplicate: confirmed, and resolved in #6697's favour.** #6697 was opened three minutes before this one, by the issue's own author, and is more complete on the overlapping part — it also corrects 67 stale `kubestellar` links inside the badge entry itself, and carries the changelog fragment. I've dropped my version of that section rewrite rather than compete with it.

**Conflicts: confirmed, now fixed.** Both were `CONFLICTING`. This branch is rebased onto current `v4` and is now `MERGEABLE`.

**But "merge one, close the other" would leave the document self-contradictory.** #6697 rewrites `### Open SSF best practices`; it does **not** touch `## Project compliance`. Since this PR was first written, `v4` gained a badge bullet *there* which says:

> - **OpenSSF Best Practices Badge**: **not yet held — and being pursued.**

The badge has been held at `passing` since 2026-08-27 (project 14261, 100% of passing criteria — re-verified live against the bestpractices.dev API just now: `level: passing | passing%: 100 | achieved: 2026-08-27T17:47:12Z`). So if #6697 merges alone, the self-assessment states in one section that the badge is held and in another that it is not — in the artifact written for CNCF TAG-Security.

So rather than closing this, I've **narrowed it to exactly that gap**: one bullet, in the one section #6697 leaves alone. The two PRs no longer touch the same lines, no longer conflict with each other, and both can merge in either order.

If maintainers would still rather have a single PR, I'm happy to close this and hand the bullet to #6697 as a suggestion instead — just say which you prefer.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `891b816c`

PR #6702, opened 2026-09-11
🌱 Publish ioscan red-team evaluation

### LINKED ISSUE #6685: 🌱 Red-team ioscan: measure prompt-injection detection instead of asserting it
Raised in CNCF TAG-Security review of the Hive security self-assessment ([cncf/toc#2286](https://github.com/cncf/toc/pull/2286)), on the `ioscan` section:

> have you red teamed to know how well this works? What should a potential user be aware of here? Is this perfect defense, works decently well, a partial mitigation, etc.?

## The honest current answer

**No red-teaming has been done.** A search of the repository for red-team, adversarial-evaluation or penetration-test artifacts returns nothing relevant. `src/pkg/ioscan` has 56 unit tests across its rule, Unicode, classifier and canary paths, but unit tests establish that *known* shapes are caught — they say nothing about unknown ones, and they cannot produce a detection rate.

This matters because `ioscan` is the control most likely to be *read* as a prompt-injection defense, when the thing actually containing the consequence of a successful injection is the proxy's hard-deny of `POST /pulls` and `PUT /pulls/{n}/merge` for every ACMM mode.

## Task

Run and publish an adversarial evaluation of the injection path:

1. Build a corpus of prompt-injection attempts against the kick path — at minimum: direct instruction override, Unicode steganography (invisible controls, tag characters, variation selectors), homoglyph substitution, base64 and multi-layer encoding, instruction smuggling via code blocks/diffs/labels/author fields, and multi-turn/split-payload attempts across issue + comments.
2. Measure what the deterministic rules catch, what the optional classifier adds, and what gets through.
3. For anything that gets through, determine whether the network-layer deny rules still contain it — that distinction is the real finding.
4. Publish the methodology and results, including the miss rate.

## Done when

- A red-team evaluation document exists in `src/docs/`
- `security-self-assessment.md` cites measured results instead of describing efficacy as unmeasured
- Any rule gaps found are filed as their own issues

clubanderson:
Closes #6685

## Summary
- add `src/docs/ioscan-red-team-evaluation.md` with a bounded adversarial corpus and measured deterministic ioscan results
- update the security self-assessment to cite measured results rather than saying ioscan efficacy is unmeasured
- document classifier and network-deny scope separately so scanner misses are not overstated as write-control bypasses

## Verification
- Documentation/configuration only; mutation-proof test requirement is not applicable because this PR publishes evaluation results rather than changing runtime behavior.

### DIFF
--- a/src/docs/ioscan-red-team-evaluation.md
+++ b/src/docs/ioscan-red-team-evaluation.md
@@ -1,0 +1,75 @@
+# ioscan adversarial evaluation (2026-09-11)
+
+This document records the first bounded red-team evaluation of Hive's `ioscan`
+untrusted-input path. It measures the deterministic scanner as implemented on
+`v4`; the optional LLM classifier is documented separately because it depends on
+operator-provided model configuration and is disabled by default.
+
+## Methodology
+
+The corpus focuses on prompt text that can reach an agent kick through issues,
+PR bodies, comments, labels, authorship fields, or concatenated task context. A
+case is counted as **caught** when `ioscan.ScanInput` returns at least one
+finding, and as **blocked** when the deterministic policy returns
+`Verdict.Blocked`. Tests were run with:
+
+```sh
+cd src
+GOMODCACHE=/Users/andan02/hive-copilot-work/gomodcache GOFLAGS=-modcacherw go test ./pkg/ioscan
+```
+
+The optional classifier was not called against a live model in this evaluation:
+no repository-default endpoint/API key exists, the classifier is off by default,
+and a live model score is not reproducible enough to use as the source of truth
+for the deterministic baseline. Its expected value is therefore stated as a
+qualitative operator control: it can add semantic coverage for plain-English
+manipulation that avoids the deterministic phrases, but it is not required for
+Hive's hard network denies.
+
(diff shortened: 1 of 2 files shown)


clubanderson:
Closing this in favour of @Danathar's #6709, which is a superset.

Both PRs answer #6685, but they answer it differently, and the difference matters:

- **This PR (#6702)** documents an evaluation — `ioscan-red-team-evaluation.md` plus the self-assessment update. The measurement lives in prose.
- **#6709** ships the corpus as an executable test (`src/pkg/ioscan/redteam_test.go`, +592/-17) *and* updates the same self-assessment rows.

#6685 asked us to **measure** detection instead of asserting it. A number written into a markdown file is still an assertion — it is true on the day it is written and silently rots afterwards. #6709's corpus re-runs in CI, so the published miss rate stays honest, and it already paid for itself: it is what found #6708 (U+FF29 and U+00AD producing *no finding at all*), which I have now fixed in #6714.

No work is lost — #6709 covers the self-assessment changes this PR made.

Thanks @Danathar. Sorry for the duplicated effort; I should have checked for an existing PR before starting this one.

Issue #6904, opened 2026-09-14
🐛 Spoke dashboard gives an auto-upgrade hive no manual upgrade action — owners click the "Queued for auto-upgrade" pill and nothing happens


Danathar:
## Reported

Two hosted-hive owners in Discord, 2026-09-11:

- **jOrge** — "I have it set to auto update. I clicked the button it doesn't seem to upgrade."
- **andy** — "I updated bluefin using hub dashboard from my phone. Bluefin is set to upgrade once a week on Tuesdays. I think there might be a bug in the spoke self upgrade function. But the hub upgrade worked. Hive owners can use either."

So: the **hub** dashboard's upgrade path worked; the **spoke** dashboard's did not. Both hives have
auto-upgrade enabled, at least one of them on the weekly (Tuesday) schedule.

I have no spoke logs or hive IDs from either reporter — everything below is read from the code at
`4fbb06c` (fresh clone today), and it finds two distinct ways the spoke Upgrade control is inert for
an auto-upgrade hive. Both produce exactly the reported symptom.

## 1. With auto-upgrade on, the spoke renders a passive pill and no action at all

`pkg/dashboard/static/index.html:16662`:

```js
if (v.autoUpgrade && !um) {
  // passive badge: <span ...>Queued for auto-upgrade</span>   (no data-action, no handler)
} else {
  // <button id="spoke-upgrade-btn" data-action="gh27" ...>↑ Upgrade available → <sha></button>
}
```

The rationale is stated in `pkg/dashboard/api.go:655-660`:

> `autoUpgrade` tells the dashboard whether the hub manages this spoke's upgrades. When true the
> manual spoke Upgrade button is hidden — the hub rolls the upgrade out automatically, so offering a
> manual button is redundant and confusing.

That is true for **instant** mode only. For `daily` the wait is up to ~24h (1pm ET), and for
`weekly` up to ~7 days (Tuesday 1pm ET — `pkg/hub/upgrade_schedule.go:35-52`). During that window
the owner has no upgrade action on the spoke dashboard at all, which is why andy had to reach for
the hub from his phone.

The hub deliberately does the opposite. `pkg/hub/saas.go:17299-17340` renders the badge **and** an
"Upgrade now" action for exactly this case, and says why:

> The old single "queued" pill WAS the click-to-upgrade affordance, and owners could not find it
> (real feedback: "that line that says 'queued' is the upgrade button — it's not straightforward").
[...]

Issue #6902, opened 2026-09-14
✨ feature(contribute): expose contributor-queue withheld items and admission reasons in Operations


Danathar:
## Problem

An operator can look at an open, apparently actionable (often `help wanted`) issue on GitHub and reasonably ask:

> "Why isn't this issue in the Contributor Queue?"

Today the Operations tab cannot answer that. The Contributor Queue deliberately does **not** contain every open issue — a candidate can be withheld for many legitimate reasons (operator hold, `blocked` workflow label, an open PR already claiming it, unresolved/unknown dependencies, work already in flight, completion or failure cooldown, assigned to someone else, title/author/label filters, tracker/umbrella issue, a live `no_work_needed` verdict, disabled repository, …). But almost every one of those decisions is a bare `continue` in the admission loop: the reason is computed and then thrown away. The only thing the operator sees is that the row is absent.

The count gap is already visible but unexplained: since #6461, `GET /api/contribute/status` reports both `candidate_items` (raw scanner population) and `actionable_items` (offerable). In #6537 the maintainer summary was that "the 116 vs 14 gap is exactly 'candidates that admission rejects'" — this issue is about showing *which* candidates and *why*.

## Desired behaviour

The Operations Contributor Queue should distinguish roughly three groups:

```
Contributor Queue

READY
  #1234  Some issue
  #1235  Another issue

ON HOLD
  #1236  Some issue
         Operator hold — "waiting on design review"

WITHHELD  (collapsed by default — "N withheld")
  #1237  Some issue
         Existing PR #4567
  #1238  Some tracker
         Tracker / umbrella issue
  #1239  Some issue
         Dependency #4321 is still open
  #1240  Some issue
         Assigned to another contributor
  #1241  Some issue
         Failure cooldown (until 14:32)
  #1242  Some issue
         Workflow label `blocked`
  #1243  Some issue
         Previous no_work_needed verdict
  #1244  Some issue
         Rejected by contributor title/author/label filter
```

- **Ready** is exactly today's offerable queue — unchanged.
- **On hold** is today's manually held rows (amber "on hold" pill, Resume control) — unchanged, and kept visually distinct from automatic withholding because it is the operator's own decision.
- **Withheld** is new: every candidate Hive knows about (it is in `ActionableIssues`) but is not currently willing to offer, with a human-readable reason and the evidence behind it when one exists (PR link, blocker issue, cooldown expiry, assignee, matched filter).

[...]

PR #6919, opened 2026-09-14
🌱 sync: merge v4 (38b582c1) into v5

clubanderson:
Forward-port of v4 `38b582c1` (v4.28.4 → v4.30.0, 104 commits) into v5.

See the merge commit message for the conflict resolutions. Highlights:
- `main.go` / `manager.go`: v5's split-file layout kept; v4 hunks (#6846 persist seam, #6500/#6767 copilot token propagation) ported into `main_helpers.go`, `manager_tokens.go`, `provider_error.go`, `manager_env.go`.
- Restored the #5958 `clearStartFailureLocked` call in `RelaunchBobAgentsAwaitingKey` that v5's earlier split dropped (caught by v4's new test).
- `contribute_sse.go`: #6902 Withheld collector supersedes #4246 `withDiagnostics`.
- Dropped `pkg/hub/terminal_assertion_alias_test.go` — v5 removed the hub alias layer in 93232b083; the same lanes are pinned in `pkg/hub/spoke` and `pkg/terminalassert`.
- `version_provenance_ui_test.go` updated to #6904's contract (Upgrade now button stays on auto-upgrade spokes).

Validation: `go build ./...`, `go vet ./...`, `go test ./... -count=1` all green locally.


clubanderson:
Merging with admin despite one red check.

The single remaining failure is `Sign-off survives the squash`, which is a **confirmed structural false positive on sync pull requests** — filed as #6923 and fixed on `v4` by #6925 (merged).

The gate enumerates `base..head`. Here the base is `v5`, so that range is all 93 v4 commits `v5` has not seen yet, and the five carrying legitimate bot sign-offs get re-flagged. Both premises the gate documents for itself fail on a sync:

- *"the branch is still writable at PR time"* — these commits are permanent history on `v4`; the `--amend -s && push --force-with-lease` remediation it prints is impossible.
- *"squash merge writes a NEW commit with the PR author"* — syncs are **merge-committed**, not squashed (#6751 `783ec871` and #6718 `2e2ebf7b` both have two parents), so the mismatched commit it guards against is never synthesised.

`d8f376e8` appearing in the flagged list is the tell: it is the very commit #6798 was written about, already landed and already waived.

Reproduced locally against this PR's real base/head:

| | inspected | result |
|---|---|---|
| no resolvable landed ref (old behaviour) | 93 | 5 FAIL, rc=1 — matches this PR's CI exactly |
| `LANDED_REFS=origin/v4` (the fix) | 0 (93 skipped) | pass, rc=0 |

This PR cannot pick the fix up itself: its head merged `v4` at `38b582c1`, before #6925 landed, and its base is `v5`, so the fix is absent from its merge ref. **The next sync PR will run the fixed gate and go green on its own.**

The two `test` failures that were also red were the known runner egress outage (#6648/#6870); both passed on re-run.

I did attempt to top this branch up with current `v4` so it would carry the fix, but `src/cmd/hive/main.go` conflicts as a single hunk spanning ~8,500 lines — `v5` has refactored it into `wireSpokeSubsystems` while `v4` keeps it inline. Hand-resolving that risks silently reverting the `v5` refactor, so I left it for the sync automation / a maintainer rather than forcing it.

Issue #6909, opened 2026-09-14
🐛 bug: contributor-protocol generation fencing is opt-in — omitting task_gen bypasses the #2568 guard


clubanderson:
## Summary

Contributor-protocol **generation fencing is opt-in by the client**. Any client can bypass the #2568 stale-generation guard on every call site simply by omitting `task_gen` from its message.

## The defect

`generationAccepted` treats a zero client generation as universally valid:

```go
func generationAccepted(clientGen, currentGen uint64) bool {
	if clientGen == 0 {
		return true
	}
	return clientGen == currentGen
}
```
<sub>`src/pkg/dashboard/contribute_ws.go:2975-2980`</sub>

`msg.TaskGen` is a plain `uint64` decoded from JSON, so an **omitted** `task_gen` field yields `0` — which is exactly the value that disables the check. The guard therefore protects only clients that volunteer to be protected.

The codebase already knows this. The comment at `contribute_ws.go:4201` says so in as many words, while documenting a past incident this same weakness contributed to:

> *"The #2568 Gate does not cover this: `generationAccepted()` returns true whenever `clientGen == 0`, and omitting `task_gen` yields 0, so the guard is opt-in from the client."*

That incident: a completion naming any other task released the assignment without revoking the lease or booking a cooldown, so *"the contributor went `ready` and was minted a SECOND live repo credential under `max_concurrent=1`"*. The `if hasTask` guard fixed that specific path — but the underlying fencing bypass it names is still live.

## Blast radius

All three fencing call sites are bypassable the same way:

- `contribute_ws.go:4104`
- `contribute_ws.go:4174`
- `contribute_ws.go:4401`

## Why this matters now

#6899 asks Hive to expose a contributor lifecycle seam to third-party agent hosts (OMP workbenches). The [DESIGN-RESPONSE](https://github.com/hivecommons/hive/issues/6899#issuecomment-5665373503) correctly identifies generation fencing as an invariant any such seam must enforce **structurally** rather than by convention. Today it is enforced by convention only. Widening the contract to external embedders while fencing remains opt-in would hand that weakness to a larger and less trusted client population.

## Proposed fix — close the downgrade path without breaking legacy clients

The `clientGen == 0` escape exists for genuine backward compatibility, so it cannot simply be deleted. But it should not be available to a client that has already demonstrated it speaks the fenced protocol.

[...]

PR #6931, opened 2026-09-14
✨ protect contributor quota headroom before accepting work

### LINKED ISSUE #6833: [v5] [Tracker] Feature: protect contributor quota headroom before accepting work
## Use case

A contributor should be able to start a Hive contribution session and leave it running without worrying that it will consume the last part of a provider's short-term or weekly subscription allowance.

Codex, Claude Code, and Agy expose rolling short-term (commonly five-hour) and weekly usage windows. Some accounts can continue through paid credits or extra usage after the subscription allowance is exhausted. Today Hive can react after a provider refuses work, but that is too late: a task may already have been assigned, credentials released, and useful context consumed. The risk is greatest when Hive starts a Medium, Complex, or Unknown task near the edge of a window.

The contributor needs a local, configurable reserve. For example, setting 20% should cause Hive to stop accepting new work when either the short-term or weekly window reaches 20% remaining, display the reason and reset time in the terminal where the session was launched, and ask whether to continue. If nobody answers, the safe behavior is to wait. A contributor who intentionally permits paid-credit use must make that choice explicitly, and may opt out of the guard when starting the session.

## Proposed solution

Add a contributor-side quota guard that normalizes provider-reported usage windows and runs an admission check before each task is accepted.

The first implementation should support subscription-authenticated Codex, Claude Code, and Agy. Provider adapters should emit a local normalized reading with:

- provider and local quota-pool identity;
- observation time and source;
- state: `available`, `guarded`, `unknown`, or `stale`;
- zero or more windows containing a stable window ID, duration, used/remaining percentage, and reset time;
- whether the provider reports that paid credits or extra usage are available, without enabling them.

This should reuse the normalized capacity concepts in #5698 instead of hard-coding provider-specific facts into the relay. Subscription limits are not API RPM/TPM limits and should remain a separate capacity source.

### Contributor configuration

Expose documented environment/config settings with these semantics (names may change to fit v5 conventions):

```dotenv
# Enabled by default for supported subscription backends.
# ask: pause at the threshold and display interactive/noninteractive controls
# pause: pause without an interactive question
# off: do not enforce the quota reserve for this session
HIVE_CONTRIBUTOR_QUOTA_GUARD=ask

# Default reserve for every recognized window. Valid range: 0..100.
HIVE_CONTRIBUTOR_QUOTA_MIN_REMAINING_PCT=20

# Optional per-window overrides; inherit the default when unset.
HIVE_CONTRIBUTOR_QUOTA_SHORT_MIN_REMAINING_PCT=20
HIVE_CONTRIBUTOR_QUOTA_WEEKLY_MIN_REMAINING_PCT=20

# Optional task-complexity reserves; inherit the effective window reserve when unset.
# This lets a contributor keep more headroom before accepting harder work.
HIVE_CONTRIBUTOR_QUOTA_COMPLEX_MIN_REMAINING_PCT=35
HIVE_CONTRIBUTOR_QUOTA_UNKNOWN_MIN_REMAINING_PCT=35
```

Equivalent settings should be available through the normal contributor configuration path. Invalid percentages must fail startup with an actionable error. `off` is the explicit launch-time opt-out requested here.

For a task and window, the required reserve is the maximum of the effective window reserve and any configured reserve for the task's v5 complexity tier. With the example above, Simple and Medium tasks use 20%, while Complex and Unknown tasks require 35%. Complexity settings are optional; by default every tier inherits the user's base percentage.

### Admission and refresh behavior

1. Discover the backend, authentication/account class, and local quota-pool identity at startup.
2. Obtain a fresh reading without sending a model prompt. For a supported subscription backend, do not advertise readiness until the first valid reading is available.
3. Refresh after every provider response/task completion, before accepting an assignment, when a provider status update arrives, and around a reported reset time.
4. Gate when **any** recognized window has `remaining_pct <= required_reserve_pct`. The equality rule makes the boundary deterministic.
5. Do not interrupt an active task solely because it crosses the threshold. Warn immediately, allow the current task to reach its normal completion point, and stop admitting subsequent work.
6. Resume automatically after a fresh reading shows every window above its effective reserve, unless the contributor explicitly chose to remain paused.
7. Store readings atomically and key them by the local credential/quota pool, not by relay process. Concurrent relays using the same provider account must see one shared guard state.
8. Treat an unknown or stale reading from a supported adapter as a safe hold after a bounded retry period, with a clear terminal message. Unsupported backends should report that the guard is unavailable and retain their existing behavior.

The reserve is a safety buffer, not a guarantee that a task will complete before a provider limit. Complexity-specific reserves give contributors a deterministic way to be more conservative without relying on guessed token costs.

### Terminal behavior and explicit overrides

When the guard activates, write a prominent message to the terminal/log surface associated with the session launcher. It should include:

- provider and account/plan class when available;
- each window's remaining percentage and reset time;
- the configured reserve that caused the hold;
- the pending task's title and complexity when a task was declined at preflight;
- a clear statement that no new work will start while paused;
- a clear statement that continuing may consume paid credits if the provider already has them enabled.

Offer controls equivalent to:

- **Pause until reset** — safe default, including when there is no TTY or no response.
- **Continue once** — admit one assignment, then evaluate the guard again.
- **Continue until this window resets** — override only the exact guarded window ID/reset epoch.
- **Disable for this contributor session** — the runtime equivalent of the launch-time opt-out.

The relay often runs detached, so correctness must not depend on reading stdin from its background process. The launcher may present an interactive question when attached, but it must also print a reliable command/control path, for example `just contribute-quota continue-once`, that reaches the running session through a local control socket or atomic state file.

Overrides must be narrowly scoped. A one-task override expires after one assignment. An until-reset override expires when that exact window/reset epoch changes. A session opt-out expires when the session exits. None of these choices should silently become a persistent authorization to spend money.

### Paid credits and provider settings

Hive must never purchase credits, enable extra usage, change a billing preference, or mutate a provider's host configuration automatically.

The continue choices only authorize Hive to admit work despite the contributor's configured reserve. The provider's own paid-credit or extra-usage setting remains authoritative. The prompt must say when continuing may use paid credits, and should link to the provider's own control when one is available.

### Provider adapters

Use supported machine-readable provider surfaces where possible and capability-detect CLI versions rather than parsing decorative terminal text:

| Backend | Initial integration |
| --- | --- |
| Codex | Read the same account's limits through the Codex app-server `account/rateLimits/read` method, including all returned windows and reset times. Do not use API-key RPM/TPM data as subscription usage. |
| Claude Code | Capture the documented status-line JSON `rate_limits.five_hour` and `rate_limits.seven_day` fields. Chain an existing user status-line command and preserve its output. Use a non-token-consuming usage refresh/bootstrap path when the fields have not appeared yet. |
| Agy | Capture the documented status-line `quota` map (`remaining_fraction`, reset fields, and plan tier), preserving an existing status line. Use its usage/quota refresh path when needed. |

Adapters must not overwrite a contributor's existing status-line configuration. Local and container launch modes should use an ephemeral overlay/wrapper where necessary. If a CLI version lacks the required interface or returns an unrecognized schema, report that explicitly and enter the configured unknown-data behavior.

Existing estimates such as `bin/token-usage.py` may remain informational fallback telemetry, but hard-coded token budgets must not become the enforcement source because provider plans, models, and accounting rules change.

### v5 assignment protocol

The clean admission point is after Hive sends task metadata but before it releases a GitHub credential or starts the agent:

1. The relay advertises a versioned `quota_preflight_v1` capability.
2. The hub sends `task_assign` with the v5 complexity tier and withholds the GitHub token.
3. The relay refreshes/evaluates local quota state.
4. On success it sends `task_accepted`; the hub then releases the token.
5. On failure it sends a new non-failure response such as `task_declined` with a coarse reason (`local_capacity_guard` or `local_headroom_mismatch`).

A quota decline means the task never started. It must not increment task or contributor failure counts, affect trust/reputation, or apply the normal failure cooldown. The hub should release the assignment immediately so another contributor can take it.

When the base window reserve is reached, the relay enters a local hold and withholds `ready`. When only a complexity-specific reserve rejects the offered task, the hub should temporarily skip that task for this contributor and may offer a lower-complexity task. This avoids a repeated assign/decline loop while still using available capacity.

Capability negotiation must preserve current behavior between mixed-version hubs and relays.

### Privacy and policy ownership

Detailed usage percentages, reset times, quota-pool identifiers, and credit balances stay on the contributor machine. The hub needs only the negotiated capability, a coarse capacity/decline reason, and any temporary per-task skip needed for scheduling.

The contributor owns this policy. A hub must not lower the contributor's reserve or authorize an override. Logs must not expose provider credentials, raw account identifiers, or billing details.

## Acceptance criteria

- [ ] A contributor can configure an arbitrary 0–100 remaining percentage globally and optional short-term, weekly, and task-complexity overrides.
- [ ] Guard enforcement is enabled by default for supported subscription backends, with a documented launch-time opt-out.
- [ ] Codex, Claude Code, and Agy adapters normalize their documented short-term/weekly quota data without consuming a model turn.
- [ ] Exact-threshold, multiple-window, stale/unknown, reset, and invalid-configuration behavior is covered by tests.
- [ ] The guard never accepts a new task or receives its GitHub credential while any applicable window is at or below the effective reserve.
- [ ] An active task is allowed to finish normally; subsequent work is held.
- [ ] Simple/Medium/Complex/Unknown v5 task metadata participates in the effective reserve, with inheritance covered by tests.
- [ ] The launcher shows remaining quota, reset timing, the triggering reserve, and scoped continue/pause controls; headless sessions pause safely.
- [ ] One-task, until-reset, and session overrides expire at their documented boundaries.
- [ ] No code path purchases credits, enables paid usage, or changes provider billing/settings.
- [ ] A quota preflight decline releases the task without failure, trust, or normal cooldown penalties.
- [ ] Two relays using the same local quota pool cannot independently oversubscribe the configured reserve.
- [ ] Existing status-line commands/configuration remain intact in local and container modes.
- [ ] Mixed-version hub/relay behavior remains backward compatible through capability negotiation.
- [ ] Contributor documentation explains that the reserve reduces risk but cannot guarantee a task will finish within a window.

## Alternatives considered

- **React only after a provider quota error.** This already helps Hive recover after exhaustion, but the assignment and model context may already have been spent.
- **Use fixed token estimates.** Provider subscription accounting varies by model, plan, cache behavior, and product policy; hard-coded estimates drift and cannot reliably represent the provider's own window.
- **Ask the hub to enforce contributor percentages.** That would disclose private account/billing state and let a remote policy influence local spending. The contributor should enforce the guard before acceptance.
- **Kill work as soon as a threshold is crossed.** Interrupting an active task is likely to waste the context the reserve is intended to protect. The main enforcement point should be between tasks.
- **Hold an assigned task while waiting for input.** This blocks other contributors. Decline/release it immediately and keep the local relay paused instead.

## Scope and compatibility

- **Target branch/version:** v5. This builds on v5 task classification and the task-metadata-before-credential admission point. It complements #6825; quota pacing remains a separate contributor policy.
- **Operator impact:** Add capability-negotiated preflight and non-failure decline semantics. Operators do not receive private quota or billing data.
- **Contributor/agent impact:** Supported subscription users gain a default safety reserve, visible pause state, and explicit scoped overrides. API-key/pay-as-you-go spend budgets are a separate feature.
- **Migration/backcompat:** Old hubs and relays retain current behavior. New behavior activates only after capability negotiation. Unsupported backends continue with an explicit guard-unavailable status.
- **Initial scope boundary:** Codex, Claude Code, and Agy. Other backends can add adapters against the same normalized interface.
- **Prediction boundary:** This issue does not require a task-cost predictor. v5 complexity-specific percentages provide deterministic headroom control; measured cost prediction can be added later.

### Maintainer note: possible v4 backport

Please consider whether a smaller safety improvement should be backported to v4 after the v5 design lands. A v4 backport could implement provider readings, the configurable reserve, terminal warnings, and between-task suppression of `ready`. The full task-aware flow and clean pre-credential `task_declined` semantics belong in v5, so a backport should not block or weaken the v5 design.

## Additional context

Relevant existing work:

- #6541 handles provider quota exhaustion and reset after the provider has already refused work.
- #5698 defines normalized provider-capacity readings that this local guard can reuse.
- #6825 covers v5 capability-aware assignment and explicitly leaves quota pacing as separate work.
- Codex exposes account rate limits through its documented app-server API: https://learn.chatgpt.com/docs/app-server
- Claude Code exposes five-hour and seven-day percentage/reset fields to status-line integrations: https://code.claude.com/docs/en/statusline
- Agy exposes remaining quota and reset data to status-line integrations: https://antigravity.google/docs/cli/statusline/

clubanderson:
Refs #6833.

This PR targets `v5`. GitHub only honors `Fixes #N` closing keywords for PRs into the default branch (`v4`), so this v5-based PR will not auto-link/auto-close #6833; I will cross-reference it from the issue instead.

## Summary

- Adds a contributor-local quota guard evaluator with `ask`/`pause`/`off`, validated `HIVE_CONTRIBUTOR_QUOTA_*` percentages, per-window reserves, and v5 complexity reserve maxing.
- Extends `rotation.Headroom` with RFC #5698-style normalized `limits` windows and populates them from existing Claude/Codex/Agy probes.
- Adds relay preflight behavior that withholds `ready` while guarded, declines via `task_declined` only to hubs advertising `quota_preflight_v1`, and keeps the hub from delivering credentials to quota-preflight-capable relays until `task_accepted`.
- Documents contributor-facing env vars and adds a changelog fragment.

## RFC #5698 reuse

I reused RFC #5698's limit-window model by adding `rotation.LimitWindow` and using that same shape in `pkg/contributorquota.Reading.Limits` and in the relay's normalized JSON reading evaluation. Relevant RFC text from `docs/rfc-5698-backend-capacity-model-inventory-placement.md`:

> Represent provider headroom as a list of limit windows, not as one scalar.

and its capacity reading example defines:

> `limits:`
> `  - kind: session`
> `  - kind: weekly`
> `  - kind: weekly_scoped`

This PR extends the existing `/api/providers/headroom` `Headroom` payload with `limits` instead of introducing a parallel quota schema.

## Deliberately descoped / follow-up

- Full machine-readable provider adapter bootstrapping for Claude Code status-line JSON, Agy status-line quota maps, and Codex app-server multi-window refresh is not complete here. Existing rotation probes now publish normalized windows, and the relay has a clean normalized reading seam (`HIVE_CONTRIBUTOR_QUOTA_READING_JSON` / `HIVE_CONTRIBUTOR_QUOTA_READING_FILE`) for provider readers to feed without hard-coding provider-specific facts into admission.
- Scoped interactive controls (`continue once`, `until reset`, control socket/state file) are represented in the Go guard state but not wired to launcher commands yet.
- Cross-process atomic quota-pool storage for concurrent relays is not implemented yet; readings can be supplied through the normalized file seam as a follow-up.

These paths are explicit rather than silent fail-open: the preflight protocol only uses `task_declined` when the hub advertises `quota_preflight_v1`, and guarded readings with `unknown`/`stale` state wait safely.

## Break-it proof

Temporarily changed the core gate from `<=` to `<` and ran the boundary test. It failed as expected:

```text
--- FAIL: TestEvaluateBoundaryRefusesAtEquality (0.00s)
    --- FAIL: TestEvaluateBoundaryRefusesAtEquality/equality_refuses (0.00s)
        guard_test.go:33: Admit=true, want false (decision={Admit:true Wait:false Reason: WindowID: Required:0 Remaining:0})
FAIL
FAIL	github.com/hivecommons/hive/pkg/contributorquota	0.531s
FAIL
```

The gate was then restored to `<=`.

## Test output

Focused/full touched Go packages:

```text
$ cd /Users/andan02/hive-6833/src && export GOMODCACHE=/Users/andan02/hive-6833/.gomodcache GOFLAGS=-modcacherw && go test ./pkg/contributorquota ./pkg/rotation ./pkg/dashboard
ok  	github.com/hivecommons/hive/pkg/contributorquota	(cached)
ok  	github.com/hivecommons/hive/pkg/rotation	(cached)
ok  	github.com/hivecommons/hive/pkg/dashboard	50.292s
```

Relay suite:

```text
$ cd /Users/andan02/hive-6833 && node bin/contributor-relay.test.js
...
ok   contributor quota preflight declines at exact reserve and admits one unit above
ok   contributor quota task_assign sends task_declined before acceptance
ok   contributor quota off fully disables task_assign guard
ok   contributor quota guarded relay withholds ready while idle

421/421 passed
```

`gofmt -l` on touched Go files produced no output.


clubanderson:
Reviewed this and pushed one fix directly to the branch.

**`Evaluate` failed open on an unrecognized limit window.** It skipped any window whose kind was not in `recognizedWindow`, so an exhausted window the build does not know about admitted work, while an identical recognized window refused it:

```
exhausted UNRECOGNIZED ("daily") -> {Admit:true}
exhausted RECOGNIZED   ("weekly") -> {Admit:false Wait:true Reason:guarded WindowID:weekly}
```

That is not hypothetical for this feature: `weekly_scoped` was itself a late addition to Claude's reporting, which this PR has to special-case. The next new kind would silently disable the guard at exactly the moment a real limit is exhausted — the same shape as the #6909 fencing bypass, where an allowlist degraded to permissive on unseen input.

Now every window is evaluated. `EffectiveWindowReserve` already falls back to `BaseReservePct` for an unknown kind, so the rule is well defined without inventing semantics. Refusals from an unknown kind report `guarded_unknown_window` so an operator can distinguish "your quota is low" from "the provider reported a window we have not taught the guard about", and a healthy unrecognized window still admits, so this cannot cause spurious pauses.

Break-it proof: restoring the `continue` makes `TestUnrecognizedWindowKindStillGuards` fail with `admitted work: {Admit:true}`; restoring the fix passes.

Also removed a dead branch in `ClaudeProber.Probe` that re-set `Kind` to `weekly_scoped` after `normalizeLimitKind` had already returned it unchanged.

**Verified independently** (not taking the original report on trust):

- `rotation.LimitWindow` genuinely reuses RFC #5698's model — multiple windows with `kind`, not collapsed to one scalar — and extends `Headroom` additively, so existing consumers still compile.
- Re-ran the equality break-it proof myself: reverting `<=` to `<` fails 3 tests, not the 1 claimed.
- `go test ./pkg/contributorquota ./pkg/rotation` → both ok.

Issue #6825, opened 2026-09-12
[RFC] [v5] Capability-aware contributor task assignment


Danathar:
> **Hold - design discussion, not a task.** This is a proposal for maintainer discussion. Please do not implement it, decompose it into sub-issues, or relay it to contributors until the design is settled.

**Status:** Proposal  
**Target:** Hive v5  
**Scope:** Contributor relay, task selection, model capability metadata, contributor portal  
**Initially out of scope:** Autonomous provider purchasing, benchmark-based blocking, changes to contributor trust, and automatic performance-based tier adjustments  
**Related design:** [RFC #5698: Backend capacity, model inventory, pacing, and placement](https://github.com/hivecommons/hive/blob/v5/docs/rfc-5698-backend-capacity-model-inventory-placement.md)

## Summary

Extend v5's contributor system so Hive can recommend—and, with contributor opt-in, select—an appropriate model configuration for each task from the configurations that contributor has offered.

The design separates three questions:

1. **Task capability requirement:** What level of engineering capability does this task appear to need?
2. **Model capability estimate:** What evidence supports using this model, harness, and reasoning configuration for that work?
3. **Availability and preference:** Can the contributor run that configuration now, and under what limits or preferences have they offered it?

Hive supplies task requirements and adopts the capability policy. The contributor supplies the configurations they are willing to run. The relay remains responsible for resolving and launching the selected local configuration.

For example, a contributor might offer:

| Local configuration | Backend | Capability estimate | Contributor preference |
| --- | --- | --- | --- |
| `everyday-agy` | Agy | T2 | First choice for routine work |
| `everyday-codex` | Codex | T2 | Alternative for routine work |
| `deep-codex` | Codex | T1 | Reserved for T1 tasks |

These are illustrative assignments, not benchmark claims about particular products. Each configuration represents a specific model and reasoning setting. Hive could recommend either routine configuration for a contained bug fix and the reserved configuration for a difficult compatibility migration.

The expected benefits are fewer avoidable task/model mismatches and better use of contributor resources. These are hypotheses to validate: this RFC does not establish that model capability is the cause of any particular failed contribution, or that a higher tier guarantees a successful PR.

The recommended first release is advisory. Automatic selection is a later opt-in extension. Benchmark-based blocking is outside the initial scope.

## Existing v5 foundation

This proposal extends existing mechanisms; capability routing is not entirely new to v5.

[...]

PR #6949, opened 2026-09-14
🐛 Fix Operations personal panels

### LINKED ISSUE #6937: Operations tab shows nothing to signed-out viewers — no sign-in prompt like the Profile tab has
## Summary

On `/contribute`, the **Operations** tab has identity-dependent panels that simply disappear when the viewer is not signed in, with nothing said about why. The **Profile** tab already handles the same condition correctly, with a `.me-signin` prompt. Operations should do the same.

Verified against `origin/v4` @ `701c76fd`.

## What a signed-out viewer sees

Open `/contribute`, click **Operations**, while not signed in. The "Your contribution" card is not rendered at all — no placeholder, no prompt, no error. The page looks complete. There is nothing to suggest that signing in would reveal anything, and nothing on the page offers a way to sign in.

The page chrome (`api_contribute.go:2011-2020`) is tabs plus a theme toggle. There is no auth chip, no sign-in button, and no anonymous-viewer banner anywhere on `/contribute`. The Profile tab is currently the only surface on the whole page that tells a visitor they are anonymous.

## Why it happens

`#cc-mine-card` is declared hidden (`api_contribute.go:2641`):

```html
<div class="ops-card" id="cc-mine-card" style="display:none;margin-bottom:20px">
```

`ccLoadMine()` (`api_contribute.go:6395`) only ever reveals it on success, and deliberately swallows the anonymous case:

```js
fetch('/api/contribute/me').then(function(r){
  // 401 (anonymous) and 403 (no profile on this hive) are the NORMAL answers
  // for a visitor, not failures: the card simply stays hidden. ...
  if(!r.ok)return null;
  return r.json();
}).then(function(d){
  if(!d||!d.github_username)return;
  ...
```

Treating 401 as normal rather than as an error is right. Rendering nothing for it is the part that costs the user. The server already produces a usable message for exactly this case and the client discards it — `api_contribute.go:7145`:

```go
jsonError(w, "Sign in with GitHub to see your contribution stats.", http.StatusUnauthorized)
```

## The Profile tab already solves this

Same condition, same page, different outcome. `#tab-profile` (`api_contribute.go:2811`) routes an anonymous viewer through `renderMeSignIn()` (`api_contribute.go:3361`):

```js
// Anonymous / not-yet-a-contributor: a subtle prompt, not an error.
function renderMeSignIn(mount,username){
  var msg=username
    ?('<b>'+esc(username)+'</b>, you don’t have a contributor profile on this hive yet. ...')
    :'<b>Sign in</b> to see your personal contributor profile — your rank, milestones, and hives.';
  mount.innerHTML='<div class="me-signin">'+msg+'</div>';
}
```

backed by the `.me-signin` style at `api_contribute.go:1257`. That distinction is already drawn carefully on that tab — `renderMeError()` right above it exists so a render failure is never disguised as a missing profile. Operations makes neither distinction: anonymous, no-profile, and broken all render identically as absence.

## Suggested fix

Reuse what is already there rather than inventing a second pattern. In `ccLoadMine()`, distinguish the outcomes instead of returning early on all of them:

- **401** — reveal `#cc-mine-card` and render a `.me-signin` body: sign in to see your own numbers. Ideally a link to the sign-in flow, since the page currently offers none.
- **403** (signed in, no contributor profile on this hive) — reveal it with the existing "ship a task to start your card" wording `renderMeSignIn` already uses for that case.
- **Transport/parse fault** — the `renderMeError` treatment, so a bug is visible as a bug.

The wording and the CSS class can be lifted from the Profile tab verbatim, which also keeps the two surfaces consistent.

## Split out

An earlier revision of this issue carried a lower-confidence note about the **"My work"** panel on the same tab rendering fleet-wide work under a personal title. That turned out to be a separate, confirmed defect with a different fix, so it now lives in its own issue: #6945. This issue is only about identity-dependent panels vanishing silently when signed out.

## Environment

- `origin/v4` @ `701c76fd`
- `/contribute` → Operations tab, signed out
- Observed 2026-09-14

### LINKED ISSUE #6945: Operations "My work" panel is fleet-wide, not scoped to the viewer
## Summary

The Operations tab's **"My work"** panel is not scoped to the viewer. It renders the whole fleet's in-flight work to every visitor, including anonymous ones, under a title that reads as personal.

Split out of #6937, which covers a different defect on the same tab (identity-dependent panels vanishing silently when signed out). This one is about scoping and naming, not about the signed-out prompt.

Verified against `origin/v4` @ `701c76fd`.

## What happens

Open `/contribute` → **Operations**, signed out. "My work" lists the tasks currently in flight across every connected clanker on the hive, each row annotated with the contributor it belongs to. For an anonymous viewer, every row is by definition someone else's.

## Why

`/api/contribute/fleet` is public. The route is registered with no auth wrapper (`api_contribute.go:316`), and the comment four lines below it states the intent for the sibling route:

```go
// Under the /api/contribute* prefix, so isPublicPath (server.go) makes it PUBLIC —
// anonymous viewers may subscribe to this read-only info.
```

`handleContributeFleet` (`api_contribute.go:7029`) performs no viewer lookup, and `FleetSnapshot()` (`contribute_ws.go:3286`) takes no viewer argument — it iterates `h.connections` in full:

```go
func (h *ContributeWSHub) FleetSnapshot() FleetSnapshot {
	...
	for _, c := range h.connections {
```

On the client, `renderWork()` (`api_contribute.go:4970`) filters only by status — `workMatchesFilter` (`api_contribute.go:4956`) tests `w.status` against all / active / review / done and nothing else. No identity is consulted at any layer.

## The pieces to scope it are already present

This is not a case of missing data.

- `FleetWorkItem.GitHubUsername` (`contribute_ws.go:3262`) is carried on every item, and `renderWork` already reads `w.github_username` to render the per-row contributor chip.
- `ccMeUsername` — the signed-in viewer's username, resolved from `/api/gh-user-auth/status` (`api_contribute.go:3127`) — is already available on this tab, and is already used for identity matching elsewhere on the same page: `lbRow` (`api_contribute.go:3140`) compares it case-insensitively to highlight the viewer's own row in the standings.

So both halves of the comparison exist and are already used together on this page. "My work" just never applies them.

## What I am not claiming

The panel may be intended as the *operator's* view of their hive rather than a personal list — the "Your army" roster directly above it reads that way, and a fleet-wide view is clearly useful. I am not asserting the data shown is wrong.

The problem is that the title claims something the contents do not honour, and it sits immediately above "Your contribution", which genuinely is per-viewer. Whichever way it is resolved, the two adjacent panels should not use the same possessive voice for different scopes.

## Options

1. **Scope it to the viewer.** Filter `renderWork`'s list by `ccMeUsername` when a viewer is known, matching `lbRow`'s case-insensitive comparison. Anonymous viewers then get an empty state — which should say "sign in to see your work" rather than the current "No work items in flight.", for the same reason as #6937.
2. **Rename it.** If fleet-wide is the intent, call it "Fleet work" or "In-flight work" and leave the behaviour alone. Cheapest option, and it removes the contradiction with the adjacent card.
3. **Both** — a fleet-wide panel plus a viewer filter toggle (`All` / `Mine`), alongside the existing status filters.

I lean toward 2 unless the panel was meant to be personal all along, in which case 1. Happy to send a PR once you pick.

## Environment

- `origin/v4` @ `701c76fd`
- `/contribute` → Operations tab
- Observed 2026-09-14

clubanderson:
## Summary
- Combined the #6937 and #6945 fixes intentionally because both defects share the Operations tab rendering path.
- Show the Profile tab .me-signin treatment for the Operations contribution card when the viewer is anonymous or lacks a contributor profile.
- Scope Operations My work (including Done activity) to the signed-in viewer and show a sign-in empty state for anonymous visitors.
- Added the required changelog fragment.

Fixes #6937
Fixes #6945

## Verification
- cd src && export GOMODCACHE=$HOME/.copilot/session-state/1fc5946e-68c2-4f54-bc30-0eaf434796c7/files/gomodcache GOFLAGS=-modcacherw && go test ./pkg/dashboard -count=1 -run TestOps(ContributionCardUsesProfileSignInPrompt|MyWorkIsScopedToViewer)
- Reverted src/pkg/dashboard/api_contribute.go with the new tests still present and confirmed those tests fail; restored the fix and confirmed they pass.
- cd src && export GOMODCACHE=$HOME/.copilot/session-state/1fc5946e-68c2-4f54-bc30-0eaf434796c7/files/gomodcache GOFLAGS=-modcacherw && go build ./... && go test ./pkg/dashboard/...


clubanderson:
Closing this in favour of #6946 and #6948, which cover the same two defects.

This was my error, not a problem with the work here. I dispatched this PR to fix #6937 and #6945 **together**, reasoning that they touch the same rendering path and separate PRs would conflict. @Danathar had already opened #6946 (17:37Z) and #6948 (17:51Z) as one PR per issue; this one landed at 17:59Z, last of the three. All three modify `src/pkg/dashboard/api_contribute.go`, so they cannot all land.

The split is also the right call, and the issues themselves say so — #6945 opens with "Split out of #6937". I should have checked for existing PRs before dispatching. This is the second time today I have caused a collision this way (the #6825 RFC), so the lesson is now explicit: **check open PRs for an issue immediately before starting work on it, not just the issue's link state.**

**The two approaches also resolve #6945 differently, and #6948's is the better reading.** This PR scoped the panel to the viewer (`workBelongsToViewer`, filtering by `ccMeUsername`) and showed signed-out users an empty state. #6948 instead renames it to **Fleet work** and adds scope chips, keeping the data visible but honestly labelled. For an anonymous viewer that is strictly better: my version showed them nothing, whereas fleet-wide activity is genuinely useful information and was never actually secret. The issue explicitly allowed either resolution; #6948 picked the one that loses no information.

One piece of this branch that may be worth keeping regardless, if a genuinely personal panel is added later: `renderWorkSignInEmpty` plus the `<div class="me-signin">Sign in to see your work.</div>` empty state, which reuses the Profile tab's `.me-signin` affordance for the work list specifically. #6946 already brings that affordance to the contribution card, so this would only matter for a future per-viewer work panel. Happy to open it separately if that is ever wanted.

No action needed on #6937 or #6945 — both are covered by the two PRs above, both of which carry proper `Fixes` keywords on `v4` and will close their issues on merge.

Issue #6954, opened 2026-09-14
[v5] ✨ contributor quota preflight: negotiate quota_preflight_v1 relay→hub instead of proxying on protocol version


Danathar:
Sub-issue of #6833. This is the part that touches the #6825 assignment boundary, split out so it can be sequenced against #6825 independently of the other three.

## Summary

#6833 requires that the preflight be capability-negotiated, and lists "Mixed-version hub/relay behavior remains backward compatible through capability negotiation" as an acceptance criterion. #6931 added `quota_preflight_v1` to the hub's **advertised** set (`serverCapabilities()` in `src/pkg/dashboard/contribute_protocol.go`) and the relay reads it correctly via `hubSupportsQuotaPreflight`. The reverse direction is not implemented.

`ContributorCapabilities` (`src/pkg/dashboard/contribute_protocol.go:142`) has **no field** in which a relay can advertise a capability — `quota_preflight_v1` exists only as a hub-outbound string. So the hub's check proxies on the declared protocol version instead:

`src/pkg/dashboard/contribute_ws.go:5055`
```go
func contributorSupportsQuotaPreflight(c *ContributorConnection) bool {
	if c == nil || c.capabilities == nil {
		return false
	}
	return strings.TrimSpace(c.capabilities.RelayProtocolVersion) != ""
}
```

That value gates credential delivery:

```go
if !h.requireExplicitAccept() && !contributorSupportsQuotaPreflight(contributor) {
	h.deliverTaskCredential(contributor, "auto_accept")
} else {
	// withhold, wait for task_accepted
}
```

So the hub now withholds the credential on the auto-accept path for **every relay that declares any protocol version**, whether or not it has ever heard of `quota_preflight_v1`. #6931 did not bump `RELAY_PROTOCOL_VERSION` (`bin/contributor-relay.js:433`, still `1.2`), so every already-deployed relay trips this.

It does not break today, because the in-tree relay has sent `task_accepted` unconditionally on `task_assign` since well before #6833 — the send at `bin/contributor-relay.js:4347` (as of `cc0d24b8`) predates #6931. That makes the current behaviour a coincidence of relay history rather than negotiation, and it is exactly the criterion above that is unmet: nothing in the wire contract establishes that a connected relay will answer.

## Why this one is separable

[...]

Issue #6953, opened 2026-09-14
[v5] ✨ contributor quota guard: scoped override controls and a cross-process quota-pool store


Danathar:
Sub-issue of #6833. These are the "interactive override controls" and "cross-process quota-pool storage" items @clubanderson descoped from #6931.

## Summary

Two of #6833's requirements have no implementation after #6931.

### `ask` mode is indistinguishable from `pause`

`HIVE_CONTRIBUTOR_QUOTA_GUARD` accepts and validates `ask`, but nothing prompts and no control path exists. #6833 asks for four scoped controls — pause until reset, continue once, continue until this window resets, disable for this session — reachable both from an attached launcher **and** from a reliable command path (for example `just contribute-quota continue-once`) via a local control socket or atomic state file, because the relay usually runs detached and correctness must not depend on reading stdin from a background process.

What exists today:

- `contributorQuotaStayPaused` (`bin/contributor-relay.js:306`) is settable **only** through the `HIVE_RELAY_TEST_MODE` export at `:4868`. There is no production caller.
- ~~`Guard.StayPaused` / `Guard.ContinueOnce` / `Guard.StartTask` / `Guard.FinishTask` exist in `src/pkg/contributorquota/guard.go` on a type nothing constructs.~~ **No longer true, and it changes this issue's scope.** #6955 deleted `src/pkg/contributorquota` under the #6951 ruling, so those four methods do not exist anywhere. The override controls are not a wiring job against an existing Go type — they have to be **built in the relay**, in JS. Strictly more implementation, but only one implementation, which is the point of the ruling.
- The JS guard has no continue-once concept at all, and no notion of an override expiring at a window/reset-epoch boundary.

So the override-expiry criteria in #6833 are currently untestable end to end, and the guard's only real escape is restarting the relay with `HIVE_CONTRIBUTOR_QUOTA_GUARD=off`.

The terminal banner that did land (`logContributorQuotaDecision`) covers most of #6833's required message content but advertises only that opt-out, and it prints no reset time because no adapter supplies one yet (see the adapter sibling).

### Guard state is per-process, so two relays oversubscribe

#6833 requires readings stored atomically and keyed by the local credential/quota pool, not by relay process, so concurrent relays on the same provider account share one guard state. Today all guard state is in-process module-level variables. Two relays under one account — which the documented `HIVE_SESSION` multi-backend setup makes ordinary — each hold their own reserve against the same pool and can independently admit work.

[...]

Issue #6952, opened 2026-09-14
[v5] ✨ contributor quota guard: normalized provider readings from Codex, Claude Code, and Agy


Danathar:
> ## ⚠️ Split into per-adapter sub-issues — claim a child, not this issue
>
> This issue described three independent adapters plus a gap that belonged to none of them. Taking it as one work item means colliding with whoever holds another third of it, so it is now a tracker:
>
> | | Status |
> |---|---|
> | #6964 — Codex adapter (+ the shared `LimitWindow` duration field) | **Claimed** by #6961 |
> | #6965 — Claude Code adapter, and deciding the source | Unclaimed |
> | #6966 — Agy adapter | Unclaimed |
> | #6967 — publish readings to the relay, and flip `unprovisioned` to a hold | Unclaimed |
> | #6980 — Copilot adapter (monthly premium-request window; outside #6833's initial scope, does **not** gate #6967) | Unclaimed |
>
> **#6967 was not in the criteria below and is the keystone.** Nothing in the tree writes `HIVE_CONTRIBUTOR_QUOTA_READING_FILE` — every reference is a read in `bin/contributor-relay.js`. All three adapters can land, be correct, and ship with real fixtures, and every default install still runs `unprovisioned` with a guard that enforces nothing.
>
> #6964 depends on nothing. #6965 and #6966 need only the shared `rotation.LimitWindow` duration field from #6964. #6967 needs at least one adapter publishing.
>
> The analysis below is retained as the shared spec — the per-adapter criteria now live in the children.

Sub-issue of #6833. This is the "full provider adapter bootstrapping" item @clubanderson descoped from #6931.

## Summary

The guard merged in #6931 has no way to obtain a real reading. `readContributorQuotaReading()` only reads `HIVE_CONTRIBUTOR_QUOTA_READING_FILE` / `HIVE_CONTRIBUTOR_QUOTA_READING_JSON`; nothing populates either, and no prober constructs a normalized reading. (This originally cited `PaidCreditsAvailable` in `src/pkg/contributorquota/guard.go`; that package was deleted by #6955 under the #6951 ruling, so the paid-credits signal now has neither a producer nor a declaration — see #6964.). The existing `rotation` probers were extended to emit `LimitWindow`s, but with gaps that matter for admission decisions.

## Codex — verified against the real schema

[...]

Issue #6951, opened 2026-09-14
🐛 contributor quota guard enforces in the JS relay, but the fail-open fix landed in an unwired Go package


Danathar:
Sub-issue of #6833. Follows the merge of #6931.

## Summary

#6931 landed the contributor quota guard **twice**: once as `src/pkg/contributorquota` (Go) and once as `evaluateContributorQuota` in `bin/contributor-relay.js`. Only the JS copy is in the admission path — `go list -deps ./...` shows **no package imports `src/pkg/contributorquota`** outside its own test.

The consequence is not theoretical. The fail-open defect #6931 describes fixing ("`Evaluate` failed open on an unrecognized limit window") was fixed in the Go copy and is **still live in the JS copy that actually gates work**:

`bin/contributor-relay.js:366`
```js
if (!recognizedQuotaWindow(kind)) continue;
```

Reproduced through the relay's own `HIVE_RELAY_TEST_MODE` export, same shape as the original report:

```
exhausted UNRECOGNIZED ("daily")  -> {"admit":true}
exhausted RECOGNIZED   ("weekly") -> {"admit":false,"wait":true,"reason":"guarded","remaining_pct":0,"required_reserve_pct":20}
```

So an exhausted window kind this build has not been taught admits work, exactly as before the fix. `weekly_scoped` was itself a late addition to Claude's reporting, so new kinds do arrive.

The same split means #6931's `ContinueOnce` rearm test and its 94.7% coverage figure describe a `Guard` type nothing constructs.

## Decision needed first

**Which implementation is authoritative?** Everything below depends on it, and a dead second copy is the mechanism that produced this bug. Two workable answers:

- **JS is authoritative** — delete `src/pkg/contributorquota` until something wires it, and keep `rotation.LimitWindow` as the shared wire shape.
- **Go is authoritative** — the relay obtains decisions from the hive binary rather than reimplementing them, and the JS copy goes away.

Either is fine. Having both is not.

## Also in scope

**No resume path.** Every `ready` send in the relay is event-driven (task completion, CLI recovery). #6541 arms `quotaHoldTimer` to release its hold and re-send `ready`; the #6833 guard arms nothing. `contributorQuotaPaused` (`bin/contributor-relay.js:293`) is written at :374/:379 and read nowhere outside the test export. Once the guard suppresses `ready` between tasks, the relay never asks for work again for the life of the process.

[...]

PR #6957, opened 2026-09-14
📝 docs(contributor): document the quota guard, and name the unknown-window hold

Danathar:
Rescoped onto `cc0d24b8`. **#6955 landed first and fixed all four defects** — this PR no longer claims any of them.

@clubanderson merged #6955 minutes before this opened, and his implementation is the one on `v5`. What remains here is the part of the original branch that was additive rather than duplicative: the documentation he asked for, plus one small observability delta.

## What is left

**1. Name the hold caused by an unrecognized window kind** (`bin/contributor-relay.js`, +25/-3)

#6955 correctly stopped skipping unrecognized window kinds, so every window is measured now. But every hold still reports the same `guarded`, and the two cases are not the same thing to the operator reading the banner:

- `guarded` — a window kind we know is genuinely low.
- `guarded_unknown_window` — the provider reported a kind this build has never seen. It is still a real limit and still refuses, but it is also the only signal that `HIVE_CONTRIBUTOR_QUOTA_SHORT_MIN_REMAINING_PCT` / `_WEEKLY_...` **did not apply to that window**, because those overrides key off the kind. A contributor who relaxed `weekly` to 5% and sees a hold at 25% is not looking at a bug.

The refusal is byte-for-byte unchanged; only the label and one banner line are new. The known-kind list is now named (`QUOTA_SHORT_WINDOW_KINDS` / `QUOTA_WEEKLY_WINDOW_KINDS`) with a comment stating it decides *which reserve* and *what the hold is called*, never *whether a window counts* — gating on it is the fail-open #6951 removed.

**2. Document the guard** (`src/docs/contributor-relay.md`, +67/-4)

The table rows for `HIVE_CONTRIBUTOR_QUOTA_READING_FILE` / `_JSON` gain the actual reading shape and two things a prober author needs: that the file is re-read per decision (so refreshing it is what lifts a hold, hence the atomic-rename requirement), and that `_JSON` is a fixed launch-time value a hold therefore *cannot* lift against. Plus a prose section under "Provider quota parks the relay" covering the five behaviours — unrecognized kinds, the two hold names, the resume timer, unreadable-is-`unknown`, and `unprovisioned`, including why that last one admits and what flips it once #6952 lands.

## Testing

`432/432` (`430` on `v5` HEAD). Two new tests; one of @clubanderson's existing assertions changes from `guarded` to `guarded_unknown_window`, which is the intended behaviour change and is called out here so it is not mistaken for a merge artifact.

Both new tests are mutation-proven — and one of them **was not** on the first attempt, in the same way he described on #6955:

| Mutation | Result |
|---|---|
| `const reason = 'guarded'` (drop the distinction) | `429/432` |
| Unknown kinds inherit the weekly override | `431/432` — *only after rewriting the test through the real `ready` path* |

The second is worth recording. The base-reserve test was first written against `evaluateContributorQuota` on the task path, where an unset complexity tier falls back to the base percentage, so `Math.max(windowReserve, tierReserve)` hides which window reserve was chosen — the test passed under its own mutation and could not have caught the regression. Rewritten through `handleMessage(auth_ok)`, where the `ready` hold uses the window reserve alone, it fails correctly.

## Answering the three questions on this PR

**The Go package — no wiring plan, do not revert the deletion.** The `+12` on `src/pkg/contributorquota/guard.go` was a `NOT WIRED` doc header for the next reader *while the authority decision was open*, nothing more; acceptance criterion 1 was explicitly not claimed. Deleting it plus the `existsSync` guard test is strictly better and permanent. It is gone from this branch.

**Criterion 5 — I did not implement it either.** This branch's original description flagged the identical deviation for the identical reason ("the acceptance criterion asks for a safe hold here. I did not make it hold" — every default install has no source, so holding stops the fleet). There was no method I found that you missed. Two independent implementations converging on overruling that criterion seems like a better argument for amending it on #6951 than either of us flagging it alone. `unprovisioned` is the better name and this PR documents it as such; the original's overloaded `unknown` conflated "broken source" with "no source".

**`HIVE_CONTRIBUTOR_QUOTA_UNKNOWN_RETRY_MS` — dropped, yours is safer.** The original admitted work for up to 5 minutes against a *configured but unreadable* source. That is a fail-open window, and it contradicts #6833 criterion 2 ("do not advertise readiness until the first valid reading is available") for the case where an adapter has not published its first sample yet. Holding immediately is right; your 60s retry already bounds the cost of a transient torn read.

Refs #6951. No sibling scope (#6952 / #6953 / #6954).


clubanderson:
Heads up, and an apology for the collision: I had #6955 open against `v5` for the same issue and merged it (`cc0d24b8`) shortly before this appeared, which is why this now reads `CONFLICTING`. That conflict is my doing, not a problem with your branch.

I'm not closing this, because on the one point where we visibly diverge **your call may be the better one** and I'd rather you decide than have my merge decide by accident:

- **I deleted `src/pkg/contributorquota`** and added a test that fails if a second unwired copy reappears, on the grounds that routing the relay through the binary would put a process exec in the `task_assign` hot path.
- **You kept it and added to it** (`+12`), so you evidently had a resolution in mind for the Go side that deletion forecloses. If that's a deprecation marker or a wiring plan, say so and I'll revert my deletion — I have no attachment to it, and "both copies exist, neither is authoritative" is the only outcome I think we should rule out.

Two other things where yours looks stronger than what I landed:

1. **Docs.** You have `+54/-2` on `src/docs/contributor-relay.md`; I added three table rows and one clause. If yours documents the reading contract properly, that's worth landing on top of mine and I'd review it.
2. I deviated from acceptance criterion 5 and flagged it on #6951: an **unconfigured** reading source yields a distinct `unprovisioned` state that admits, rather than `unknown` which holds — because nothing populates either env var until #6952 (on `hold`), so holding would stop every default install from ever asking for work again. If you implemented the criterion as written, I'd genuinely like to know how you avoided that, because I could not see a way.

For what it's worth, what landed covers: unrecognized-kind fail-open, a resume timer (the guard armed nothing, so one suppressed `ready` between tasks meant the relay never asked for work again), guarded parsing of torn readings, and the `ready` hold moved to the base window reserve. All four are mutation-proven — including one test I had to rewrite after it passed under its own mutation.

Rebasing onto `cc0d24b8` should leave only your genuine delta. Sorry again for the churn.

clubanderson (approved):

Danathar:
No apology needed — you merged first and your implementation is the one on `v5`. I've rescoped this PR onto `cc0d24b8` rather than rebasing it as-was, because "my genuine delta" turned out to be much smaller than the diff suggested: `+134/-8` now, from `+492/-233`, and mostly docs. Answering your three points in order.

**The Go package: no wiring plan. Don't revert the deletion.** Those `+12` lines were a `NOT WIRED` doc header for the next reader *while the authority decision was open* — that's why acceptance criterion 1 was left unchecked rather than claimed. It was a note that the decision existed, not a resolution to it. Your deletion plus the `existsSync` guard test does the same job permanently and better, and it's out of this branch entirely. `src/pkg/contributorquota` should stay gone.

**Criterion 5: I didn't implement it either.** There's no method you missed. The original description on this PR flagged the identical deviation for the identical reason — quoting it, since it's now buried under the rescope:

> Judgment call worth flagging: the acceptance criterion asks for a safe hold here. I did not make it hold. […] holding would stop every contributor relay from asking for work — a far larger regression than the fail-open it closes.

Same conclusion, same argument, arrived at separately. That seems worth more on #6951 than either of us flagging it alone: two independent implementations both declined to implement criterion 5 as written, which is a case for amending the criterion rather than leaving it permanently unmet. Your `unprovisioned` is also the better name — mine overloaded `unknown` to mean both "the source is broken" and "there is no source", which are the two cases the issue was conflating in the first place. This PR documents yours.

**`HIVE_CONTRIBUTOR_QUOTA_UNKNOWN_RETRY_MS`: dropped, yours is safer.** Mine admitted work for up to five minutes against a *configured but unreadable* source. That's a fail-open window, and it's the wrong side of #6833 criterion 2 ("do not advertise readiness until the first valid reading is available") for the case where an adapter hasn't published its first sample yet. Holding immediately is right, and your 60s retry already bounds what a transient torn read costs.

**What's left, and one behaviour change.** The docs you asked for, rewritten against `v5` HEAD rather than against my branch — `HIVE_CONTRIBUTOR_QUOTA_RETRY_MS`, `unprovisioned`, hold-on-unknown-immediately, plus the reading-file shape and the atomic-rename requirement for whoever takes #6952.

The one code delta is that a hold caused by an *unrecognized* window kind now reports `guarded_unknown_window` instead of plain `guarded`, with the kind named in the banner. The refusal is unchanged — the point is that it's the only signal the short/weekly overrides **didn't apply to that window**, since they key off the kind. A contributor who set `_WEEKLY_MIN_REMAINING_PCT=5` and sees a hold at 25% remaining is not looking at a bug, and right now nothing tells them that.

This changes one of your existing assertions from `guarded` to `guarded_unknown_window`. Intentional, flagging it so it doesn't read as a merge artifact.

**On your mutation note — it got me too, and in a place worth recording.** My base-reserve test was first written against `evaluateContributorQuota` on the task path. An unset complexity tier falls back to the base percentage there, so `Math.max(windowReserve, tierReserve)` hides which *window* reserve was chosen: the test passed under its own mutation and could not have caught the regression it was for. Rewritten through `handleMessage(auth_ok)`, where the `ready` hold uses the window reserve alone, it kills at `431/432`. Same failure mode you hit, different mechanism — the task path structurally cannot observe a window-reserve bug. Worth knowing for anything else that tests reserves. `432/432` overall; both new tests kill their mutations.

**One thing you should know about, unrelated to the review.** Between your merge and this rescope, something auto-rebased this branch (`0fda836f`, 19:00Z) and resolved every conflict by taking my side wholesale — it deleted `unprovisioned`, `armContributorQuotaRetry`, `baseReserveOnly` and `HIVE_CONTRIBUTOR_QUOTA_RETRY_MS`, i.e. it would have reverted your merge from twelve minutes earlier, and GitHub reported it `MERGEABLE`. Your `lgtm` was correctly dropped when it pushed, so nothing was at risk here, but I'm mentioning it because that resolution was automatic and would presumably do the same thing on the next collision. I force-pushed over it after checking what it was.

**Finally: #6952 no longer carries `hold`** — it's `help wanted` / `priority/important-soon` now, same as this issue. Both of our PR descriptions assume it's held, and it's the one that makes `unprovisioned` flip to a real guard. Deliberate, or did it come off with the rest of the triage sweep?

clubanderson (approved):

PR #6961, opened 2026-09-14
🐛 normalize Codex contributor quota readings from the real schema

clubanderson:
Normalizes the Codex quota reading against the provider's real schema, so the guard shipped in #6931 has something trustworthy to evaluate. Base is `v5`.

## Criteria

| # | Criterion | Status |
|---|---|---|
| 1 | `LimitWindow` carries window duration | ✅ `DurationMins` added |
| 2 | Codex reads `primary` **and** `secondary`; kind derived from `windowDurationMins` | ✅ both parsed; kind derived |
| 3 | `credits` populates paid-credits-available; `planType` surfaced | ✅ both, as tri-state/string |
| 4 | `ordinaryUsageAllowed == null` ⇒ unknown, never recovery | ✅ tri-state that can only remove headroom |
| 5 | Test asserts no spend/billing mutation | ✅ `TestNoCodexSpendOrBillingMutation` |
| 8 | Unrecognized schema reports explicitly, no silent permissive fallback | ✅ returns an error |
| 9 | No adapter consumes a model turn | ✅ unchanged — `account/rateLimits/read` over app-server |
| 6, 7, 10 | Claude / Agy source decisions and fixtures | ⛔ **deliberately not attempted — see below** |

Behavioural notes worth review:

- **Availability now binds to the most-used window.** Previously only `primary` existed, so an exhausted `secondary` was invisible. Reporting the roomier window's headroom is the gap that made parsing `secondary` worth doing.
- **Unknown durations produce an unrecognized kind, on purpose.** `window_<N>m` rather than a guess. Since #6955 the guard evaluates unknown kinds against the default reserve, so an unfamiliar window is still *enforced* — a fabricated familiar label would not be.

## Where I deviated on `ordinaryUsageAllowed`

The criterion reads "`null` maps to `unknown`, never to recovery". My first implementation returned a hard error when the field was absent, and **it broke two existing prober tests**, which is what made me look again: real payloads in this repo's own fixtures omit it, so a strict reading would have made the Codex probe permanently unusable for rotation.

"Never to recovery" is a rule about not *granting* headroom. So the field is now tri-state and can only ever **remove** it:

- `false` ⇒ unavailable, overriding healthy percentages
- absent ⇒ grants nothing, carried as `nil` for consumers that need a definite answer, availability falls back to the windows alone

Nothing is manufactured from a nil, and rotation keeps working. If you intended the stricter reading, say so and I will flip it.

## Claude and Agy are NOT touched, deliberately

Criteria 6, 7 and 10 need a decision *and* a real payload, and I have neither. The issue makes the risk explicit:

> Doc-only adapters are the main risk in this whole feature — a guard that silently misreads is worse than no guard.

I have no live credentials for any provider on this host, so writing a status-line parser for Claude or replacing Agy's regex scrape would be exactly the doc-only guesswork the issue warns against. They keep their current behaviour, which is honest, rather than gaining an unverified new one. **#6952 should stay open for those three criteria.** Codex is the one adapter whose schema the issue independently verified (codex-cli 0.154.0), which is why it is the one I changed.

## Verification

`pkg/rotation`, `pkg/watchdog`, `pkg/dashboard`, `cmd/hive` all pass; `gofmt` clean; `go build ./...` clean.

Mutation-proven — each mutation leaves the suite with exactly one failing test:

| Mutation | Result |
|---|---|
| Hard-code `Kind: "weekly"` (the original defect) | 1 failing |
| Stop parsing `secondary` | 1 failing |
| Ignore `ordinaryUsageAllowed == false` | 1 failing |
| Accept an unrecognized schema as available | 1 failing |

Baseline and restored: 0 failing.

Refs #6952


Danathar:
Heads up — I split #6952 into per-adapter sub-issues, and **this PR now corresponds to #6964** (Codex adapter, plus the shared `rotation.LimitWindow` duration field, which I confirmed this PR already lands).

No action needed on the PR; #6964 records it as claimed so nobody starts a second Codex implementation. The reason for the split is that #6952 bundled three independent adapters, so a contributor picking it up would have collided with this.

Splitting it also surfaced something that was in nobody's criteria, now #6967: **nothing writes the reading.** Every reference to `HIVE_CONTRIBUTOR_QUOTA_READING_FILE` / `_JSON` in the tree is a read in `bin/contributor-relay.js` — the adapters normalize into `rotation.LimitWindow` inside the hive binary, the guard reads JSON from a file path in the relay process, and nothing connects the two. So all three adapters can land and be correct and every default install still runs `unprovisioned` with a guard enforcing nothing. That also means #6967, not the adapters, is what gates flipping `unprovisioned` to a hold.

Not a criticism of this PR's scope — it does what #6952 asked for. The gap was in the issue.

PR #6978, opened 2026-09-14
🐛 fix: honor CODEX_HOME in contributor containers

TB516:
This PR was created by GPT-6.

When `CODEX_HOME` points outside `~/.codex`, contributor setup finds the credentials, but container startup stages the wrong directory. Use `CODEX_HOME` for container configuration staging, with the existing `~/.codex` fallback.

Includes a changelog fragment and coverage in the existing container-launch test.

## Testing

- `bash src/deploy/test_contributor_container_resources.sh`: 30 passed, using fake credentials and stubbed Docker/Podman.
- Verified the custom-path checks fail for both runtimes without the fix.
- Shell syntax and `git diff --check` passed.
- No live container or contributor task was started; Go tests were not run because this change is confined to the shell recipe.


Danathar (approved):

clubanderson (approved):
LGTM — CODEX_HOME honored with fallback, covered by container-resource staging tests, changelog fragment included.

Issue #6980, opened 2026-09-14
[v5] ✨ contributor quota adapter: Copilot readings, monthly premium-request window and deciding the source


Danathar:
Sub-issue of #6952. Fourth adapter alongside #6964 (Codex), #6965 (Claude Code), and #6966 (Agy).

**Unclaimed.** Independent of the other adapters except for the shared `rotation.LimitWindow` duration field landing in #6964 first. **Not a blocker for #6967** — #6833's initial scope boundary is "Codex, Claude Code, and Agy. Other backends can add adapters against the same normalized interface." This is the first of those other backends, and #6967's `unprovisioned → hold` flip must not wait on it.

## Why Copilot needs an adapter at all

Copilot is a first-class backend (`KNOWN_BACKENDS` in `config/backends.conf`), and it is the backend the guard's own tests exercise — every `loadRelay({ backend: 'copilot', … })` in `bin/contributor-relay.test.js` §6953. But there is **no `CopilotProber`**: `NewManager` in `src/pkg/rotation/rotation.go` wires only `claude`, `codex`, `agy`, and `deepseek`. So once #6967 lands, every Copilot contributor gets "guard unavailable" — on a backend a large share of contributors run.

The overage model is also exactly the thing #6833 exists to protect against. A Copilot plan carries a **monthly premium-request allowance**; when it is exhausted, requests either stop or — if the contributor has enabled a budget for additional premium requests — roll into pay-per-request billing. That budget is the Copilot equivalent of #6833's "paid credits / extra usage available, without enabling them" signal, and a Hive session left running unattended is precisely how a contributor burns through it.

## Why this is not a copy of #6965

The window shape and the data source are both different enough to need their own decisions.

### Window shape

There are no rolling five-hour / weekly windows. The unit is a **count of premium requests per calendar month**, with a plan-dependent allowance, a reset on the first of the month, and per-model multipliers — some models are 0× and never count. The relay's normalized contract only knows `kind: five_hour` and `kind: weekly` (`evaluateContributorQuota` in `bin/contributor-relay.js`, documented in `src/docs/contributor-relay.md`). A `kind: monthly` window is enforced against the base reserve and holds as `guarded_unknown_window`, but `HIVE_CONTRIBUTOR_QUOTA_SHORT_MIN_REMAINING_PCT` / `_WEEKLY_` do not apply to it.

[...]

Issue #6967, opened 2026-09-14
[v5] ✨ contributor quota guard: publish normalized readings to the relay (opt-in)


Danathar:
Sub-issue of #6952. This one is not an adapter — it is the missing link between them and the guard, and it was not covered by #6952's acceptance criteria.

## The gap

Nothing writes the reading. On `v5`, every reference to `HIVE_CONTRIBUTOR_QUOTA_READING_FILE` / `_JSON` in the entire tree is a **read** in `bin/contributor-relay.js`:

```
$ grep -rn "CONTRIBUTOR_QUOTA_READING" --include="*.go" --include="*.js" --include="*.py" --include="*.sh" .
bin/contributor-relay.js:291   const QUOTA_READING_FILE  = ...   (read)
bin/contributor-relay.js:292   const QUOTA_READING_JSON  = ...   (read)
bin/contributor-relay.js:393   parse('...READING_JSON', ...)     (read)
bin/contributor-relay.js:396   parse('...READING_FILE', ...)     (read)
bin/contributor-relay.js:414   warning naming both vars          (read)
```

The adapters (#6964 / #6965 / #6966) normalize provider data into `rotation.LimitWindow`s inside the hive binary. The guard reads JSON from a file path in the relay process. **Nothing connects the two.** So all three adapters can land, be correct, and ship with real fixtures, and every default install still runs with `state: unprovisioned` and a guard that enforces nothing.

This is the keystone of the feature, and it is currently nobody's criterion.

## Scope

A publisher that takes a normalized reading and makes it visible to the relay, plus the flip that turns the guard on.

## Acceptance criteria

- [ ] A normalized reading reaches the relay without the contributor hand-configuring an env var — a default install of a supported backend gets a real reading.
- [ ] Written **atomically** (temp file + rename), per #6833's "store readings atomically". The relay detects a torn read and holds rather than crashing or admitting blind, but detection is a safety net, not a substitute: a torn read costs a full `HIVE_CONTRIBUTOR_QUOTA_RETRY_MS` before the relay re-reads.
- [ ] Refreshed on the cadence #6833 asks for — after each task completion, before accepting an assignment, on provider status updates, and around a reported reset time. A stale file holds a relay indefinitely even when the real quota has recovered.
- [ ] Keyed by the local credential/quota pool, not the relay process, so concurrent relays on one provider account see one shared guard state (#6833 criterion 7 and the cross-process half of #6953 — coordinate scope with that issue rather than duplicating it).
[...]

Issue #6966, opened 2026-09-14
[v5] ✨ contributor quota adapter: Agy readings from the documented status-line quota map


Danathar:
Sub-issue of #6952. Splitting that issue per adapter so the three can proceed in parallel without colliding.

**Unclaimed.** Independent of the Codex adapter (#6964) except for the shared `rotation.LimitWindow` duration field landing there first.

## Scope

The Agy adapter, which today is the weakest of the three.

`AgyProber` regex-scrapes the text of `agy --print "/usage"` (`agyWeeklyRe`). #6833 explicitly rules that out in favour of the documented status-line `quota` map (`remaining_fraction`, reset fields, plan tier), and asks for CLI capability detection rather than parsing decorative terminal text — decorative output is not a stable interface and changes without notice.

The emitted window also carries **no `ResetAt`**, so the terminal message #6833 requires cannot show reset timing for Agy. A contributor told "paused on quota" with no reset time cannot tell a four-hour wait from a four-day one.

## Acceptance criteria

- [ ] Reads the documented status-line `quota` map (`remaining_fraction`, reset fields, plan tier) rather than scraping `/usage` text.
- [ ] Emitted windows carry `ResetAt`, so the guard's terminal message can show reset timing.
- [ ] An existing contributor status line is preserved, in local **and** container launch modes, with a test covering a contributor who already has one configured.
- [ ] CLI capability detection rather than version-sniffing or text matching; a CLI lacking the interface, or returning an unrecognized schema, reports that explicitly and enters the configured unknown-data behaviour — no silent fallback to a permissive reading.
- [ ] Obtaining a reading consumes no model turn; the documented usage/quota refresh path is used where one is needed.
- [ ] A recorded real Agy payload ships as a test fixture, or the adapter ships reporting `unknown` until one exists.

That last one is the main risk in this whole feature, and it is sharpest here: the current scraper's failure mode is a *confident wrong number*, not an error. A guard that silently misreads is worse than no guard.

## Note

This normalizes readings; it does not publish them. Nothing writes `HIVE_CONTRIBUTOR_QUOTA_READING_FILE`, so the relay's guard stays `unprovisioned` and inert even after this lands — that gap is tracked separately under #6952.

Code verified against `v5` at `7f46ad8`; agy 1.1.22 exposes no auth surface on the verifying host, so no live reading was obtainable.

clubanderson:
[...]

Issue #6965, opened 2026-09-14
[v5] ✨ contributor quota adapter: Claude Code readings, and deciding the source


Danathar:
Sub-issue of #6952. Splitting that issue per adapter so the three can proceed in parallel without colliding.

**Unclaimed.** Independent of the Codex adapter (#6964) except for the shared `rotation.LimitWindow` duration field landing there first.

## Scope

The Claude Code adapter — and first, a decision about which source it reads.

#6833 specifies the documented status-line JSON `rate_limits.five_hour` / `rate_limits.seven_day` fields, chaining any existing user status-line command and preserving its output. `ClaudeProber` instead probes an HTTP endpoint using stored credentials.

That may well be the better source — it avoids depending on the contributor having a status line configured at all, and it does not require an overlay in container mode. But it is a *different* source than #6833 and its acceptance criteria describe, and the divergence is currently implicit in the code rather than recorded anywhere. **Decide it here and write the decision down**, either way:

- **Status line** — then the chaining requirement is load-bearing: a contributor's existing status-line command must be preserved and must keep working, in local **and** container launch modes, via an ephemeral overlay/wrapper where necessary. Overwriting a contributor's configuration to enable a safety feature is not an acceptable trade.
- **HTTP endpoint** — then #6833's adapter table should be amended to match, with the reason, so the next reader does not "fix" the adapter back to the status line.

## Acceptance criteria

- [ ] The source is decided and documented, and #6833's adapter table agrees with what shipped.
- [ ] If status-line: an existing contributor status-line command is chained and its output preserved, in local **and** container launch modes, with a test covering a contributor who already has one configured.
- [ ] A CLI version lacking the interface, or returning an unrecognized schema, reports that explicitly and enters the configured unknown-data behaviour — no silent fallback to a permissive reading.
- [ ] Obtaining a reading consumes no model turn. A non-token-consuming usage refresh/bootstrap path is used when the fields have not appeared yet.
- [ ] Emitted windows carry reset timing, so the terminal message #6833 requires can show it.
- [ ] A recorded real Claude Code payload ships as a test fixture, or the adapter ships reporting `unknown` until one exists.

That last one is the main risk in this whole feature: a guard that silently misreads is worse than no guard, because it manufactures confidence that Hive will stop in time.

## Note

[...]

PR #6993, opened 2026-09-14
✨ rotation: normalize GitHub Copilot contributor quota readings

clubanderson:
Refs #6980

Fourth contributor-quota adapter (`CopilotProber`) in `src/pkg/rotation/rotation.go`, alongside Codex (#6964), Claude Code (#6965), and Agy (#6966). Without it, once #6967 flips `unprovisioned -> hold`, every Copilot contributor gets "guard unavailable". This issue **normalizes** readings only; it does not publish them (#6967 carries them to the relay), so `bin/contributor-relay.js` behaviour is unchanged.

> Base is `v5` per the directive that all of @Danathar's issues are v5 issues. Closing keywords are silently ignored on a non-default base, so this uses `Refs #6980`; the issue is closed by hand.

## Decision 1 - window shape: monthly, governed by the base reserve
Copilot's unit is a count of premium requests per **calendar month** (reset on the 1st), not a rolling window. Rather than teach the relay a new `monthly` reserve override (a JS behaviour change out of scope here), the emitted window carries `kind: monthly` with its duration and a **first-of-next-month `ResetAt`**, and the guard evaluates it against the **base reserve**, holding as `guarded_unknown_window`. `HIVE_CONTRIBUTOR_QUOTA_SHORT_/_WEEKLY_MIN_REMAINING_PCT` do not apply; only `HIVE_CONTRIBUTOR_QUOTA_MIN_REMAINING_PCT` does. Documented in `src/docs/contributor-relay.md`.

## Decision 2 - data source: documented billing endpoint, honest `unknown`
`GET /users/{username}/settings/billing/premium_request/usage` reports **consumed** premium requests (`grossQuantity`/`discountQuantity`/`netQuantity`) but **not** the plan allowance, so `pct_remaining` **cannot be derived** - the adapter reports the headroom as **`unknown`** (a probe error the guard holds on, never a fabricated percentage) while surfacing the consumed count informationally. `netQuantity > 0` populates the paid-overage signal (a documented direct signal). Rejected alternatives, recorded in the docs: the Copilot CLI has no documented non-interactive usage output; the undocumented IDE quota endpoint is **not** read; and `~/.copilot/session-state/*/events.jsonl` (already read by `detectCopilotModel`) was verified to carry only model-selection fields, not premium-request accounting.

## Safety
- **No budget mutation.** No code path touches any budget-mutating endpoint - asserted by `TestNoCopilotBudgetMutation` (static scan) and `TestCopilotProbeUsesOnlyGetVerb` (runtime: every request is a GET), per #6833's "no code path purchases credits".
- **No permissive fallback.** Missing token scope (401/403), a logged-out CLI (`gh auth token` / `GET /user` failure), and an unrecognized schema each report explicitly and enter the unknown-data behaviour.
- **No spend of a turn or premium request.** Obtaining a reading is a plain authenticated GET.
- Emitted window carries duration and `ResetAt`.

## Fixture provenance
`testdata/copilot_premium_request_usage.json` is **schema-derived, not a live capture** (no Copilot billing token on this host), labeled as such in `testdata/README.md`; the reject-unrecognized-schema path is verified independently of the accepted-shape field spellings.

## Validation (`GOMODCACHE=/tmp/i6980-gomod GOFLAGS=-modcacherw` from `src/`)
- `go test ./pkg/rotation/...` - ok
- `go vet ./pkg/rotation/...` - clean
- `go build ./...` - ok
- `gofmt -l` - clean

Mutation testing (each guard neutralized one at a time - not deleted - then restored; all produced real test failures):

| # | Neutralized guard | Test | Observed failure |
|---|---|---|---|
| 1 | `if !recognized {` -> `&& false` | `TestCopilotHeadroomRejectsUnrecognizedSchema` | `err = nil, want an explicit unrecognized-schema error` |
| 2 | drop `ProbeErr: errCopilotEntitlementUnknown` | `...ReportsUnknownWithConsumedAndOverage` | `ProbeErr = <nil>, want errCopilotEntitlementUnknown` |
| 3 | `if net > 0` -> `if false` | `...ReportsUnknownWithConsumedAndOverage` | `PaidCreditsAvailable should be true` |
| 4 | remove `NewManager` `case "github"` | `TestNewManager_DefaultProbers` | `len(probers) = 4, want 5` |
| 5 | window `Kind "monthly"` -> `"weekly"` | `...ReportsUnknownWithConsumedAndOverage` | `window Kind = "weekly", want monthly` |
| 6 | 403/401 scope guard return -> fall through | `TestCopilotProber_MissingScopeIsExplicitUnknown` | `ProbeErr = unexpected end of JSON input, want it to name the 403` |
| 7 | inject `.../billing/budgets` reference | `TestNoCopilotBudgetMutation` | `references a budget endpoint "/billing/budgets"` |
| 8 | drop window `ResetAt` | `...ReportsUnknownWithConsumedAndOverage` | `window ResetAt = 0001-01-01..., want first of next month 2026-10-01` |
| 9 | neutralize needs-login token guards | `TestCopilotProber_NeedsLogin` | `ProbeErr = ...connection refused, want it to name the needs-login state` |

DCO: `author=andy@clubanderson.com signoff=Andrew Anderson <andy@clubanderson.com>`.


clubanderson:
Closing this in favour of **#6996**, which implements the same issue.

Both PRs were written concurrently — mine opened 21:22Z, #6996 at 21:42Z — and both independently satisfy #6980's acceptance criteria, down to converging on the same `TestNoCopilotBudgetMutation` test name and the same `monthly` window decision.

**I am closing mine even though it was first, because #6996 has the better answer to this issue's hardest problem.**

#6980's central difficulty is that `GET /users/{username}/settings/billing/premium_request/usage` reports *consumed* premium requests (`grossQuantity`/`discountQuantity`/`netQuantity`) but **never the plan's allowance**, so a `pct_remaining` cannot honestly be derived from it alone. My PR took the conservative route the issue permits: report `unknown` and surface the consumed count informationally.

#6996 goes further and adds `ProviderRotationConfig.MonthlyAllowance`, letting the operator state the entitlement the API refuses to expose:

> Their documented usage endpoints report only consumption, never the entitlement, so a remaining percentage exists only when the operator states the allowance here. 0 (unset) leaves the reading an explicit unknown rather than a guessed one.

That converts a permanently-unknown reading into a real one **when, and only when, an operator supplies the missing fact** — and it keeps the honest `unknown` as the default rather than guessing. For a guard whose whole purpose is to stop a session before it bills a contributor, an adapter that can actually produce a percentage is materially more useful than one that structurally cannot, and it does it without inventing the number. That is the right trade.

**One thing worth carrying over from here:** this branch included `src/pkg/rotation/window_kind_test.go` (+104 lines) covering window-kind banding, which #6996 does not have. I will open a small follow-up adding that coverage once #6996 lands, rather than let it be lost in the duplicate.

My fault for the collision — I dispatched this without re-checking for an in-flight PR first, which is exactly the check I know to run in this repo after #6949. No wasted work on the reviewer's side, at least: the agreement between two independent implementations is reasonable evidence that the issue's stated design was unambiguous.

Issue #6986, opened 2026-09-14
[v5] 🐛 contributor quota adapter: AgyProber cannot parse real agy output, reports `unknown` forever


Danathar:
Sub-issue of #6833. Follow-up to #6966, which shipped in #6981 (merged to `v5` as `04d3f310`) and was closed with two gaps recorded in its closing comment but no issue carrying them. This is that issue.

The adapter as landed is **safe** — every gap below fails toward `unknown` (a hold), never toward a confident wrong number, which was the whole point of replacing the `agyWeeklyRe` scraper. What is unresolved is whether it ever produces a reading *at all*.

## Gap 1 — the accepted-shape path is unproven against real agy output

`agyHeadroom` (`src/pkg/rotation/rotation.go:744`) parses the struct defined at `rotation.go:685`/`:693`:

```go
type agyUsageResponse struct {
	Quota *struct {
		Plan     string                     `json:"plan"`
		PlanTier string                     `json:"plan_tier"`
		Windows  map[string]*agyQuotaWindow `json:"windows"`
	} `json:"quota"`
}

type agyQuotaWindow struct {
	RemainingFraction *float64   `json:"remaining_fraction"`
	ResetAt           *time.Time `json:"reset_at"`
	ResetsAt          *time.Time `json:"resets_at"`
	DurationMins      int        `json:"duration_mins"`
}
```

Those **field names came from #6833/#6966 prose, not from a captured payload.** The nesting (`quota.windows`), the spelling (`reset_at` vs `resets_at` — both are accepted precisely because the real one is unknown), and `duration_mins` are all best-effort. `testdata/agy_usage.json` is an instance of that guess, not a recording; `testdata/README.md` and a `PROVENANCE CAVEAT` block on `AgyProber` both say so. agy 1.1.22 exposed no auth surface on the verifying host, so no real payload was obtainable.

**Consequence:** if the real shape differs in any of those details, `parsed.Quota == nil || len(parsed.Quota.Windows) == 0` fires and the Agy adapter reports `unknown` *permanently and silently* — indistinguishable, from the outside, from a working adapter on a host with no agy credentials. The guard holds rather than misreads, which is the correct direction, but the adapter would be dead code that looks alive.

This matters more now than it did at merge time: #6967 landed the publisher (`56897a33`), so a reading that never materializes now propagates to a real relay guard decision instead of stopping at an unused struct.

### Acceptance criteria
[...]

PR #6994, opened 2026-09-14
🐛 rotation: parse agy's real /usage envelope, and select the applicable quota pool

Danathar:
🐛 Make the Agy contributor quota adapter parse agy's **real** `/usage` output. Today it cannot, and reports `unknown` on every call.

`Refs #6986`
`Refs #6966`

> Based on **v5** (`--base v5`). Closing keywords only fire when the base is the default branch (`v4`), so `Refs` is deliberate and this will not auto-close/auto-link — I comment on the issue with the link. Please merge to v5.

## The bug

#6966 modelled agy's structured `/usage` response on the field names #6833 described in prose, because agy 1.1.22 exposed no auth surface on the verifying host and no live payload was obtainable. That was disclosed at the time, in the code, in `testdata/README.md`, and in the closing comment — this PR is the reconciliation those notes asked for, not a surprise.

Captured on a host that does have credentials (**agy 1.2.1**, which authenticates fine and serves `/usage` headless), then run through this branch's own parent-commit structs and guard:

```
agyHeadroom REJECTS the real payload: agy usage: no quota groups (unrecognized schema)
=> failOpen(provider, err) => published state: unknown (a permanent HOLD)
```

`--output-format json` turns out to be a **print-mode response formatter, not a quota API**, so the reply is a result envelope and the quota data lives under `command.data`:

```
command.data.groups[].buckets[] → {id, window ("weekly"|"5h"),
                                   remaining_fraction (0..1), reset_time}
```

| `rotation.go` expected | agy 1.2.1 emits |
|---|---|
| root key `quota` | `command.data` → `parsed.Quota == nil`, fatal on every call |
| flat `windows` map | `groups[]` → `buckets[]` |
| `reset_at` / `resets_at` | `reset_time` — both accepted spellings missed |
| `duration_mins` | `window` label: `"weekly"`, `"5h"` |
| `plan` / `plan_tier` | absent entirely |
| `remaining_fraction` | ✅ the one correct guess |

Failing to `unknown` is the right direction, and that is exactly why this went unnoticed: on a correctly configured host it looks identical to having no agy credentials installed. With #6967 now publishing readings, it feeds a real guard decision.

## The part that is a design correction, not a tag fix

Agy meters **two independent quota pools** — `gemini-*` (Flash/Pro) and `3p-*` (Claude Opus/Sonnet, GPT-OSS) — each with its own weekly and 5h bucket, routinely far apart. In the capture: **48%** remaining on Gemini, **9%** on third-party.

`agyHeadroom` folds worst-binds across all windows. Correcting only the struct tags would bind on **9%** and hold the relay for a contributor running Gemini Flash whose actual pool is at 48%.

So `AgyProber` gains an optional `Model`, which selects the applicable pool. With no model known the fold widens to every pool: that can hold work unnecessarily, but it cannot admit work against an exhausted pool — the conservative direction, chosen deliberately over guessing a pool. Bucket IDs stay group-qualified (`gemini-weekly`, `3p-weekly`) so the guard's message names the pool that actually bound.

**Not plumbed yet, on purpose:** nothing sets `Model` — `NewManager` and `main_helpers.go` construct `AgyProber{ThresholdPct: threshold}` and keep today's widen-to-all behaviour. Threading the launched model through config is a separate change with its own decisions, and I did not want to bundle it with a parse fix. The hook and its semantics are here and tested; wiring it is follow-up on #6986.

## Also corrected

- **Capability detection.** Now keys on the envelope (`status: SUCCESS`, `command.name == "usage"`, populated `command.data`) rather than on `--output-format json` being accepted — a general print-mode flag that says nothing about whether `/usage` exists.
- **Plan tier dropped** from the agy path. It is not in the payload; claiming one was a guess.
- **`num_turns: 0` / `usage.total_tokens: 0`** in the captured envelope confirm a reading consumes no model turn. #6981 could only argue that by design.

## Tests

`go test ./pkg/rotation/` ok, **62 → 63** top-level tests, `go vet` clean, `go build ./...` ok. `bin/` untouched, so `bin/contributor-relay.test.js` not run.

`TestAgyHeadroomRejectsUnrecognizedSchema` passes **unchanged** — its bodies carry no `command.data` either, exactly as its own comment predicted ("holds even if the fixture's schema is later corrected").

### Mutation evidence (production change reverted, test kept → must FAIL)

| # | Mutation | Test | Observed |
|---|---|---|---|
| M1 | pool narrowing neutralized | `TestAgyHeadroomSelectsPoolForModel` | FAIL — `len(Limits) = 4, want 2`, and the 9% `3p-weekly` bucket binds |
| M2 | `reset_time` tag reverted to `reset_at` | `TestAgyHeadroomFromFixture` | FAIL — `3p-weekly carries no ResetAt` |
| M3 | unrecognized-schema guard neutralized | `TestAgyHeadroomRejectsUnrecognizedSchema` | FAIL — all five bodies return `err = nil` |
| M4 | `"5h"` label unmapped | `TestAgyHeadroomFromFixture` | FAIL — `Kind = "5h"` want `five_hour`; `DurationMins = 0` want `300` |

Each applied to `rotation.go` only, tests unchanged; the named test failed, the file was restored, the suite re-confirmed green. M1 was first attempted by *deleting* the narrowing, which produced a build error — not evidence — so it was redone by neutralizing the condition with the locals still referenced.

## Fixture

`testdata/agy_usage.json` is now the real recording, replacing the schema-derived guess. Unredacted: the envelope carries no account identifier (`conversation_id` is empty), no email, no plan tier. The percentages are the capturing account's real levels, which is what makes the fixture useful — the 9% bucket is what exercises worst-binds and the threshold crossing.

## Caveat

**One account, one plan tier, agy 1.2.1.** This proves the shape #6966 assumed was wrong and pins the shape that exists today. It does not prove the field set is complete across tiers — a different tier may expose groups or fields this capture does not. `testdata/README.md` says so rather than implying the shape is now fully settled.

The status-line-preservation criterion left unchecked on #6966 is **not** addressed here; it remains open on #6986 alongside wiring `Model`.


clubanderson (approved):
Agy /usage envelope parsing + scoped window selection verified against tests. LGTM.

PR #7045, opened 2026-09-15
🐛 fix(terminal): resolve the hive id from disk on standalone spokes (#7043)

### LINKED ISSUE #7043: Terminal reconnect loop on standalone spokes: the Node proxy never learns the hive id the dashboard mints assertions with
> **Edited after the fix merged** to add the answer to "when did this break?" — it was [#6040](https://github.com/hivecommons/hive/pull/6040) on 2026-09-04, not #6489 as this issue first said. The rest is rewritten to be easier to follow. Fixed by #7045.

## What happens

On a self-hosted hive (podman or compose, not hub-provisioned), clicking **▶ terminal** opens a page that says **"Reconnected"** and never shows a shell.

"Reconnected" is misleading. The terminal is not reconnecting after a hiccup — it is connecting, failing, and retrying dozens of times a second. You are seeing one frame of that loop.

Seen on a rootless-podman hive, `v4` @ `01050b2`, Firefox 155.

## When it broke

**2026-09-04, in [#6040](https://github.com/hivecommons/hive/pull/6040)** (`79d6be3c`, "fix dashboard token query auth").

Measured on the affected host — the same machine, same hive, the gateway's own access log:

| date | `GET /terminal/token` | |
|---|---|---|
| Aug 29 – Sep 5 | **200** ×114 | terminal worked |
| Sep 5, 22:16 | **200** | last time it ever worked |
| Sep 15, 09:06 | **401** | first failure |
| Sep 15 | **401** ×11,174 | one session of the loop |

Nobody opened a terminal between Sep 6 and Sep 14, so the host can't narrow it further than that window — but #6040 merged on Sep 4 and is the only change that touches this path.

## Why #6040 broke it

Before #6040, a self-hosted hive's `/terminal` requests were **not checked by the Node proxy at all**:

```js
if (isHosted) {            // self-hosted hives fell straight through
```

ttyd did its own check instead, with HTTP Basic auth (`hive:<dashboard token>`). That matters because Basic auth is sent on **every** request the browser makes — the page, the token fetch, the websocket. All three were covered.

#6040 replaced that with a proxy-side check:

```js
if (isHosted || DASHBOARD_TOKEN) {   // self-hosted hives now checked here too
```

The new check accepts two credentials: a hub login cookie, or a signed "terminal assertion" cookie. A self-hosted hive has no hub, so the assertion cookie is the only option — and it never verifies. See below.

#6040 also removed the old `?token=` credential from the websocket path, so there was no fallback left.

## Why the assertion cookie never verifies

The cookie says which hive it is for, and the proxy refuses it unless that matches the proxy's own idea of which hive this is:

```js
function verifyTerminalAssertion(key, token, expectedHiveID, nowSec) {
  if (!key || !token || !expectedHiveID) return null;   // <- no id, no entry
```

The two sides of the container disagree about what that id is:

- **The dashboard (Go)** mints the cookie using `cfg.HiveID`. On a self-hosted hive there is no id to be given one, so `cmd/hive` **makes one up** at first boot (`hive-<name>`) and saves it to `/data/hive-id`.
- **The proxy (Node)** only ever read `process.env.HIVE_ID`.

On a hub-provisioned hive the hub sets `HIVE_ID` in the container environment and both sides read the same value, so this never shows. On a self-hosted hive nothing sets it — `bin/hive-podman-setup.sh` writes `HIVE_DASHBOARD_TOKEN` and nothing else.

The Go side does try to publish it:

```go
cfg.HiveID = loadOrGenerateHiveID(logger)
_ = os.Setenv("HIVE_ID", cfg.HiveID)   // src/cmd/hive/main.go
```

but `os.Setenv` only changes **that process's own** environment. The Node proxy is a separate process started by the same `entrypoint.sh`, and it never sees it.

So the proxy sits there with an empty hive id and rejects every cookie the dashboard six inches away just issued.

Confirmed on the affected host: `GET /api/status` returns `"hiveId":"hive-wild-mole"` — a made-up name, which only happens when `HIVE_ID` is absent.

## Why it loops instead of just failing

The dashboard hands the terminal a one-time `?code=` in the URL. It works exactly once. ttyd's browser client then makes two more requests, and **only one of them keeps the query string**:

| request | has `?code=` | handled by | result |
|---|---|---|---|
| `GET /terminal/?arg=…&code=X` | yes | Go dashboard, spends the code | `200` — page loads |
| `GET /terminal/token` | **no** | Node proxy | **`401`** |
| `GET /terminal/ws?arg=…&code=X` | yes | Go dashboard, checks its own cookie | `101` — socket opens |

The websocket opens, but ttyd immediately closes it because the client never got its token. The client retries. Round and round:

```
"GET /terminal/token HTTP/1.1" 401 108
"GET /terminal/ws?arg=hive-scanner&code=d3877bb… HTTP/1.1" 101 100
"GET /terminal/token HTTP/1.1" 401 108
"GET /terminal/ws?arg=hive-scanner&code=d3877bb… HTTP/1.1" 101 100
…
```

## Where #6489 fits in

[#6489](https://github.com/hivecommons/hive/pull/6489) (2026-09-10) did **not** cause this, but it changed what you see.

It fixed a sibling problem — the two sides also disagreed about the *signing key* — by having `entrypoint.sh` generate one and export it so both processes agree. Good fix. But it only did the key. The hive id stayed split.

Its side effect: `/api/terminal/handoff` started succeeding on self-hosted hives, where it used to return `503 terminal handoff requires terminal signing key and hive id`. So between Sep 4 and Sep 10 the terminal failed loudly (a 503 the dashboard showed you). After Sep 10, it fails quietly, as this loop.

Confirmed on the affected host — this prints on every boot, and only prints when the hive is *not* hub-provisioned:

```
[entrypoint] auto-provisioned a per-instance terminal signing key for this standalone spoke (#6489); set HIVE_TERMINAL_KEY to override
```

So the key halves already agree. The hive id was the last piece.

## Why no test caught it

Every terminal test in `src/proxy/` sets `HIVE_ID` in the proxy's environment, because that's what a hosted hive looks like. The self-hosted shape — cookie minted, `HIVE_ID` missing — had no test at all. The one setup where the two halves disagree was the one setup nothing exercised.

## Who is affected

Every self-hosted hive (podman, compose) on `v4` since 2026-09-04. **▶ terminal** does not work on any of them.

Hub-provisioned hives are fine.

There's no workaround unless you happen to know to set `HIVE_ID` in `hive.env` by hand to whatever `/api/status` reports — and nothing tells you that, because the screen says "Reconnected".

## The fix (#7045)

Have the proxy find the hive id the same way the Go side does: use `$HIVE_ID` if set, otherwise read `/data/hive-id`.

Read it when a request arrives, not at startup — the Go binary writes that file early in `main()`, but the proxy may be running first, and a startup read would cache an empty value forever.

The environment variable still wins, so hub-provisioned hives are untouched.

## One loose end, not fixed here

`/terminal/ws?…&code=X` goes to the Go dashboard instead of to ttyd only because a **spent, useless** code is still sitting in the query string that ttyd copies from `window.location.search`.

It works today — the Go dashboard's check is the stronger of the two — but the websocket taking a different route through the stack than every other terminal request, because of a dead credential, is worth its own look.

Danathar:
> **Edited after merge** to add the answer to "when did this break?" — it was #6040 on 2026-09-04, not #6489 as this PR first implied. The rest is rewritten to be easier to follow. The code did not change.

Fixes #7043.

## What was broken

On a self-hosted hive (podman or compose), clicking **▶ terminal** opened a page that said **"Reconnected"** and never showed a shell.

That message was misleading. The terminal wasn't recovering from a hiccup — it was connecting, failing, and retrying dozens of times a second. You were seeing one frame of the loop.

## When it broke

**2026-09-04, in #6040** (`79d6be3c`, "fix dashboard token query auth").

Measured on the affected host, from the gateway's own access log:

| date | `GET /terminal/token` | |
|---|---|---|
| Aug 29 – Sep 5 | **200** ×114 | worked |
| Sep 5, 22:16 | **200** | last time it worked |
| Sep 15, 09:06 | **401** | first failure |
| Sep 15 | **401** ×11,174 | one session of the loop |

## Why #6040 broke it

Before #6040, a self-hosted hive's `/terminal` requests weren't checked by the Node proxy at all:

```js
if (isHosted) {            // self-hosted hives fell straight through
```

ttyd checked them instead, with HTTP Basic auth (`hive:<dashboard token>`). That mattered, because the browser sends Basic auth on **every** request — the page, the token fetch, and the websocket. All three were covered.

#6040 moved the check into the proxy:

```js
if (isHosted || DASHBOARD_TOKEN) {   // self-hosted hives now checked here too
```

That check takes a hub login cookie or a signed "terminal assertion" cookie. A self-hosted hive has no hub, so the assertion cookie is the only option — and it never verified. #6040 also removed the old `?token=` credential from the websocket path, so nothing was left to fall back on.

## Why the assertion cookie never verified

The cookie names the hive it's for, and the proxy refuses it unless that matches the proxy's own idea of which hive this is:

```js
function verifyTerminalAssertion(key, token, expectedHiveID, nowSec) {
  if (!key || !token || !expectedHiveID) return null;   // <- no id, no entry
```

The two halves of the container disagreed about that id:

- **The dashboard (Go)** mints the cookie with `cfg.HiveID`. On a self-hosted hive nobody hands it an id, so `cmd/hive` invents one at first boot (`hive-<name>`) and saves it to `/data/hive-id`.
- **The proxy (Node)** only read `process.env.HIVE_ID`.

On a hub-provisioned hive the hub sets `HIVE_ID` in the container environment and both sides read it, so this never showed. Self-hosted, nothing sets it — `bin/hive-podman-setup.sh` writes `HIVE_DASHBOARD_TOKEN` and nothing else.

The Go side does try to publish it:

```go
_ = os.Setenv("HIVE_ID", cfg.HiveID)
```

but that only changes **its own** process environment. The proxy is a separate process started by the same entrypoint and never sees it.

So the proxy sat with an empty hive id and rejected every cookie the dashboard had just issued.

## Why it looped instead of just failing

The handoff `?code=` in the URL works exactly once. ttyd's client then makes two more requests, and only one keeps the query string:

| request | has `?code=` | handled by | result |
|---|---|---|---|
| `GET /terminal/?arg=…&code=X` | yes | Go dashboard, spends the code | `200` — page loads |
| `GET /terminal/token` | **no** | Node proxy | **`401`** |
| `GET /terminal/ws?arg=…&code=X` | yes | Go dashboard, checks its own cookie | `101` — socket opens |

The socket opened, ttyd closed it because the client never got its token, the client retried, forever.

## Where #6489 fits

#6489 didn't cause this, but it changed the symptom.

It fixed a sibling problem — the two sides also disagreed about the *signing key* — by having `entrypoint.sh` generate one and export it so both agree. That fixed the key and left the hive id split.

Its side effect: `/api/terminal/handoff` started succeeding on self-hosted hives instead of returning `503 terminal handoff requires terminal signing key and hive id`. So from Sep 4 to Sep 10 the terminal failed loudly with a message you could read, and after Sep 10 it failed quietly as this loop.

## The change

The proxy now finds the hive id the way the Go side does: `$HIVE_ID` if set, otherwise `/data/hive-id` (`HIVE_ID_FILE` for tests).

It's read when a request arrives, not at startup. The Go binary writes that file early in `main()`, but the proxy may be up first, and a startup read would cache an empty value forever. It's cached once it finds something, so steady state is one read.

The old, now-unused `const HIVE_ID` is removed, so nobody later reaches for the wrong one.

### This can't loosen anything

- The environment variable still wins, so hub-provisioned hives are byte-for-byte unchanged — a stale or planted id in the data directory can't redirect a hosted hive.
- The file is written by the hive's own Go binary, in its own private data directory.
- The value is only the id a cookie has to **match**. It's a requirement the cookie must satisfy, never permission to do anything.
- If the file is missing or unreadable, the id stays empty and the proxy denies — exactly what it does today.

## Tests

Every existing terminal test in `src/proxy/` sets `HIVE_ID`, because that's what a hosted hive looks like. The self-hosted shape — cookie minted, `HIVE_ID` missing — had no test at all, which is why the one setup where the halves disagree was the one setup nothing exercised.

`src/proxy/terminal_standalone_hive_id.test.js` starts the real `server.js` with `HIVE_ID` **deleted** from the environment (deleted, not empty, so a stray value in a developer's shell can't make it pass for the wrong reason).

Against the unfixed proxy it fails with exactly the reported symptom:

```
AssertionError: GET /terminal/token with a valid assertion for THIS hive must reach ttyd.
A 401 here is the ttyd reconnect loop: …
  actual: 401
  expected: 200
```

It also checks that the new lookup can't become a hole:

- a cookie for **another hive** → refused (HTTP and websocket)
- a tampered signature → refused
- no cookie at all → refused
- a valid **read-only** cookie → `403`, not quietly upgraded to a shell
- an **expired** cookie → refused
- `HIVE_ID` **beats** the file — the file is a fallback, never an override
- neither available → denied, same as before

`npm test` in `src/proxy/` is green, all suites.

## Checked on the affected host

- `GET /api/status` returns `"hiveId":"hive-wild-mole"` — a made-up name, which only happens when `HIVE_ID` is absent.
- The entrypoint logs `auto-provisioned a per-instance terminal signing key for this standalone spoke (#6489)` on every boot, which only prints when the hive isn't hub-provisioned. So the key halves already agreed and the hive id was the last piece.

## One loose end, not fixed here

`/terminal/ws?…&code=X` goes to the Go dashboard instead of to ttyd only because a **spent** code is still sitting in the query string ttyd copies from `window.location.search`.

It works — the Go dashboard's check is the stronger of the two — but a websocket taking a different route through the stack than every other terminal request, because of a dead credential, deserves its own look. Noted in #7043 rather than folded in here.


clubanderson (approved):
Reviewed: lazy hive-id resolution from HIVE_ID_FILE with env precedence and fail-closed fallback; test included. LGTM.

Danathar:
Follow-up after merge — I went back and pinned **when** this broke, which the original write-up didn't answer. Both this PR's description and #7043 are updated (code unchanged).

**It broke on 2026-09-04 in #6040** (`79d6be3c`), not #6489.

Before #6040, self-hosted hives weren't checked by the Node proxy at all — `if (isHosted)` let them fall straight through to ttyd, which did its own HTTP Basic auth. Basic auth rides on every request the browser makes, so the page, the token fetch and the websocket were all covered.

#6040 changed that to `if (isHosted || DASHBOARD_TOKEN)` and, in the same commit, removed the old `?token=` credential from the websocket path. From then on a self-hosted hive's only possible credential was the terminal-assertion cookie — which never verified, because the proxy had no hive id to check it against.

#6489 didn't cause it, but it changed the symptom: once `/api/terminal/handoff` started succeeding, the failure moved from a `503` you could read to the silent reconnect loop.

Measured on the affected host, same machine and hive, from the gateway access log:

| date | `GET /terminal/token` | |
|---|---|---|
| Aug 29 – Sep 5 | **200** ×114 | worked |
| Sep 5, 22:16 | **200** | last time it worked |
| Sep 15, 09:06 | **401** | first failure |
| Sep 15 | **401** ×11,174 | one session of the loop |

Nobody opened a terminal between Sep 6 and Sep 14, so the host can't narrow the window further — but #6040 is the only change in that window that touches this path.

Practical upshot: every self-hosted hive has had a broken **▶ terminal** for eleven days, and `v4` shipped releases in that window. Might be worth a backport if any of those lines are still live.

PR #7038, opened 2026-09-15
dashboard: read the coverage badge from the hive's own repo, not the flagship gist

gregoryhunt:
## Problem

The coverage percentage on the ci-maintainer card (also the ACMM advisor's coverage-floor input) comes from `HIVE_COVERAGE_BADGE_URL`, with a hard-coded fallback to the hivecommons/hive maintainer's public gist. A hive that never set the variable displayed **hivecommons/hive's** coverage as its own. When that gist began returning `0%`, every such hive dropped to 0 at once — that is how this surfaced on a standalone hive managing a private repo whose real coverage badge reads 98.5%.

There was no way for that hive to point at its own number: the repo is private, the collector fetches unauthenticated, and octocov publishes SVG, not shields JSON.

## Change

- **`repo://<ref>/<path>`** form for `HIVE_COVERAGE_BADGE_URL`: the badge file is read from the primary repo through the GitHub App client (`GetFileContentRef`), so it works for private repos. `repo://badges/coverage.svg` is octocov's default layout.
- **SVG badges parse**: the first percentage in the body is used (fractions truncated), so octocov / shields.io SVGs work directly; shields JSON still reads `message`.
- **Flagship gist gated**: the built-in gist is the default only for `project.org` `hivecommons`/`kubestellar`, the same gate `collectOutreach` uses for adopters/ACMM. Other hives configure a source or honestly show 0 — never another project's number.
- `collectCoverage` takes the collector's context (App fetch and HTTP GET are cancellable with the loop).

## Tests

- `TestMetricsCollector_CollectCoverage_RepoScheme`: `repo://badges/coverage.svg` hits `/repos/myorg/repo1/contents/coverage.svg?ref=badges` via `NewClientForTest` and yields 85.
- `TestMetricsCollector_CollectCoverage_SVGBadge`: an SVG badge over HTTP yields 98 from `98.5%`.
- `TestMetricsCollector_CollectCoverage_RepoScheme_NoClient`: no client / malformed `repo://` → 0, no HTTP attempt.
- `TestParseCoverageBadge`: shields JSON, fractional, 0%, non-numeric, SVG, >100% rejected, empty.
- Existing badge tests updated for the context parameter; `go test ./pkg/dashboard/ ./cmd/hive/` green.

Docs: `env-vars.md` and `operator-reference.md` rows updated; changelog fragment `changed-coverage-badge-managed-repo.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #7048, opened 2026-09-15
dashboard: read the coverage badge's text, not its markup

gregoryhunt:
Follow-up to #7038. The SVG path scanned the raw badge bytes for the first percentage. A shields-style badge — the shape octocov publishes — opens with `<linearGradient x2="0" y2="100%">` before any `<text>` node, so the gradient's extent was matched: a repo whose badge reads 98.5% showed **100** on the dashboard (observed on the first hive to use `repo://badges/coverage.svg`).

**Fix:** strip tags, and with them every attribute, before the scan, so only text nodes are read. JSON badges are unchanged.

**Tests:** `TestParseCoverageBadge` and the two SVG collector tests now use the real octocov badge markup (gradient first) and expect 98; added a `0%` text case and a badge whose text carries no number (rejected). `go test ./pkg/dashboard/` green.

Changelog fragment: `fixed-coverage-badge-svg-gradient.md`.


clubanderson (approved):
Strips markup before percentage scan — fixes gradient y2="100%" false match; test covers shields SVG. LGTM.

Issue #7109, opened 2026-09-15
Cadence tooltip and agent card disagree on next kick: tooltip anchors on now, card on last kick


clubanderson:
## Summary

The cadence matrix tooltip and the agent card disagree about when an agent is next kicked. On the console spoke at 3:38 PM EDT, scanner showed:

| surface | "next" |
|---|---|
| cadence table tooltip | `1h — next: 9/15 4:37 PM EDT` |
| agent card | `NEXT KICK 9/15 3:38 PM EDT` · `next kick in 58s` |

Both describe the same agent, same cadence, at the same moment, and they are ~59 minutes apart. The card is the correct one: scanner's last kick was 2:38 PM and its surge cadence is `1h`, so the next kick is 3:38 PM — and it did fire within the minute.

## Cause

The two surfaces compute "next" from different anchors, and only one of them accounts for the last kick.

The agent card anchors on `LastKick` for interval cadences (`src/pkg/dashboard/status_builder.go:1284`):

```go
func computeNextKickFromCadence(lastKick *time.Time, cadence config.Cadence) string {
	...
	base := time.Now()
	if lastKick != nil && cadence.Mode() == config.CadenceModeInterval {
		base = *lastKick
	}
	next, ok := cadence.NextAfter(base)
```

The cadence tooltip anchors on `time.Now()` and never consults `LastKick` (`src/pkg/dashboard/status_builder.go:1886`):

```go
func cadenceTooltip(c config.Cadence) string {
	...
	next, ok := c.NextAfter(time.Now())
	...
	return c.HumanSummary() + " — next: " + formatHumanTime(next)
}
```

For interval cadences `NextAfter` is simply `after.Add(d)` (`src/pkg/config/cadence.go:204-209`), so the tooltip renders **now + interval** rather than **last kick + interval**. With a 1h cadence read at 3:37 PM that is 4:37 PM, which is what the screenshot shows.

The error equals the time already elapsed since the last kick, so the tooltip is at its most wrong right before a kick fires — exactly when someone is most likely to hover it to ask "when does this run?". It always reads late, never early.

## Scope

**Interval cadences only.** For cron cadences both paths pass `time.Now()` into `schedule.Next(after)` (the `if` above only rebases for `CadenceModeInterval`), so they already agree. Any agent on an interval cadence is affected — on this spoke that is scanner (`1h`), ci-maintainer (`3h`), security (`1h`), quality, telemetry and operations.

## Suggested fix

[...]

PR #7114, opened 2026-09-15
🐛 Cadence matrix follows status.agents so a running out-of-pack agent keeps its row (#7112)

### LINKED ISSUE #7112: Cadence matrix still hides a running out-of-pack agent: acmmPackAgents is the raw roster, not the effective set #7110 used
## Summary

#7110 fixed the ACMM pack filter for the sidebar and agent cards, but the **cadence matrix still re-applies the same raw pack roster** with no "active outside pack" escape hatch. A running, enabled agent outside the level's pack is published by the API and now renders its nav item and card correctly — but its **cadence row is still dropped from the Governor matrix**.

This is the same defect class as #7099, in a second consumer that the fix did not reach.

## Mechanism

`renderCadenceMatrix` filters rows by `acmmPackAgents` (`src/pkg/dashboard/static/index.html:7580-7584`):

```js
const packAgentNames = (window._lastStatus && window._lastStatus.acmmPackAgents) || null;
const packSet = packAgentNames ? new Set(packAgentNames) : null;
...
const matrixNames = matrix.map(r => r.agent).filter(n => !packSet || packSet.has(n));
```

And the server publishes `acmmPackAgents` as the **raw roster**, not the effective set it actually used (`src/pkg/dashboard/status_builder.go`, `buildACMMPackAgents`):

```go
func buildACMMPackAgents(cfg *config.Config) []string {
	level := detectACMMLevel(cfg)
	pack, err := config.ACMMPackByLevel(level)
	if err != nil {
		return nil
	}
	names := make([]string, 0, len(pack.Agents))
	for _, a := range pack.Agents {
		if !a.Hidden {
			names = append(names, a.Name)
		}
	}
	return names
}
```

It never consults `activeOutsidePack` and never sees running state, so it cannot contain a custom or explicitly-enabled agent running beside the pack — while `status.agents`, built by `buildAgentsWithHidden` *with* that exception, does.

## Effect

For the `review` agent from #7099 (running, L5, outside the pack):

| surface | before #7110 | after #7110 |
|---|---|---|
| sidebar nav item | hidden | **visible** |
| agent card | hidden | **visible** |
| cadence matrix row | hidden | **still hidden** |

So the operator can now reach the agent's terminal and Login button, but the Governor matrix still shows no cadence for it — there is no way to see or reason about when it is kicked, and it is silently absent from the table rather than shown as unscheduled.

It is also invisible to the `hiddenAgents` diagnostic (#6581/#6652) for the same reason as #7099: the suppression happens client-side, so the server truthfully records nothing hidden.

This interacts with #7111, which just corrected the matrix tooltip's "next kick" anchoring — that fix is correct but cannot be seen for any agent whose row is filtered out here.

## Suggested fix

Prefer the structural option (option 3 from #7099), because it fixes every present and future consumer at once rather than this one call site:

**Publish the effective set.** Have `buildACMMPackAgents` return the set the server actually used — i.e. apply the same `activeOutsidePack` exception as `buildAgentsWithHidden` — so `acmmPackAgents` can never disagree with `status.agents` by construction. Every client consumer then becomes correct without knowing the rule.

This requires threading the same `cfg`/process state `activeOutsidePack` needs into `buildACMMPackAgents`, which currently takes only `cfg`.

Alternatively (narrower, same spirit as #7110): drop the `packSet` filter from `renderCadenceMatrix` and let the matrix rows follow `status.agents`, which is already authoritative.

Please do **not** fix this by copying the `activeOutsidePack` rule into a third place — that is the duplication that produced #7099 and #7109.

## Acceptance criteria

- [ ] A running/enabled agent outside the pack roster keeps its cadence-matrix row.
- [ ] Genuinely pack-inactive "ghost" agents are still suppressed.
- [ ] A regression test pins the invariant that `acmmPackAgents` (or whatever the matrix filters on) never omits an agent present in `status.agents`.
- [ ] No third copy of the pack-exception rule.

clubanderson:
## Summary

#7110 stopped the dashboard hiding a running agent the API deliberately surfaces — running/enabled but outside the ACMM level's pack roster (e.g. a custom `review` agent at L5) — for the **sidebar nav item and card**. But the Governor **cadence matrix** still re-applied the raw pack roster with no escape hatch, so that agent's cadence row was still dropped.

`renderCadenceMatrix` (`static/index.html`) filtered rows against `acmmPackAgents`, published by `buildACMMPackAgents` (`status_builder.go`), which iterates `pack.Agents`, skips `Hidden`, and **never consults `activeOutsidePack` or any running state**. So the roster cannot contain a running out-of-pack agent — while `status.agents`, built by `buildAgentsWithHidden` **with** that exception, does. Net effect: the operator could reach the agent's terminal but the matrix showed no cadence for it — silently absent, and invisible to the `hiddenAgents` diagnostic (#6581/#6652) because the suppression was client-side. It also masked the #7109/#7111 tooltip fix for any filtered-out agent.

## Approach chosen: the #7110-style fix (matrix follows `status.agents`)

I first investigated the preferred structural option 3 (make `buildACMMPackAgents` publish the *effective* set). **It is not cleanly workable here, and would introduce a new bug**, because `acmmPackAgents` is not a single-consumer channel:

- **`_packAgentNote` (`index.html`) is a genuine raw-roster consumer.** It warns, when deleting an agent, that *"X is part of this hive's ACMM pack … no restart or ACMM level change will bring it back."* Redefining `acmmPackAgents` to include active out-of-pack agents (like `review`) would make that dialog falsely claim `review` is a pack member — a real UI regression. The premise "every consumer becomes correct without knowing the rule" does not hold for this consumer.
- **`acmmPackAgents` is fed from two server sources.** Besides the status builder, the apply-pack endpoints (`api_packs.go:557`, `:710`) return the **raw** roster as `packAgents`, which the client writes straight into `window._lastStatus.acmmPackAgents` (`index.html:13611`, `:17305`, `:17345`). Option 3 in `buildACMMPackAgents` alone would leave those endpoints publishing the raw roster (re-dropping `review` right after a pack apply until the next status), so a full option 3 would also have to thread process state into those handlers — expanding the blast radius while still breaking `_packAgentNote`.

So I took the explicitly-sanctioned fallback, which is exactly what #7110 did for the cards/nav: **drop the redundant pack filter and let the matrix follow `status.agents`**, the set the server already gated with the pack rule *and* its `activeOutsidePack` escape hatch. This adds **no third copy** of the exception rule, keeps `acmmPackAgents` meaning intact for `_packAgentNote` and the apply-pack flow, and still suppresses genuinely pack-inactive "ghost" agents because they never appear in `status.agents`.

`renderCadenceMatrix` already loaded `status.agents` (`window._lastAgents`) for its sidebar ordering, so the change is one line of filter logic plus a comment.

### Every `acmmPackAgents` consumer found (repo-wide grep)

| location | role | affected? |
|---|---|---|
| `renderCadenceMatrix` (`index.html:7580`) | matrix row filter | **fixed** — now follows `status.agents` |
| `_packAgentNote` (`index.html:23042`) | delete-confirm "is a pack agent" note | unchanged — still wants the raw roster (why option 3 was rejected) |
| `applyACMMVisibility` via `index.html:4737`, `:16904` | param is **vestigial** after #7110 (no longer used to hide) | unchanged |
| `updateACMMBadge` (`index.html:13733`) | passes through to `applyACMMVisibility` (unused) | unchanged |
| `window._lastStatus.acmmPackAgents = …` (`index.html:13611`, `:17305`, `:17345`) | writes raw roster from apply-pack responses | unchanged |
| producers: `status_builder.go:298`, `server.go:1765` (`buildACMMPackAgents`); `api_packs.go:557`, `:710` | publish raw roster | unchanged |

## Tests

Following the #7110 pattern (a Go server-contract test + a node-executed JS test):

- **`TestRenderCadenceMatrixKeepsOutOfPackRunningRow7112`** — *executes* `renderCadenceMatrix` under node against a fixture where `status.agents = {supervisor, review}`, `acmmPackAgents = {supervisor}` (raw roster), and the cadence matrix has rows for `supervisor`, `review` and a pack-inactive ghost `linter`. Pins all three invariants: the matrix never omits a `status.agents` member, the out-of-pack running `review` keeps its row, and the ghost `linter` stays suppressed. Skips loudly if node is absent.
- **`TestCadenceMatrixFollowsStatusAgents_7112`** — pins the server-side contract the JS filter relies on: `review` (running, out-of-pack) is in `status.agents`; the ghost `linter` (out-of-pack, disabled) is not; and the matrix's visible set (`buildCadenceMatrix` ∩ `status.agents`) keeps `review` and drops `linter`.

`go build ./...`, `go vet ./pkg/dashboard/`, full `go test ./pkg/dashboard/` (52s, all pass), and the Dashboard Lint inline-JS checker (`.github/scripts/check-inline-js.js`) all pass.

### Mutation testing

1. **Reinstate the unconditional `acmmPackAgents` packSet filter** in `renderCadenceMatrix` → `TestRenderCadenceMatrixKeepsOutOfPackRunningRow7112` **FAILS**: `FAIL out-of-pack running agent keeps its cadence row` (reproduces #7112).
2. **Remove the `status.agents` filter entirely** (`matrixNames = matrix.map(r => r.agent)`) → same test **FAILS**: `FAIL pack-inactive ghost agent row stays suppressed`.
3. **`activeOutsidePack` always returns `false`** → `TestCadenceMatrixFollowsStatusAgents_7112` **FAILS**: `running out-of-pack agent 'review' must be in status.agents; got [supervisor]`.

Each mutation was reverted and the suite reconfirmed green.

Fixes #7112


clubanderson:
## Follow-up: restore the fail-open bypass (empty `_lastAgents`)

Addressing review feedback: the first commit swapped the raw-roster `packSet` filter for one keyed on `status.agents` but dropped the old `!packSet ||` bypass, so an empty `window._lastAgents` would blank the **entire** cadence matrix.

**Reachability — confirmed reachable.**
- The SSE `agent-status` handler assigns `window._lastAgents = payload.agents || []` from a **delta** (`static/index.html`, ~line 16891) and updates `_lastStatus.agents` but **not** `_lastStatus.cadenceMatrix`. A delta carrying an empty/absent roster leaves `_lastAgents = []` while `_lastStatus.cadenceMatrix` still holds rows from the last full status.
- `renderGovernor(lastData.governor, lastData.cadenceMatrix || [], lastData)` is then invoked from `togglePin` (~15674), `switchModel` (~15641), `switchBackend` (~15572), `_saveSidebarLayout` (~4732) and the config-save path (~23359), each reading `window._lastAgents`. After such a delta, `statusAgentSet` is empty and the matrix renders **zero rows**.
- The pre-#7112 filter failed open in exactly this "guard set absent" case (`!packSet ||`).

**Fix.** Restore an equivalent guard: `const filterByStatus = statusAgentSet.size > 0;` and `filter(n => !filterByStatus || statusAgentSet.has(n))`. With no authoritative set, show every row (pre-#7112 behaviour); when the set is populated — the normal case, including the out-of-pack `review` agent — filtering proceeds and pack-inactive ghosts stay suppressed. Failing open (a stale row) is a far smaller harm than blanking the operator's whole matrix.

**Regression test.** `TestRenderCadenceMatrixFailsOpenWhenAgentsUnpopulated7112` executes `renderCadenceMatrix` under node with an empty `window._lastAgents` and a non-empty matrix, asserting all three rows still render.

**Mutation.** Break the guard (`const filterByStatus = true;`, i.e. unconditional filter) → the new test FAILS:
```
FAIL fail-open renders supervisor row when _lastAgents empty
FAIL fail-open renders review row when _lastAgents empty
FAIL fail-open renders linter row when _lastAgents empty
```
Reverted → green. The original `TestRenderCadenceMatrixKeepsOutOfPackRunningRow7112` still passes under that mutation (its set is populated), confirming the guard is specific to the empty case.

`go build ./...`, `go vet ./pkg/dashboard/`, full `go test ./pkg/dashboard/` (49s, all pass), and the Dashboard Lint inline-JS checker all pass. Rebased on latest `origin/v4`; DCO author and `Signed-off-by` both `Andrew Anderson <andy@clubanderson.com>` on both commits.

Issue #7175, opened 2026-09-16
✨ feature: forward OPENROUTER_API_KEY for Goose backend


mendezr:
## Use case

I use Hive contributions with:

    Hive → Goose → OpenRouter → selected model

This works with Goose when run directly, but the containerized `just contribute-hive` workflow currently does not forward `OPENROUTER_API_KEY` into the contributor container.

As a result, Goose can be configured with the OpenRouter provider and model, but the container fails because the OpenRouter credential is unavailable inside the container.

This is useful for contributors who already use Goose and want to use different models available through OpenRouter without switching to the LiteLLM + Claude Code workflow.

## Current behavior

The contributor relay forwards several provider-specific environment variables into the container, including:

    OPENAI_API_KEY
    GOOSE_API_KEY
    GOOSE_PROVIDER
    GOOSE_MODEL
    HIVE_LITELLM_ENDPOINT
    HIVE_LITELLM_API_KEY
    ...

but does not currently include:

    OPENROUTER_API_KEY

When using Goose with OpenRouter, this results in the credential being unavailable inside the contributor container.

The observed error is:

    Configuration value not found: OPENROUTER_API_KEY

## Proposed solution

Forward `OPENROUTER_API_KEY` alongside the other provider environment variables in the containerized `contribute-hive` workflow.

For example, my local proof-of-concept adds `OPENROUTER_API_KEY` to the existing provider environment allowlist:

    - for name in ANTHROPIC_API_KEY OPENAI_API_KEY OPENAI_HOST OPENAI_BASE_PATH GOOGLE_API_KEY GOOSE_API_KEY GOOSE_PROVIDER GOOSE_MODEL BOBSHELL_API_KEY HIVE_LITELLM_ENDPOINT HIVE_LITELLM_API_KEY KILO_AUTH_CONTENT KILO_CONFIG_CONTENT KILO_API_KEY KILO_ORG_ID META_API_KEY; do
    + for name in ANTHROPIC_API_KEY OPENROUTER_API_KEY OPENAI_API_KEY OPENAI_HOST OPENAI_BASE_PATH GOOGLE_API_KEY GOOSE_API_KEY GOOSE_PROVIDER GOOSE_MODEL BOBSHELL_API_KEY HIVE_LITELLM_ENDPOINT HIVE_LITELLM_API_KEY KILO_AUTH_CONTENT KILO_CONFIG_CONTENT KILO_API_KEY KILO_ORG_ID META_API_KEY; do

[...]

PR #7176, opened 2026-09-16
✨ feat: forward OPENROUTER_API_KEY into the contributor container

clubanderson:
Fixes the gap reported in #7175 by @mendezr.

## Why it failed

`just contribute-hive` deliberately does **not** hand the container the whole environment. It forwards an explicit allowlist of provider credentials, so an unconfined agent gets only what it needs and provider choice isn't made observable through unrelated secrets (#5039):

```
for name in ANTHROPIC_API_KEY OPENAI_API_KEY … META_API_KEY; do
  add_provider_env "$name"
done
```

`OPENROUTER_API_KEY` was simply absent. So Goose could be configured correctly — `GOOSE_PROVIDER=openrouter`, `GOOSE_MODEL=…` — and still fail inside the container with:

```
Configuration value not found: OPENROUTER_API_KEY
```

Worth noting *why* this looked inconsistent: the **`pi` backend already resolves `OPENROUTER_API_KEY`** per model through `bin/pi-backend.js` (`openrouter: ['OPENROUTER_API_KEY']`), which takes a different branch above this loop. OpenRouter therefore worked on that path and not for Goose — the credential was known to the codebase, just not to this allowlist.

I checked for a second copy of the list before changing it, since duplicated-and-drifted config has bitten this repo before (the `PKG_THRESHOLD` map is duplicated across two workflows). **This is the only allowlist of its kind in the tree** — `Justfile:1565` — so there is nothing to keep in sync.

## The change

One name added to the allowlist, in the placement @mendezr tested, plus a row in `src/docs/env-vars.md` next to the other forwarded provider vars.

**Only the variable *name* is forwarded.** `add_provider_env` emits `-e NAME`, and the container runtime resolves the value from the calling process — so the secret never appears in the runtime command's argv. And like every other entry, it's forwarded only when actually set, so nothing changes for contributors who don't use OpenRouter. The existing LiteLLM + Claude Code workflows are untouched, as the issue requested.

## Verification

`just --list` parses. Beyond that I replicated the loop against the **real `Justfile`** (parsing the live line rather than a copy, so the test can't drift from the source) and checked behaviour:

| Case | Expected | Result |
|---|---|---|
| `OPENROUTER_API_KEY` unset | not forwarded | ✅ |
| set | forwarded as `-e OPENROUTER_API_KEY` | ✅ |
| set | secret **value** absent from argv | ✅ |
| `ANTHROPIC_API_KEY`, `GOOSE_PROVIDER` | still forwarded | ✅ |

Mutation check: removing the entry again makes that check fail, reproducing the exact reported error — so the fix is load-bearing rather than incidental.

I kept this to the issue's stated scope and did not add a CI test for the allowlist, since that would mean new CI surface for a single-token list; happy to add one if maintainers would prefer it pinned.

Refs #7175

---
@mendezr — could you confirm this works against your setup? I've used `Refs` rather than `Closes` so this stays open until you've verified it on your end.


mendezr:
I tested this locally with my Goose + OpenRouter setup and confirmed it works.

For context, `hive-custom` is a small local Bash wrapper I use around `just contribute-hive` to load the required environment variables. It configures Hive to use the Goose backend, sets `GOOSE_PROVIDER=openrouter`, selects the model passed as an argument, sources my `OPENROUTER_API_KEY`, and then runs `just contribute-hive`.

I ran:

hive-custom deepseek/deepseek-v4-flash-0731

With this PR, `OPENROUTER_API_KEY` is successfully forwarded into the contributor container, and Goose starts with:

openrouter deepseek/deepseek-v4-flash-0731

The contributor agent then started successfully and received the assigned task.

So the OpenRouter API key forwarding works as expected from my setup. Thanks!

PR #7178, opened 2026-09-16
🐛 fix: stop nested packages from sharing a top-level package's coverage floor

clubanderson:
## What

`floor_key()` in both coverage gates keyed every non-`cmd` package by its **bare directory name**. That makes any nested package an alias for the top-level package sharing that name, so the two silently share a coverage floor.

## Why now

This blocks the v5 sync that #7143 needs.

v5 carries **both** `pkg/spoke` and `pkg/hub/spoke`; v4 carries only `pkg/spoke`. Measured on a trial merge of v4 into v5:

| package | coverage |
|---|---|
| `pkg/spoke` | 96.3% |
| `pkg/hub/spoke` | 90.7% |

Both keyed to `spoke`. v4's `[spoke]=95` — set by #7146 for `pkg/spoke` specifically — would therefore be enforced against `pkg/hub/spoke` too and fail it at 90.7%. That lands the breakage on v5, which is the branch #7143 needs green to evidence the GA bar.

The function already prefixes `cmd` packages to prevent precisely this (`cmd/hivectl` vs `pkg/hivectl`, #4701). This extends the same rule to `pkg`.

## No-op on v4

Verified by extracting `floor_key` from each file and running it against the real `go list ./pkg/... ./cmd/...` output:

- all 18 v4 floor keys still resolve to a package — **no orphans**
- no two v4 packages share a key — **no collisions**

So no floor changes value or target on v4; only the mechanism changes. Both workflow files were edited together per their KEEP IN SYNC note, and both parse.

Worth noting for the follow-up sync: v5's map has three keys that *are* nested — `mutation`, `webstatic`, `automerge` — which become `convergence/mutation`, `dashboard/webstatic`, `github/automerge`. Those renames belong in the v5 sync PR, where those keys actually exist; they are not present on v4.

Refs #7143


clubanderson:
## Correction: the stated CI consequence was wrong

I claimed that a naive `v4` → `v5` sync would **fail CI** because `pkg/hub/spoke` (90.7%) would be enforced against the `[spoke]=95` floor that #7146 set for `pkg/spoke`, since `floor_key()` keyed both by bare basename.

That consequence is wrong, and I want the record straight.

**The evidence that contradicted it:** PR #7182 is a naive sync carrying `[spoke]=95` onto `v5` with no `floor_key` fix, and `v5` has both packages. Its `coverage` check **passed**. Run `35116595631` shows `pkg/hub/spoke` tested at 90.6% — under the 95 floor — with no failure.

**Why:** the `coverage` job in `v2-tests.yml` has been scoring **zero packages** and passing unconditionally. It selects this module's packages with the pre-rename module path `github.com/kubestellar/hive`, while `src/go.mod` declares `github.com/hivecommons/hive`. Both filters match nothing, so `TOTAL` is empty and `pkg-states.txt` is empty, so the loop that reads `PKG_THRESHOLD` never executes its body. The floor map has not been consulted in this job at all.

I verified this by replaying the real shard artifacts from run `35113609994` through the gate script extracted verbatim from the workflow: current `v4` exits 0 with **0 packages scored**; with only the module path corrected it reports **91.9%** across **79 packages**.

That is filed as #7193 and fixed in #7194.

## What this does and does not change

- The `floor_key` change itself is **still correct**. Keying non-`cmd` packages by bare basename genuinely makes every nested package an alias for the top-level one sharing its name, and it was verified a strict no-op on `v4` (zero orphans, zero collisions against real `go list` output).
- What was overstated is the **timing**: this was a latent aliasing defect, not an imminent breakage. It could not have failed anything while the gate was blind.
- It becomes a real hazard the moment #7194 lands and the gate starts evaluating floors again. So the ordering matters: the `floor_key` fix needs to be in place before enforcement is restored, which it now is.

Apologies for asserting a consequence I had not observed. The underlying change stands; the justification I gave for it did not.

PR #7187, opened 2026-09-16
🌱 chore: top up v5 with v4 so the apt/cgo fix reaches the branch that keeps failing

clubanderson:
Tops up `v5` with `v4` (23 commits), so the GA-bar row #7143 tracks can finally be evidenced on the branch the failures actually occur on.

## Why this is needed

#7143 reports the apt-mirror / cgo-toolchain failure recurring **5x on v5**. The real fix — #7170 — landed on **v4 only**. `attempt_deadline_at` was absent from v5's `ci-install-tool.sh` entirely, so every failure #7143 cites was still live. Root cause was that `APT_DEADLINE_SECONDS` is an end-to-end ceiling, so an attempt that fails *by timing out* consumed the whole budget and the retry loop could never make a second attempt.

## The trap this merge had to clear

A naive merge builds cleanly but **breaks v5 CI**. The coverage gate keyed packages by bare directory name, and v5 carries both `pkg/spoke` (96.3%) and `pkg/hub/spoke` (90.7%) — both keying to `spoke`. v4's `[spoke]=95`, set by #7146 for `pkg/spoke` alone, would then be enforced against `pkg/hub/spoke` and fail it at 90.7%.

#7178 fixed the mechanism on v4 (keys are now the full path below `pkg/`). This PR carries it over and spells out the three v5 floors that point at nested packages, which would otherwise stop resolving and silently fall back to the 90 default:

- `mutation` → `convergence/mutation`
- `webstatic` → `dashboard/webstatic`
- `automerge` → `github/automerge`

Values unchanged. Verified against `go list` **on this branch**: every floor key resolves to a real package (no orphans) and no two packages share a key. Both workflow files were updated together per their KEEP IN SYNC note and both parse.

## Conflicts, and how they were resolved

**`src/cmd/hive/main.go`** — v5's side was 1 line (`wireSpokeSubsystems(...)`), v4's was 8667 lines defining ~120 symbols, because v5 split `main.go` apart. Took v5's side, but only after checking every one of those symbols still exists on v5. Exactly two did not:

- `providerBudgetSuppresses` — v5 moved it to `governor.ProviderBudgetSuppresses`; v5's test already calls the new one (the old name survives only inside a `t.Errorf` string). Nothing lost.
- `hiveIDFilePath` — see below.

Resolved the conflicted hunk in place rather than `git checkout --ours`, so #7181's `var hiveIDFilePath` change elsewhere in the file was not discarded.

**`src/docs/env-vars.md`** — took v4's three added rows (`OPENAI_HOST`, `OPENROUTER_API_KEY`, `OPENAI_BASE_PATH`) from #7176.

## One follow-on fix this merge required

#7181 changed `hiveIDFilePath` from `const` to `var` in `cmd/hive/main.go` so `loadOrGenerateHiveID`'s disk branches are testable. On v5 that symbol lives in `main_helpers.go` instead, still as a `const` — so #7181's incoming `hive_id_identity_test.go` would **not compile** on v5. Applied the same seam at v5's location. Value and runtime behaviour unchanged.

## Verification on the merged tree

- `go build ./...` — clean
- `go vet` on `cmd/hive`, `pkg/dashboard`, `pkg/spoke`, `pkg/hub/spoke`, `pkg/knowledge`, `pkg/github` — clean (this compiles the test binaries, which is what catches the `const`/`var` break above)
- synced tests pass: `TestLoadOrGenerateHiveID`, `TestCovD_CopilotAPIHost`, `TestValidateIssueTemplate*`, `TestBodyRuleStays*`
- coverage measured on this branch: `pkg/spoke` 96.3% (floor 95 ✓), `pkg/hub/spoke` 90.7% (default floor 90 ✓)
- `attempt_deadline_at` now present in `ci-install-tool.sh`

Note `pkg/hub/spoke` clears the default floor by only 0.7 points. Worth a ratchet entry once it has real headroom, but deliberately not touched here — this PR should not change any floor's effective value.

Refs #7143


clubanderson:
## Correction: the stated CI consequence was wrong

I claimed that a naive `v4` → `v5` sync would **fail CI** because `pkg/hub/spoke` (90.7%) would be enforced against the `[spoke]=95` floor that #7146 set for `pkg/spoke`, since `floor_key()` keyed both by bare basename.

That consequence is wrong, and I want the record straight.

**The evidence that contradicted it:** PR #7182 is a naive sync carrying `[spoke]=95` onto `v5` with no `floor_key` fix, and `v5` has both packages. Its `coverage` check **passed**. Run `35116595631` shows `pkg/hub/spoke` tested at 90.6% — under the 95 floor — with no failure.

**Why:** the `coverage` job in `v2-tests.yml` has been scoring **zero packages** and passing unconditionally. It selects this module's packages with the pre-rename module path `github.com/kubestellar/hive`, while `src/go.mod` declares `github.com/hivecommons/hive`. Both filters match nothing, so `TOTAL` is empty and `pkg-states.txt` is empty, so the loop that reads `PKG_THRESHOLD` never executes its body. The floor map has not been consulted in this job at all.

I verified this by replaying the real shard artifacts from run `35113609994` through the gate script extracted verbatim from the workflow: current `v4` exits 0 with **0 packages scored**; with only the module path corrected it reports **91.9%** across **79 packages**.

That is filed as #7193 and fixed in #7194.

## What this does and does not change

- The `floor_key` change itself is **still correct**. Keying non-`cmd` packages by bare basename genuinely makes every nested package an alias for the top-level one sharing its name, and it was verified a strict no-op on `v4` (zero orphans, zero collisions against real `go list` output).
- What was overstated is the **timing**: this was a latent aliasing defect, not an imminent breakage. It could not have failed anything while the gate was blind.
- It becomes a real hazard the moment #7194 lands and the gate starts evaluating floors again. So the ordering matters: the `floor_key` fix needs to be in place before enforcement is restored, which it now is.

Apologies for asserting a consequence I had not observed. The underlying change stands; the justification I gave for it did not.

Issue #7251, opened 2026-09-17
Health tab: Healthcheck Interval / Restart Cooldown are dead config — consolidate into Agent Watchdog


clubanderson:
## Problem

Settings → Health shows two overlapping liveness systems, and one of them is dead:

| Field | What the tooltip promises | What runs |
|---|---|---|
| **Healthcheck Interval (sec)** | "How often the governor checks each agent's tmux session for liveness" | **Nothing.** `Governor.Health.HealthcheckInterval` is parsed (`config.go:2942`), defaulted (`:4881`), validated (`dashboard/validation.go:264`), persisted (`api_governor.go:729`, `snapshot/state.go:67`) and rendered — but no runtime loop reads it. |
| **Restart Cooldown (sec)** | "Minimum time between restart attempts for the same agent" | **Nothing.** Same lifecycle as above. Actual restart cooldowns are hard-coded in `pkg/agent/manager.go` (`tokenRestartCooldownSec=60`, `tlsErrorRestartCooldownSec=120`) and the watchdog uses its own backoff ladder. |
| **Agent Watchdog** (RFC #4665) | Probe interval, mode, give-up, healthy reset, backoff ladder | The real thing — `pkg/watchdog/reconciler.go` self-gates on `ProbeInterval` from the governor tick. |

An operator lowering Healthcheck Interval to 60 s expecting faster detection gets no change, while the effective cadence lives further down the tab under a different name. **Model Lock** also sits in this block although it is a budget behaviour, not a health one.

## Proposal

1. **Remove** `governor.health.healthcheck_interval` and `governor.health.restart_cooldown` from the UI, `config.HealthConfig`, `api_governor.go` (GET/PUT), `validation.go`, `snapshot/state.go`, `cmd/hive/main.go` overrides, and `deploy/hive*.yaml` samples.
2. **Back-compat alias**: if `governor.watchdog.probe_interval` is unset and `governor.health.healthcheck_interval` is set, seed the watchdog probe interval from it and log a deprecation warning once. `restart_cooldown` is dropped outright — the watchdog ladder (1→2→4→8→16 m, reset after healthy window) supersedes it.
3. **Move Model Lock** to Settings → Budget beside the downgrade behaviour it disables.
4. Health tab order becomes **Agent Watchdog → Escalation Breaker**.
5. Guard: a `pkg/config` test asserting `HealthConfig` has no fields other than `ModelLock` (or none, if it moves) so dead knobs can't creep back.

Applies to v4 and v5.

PR #7310, opened 2026-09-17
Health tab: watchdog activity count + 30-day sparkline + per-agent liveness so Observe→Heal is a data decision (#7254)

### LINKED ISSUE #7254: Health tab: watchdog activity count + 30-day sparkline + per-agent liveness so Observe→Heal is a data decision
## Problem

Settings → Health → Agent Watchdog tells the operator to "review `watchdog-*-observed` entries in the Audit Log, then switch to Heal" — but gives no signal about whether any exist, how many, or how that has trended. Promotion to Heal is meant to be a data-driven decision (RFC #4665, `pkg/watchdog/settings.go:101-107`); today the data lives only in `/data/audit.jsonl` behind a manual search.

Verified on `hosted-kubestellar-console-4vkt` (v4.39.x): 6 h in Observe, 6 `watchdog: liveness transition` log lines, **0** `watchdog-*` audit entries out of 28,775 — nothing on the page conveys "0 would-have-acted events in 6 h".

## Proposal

Add a **Watchdog activity** strip directly under the mode banner:

1. **Number** — total `watchdog-*` audit actions in the trailing **30 days**, broken down by action (`restart`, `crashloop-pause`, `giveup`, `healthy-reset`, and their `-observed` twins). In Observe mode the headline is "N actions the watchdog WOULD have taken in 30 d"; in Heal it is "N actions taken".
2. **Sparkline** — 30-day daily histogram of those actions, reusing the existing `sparkline(name, color)` / `.spark-row` helper. Colour by severity: observed = muted, restart = amber, pause/giveup = red.
3. **Per-agent liveness** — current watchdog state per agent (ready / stuck-overlay / dead-pane / crash-loop) with time in state, from the reconciler's condition set.
4. **Promotion hint** — when mode is Observe and there are 0 pause/giveup verdicts in 30 d (or the observed set looks sane), show "Safe to promote to Heal" next to the mode select; when Heal, show the last action time.

### Backend

- New `GET /api/watchdog/activity?days=30` → `{ total, byAction: {...}, daily: [{date, count, byAction}], agents: [{name, state, since}] }`.
- Source: `pkg/dashboard/audit.go` audit store filtered on `action` prefix `watchdog-`; keep it cheap by scanning from the tail with a 30-day time bound (audit.jsonl is append-only, ts-ordered).
- Liveness states from `pkg/watchdog/reconciler.go` — expose the last-known state map via the existing `Watchdog` handle the dashboard already holds for `/api/governor` mode changes.

### Tests

- Handler test with a synthetic audit log covering all action names, mixed dates, and entries older than 30 d (excluded).
- Static-HTML guard that the Health tab renders the activity strip element ids.

Related: #7251 (consolidate Healthcheck fields into the Watchdog block — this strip is where they should be replaced). Applies to v4 and v5.

Danathar:
## Summary

The Health tab's **Agent Watchdog** section tells you to "review the `watchdog-*-observed` entries in the Audit Log, then switch to Heal" — but never shows whether any exist, how many, or how that has trended. This PR puts that data on the page: a **Watchdog activity** strip directly under the mode banner with the 30-day count of actions the watchdog took (or, in Observe, *would have* taken), a per-day sparkline, the current liveness verdict for every agent it has probed, and a **Safe to promote to Heal** / **Not yet safe** hint beside the mode select. A hive that has sat in Observe with nothing to act on now shows **0**, which is an answer rather than an absence.

Closes #7254.

### Backend — `GET /api/watchdog/activity?days=30`

- New handler in `pkg/dashboard/api_watchdog_activity.go`. Gated at read-write role — the same tier as `GET /api/audit`, which it is derived from.
- **Source:** the on-disk audit log (`/data/audit.jsonl` plus lumberjack's rotated/compressed backups), not the 500-entry in-memory ring, so the window survives pod restarts. Falls back to the ring only when no file exists (no `/data` volume). Lines are pre-filtered on the serialized `"action":"watchdog-` fragment before JSON decoding — on the hive in the issue that means decoding 0 of 28,775 lines rather than all of them.
- **Response:** `days` / `since` (window actually applied), `mode` / `acting` (the mode **in force**, kill switch applied), `total` / `taken` / `observed`, `byAction` keyed by the audit action minus the `watchdog-` prefix (`restart`, `crashloop-pause`, `giveup`, `healthy-reset`, and their `-observed` twins), `daily` (one zero-filled bucket per UTC calendar day, oldest first), `agents` (per-agent liveness), `lastAction` / `lastActionAt`, and in observe mode a `promotion` hint.
- **Per-agent liveness** comes from the condition set the reconciler already publishes onto each agent (`Fleet.SetConditions` → `AgentProcess.WatchdogConditions`): the `Ready` condition's reason is the pane class (`ready`, `stuck-overlay`, `shell-prompt`, `no-output`, `no-session`, `auth-required`, `unknown`), its `lastTransitionTime` is "since". An agent the watchdog paused (`PausedTrigger == watchdog-crashloop`) reports `crash-loop` since the pause, outranking the pane. Agents never swept, and agents disabled in config, are omitted rather than listed as healthy.
- **Promotion rule** (from the issue): safe iff the window holds no crash-loop pause and no give-up verdict, taken *or* observed. The reason string states how many would-be restarts the verdict rests on, so "safe" on a window with zero events reads as "nothing happened".
- `days` defaults to 30 and is clamped to the audit log's 90-day retention; non-positive / non-integer is a 400.
- `pkg/watchdog` now exports its audit vocabulary (`AuditActionPrefix`, `AuditObservedSuffix`, `AuditActionRestart/Pause/GiveUp/HealthyReset`) so the dashboard classifies entries by the constants the reconciler writes rather than by string literals of its own.
- `AuditLog.OutputActionsSince` is refactored onto a shared scanner (`actionsSince`) that the new `ActionsWithPrefixSince` also uses; behaviour of the existing caller is pinned by a test.

### Frontend (Health tab)

- `renderWatchdogActivityShell()` renders the strip between the mode banners and the mode select; `loadWatchdogActivity()` fetches when the Health tab opens and `fillWatchdogActivity()` paints in place (the `loadConvergenceSoak` pattern) so a late response never repaints an unsaved mode edit. Cached 60 s; invalidated when watchdog settings are saved so the headline follows the new mode.
- Headline: "**N** actions the watchdog WOULD have taken in 30 d" (observe) / "**N** actions taken by the watchdog in 30 d" (heal) / "Watchdog is off" — plus the per-action breakdown.
- Sparkline: the daily histogram through the existing `axisSparkSvg` (the fact/cost 30-day pattern, with real axes). Deliberately **not** `sparkSvg`, which carries the last non-zero value forward through zero days — right for a gauge, wrong for an event count where a zero day is the news. Coloured by the worst thing in the window: observed-only muted, taken restart amber, pause/give-up red.
- Per-agent liveness list: name, state (colour-coded), time in state, `auth` / `producing` axes, `(paused)` marker.
- Promotion hint under the mode select: observe → "Safe to promote to Heal — …" / "Not yet safe to promote — N crash-loop pause / give-up verdict(s) …"; heal → "Last action: restart 3h ago" or "No action taken in 30 d."
- Passes `.github/scripts/check-inline-js.js` (syntax + no-undef).

### Docs / spec

- `dashboard/openapi.json`: new path (route-parity guard) with a typed schema; the endpoint is added to `TestOpenAPISchemaFieldsExistOnGoTypes` so the spec cannot invent fields.
- `src/docs/api-reference.md` row (route registered at the tail of `RegisterAPI` so no existing `pkg/<file>:<line>` citation shifts — `check-api-reference-citations.sh` verifies clean).
- `src/docs/agent-watchdog.md`: new "Watchdog activity strip" section.
- `changelog.d/added-7254-watchdog-activity-strip.md`.

## Tests

- `api_watchdog_activity_test.go`: synthetic audit log covering every action name (taken and observed), interleaved noise (including `config_governor_watchdog`, which contains the word but not the prefix), entries older than 30 d, and a malformed timestamp; asserts totals, taken/observed split, `byAction` keys, 30 ascending zero-filled buckets that sum to the total, `since`, `lastAction`, and the promotion verdict. Also: `days` narrowing / clamping / 400s; heal-mode headline and no hint; kill-switch downgrade reports observe; ring fallback without a file; the zero-events case; role gate fails closed; route registered on the mux; `watchdogAgentLiveness` derivation (ready / stuck-overlay / crash-loop / operator-pause / unswept omitted / disabled omitted / live config wins); `ActionsWithPrefixSince`; `OutputActionsSince` unchanged by the refactor.
- `watchdog_ui_test.go`: static-HTML guards — strip placed between the banners and the mode select, every element id declared, fetched on Health open from the right endpoint with the 30-day window, save path invalidates the cache, honest observe/heal wording, severity rule, `axisSparkSvg` usage, and no undefined callees.

Verified locally: `go test ./pkg/dashboard/` passes except two pre-existing `backends.conf` tests that fail only because this sandbox mounts a read-only, locally modified `config/backends.conf` (unrelated; untouched by this PR). `go test ./pkg/watchdog/` passes. One `pkg/agent` test (`TestDeadSessionRecoveryOwnershipTransfer`) fails in this sandbox on pristine `upstream/v4` too and is green in v4 CI — unrelated.

## Notes for reviewers

- The observed-vs-taken distinction lives in the audit action name (`-observed` suffix), so the count can never mistake "would have paused" for "paused" — the same honesty rule the existing banners follow.
- An observed pause/give-up is treated as **severe** (red, blocks the hint) even though nothing was done: it is the verdict that decides promotion, taken or not.
- Related: #7251 / #7263 already removed the dead Healthcheck fields this strip was meant to replace.


clubanderson (approved):
Reviewed: the new /api/watchdog/activity endpoint is correctly gated at read-write role (same tier as the audit log it derives from), days param is clamped, rotated backups are scanned, and the strip/liveness/promotion logic is covered by tests with OpenAPI parity maintained. LGTM.

Issue #7206, opened 2026-09-16
🌱 [ci] bake gcc/libc6-dev into the ARC runner image — root fix for the recurring apt-egress failure class


clubanderson:
## Summary

The apt-egress failure class (#6648 → #6870 → #6935 → #7124 → #7201) has been closed four times and recurred every time, because each fix so far has been a repo-side mitigation of a runner-image gap. The gap: **`gcc`/`libc6-dev` are not in the ARC runner image**, but every `-race` shard needs a C toolchain (`CGO_ENABLED=1`). So each shard installs ~100 packages from the Ubuntu mirrors, and any pod that has lost egress fails the run.

This issue tracks the one remediation that removes the dependency instead of softening it — the one `ci-install-tool.sh` itself ranks first.

## Why the existing mitigations cannot close this

They are all good, and all of them are downstream of the gap:

| Fix | What it did | Why it doesn't end the class |
|---|---|---|
| #6648 | fail fast, one clear annotation | turns a 2+ min hang into a fast failure — still a failure |
| #6870/#6935 | bounded retry with backoff | can't manufacture a route to the mirrors |
| #7009 | offline `.deb` cache | the entry is evicted (see below) |
| #7124 | prune dead-branch caches | dead-branch waste is now **0 GB** |

## Measurement (2026-09-16)

Today's two v4 failures (runs `35126475468`, `35122417180`) both logged:

```
Cache not found for input keys: hive-cgo-race-apt-Linux-X64-gcc-libc6-dev-v1
```

Actions cache state at the time: **9.99 GB / 10 GB, 364 entries.**

- buildkit blobs are **9.65 GB — 96.6%** of the quota
- **100%** of them sit on branches that still exist (`refs/heads/v4` alone: **7.96 GB**, 224 entries)
- **6.57 GB of blobs were created on 2026-09-16 alone** — 66% of the entire quota in one day

Both prune strategies were measured and neither is viable:

- **by last access:** 0.00 GB reclaimable at *any* threshold (1/7/14 days) — `cache-from: type=gha` touches all 276 entries every build, so they are permanently hot
- **by creation date:** 2.50 GB at >= 2 days, against 6.57 GB/day of new writes

At ~6.5 GB/day into a 10 GB ceiling, a few-tens-of-MB apt entry cannot survive. #7009 is sound engineering placed in an environment where it cannot win.

## Proposed fix

Bake `gcc` and `libc6-dev` into the ARC runner image used by the `hive` runner scale set.

[...]

Issue #6648, opened 2026-09-11
[ci] self-hosted runners lost egress to archive.ubuntu.com — every -race shard fails at 'Prepare cgo toolchain' since ~07:00Z 2026-09-11


clubanderson:
## What happened
Starting with v2 Tests run 34572490529 (07:01Z) every job scheduled on the `self-hosted,hive` runners (`hivecommons-hive-runners-lbz5b-*`, labels self-hosted,Linux,X64,openshift,hive) fails at **Prepare cgo toolchain for race tests** or **Install tmux**: the runner image has neither gcc nor tmux, and `apt-get update` cannot reach any Ubuntu mirror:

```
W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/noble/InRelease  Cannot initiate the connection to archive.ubuntu.com:80 (2620:2d:4002:1::101). - connect (101: Network is unreachable) ... Could not connect to archive.ubuntu.com:80 (185.125.190.81), connection timed out
```
IPv6 is unreachable and IPv4 port 80 times out — cluster egress, not the workflow. Runs at 06:10Z/06:14Z on the same runners were green. Affected: v4 pushes 34572842434, 34573649106; PRs #6638, #6644, #6647 and the post-merge runs of #6637/#6639.

## Mitigation applied
Repo variable `HIVE_RUNNER_LABELS` switched from `["self-hosted","hive"]` to `["ubuntu-latest"]` at 07:49Z so the queue keeps moving on GitHub-hosted runners (gcc/tmux preinstalled). **Revert once the runner cluster's egress is fixed.**

## Follow-ups
- [ ] Find/fix the egress path on the OpenShift runner cluster (port 80 to Ubuntu mirrors; or bake gcc+tmux into the runner image so the job has no apt dependency at all — the better fix).
- [ ] Consider making the toolchain step fall back to a GitHub-hosted runner or fail fast with a clearer ::error rather than 2+ minutes of apt retries.
- [ ] Revert `HIVE_RUNNER_LABELS`.

clubanderson:
Note: re-running failed jobs keeps the original `runs-on` labels (self-hosted), so the variable flip only applies to NEW runs. Re-triggered #6638/#6644/#6647 by close/reopen; they now schedule on `ubuntu-latest`.

clubanderson:
Investigated this from the repository side.

[...]

PR #7315, opened 2026-09-17
🌱 ci: build and push the toolchain runner image from CI (#7289)

### LINKED ISSUE #7289: [ci] 'Prepare cgo toolchain' still fails intermittently on self-hosted hub shards — #6648's top remediation (bake gcc into the runner image) appears unapplied
## What happened

The `Prepare cgo toolchain for race tests` step failed on self-hosted hub shards **twice within ~90 minutes** today, on two unrelated PRs:

| PR | Run | Job | Shard | Step duration |
|---|---|---|---|---|
| #7276 | 35177793677 | 105063232272 | `test (hub 1/3)` | 5m52s |
| #7287 | 35180255265 | 105070658701 | `test (hub 2/3)` | 7m12s |

In both cases the job failed **before running any tests** — `Test slice of pkg/hub` never started — and both **passed on rerun with no code change**. Neither failure had anything to do with the PR's contents (one was a config default change, the other a `cmd/hive` refactor).

## Why this is worth reopening the topic

This is the failure mode #6648 documented and mitigated. That issue is CLOSED, and `ci-install-tool.sh` does exactly what it promised: it bounds the cost and emits one clear annotation instead of a wall of apt warnings. **The mitigation is working as designed.**

What it explicitly does *not* do — and says so in its own header — is fix egress: *"This script does not try to FIX egress; it cannot."* The remediations it names are ranked, and the top one is **bake the tool into the runner image**, which removes the apt dependency entirely rather than making its failure cheaper.

Today's recurrences suggest the egress drop is **intermittent rather than resolved**, so the ranked-first remediation has not been applied. The current state is that a random self-hosted shard can fail any PR at any time for reasons unrelated to the change, and the only recovery is a human noticing and rerunning.

## Impact

- Every occurrence costs a full rerun cycle (~10+ min) and needs a human to distinguish an infra failure from a real one. The tell is subtle: `gh run view --log-failed` returns **empty**, and only the job step list shows `X Prepare cgo toolchain for race tests`.
- It is indistinguishable at a glance from a genuine test failure, so the safe reaction is to investigate every time.

## Suggested fix

Apply #6648's own first-ranked remediation: **preinstall `gcc` + `libc6-dev` (and `tmux`) in the self-hosted runner image.** The script's fast path already returns immediately when the tool is present, so a baked image makes the common path touch no network at all and this class of failure disappears rather than being retried more cheaply.

Filed after hitting this twice in one session; happy to be closed as a duplicate if the runner-image work is already tracked elsewhere.

Danathar:
Refs #7289

## Summary

#7206 wrote the Dockerfile that bakes `gcc`/`libc6-dev`/`tmux` into the self-hosted runner image so race shards stop failing at **Prepare cgo toolchain** when the cluster's apt egress drops. It merged on 2026-09-16 — and a day later the image still did not exist on GHCR and two more `test (hub 2/3)` shards had died the same way, because *publishing* the image was left to "an operator with `docker` and GHCR write access".

This PR removes that half of the rollout blocker: a **CI Runner Image** workflow builds the Dockerfile on a GitHub-hosted runner and pushes it to `ghcr.io/hivecommons/hive-ci-runner` with the job's own `GITHUB_TOKEN`. The only step left for an operator is the one CI genuinely cannot do — `kubectl patch` the RunnerDeployment in `arc-systems` to the tag the run prints — and the job summary hands them that command with the digest filled in.

## How it works

- **`workflow_dispatch`** builds and pushes. Inputs: `runner_base_image` (default matches the live runners) and `tag_suffix`. The pushed tag is `<runner version>-<suffix>`, e.g. `v2.337.0-ubuntu-24.04-toolchain-1`, so the tag always records which runner release it was built on (what the #6648 suite caveat needs an operator to read off the cluster). An existing tag is **refused, never overwritten** — ARC does not re-pull an unchanged tag.
- **`pull_request`**, path-filtered to the Dockerfile and the workflow, builds only. The Dockerfile's own `gcc --version` / `tmux -V` verification becomes the gate that keeps a Dockerfile change from shipping an image that quietly sends CI back to the mirrors.
- **`runs-on: ubuntu-latest`, not `HIVE_RUNNER_LABELS`** — the image exists to remove CI's dependency on the cluster's egress, so building it behind that egress would defeat it.
- **No `type=gha` layer cache** — the Actions cache is the resource #7206 measured full (9.99/10 GB, 96.6% buildkit blobs), the reason the #7009 offline `.deb` cache never survived.
- **No `push:` trigger / no `branches:` filter** — rolling a new tag restarts every runner pod, so *when* stays an operator decision; the release-line guard has nothing to classify (verified with `check-release-lines.sh`, `check-action-pins.sh`).
- Attestations off (`provenance: false`, `sbom: false`) so the pushed tag is a plain image manifest, per the #3760 contract.

## Also

- `src/deploy/ci-runners/README.md` now leads with `gh workflow run ci-runner-image.yml --ref v4 …`, keeps the laptop `docker build` as a fallback, marks the `kubectl patch` as the step that still needs `vllm-d` access, and adds the first-push caveat: GHCR creates the package **private** and ARC pulls anonymously, so `hive-ci-runner` must be made public (like `hive`/`hive-hub`/`hive-contributor`) before patching or the pods land in `ImagePullBackOff`.
- `pkg/config/ci_runner_image_workflow_test.go` pins the properties above (off-cluster, no gha cache, attestations off, SHA-pinned actions, push only on dispatch, README documents the workflow and the visibility caveat).
- `changelog.d/changed-7289-ci-runner-image-workflow.md`.

## Testing

- `go test ./pkg/config/ -run CIRunner` — all pass (existing Dockerfile guards + 6 new).
- `src/scripts/check-release-lines.sh`, `check-action-pins.sh`, `check-no-image-attestations.sh` — PASS.
- The tag-resolution step was exercised locally for dispatch, PR, untagged base (rejected), and bad suffix (rejected).
- The `pull_request` lane of the new workflow should run on this PR itself (it touches the workflow file), which is the first real build of the Dockerfile.

## After merge

```console
gh workflow run ci-runner-image.yml --ref v4 -f tag_suffix=toolchain-1
```
then follow the run's job summary for the `kubectl patch`. Close #7289 on the first green race shard whose **Prepare cgo toolchain** step reports the tool already present.


clubanderson (approved):
CI-built runner image off-cluster with immutable tags and a build-only PR gate — this closes the gap that left the cgo-toolchain shards flaking since #7206. LGTM.

clubanderson:
Dispatched the workflow after merge — run 35215122771 built and pushed `ghcr.io/hivecommons/hive-ci-runner:v2.337.0-ubuntu-24.04-toolchain-1`. Remaining operator step: `kubectl patch` the RunnerDeployment in `arc-systems` to that tag per `src/deploy/ci-runners/README.md` — the cgo-toolchain flakes persist until the runner pods roll onto it.

Issue #7317, opened 2026-09-17
✨ feature: let an operator troubleshoot a struggling contributor from the dashboard, without shell access to the hub


Danathar:
## Summary

Hive's contributor dashboard (the `/contribute` Operations tab) is the only view an operator has of the contributor ("clanker") fleet when they don't have shell access to the host or pod the hub runs on — which is the normal case for a hosted hive. Today, when a contributor is visibly struggling (tasks picked up and handed back over and over, nothing completing), that view cannot say **why**: the reason strings, pane output and hub decisions that would explain it are either held only in memory for the life of the WebSocket connection, written to a JSONL file no endpoint reads per-run, or logged to the hub process's stdout and nowhere else.

This is a request to make the diagnostics the hub already has reachable from the dashboard/API, so an operator can troubleshoot a contributor without `podman logs` / `kubectl logs` on the hub.

Observed 2026-09-17 on `hosted-projectbluefin-knuckle-gjvq`: a `litellm`-backend contributor picked up 11 tasks over ~2h and completed none. From the dashboard, all an operator could see was the activity rail cycling `picked up` → `released: gave the task back`. Blast radius of the fix: dashboard + read API only; no change to routing, cooldowns or the relay protocol.

## What an operator can see today, per layer

Traced against `v4` at `65f2f6f16`.

**Rendered in the dashboard**

- The **"Log" rail** on the Operations tab is the narrated activity feed (`joined` / `picked up` / `released: …` / `failed`), capped, with no reason text. Comment at `src/pkg/dashboard/api_contribute.go:2806` — it is the same SSE stream as the Onboarding tab's Live Activity, renamed in #2591.
- The **fleet cards** show backend, model, tier, connect time.

**In the API but not rendered**

- `GET /api/contribute/fleet` carries `last_failure` (reason, kind, permanent, task) and `tmux_output` (the pane tail from the relay's last `task_progress`). Both live on the connection struct (`contribute_ws.go:160-170`, `:3231`, `:339`) and are gone the moment the contributor disconnects. I found no consumer of either field in the ops-page JS, so even while a contributor is connected an operator has to `curl` the endpoint to read them.

**Durable, but only as aggregates**

[...]

Issue #4260, opened 2026-08-20
A relay that reconnects inside the backoff window cannot resume its task — in-flight work is interrupted mid-turn


Danathar:
## Summary

When a contributor relay's WebSocket drops and it reconnects a second later, it re-asserts the task it is still working on — exactly as the disconnect path's own comment says it should. The hub replies `task_revoke: no active lease for this task`, then assigns the same issue again as a new task. The relay types the new prompt into the pane, which interrupts the agent mid-turn.

The disconnect handler explicitly expects the opposite:

> the original relay — which keeps `currentTask` locally and re-asserts it via `task_progress` on reconnect — is still working it […] The short window comfortably outlasts the reconnect backoff, so **the returning session re-asserts and resumes** (repopulating `activeIssues`) before the cooldown lapses

That resume does not happen. I have two captures, on different backends, 18.7 hours apart.

## Reproduction 1 — `codex`, issue #4203

```
Task assigned: issue kubestellar/hive#4203 — 🧪 Podman: evaluate Quadlet .kube reuse for standalone Hive
Task prompt sent to CLI
GitHub token refreshed
Connection to wss://…/api/contribute/ws closed. Reconnecting in 1000ms...
Connecting to wss://…/api/contribute/ws...
Connected to wss://…/api/contribute/ws
Authenticated with wss://…/api/contribute/ws as c-ca8187f7480c (tier: contributor)
Hub protocol 1.2; capabilities: token_refresh, task_unavailable_reasons, prompt_preview, …
Reconnected while working on kubestellar/hive#4203 — resuming
Task revoked: ct-kubestellar/hive-4203-1787228212 — no active lease for this task
Task assigned: issue kubestellar/hive#4203 — 🧪 Podman: evaluate Quadlet .kube reuse for standalone Hive
Task prompt sent to CLI
```

What that did to the agent, from its own pane at the same moment:

```
• Ran test -e src/docs/podman-quadlet-kube-spike.md && echo exists || echo absent
■ Conversation interrupted - tell the model what to do differently.
› You are a contributor to the kubestellar/hive hive. Work on issue kubestellar/hive#4203 …
• I'm continuing with the GitHub contribution workflow […] The checkout, live issue/duplicate
  checks, and isolated Podman/Quadlet spike are already complete; I'm first rechecking the
  branch and worktree after the interruption, then I'll finish the shippable report …
```

[...]

PR #7322, opened 2026-09-17
✨ feat(contribute): serve per-run task history and record abandonments (#7317)

Danathar:
## Summary

An operator watching a struggling contributor on the `/contribute` Operations tab could see *that* it was failing and never *why*. On `hosted-projectbluefin-knuckle-gjvq` a litellm contributor picked up 11 tasks in ~2h and completed none; the whole session read as `picked up` → `released: gave the task back`, eleven times, with no reason text anywhere. The hub had the answers and threw them away.

This lands **items 1 and 2** of #7317 — the two that need no new in-memory state and no UI framework. Refs #7317; it stays open for items 3 and 4.

## Item 2 — abandonments are recorded at all

`task_runs.jsonl` only ever gained a row for an *accepted* `task_complete`/`task_failed`. A relay asking for new work while still holding a task, or a socket dropping with one held, wrote nothing durable — so that 11-task session left **one row**, and `run-stats` reported one failure for two hours of failure. Both paths now append a record:

```json
{"outcome":"abandoned","abandon_cause":"handback",
 "reason":"abandoned: relay asked for new work while still holding this task",
 "duration_s":1560.4,"scenario":"abandoned_handback","username":"...","backend":"litellm"}
```

**The duration is the point.** On that session the hand-backs land on the relay's own timeouts — ~26 min is `PANE_STALL_TIMEOUT_MS`, ~10 min is `CLI_READY_TIMEOUT_MS` — so the number says which one fired. Today that is only reachable by reverse-engineering relay constants from activity timestamps, which the issue rightly says an operator should not have to do. It is left **unset, not zero**, for a task adopted on the resume path with no fresh assignment: absent means unknown, and a `0` would read as an instant hand-back — the very thing being told apart from a stall.

### Why `abandoned` is not `failed`

#4260 established that a dropped socket is not a failure of the work; booking it as one is what turned three dropped sockets into a quarantine of an issue nobody had failed. The run log inherits that distinction. `run-stats` gains an `abandoned` bucket and keeps abandonment durations **out of the percentile pool**, so neither number the backend ratchet reads moves merely because these rows started existing:

| | before | after |
|---|---|---|
| `failed` | work that failed on its merits | unchanged |
| `abandoned` | — | new bucket |
| `duration_p50_s` | completed + failed | unchanged (a 26-min stall is not evidence about run length) |
| `chrome_idle_share` | `idle_complete / completed` | unchanged |

The issue notes totals will rise. They do — in a bucket that says what it is.

## Item 1 — the read path `task_run_log.go` was missing

`GET /api/contribute/runs?username=<u>&days=N&limit=N` serves the records behind `run-stats`' aggregates: outcome, failure kind, **reason**, duration, scenario, per task. Every field was already being written; there was simply no endpoint that served a single record, so a failure's `reason` was reachable only by reading the file on the hub host — precisely the access a hosted-hive operator does not have.

Newest-first, bounded (default 100, max 500), username matched case-insensitively so a login pasted out of the activity rail works.

**Posture is inherited, not chosen.** Public read-only like the sibling `/api/contribute*` GETs: the username is already public on `/api/contribute/activity` and the leaderboard, and `reason` is already served as-is on `/api/contribute/fleet`'s `last_failure` while the contributor is connected. What changes is *durability, not audience*. No token, no pane output — item 3 wants a gate and is not here.

Registered at the **tail** of `RegisterAPI`: `api-reference.md` cites every route by `file:line`, so an insert mid-list rewrites the citation of every route below it and `check-api-reference-citations.sh` goes red for a change that touched none of them.

## Verification

- Full `pkg/dashboard` suite green (48.9s); `go vet` and `gofmt` clean; `check-api-reference-citations.sh` passes; the OpenAPI route-parity test passes (it caught the missing path first, which is how I found it).
- **End to end through the real WebSocket handler.** `TestTaskRunLog_RecordedOnHandback` registers, authenticates, goes `ready`, takes an assignment, then sends `ready` while still holding it — the exact incident shape — and asserts the record comes back off the new endpoint *by username* and counts as `abandoned`, not `failed`, in `run-stats`.
- **8 mutants, one at a time, whole suite after each — 8/8 killed, 0 survived:**

| mutant | killed |
|---|---|
| handback record not written | ✓ |
| disconnect record not written | ✓ |
| abandonment duration dropped | ✓ |
| `abandoned` folded back into `Failed` | ✓ |
| abandonment durations back in the percentile pool | ✓ |
| runs endpoint ignores the username filter | ✓ |
| runs returned oldest-first | ✓ |
| scenario ignores the abandon cause | ✓ |

Each anchor was asserted to match exactly once, so a mutant that failed to apply could not be miscounted as a kill, and `contribute_ws.go`/`task_run_log.go` were restored byte-for-byte after.

Worth flagging: **the duration mutant survived the first round.** My end-to-end test asserted everything about the record except the field my own comment calls the most diagnostic one. That is why `TestAppendAbandonedRun_DurationOnlyWhenKnown` exists, covering both the known and adopted-task branches.

## Not in this change

- **Rendering any of it in the dashboard UI.** The endpoint is reachable from an operator's browser, but item 1's "render it on the dossier page / fleet card" is not done.
- **Items 3 and 4** — per-username ring buffers for `last_failure`/`tmux_output`, and the hub's contributor-scoped decisions. Both want in-memory state with a retention policy and a sensitivity gate (the issue itself suggests gating them on owner/read-write), which is a separate argument from this one and a larger change.

No routing, cooldown, trust or relay-protocol behaviour changes — the two new call sites run *after* their paths' existing routing and feed nothing back into it, matching the DECLARE-never-ROUTE boundary `task_run_log.go` already draws.


Danathar:
@clubanderson — thanks for the `scratchRunLog` fix, and it was a real one, not a theoretical one. I checked rather than taking it on trust:

| | `-race -count=3`, run-log tests |
|---|---|
| with your guard | clean |
| with the pre-fix unguarded swap/restore | **`WARNING: DATA RACE`** on `taskRunLogPath` |

And the trigger is mine: the disconnect-abandonment append this PR adds is exactly what makes the deferred handler read that variable after the test has returned, so the race did not exist before this branch. Good catch.

**I had to re-author your commit, and I want to be upfront about it.** `Sign-off survives the squash` went red on `5c9cc386`: the squash merge stamps the merged commit with the *pull request's* author (me) and keeps the message, so a commit signed by anyone else lands on protected `v4` with its author and sign-off disagreeing — the #6798 / #6971 shape, unrepairable once merged. The check's own guidance is to re-sign as the contributor of record and credit the author with `Co-authored-by`.

So `95752849` is your change with:

- the **tree byte-identical** to `5c9cc386` (`git diff 5c9cc386 95752849` is empty),
- your commit message body kept verbatim,
- **your `Signed-off-by` kept**, not replaced — mine is added alongside it, because the merged commit will carry my name. Your DCO certification still stands on its own line.
- `Co-authored-by: clubanderson` added, plus the `Co-authored-by: Copilot` you had.

`src/scripts/check-squash-signoff-attribution.sh upstream/v4 HEAD Danathar` passes locally now. Pushed with `--force-with-lease` pinned to your SHA, so it could only have landed on exactly the commit I read.

If you would rather own that commit yourself — land it on `v4` separately and I will rebase this PR onto it — say so and I will revert the re-authoring. I did not want to leave a merge-blocking check red on a branch I own, but it is your commit and your call.

clubanderson (approved):
Items 1 and 2 of #7317 with the right accounting: a distinct abandoned bucket keyed on hub-observed cause, durations kept out of the percentile pool, and the unset-vs-zero duration distinction on resume. I fixed the scratchRunLog restore race under taskRunMu (5c9cc38). LGTM.

PR #7329, opened 2026-09-17
✨ feat(contribute): show last failure, agent pane, and run history on the Operations tab (#7317)

Danathar:
## Summary

The contribute Operations tab is the only view an operator has of the contributor ("clanker") fleet when they can't shell into the host or pod the hub runs on — the normal case for a hosted hive. #7322 made the per-run evidence *reachable* (`GET /api/contribute/runs`) and made hand-backs *exist* in that log. What was still missing is the half the issue calls item 3: nothing on the page **shows** any of it, and the one piece of evidence that explains a stall — what was on the agent's terminal — was read by the hub on every progress report and kept nowhere.

This PR renders the diagnostics the hub already had and stores the one it was dropping. Owner/read-write viewers get the pane; everyone else gets exactly what they got before. Refs #7317 (item 3; item 4 is filed separately). Blast radius: dashboard page, two read endpoints gain gated fields, run-log records gain a field. No routing, cooldown, trust, or relay-protocol change.

One correction to the issue text, found while doing this: I wrote there that `/api/contribute/fleet` already carried `tmux_output`. It did not — the connection stored it (`contribute_ws.go:131`) and `ActiveConnections()` copied it, but no handler served it. `last_failure` was the only diagnostic on the wire, and it was unrendered.

### Fleet row (Operations → Connected clankers)

- **Last failure line.** `last_failure` has travelled on the fleet snapshot since #2547 and was never drawn. Now: `last failure: environment CLI never became ready · projectbluefin/server#96 · 5m ago`, amber, with the full reason as a tooltip. Byte-for-byte the old row when there is none.
- **Agent pane.** While a task is in flight, a collapsed `<details>` showing the terminal the relay last reported in `task_progress`. Open/closed state is remembered per clanker across the 4-second re-render, so the pane does not slam shut while the operator is reading it (toggle events don't bubble; the listener uses capture).
- **"run history →" link** that fills the new card below with this login. For every viewer, not just admins — it's a public read.

### Contributor run history card (new, under Fleet work)

A lookup **by GitHub login** rather than something hung off a fleet row, because the operator who needs this is usually looking at a contributor that has already dropped off: the fleet row is gone, the log rail has scrolled, and the run log is the only thing left. Renders `/api/contribute/runs?username=…&days=7&limit=50`: outcome pill (completed / failed / abandoned), task link, duration, model, when; the reason (or scenario label when there is none); PR link on completions; and — for admitted viewers — the pane under each failed/abandoned run. When the server withheld panes and there is a failed/abandoned run in the list, one line says so; a list of clean completions says nothing.

### Run log: `pane_tail`

`TaskRunRecord` gains `pane_tail`, set on `failed` (from the report's own `tmux_output` — the relay captures it *before* stopping the agent for exactly this purpose, see `failCurrentTask`) and on both `abandoned` paths (from the connection's last progress report). For a stall this is the frozen screen itself. Not set on `completed`: the evidence there is the PR, and a 15-line tail on every completion would spend the 10 MB log for nothing.

Bounded at write time by `boundPaneTail`: last 30 lines (above the relay's 15 so a future relay isn't silently clipped), 400 runes per line, `redactTokens` on each. Nil in → nil out, so the JSON field is omitted, not `[]`.

### The gate

Pane text is what the agent printed — more sensitive than the bounded reason string the same endpoints already serve anonymously. `paneTailViewer` follows the line the per-clanker controls already draw: `owner` and `read-write` see it; `read`, unknown, and absent roles do not. The empty-role case mirrors `requestRoleAllowsOwner`, not `requireContributorWrite`: on a spoke with any auth boundary an absent `X-Hive-Role` is an anonymous caller and gets nothing; on a genuinely open spoke (no token, no allowlist) the whole dashboard is already anonymous and hiding one field from its only operator would be theatre.

Applied at the boundary in `handleContributeFleet` and `handleContributeRuns` — the snapshot and the record always carry the pane; the handler decides who gets it, so a live and a disconnected contributor are gated identically. Both responses add `pane_tail_visible` so the page can say "shown to owners" instead of showing nothing. `dashboard/openapi.json` documents the fields and the posture.

Known limit, inherited not introduced: on a direct-route spoke, `/api/contribute/*` is `isPublicPath`, and `authenticate` resolves a session cookie only for non-public paths — so a device-flow-logged-in owner of a direct-route spoke reads these two endpoints as anonymous. Hub-proxied hives (the hosted ones this issue is about) inject `X-Hive-Role` and are fine. Worth its own issue if a direct-route operator hits it.

## Verification

- `go build`, `go vet` clean; `gofmt -l` reports 9 files, all pre-existing on `v4` (checked against a stash).
- Full `pkg/dashboard` suite green, twice (77s / 75s). `check-api-reference-citations.sh` passes.
- New `task_run_pane_tail_test.go`:
  - `boundPaneTail`: keeps the *last* 30, rune-safe truncation, redacts a `ghp_` token, does not alias the input.
  - `paneTailViewer`: 7-row table across guarded/open spokes × owner/read-write/read/unknown/absent.
  - `handleContributeRuns`: same record served to anonymous (stripped, `pane_tail_visible:false`, reason intact), owner, read-write (present), read (stripped).
  - **End to end through the real WebSocket handler**, on a guarded spoke: `task_progress` with a pane containing a token → owner fleet read shows the pane redacted, anonymous fleet read has no `pane_tail` but the row is otherwise intact → `ready` while holding (hand-back) → second task `task_failed` with a pane on the report → both log records carry the right pane, and the anonymous `/runs` read has the reasons and no `"pane_tail":`.
- The served page's six inline `<script>` blocks pass `node --check` (caught one precedence typo of mine before commit). The render helpers were run in node against sample fleet/run payloads and the HTML inspected.

## Not in this change

- **Item 4** — the hub's contributor-scoped decisions (stale-generation rejections, unassigned-task ignores, abandon-without-completion, no-tasks-available). Filed separately with the gating decision made up front.
- The dossier page. The run-history card covers the disconnected case; a "recent runs" section on `/contribute/dossier/{username}` would be the same renderer in a second place and can follow if wanted.

## Related issues

Refs #7317

## Testing

- [x] `cd src && go test ./pkg/dashboard/...` (full package; `./...` not run — the change is confined to `pkg/dashboard`)
- [ ] Other / not run (explain):

## Contributor checklist

- [x] PR targets `v4` (the current default branch).
- [x] Docs, examples, and policies are updated when behavior changes (`dashboard/openapi.json`).
- [x] A `changelog.d/added-7317-fleet-diagnostics.md` fragment carries the changelog entry.


clubanderson (approved):
Items 3 and 4 of #7317 — last-failure surfacing, pane tail, and run history on the Operations tab. LGTM.

Issue #7330, opened 2026-09-17
✨ feature: surface the hub's per-contributor decisions (stale-gen rejections, refusals, ignored reports) on the Operations tab — #7317 item 4


Danathar:
## Summary

Hive's contributor dashboard (the `/contribute` Operations tab) is what an operator has when they can't shell into the hub's host or pod — the normal case for a hosted hive. After #7322 and #7329, a struggling contributor's **runs** are visible from the page: outcome, reason, duration, and (for owners) the agent's terminal when each run stopped. What is still invisible is the hub's **own decisions about that contributor** — the moments where the hub refused, ignored, or fenced something the relay sent. Those are `slog` lines on the hub's stdout and nowhere else, so from the dashboard a contributor whose reports are being silently dropped looks identical to one whose relay never sent them.

This is item 4 of #7317, split out so the retention and gating decisions are made up front rather than left to whoever picks it up. Blast radius: additive read API + one panel; no routing, cooldown, trust, or relay-protocol change.

## Why this matters — the motivating case

On `hosted-projectbluefin-knuckle-gjvq` on 2026-09-17, a `litellm` contributor picked up 11 tasks in ~2h and completed none. 10 ended as `released: gave the task back` — the hub received a `ready` while it still held a task — and none as `failed`. Two of those hand-backs took 26 min and two took 10 min, the exact durations of the relay's pane-stall and CLI-ready timeouts, both of which end in `failCurrentTask()`, which sends `task_failed` *before* `ready`. So either the relay's `task_failed` never arrived, or the hub rejected it. Reading the hub: the `#2568` generation fence (`generationAccepted`, `contribute_ws.go:3028`) drops any `task_failed` whose `task_gen` is 0 once the connection has ever sent a non-zero one, and the relay's up-front rejections (`isGivenUp`, "already has active task", "not the active polling slot" — `bin/contributor-relay.js:4240-4257`) send `task_failed` **without** `task_gen`. If that is what happened, the hub logged `stale-generation task_failed rejected` eleven times and no operator could see it. If it isn't, something else happened and no operator can see that either. The point of this issue is that the question should be answerable from the page.

## What the hub decides and currently only logs

Contributor-scoped `Warn`/`Info` lines in `contribute_ws.go` that name a username and never reach any endpoint (line numbers at `v4` `c67890c1b`):

| line | decision |
|---|---|
| 3837 | `model rejected` |
| 3852 | `agent role claim rejected` |
[...]

PR #7332, opened 2026-09-17
✨ feat(contribute): serve the hub's per-contributor decisions (#7330)

Danathar:
## Summary

Refs #7330 (item 4 of #7317). After #7322 a struggling contributor's **runs** are readable from the dashboard. The other half of the conversation was not: the moments the hub **refused, ignored or fenced** something the relay sent were `slog` lines on the hub's stdout and nowhere else — so from the page, a contributor whose reports are being silently dropped looked exactly like one whose relay never sent any.

That is what the motivating session turns on. A litellm contributor picked up 11 tasks in ~2h, completed none, and ended 10 as "gave the task back" with **no failure anywhere** — even though the relay's pane-stall and CLI-ready timeouts both call `failCurrentTask()`, which sends `task_failed` *before* `ready`. Either those never arrived or the `#2568` generation fence rejected them. The hub knew which; the operator could not.

## 1. Record

A bounded per-login ring (`hub_decisions.go`) behind a closed event vocabulary — `stale_gen_rejected`, `unassigned_ignored`, `abandoned`, `resume_rejected`, `lease_expired`, `refused` — written at the decision sites in `contribute_ws.go` (the three generation fences, both unassigned-report drops, abandonment, both resume rejections, lease expiry, and all six refusal paths). Every site **keeps the slog line it already had**; the ring is additional, never a replacement.

DECLARE-never-ROUTE, the same boundary `task_run_log.go` draws: its own mutex, never `h.mu` or `contributor.mu`, and nothing routing reads it.

**Bounded twice, because one cap is not enough.** 200 entries per login stops a flapping relay; 200 distinct logins stops a stream of *new* logins growing the map for the life of the process. Overflow evicts the least recently active login whole.

## 2. Serve

`GET /api/contribute/decisions?username=<u>&limit=N`, newest first.

**Owner/read-write only**, enforced in-handler because the `/api/contribute` prefix is public in `isPublicPath`. It **403s rather than stripping**: unlike a run's failure reason — already served anonymously as `last_failure` on `/api/contribute/fleet` — these carry the hub's internal protocol state (generation numbers, lease identity, configured rate-limit thresholds), and nothing useful is left once those are removed. Empty-role follows `requestRoleAllowsOwner`: anonymous behind a boundary gets nothing; on a genuinely open spoke the whole dashboard is already anonymous.

In memory only, and the response says so — `since` and `in_memory_only` are returned, so an empty list after a restart reads as "nothing since boot" rather than "nothing happened".

Registered at the tail of `RegisterAPI` so `api-reference.md`'s `file:line` citations don't shift.

## Two things the tests caught — both real, not test artifacts

**Ordering could not use `TS`.** It's RFC3339 second-resolution (matching the run log, which operators read), so a burst of fences — the exact thing this ring exists to capture — lands many entries inside one second. The newest-first sort then degenerated to insertion order and returned **oldest first**, and the eviction picked an *arbitrary* login rather than the least recently active one. Entries now carry an unexported arrival counter and the read is a reverse, not a sort.

**The refusal sites log `identityOf(c)`, which is `ContributorID[#session]`** — a different key space from the GitHub login the ring is keyed by and the endpoint is queried by. Keying on it would have produced an endpoint that always answered empty for a contributor the operator could watch being refused in the logs. `decisionUsername()` exists for that, and a test asserts the two genuinely differ so it cannot quietly become vacuous.

## Verification

- Full `pkg/dashboard` suite green (47.8s); `go vet` and `gofmt` clean; `check-api-reference-citations.sh` passes; the OpenAPI route-parity test passes (it caught the missing path first); the `internal/testutil` sleep ratchet is clean.
- **The motivating case, end to end through the real WebSocket handler**: register, auth, `ready`, take an assignment, establish a non-zero generation with a `task_progress`, then send `task_failed` with **no `task_gen`** the way the relay's up-front rejection paths do — and assert the fence is readable by username off the gated endpoint afterwards.
- **9 mutants, one at a time, whole suite after each — 9/9 killed, 0 survived:**

| mutant | killed |
|---|---|
| gate admits read-only | ✓ |
| gate admits anonymous behind a boundary | ✓ |
| handler serves instead of 403 | ✓ |
| per-user ring unbounded | ✓ |
| distinct-user map unbounded | ✓ |
| results returned oldest-first | ✓ |
| blank username still recorded | ✓ |
| `decisionUsername` returns the contributor id | ✓ |
| stale-gen `task_failed` not recorded | ✓ |

Each anchor was asserted to match exactly once, so a mutant that failed to apply couldn't be miscounted, and both source files were restored byte-for-byte after.

## Note for whoever merges this after #7329

#7329 introduces `paneTailViewer` with **this exact predicate** for pane output. They are the same gate for the same reason and should collapse into one helper once both have landed — kept separate here only because #7329 is still open and this branch is based on `v4` as the issue asks. Happy to do the unification in whichever lands second.

## Not in this change

**The render half.** Item 3 of the issue wants these interleaved into the *Contributor run history* card — and that card arrives with #7329, which is still open, so there is nothing on `v4` to render into yet. The endpoint is reachable from an operator's browser in the meantime.

No routing, cooldown, trust or relay-protocol behaviour changes: every call site records *after* the decision it observes and feeds nothing back into it.


clubanderson (approved):
The right half of the story after #7322: hub refusals recorded in a doubly-bounded ring at every decision site, slog lines kept, and served owner/read-write with an in-handler 403 — the stricter posture these entries need. LGTM.

PR #7339, opened 2026-09-17
✨ Serve hub-side contributor decisions from the dashboard (#7317 item 4)

clubanderson:
## What

Implements **item 4** of #7317: the hub's own decisions about a contributor are now kept in a bounded per-contributor buffer and served at

```
GET /api/contribute/decisions?username=<u>&limit=N
```

## Why this item, and why this shape

When a contributor is visibly struggling, an operator asks two questions. `GET /api/contribute/runs` (already on `v4`) answers *"what happened to that task?"*. Nothing answered *"why did the hub refuse that report?"* — the hub made the decision, logged it to its own stdout, and threw it away. That is exactly the access a hosted-hive operator does not have, and `task_run_log.go`'s own header already concedes the shape of the problem: *"the slog line rotates away"*.

The issue offered two routes and named a preference — a **structured, bounded, per-username event buffer** over an operator-facing log tail, because a tail turns the dashboard into a log viewer and hands back unstructured text the operator still has to parse. This is that buffer.

## Scope check — items 1 and 2 were already done

Before writing anything I checked the issue against `v4`:

| Item | State |
|---|---|
| 1. Per-contributor run history endpoint | **already on `v4`** (`GET /api/contribute/runs`) |
| 2. Record hand-backs as run records | **already on `v4`** (`scenarioAbandonedHandback` / `…Disconnect` / `…Other`) |
| 3. Pane output + failure persistence | **deliberately not included** — wants a gate |
| 4. Hub decision buffer | **this PR** |

Item 3 is excluded on purpose, and a test locks the response surface to the four declared fields so pane output cannot arrive here by accident.

## The five decision sites

Each already had an slog line; each now records beside it.

| Kind | Meaning |
|---|---|
| `task_abandoned` | asked for new work while still holding a task |
| `no_tasks_available` | hub had nothing to offer |
| `stale_progress_rejected` | `task_progress` carried an older generation |
| `stale_failure_rejected` | `task_failed` carried an older generation |
| `unassigned_complete_ignored` | `task_complete` named a task they did not hold |

`stale_failure_rejected` is the one that explains the silent hand-back: the relay **did** report a failure and the hub refused it, so the task ended as a bare `ready`.

## Safety

**Declare, never route.** Recording is observational only — nothing here feeds cooldowns, trust, routing or offers, and a full buffer never changes a decision, it only forgets the oldest. That is the same boundary `task_run_log.go` draws, and it is what makes this safe to call from inside the message loop.

Bounded twice — 50 per contributor, 500 contributors, evicting the least-recently-recorded — so a busy or hostile fleet cannot grow it without limit. In memory on purpose: unlike `task_runs.jsonl` these are not per-run facts worth a disk write, they are the recent context around one.

Public read-only, a posture **inherited rather than chosen**, for the reason `handleContributeRuns` already gives: username and task id are already public on `/api/contribute/activity`, and everything else is the hub's own bookkeeping. No contributor-supplied text, no tokens, no pane output.

## A bug found on the way

`forUser` originally sorted by timestamp. A **burst** of decisions — precisely what this endpoint exists to explain — can share one coarse clock reading, and a stable sort then leaves the tied entries in arrival order, i.e. **oldest-first**, silently inverting the answer for the worst case. It surfaced as my own tests passing under `-race` (slower, so timestamps differed) and failing without it.

Insertion order is the ground truth, so the ring is now reversed rather than sorted. `TestContributorDecisionsNewestFirstWithIdenticalTimestamps` pins it, and I **verified it fails against the original code**:

```
--- FAIL: TestContributorDecisionsNewestFirstWithIdenticalTimestamps
    position 0 = "task-1", want "task-3"
```

## Testing

- 14 new tests: ordering, tied timestamps, case-insensitive lookup, per-user scoping, ring eviction, user-map eviction, limit clamping, empty-array-not-null, copy-on-read, nil-Server no-op, concurrent record/read under `-race`, and the response-surface lock.
- New file at **100.0%** statement coverage.
- Full `pkg/dashboard` suite green — **89.0%**, up from 88.9% — and green under `-race`.
- `go build ./...`, `go vet`, `gofmt` clean.
- `check-api-reference-citations.sh` clean; route registered in `dashboard/openapi.json` and `src/docs/api-reference.md`.

Refs #7317


clubanderson:
Closing as a duplicate of #7332, which was opened ~20 minutes earlier and is the better implementation.

I checked before closing rather than assuming, and #7332 is strictly ahead on every axis I care about:

- **Broader vocabulary** — six kinds (`stale_gen_rejected`, `unassigned_ignored`, `abandoned`, `resume_rejected`, `lease_expired`, `refused`) against my five; it covers the resume-reject and lease-expiry paths I did not.
- **Better eviction** — a monotonic `seq` and eviction of the login whose newest entry is oldest, rather than my recency list.
- **It already gets the ordering subtlety right.** I hit a real bug in my own version: `forUser` sorted by timestamp, and a burst of decisions — precisely what this endpoint exists to explain — can share one coarse clock reading, so a stable sort leaves the tied entries in arrival order and silently inverts newest-first. #7332 reasons about exactly this and reverses the ring instead of sorting. That was the one thing I thought might be additive; it is not.

No reason to keep two implementations of the same buffer. Branch deleted.

Separately, while reviewing this area I picked up #7334 (the `redactTokens` bypass on the sibling `/api/contribute/runs`), which is still unowned — that one I am taking forward.

PR #7353, opened 2026-09-17
🐛 fix(proxy,dashboard): /terminal is unauthenticated on the hivecommons apex — gate every hosted apex, fail closed

clubanderson:
## 🔴 Unauthenticated shell on every hosted spoke

`/terminal` on hosted spokes is **completely unauthenticated** on the `hive.hivecommons.dev` apex. Verified from outside the cluster with **zero credentials** — no cookie, no SSO session, no token:

```
GET  https://<spoke>.hive.hivecommons.dev/            -> 302  (SSO, correctly protected)
GET  https://<spoke>.hive.hivecommons.dev/api/status  -> 302  (SSO, correctly protected)
GET  https://<spoke>.hive.hivecommons.dev/terminal/   -> 200  (ttyd page)        <-- EXPOSED
GET  https://<spoke>.hive.hivecommons.dev/terminal/token -> 200 {"token": ""}    <-- EXPOSED
WSS  .../terminal/ws?arg=hive-<agent>  -> UPGRADED (sec-websocket-protocol: tty) <-- EXPOSED
```

That is an interactive shell in a pod holding the GitHub App private key, every agent token and the hub secret. **CWE-862 / CWE-306.**

## Root cause

`src/proxy/server.js` hardcoded a single apex:

```js
const HOSTED_SUFFIX = '.hive.kubestellar.io';
```

A hosted spoke runs with `HIVE_DASHBOARD_TOKEN` deliberately unset (identity comes from the hub cookie). So on the rebranded apex **both** halves of `if (isHosted || DASHBOARD_TOKEN)` are false and the gate is skipped — for the HTTP handler *and* the WebSocket upgrade. The apex constant was never updated for the kubestellar → hivecommons rebrand, and every existing terminal-gate test addresses the spoke at `*.hive.kubestellar.io`, i.e. the one host the broken constant happened to cover, so nothing exercised the failing-open case.

## Fix

- **Multi-apex + operator escape hatch.** `DEFAULT_HOSTED_SUFFIXES` lists every known hosted apex; `HIVE_HOSTED_HOST_SUFFIXES` adds more without a rebuild.
- **Fail CLOSED, not open.** `IS_HOSTED` (a resolved hub session key — not attacker controlled) is now **authoritative** in `isHostedHost()`. The `Host` header can only *add* hosts to the gated set, never remove one, so a vanity alias, a bare Service IP, a future apex and a forged `Host` are all gated. Combined with the existing boot guard (which refuses to start a proxy that is neither hosted nor token-protected), **no production configuration is left in which an unrecognized host skips terminal authentication.**
- **One predicate for both gates.** The WS upgrade branched `if (isHosted) … else if (DASHBOARD_TOKEN) …`, leaving a silent third case that reached the same ttyd ungated. Both paths now call `terminalGateApplies()`.
- **Dashboard (`static/index.html`)**: `isHostedHiveHost()` had the same hardcode, so `renewTerminalAssertion()` no-op'd on the new apex and the 15-minute `hive_terminal_assertion` expired with nothing to refresh it. `dashboardTokenConfigured()` mapped the hosted **404** from `/api/auth/token` to "token configured", so a failed handoff made `openTerminal()` `tab.close()` the tab it had just opened instead of falling back to the plain terminal URL.

The OpenShift-Route lane is untouched: `HIVE_INGRESS_AUTHZ` and `isAuthorizedForThisHive()`'s fail-closed empty-allowlist policy are unchanged.

## No risk to legitimate authenticated terminal access

The SSO handoff lane (`/terminal?code=…`) short-circuits to the Go API **before** the gate and is unchanged — that is the path the dashboard's ▶ terminal button drives. After redemption the browser carries the spoke-minted assertion cookie, which is the gate's PRIMARY credential and is independent of the allowlist. Both are covered by explicit **control** assertions in the new test (`?code=` still 200; a valid hub-signed, per-hive-authorized cookie still opens HTTP **and** WS).

## Tests — red before, green after

New `src/proxy/terminal_hosted_domain.test.js` spawns the real `server.js` in the exact shape a hub-provisioned spoke runs in (hub master set, `HIVE_DASHBOARD_TOKEN=''`) and addresses it at the second apex with `http.request` (not `fetch`, which drops a custom `Host`).

**Against unfixed code:**

```
AssertionError [ERR_ASSERTION]: anonymous GET /terminal at hive-rebrand.hive.hivecommons.dev
  must be denied — an unknown hosted apex must not skip the gate
  actual: 200,
  expected: 401,
```

**Against this branch:**

```
  ✓ anonymous terminal denied on the rebranded hosted apex (HTTP + WS)
  ✓ forged cookie denied on the rebranded hosted apex
  ✓ anonymous terminal denied at a vanity host on a hosted spoke
  ✓ unrecognized / forged Host headers fail closed
  ✓ SSO handoff (?code=) lane still reaches the API
  ✓ valid hub-signed + authorized cookie still granted (HTTP + WS)
PASS terminal_hosted_domain.test.js
```

Full `npm test` in `src/proxy` passes (13 suites). The Go pin test fails before / passes after:

```
--- FAIL: TestStaticTerminalHostedApexWiring (0.00s)
    static dashboard hosted-hive terminal wiring missing "const HOSTED_HIVE_HOST_SUFFIXES = [...]"
```
```
--- PASS: TestStaticTerminalLinksRenewAssertionBeforeOpening (0.01s)
--- PASS: TestStaticTerminalHostedApexWiring (0.00s)
ok  github.com/hivecommons/hive/pkg/dashboard
```

## ⚠️ This proxy gate is the ONLY containment that reaches the live fleet

The live `hive-terminal` Ingress on **all 7** hosted spokes is missing its auth annotations:

```
aslom-hive-agent-ua5i                        auth-url=*** MISSING ***
available-oke-260812-0nzc-available-ok-ba44  auth-url=*** MISSING ***
available-oke-260812-ixya-available-ok-3do3  auth-url=*** MISSING ***
available-oke-260812-jlem-available-ok-nnpj  auth-url=*** MISSING ***
daviddiaz0317-visual-hive--1jos              auth-url=*** MISSING ***
kubestellar-console-4vkt                     auth-url=*** MISSING ***
projectbluefin-knuckle-gjvq                  auth-url=*** MISSING ***
```

The template renders them correctly and `saas_provision_terminal_authz_test.go` asserts it — the **deployed** state has drifted. And nothing ever repairs it: the only post-provision Ingress mutation is the vanity-host attach loop (`saas_provision.go:3801`), which appends `spec.rules` entries and TLS hosts and **never reads or writes `metadata.annotations`**.

So layer 2 protects only *newly* provisioned spokes and will never reach the 7 exposed ones. **This proxy gate is the only containment that does**, because it ships in the container image (`imagePullPolicy: Always`) and takes effect on the spoke's next restart.

### Acceptance: does a restarted spoke self-protect with NO config change? **Yes.**

Two independent mechanisms, either alone sufficient, both requiring zero new env vars:

1. **`.hive.hivecommons.dev` is a compiled-in DEFAULT.** It lives in `DEFAULT_HOSTED_SUFFIXES`, not in `HIVE_HOSTED_HOST_SUFFIXES`. That env var is purely *additive* for future apexes; **unset is the safe state**, and the fix does not depend on it.
2. **`IS_HOSTED` short-circuits before the Host header is even consulted.** `IS_HOSTED = SESSION_KEY !== '' || SESSION_PUBLIC_KEY !== ''`, and `HIVE_SESSION_PUBLIC_KEY` is injected unconditionally into every hosted spoke (`saas_provision.go:3446`). All 7 already carry it, so `isHostedHost()` returns `true` for them at *any* hostname.

The new `— ACCEPTANCE —` block in the test pins exactly this: it spawns the proxy with **no hub secret, no session key, no dashboard token and `HIVE_HOSTED_HOST_SUFFIXES` unset**, so only the built-in list can be doing the work, and requires both hosted apexes to deny:

```
✓ both hosted apexes gated by DEFAULT, no env var set, no hub key, no token
✓ control: the suffix list is what denies (non-hosted host still 200 here)
```

Red against unfixed code, with all the other assertions stripped so this one is proven on its own:

```
AssertionError: the CURRENT hosted apex must be gated by the BUILT-IN suffix list,
  with HIVE_HOSTED_HOST_SUFFIXES unset — the live fleet gets no config change, only a new image
  actual: 200,  expected: 401
```

The positive control (a genuinely non-hosted host still returns 200 on that same stripped proxy) is what makes the denials non-vacuous rather than some unrelated blanket failure.

## Commit 3 — secondary, droppable: Ingress annotation reconcile

Separate commit, **defense in depth only**. A bounded sweep modelled on `reconcilePerHiveEnv` (same eligibility filter, reachability breaker, kubectl timeout, per-cycle patch cap, idempotent/non-fatal posture) merge-patches the missing annotations back onto already-provisioned spokes. It compares **values, not presence**, so an `auth-url` left pointing at a previous hub URL or a *different* hive counts as drift instead of looking converged. The merge patch adds only the `auth-*` keys, leaving cert-manager and proxy-timeout annotations in the same map untouched. OpenShift-Route spokes have no such Ingress, so the `get` fails and they are skipped — correct, since Routes carry no nginx auth annotations and that lane's gate is the proxy's fail-closed allowlist.

Drop this commit freely; it does not affect the proxy fix. **No live cluster state was changed by this PR.**


clubanderson:
**v4/v5 divergence note.** This is the v5 counterpart of #7352 (v4). `src/proxy/server.js` and `src/proxy/package.json` are byte-identical on both branches (same blob `e5be38c3`), so the proxy diff is identical. `src/pkg/dashboard/static/index.html` has diverged (~1000 lines), but both hunks — `isHostedHiveHost()` and `dashboardTokenConfigured()` — are unchanged between the branches and were patched by anchored match, not a blind cherry-pick. The Go pin test lives in a different file: v4 still has `static_index_test.go`, whereas on v5 the SPA-content assertions were moved to `static_terminal_links_test.go` (#5565), so the new `TestStaticTerminalHostedApexWiring` was appended there instead. Both branches were tested independently; results in the PR body reproduce on each.

PR #7364, opened 2026-09-17
🐛 fix(contribute): stop handing anonymous hub-proxied visitors the owner's identity (#7362)

### LINKED ISSUE #7362: 🐛 bug: signed-out visitors to a hub-proxied Operations tab are shown the hive owner's contribution stats
## What a visitor sees

Open `/contribute` on the projectbluefin hive, click **Operations**, while **not signed in**. The "Your contribution" card shows the hive owner's numbers — 300 issues worked, 3066 failed — as if the visitor were the owner. Anyone on the internet gets this. They did not sign in and are not the owner.

The sign-in prompt that #6946 added for this case never appears, because the server told the page the visitor *is* signed in.

## Why

`/api/contribute/me` works out who is asking in `resolveViewerUsername` (`src/pkg/dashboard/api_contribute.go`). It tries three things in order:

1. A dashboard session cookie.
2. The `X-Hive-User` header the hub's nginx adds for signed-in visitors.
3. **Fallback:** read the owner's saved GitHub token from disk and treat the caller as the owner.

Step 3 exists for a single-owner spoke: one person runs the dashboard on their own machine, so whoever reaches it is the owner. It is already skipped when direct-route auth is on.

It is **not** skipped on a hub-proxied spoke. But there, the hub sets `X-Hive-User` for every signed-in visitor, so a request *without* it means "anonymous". `/api/contribute*` is public, so anonymous requests arrive all the time — and each one falls through to step 3 and is handed the owner's identity.

`handleGHUserAuthStatus` in `api.go` has the same fallback and the same gap, which is why the page then reports `logged_in: true` for the owner. `handleAuthToken` already guards against this with `hubProxied()`; the two handlers above just don't.

## Fix

Skip the owner-token fallback when `s.hubProxied()` is true, in both `resolveViewerUsername` and `handleGHUserAuthStatus`. Single-owner spokes keep the fallback. PR to follow.

## Environment

- projectbluefin hive, hub-proxied spoke
- `origin/v4` @ `d1c94f97d` still has the fallback unguarded

Danathar:
Fixes #7362

## What a visitor saw

Open `/contribute` → **Operations** on the projectbluefin hive while signed out. The "Your contribution" card showed the hive owner's numbers (300 issues worked, 3066 failed) to anyone who visited, and the page believed the visitor was logged in as the owner. The sign-in prompt from #6946 never appeared because the server said there was nothing to prompt for.

## Why

Two handlers decide who is asking by checking, in order: a session cookie, then the `X-Hive-User` header the hub's nginx adds for signed-in visitors, then — as a last resort — the owner's saved GitHub token on disk.

That last step is for a single-owner spoke, where the one person who can reach the dashboard is the owner. It already stays off when direct-route auth is on. It did not stay off on a hub-proxied spoke — but there, a request with no `X-Hive-User` header *is* an anonymous visitor, and `/api/contribute*` is public, so those requests arrive constantly. Each one fell through to the owner's token.

## The change

Skip the owner-token fallback when `s.hubProxied()` is true, in `resolveViewerUsername` and `handleGHUserAuthStatus`. This is the same rule `handleAuthToken` already applies. Single-owner spokes are unchanged.

```go
if s.directRouteAuthzEnabled() || s.hubProxied() {
    return ""
}
```

## Tests

New file `contribute_me_hub_proxied_test.go`:

| test | checks |
| --- | --- |
| `TestContributeMeHubProxiedAnonymousIsNotTheOwner` | anonymous request on a hub-proxied spoke gets 401, not the owner's record; a hub-identified visitor still gets their own record |
| `TestContributeMeSingleOwnerSpokeStillResolvesOwner` | single-owner spoke still resolves the owner from the saved token |
| `TestGHUserAuthStatusHubProxiedAnonymousLoggedOut` | `/api/gh-user-auth/status` reports `logged_in: false` for the same anonymous request |

`go build`, `go vet` and `gofmt` clean; full `./pkg/dashboard/` suite passes (57s).


clubanderson (approved):
Identity leak closed — anonymous hub-proxied visitors no longer inherit the owner's /api/contribute/me. Test pins both postures. LGTM.

Issue #5325, opened 2026-08-31
🐛 Kick reports 504 while succeeding: the handler blocks up to 120s waiting for the CLI prompt


clubanderson:
## Bug

Kicking an agent can report **`Kick failed: server returned 504 Gateway Time-out`** while the kick actually succeeds — the agent receives the prompt and runs the session to completion. Observed live on a GHE spoke: the toast showed 504 while the operator watched the agent work in the terminal.

## Root cause

`POST /api/kick/{name}` is synchronous, and its slow path can exceed any reasonable proxy timeout.

`handleKick` (`src/pkg/dashboard/api.go:1489`) calls `SendKick` inline and only responds afterwards:

```go
if err := s.deps.AgentMgr.SendKick(name, msg); err != nil {
```

`SendKick` (`src/pkg/agent/manager.go:4686`) captures the pane and then **waits for the CLI's input prompt before typing**:

```go
// Wait for the input prompt (❯) before sending — the CLI may be
// showing a trust prompt or still initializing even though
// tmuxPaneHasCLI matched a broad marker like "Copilot".
m.mu.Unlock()
if !m.waitForInputPromptForAgent(agent) {
    ...
}
```

That wait is bounded by `inputPromptTimeout = 120 * time.Second` (`manager.go:5930`).

**120s exceeds a typical ingress/proxy idle timeout (commonly 60s).** So whenever an agent's CLI is slow to present its prompt — busy finishing a previous turn, still initializing, mid-restart — the proxy cuts the connection and returns 504 *while the wait is still in progress*. The wait then completes server-side, the prompt is typed, and the agent runs normally. The client has already been told it failed.

## Why the report is actively misleading

The operator is shown a failure for an operation that succeeded. The natural response is to click Kick again, which delivers the prompt a second time — the agent gets duplicate work, and on a hold-gated lane that can mean duplicate advisory comments or beads.

Note this is **not** the JSON-parse bug fixed in #5306. That fix is working correctly here: it is what surfaced the honest `504 Gateway Time-out` instead of `Unexpected token '<'`. This issue is that the 504 itself should not be reachable.

## Proposed fix

Make the kick dispatch return promptly and report delivery separately.

1. **Return `202 Accepted` as soon as the kick is queued**, and move the prompt wait off the request path. The UI already polls agent state and has an SSE channel; delivery success/failure belongs there, not in the POST response.
[...]

PR #7378, opened 2026-09-17
🐛 fix(agent): restart cancels the pending kick and breaks the restart/kick loop (#7363)

### LINKED ISSUE #7363: 🐛 restart does not cancel a pending kick — operator restart re-delivers the same 71 KB prompt and loops forever
## Summary

A dashboard **restart** does not cancel the agent's **pending kick**. The restart completes, the queued kick is re-delivered, and ~10 s later another restart lands and destroys the turn again. The result is an unbreakable loop: the agent never completes a turn, `restart_count` climbs, and every cycle burns a full large-prompt delivery.

Operator-visible symptom: *"I interrupted the agent accidentally, clicked **restart** on the dashboard, and the kick is still being sent — the restart won't interrupt it."*

## Observed (live, projectbluefin spoke)

Two consecutive cycles, identical shape:

```
12:12:03  audit: agent kicked                         trigger=dashboard-api
12:15:12  audit: agent kicked  message_len=71139      trigger=send-kick
12:15:22  archived kick log    reason=restart  bytes=54115
12:15:22  audit: turn interrupted mid-flight  reason=restart
          since_kick_s=10.43  producing=true  interruptions_total=13
12:15:22  audit: agent restarting  restart_count=52  reason=operator
12:15:37  audit: agent restarted   trigger=dashboard-api
12:15:38  audit: agent kicked      trigger=dashboard-api     <-- queued kick replayed

12:19:00  audit: agent kicked  message_len=71155      trigger=send-kick
12:19:11  archived kick log    reason=restart  bytes=17791
12:19:11  audit: turn interrupted mid-flight  reason=restart
          since_kick_s=10.48  producing=true  interruptions_total=14
12:19:11  audit: agent restarting  restart_count=53  reason=operator
12:19:22  audit: agent restarted   trigger=dashboard-api
```

`since_kick_s` is **10.43** then **10.48** — near-identical, so this is a deterministic trigger, not an operator clicking repeatedly. The operator clicked restart **once**.

Confirmed at the pane level: two `capture-pane` snapshots 25 s apart showed the agent mid-output (`Claude Opus 5`, triage text on screen), then a fresh CLI banner (`Loading: 1 instruction, 22 skills`, empty prompt) — i.e. the session was torn down and relaunched between samples.

`restart_count` went 47 -> 53 while the agent produced nothing.

## Aggravating factor: the kick payload is ~71 KB

`message_len=71139` / `71155`. Delivery is not instant — the kick was queued at 12:12:03 and did not reach `send-kick` until 12:15:12 (~3m09s). Because the restart lands ~10 s after delivery begins, the turn is **guaranteed** to be destroyed before it can produce anything useful. The larger the prompt, the more certain the loss.

Note `bytes=54115` in cycle 1 vs `bytes=17791` in cycle 2 — the agent got measurably *less* far on the retry, so this does not self-heal.

## Expected

1. A restart should **cancel any pending/queued kick** for that agent. A restart is an operator saying "stop what you are doing"; it should not be followed by an automatic replay of the very work that was interrupted.
2. If a kick is genuinely mid-delivery, a restart should either preempt it or the UI should refuse and say so — not silently queue behind it and then replay it.
3. A restart must not be able to schedule a kick that will itself be restarted, i.e. the loop must have a breaker.

## Actual

The restart tears down the session, and immediately afterwards `audit: agent kicked trigger=dashboard-api` fires again with the same payload, re-entering the loop.

## Code pointers

- `pkg/agent/manager.go:10103` `Restart()` -> `RestartWithReason()` — takes `m.mu.Lock()` and tears the agent down. I see no cancellation of a queued/in-flight kick here.
- `pkg/agent/kick_async.go:306` and `pkg/agent/manager.go:5166` — both call `m.deliverKickLocked(agent, message, "send-kick")`.
- The `audit: turn interrupted mid-flight` accounting (`since_kick_s`, `producing`, `interruptions_total`) already *knows* the restart is destroying a producing turn — `interruptions_total` reached 14 — but nothing acts on that signal to break the cycle.

Suggested: have `RestartWithReason` drop the agent's pending kick queue as part of the teardown, and/or add a breaker that refuses to auto-kick after a restart that itself interrupted a producing turn.

## Environment

- hive-id: `hosted-projectbluefin-knuckle-gjvq` (hosted spoke, ACMM L5)
- image: `ghcr.io/hivecommons/hive:candidate`
- digest: `sha256:5e9abda0ff0ec5539f8bbc79e9a5a240a14997cf826f1e5fda9f0ed2b705ffa6`
- agent: `scanner`, backend `copilot`, mode `ISSUES_AND_PRS`, `kick_template: scanner-holdgated.md`

Danathar:
## Summary

When an operator clicked **Restart** on an agent, the restart killed the session — but the kick that was still being delivered to that agent (waiting for the CLI's input prompt) did not die with it. It kept waiting, saw the *relaunched* CLI's prompt, and typed the same 71 KB prompt into the fresh session, so the "stop" the operator asked for turned into a replay of exactly the work they interrupted. Ten seconds later the next restart destroyed that turn too, and the agent looped: `restart_count` 47→53, every turn interrupted while producing, nothing ever finished. This PR makes a restart cancel the agent's pending kick (and say so), and adds a breaker so a restart that destroyed a producing turn cannot be followed within seconds by another kick that will itself be restarted.

Closes #7363.

## What changed

All three expectations in the issue, in `pkg/agent`:

**1. A restart cancels any pending/queued kick.** `AgentProcess.kickEpoch` is incremented by every restart teardown (`RestartWithReason`, `RestartWithBootstrap`). Both kick entry points — `SendKick` and `SendKickAsync`'s background `deliverKickAsync` — capture the epoch before they start waiting for the input prompt and re-check it under the lock immediately before typing; a moved epoch drops the message with a wrapped `errKickCancelledByRestart` instead of delivering it. `launchGen` was not reused because it only moves on a *completed* launch; a pending kick has to die the moment the restart begins.

**2. Mid-delivery kicks are preempted, and the UI says so.** The prompt wait (`waitForInputPromptForAgentUnless`) polls the epoch and returns as soon as it moves, so the goroutine is released at restart time rather than when the relaunched CLI shows a prompt. The dispatch registry entry is failed at the same moment (`cancelled: agent restarted (<reason>) while the kick was pending`), so `GET /api/kick/{agent}/status` stops reporting `pending`; `POST /api/restart/{agent}` answers `kickCancelled: true` and the dashboard toast reads "restarted — the pending kick was cancelled, not replayed". `openapi.json` documents the field.

**3. The loop has a breaker.** `tearDownTurnLocked` now reports whether it interrupted a turn and whether that turn was *producing* — the `producing=true` the turn-loss accounting already logged on every cycle of the observed loop. A restart that destroyed a producing turn arms `kickHoldUntil` (60 s, `restartInterruptedKickHold`); during the hold `SendKick`/`SendKickAsync` — governor kicks and the dashboard endpoint alike — refuse with `errKickHeldAfterRestart` naming the reason and remaining time. The observed loop re-kicked one second after each restart; scheduler cadences are minutes. `ResetRestartCount` (the dashboard's "reset restart counter") lifts the hold.

**Exemptions, so recovery still works.** A kick's own crash-recovery restart (bare shell / consent screen) now goes through `restartForKick`: it still invalidates *other* pending kicks but neither fails its caller's dispatch nor arms the breaker, because the caller captures the post-restart epoch and is the intended next delivery — `TestSendKick_RestartsCrashedCLIThenDelivers` is unchanged. `RestartWithBootstrap` cancels pending kicks (the bootstrap is the next turn) without arming the hold.

**Registry hardening.** `kickDispatchRegistry.settleDispatch` settles by dispatch pointer, so a cancelled delivery's late failure can never overwrite the fresh dispatch a later `SendKickAsync` registered for the same agent (the exactly-once property #5325 introduced is preserved).

## What I could not determine

The issue's log shows a `POST /api/kick` landing one second after each dashboard restart completed and a `POST /api/restart` ~10 s after each delivery, both `trigger=dashboard-api`. Nothing in this repo's UI, hub, Discord bot, or hivectl chains those automatically, so the source of that second pair is outside what I can see. This change is independent of it: the pending-kick cancellation stops the replay regardless, and the breaker refuses the 1-second-later kick whenever the restart it follows destroyed a producing turn.

## Testing

- New `kick_cancel_test.go`: registry `cancel`/`settleDispatch` semantics; the teardown bookkeeping (epoch bump, dispatch failure, hold arming, kick-initiated exemptions, hold expiry, `ResetRestartCount` lifting it); both kick entry points refusing during the hold with no dispatch registered and nothing delivered; `tearDownTurnLocked`'s producing report; and `TestRestart_CancelsPendingAsyncKick`, a live-tmux regression that queues a kick against a promptless CLI, runs `RestartWithReason`, asserts the dispatch reads failed/cancelled immediately, that the relaunched session never receives the message once it does show a prompt, that no hold was armed (no turn had landed), and that a deliberate follow-up kick still delivers.
- `go test ./pkg/agent/ -short` and `./pkg/dashboard/ -short -run 'Restart|Kick|OpenAPI'`: the unit tests pass; the tmux-backed test skips here. This sandbox cannot create tmux sockets, so the agent package has 41 pre-existing environment failures — I ran the same suite on a pristine `upstream/v4` worktree and the failure sets are **identical**, i.e. nothing new. The live-tmux test therefore runs for the first time in CI; I traced its flow against the existing `TestSendKick_RestartsCrashedCLIThenDelivers` harness (same `forceFastPaneShell` / `quiesceThenMarkReady` helpers) and it follows the same launch path.
- `go vet` clean on `pkg/agent`, `pkg/dashboard`; `node .github/scripts/check-inline-js.js` passes on the edited `index.html`; `openapi.json` parses and the schema-parity tests pass.


clubanderson:
Rebased over v4 head: the #7370 kick-domain extraction had moved notRunningReason/SendKick/deliverKickLocked into pkg/agent/manager_kick.go, so this PR's changes to those functions were relocated there (verbatim). Build, vet, and kick/restart tests pass locally.

clubanderson:
Rebased again after #7375 moved the restart domain into pkg/agent/manager_restart.go — this PR's changes to SetBootstrapOverride/RestartWithBootstrap/RestartThenSendKick/Restart/RestartWithReason (and the new restartForKick/restartWithReason) now live there. Build, vet, and kick/restart tests pass locally.

clubanderson:
Sleep ratchet was at 201 (three new sleeps in kick_cancel_test.go). Kept the two deliberate interleaving/negative waits and paid for them: converted the follow-up-kick settle loop plus two poll loops in tmux_coverage_test.go to testutil.EventuallyEvery(Func). Ratchet back at 198; affected tests pass locally.

clubanderson (approved):
Restart-cancels-pending-kick fix reviewed across two rebases; relocated hunks live in manager_kick.go/manager_restart.go; sleep ratchet paid via EventuallyEvery conversions. Only red is the known sign-off false positive from my signed rebase commits. LGTM.

PR #7379, opened 2026-09-17
🐛 fix(scheduler): cap the PR list in kick prompts, configurable with the issue cap (#7368)

### LINKED ISSUE #7368: 🐛 PR list in kick prompts is uncapped while issues are capped at 100 — 302 PRs produced a 69.5 KiB kick (3x the documented budget)
## Summary

The kick prompt caps the **issue** list at `maxIssuesPerKick = 100` but applies **no cap at all to the PR list**. On a hive with a large open-PR backlog the PR section alone becomes the dominant term in the prompt, pushing kicks to ~3x the size the codebase's own documented budget assumes.

## The asymmetry

Issues are capped in five places, e.g.:

```go
// pkg/scheduler/scheduler.go:730
const maxIssuesPerKick = 100

// pkg/scheduler/scheduler.go:612
if len(agentIssues) > maxIssuesPerKick {
    agentIssues = agentIssues[:maxIssuesPerKick]
}
```

PRs are not capped anywhere:

```go
// pkg/scheduler/scheduler.go:451
func (s *Scheduler) formatPRListWithPolicy(actionable *github.ActionableResult) (string, bool) {
    ...
    for _, pr := range actionable.PRs.Items {   // <-- unbounded
        ...
        b.WriteString(fmt.Sprintf("  %s#%d by @%s %s\n", pr.Repo, pr.Number, author, title))
    }
```

Note the per-line trims are already there (`maxPRTitleRunes = 70`), so the intent to bound this section exists — only the item count was missed.

## Measured impact

`pkg/dashboard/prompt_history.go:53` documents the intended size budget, and explicitly sizes the PR section for **30** PRs:

```
   template body      ~3.0 KiB
   issue list         up to 12.7 KiB (maxIssuesPerKick=100 issues x ~127 B ...)
   PR list            ~3.5 KiB at 30 open PRs x ~120 B per line
   knowledge section  ~3.0 KiB
   repos section      ~0.4 KiB
 giving ~22 KiB for a worst-case prompt and ~8 KiB for a typical one.
```

On the projectbluefin spoke (302 open PRs, 361 actionable issues, 16 repos) the delivered kick measured:

```
audit: agent kicked  name=scanner  message_len=71139  trigger=send-kick
audit: agent kicked  name=scanner  message_len=71155  trigger=send-kick
```

**69.5 KiB — over 3x the documented worst case.** Reconciling against the budget:

| section | budgeted | actual here |
|---|---|---|
| issue list | 12.7 KiB (capped at 100) | ~12.7 KiB — cap working |
| PR list | ~3.5 KiB at 30 PRs | **~36 KiB at 302 PRs** |
| template | ~3.0 KiB | 7.8 KiB (scanner-holdgated) |
| knowledge | ~3.0 KiB | ~3 KiB |

The PR list is more than half the prompt and is the only unbounded term.

## Why it matters beyond size

Delivery is not instant. On the spoke above, a kick queued at 12:12:03 did not reach `send-kick` until 12:15:12 — **~3m09s** to deliver one prompt. That long delivery window is what makes #7363 (restart does not cancel a pending kick) unrecoverable in practice: the restart lands ~10 s into a 3-minute delivery, so the turn is destroyed every time.

Bounding the PR list shortens the window and reduces the blast radius of that bug.

Secondary effects:
- `promptHistoryMaxSizeMB` sizing assumes ~8 KiB typical / ~22 KiB worst case; at 69.5 KiB the retention window is ~3x shorter than documented.
- Token cost per kick scales with the list.
- The tail of a 302-item PR list is unlikely to influence the agent's behaviour, so most of those bytes buy nothing.

## Suggested fix

Introduce a `maxPRsPerKick` mirroring `maxIssuesPerKick` and apply it in `formatPRListWithPolicy`. Truncating with an explicit "… and N more" marker would preserve the agent's awareness that the list is partial, which a silent slice would not.

Worth considering (raised by an operator): make the cap configurable per hive rather than a bare constant, since backlog size varies enormously across hives. Note the issue cap is currently a hardcoded constant too, so a single shared setting could cover both.

## Environment

- hive-id: `hosted-projectbluefin-knuckle-gjvq` (hosted spoke, ACMM L5, 16 repos)
- image: `ghcr.io/hivecommons/hive:candidate` @ `sha256:5e9abda0ff0ec5539f8bbc79e9a5a240a14997cf826f1e5fda9f0ed2b705ffa6`
- agent: `scanner`, mode `ISSUES_AND_PRS`, `kick_template: scanner-holdgated.md`
- backlog at time of measurement: 361 actionable issues, 302 open PRs

Related: #7363 (restart does not cancel a pending kick)

Danathar:
## Summary

Every kick prompt lists the agent's open issues and open PRs. The issue list has always been cut at 100, but the PR list was rendered in full — so on a hive with 302 open PRs the PR section alone was ~36 KiB and the whole kick came to 69.5 KiB, three times the size the code's own budget comment assumes. That much text takes ~3 minutes to type into the agent's terminal, costs tokens on every kick, and the agent never gets near the tail of the list anyway. This PR caps every PR list in the kick (default 50) with an explicit "… and N more" line so the agent knows the list is partial, and makes both caps a per-hive setting, `governor.kick_limits`, as the issue suggested.

Closes #7368. Shrinks the delivery window that makes #7363 unrecoverable in practice (fixed separately in #7378).

## What changed

- **`formatPRListWithPolicy`** (the `${PR_LIST}` every template uses) and the hardcoded scanner fallback's actionable-PR loop stop at the cap and append `  … and N more open PRs not listed (cap N per kick; they return on later kicks as this list drains)`. The section header still carries the true count.
- The scanner fallback's **stale-draft list** and the file-backed **merge-eligible** and **CI-failing** lists had the same unbounded shape; they share the cap and the marker.
- **`governor.kick_limits { max_issues, max_prs }`** — defaults 100 (the historical constant, unchanged) and 50. The existing issue cap sites (work list, lane lists, held-PR list, `issueRefsForAgent`) now read the same block through `Scheduler.issueCap()` / `prCap()`. Zero or negative means the default; there is deliberately no way to spell "uncapped". Documented in `hive.yaml.example` and `operator-reference.md`.
- The prompt-size budget comment in `pkg/dashboard/prompt_history.go` (which the issue quotes) now states the real PR term (≤ ~6 KiB at the 50 default) and the ~25 KiB worst case.

**Why 50 rather than mirroring 100:** the budget comment sized the section for 30; 50 keeps the worst-case prompt within a few KiB of the documented figure while leaving headroom over a typical backlog, and no agent works 50 PRs in one turn. Operators with a genuine need can raise it.

## Measured

Rendering the reported shape (302 dependabot-style PRs) through `formatPRList`:

| | lines | bytes |
|---|---|---|
| parent commit (`upstream/v4`) | 302 | 36,328 |
| this PR | 50 | 6,038 |

## Testing

Per CONTRIBUTING's test policy, the regression test fails without the fix: `TestFormatPRList_CappedWithOverflowMarker` does not compile on the parent (the cap symbols are new), and a symbol-free probe of the same assertion run on a pristine `upstream/v4` worktree failed with `PR list is unbounded: 302 lines, 36328 bytes` (that probe is not committed — it was only for the comparison). Also added: the scanner fallback capping both its PR and stale-draft lists; `governor.kick_limits` overriding both caps across the issue list, the scanner fallback and `issueRefsForAgent`, with zero/negative/nil falling back to defaults; merge-eligible and CI-failing capping with the marker; the config block loading from `hive.yaml`.

`go build ./...`, `go vet` on `pkg/scheduler`, `pkg/config`, `pkg/dashboard`, and `go test ./pkg/scheduler/ ./pkg/config/ -short` pass (the only failure in `pkg/config` is the pre-existing root-only `TestGateFailsClosedWithoutIP6Tables`).


clubanderson (approved):
Reviewed: caps all five unbounded list sites with explicit overflow markers, defaults preserve the historical issue cap, zero/negative fall back to defaults with no uncapped spelling, regression tests fail without the fix, docs updated. Chosen over #7380 (fewer sites) — UI exposure noted there as follow-up.

clubanderson:
Agreed this is the better vehicle — I've closed my #7380 in favour of it. Your extra two sites (file-backed merge-eligible and CI-failing) are ones I missed.

One gap worth flagging before this lands: **there's no UI.** The operator request on this was specifically that both caps be adjustable from **Settings → Repos**, not just `hive.yaml`:

> the cap for issues and prs should be adjustable in the settings on the repo tab

This PR adds `governor.kick_limits` plus `hive.yaml.example` and operator-reference docs, but no dashboard field and no API write path — so on a hosted spoke, where operators don't edit `hive.yaml` directly, the caps aren't reachable.

I have that half already built against my own config shape (Repos-tab number fields, `markDirty` wiring, effective-value read path, pointer-typed PUT so editing repos never resets the caps, and validation). I'll rebase it onto `governor.kick_limits` and open it as a follow-up once this merges — no need to grow this PR.

Two small things I'd raise on the config shape while it's still open:

1. **No ceiling.** `PRsPerKick()` clamps `<= 0` to the default but accepts any positive value, so `max_prs: 100000` is honoured and reproduces the original bug through the setting meant to prevent it. I'd clamp to something like 500.
2. **Default of 50 for PRs.** `prompt_history.go` sizes the PR section at *"~3.5 KiB at 30 open PRs x ~120 B per line"*, so 50 lands around 6 KiB — above the documented budget the same PR updates. Worth either using 30 or adjusting that budget line to match.

Neither is a blocker.

For the record on truncation safety, since it's easy to assume these lists are newest-first: they aren't. Issues sort oldest-first by age (`pkg/github/client.go:614`) and PRs sort by review class then oldest-first within class (`SortPullRequestsForReview`, `pkg/github/review_priority.go:185`), so the dropped tail is the newest, lowest-priority work. That's what makes the cap safe, and it's worth stating in the operator-reference text so nobody lowers the cap fearing it hides urgent items.

PR #7380, opened 2026-09-17
🐛 Cap the PR lists in kick prompts and make both caps configurable (#7368)

### LINKED ISSUE #7368: 🐛 PR list in kick prompts is uncapped while issues are capped at 100 — 302 PRs produced a 69.5 KiB kick (3x the documented budget)

clubanderson:
Closes #7368

## What was wrong

The kick prompt capped the **issue** list at 100 but emitted the **PR** lists in full, so prompt size tracked backlog size. On the reporting spoke (302 open PRs) the delivered kick measured **69.5 KiB** — over 3x the ~22 KiB worst case that `pkg/dashboard/prompt_history.go` sizes `promptHistoryMaxSizeMB` against.

## The issue named one unbounded loop; there were three

| site | path | was |
|---|---|---|
| `formatPRListWithPolicy` | `${PR_LIST}` in kick templates — **the live path on the reporting spoke** | unbounded |
| `buildScannerMessage` → ACTIONABLE PRs | legacy hardcoded fallback, still live for hives with no template | unbounded |
| `buildScannerMessage` → STALE DRAFT PRs | same fallback | unbounded |

I confirmed the reporting spoke (`kick_template: scanner-holdgated.md`) goes through `formatPRListWithPolicy`, so the issue did identify the site that produced its own 69.5 KiB measurement. The other two are still reachable and are fixed here too.

All three now emit an explicit `... N more open PRs not shown (list capped at N)` marker, so the agent is told the list is partial rather than silently concluding it has seen everything.

## Truncation is safe — and not for the reason you might assume

It is worth being precise, because "cap the list" sounds lossy. Neither list is in arrival order:

- **Issues** are sorted by `AgeMinutes` descending — **oldest first** (`pkg/github/client.go:614`).
- **PRs** are sorted by review class (fixes > refactors/docs > tests), then `CreatedAt` ascending — **oldest first within class** (`SortPullRequestsForReview`, `pkg/github/review_priority.go:185`).

So the cap keeps the oldest/highest-priority work and drops the **newest, lowest-priority** tail. It is not "the most recent N".

## Both caps are now operator settings

Per the issue's secondary suggestion, these are settings rather than bare constants — backlog size varies enormously between hives. They live on **Settings → Repos**, stated plainly in the UI:

> These caps limit how many issues and PRs are listed in an agent's kick prompt. They exist to keep the prompt a manageable size for the agents that read these lists — **scanner** above all, which receives both in full.

- `project.max_issues_per_kick` (default 100 — unchanged behaviour)
- `project.max_prs_per_kick` (default 30, taken from the PR-list line in the `prompt_history.go` budget: *"~3.5 KiB at 30 open PRs x ~120 B per line"*)
- Range 1–500. **Zero means "use the default", deliberately not "unlimited"** — an uncapped list is the bug these settings exist to prevent.
- The GET path sends the **effective** value, so the fields show the number actually in force rather than a blank box.
- Pointer-typed on the PUT path, so editing repos never resets the caps and vice versa. Validation runs before any mutation, so a rejected cap cannot half-apply.

## One adjacent correction

`formatHeldPRClaimsWithPolicy` is a **PR** list that was bounded by `maxIssuesPerKick`. It now uses the PR cap. Its `STAND DOWN this kick` marker reports the full omitted count either way, so the gate stays correct when the list is truncated.

## Verification

- `go test ./pkg/scheduler/` — ok
- `go test ./pkg/dashboard/` — ok (63 s)
- `go test ./pkg/config/` — only the known pre-existing `TestEntrypointHardensRuntimeConfigItRecreates` macOS chown failure, confirmed failing identically on a clean unmodified `v4`
- `go vet`, `gofmt` — clean on all changed files
- `.github/scripts/check-inline-js.js` over both dashboard trees — passed syntax and `no-undef`

**Mutation-tested — 11 mutations, all killed:**

| mutation | tests failed |
|---|---|
| PR cap removed (pre-fix state) | 4 |
| off-by-one `>=` → `>` | 3 |
| omission marker removed | 2 |
| scanner PR cap removed | 1 |
| stale-draft cap removed | 1 |
| zero treated as unlimited | 12 |
| ceiling clamp removed | 1 |
| `capsOnly` dropped from the empty-body guard | 3 |
| cap validation neutered | 6 |
| cap applied when key omitted | 1 |
| GET sends raw instead of effective cap | 1 |

Baseline after restoring every mutation: 0 failures.

Note one mutation (`capsOnly`) initially appeared to survive; it had actually failed to compile (unused variable) and my harness was counting a build failure as zero test failures. Re-run correctly, it kills 3 tests.

## Out of scope

The issue's related note about kick **delivery latency** (~3m09s, and its interaction with #7363) is not addressed here. Bounding the list shortens that window but does not fix the cancellation bug.


clubanderson:
Closing in favor of #7379, which covers the same three sites plus the two other unbounded lists (file-backed merge-eligible and CI-failing), threads both caps through the existing issue-cap sites via governor.kick_limits as #7368 suggested, and documents them in hive.yaml.example/operator-reference. The Settings → Repos UI exposure from this PR would make a good follow-up layered on the governor.kick_limits block.

Issue #7365, opened 2026-09-17
🐛 dashboard model list disagrees with the CLI — Copilot SDK discovery always fails (missing platform package), silently falling back to the chat-completions catalog


clubanderson:
## Summary

On the hosted spokes, the dashboard's Copilot model dropdown lists a **completely different set of models** from the ones the Copilot CLI actually offers. The dropdown shows a legacy chat-completions catalog (`gpt-4o`, `gpt-3.5-turbo`, …) while `/model` inside the very same agent's CLI lists the real agent catalog (`Claude Opus 5`, `GPT-5.6 Sol`, `Gemini 3.8 Flash`, `Grok 4.6`, `Kimi K3`, …).

Root cause: **probe #1 (the official-SDK helper) fails on every discovery cycle**, so discovery silently falls back to probe #2, the raw HTTP `GET <copilot-api>/models` — which returns a different catalog than the CLI uses.

## The failing probe

Every cycle, once per agent, in the spoke log:

```
level=INFO msg="copilot SDK model discovery unavailable, falling back to HTTP probe"
err="sdk helper: exit status 1 (stderr: copilot-models: Could not find a @github/copilot
platform package (tried @github/copilot-linux-arm64, @github/copilot-linuxmusl-arm64).
Searched 11 paths. Ensure @github/copilot is installed, or pass cliPath/cliUrl to
CopilotClient.)"
```

Confirmed in the running image — the platform package is genuinely absent:

```
$ ls /usr/local/lib/node_modules/@github
copilot
copilot-sdk          # <-- no copilot-linux-arm64

$ ls -l /usr/local/bin/copilot
/usr/local/bin/copilot -> ../lib/node_modules/@github/copilot/npm-loader.js
```

## Why this is NOT "arm64 is unsupported"

Three facts rule that out:

1. **The package exists upstream.** `@github/copilot-linux-arm64@1.0.85` and `@github/copilot-linuxmusl-arm64@1.0.85` are both published on npm.
2. **The CLI itself runs fine on this box.** The agents are actively driving `copilot` on these same arm64 nodes, and `/model` inside that CLI renders the full, current catalog. So the platform binary is reachable *at runtime for the agent*, just not to the SDK helper.
3. **The error message names the fix.** "Ensure `@github/copilot` is installed, **or pass `cliPath`/`cliUrl` to CopilotClient**" — and a working CLI entrypoint is sitting right there at `/usr/local/bin/copilot`.

[...]

Issue #4262, opened 2026-08-20
🐛 copilot: model nomenclature drift — catalog returns claude-fable.5 but CLI accepts claude-fable-5


clubanderson:
## Observed on a live spoke

Copilot CLI v1.0.78 was launched by hive with `--model claude-fable.5` (dotted) and replied:

```
Model "claude-fable.5" from --model flag is not available. Using "claude-sonnet-5" instead.
Model changed from claude-sonnet-5 (medium) to claude-fable-5 (medium)
```

So the CLI accepts **claude-fable-5** (dashed) but hive passed **claude-fable.5** (dotted).

## Root cause

`normalizeModelName` in `pkg/agent/manager.go` blindly rewrites a trailing `-<digits>` group to `.<digits>` for the copilot backend (a YAML-friendliness conversion from the `claude-sonnet-4.5` era). The Copilot CLI nomenclature has since drifted: the `-5` model family is DASHED (`claude-fable-5`, `claude-sonnet-5`, `claude-opus-5`) while older families remain DOTTED (`claude-opus-4.6`, `gpt-5.5`, `gemini-2.5-pro`). The blind rewrite corrupts every dashed-family id.

This is a recurring class of bug: the Copilot models catalog (`/models` endpoint and/or the SDK helper probe) periodically returns ids in different nomenclature than the copilot CLI `--model` flag accepts, in BOTH directions.

## Required fix

1. Alias-based canonicalization of Copilot model ids: match against the known CLI-accepted list separator-insensitively (`.` vs `-` drift only) and normalize to the known-good form; unknown ids and ids with legitimate dots pass through untouched.
2. Apply at launch time (before constructing `--model`) so already-stored bad ids self-correct on existing spokes without operator action.
3. Apply at discovery time so the dropdown shows canonical ids.
4. Model auto-heal must treat `claude-fable.5` and `claude-fable-5` as the same model — no false "model revoked" churn (the frontend `_normalizeModel` regex only handles `<digit>.<digit>$`, missing `fable.5`).

PR #7382, opened 2026-09-17
🐛 fix(dashboard): apply PUT /api/config/agent/{name}/models to the live launch config (#7374)

### LINKED ISSUE #7374: 🐛 PUT /api/config/agent/{name}/models persists but never applies — restart relaunches with the old model
## Summary

`PUT /api/config/agent/{name}/models` persists the new model to the agent's overlay file and returns `{"ok":true,"status":"updated"}`, but it does **not** update the live launch configuration. A subsequent agent restart spawns a **fresh process that still uses the old model**, even though the config on disk shows the new one.

This leaves config and runtime silently disagreeing, with a success response and no warning.

## Reproduction (observed on a v4 hosted spoke)

Six agents were pinned to `gpt-4o-mini-2024-07-18`. For five of them:

1. `PUT /api/config/agent/<agent>/models` with `{"backend":"copilot","model":"auto"}` → `200 {"ok":true,"status":"updated"}`
2. Overlay file confirms the write:
   ```
   /data/agent-configs/architect.yaml -> model: auto
   ```
   `/etc/hive/hive.yaml` also reads `model: auto`.
3. `POST /api/restart/<agent>` → `200`. Logs confirm a real restart:
   ```
   audit: agent restarting  name=quality reason=operator restart_count=41
   audit: agent restarted   agent=quality trigger=dashboard-api
   ```
4. The **new** process still carries the old flag. PIDs changed (14350 -> 50930 etc.), so this is a genuine respawn, not a stale process:
   ```
   uid=2001 pid=50930 model=gpt-4o-mini-2024-07-18
   uid=2002 pid=51344 model=gpt-4o-mini-2024-07-18
   uid=2003 pid=51822 model=gpt-4o-mini-2024-07-18
   uid=2004 pid=52153 model=gpt-4o-mini-2024-07-18
   ```

So: config says `auto`, a freshly launched process says `gpt-4o-mini-2024-07-18`.

## What does work

`POST /api/model/{agent}/{model}` — the endpoint the dashboard actually uses — persists *and* relaunches correctly. The frontend documents this distinction:

`pkg/dashboard/static/index.html`:
```js
switchModel(a.name, first);           // persists + relaunches via /api/model
```

Re-running the same change through `POST /api/model/<agent>/auto` immediately produced the intended result on all six:
```
uid=2001 pid=53554 running=auto
uid=2002 pid=53984 running=auto
uid=2003 pid=54412 running=auto
uid=2004 pid=54644 running=auto
uid=2005 pid=50251 running=auto
uid=2006 pid=55147 running=auto
```

## Why this matters

The model is a launch-time flag, so a fresh process is exactly when the new value should be picked up. Because it isn't, an operator who uses the config endpoint plus a restart gets a success response, correct-looking config, a real restart — and no behaviour change. The disagreement is only visible by inspecting `/proc/<pid>/cmdline`.

## Suggested fix

Either:

- have `PUT /api/config/agent/{name}/models` update the in-memory launch config the same way `/api/model` does (preferred — the endpoint's name implies it sets the model), or
- make the launch-command builder read the model from the overlay/config at spawn time, so any fresh process self-corrects, or
- if the config endpoint is deliberately persist-only, say so in the response (e.g. `"status":"persisted, not applied to running session"`) and document that `/api/model` is required to apply it.

Related: #7365 (bad model catalog) is what pinned these agents to `gpt-4o-mini-2024-07-18` in the first place; this issue is what made the correction appear to fail.

Danathar:
## Summary

Setting an agent's model through `PUT /api/config/agent/{name}/models` wrote the new value to disk and answered `{"ok":true,"status":"updated"}`, but the next restart launched the agent on the **old** model. The reason: the launcher prefers an agent's runtime model/backend *override* — left behind whenever `/api/model`, `/api/switch`, a pin, or the governor's auto-selection had set one — over the config value, and this endpoint only refreshed the config. So config said `auto`, a freshly spawned PID said `gpt-4o-mini-2024-07-18`, and nothing warned. This PR makes the endpoint apply the change to the live launch configuration the same way the agent-config dialog and the card dropdowns already do, restart the agent once when something launch-relevant actually changed, and say in its response what it did.

Closes #7374.

## Root cause

`launchInTmux` (`pkg/agent/manager.go`):

```go
model := agent.Config.Model
if agent.ModelOverride != "" {
    model = agent.ModelOverride
}
```

`handleAgentConfigModels` updated `Config.Agents[name]`, called `AgentMgr.UpdateConfig` (which only assigns `agent.Config`), persisted, and returned. `agent.ModelOverride` — set to `gpt-4o-mini-2024-07-18` by the earlier `/api/model` calls (#7365) — was never touched, so it kept winning at launch. `POST /api/model/{agent}/{model}` works because it calls `SetModelOverride` and restarts. `handleAgentConfigGeneral` (the dialog's other save path) already had the same fix in place for the same reason; the `/models` handler simply never got it. The published OpenAPI description even said *"Does not restart the agent itself; syncs the change into the live process config"* — that contract is what produced the bug.

## What changed

`pkg/dashboard/api.go`, `handleAgentConfigModels`:

- Compares the request against the agent's **effective** model/backend (runtime override first, then config) so a change is detected even when the config field already matched.
- When the model or backend changed, sets the override to the saved value (`SetModelOverride` / `SetBackendOverride`) — exactly what the general handler and the dropdowns do — so config, status, and the next launch agree.
- Restarts the agent **once** when the model, backend, or reasoning effort actually changed (all three are launch-time flags; the effort already lived in `agent.Config`, which is synced). An unchanged value neither restarts nor touches the override, so the dialog can re-save an unchanged form harmlessly.
- The response reports `applied: true` and `restarted: true|false`; if the restart itself fails the status says so explicitly and carries `restart_error`, instead of the bare `"updated"` that made this invisible (the issue's third suggestion, kept for the failure case).

`dashboard/openapi.json`: corrected description; documented `applied` / `restarted` / `restart_error`; added `reasoning_effort` to the request schema the handler already accepted.

## Testing

Per CONTRIBUTING's test policy, the new tests fail on the parent commit. Run on a pristine `upstream/v4` worktree:

```
--- FAIL: TestHandleAgentConfigModels_AppliesOverStaleOverride
    live ModelOverride = "gpt-4o-mini-2024-07-18", want auto — a restart would relaunch on the old model (#7374)
--- FAIL: TestHandleAgentConfigModels_BackendAppliesOverStaleOverride
    live overrides = ("copilot", ""), want (gemini, gemini-2.5-pro)
--- FAIL: TestHandleAgentConfigModels_NoChangeNoRestart
```

and pass on this branch. They mirror the report: seed the stale `gpt-4o-mini-2024-07-18` override, `PUT {"backend":"copilot","model":"auto"}`, assert the value the next launch will use. `go test ./pkg/dashboard/ -short` is fully green, `go vet` clean, `openapi.json` parses and the schema/route parity tests pass, and `src/scripts/check-api-reference-citations.sh` verifies (no route registration lines moved).

Not exercised here: a live tmux relaunch — this sandbox cannot open tmux sockets — but the restart call is the same `AgentMgr.Restart` the general handler and `/api/model` already make on the same code path.


clubanderson:
Duplicate-PR note first: I collided with you on this one — my #7383 was 68 s later and is now closed. Yours is the better change (the `reasoning_effort` handling, the openapi update, and `applied`/`restarted`/`restart_error` so a failed restart isn't a silent success). One substantive finding from the other branch, though:

### The order of the two `Set*Override` calls matters, and I think it's backwards here

This PR sets model first, then backend:

```go
if err := s.deps.AgentMgr.SetModelOverride(name, agentCfg.Model); err != nil {
...
if err := s.deps.AgentMgr.SetBackendOverride(name, agentCfg.Backend); err != nil {
```

`SetModelOverride` canonicalizes the model **only when the backend is already copilot** — `pkg/agent/manager.go:9993`:

```go
if agent.effectiveBackend() == "copilot" {
    model = CanonicalizeCopilotModel(model)
}
```

So on a request that switches an agent **to** copilot, `effectiveBackend()` still reports the *old* backend when `SetModelOverride` runs, the canonicalization is skipped, and a non-canonical id (the `claude-fable.5` vs `claude-fable-5` separator drift from #4262) is stored raw.

Reversing the two calls fixes it, and costs nothing: `SetBackendOverride` does not touch the model, and `SetModelOverride` does its own inference re-route afterwards reading `BackendOverride` (`manager.go:10018-10022`), so routing still lands on both new values:

```go
effectiveBackend := agent.Config.Backend
if agent.BackendOverride != "" {
    effectiveBackend = agent.BackendOverride
}
if m.routableBackend(effectiveBackend) && m.inferenceRouteCallback != nil {
    m.inferenceRouteCallback(name, effectiveBackend, model)
}
```

With the current order, `SetBackendOverride`'s own route callback fires first with the *old* model, and is then immediately corrected by `SetModelOverride` — harmless, just a redundant call.

### Severity: low, and self-limiting

The comment right above the canonicalization says launch re-applies it (`manager.go:9990-9992`), so a raw id self-corrects at spawn. The persisted override and the dropdown preselect are what stay un-canonical in the meantime. So this isn't a blocker — but since the fix is a two-line swap, it seems worth doing while you're in here.

Not tested on your branch — flagging from a read of the manager, so worth a second look before you act on it.

clubanderson (approved):
Reviewed: mirrors the general-section handler's override sync, compares against effective (override-first) values so a config-equal change is still detected, single restart only when launch-relevant fields changed, OpenAPI contract corrected, regression test included. Chosen over #7383 (no OpenAPI update).

PR #7383, opened 2026-09-17
🐛 Apply model/backend from PUT /api/config/agent/{name}/models to the live launch config (#7374)

### LINKED ISSUE #7374: 🐛 PUT /api/config/agent/{name}/models persists but never applies — restart relaunches with the old model

clubanderson:
Closes #7374

## Real root cause

The issue says the endpoint "does not update the live launch configuration". That is right in effect, but the mechanism is more specific than "it only writes to disk" — `handleAgentConfigModels` *already* called `AgentMgr.UpdateConfig(name, agentCfg)`, which does update the in-memory `AgentProcess.Config`.

The problem is that `AgentProcess.Config` is **not** what the launch command is built from:

```go
// pkg/agent/manager.go
func (a *AgentProcess) effectiveModel() string {
	if a.ModelOverride != "" {
		return a.ModelOverride
	}
	return a.Config.Model
}
```

`effectiveModel()` / `effectiveBackend()` prefer the per-agent **override** (`ModelOverride` / `BackendOverride`). Those overrides are set by `POST /api/model/{agent}/{model}` (the card dropdown), by the governor's auto-selection and by auto-heal — but never cleared or retargeted by the config endpoint. So on the reporter's spoke, where the agents had previously been pinned to `gpt-4o-mini-2024-07-18` (#7365), `ModelOverride` still held that value. The config PUT wrote `auto` into `hive.yaml` and `/data/agent-configs/<agent>.yaml`, the process `Config` was updated, and then every relaunch still resolved `effectiveModel() == "gpt-4o-mini-2024-07-18"`. Config and runtime disagreed exactly as described, silently.

`POST /api/model` worked because it goes through `SetModelOverride`, i.e. it writes the field the launcher reads.

## What changed

`src/pkg/dashboard/api.go` — `handleAgentConfigModels` now mirrors the config dialog's *general* section (`handleAgentConfigGeneral`, which already got this treatment):

* capture backend/model/reasoning_effort before the edits,
* after persisting, push a changed backend through `SetBackendOverride` and a changed model through `SetModelOverride` (backend first, because `SetModelOverride` canonicalizes copilot model ids against the effective backend, and it also retargets an existing pin),
* restart the session once if backend, model or reasoning effort changed — the model and the effort exist only on the launch command line, so a running CLI never picks them up in place (same rationale as `handleEffortSet`).

Only *real* changes apply an override or restart, so the dialog's save-everything path cannot bounce sessions on a no-op save.

## Verification

New tests in `src/pkg/dashboard/agent_config_models_apply_test.go`:

* `TestHandleAgentConfigModels_AppliesToLaunchConfig` — seeds a stale `ModelOverride` of `gpt-4o-mini-2024-07-18` (the reported starting state), PUTs `{"backend":"copilot","model":"auto"}`, then asserts via `GetStatus` that the live override is now `auto`/`copilot`.
* `TestHandleAgentConfigModels_NoChangeLeavesOverrides` — a PUT with the already-configured values must not rewrite the override (and hence must not restart).

Commands run (from `src/`, with an isolated `GOMODCACHE`):

```
go test ./pkg/dashboard/ -run AgentConfigModels -count=1   # ok
go test ./pkg/dashboard/ -count=1                          # ok  (67.4s, full package)
go vet ./pkg/dashboard/                                    # clean
gofmt -l pkg/dashboard/                                    # api.go and the new test not listed (the files it does list are pre-existing on v4)
```

### Mutation testing

1. Replaced the two `Set*Override` calls with `_ = backendChanged; _ = modelChanged` (i.e. the pre-fix behaviour):
   `TestHandleAgentConfigModels_AppliesToLaunchConfig` FAILED — `launch model override = "gpt-4o-mini-2024-07-18", want "auto"` and `launch backend override = "", want "copilot"`.
2. Forced the guards true (`... || true`) so every PUT applies:
   `TestHandleAgentConfigModels_NoChangeLeavesOverrides` FAILED — `unchanged PUT rewrote the override: got "sonnet", want "opus"`.

Both mutations restored afterwards; the committed tree is the fixed version and passes.

## Out of scope

* No change to `POST /api/model/{agent}/{model}` — it already behaves correctly.
* Did not alter the response shape (`{"ok":true,"status":"updated"}`); the endpoint now genuinely applies the change, so the "persisted, not applied" wording suggested as an alternative in the issue is no longer needed.
* `src/pkg/dashboard/server.go` untouched, so no `api-reference.md` citation drift.


clubanderson:
Closing as a duplicate of #7382, which was opened 68 seconds earlier (17:48:09Z vs 17:49:17Z) — we collided on the same issue.

#7382 is the better change and a strict superset of this one: same root cause (a stale `ModelOverride`/`BackendOverride` shadowing the saved config at launch), plus `reasoning_effort` handling, the `openapi.json` update, and `applied`/`restarted`/`restart_error` in the response so a failed restart is not a silent success.

Nothing here needs carrying over, but one detail from this branch is worth a look — I've left it as a comment on #7382: the order of the two `Set*Override` calls matters, and #7382 has them the other way round.

PR #7388, opened 2026-09-17
🐛 fix(dashboard): audit model discovery failures, label them, and keep auto-heal off non-authoritative lists (#7384)

### LINKED ISSUE #7384: 🐛 bug: model discovery failure is silent — no audit log entry, no dropdown error, and auto-heal downgrades agents off a fallback list
## Summary

When CLI model discovery fails, hive fails **silently**: there is no audit log entry, and the dashboard model dropdown renders a static fallback list that is visually indistinguishable from a real, live catalog. Auto-heal then acts on that bad data and downgrades agents.

This is not hypothetical — it is the root cause of a production incident today on the `projectbluefin` hosted spoke.

## What happened

1. On arm64 (OKE), `copilot-models.mjs` failed with `Could not find a @github/copilot platform package (tried @github/copilot-linux-arm64...)` — filed and fixed as #7365 / PR #7372.
2. Discovery therefore returned no live catalog and the UI fell back to the legacy chat-completions list.
3. The browser-side auto-heal (`_confirmAbsent` in `src/pkg/dashboard/static/index.html`) concluded that the agents' configured models were "absent" and switched **all six** bluefin agents to the weakest entry, `gpt-4o-mini-2024-07-18`.
4. **Nothing anywhere reported a problem.** Not the audit log, not the dashboard. The configs looked fine; only `/proc/<pid>/cmdline` on the running agents revealed the downgrade.
5. Blast radius: for roughly two days every PR the bluefin spoke opened was authored by `gpt-4o-mini`. This is very likely a major contributor to the downstream community complaint that "our hive agents keep opening PRs that can't pass CI."

The failure was only caught by manually diffing bluefin's agent config against a healthy spoke.

## Requested changes

**1. Audit log entry on discovery failure**

Emit a structured audit event whenever CLI model discovery fails or falls back, including: agent/backend, platform/arch, the underlying error text, and an explicit flag for whether a fallback list was served. Today this path is completely silent, which is what made a 2-day-old regression invisible.

**2. Surface the failure in the model dropdown**

The dropdown should show an explicit error state — e.g. `⚠ model discovery failed — showing fallback list` — rather than silently rendering stale/static entries that look authoritative. An operator seeing a plausible model list has no way to know discovery is broken.

**3. Do not let auto-heal act on a failed or fallback discovery**

This is the most important one. `cli_models.go` already implements deliberate hysteresis for exactly this class of problem — `cliModelDropAfterMisses = 3` with the comment noting the API returns "31 vs 35 chat-capable models across calls seconds apart," and that treating one thin sample as authoritative previously "made the dashboard auto-heal switch every agent off a model the account is in fact entitled to."

That protection only smooths **live** results. A hard discovery failure that serves a static fallback bypasses the retention logic entirely, and auto-heal treats the fallback as ground truth. Auto-heal must require a *successful* live probe before concluding a model is unavailable; absence in a fallback list must never trigger a downgrade.

## Relevant code

- `src/pkg/dashboard/cli_models.go` — `cliModelCacheTTL = 30 * time.Second`, `cliModelDropAfterMisses = 3`, with the rationale comment
- `src/pkg/dashboard/static/index.html` — `_confirmAbsent` auto-heal, runs browser-side
- `bin/copilot-models.mjs` — the discovery helper that failed

## Related

- #7365 / PR #7372 — the arm64 discovery failure that triggered this
- #7374 — `PUT /api/config/agent/{name}/models` persists but never applies to the live launch config

A useful diagnostic detail for testing: post-fix, the helper's *error text* is the signal. On an unauthenticated spoke it returns `Not authenticated` (meaning the SDK resolved and genuinely issued `models.list`), versus the pre-fix `Could not find a @github/copilot platform package`. Success and failure were distinguishable only by reading that string by hand — which is the observability gap this issue asks to close.

Danathar:
## Summary

When CLI model discovery broke on the projectbluefin spoke (the Copilot SDK helper failing on arm64, #7365), nothing said so: the dashboard's model dropdown showed a plausible-looking list, the audit log was silent, and the dashboard's auto-heal — trusting that list — moved all six agents onto `gpt-4o-mini-2024-07-18` for two days. This PR makes a discovery failure loud and harmless: it is written to the audit log, the dropdown labels it as a failure with the probe's own error text as the tooltip, and no list that isn't a clean live probe can ever trigger a downgrade.

Closes #7384.

## What actually happened (and why the existing guard didn't catch it)

`reconcileModelsAfterDiscovery` already refuses to heal off a `fallback` (static) list, and `stabilize()` already smooths noisy live samples. The incident slipped between them: when the *installed* SDK helper failed, `discoverCopilotModels` fell through to the raw-HTTP `/models` probe, which **succeeded** — a `fallback=false` "live" result that is the chat-completions catalog, not the CLI's. Every authority check keyed on `fallback` let it through: the dropdown rendered it as real, auto-heal saw the agents' CLI models "absent" from it and healed to `available[0]` (copilot has no vetted-target list, so that was `gpt-4o-mini`), and the only record was a WARN log line.

## Changes (all `pkg/dashboard`)

**1. Audit log entry on failure/fallback** — `queryCLIModels` writes `model_discovery_failed` (user `system`; detail: `backend`, `platform` os/arch, `fallback`, `degraded`, `served`, and the probe's token-free `error` text) when a backend's discovery fails or degrades, and `model_discovery_recovered` (with `previous=` the failure) when a clean probe answers again. Transitions only — the cache tracks the last audited state per backend — so the log gains one line when discovery breaks and one when it heals, not one every 30 s. Not-configured fallbacks (no credential, no helper installed) carry no error and are not failures, so they are unaudited, matching the existing notice policy. Documented in `docs/audit-log.md`.

**2. Dropdown error state** — `cliModelResult` gains `degraded` and `discoveryErr`; `/api/config/backends` carries `degraded` and `discovery: {ok:false, error, fallback, degraded}`. The dropdown labels entries `⚠ model discovery failed — showing fallback list` (or `⚠ model discovery degraded — not the CLI catalog`) and puts the error in the tooltip — exactly the string (`Could not find a @github/copilot platform package…` vs `Not authenticated`) the issue notes an operator previously had to read off a server log. An entitlement notice (#6500) still wins the label since it is the more specific diagnosis.

**3. Auto-heal requires a clean live probe** — the HTTP catalog served after an *installed* helper failed is now marked `degraded` (an absent helper — dev machines, CI — stays a clean live result, as today). A result is `authoritative()` only when neither fallback nor degraded, and that gates three places: `reconcileModelsAfterDiscovery(…, !!b.fallback || !!b.degraded)` skips it, `modelIDsForBackend` treats it as unverifiable (so `/api/model` doesn't reject CLI models missing from the HTTP catalog), and `stabilize()` is not fed it (so the HTTP inventory can't leak into the CLI catalog's retention). gemini/claude probe failures now carry their error text too; codex/agy/goose "binary absent" stays the documented normal state.

## Testing

`cli_models_discovery_failure_test.go` reproduces the incident shape — helper failing with the verbatim arm64 error, HTTP probe answering with `gpt-4o-mini-2024-07-18` — and pins: degraded-not-fallback with the error carried; kept out of validation and retention; absent helper + HTTP success stays live; no-token failure carries the error; the `/api/config/backends` wire shape (and no leak onto other backends); audit transitions (three identical failures → one entry, changed error text → a second, recovery → one `recovered` entry naming the previous failure); healthy and unconfigured hives audit nothing; and the dashboard strings plus the auto-heal gate. The new tests reference new symbols so they do not compile on the parent; the behavioural assertion (`degraded`, audit entries) is what the parent lacks.

`go test ./pkg/dashboard/ -short` fully green (37 s), `go vet` clean, `node .github/scripts/check-inline-js.js` passes on `index.html`, `openapi.json` parses.

Not exercised: the real helper on arm64 (fixed separately in #7372) — the seams used here are the same `runCopilotSDKHelper` / `copilotUserEndpointURL` the existing discovery tests use.


clubanderson (approved):
Model discovery failures audit — reviewed: complete site coverage, no sleeps added. LGTM.

PR #7391, opened 2026-09-17
🐛 fix(scheduler): scope held-PR snapshot cap and stand-down per repo

### LINKED ISSUE #7385: 🐛 bug: >100 held PRs permanently disables the scanner PR channel — STAND DOWN fires on render-cap overflow, not real conflict
## Summary

A spoke with more than 100 held PRs permanently disables its own scanner PR channel. The trigger is a **render cap overflow**, not any real conflict. This is a self-reinforcing deadlock: the condition that causes the stand-down is the backlog, and the stand-down prevents the backlog from being worked down.

Observed live on the `projectbluefin` hosted spoke, which has ~304 open PRs across 16 repos and a queue that never shrinks.

## Mechanism

`formatHeldPRClaimsWithPolicy` in `src/pkg/scheduler/scheduler.go` (~L1248-1262):

```go
for _, item := range actionable.Hold.Items {
    if item.Type != "pr" { continue }
    if shown >= maxIssuesPerKick {   // maxIssuesPerKick == 100
        omitted++
        continue
    }
    ...
    shown++
}
if omitted > 0 {
    b.WriteString(fmt.Sprintf("  ... %d additional open held PRs omitted; STAND DOWN this kick\n", omitted))
}
```

`omitted` is incremented **solely** because the list exceeded the display cap of 100. It carries no information about whether any ground is genuinely contested. Yet the emitted text instructs a full stand-down.

## Why the agent obeys (correctly)

The scanner's policy reads an omitted snapshot as "unseen occupied ground" — if it cannot see the complete set of held PRs, it cannot prove its intended target is disjoint from them, so it must not open new PRs. **That reasoning is sound.** The agent is behaving exactly as designed.

The defect is that hive truncates its own list and then reports the truncation to the agent as a trust signal. From the scanner's own session summary today:

> What actually blocked me: the PR channel. The hold-gated snapshot ended with "1 additional open held PRs omitted; STAND DOWN this kick" — and policy says an omitted snapshot means unseen occupied ground, so no new PRs. That is explicit, and I hold to it.

Note the count: **1**. A single PR past the cap zeroed out the entire PR channel across all 16 repos.

## Impact

- Any spoke crossing 100 held PRs stops opening PRs permanently.
- Because it can no longer open PRs, the backlog only grows, so `omitted` stays > 0 forever. There is no path out without operator intervention.
- The session that prompted this investigation produced: zero PRs, zero issues, zero comments.

## Suggested fixes

The display cap and the safety signal must be decoupled — they are currently the same variable.

1. **Compute the claim/disjointness check over the full `Hold.Items` set**, independent of how many are rendered. Truncating for prompt size is fine; truncating the *safety evaluation* is not.
2. If a genuine stand-down is warranted, base it on an actual overlap test against the agent's intended targets, not on list length.
3. If the full set truly cannot be evaluated, prefer a narrower restriction — e.g. stand down only for the repos that have held PRs — rather than a global halt across every repo.
4. Stop reusing `maxIssuesPerKick` as a held-PR cap. It is an issue-list constant (see #7368) and its reuse here silently couples two unrelated limits.

## Related

- #7368 — kick list caps (issues capped at 100, PRs uncapped)
- Test asserting the current string: `src/pkg/scheduler/held_pr_coordination_test.go:122`

clubanderson:
Fixes the deadlock in #7385.

## Problem

`formatHeldPRClaimsWithPolicy` capped the hold-gated coordination snapshot **globally** across all repos, reusing `maxIssuesPerKick` (100) to do it:

```go
if shown >= maxIssuesPerKick {
    omitted++
    continue
}
...
if omitted > 0 {
    b.WriteString("... STAND DOWN this kick")
}
```

So the stand-down fired on **render-cap overflow**, not on real contention. On the projectbluefin spoke a single repo pushed the combined list one item past 100 (`omitted = 1`), which stood the scanner down across all 16 tracked repos. Because it could no longer open PRs, the backlog causing the overflow could never drain — a self-reinforcing deadlock with no exit without operator intervention.

The agent's behavior was correct throughout; it was obeying a signal that carried no real information. From its session log:

> What actually blocked me: the PR channel. The hold-gated snapshot ended with "1 additional open held PRs omitted; STAND DOWN this kick" — and policy says an omitted snapshot means unseen occupied ground, so no new PRs. That is explicit, and I hold to it.

## Fix

Disjointness is only ever evaluated against PRs **in the repo the agent is about to touch**, so a global cap is the wrong scope. This change:

- adds a dedicated `maxHeldPRsPerRepoPerKick` constant and applies the cap per repository, decoupling it from `maxIssuesPerKick` (which is an issue-list limit — see #7368)
- groups the snapshot by repo, preserving first-seen order
- makes the omission notice name its repo: `STAND DOWN for <repo> this kick`
- updates the coordination prompt so the agent stands down for the flagged repo and knows fully-listed repos remain workable

## Fail-closed is preserved

This is deliberately **not** a relaxation of the safety property. The agent still refuses to act anywhere the snapshot is incomplete — `TestHeldPRCoordinationFailsClosedWhenSnapshotDoesNotFit` still passes, updated only for the repo-scoped wording. What changes is blast radius: one crowded repo no longer blanks the snapshot for repos that *are* fully listed.

## Tests

- updated `TestHeldPRCoordinationFailsClosedWhenSnapshotDoesNotFit`
- added `TestHeldPRCoordinationTruncationIsScopedToTheCrowdedRepo`, which pins the actual regression: an over-cap repo must fail closed while a quiet repo in the same sweep still renders and is not stood down

`go build ./...` clean; `go test ./pkg/scheduler/ -count=1` passes in full.

Refs #7385

Closes #7385


clubanderson:
Rebased over v4 head: #7379's kick_limits change had replaced the maxIssuesPerKick reference this PR removes with s.issueCap(); resolution keeps this PR's per-repo grouping (the render cap here is deliberately maxHeldPRsPerRepoPerKick, decoupled from the issue cap). Build, vet, and pkg/scheduler tests pass locally.

clubanderson:
Scoping the cap per repo is clearly right — a single crowded repo blanking the snapshot for all 16 was the actual deadlock, and the regression test pins exactly that.

One thing worth a second look, because it cuts across #7368 which landed in parallel:

The previous code took this limit from a **configurable** source, while this PR introduces `const maxHeldPRsPerRepoPerKick = 100`. Decoupling the held-PR list from the *issue* cap is correct — it's a PR list and never belonged on the issue cap — but the replacement is hardcoded, so the net effect is that operators lose the ability to tune it at the same moment #7379 made every other kick list tunable via `governor.kick_limits`, and #7395 puts those on **Settings → Repos**.

That leaves three kick-list limits with two different policies, which is the kind of split that's much cheaper to fix now than after it ships. Since `governor.kick_limits.max_prs` already exists and the held-PR list *is* a PR list, sourcing the per-repo cap from `KickLimits.PRsPerKick()` would fold it in without new config surface. Not a blocker, and if you'd rather keep it fixed for fail-closed predictability that's a defensible call — worth stating in the comment either way so the next reader doesn't "fix" it by accident.

For the record on truncation safety, since it's easy to assume otherwise: these lists are not newest-first. Issues sort oldest-first by age (`pkg/github/client.go:614`) and PRs by review class then oldest-first within class (`SortPullRequestsForReview`, `pkg/github/review_priority.go:185`), so a cut always drops the newest, lowest-priority tail.

clubanderson:
## This is now blocking a second agent

Adding live evidence, since this PR has been waiting on review: the truncated held-PR snapshot is no longer just a scanner problem. **`quality` is now fully idle for the same reason.**

Pane capture (two samples 90s apart, unchanged, both ending at the CLI prompt):

```text
● STAND DOWN.

Preflight snapshot truncated — "1 additional open held PRs omitted". Unseen occupied
ground can't be proven disjoint against any candidate cluster. Policy: stand down this kick.

No issue opened, no PR opened, no bead created, no hold touched.
```

Note the magnitude: **one** omitted PR is enough to stand the agent down completely. The preflight tells the agent that any truncation means unseen ground cannot be proven disjoint, so the agent — correctly following policy — refuses to do anything at all.

This also rules out the "weak model" explanation for quality being stuck. It was previously running `Auto → Claude Haiku 4.5`; it is now pinned to `claude-opus-5` (confirmed in `/proc`: `copilot --model claude-opus-5`, and `/api/status` reports `pinnedBoth: true`). The stand-down persists on Opus. It is the snapshot, not the model.

Worth stating the shape of the deadlock plainly, because it's self-reinforcing:

- hive's own top recommendation on this spoke reads `open_holds=109`
- >100 held PRs is exactly what trips the #7385 STAND DOWN
- so hive advises draining held PRs **while its scheduler forbids the agents from opening anything, because too many are held**

The per-repo cap in this PR breaks that loop by keeping the snapshot complete within each repo rather than truncating globally.

/cc for review — this is currently costing us two agent lanes.

PR #7396, opened 2026-09-17
🐛 fix(scheduler): make the repair queue fork-aware — comment-only for PRs the hive cannot push to (#7386)

### LINKED ISSUE #7386: 🐛 bug: repair queue has no fork awareness — scanner burns sessions on fork PRs it cannot push to, and can create stray branches
## Summary

The CI-failing "repair queue" handed to the scanner contains PRs from **forks**, which the agent has no permission to push to. Nothing in the kick marks which PRs are cross-repository, so the agent can only discover this by attempting a push and failing. On a large spoke this wastes an entire session and can leave stray branches behind.

`grep -rni "isCrossRepository\|cross_repository\|is_fork\|fromFork" --include=*.go` across the repo returns **zero matches** — there is no fork awareness anywhere in the scheduler or GitHub layer.

## Observed today (projectbluefin spoke)

The scanner was given 107 CI-failing PRs. From its own session accounting:

> **The work list.** 120+ actionable issues sat untouched. I spent the whole session on the CI-failing list, and that list turned out to be other people's PRs (many from forks) that I cannot push to.

> **Fork check.** I should have tested `isCrossRepository` across the 107 up front — that would have collapsed the repair queue in two minutes instead of after a bad push.

The "bad push" it refers to produced a stray branch on `projectbluefin/testsuite`:

1. `git push origin HEAD:sec-check-dashboard` — accidental stray branch, created because the fork PR #839 head name collided
2. `git push origin --delete sec-check-dashboard` — self-reverted

So the failure mode is not merely wasted effort; it caused an unintended write to a shared repo under a branch name belonging to a *different* agent (`sec-check`), because the agent was reasoning about a head ref it did not actually control.

## Impact

- Sessions burn their full budget on an unreachable queue while actionable issues go untouched.
- Push attempts against fork head refs can collide with local branch names and create stray branches.
- The repair queue's apparent size is badly misleading — 107 "fixable" PRs may be mostly unfixable by this agent.

## Suggested fixes

1. Capture `isCrossRepository` (or `head.repo.full_name != base.repo.full_name`) when fetching PRs and store it on the PR item.
2. **Either** filter fork PRs out of the push-repair queue entirely, **or** render them in a clearly separate section, e.g. `FORK PRs (review/comment only — you cannot push)`.
3. Annotate each queue entry with the reachable action so the agent does not have to probe: push-fixable vs. comment-only.
4. Guard the push path: never push to a branch name derived from a fork PR head ref without confirming the head repo is the same repo.

## Related

- #7385 — held-PR stand-down deadlock (same session, same spoke)

Danathar:
## Summary

The hive's repair queue treated every CI-failing PR as something an agent could fix by pushing — but the hive's App token can only push to the base repository, and a PR whose head branch lives in a contributor's fork cannot be pushed to at all. Nothing recorded which PRs were forks (`grep -rni isCrossRepository` returned nothing), so on the projectbluefin spoke a scanner was handed 107 red PRs, 66 of them forks, discovered this by pushing, burned its session on an unreachable queue while 120+ issues sat idle, and the push left a stray `sec-check-dashboard` branch on the base repo under the fork's head-ref name. This PR computes fork origin once at scan time and carries it through every queue an agent reads, so "comment only" is known before any checkout or push.

Closes #7386.

## What changed

All four suggested fixes from the issue, at the data source rather than in the prompt:

1. **Capture it** — `github.PullRequest` gains `HeadRef`, `HeadRepo`, `FromFork`, populated in `fetchPRs` by `prHeadOrigin` (head vs base `repo.full_name`, case-insensitive, GitHub's `isCrossRepository`); a head whose repository was deleted counts as a fork, since it is equally unpushable. `ReachableAction(pr)` renders it as `push` / `comment-only`.
2. **Separate section** — the `CI_FAILING` work list now renders two queues: push-repairable first (`(none you can push to)` when empty, so an all-fork queue doesn't look empty by accident), then `FORK PRs (N — review/comment only, you CANNOT push to these)`, each with `[fork: <owner/repo>:<head_ref> — comment only]`.
3. **Annotate every entry** — `ci-failing.json` rows carry `head_ref`, `head_repo`, `from_fork` and an explicit `reachable_action`; `${PR_LIST}` and the scanner fallback list annotate fork PRs inline as `[fork: <owner/repo> — comment only, cannot push]`.
4. **Guard the push path** — the fork section states outright: *never `git push origin HEAD:<head_ref>` — that creates a stray branch on the base repo under a name you do not own* (the exact command from the incident), and the head repo is printed next to every fork PR so the agent is never reasoning about a ref it doesn't control.

**The compounding factor from the issue's comment.** `FIX-BEFORE-NEW` defaults unattributed red PRs to the scanner, so 66 contributor fork PRs became a gate the scanner was obliged to clear before any new work — and structurally could not. `formatRedPRFixData` now never lists a fork PR as the agent's own regardless of attribution, says how many were withheld and why, and raises no gate at all when only fork PRs are red.

**Not changed, deliberately:** `merge-eligible.json` (merging a fork PR is a base-repo write the App *can* perform), and nothing at the git level intercepts a push — the guard here is classification plus explicit kick text. A pre-push hook in agent checkouts would be a new mechanism (agent checkouts get no hooks today) and is left as a follow-up if maintainers want a hard stop.

## Testing

Per CONTRIBUTING's test policy the regression is demonstrated on the parent: a symbol-free probe of `TestFormatRedPRFixData_ForkPRsNeverYours` run on a pristine `upstream/v4` worktree fails with `unattributed fork PR #839 was listed as scanner's FIX-BEFORE-NEW work` (probe not committed — the committed test file references the new `FromFork` field and doesn't compile on the parent). Also added: a `prHeadOrigin` table (same repo, case-insensitive, fork, deleted fork, nil, abbreviated payload), `ReachableAction`, `fetchPRs` end to end from a REST fixture with head/base objects, `ci-failing.json` carrying the fields with `reachable_action`, the `CI_FAILING` split with the warning text, and the inline annotation in both PR lists.

`go test ./pkg/github/ ./pkg/scheduler/ ./cmd/hive/ -short`: scheduler and cmd/hive green; the only `pkg/github` failure is the pre-existing sandbox-only `TestHiveOpenPRScript_MultilineBodyIsValidJSONWithoutPython3` (needs `dirname` on a stripped PATH). `go vet` clean on all three.


clubanderson (approved):
Fork-aware repair queue reviewed: comment-only path for unpushable fork PRs, no sleeps, clean seam. LGTM.

PR #7401, opened 2026-09-17
🐛 fix(scheduler): report a dangling kick_template instead of silently falling back (#7390)

### LINKED ISSUE #7390: 🐛 bug: a dangling kick_template resolves to nothing silently — blank prompt editor, 404 link, no warning, and saving it creates a live override
## Summary

An agent configured with a `kick_template` that does not resolve fails **silently**. The dashboard prompt editor renders an empty textarea, the "Kick template:" link 404s, and nothing in the logs says the template was not found. Worse, typing into that empty editor and saving creates a real override that *does* take effect — so a box that looks broken is actually live.

Observed on the `projectbluefin` hosted spoke, whose `review` agent has `kick_template: review.md`.

## Evidence

`src/pkg/policies/defaults/` contains no `review*.md` — not `review.md`, not `review-full.md`, not `review-holdgated.md`. Every other agent has one (`scanner-holdgated.md`, `quality-advisory.md`, `sec-check-full.md`, etc.). The UI link therefore 404s:

> The `d34e4c5939f0fbadf2137e2fcbf07aee18c1d254` branch of hive does not contain the path `src/pkg/policies/defaults/review.md`.

`grep -rn 'review\.md' --include=*.go src/` returns no policy-template reference — the only hits are unrelated (`acmm_criteria.go` patterns, skills tests).

Review prompts are in fact built **in Go**, not from a kick template: `src/pkg/review/prompts.go:31` emits `[review-perspective:%s]`. Live kicks confirm it — `[review-perspective:security]`, `[review-perspective:correctness]`. So the `kick_template` field on `review` is inert and probably should not be set at all.

## Mechanism

`loadNamedTemplate` (`src/pkg/scheduler/scheduler.go:218-243`) tries the user override dir, the cloned examples dir, the configured local dir, then embedded defaults — and if all miss:

```go
return ""
```

No error return, no log line. The caller (`scheduler.go:800-809`) then silently falls through:

```go
if agentCfg, ok := s.cfg.Agents[baseName]; ok && agentCfg.KickTemplate != "" {
    if template := s.loadNamedTemplate(agentCfg.KickTemplate); template != "" {
        s.logger.Info("using config kick_template", ...)
        ...
    }
}
// falls through to ACMM pack default with no indication the configured template was missing
```

Note the asymmetry: the **success** path logs `using config kick_template`, but the failure path logs nothing at all.

## Why this is more than cosmetic

`handleAgentPromptSave` writes the edited template to `/data/policies/<KickTemplate>`, which is the **first** path `loadNamedTemplate` checks. So an operator who sees the blank editor, assumes the template was lost, types something and saves, silently promotes a dangling reference into an active override — changing agent behavior via a control that appeared non-functional.

## Requested changes

1. **Log a warning when a configured `kick_template` does not resolve**, naming the agent, the template, and the paths tried. The success path already logs; the failure path must too.
2. **Surface it in the prompt editor** — show `template not found: review.md (falling back to <actual source>)` rather than an empty box, and do not render a repo link for a path that does not exist.
3. **Validate `kick_template` at config load / overlay-reject time**, alongside the existing invalid-overlay rejection, so a dangling name is caught when set rather than never.
4. Separately: consider clearing `kick_template` on `review`, since the review swarm does not consume kick templates.

## Related

- #7385 — held-PR stand-down deadlock (same spoke)
- #7386 — repair queue has no fork awareness (same spoke)

Danathar:
## Summary

When an agent's `kick_template` names a file that exists nowhere, the hive said nothing: the kick quietly fell through to the pack/convention template, the dashboard's Prompt Template tab showed an empty editor with a repo link that 404'd, and — because the save path writes to the *first* directory the loader checks — typing into that empty box and saving turned the dangling reference into a live override. This PR makes the miss visible everywhere it matters (boot log, per-kick log, the API, the editor), stops the editor linking to files the repo doesn't ship, and refuses to *set* a new name that doesn't resolve.

Closes #7390.

## What changed

**1. Warn when the template does not resolve (issue item 1).** `resolveNamedTemplate` now returns the content, the source that served it, and every path tried; `loadNamedTemplate` wraps it. `BuildAgentMessage` logs the miss at WARN — agent, template, the fallback it will actually use, and the paths tried — with the same weight the success line has (`using config kick_template` now also names its source). `WarnDanglingKickTemplates()` runs once at boot (wired in `cmd/hive/main.go`) and warns per agent.

**2. Surface it in the prompt editor (item 2).** `GET /api/config/agent/{name}/prompt` gains `template` — the scheduler's own `TemplateResolution` (`kickTemplate`, `resolved`, `source`, `fallback`, `pathsTried`, `embeddedDefaultExists`) — and omits the repo `url` from `sourceFiles` when `pkg/policies/defaults` does not ship the file. The Prompt Template tab renders `⚠ template not found: review.md — kicks fall back to <fallback>`, the paths tried, and a note that saving creates `/data/policies/<name>` as a live override (clear `kick_template` in General to keep the fallback); the source-files list shows the missing path struck through with no link.

**3. Validate when set (item 3).** `PUT /api/config/agent/{name}/general` refuses a **new** `kickTemplate` that resolves nowhere (400 naming the missing file and that saving the prompt first creates it). An unchanged dangling value is deliberately left alone so a stale field can't block an unrelated display-name/model edit. At config load and in the per-agent overlay guard, `ValidateKickTemplateName` rejects a value carrying a path separator or `..` — the loader joins the value under the policy dirs, so a path would read outside them. Whether a bare name *resolves* stays a warning rather than a load error on purpose: a template can legitimately be created later by the prompt editor, and `hive-quickstart.yaml` ships names (`scanner-CLAUDE.md`, `sec-check.md`, …) served only by the cloned `examples/kubestellar/agents/` checkout — rejecting overlays for those would drop working agents.

**4. `review`'s `kick_template` (item 4).** No shipped config or ACMM pack sets `kick_template` on a `review` agent — it lives in the projectbluefin spoke's own config — so there is nothing to clear in this repo. The docs now say agents whose prompts are built in Go (the review swarm's `[review-perspective:…]` kicks) should carry no `kick_template`; with this PR the spoke's boot log and editor will both point at it.

Docs: `agent-configuration.md` (a "dangling template" paragraph after the resolution order), `openapi.json` description, changelog fragment.

## Testing

Per CONTRIBUTING's test policy the regression is demonstrated on the parent: a symbol-free probe (build a kick for an agent with `kick_template: review.md`, expect a WARN) run on a pristine `upstream/v4` worktree fails with an **empty** log — literally nothing was written — and passes on this branch. Committed tests cover: `resolveNamedTemplate` provenance (embedded source; dangling → paths tried starting at the override dir; a user override wins and is named), `ResolveTemplate` for dangling / shipped / unset agents plus `TemplateExists` and nil-safety, the WARN on a miss and its absence on a hit, `WarnDanglingKickTemplates`, `ValidateKickTemplateName` and the overlay guard, the prompt endpoint's resolution and no-link behaviour, the general endpoint's refuse-new / allow-unchanged / allow-shipped / allow-clear matrix, and the dashboard strings.

`go build ./...`; `go test ./pkg/scheduler/ ./pkg/dashboard/ ./pkg/config/ -short` — scheduler and dashboard fully green, config green except the pre-existing root-only `TestGateFailsClosedWithoutIP6Tables`; `go vet` clean; `node .github/scripts/check-inline-js.js` passes on `index.html`; `openapi.json` parses and the parity tests pass.


clubanderson (approved):
Dangling kick_template visibility reviewed: warn-on-load/validate-on-set split is right (quickstart names resolve only after clone), path-separator rejection closes the traversal read, editor stops linking 404s. LGTM.

PR #7408, opened 2026-09-17
🐛 fix(bin): hive-open-issue honours --dry-run and refuses unknown flags instead of dropping them (#7400)

### LINKED ISSUE #7400: 🐛 bug: hive-open-issue wrapper silently ignores --dry-run and files a real issue an agent cannot retract
## Summary

The `hive-open-issue` wrapper silently ignores `--dry-run` and creates a real issue. An agent that used the flag as a safe capability probe published a live issue with the body `placeholder` into `hivecommons/hive`, and then could not clean it up.

## What happened

The scanner agent on the `projectbluefin` spoke ran `gh issue create --dry-run` expecting a no-op. The wrapper submitted it, producing **#7393** with a literal `placeholder` body.

The agent could not repair its own mistake: its scoped token **can create** issues on this repo but returns **403 on both edit and comment**. So the only recovery available to it was to file a second, correct issue (**#7397**) that opens with `> **Supersedes #7393.**` and to ask a human to close the first. In its own words:

> `gh issue create --dry-run` is **not** honored by the `hive-open-issue` wrapper — my access probe filed hive#7393 with the body "placeholder". The token can create issues on that repo but not edit or comment (403 on both), so I could not repair it. #7397 supersedes it and asks for #7393 to be closed. Worth a wrapper fix: it should reject or pass through `--dry-run`.

#7393 has now been closed as superseded.

## Why it matters

`--dry-run` is the conventional way to probe a capability without side effects. A wrapper that accepts the flag and ignores it converts a safe probe into an unrecoverable write — unrecoverable specifically because the create and edit scopes are asymmetric. Every agent that probes this way will leak a junk issue and be unable to retract it.

The asymmetry compounds the damage: if the token could comment or edit, the agent would have self-repaired and no human action would have been needed.

## Requested changes

1. **Honor `--dry-run`**: validate and report what *would* be created, then exit without writing.
2. **Or reject it explicitly**: fail with a clear "unsupported flag" error. Either is acceptable; silently dropping a flag whose entire purpose is to prevent writes is not.
3. Audit the wrapper for other silently-dropped `gh` flags, since the same swallow pattern would hide any of them.
4. Consider granting agents edit/comment on issues they themselves authored, so a bad create is self-recoverable.

## Related

- #7393 — the placeholder issue this produced (closed, not planned)
- #7397 — the correct issue filed to supersede it

Danathar:
## Summary

`hive-open-issue` — the relay every agent's `gh issue create` is redirected through — ended its argument parser with a catch-all that silently discarded any flag it didn't recognise and wrote the request anyway. A scanner probing its access with `gh issue create --dry-run` therefore filed a real issue whose body was `placeholder` (#7393), and because its scoped token can create issues but not edit or comment on them, it could not retract it. This PR makes `--dry-run` do what it says on every shape, and makes any other unknown flag a hard error instead of a silent drop.

Closes #7400.

## What changed (`bin/hive-open-issue.sh`)

**`--dry-run` / `-n` is honoured (requested change 1).** On `issue`, `comment`, `claim` and `close` it validates exactly as a real run would (a missing `--body` still exits 2), prints the request JSON that *would* have been queued and the path it would have gone to, and exits 0 — without creating the request directory or writing a file.

**Unknown flags are refused (changes 2 and 3 — the audit).** The catch-all was the only place flags could be swallowed; it now applies only to the positional issue number/URL. Anything else that looks like a flag exits 2 with `unsupported flag: <flag> (supported: …) … nothing was created`. This closes the same hole for every gh flag whose purpose is to change or prevent a write — `--edit-last`, `--create-if-none`, `--recover` — not just `--dry-run`. The documented gh no-op flags (`--assignee`, `--milestone`, `--project`, `--template`, `--web`, `--editor`, plus their `=` forms) stay accepted and ignored, so existing agent invocations are unaffected.

**Self-repair path (change 4).** Granting agents edit/comment scope is a token-policy change beyond this wrapper, but the retraction path already exists and is now documented: `hive-open-issue close --repo <owner/repo> <number>` runs in the watcher with the App token, and I checked that the reporter-confirmation close gate (`humanFiledBugReason`) only applies to *human-filed* bug issues — an agent can close its own mistake without any edit scope. `src/docs/hive-open-issue.md` gains the `--dry-run` row, the refuse-unknown rule, and that retraction note.

The `gh` wrapper passes the agent's arguments straight through (`exec hive-open-issue "${args[@]}" …`), so `gh issue create --dry-run` now reaches the honoured flag with no wrapper change.

## Testing

`bin/test_hive_open_issue.sh` — 44 cases, 0 failures (26 pre-existing, contract untouched, plus 18 new): the incident shape (`--dry-run` create writes nothing and prints the request with its body), `-n` and flag-first ordering, validation under dry-run, dry-run on comment/claim/close, unknown long/short flags refused with no request written and the flag named in the error, and the tolerated gh flags still creating. Per CONTRIBUTING's test policy: the same suite run against the parent commit's shim fails 11 of 44, first among them **"dry-run create writes no request"** — the parent wrote it.


clubanderson (approved):
hive-open-issue --dry-run + unknown-flag hard error reviewed; closes the silent-drop that filed #7393's placeholder issue. LGTM.

PR #7413, opened 2026-09-17
🐛 fix(dashboard): never fabricate an all-zero config when the session expires

### LINKED ISSUE #7405: 🐛 Agent config dialog renders a fabricated all-zero config when the session expires
## What happened

Opening any agent's **Configuration → Cadences** tab rendered a fully populated, authoritative-looking config of **all zeros** — `Agent power: 0 = disabled` and idle/quiet/busy/surge all `0` — while the Governor cadence table *on the same page* simultaneously showed the true values for that agent (`scanner`: idle 4h, quiet 4h, busy 2h, surge 2h).

Three toasts appeared: `Failed to load config: Load failed`, `Could not refresh terminal access: Load failed`, `Could not prepare terminal access: Load failed`.

Actual cause: **the GitHub session had expired.** Refreshing and signing in again fixed it completely. Nothing was wrong with the agent or its config.

## Root cause

`loadConfigData` swallows every failure and substitutes an empty object:

```js
async function loadConfigData(type, agentName) {
  try {
    const url = type === 'governor' ? '/api/config/governor' : `/api/config/agent/${agentName}`;
    const res = await fetch(url);
    if (!res.ok) throw new Error(`HTTP ${res.status}`);
    _configState.data = await res.json();
  } catch (err) {
    showToast(`Failed to load config: ${err.message}`, 'error');
    _configState.data = {};          // <-- fabricated config
  }
}
```
`src/pkg/dashboard/static/index.html:17906-17916`

The dialog then opens and renders from `{}`. Every field falls back to its zero value, so a **failed load is indistinguishable from a real config in which the agent is disabled and every cadence is off.** The toast is the only signal, and it is generic.

## Why the existing 401 handling never fires

Several call sites already handle auth expiry properly — e.g. the terminal path at `index.html:6275-6276`:

```js
if (resp.status === 401) showToast('Sign in to this hive before opening a terminal.', 'error');
else if (resp.status === 403) showToast('Your current hive role cannot open terminals.', 'error');
```

But in this incident even *that* site printed `Could not refresh terminal access: Load failed` — the `e.message` branch at `:6279`, not the 401 branch. So the fetch never saw a 401 status; it **rejected** at the network layer. `Load failed` is Safari's generic `TypeError` for a fetch that could not complete.

That means no amount of `res.status === 401` checking can detect session expiry here. The status-code checks scattered through the dashboard are dead code on this path.

## Severity

Not silently destructive: `saveConfig` (`index.html:23846`) sends only `_configState.dirty`, so merely opening the dialog and pressing Save does **not** write the phantom zeros back.

The harm is diagnostic. An operator who opens this dialog is told, with no hedging, that the agent is **disabled with all cadences at zero**. That is a false baseline, and it invites "fixing" a problem that does not exist — or worse, editing one field while believing the other three really are `0`.

This is the same failure shape as #7384 and #7390: a component degrades, produces a default, and then presents the default as ground truth.

## Suggested fix

1. **Do not fabricate config.** On failure, leave `_configState.data` unset and render the dialog body as an explicit error state ("Could not load configuration — not showing values") with **Save disabled**. Never render zeros that the server did not send.
2. **Detect session expiry independent of status code.** A rejected fetch on an authenticated endpoint should trigger a session probe; if unauthenticated, surface `Your session expired — sign in again` with a sign-in action, instead of `Load failed`.
3. **Make the API path return a real 401.** If an expired session currently produces a redirect toward the identity provider, API routes under `/api/` should return `401` with a JSON body rather than a cross-origin redirect, so client status checks work as intended.
4. Route all three toasts through that shared handler so one expired session yields **one** actionable message, not three generic ones.

## Reproduction

1. Sign in to a hosted spoke dashboard.
2. Let the GitHub session expire (or clear the session cookie).
3. Open any agent's ⚙️ → **Cadences**.
4. Observe all-zero cadences, `Agent power 0`, and generic `Load failed` toasts — while the Governor table above still shows the correct cadences.

Observed on `hosted-projectbluefin-knuckle-gjvq`, `candidate (v4)` `cb3a49e`, Safari.

Danathar:
## Summary

When a hosted hive's hub session expired, opening an agent's config dialog showed a config that was never the agent's: every cadence `0`, stale timeout `0`, restart strategy off, model blank — with **Save** enabled, so one click would have persisted those zeros. The only hint was three toasts that each said `Load failed`. The cause is a chain: nginx turns the hub auth-check's 401 into a cross-origin redirect to the login page, `fetch()` cannot follow that, Safari reports `TypeError: Load failed`, none of the dashboard's `res.status === 401` branches ever fire, and `loadConfigData` answered the rejection with an empty `{}` that every tab rendered as real values. This PR removes the fabrication, makes an expired session say so with a sign-in action, funnels the three toasts through one handler, and makes the hosted `/api` routes answer a real `401` instead of a redirect.

Closes #7405.

## What changed

**1. No fabricated config (`static/index.html`).** `loadConfigData` no longer falls back to `_configState.data = {}`; it records `loadError` (`'session'` or `'error'`). `openConfigDialog` and `renderConfigTab` then render an explicit *Could not load configuration — not showing values* state (with a **Sign in** or **Retry** action) and never a tab from the placeholder. `syncRepoSaveGuard` disables Save while nothing is loaded and `saveConfig` refuses a racy click.

**2. Session expiry detected independent of status code.** A new shared handler `reportFetchFailure(context, err)`: on a rejected fetch (or an explicit 401/403) it runs one session probe against `/api/gh-user-auth/status` — shared across concurrent callers — and classifies the session as ok / expired / unknown from whatever it gets back: an explicit 401/403 (the gate), `logged_in:false` after this page had an identity (device-flow session gone), or a rejected probe on a hosted host (the redirect itself). Expired ⇒ a single sticky toast *Your session expired — sign in again* with a **Sign in** button (hosted: a full navigation re-enters the login redirect, which the browser *can* follow; self-hosted: the GitHub device flow). Anything else is reported under the caller's context, with a network failure phrased as "could not reach this hive".

**3. One handler, one message.** `loadConfigData`, `renewTerminalAssertion` and `createTerminalHandoff` all go through `reportFetchFailure`; the toast is deduplicated, so one expired session yields one actionable message instead of `Failed to load config: Load failed` + `Could not refresh terminal access: Load failed` + `Could not prepare terminal access: Load failed`.

**4. `/api` answers a real 401 on hosted spokes (`saas_provision.go`, `deploy/k8s/error-pages.yaml`).** A new `hive-api-xhr` Ingress covers the `/api` prefix with the *same* per-hive `auth-url` and `auth-response-headers` as the page Ingress but **no `auth-signin`**, and intercepts `401,502,503` into `hive-error-pages`. The error backend gains `unauthorized.json` (`{"ok":false,"error":"unauthorized",…}`) and an HTML twin (for a browser that navigated to an `/api/` URL directly — it links back to `/`, where the login redirect still lives), maps `error_page 401` onto them and reproduces `X-Code: 401` instead of the 503 fallback. `/api/v1` (bearer-token) and `/api/contribute` (public) are longer prefixes and keep their own ungated Ingresses; the page and terminal Ingresses keep `auth-signin`. The vanity-host patch loop includes the new Ingress. `docs/security-model.md` documents the page-vs-API distinction.

**Rollout note.** The Ingress is provisioned per hive by the hub; the `hive-error-pages` ConfigMaps live in `ingress-nginx` and are applied by operators — re-apply `deploy/k8s/error-pages.yaml` to get the 401 document (until then an intercepted 401 gets the backend's 503 fallback, which the client still handles as an ordinary failure, not as zeros).

## Testing

- `pkg/dashboard/config_dialog_no_fabricated_config_test.go` (3 tests): the `{}` fallback is gone and `loadError` is recorded; the error state exists and guards both the initial render and every tab switch; Save is disabled and `saveConfig` refuses; the shared handler, probe and message exist; the probe reads 401/403 and a rejected hosted probe as expired; all three sites call the handler and none raises its own toast; the toast deduplicates.
- `pkg/hub/saas_api_xhr_ingress_test.go` (4 tests): `hive-api-xhr` is exactly `Prefix /api`, carries the per-hive auth gate and identity headers, has no `auth-signin`, intercepts 401 into `hive-error-pages` (both TLS modes); page/terminal Ingresses keep `auth-signin` and `/api/v1`/`/api/contribute` stay ungated; the vanity patch loop lists it; the error backend returns 401, maps it onto `unauthorized.{json,html}`, and the JSON is valid with `ok:false` / `error:"unauthorized"`.
- Per CONTRIBUTING's test policy, the same test files run against the parent commit (`9d2b1ff0`) fail 3 of 3 (dashboard) and 3 of 4 (hub — the sibling-Ingress test is a guard that also holds on the parent).
- `go test ./pkg/dashboard/ ./pkg/hub/` green; `check-inline-js.js` on `index.html` passes. `saas_api_ingress_test.go` matched `name: hive-api` as a prefix and picked up the new Ingress, so it now matches `name: hive-api\n` exactly; the wildcard-TLS helpers know the fifth Ingress.


clubanderson (approved):
No-fabricated-config chain reviewed: loadError states with sign-in/retry, status-code-independent expiry probe with deduped toast, and the hive-api-xhr Ingress giving /api a real 401 (no auth-signin, error-pages intercept, /api/v1 + /api/contribute untouched). Tests fail-on-parent per policy. LGTM.

PR #7418, opened 2026-09-17
🐛 fix(dashboard): stop shipping ci-maintainer's CI/coverage strip to other agents (#7411)

### LINKED ISSUE #7411: 🐛 Newly created agents inherit ci-maintainer's CI/coverage stat strip (reviewer shows a fabricated COVERAGE 0% vs goal 91%)
## What happened

The `reviewer` agent's card renders a full CI/coverage indicator strip that has nothing to do with it:

```
COVERAGE  0% current   goal: 91%   ● Brew ● Helm  0% CI ● Weekly ● Nightly Tests
● Compliance ● Dashboard ● gh-aw ● Playwright ● Nightly Rel ● Weekly Rel ● vLLM-d ● PokProd
```

`reviewer` is an **ADVISORY, on-demand** agent. It owns no repository, runs no CI, and produces structured review verdicts — it has no coverage and no release pipeline. The card reports `COVERAGE 0% against goal: 91%`, which is not a measurement of anything.

## Evidence

`GET /api/config/agent/reviewer` returns a `stats` array of exactly 14 entries:

```
['coverage', 'brew', 'helm', 'ci', 'weekly', 'nightly', 'nightlyCompliance',
 'nightlyDashboard', 'nightlyGhaw', 'nightlyPlaywright', 'nightlyRel',
 'weeklyRel', 'deploy_vllm_d', 'deploy_pok_prod']
```

That is, verbatim and in order, the `"ci-maintainer"` entry of `defaultStatsConfig` — `src/pkg/dashboard/status_builder.go:1141-1155`. The coverage entry matches field-for-field, including the target that produces the rendered `goal: 91%`:

```json
{"field":"coverage","key":"coverage","label":"Coverage","source":"agentMetrics","style":"pct-bar","target":91}
```

For contrast, `GET /api/config/agent/scanner` returns its own correct three:

```
['actionable', 'openPrs', 'mergeable']
```

## Why this is not the defaults fallback

`defaultStatsConfig` is keyed by agent name and returns **empty** for any name it does not know (`status_builder.go:1171-1175`):

```go
if d, ok := defaults[name]; ok {
    return d
}
return []any{}
```

`reviewer` is not a key in that map, so the defaults path would correctly yield no stats at all. The 14 entries are therefore coming from the agent's **own persisted `StatsConfig`** — they were written into the reviewer's stored config when the agent was created (via `POST /api/agents`), cloning ci-maintainer's set rather than starting empty.

`CollectAgentStats` (`status_builder.go`) keys results by agent name, so this populates `AgentMetrics["reviewer"]["coverage"] = 0` separately from `AgentMetrics["ci-maintainer"]["coverage"]`. The ACMM coverage gate reads the ci-maintainer key (`api_acmm_recommendation.go:24,161`), so this does **not** appear to corrupt the ACMM evaluation. The damage is display-level — but it is a fabricated number presented as a measurement.

## The internal inconsistency

The very same card already knows how to say "no data". Directly above the coverage strip it renders:

```
LAST KICK  —          AVG/PASS  —
```

`—` for the fields with nothing to report, but `0%` for coverage. A `pct-bar` with no underlying metric should not resolve to `0%` against a `91%` goal; `0%` is indistinguishable from a real, catastrophic coverage result. Same failure shape as #7384 / #7390 / #7405: absence of data rendered as a confident value.

## Suggested fix

1. **Do not clone another agent's stat set at creation.** A newly created agent should persist `stats: []` (or the defaults for its own name) — never ci-maintainer's CI strip. This is the actual bug.
2. **Scope stats to role.** Stats with `"source": "health"` describe the primary repo's workflows and only make sense for the agent that owns CI. An ADVISORY/on-demand agent should not be offered them at all.
3. **Never render a `pct-bar` with no value.** In `renderStatHtml` (`src/pkg/dashboard/static/index.html:6064`, `pct-bar` branch at ~6088) the code does `const pct = v || 0`, which turns missing/undefined into a hard `0%`. Render `—` when the metric is absent, matching LAST KICK and AVG/PASS on the same card.
4. Clean up the already-persisted `stats` on the existing `reviewer` agent on this spoke.

## Environment

Observed on `hosted-projectbluefin-knuckle-gjvq`, `candidate (v4)` `cb3a49e`, ACMM L5. The `reviewer` agent was created through `POST /api/agents` with `role: reviewer`, `mode: ADVISORY`, `on_demand: true`.

clubanderson:
Closes #7411

## Root cause — different from the one reported

The issue concluded that `POST /api/agents` clones ci-maintainer's stat set into a new agent's persisted config. That is not what happens: `handleAgentCreate` (`src/pkg/dashboard/api_agents.go`) never writes `stats.json`, and the only writer is the explicit `PUT /api/config/agent/{name}/stats` editor.

The 14 entries came from a **shipped seed**: `src/deploy/data/agents/reviewer/stats.json` was a byte-for-byte copy of `src/deploy/data/agents/ci-maintainer/stats.json` (verified with `diff` — identical). `src/Dockerfile:510` copies `src/deploy/data/` to `/opt/hive/seed-data/`, and `src/deploy/entrypoint.sh:866-869` does `cp -rn /opt/hive/seed-data/* /data/` on every start. So `/data/agents/reviewer/stats.json` exists on every spoke before any agent named `reviewer` is created, and `loadStatsConfig` reads it ahead of the (correctly empty) name-keyed defaults. `reviewer` is a retired roster name — `entrypoint.sh:874` already calls it "the retired `reviewer` seed".

The rest of the issue's analysis holds: the defaults map is not the source, `AgentMetrics["reviewer"]["coverage"]` is never collected (`metrics_collector.go:108` only populates `ci-maintainer`), and the ACMM gate reads the ci-maintainer key, so the damage is display-level — a fabricated number presented as a measurement.

## Fix

1. **Remove the retired seed** `src/deploy/data/agents/reviewer/stats.json`. This is the actual creation-path leak; the other four seeds (`scanner`, `architect`, `outreach`, `ci-maintainer`) were checked and are correct for their own agents.
2. **Scope `source: "health"` stats to the CI owner on read** (`scopeHealthStats` / `agentOwnsCIHealth` in `status_builder.go`, applied in both `FrontendAgent` builders and in `loadAgentStats`). Health stats describe the primary repo's workflows, not a per-agent measurement. Doing this on read repairs spokes where the seed already landed in `/data` — `cp -rn` never overwrites, so deleting the seed alone would not have helped them — without rewriting the operator's file. The owner is the agent named `ci-maintainer`, an agent whose `role` is `ci-maintainer`, or a replica of it, using the same key the ACMM coverage gate uses.
3. **Never render a percentage that was not measured.** `renderStatHtml` in `src/pkg/dashboard/static/index.html` did `const pct = v || 0` (`pct-bar`) and `${v || 0}%` (`pct` — the same shape, one stat over; the issue named only `pct-bar`). Both now treat only a finite number as a measurement and otherwise render `—` with a "not measured" label and an empty bar, matching LAST KICK / AVG PASS on the same card. A genuine `0` still renders `0%`.

Item 4 of the issue (cleaning the persisted file on the affected spoke) is handled by the read-time scoping rather than by a destructive migration of operator data.

`dashboard/index.html` has the same `v || 0` shape but is the stale non-shipped copy, so it is left alone.

## Validation

All commands run from the clone, `GOMODCACHE` pinned:

- `go test ./pkg/dashboard/ -count=1` → `ok github.com/hivecommons/hive/pkg/dashboard 70.617s`
- `go test ./pkg/dashboard/ -run 'SeedStats|ScopeHealth|RenderStatHtml' -count=1 -v` → 6 new tests PASS (the node-backed one ran, it did not skip)
- `go vet ./pkg/dashboard/` → clean
- `gofmt -l` on the three touched Go files → no output
- `find dashboard src/pkg/dashboard/static -name '*.html' -print0 | xargs -0 node .github/scripts/check-inline-js.js` → "All script blocks passed syntax and no-undef checks."

One pre-existing nil-`deps` test (`TestLoadAgentStats_NonexistentFile`) panicked on the first draft because `loadAgentStats` reached `s.deps.Config` on a bare `NewServer`; the config lookup is now nil-guarded and the suite is green.

## Mutation evidence

Each mutation was applied, the test run, then reverted. Build failures were checked for explicitly (one mutation produced `declared and not used` and was re-done in a compilable form before being counted).

| Mutation | Result |
| --- | --- |
| `measured = true` in `renderStatHtml` (restores `v \|\| 0` behaviour) | `TestRenderStatHtmlDoesNotFabricateZeroPercent` FAILS — unmeasured pct-bar renders `NaN%`/no dash, unmeasured pct renders no dash |
| Restore `src/deploy/data/agents/reviewer/stats.json` from HEAD | `TestSeedStatsShipHealthStripOnlyToCIOwner` and `TestSeedStatsHaveNoRetiredReviewerCopy` FAIL (13 health stats + the coverage stat flagged) |
| `scopeHealthStats` short-circuited to `return stats` | all three `TestScopeHealthStats*` FAIL |
| `agentOwnsCIHealth` role/replica check → `return false` | `TestScopeHealthStatsKeepsHealthForCIOwner` FAILS (renamed CI agent and replica lose their strip) — the guard against over-filtering has teeth too |

## Files changed

- `src/deploy/data/agents/reviewer/stats.json` (deleted)
- `src/pkg/dashboard/status_builder.go`
- `src/pkg/dashboard/api.go`
- `src/pkg/dashboard/static/index.html`
- `src/pkg/dashboard/agent_stats_scope_7411_test.go` (new)
- `changelog.d/fixed-7411-agent-stat-strip-inheritance.md` (new)


clubanderson:
Closing in favour of #7425, which found the same root cause (the shipped reviewer seed byte-copy) and goes further on the issue's fix list: boot-time healing that removes the verbatim 14-key strip already copied into /data on affected spokes (cp -rn means read-time scoping alone leaves the stale file forever), role-based scoping (ADVISORY/on-demand) rather than name/role match on ci-maintainer only, a per-agent stat-sources catalogue with a '(not offered to this agent)' affordance, the same pct/pct-bar '—' fix, and docs. Salvageable from here: the mutation-evidence rigor and the nil-guard on s.deps.Config in loadAgentStats — if that panic path still exists after #7425, it deserves its own small PR.

PR #7420, opened 2026-09-17
🐛 Stop settings dialog hanging on Loading and fabricating a zeroed config

### LINKED ISSUE #7417: 🐛 Global Manager.mu held across tmux/subprocess restart stalls every agent settings dialog (11s+ lock holds)
## Symptom

Opening any agent's Configuration dialog frequently hangs on `Loading…` forever. Saving settings is intermittently very slow or appears not to take effect. The operator's words:

> there is nothing more frustrating than not being able to see settings immediately AND needing to set things multiple times for them to take effect

It is intermittent: sometimes instant, sometimes a hard timeout.

## Measured behaviour

Probes taken from inside the pod during a hang:

```
/api/config/agent/reviewer         curl exit 28  (TIMED OUT at 15s)
/api/config/agent/ci-maintainer    http=200  t=3.853s
/api/config/governor               http=200  t=0.0055s   <-- fast
/api/status                        http=200  t=0.015s    <-- fast
```

Minutes earlier, healthy:

```
/api/config/agent/scanner   http=200 size=83159 t=0.0036s
/api/config/agent/reviewer  http=200 size=2985  t=0.0025s
/api/config/agent/quality   http=200 size=84474 t=0.0043s
```

Same auth, same session, same process. Only `/api/config/agent/{name}` degrades.

## Root cause

`GET /api/config/agent/{name}` blocks acquiring a **read** lock on the global agent-manager mutex, while the model-switch/restart path holds the **write** lock across synchronous subprocess, tmux and sleep work.

The mutex is `Manager.mu sync.RWMutex` — `src/pkg/agent/manager.go:480-483`.

**The read side.** `src/pkg/dashboard/api.go:135` routes `GET /api/config/agent/{name}` to `handleAgentConfigGet`, whose very first action is the contended call (`api.go:3330-3339`):

```go
func (s *Server) handleAgentConfigGet(w http.ResponseWriter, r *http.Request) {
    name := r.PathValue("name")
    ...
    proc, err := s.deps.AgentMgr.GetStatus(name)
```

`GetStatus` waits on the global lock (`src/pkg/agent/manager.go:5277-5286`):

```go
func (m *Manager) GetStatus(name string) (*AgentProcess, error) {
    m.mu.RLock()
    defer m.mu.RUnlock()
    ...
}
```

**The write side.** `POST /api/model/{agent}/{model}` (`api.go:92`) persists the model and then restarts the agent **synchronously, before responding** (`api.go:2167-2185`):

```go
if err := s.deps.AgentMgr.SetModelOverride(name, model); err != nil { ... }
...
if err := s.deps.AgentMgr.Restart(s.deps.Ctx, name); err != nil { ... }
```

`restartWithReason` takes `m.mu.Lock()` at `src/pkg/agent/manager_restart.go:706-707` and, apart from a brief release for token minting at `:797-800`, holds it through to `launchInTmux` at `:809`:

```go
m.mu.Lock()
defer m.mu.Unlock()
...
m.mu.Unlock()
m.mintAgentTokenUnlocked(ctx, agent)
m.mu.Lock()
...
return m.launchInTmux(ctx, agent)
```

**What is held under that write lock** — all of it synchronous, none of it in-memory:

| Work | Location |
|---|---|
| `C-c` through tmux before teardown | `manager_restart.go:721-724` |
| capture/archive turn output | `manager_restart.go:728-732` |
| reap and possibly kill agent-owned processes | `manager_restart.go:749-755` |
| `tmux kill-session` | `manager_restart.go:760` |
| `ensureTmuxSession` (shells out, `cmd.CombinedOutput()`) | `manager_restart.go:766` → `manager.go:2292-2314` |
| `launchInTmux` under lock | `manager_restart.go:809` |
| `installCavemanForAgent` | `manager.go:2831-2833` |
| `tmuxPaneHasCLIForAgent` | `manager.go:2902` |
| `time.Sleep(preLaunchShellClearDelay)` | `manager.go:2946-2949` |
| `applySecretEnv`, `send-keys`, further sleeps, literal command typing, Enter | `manager.go:2971-2988` |

`src/pkg/agent/terminal.go:106-115` confirms these are real subprocess invocations (`t.m.tmuxCmd(...).Run()`).

## Live evidence: an 11.25 second lock hold

```json
{"time":"2026-09-17T14:58:45.793432102-04:00","msg":"audit: agent restarting","name":"ci-maintainer","restart_count":22}
{"time":"2026-09-17T14:58:46.314600627-04:00","msg":"tmux session created","name":"ci-maintainer"}
{"time":"2026-09-17T14:58:46.56179836-04:00","msg":"installing caveman","agent":"ci-maintainer","backend":"copilot","mode":"lite"}
{"time":"2026-09-17T14:58:57.048225311-04:00","msg":"audit: agent started","name":"ci-maintainer","model":"claude-opus-5"}
```

`14:58:45.793` → `14:58:57.048` = **11.25s** with the global write lock held. Every settings dialog opened in that window blocks. Three restarts were issued in under 30 seconds:

```json
{"time":"...14:58:18.417...","msg":"audit: agent restarting","name":"ci-maintainer","restart_count":20}
{"time":"...14:58:19.389...","msg":"audit: agent restarting","name":"ci-maintainer","restart_count":21}
{"time":"...14:58:45.793...","msg":"audit: agent restarting","name":"ci-maintainer","restart_count":22}
```

A config read landing behind several of these exceeds a 15s timeout — exactly what was observed.

## Why the gradient looks the way it does

- **`/api/status` unaffected (15ms).** `handleStatus` serves the cached payload under `Server.statusMu.RLock()` and never calls `AgentMgr.GetStatus` — `src/pkg/dashboard/server.go:2646-2656`.
- **`/api/config/governor` unaffected (5ms).** It does not go through the agent manager.
- **`/api/config/agent/*` degrades.** Every agent config read calls `GetStatus`. **Agents do not have independent locks** — one global mutex serialises all of them, so restarting *one* agent stalls the settings dialog for *every* agent.
- **`reviewer` timed out hardest** not because it is special — the handler treats it identically — but because it landed behind the global lock at the worst moment.
- **Switching tabs sometimes "fixes" it** because the frontend fetch has no timeout or abort. When the blocked request finally settles, `finally(... switchConfigTab(initialTab))` renders against whatever tab is now current.

## Disproved hypothesis

The config GET does **not** synchronously query live tmux or network state. After `GetStatus` it only reads config/file-backed data: restrictions via `os.ReadFile` (`api.go:3479-3508`), prompt template (`api.go:3566`), stats (`api.go:3894`), in-memory pipeline/hooks helpers (`api.go:6518`, `:6531`). The stall is entirely the first lock acquisition.

## This is not fixed by upgrading

This spoke runs `candidate (v4)` at `cb3a49e` and is 9 behind `a9f5ed2`. Upgrading does **not** fix it:

- `a9f5ed28` (#7374) fixes a *different* bug — `PUT /api/config/agent/{name}/models` persisted config without updating the live override. Its diff **adds another `AgentMgr.GetStatus` and a conditional `AgentMgr.Restart`** to the config PUT path, so it adds contention rather than removing it.
- `05ccb7c5` (#7363) fixes restart/kick replay loops; restart still takes `m.mu.Lock()` and still calls `launchInTmux` under it.
- `origin/v5` has the same structure: unlock for token/bob preflight, re-lock, then `return m.launchInTmux(ctx, agent)` under `Manager.mu`.

No existing issue covers this. #7374 is the closest and is a distinct bug.

## Recommended fix

**Backend — stop holding `Manager.mu` across I/O.** Restart should be phased:

1. Under `m.mu`: mark the agent restarting, capture immutable launch data, increment the launch generation, update state — then **release**.
2. Outside `m.mu`: `tmux kill-session`, `ensureTmuxSession`, token/preflight, `applySecretEnv`, `send-keys`, sleeps, caveman install, gateway/model setup.
3. Re-acquire `m.mu` only to verify the launch generation still matches and publish final state.

The launch-generation check already exists (`genBeforeMint`), so the pattern is established — it just needs to cover the whole launch rather than only the token-mint window.

Add a regression test that blocks a fake restart subprocess hook and asserts `GetStatus` / `GET /api/config/agent/{name}` still return promptly.

Consider also making `POST /api/model/{agent}/{model}` return once the override is persisted and the restart is *scheduled*, rather than blocking the HTTP response on a full tmux relaunch.

**Frontend — never let a slow server become an infinite spinner.** Wrap config fetches in an `AbortController` with a finite timeout, abort stale fetches when the dialog closes or another agent/tab is opened, and render an explicit error state with Retry instead of leaving `Loading…` standing. (Related: #7405, where a failed load is rendered as a fabricated all-zero config.)

## Environment

`hosted-projectbluefin-knuckle-gjvq`, `candidate (v4)` `cb3a49e`, ACMM L5, 6 agents. `pprof` is not exposed (`/debug/pprof/...` on 3001 returns the SPA; 3002 returns 404) — worth enabling behind a flag, since it would have answered this in one request.

clubanderson:
## Problem

Two complaints, one dialog:

1. **It hangs.** Opening agent settings frequently sits on `Loading…` forever. Root cause is backend (#7417): `handleAgentConfigGet` calls `GetStatus`, which takes the **global** `Manager.mu` — and `restartWithReason` holds that same lock across the entire tmux relaunch. Measured hold: **11.25s**. Restarting one agent stalls settings for *every* agent. The frontend had no timeout, so it just waited.

2. **It lies.** When the request fails outright, `loadConfigData` swallowed the error and set `_configState.data = {}`. The dialog then rendered a confident, fully-populated-looking config of **all zeros** — agent power `0 = disabled`, every cadence `0` — while the Governor table on the same page showed the real values (#7405). An expired GitHub session rejects at the *network* layer (Safari: `TypeError: Load failed`), not with a 401, so every `res.status === 401` check in the dashboard is dead code on that path.

Together these produce the operator experience: *"not being able to see settings immediately AND needing to set things multiple times for them to take effect."*

## Change

Frontend only. Makes the stall **visible and recoverable** instead of silently wrong.

- `loadConfigData` — `AbortController` with a 12s timeout; a `_configLoadSeq` staleness guard so a slow in-flight response can't overwrite a newer dialog; records `loadFailed`/`loadError` rather than fabricating `{}`.
- `renderConfigLoadError` — explicit error state that shows **no values at all**, disables Save (so you can't write zeros back), and offers Retry.
- `closeConfigDialog` — bumps the sequence and aborts any in-flight fetch.

`saveConfig` only sends `_configState.dirty`, so the zeroed render was never mass-overwriting config on save — but it was still driving operators to "correct" settings that were never actually wrong.

## Scope

This does **not** fix #7417. The lock still stalls the endpoint; this just means you see `Timed out` + Retry in 12s instead of an infinite spinner. The backend fix — phasing `restartWithReason` so the tmux relaunch happens outside `m.mu` — is tracked separately in #7417.

## Validation

- `go build ./...` clean
- `go test ./pkg/dashboard/ -count=1` → **ok 76.5s**
- Inline dashboard JS extracted and `node --check`'d clean

Note: `TestPromptTemplateStateResetOnOpenAndClose` asserts on an **exact literal** `_configState = { … }` initializer string, so new state fields must be inserted before `promptTemplateDirty` or the test fails on a pure substring miss.

Refs #7417, #7405


clubanderson:
Closing in favour of #7413, which fixes the same fabrication plus the layer underneath it: hosted /api now answers a real 401 (new hive-api-xhr Ingress + error-pages 401 document), session expiry is detected independent of status code with a single actionable sign-in toast, and it ships tests + security-model docs. Salvageable from here and worth re-cutting atop #7413 as the #7417 UX follow-up: the 12s AbortController timeout, the _configLoadSeq staleness guard, and abort-on-close — #7413 has no client timeout, so the lock stall in #7417 still shows as a long spinner until the backend fix lands.

PR #7425, opened 2026-09-17
🐛 fix(dashboard): stop seeding ci-maintainer's stat strip onto a new reviewer agent

### LINKED ISSUE #7411: 🐛 Newly created agents inherit ci-maintainer's CI/coverage stat strip (reviewer shows a fabricated COVERAGE 0% vs goal 91%)

Danathar:
## Summary

A freshly created `reviewer` agent — ADVISORY, on-demand, owning no repository — showed ci-maintainer's entire CI/coverage strip on its card, headed by `COVERAGE 0% current, goal: 91%`. That number measured nothing. The issue traced the 14 stats to the agent's persisted `stats.json` and guessed the create API had cloned them; the real source is one step earlier: the image shipped `src/deploy/data/agents/reviewer/stats.json`, a retired seed that was a byte-copy of ci-maintainer's set, and the entrypoint copies every seed into `/data` on boot (`cp -rn`). Any spoke that later created an agent named `reviewer` inherited the file. This PR removes the seed, heals spokes that already have the copy, scopes the CI health strip to agents that can own CI, and stops a missing percentage from rendering as `0%`.

Closes #7411.

## What changed

**1. The seed is gone (the actual bug).** `deploy/data/agents/reviewer/stats.json` is deleted. `TestSeededStatsNeverCloneCIMaintainer` parses every remaining `deploy/data/agents/*/stats.json` and fails if any seed other than ci-maintainer's carries that strip (or a coverage stat at all), and if the reviewer seed ever comes back. The entrypoint comment that cited the "retired reviewer seed" as an example now says why it is gone.

**2. Existing spokes heal themselves (issue fix 4).** `cp -rn` never removes what it copied, so the observed spoke still has the file. The three stats readers (`loadStatsConfig`, `LoadStatsConfigWithCfg`, `loadAgentStats`) now share one `readAgentStatsFile`, which recognises ci-maintainer's strip — the same 14 keys in the same order — on any agent other than ci-maintainer, treats it as absent and removes the file (one log line per agent). ci-maintainer's own file and any custom or merely overlapping set are untouched. `agentStatsDataDir` is a variable so this is tested against a temp dir rather than `/data`.

**3. `health` stats are scoped to role (fix 2).** `healthStatsApply` is false for an `ADVISORY` or `on_demand` agent — those checks describe the primary repo's pipelines, not the agent. `GET /api/config/stat-sources` takes `?agent=<name>`, and `GET /api/config/agent/{name}` now carries a `statSources` catalogue scoped to that agent, which the Stats tab renders from synchronously (the page-wide catalogue stays as the fallback). A leftover row whose source is no longer offered stays visible, labelled *(not offered to this agent)*, so it can be removed.

**4. No value ≠ 0% (fix 3).** In `renderStatHtml`, `pct` and `pct-bar` render `—` with an empty bar and no colour class when the metric is absent — the same convention as LAST KICK / AVG/PASS on the card — instead of `const pct = v || 0` turning "no measurement" into a red `0%` against the goal.

`docs/agent-configuration.md` notes the scoping and the `—` rule on `stats_display`.

## Testing

`pkg/dashboard/agent_stats_not_cloned_test.go` (5 tests): the seed invariant; reader pruning through all three readers in both file shapes (bare array = the seed, `{"stats":[…]}` = the API), with ci-maintainer's own file and a custom reviewer set kept; `healthStatsApply` / `statSourcesFor` across scheduled, autonomous, advisory, on-demand and the reviewer-as-created shapes; the `?agent=` endpoint and the agent-config catalogue (and that a fresh reviewer has zero stats); and the renderer strings. Per CONTRIBUTING's test policy: on the parent commit the seed and renderer assertions fail (`reviewer seed is ci-maintainer's strip verbatim: [coverage brew helm ci …]`, `pct-bar renders an absent value as 0%`); the other three exercise helpers this change introduces.

`go test ./pkg/dashboard/` green, `check-inline-js.js` passes, `check-api-reference-citations.sh` verified (no route registrations moved).


clubanderson:
Heads-up from #7411's owner clarification, posted here since #7418 was closed in favour of this PR (agreed — the boot-time healing and the seed removal are the right core).

> **No stats should be listed in the reviewer at all.** … do not replace it with a "not measured" placeholder either — render **nothing**: no strip, no empty container.

Two residual paths in this PR still render a strip for a non-CI agent:

1. **The healing predicate is an exact match.** `isClonedCIMaintainerStrip` requires ci-maintainer's 14 keys *in the same order*. A spoke where the operator deleted one row from the Stats tab (or a partial copy) keeps the file, and the remaining `{"key":"coverage","source":"agentMetrics","style":"pct-bar","target":91}` then renders `COVERAGE — goal: 91%` — the placeholder strip the owner ruled out.
2. **`coverage` itself is not scoped.** `collectCoverage` only ever files coverage under `AgentMetrics["ci-maintainer"]` (`metrics_collector.go:108`), so a `coverage` stat on any other agent can never hold a value — yet only `source: "health"` is scoped, and that scoping keys on `ADVISORY`/`on_demand`, so a non-advisory, non-CI agent with a copied strip keeps the health dots too.

Suggested delta on top of what you have: filter on **CI ownership** (name `ci-maintainer`, `role: ci-maintainer`, or a replica of it) and treat the strip as *health-sourced stats + the coverage metric*, applied on read. Then a non-owner's `statsConfig` is empty, `renderStatsFromConfig` returns `''` early, and `agentIndicators` emits nothing — no container, no placeholder — while ci-maintainer keeps its strip and its real numbers.

I implemented and verified exactly that on `fix/7411-agent-stat-strip-inheritance` (closed PR #7418, second commit) if it is useful to cherry-pick:

- `scopeCIOwnerStats` / `agentOwnsCIStats` / `isCIOwnerStat` in `status_builder.go`, applied in both `FrontendAgent` builders, `LoadStatsConfigWithCfg`, and `loadAgentStats`.
- Tests pin the three cases the owner asked for, with assertions scoped to **one agent's own rendered section** (a page-wide `strings.Contains` for `0%`/`91` matches unrelated dashboard content): `reviewer` and a freshly created agent render `''`; `ci-maintainer` renders `COVERAGE`, `>88%<`, `goal: 91%`, `>95%<`, health dots; a genuine `0` still renders `0%`. The renderer is *executed* under node via the existing `jsFunc` harness, fed the actual output of the Go scoping.
- Mutation-tested (each reverted, `build failed` checked explicitly so a compile error is not miscounted): strip renders unconditionally → 5 tests fail (the render test prints the exact placeholder strip the owner rejected); retired seed restored → both seed tests fail; `coverage` dropped from the CI-owner predicate → 4 tests fail; role/replica ownership check removed → the over-filtering guard fails.

Full `go test ./pkg/dashboard/ -count=1` green (62.5s), `go vet` clean, `gofmt -l` clean, and `check-inline-js.js` passes on both HTML files.

Happy for this to land in your PR rather than a second one — flagging it so the "render nothing" requirement is not left half-satisfied.

clubanderson (approved):
Reviewer stat-strip seed fix reviewed: seed removed with invariant test, boot-time healing of the verbatim strip on affected spokes, role-scoped health stats, per-agent stat-sources catalogue, pct '—' rendering. LGTM.

PR #7427, opened 2026-09-17
🐛 Stop one agent restart freezing every settings dialog

clubanderson:
Partial fix for #7417 — the backend half of the settings-dialog stall. Frontend half is #7420.

## The stall

`Manager.mu` is a **single global lock**. `restartWithReason` holds it in *write* mode across the entire tmux relaunch:

`C-c` → output archive → process reap/kill → `tmux kill-session` → `ensureTmuxSession` (`cmd.CombinedOutput`) → `installCavemanForAgent` → `tmuxPaneHasCLIForAgent` → `applySecretEnv` → `send-keys` → three `time.Sleep`s.

Measured on the `projectbluefin` spoke:

```text
14:58:45.793  restart
14:58:57.048  started      = 11.25s hold
```

Three restarts fired inside 30 seconds.

`handleAgentConfigGet`'s **first action** was `GetStatus`, which takes `m.mu.RLock()`. So opening settings for *any* agent waited out *any other* agent's relaunch. Agents do not have independent locks. `/api/status` stays fast only because `handleStatus` serves a cached payload under `Server.statusMu` and never calls `GetStatus`; `/api/config/governor` never touches the agent manager. That asymmetry is exactly why the dialog felt randomly broken while the rest of the page looked fine.

Worth noting: **upgrading does not fix this.** `a9f5ed28` (#7374) addresses a different bug and its diff *adds* another `GetStatus` plus a conditional `Restart`. `origin/v5` has the same lock structure.

## The change

`GetStatusFast`: `TryRLock`, and if that fails, serve the last snapshot instead of blocking. Go's `RWMutex` fails `TryRLock` for a *pending* writer as well as a holder — which is the behaviour we want, since a reader that queues behind a pending restart is the bug.

This is safe because of what the caller actually needs. `handleAgentConfigGet` reads exactly four display fields — `BackendOverride`, `ModelOverride`, `LastKickMessage`, `PinnedCLI` — and **already** guards every one with `err == nil`, degrading to the configured values. So both new paths are pre-handled:

- **stale snapshot** → at worst one restart out of date, versus a 12-second spinner
- **uncached + locked → error** → caller falls back to config, which is correct. Deliberately *not* a zero-valued `AgentProcess`, which would render a fabricated all-zero config — the #7405 failure mode.

The snapshot cache carries its own mutex, never `m.mu`, following the convention already documented in this file for `thrashMu` and `consentWedges`.

## What this does NOT do

It does not remove the 11s hold. `restartWithReason` still needs phasing so the relaunch runs outside `m.mu` — but `launchInTmux` mutates agent fields directly under the caller's lock, so unlocking around it requires per-agent field synchronisation. Given the documented history in that function (relaunches born with dead contexts; a generation check that turned restarts into silent kill-without-relaunch), that refactor deserves its own PR rather than being smuggled into this one. **#7417 stays open.**

What this does do is stop the hold being visible to readers who never needed serialising against a restart in the first place.

## Validation

- `go build ./...` clean
- `go test ./pkg/agent/ ./pkg/dashboard/ -count=1` → **ok 368.4s / ok 62.1s**
- New tests pass under `-race`:

```text
--- PASS: TestGetStatusFastDoesNotBlockOnHeldWriteLock
--- PASS: TestGetStatusFastReturnsErrorForUnknownAgentWhenLockIsFree
--- PASS: TestGetStatusFastErrorsWhenUncachedAndLocked
```

The first test also asserts that plain `GetStatus` *still blocks*, so a later refactor cannot quietly alias the two and silently delete the protection.

Refs #7417, #7405


clubanderson:
Pushed a fix for the golangci-lint red: govet copylocks flagged `cached := snap` / `stale := *cached` (AgentProcess carries paneMu). The snapshot cache now stores the same pointer the fast path returns and the stale path returns the cached pointer without a copy — snapshots are read-only display values, and refreshes replace the map entry. build/vet/targeted tests green locally.

PR #7441, opened 2026-09-17
🐛 fix(dashboard): make an expired session readable instead of an opaque "Load failed" (#7433)

### LINKED ISSUE #7433: Expired sessions 302 to a cross-origin login URL instead of 401, so every API call rejects as opaque "Load failed"
## Summary

Authenticated API endpoints answer **302 to a cross-origin login URL** instead of **401** when a session lapses. `fetch()` follows the redirect, the cross-origin target carries no CORS headers, and the request **rejects** with `TypeError: Load failed`. Every expired-session API call in the dashboard therefore surfaces as an opaque network error rather than a clean, actionable 401.

This is the root cause of the whole `Load failed` toast family, including #7405.

## Reproduction

```console
$ curl -s -D - -o /dev/null -X POST \
    https://hosted-projectbluefin-knuckle-gjvq.hive.hivecommons.dev/api/terminal/assertion/renew

HTTP/2 302
location: https://hive.hivecommons.dev/login?redirect=https://hosted-projectbluefin-knuckle-gjvq.hive.hivecommons.dev/api/terminal/assertion/renew&rd=...
```

Note the origin change: **spoke host → apex host**. No `access-control-allow-origin` on the response.

Content negotiation does not exist — the same request with explicit XHR/CORS hints is still redirected:

```console
$ curl -s -D - -o /dev/null -X POST \
    -H "Accept: application/json" \
    -H "Sec-Fetch-Mode: cors" \
    -H "X-Requested-With: XMLHttpRequest" \
    .../api/terminal/assertion/renew

HTTP/2 302
location: https://hive.hivecommons.dev/login?...
```

## Why this produces `Load failed`

`renewTerminalAssertion()` issues a normal same-origin `fetch` with `credentials: 'same-origin'`. Redirects are followed transparently, so the request becomes cross-origin, fails the CORS check, and rejects before any status is observable. The status branches can never run:

```js
if (resp.status === 204) return true;
if (resp.status === 401) showToast('Sign in to this hive before opening a terminal.', 'error');
else if (resp.status === 403) showToast('Your current hive role cannot open terminals.', 'error');
```

The code already knows this empirically — the catch block says so:

```js
// A rejected fetch is how an expired hosted session shows up (the
// status branches above never see it); the shared handler decides
// whether this is "sign in again" or a real failure (#7405).
```

That comment documents the symptom. This issue is the cause: the server never sends 401 to begin with.

## Impact

- **Operator-visible:** `Could not refresh terminal access: Load failed`, with no indication that the remedy is simply to sign in. Observed repeatedly with no deploy or restart preceding it — the `hive_terminal_assertion` cookie is 15 minutes, so it just lapses.
- **Fleet-wide class, not one endpoint:** any authenticated `/api/*` call inherits this. #7405 was the same failure on `GET /api/config/agent/{name}`, where it rendered a fabricated all-zero config.
- **Every `res.status === 401` check in the dashboard is dead code** on this path, because a 401 is never delivered.
- It defeats client-side remediation: the frontend cannot distinguish "signed out" from "hive unreachable", so it cannot reliably offer **Sign in**. Both #7405 and the terminal path had to resort to probing the session separately to guess.

## Proposed fix

Content-negotiate at the auth layer. Redirect only real browser navigations; answer API calls with a status the client can read:

- `Sec-Fetch-Dest: document` (or `Accept: text/html`) → keep the existing 302 to `/login`.
- Everything else, and unconditionally for `/api/*` → **`401`** with a small JSON body, e.g. `{"error":"session_expired","sign_in":"https://hive.hivecommons.dev/login?..."}`.

Returning the sign-in URL *in the body* lets the dashboard show a real **Sign in** action instead of inferring intent from a failed fetch. Once this lands, the session-probing workarounds in `reportFetchFailure` can be simplified, and `401` handling across the dashboard stops being dead code.

If a redirect must be preserved for some non-API route, it should at least be **same-origin** so the fetch does not reject.

## Context

- Observed on `hosted-projectbluefin-knuckle-gjvq`, spoke at `5e7d2a9`.
- Related: #7405 (fabricated all-zero config from this same rejection), #7417 (config dialog hang).

clubanderson:
## Summary

An expired session on a hosted spoke answers **302 to the apex login host**. Under `fetch()`'s default `redirect: 'follow'` the browser chases that cross-origin, fails the CORS check, and rejects with an unclassifiable `TypeError: Load failed`. Every `res.status === 401` branch in the dashboard is therefore dead code on that path, and the UI cannot tell *"you are signed out"* from *"this hive is unreachable"*.

Closes #7433.

## What #7413 already covered, and what was left

**#7413** (merged, `Closes #7405`) fixed the **server half for newly provisioned hives**: a `hive-api-xhr` Ingress covering `Prefix /api` with the same per-hive auth gate but **no `auth-signin`**, plus `unauthorized.{json,html}` in `hive-error-pages` and `error_page 401` mapping, so the gate answers a real `401`. It also added the client-side `reportFetchFailure` / `probeSession` funnel for three specific call sites.

Two things were left, and this PR addresses them:

1. **That Ingress is provisioned per hive** (`provisionHive` → `kubectl apply`; the vanity patch loop only patches Ingresses that already exist). A hive created before #7413 never gets it, and `deploy/k8s/error-pages.yaml` has to be re-applied by an operator. The re-verification comment on #7433 was posted *18 minutes after #7413 merged* and still observed the 302 on all three endpoints — consistent with exactly this. So the client cannot depend on the gate having been fixed.
2. **The client still could not classify the redirect.** `probeSession` falls back to `isHostedHiveHost() ? expired : unknown` — a *guess* from the hostname, which is precisely the "cannot distinguish signed out from hive unreachable" complaint in the issue. And the other ~245 `/api` `fetch()` sites in `index.html` were untouched.

Verified not in scope / already correct: the spoke's own `Server.authenticate` already returns `{"error":"unauthorized"}` with `401` (`server.go:1453`) — the 302 is produced by nginx forward-auth, not by Go, so a change confined to Go handlers would not alter the observed behaviour. `/api/terminal/...` matches the `/api` prefix (the `hive-terminal` Ingress is `/terminal`), so it is not a separate gap.

## What changed

**`src/pkg/dashboard/static/index.html`** — the dashboard already wraps `window.fetch` (to inject the dashboard token on mutations). That wrapper is now also the session-redirect guard, so **all 248 `fetch` call sites are covered without touching one of them**:

- `_isGuardedApiRequest(url)` — same-origin **and** `/api` prefix only, with `GET /api/gh-user-auth/session` exempt (the device-flow landing: a real browser navigation whose own 302 back to `/` is its *success* path, not an XHR). Cross-origin and non-`/api` requests fall through to an unmodified `fetch`.
- `_fetchWithSessionGuard(url, opts)` — issues guarded requests with `redirect: 'manual'`. The gate's 302 then arrives as an observable `opaqueredirect` response instead of a CORS rejection, and is converted into `401 {"error":"session_expired","sign_in":"/", "detail":…}` with `Content-Type: application/json` — the 401 the gate should have sent, in the shape the issue proposed.
- Every network call the wrapper makes now goes through the guard: the primary request, the stale-token retry, and `_ensureAuthToken`'s own `/api/auth/token` and `/api/gh-user-auth/status` probes (whose failures were swallowed by a bare `catch`).

Consequence: `probeSession()` now reads a **fact** (`r.status === 401`) rather than falling through to the hostname heuristic, and every `res.status === 401` branch in the dashboard becomes live again — on old hives and new ones alike.

**`src/docs/security-model.md`** — documents that the client no longer depends on the per-hive Ingress, and the one exemption.

**Not changed:** no Go handler, no Ingress template. The redirect is produced by nginx, and #7413 already fixed the manifest.

## Testing

`src/pkg/dashboard/api_session_redirect_guard_test.go` — 5 tests, each scoped to the fetch-wrapper IIFE (and, where relevant, to the `window.fetch` override body) rather than page-wide `strings.Contains`, so they cannot pass because some unrelated part of a 26k-line file matches.

- `TestApiFetchUsesManualRedirect` — the guard exists and forces `redirect: 'manual'`.
- `TestApiRedirectBecomesReadable401` — `opaqueredirect` is detected and answered with a JSON `401` carrying `error: 'session_expired'` and `sign_in`.
- `TestGuardIsTheOnlyDoorForApiFetches` — primary request, stale-token retry and both internal `/api` probes route through the guard; `_origFetch` has exactly the two call sites inside the guard and none in the override.
- `TestGuardScopeExcludesTheDeviceFlowLanding` — the exemption exists *and* is consulted.
- `TestGuardScopeIsThisHivesOwnApi` — same-origin + `/api` restriction, and the unguarded fall-through.

**Against the parent commit (`f8296b3c`, i.e. with `index.html` reverted): 5 of 5 FAIL.**

**Mutation testing — 6 mutants, 6 killed, 0 build failures** (HTML mutations cannot break the Go build; each run was checked for `--- FAIL`, and the tree was restored to green afterwards):

| # | Mutation | Killed by |
|---|---|---|
| 1 | `redirect: 'manual'` → `'follow'` | `TestApiFetchUsesManualRedirect` |
| 2 | synthesized `status: 401` → `403` | `TestApiRedirectBecomesReadable401` |
| 3 | drop the `if (res.type !== 'opaqueredirect')` check | `TestApiRedirectBecomesReadable401` |
| 4 | make the exemption list inert (`return true`) | `TestGuardScopeExcludesTheDeviceFlowLanding` |
| 5 | stale-token retry bypasses the guard (`_origFetch`) | `TestGuardIsTheOnlyDoorForApiFetches` |
| 6 | drop the same-origin restriction | `TestGuardScopeIsThisHivesOwnApi` |

Commands run:

- `cd src && GOMODCACHE=… go test ./pkg/dashboard/ -count=1` → `ok … 55.069s` (full suite, no flakes observed)
- `go vet ./pkg/dashboard/` → clean; `gofmt -l` on the new test file → empty
- `find dashboard src/pkg/dashboard/static -name '*.html' -print0 | xargs -0 node .github/scripts/check-inline-js.js` → *All script blocks passed syntax and no-undef checks.*
- `src/pkg/dashboard/server.go` was not modified, so `check-api-reference-citations.sh` was not required.

**I could not observe the live cluster** — the `curl` evidence in the issue is the reporter's, not re-run by me. The redirect→`opaqueredirect`→401 conversion is browser behaviour and stays manual-verify; what is automated here is that the guard exists, is the only door to the network for `/api`, and does not swallow the one `/api` route whose redirect is real.


clubanderson:
Reviewed this, and independently verified the one thing that would sink it.

The `redirect: 'manual'` guard rests entirely on the premise that *"an authenticated /api call has no legitimate reason to redirect"* — if any other `/api` route redirects as part of its success path, this converts that into a spurious `session_expired` 401 and breaks it. So I enumerated every redirecting handler in `pkg/dashboard` and traced each to its registered route:

| handler | route | under `/api`? |
|---|---|---|
| `handleGHUserAuthSession` (`api.go:3223`) | `GET /api/gh-user-auth/session` (`api.go:128`) | **yes — exempted by this PR** |
| `handleSSO` (`api.go:3097`) | `GET /sso` (`server.go:1027`) | no |
| `handleGitHubAppSetupCallback` (`api.go:5419`) | `GET /gh-setup` (`server.go:1022`) | no |
| `redirectLinearAgent` (`api_linear_agent.go:438`) | `/linear/callback` (`api_linear_agent.go:36`) | no |
| `redirectOpenRouter` (`openrouter.go:308`) | `/openrouter/callback` (`openrouter.go:44`) | no |
| `handleContributorDossierPage` (`api_contribute.go:680`) | `/contribute/...` | no |
| `handleLeaderboardPage` (`api_contribute.go:8938`) | `/contribute/leaderboard` | no |

So the exemption list is exactly right, and it is exactly one entry long. The premise holds against the current tree.

Two things I like about the shape here. Wrapping the existing fetch wrapper covers all ~248 call sites with zero call-site edits, which is the only version of this fix that doesn't rot the moment someone adds endpoint 249. And scoping to same-origin means a cross-origin call keeps its own redirect contract rather than inheriting ours.

The finding that the `hive-api-xhr` Ingress from #7413 is applied **per hive at provision time** is the real substance — it means every pre-existing hive still 302s today, which matches the re-verification comment posted after #7413 merged. Server-side alone was never going to close this.

One forward-looking note, not a blocker: the table above is a fact about the current routing tree, not an invariant anyone has pinned. If someone later adds a redirecting `/api` route, this guard turns it into a phantom expired-session prompt, and the failure would be confusing to trace. A cheap Go test asserting that no route registered under `/api` calls `http.Redirect` (exempting `gh-user-auth/session`) would make that assumption self-enforcing.

PR #7447, opened 2026-09-17
✨ feat(dashboard): let operators turn off an agent's On Demand flag

### LINKED ISSUE #7446: ✨ on_demand is a real per-agent field with no way to change it — and it silently zeroes reviewCapableAgents
## Summary

`on_demand` is a real per-agent config field (`config.AgentConfig.OnDemand`, `pkg/config/config.go:1012`), but the settings dialog only ever **renders** it. The agent card shows a read-only `on demand` badge and there is no control anywhere to change it. An operator can see that an agent is on-demand and has no way to take it off short of hand-editing `hive.yaml`.

Concretely:

- `GET /api/config/agent/{name}` does not return the field, so the dialog cannot render a control even if one existed.
- `PUT /api/config/agent/{name}/general` has no `onDemand` branch, so the value is silently dropped if sent.
- The frontend reads `a.onDemand` in 9 places — all of them display-only (badge, "no cadence" hints, next-kick tooltips).

## Why this is more than a missing checkbox

On-demand is a *hard gate* in three separate places, and one of them is not about scheduling at all:

1. `Governor.AgentEligibleForCELKick` (`pkg/governor/governor.go:901`) — on-demand agents are never event/CEL-kicked.
2. `NoCadenceAgents` (`pkg/governor/no_cadence.go:30`) — on-demand agents are excluded from the no-cadence warning, so the fleet banner will not flag them as idle either.
3. **`reviewCapableAgents` (`pkg/review/dispatch.go:302`)** — `if a.Name == "" || !a.Enabled || a.Paused || a.OnDemand || !a.UsesKick { continue }`.

That third one produces a catch-22 for a reviewer agent. The `reviewer` agent exists to wake per-PR and judge that change — but because it is marked on-demand, it is excluded from the very dispatch path that would summon it. `PlanDispatch` then short-circuits:

```go
if len(reviewers) == 0 || availableSlots <= 0 {
    continue
}
```

## Observed on a live spoke

`hosted-projectbluefin-knuckle-gjvq`, 16 repos:

```json
{"reviewRequireApproval":true,"reviewFanOut":true,"reviewCapableAgents":0,"totalAgents":8}
```

Review fan-out is enabled and approval is required, yet there are **zero** review-capable agents, so nothing is ever dispatched. Downstream, `latestHiveQueueApproval` (`pkg/github/automerge_sweep.go:728`) needs an APPROVED review carrying the hive-queue marker and authored by the App bot. With no reviews, that approval never exists, and the sweep skips everything:

```
automerge sweep complete  seen=37  merged=0  skipped=37
```

almost all with `reason=no-hive-queue-approval`. The spoke is sitting on **305 open PRs** and a 113-item hold list, and the issue backlog for architect/quality/ci-maintainer logs as `suppressing issue already claimed by a hive PR` — their work is all parked behind PRs that cannot merge.

I am not claiming on-demand is the sole cause of that backlog, but `reviewCapableAgents: 0` is real, it is caused by this flag, and an operator currently cannot fix it from the dashboard.

## Proposed fix

Carry `onDemand` on the read and general-save paths and add a toggle to the General tab.

Two things the control should say, because clearing the flag is necessary but not sufficient:

1. The agent also needs a cadence entry in the **active** governor mode before the governor will kick it.
2. On-demand agents are excluded from review fan-out.

A general save that omits `onDemand` must leave it alone — the dialog PUTs only dirty keys, so editing an unrelated field on an on-demand agent must not silently un-demand it.

clubanderson:
Closes #7446

`on_demand` is a real per-agent config field, but the settings dialog only ever *rendered* it: the agent card showed a read-only `on demand` badge and offered no control. `GET /api/config/agent/{name}` did not return the field and `PUT /api/config/agent/{name}/general` dropped it, so the only way to clear the flag was to hand-edit `hive.yaml`.

### Why it matters beyond scheduling

`reviewCapableAgents` (`pkg/review/dispatch.go:302`) skips any agent with `OnDemand` set:

```go
if a.Name == "" || !a.Enabled || a.Paused || a.OnDemand || !a.UsesKick {
    continue
}
```

So a reviewer agent marked on-demand is excluded from the very dispatch path meant to summon it. `PlanDispatch` then short-circuits on `len(reviewers) == 0`, no PR ever receives a verdict, and `latestHiveQueueApproval` never finds the App-authored approval the auto-merge sweep requires.

Observed live on `hosted-projectbluefin-knuckle-gjvq`:

```json
{"reviewRequireApproval":true,"reviewFanOut":true,"reviewCapableAgents":0}
```

```
automerge sweep complete  seen=37  merged=0  skipped=37
```

— almost all `reason=no-hive-queue-approval`, against 305 open PRs. Being unable to clear the flag from the dashboard made that state unrecoverable without editing config by hand.

### Change

- `GET /api/config/agent/{name}` returns `general.onDemand`.
- `PUT /api/config/agent/{name}/general` honours `onDemand`.
- General tab gains an **On Demand** toggle beside Clear on Kick, rendering the same `on demand` badge inline so the dialog agrees with the agent card.

The tooltip names both gotchas, because clearing the flag is necessary but not sufficient: the agent also needs a cadence entry in the **active** governor mode before the governor will kick it, and on-demand agents are additionally excluded from review fan-out.

### Scope

Deliberately limited to the edit path. The agent *import* path (`agentConfigFromDefinition`) has no `OnDemand` in its spec; wiring that up is a separate change and not needed to fix the reported problem.

### Tests

Three new tests in `agent_config_on_demand_test.go`:

- `TestAgentConfigGet_ExposesOnDemand` — asserts key **presence**, not just value: a missing key renders as an unchecked switch, which would display "not on demand" for an agent that is.
- `TestAgentConfigGeneral_ClearsAndSetsOnDemand` — both directions, so the control is not one-way.
- `TestAgentConfigGeneral_OmittedOnDemandIsPreserved` — the dialog PUTs only dirty keys, so editing an unrelated field on an on-demand agent must not silently un-demand it.

Verified: `go build ./...`, full `go test ./pkg/dashboard/` (54s, ok), `go vet`, and the inline dashboard JS extracted and `node --check`ed.


clubanderson:
Renamed the fragment to `added-7446-on-demand-toggle.md` — changelog-fragment-guard requires the `<category>-` prefix (added/changed/deprecated/fixed/security).

Issue #7369, opened 2026-09-17
🌱 Unify v4/v5 file layout so the v5 top-up stops conflicting on every main.go/config.go/manager.go/saas.go change


clubanderson:
## Problem

v5 split four files that v4 still edits as monoliths:

| file | v4 | v5 |
|---|---|---|
| `src/cmd/hive/main.go` | ~9,500 lines | 136 lines + `main_helpers.go`, `hubwire.go`, `governorwire.go`, `proxywire.go`, `*_wire.go` … |
| `src/pkg/config/config.go` | ~6,400 | 109 + `defaults.go`, `governor_config.go`, `governor_runtime_config.go` … |
| `src/pkg/agent/manager.go` | ~11,000 | 1,401 + `manager_tokens.go`, `manager_copilot_auth.go` … |
| `src/pkg/hub/saas.go` | ~23,000 | 525 + `saas_hives_api.go`, `saas_clusters.go` … |

Git cannot follow those moves, so **every** v4 commit touching one of them conflicts with the whole split. Both branches also independently extracted `pkg/apphealth`, `pkg/advisory`, `pkg/inference` with different APIs (reconciled toward v4 in #7351).

The automated `v5 Top-up` (#7297) only ever merged cleanly because v4 happened not to touch those files in a given window. It failed twice in one day (#7337): #7351 hand-ported 18 v4 commits, and #7366 hand-ported the very next one (c6fb9727e) minutes later.

## Proposal

Port v5's file split back to v4 as a **refactor-only** PR (no behaviour change, `git mv`-style extractions so blame survives). Once both lines share the layout the forward-merge becomes mechanical again and the cadence can run unattended.

Also unify the three duplicated package extractions on the v4 API (already done on v5 by #7351; v4 is the source of truth).

## References

- #7337 (the CI failure this fixes for good), #7199 (earlier precedent)
- #7351, #7366 (hand ports; see the first comment on #7337 for the full diagnosis)

clubanderson:
**Direction decided (with @clubanderson): v5 adopts v4's file layout.**

- #7450 collapses v5's `cmd/hive`, `pkg/config`, `pkg/agent`, `pkg/hub` back to v4's file set, decl-hash-verified (zero behaviour change), and folds the `spokeWire` step methods back into `main()`.
[...]

PR #7450, opened 2026-09-17
🌱 v5: adopt v4's file layout for cmd/hive, pkg/config, pkg/agent, pkg/hub

clubanderson:
Refs #7369

v5 adopts v4's file layout for `cmd/hive`, `pkg/config`, `pkg/agent`, `pkg/hub` so v4→v5 top-ups stop conflicting on every commit that touches `main.go` / `config.go` / `manager.go` / `saas.go` (the last three top-ups #7351, #7366, #7376, #7387, #7426 were all hand-ported for this reason).

**Behaviour-preserving, mechanically verified**

- Every top-level declaration (func/type/var/const) was compared by name + sha1 of normalised source before and after: identical sets for all four packages, except `spokeWire.*` methods which were folded back into `main()`.
- For `main()`: every line of the 41 `spokeWire` step/helper method bodies is present in the new `main()` after the mechanical `w.` prefix strip and `w.addCleanup(...)` → `defer ...` rewrite (verified by normalised line-set inclusion). Step bodies are inlined in `run()` order; helpers become closures; method-value callbacks become inline literals; `linearCredentialResolver` is a single closure as on v4.
- v5-only extractions that v4 does not have are **kept** as separate packages: `pkg/hub/spoke`, `pkg/dashboard/collect`, `pkg/github/automerge`, `pkg/requestwatch`, `pkg/appkey`. Per #7369 discussion these will be proposed to v4 as separate refactor PRs. v5-only decls in v5-only files (`provider_gateways.go`, `main_helpers.go`, `hookwire.go`, …) stay put.

**Commits**
1. import `pkg/hub/spoke` as `spoke` (never `hub`) so v4's `hub.` references don't alias-collide
2. collapse `cmd/hive` to v4's file set
3. collapse `pkg/config`, `pkg/agent`, `pkg/hub`
4. fold `spokeWire` methods back into `main()`; delete emptied `*wire.go`
5. re-point source-scanning tests at `main.go`/`manager.go` (v4's copies where they exist), drop v5-only `config.go` line-count ratchet, refresh api-reference citations

`go build ./... && go vet ./... && go test ./...` green locally.

Structural splits of these files should land **v4-first** going forward (#7238, #7278, #7303, #7435) so v5 inherits them through the normal top-up.


clubanderson:
Pushed the quadlet-gate fix: the layout adoption deletes `pkg/config/defaults.go`, but v5's `src/deploy/test_quadlet_config_contract.sh` still pointed `GO_DEFAULTS` at it. Took v4's version of the script (`GO_CONFIG=config.go`) — the same layout this PR adopts. 14 assertions pass locally.

Issue #4081, opened 2026-08-18
🐛 Spoke owners cannot see Upgrade option in hub or spoke navbar


clubanderson:
## Symptom
A spoke OWNER (e.g. the owner of a hosted spoke) cannot see the "Upgrade" affordance anywhere:
1. **Hub "My Hives" dashboard** — the owner's row shows no Upgrade link in the Version cell even when the hive is behind the latest release.
2. **Spoke dashboard navbar** — owner-gated actions on the spoke (including `POST /api/self-upgrade` behind `requireOwnerRole`) are refused because the spoke sees the user as a non-owner.

The hub admin CAN see and launch upgrades for the same spoke.

## Root cause (v2 branch)
The hub normalizes a stale/demoted stored role to `owner` **only for the hub admin**, never for the hive's actual owner. A user's stored role for their own hive (`SaaSUser.Hives[hiveID]`) can end up as a non-owner value (e.g. `handleApproveAccess` unconditionally overwrites `target.Hives[hiveID] = "read"` even for the existing owner — `v2/pkg/hub/saas.go:6036-6038`). Once that happens:

1. **`handleMyHives` (`v2/pkg/hub/saas.go:2646-2650`)**: the stored-role branch wins before the `canonicalEqual(h.Owner, username)` owner check is ever reached, and the admin-only normalization
   `if isAdmin && role != "owner" { role = "owner" }` (line 2647)
   leaves the true owner with the demoted role. The dashboard renders the Upgrade link / queued pill / auto-upgrade select only when `h.role === 'owner'` (`v2/pkg/hub/saas.go:13371-13387`), so the owner sees nothing while the admin (force-normalized to `owner` on every row) sees it.
2. **`handleAccessStatus` (`v2/pkg/hub/saas.go:2951`)**: same stored-role-first pattern, same demotion.
3. **`handleSaaSAuthCheck` (`v2/pkg/hub/saas.go:7793-7800`)**: the spoke auth proxy forwards `user.Hives[hiveID]` verbatim as `X-Hive-Role` with no owner elevation, so the spoke's `requireOwnerRole` (`v2/pkg/dashboard/api.go:1246`) rejects the true owner's `/api/self-upgrade` and the spoke UI treats them as read-only.

[...]

PR #7457, opened 2026-09-17
🐛 fix(contribute): identify a signed-in visitor on /api/contribute on hosted spokes

### LINKED ISSUE #7453: 🐛 bug(contribute): on a hosted spoke the Operations tab never recognises a signed-in user after #7364 (/api/contribute/me is always 401)
## What happens

On a hosted spoke (`https://hosted-projectbluefin-common-nmq5.hive.hivecommons.dev`), I open `/contribute/operations`. The "Your contribution" card says I am not signed in and shows a **Sign in with GitHub** link.

Clicking it goes to `/` (the dashboard root). I sign in through the hub, and land on the dashboard, which shows me as signed in. If I then go back to `/contribute/operations`, the card still says I am not signed in. There is no way to get the Operations tab to recognise me.

I am the owner of this hive.

## What I think is going on

This looks like a regression from #7364 (the fix for #7362, merged today). That fix stopped the owner-token fallback on hub-proxied spokes, which was right — but the identity it expects to use instead never reaches `/api/contribute/*` on a hosted spoke:

1. The Operations card calls `GET /api/contribute/me`.
2. On a hosted spoke that path is served by the `hive-contribute` Ingress, which has **no `auth-url` annotation** (`saas_provision.go`, the `hive-contribute` block). nginx never asks the hub who the caller is, so `X-Hive-User` is never injected.
3. Even if it did ask, the hub's `handleSaaSAuthCheck` returns `200` for `publicPaths` (which includes `/api/contribute`) **before** it resolves the user, so it would not set `X-Hive-User` anyway.
4. `resolveViewerUsername` on a hub-proxied spoke now goes: no spoke session cookie (hosted spokes do not use the device flow) → no `X-Hive-User` → `hubProxied()` is true → return `""`. So `/api/contribute/me` answers `401` for everyone, including the signed-in owner.

Before #7364, step 4 fell through to the persisted owner token, which is why the card used to show something (the owner's stats, to everyone — the bug #7362 fixed). Now it shows nothing to anyone.

The test added in #7364 says "On a hub-proxied spoke the hub's nginx injects X-Hive-User for every signed-in visitor" — that is true for `/api/...` paths gated by the `hive-api-xhr` Ingress, but not for `/api/contribute/...`.

The **Sign in with GitHub** link is a smaller, related problem: `ccSignInCTA` links to `/`, so after login the user lands on the dashboard instead of coming back to the tab they were on.

## What I can confirm from outside

Anonymous `curl` against the spoke:

```
/contribute/operations       200
/api/contribute/me           401
/api/gh-user-auth/status     302 -> hub /login
/api/role                    302 -> hub /login
/api/contribute/leaderboard  200
```

So `/api/contribute/*` is not behind the hub auth gate at all, while the rest of `/api/*` is. I have not yet captured the signed-in requests in devtools; I expect `/api/gh-user-auth/status` to say `logged_in: true` while `/api/contribute/me` still says `401`. I also have not checked which hive version this spoke is on, so "regression from #7364" is my reading of the code, not something I have bisected.

## Possible fix

Two halves are needed, since either one alone does nothing:

- **Hub:** in `handleSaaSAuthCheck`, for `publicPaths`, still try to resolve the user when a valid hub session is present and set `X-Hive-User` / `X-Hive-Role` / `X-Hive-Proxy-Auth` on the `200`. Never `401`/`403` on a public path — anonymous access to the leaderboard and the contributor relay (`/api/contribute/ws`, registration tokens) must keep working.
- **Provisioning:** give the `hive-contribute` Ingress the `auth-url` and `auth-response-headers` annotations (but not `auth-signin`, for the same reason `hive-api-xhr` omits it — a `fetch()` cannot follow the cross-origin login redirect).

Separately, `ccSignInCTA` could link to `/login?redirect=<current URL>` on hub-proxied spokes (or simply to the current page, since the ingress `auth-signin` already carries the return URL) so the user comes back to the tab they were on.

Danathar:
## Summary

On a hosted spoke the Operations tab's "Your contribution" card told everyone — the hive's own owner included — that they were not signed in, and its **Sign in with GitHub** link dropped them on the dashboard instead of bringing them back. #7364 correctly stopped the owner-token fallback, but the identity it relies on instead never reached `/api/contribute/*`: the `hive-contribute` Ingress has no `auth-url`, and the hub's auth-check answered a bare `200` for public paths before it resolved the user. The reporter's DevTools capture confirms it: `/api/gh-user-auth/status` and `/api/role` (gated Ingress) name the owner, `/api/contribute/me` (contribute Ingress) arrives anonymous and answers `401`. This PR ships both halves the issue asks for, plus the sign-in return link.

Closes #7453.

## What changed

**Hub (`pkg/hub/saas_auth.go`).** On a public path the auth-check still answers `200` for anyone — anonymous access to the leaderboard and the contributor relay (`/api/contribute/ws`, registration tokens) is unchanged — but `setPublicPathIdentity` now sets `X-Hive-User` / `X-Hive-Role` / `X-Hive-Proxy-Auth` when the caller has a hub session: their grant on the hive (the canonical owner elevated exactly as the gated path does, #4081), or `publicPathGuestRole` = `read` for a signed-in user with no grant. `read` grants nothing beyond anonymous on a public path (the spoke's public handlers hand extra capability only to owner/read-write — `requireContributorWrite`, `paneTailViewer`) and never reaches a gated path, which still `403`s a grant-less user. An anonymous caller gets no identity headers and no proof, so the spoke treats the request as anonymous, exactly as before. The `X-Hive-Proxy-Auth` CONTRACT comment is updated accordingly.

**Provisioning (`pkg/hub/saas_provision.go`).** `hive-contribute` carries the same `auth-url` and `auth-response-headers` as the page Ingress, and **no** `auth-signin` — for the reason `hive-api-xhr` has none (`fetch()` cannot follow the cross-origin login redirect, and a public path never needs one). The relay's long-poll timeouts are untouched. The spoke side already trusts hub identity on public paths when the proof header is present (`trustProxyIdentity`), so no spoke auth change is needed.

**Spoke (`pkg/dashboard`).** `ccSignInCTA` linked to `/`. On a hub-proxied spoke it now links to `GET /auth/return?to=<current tab>` — deliberately **not** a public path, so the ingress sends an anonymous visitor through the hub login and the handler then `302`s to `to`. Linking to `/contribute/operations` itself would not sign anyone in (public path → the ingress never asks). `safeReturnPath` accepts only same-origin absolute paths (no `//`, `/\`, scheme, host, or CR/LF/NUL); anything else goes to `/contribute`. Self-hosted spokes keep `/` — their device-flow sign-in page *is* the dashboard root. The page learns its auth shape from a `{{HIVE_HUB_PROXIED}}` sentinel substituted like `{{HIVE_BRANCH}}` (no new positional template argument). The route is tail-registered and documented in `api-reference.md`; `check-api-reference-citations.sh` passes.

`docs/security-model.md` describes the public-path identity rule.

## Testing

- `pkg/hub/saas_authcheck_public_identity_test.go`: the owner (demoted stored role → elevated), a granted user, and a grant-less user are identified on `/api/contribute/me` with the right role while the gated path still refuses the grant-less user; anonymous requests to `/api/contribute/me`, `/api/contribute/ws`, `/contribute/operations`, `/api/leaderboard`, `/snapshot/x` and `/sso` are `200` with no identity and no proof; the contribute Ingress has `auth-url` + `auth-response-headers` and no `auth-signin` in both TLS modes. The #7405 sibling test is updated (it asserted the contribute Ingress had no gate).
- `pkg/dashboard/contribute_signin_return_test.go`: `safeReturnPath` table (relative, protocol-relative, backslash, scheme, `javascript:`, header injection, NUL); `/auth/return` is not public and `302`s to the tab (off-origin → `/contribute`); the page carries `ccHubProxied` `true`/`false` with no leaked sentinel and the CTA routes through `ccSignInHref`. The existing CTA-is-a-link test now checks the self-hosted branch still returns `/`.
- Per CONTRIBUTING's test policy: on the parent commit the two hub tests fail (`X-Hive-User = ""` on a public path; `hive-contribute has no per-hive auth-url`).
- `go test ./pkg/hub/ ./pkg/dashboard/` green; `go vet` clean; api-reference citations verified.
- Not verified here (needs the hosted cluster): the live round trip once the hub and a re-provisioned spoke carry this — the reporter's capture is the acceptance check to repeat.


clubanderson:
I independently implemented this same fix for #7453 before seeing this PR, so rather than open a duplicate I am posting what I verified against the code. The two halves here match what I arrived at, and all four of the reporter's numbered claims hold up when read against `v4` (`ccb63d07`):

1. Confirmed — the Operations card calls `GET /api/contribute/me`.
2. Confirmed — the `hive-contribute` block in `saas_provision.go` had only the two `proxy-*-timeout` annotations, no `auth-url`.
3. Confirmed — `handleSaaSAuthCheck` (`saas_auth.go`) did `w.WriteHeader(http.StatusOK); return` inside the `publicPaths` loop, before `getAuthUser`.
4. Confirmed — `resolveViewerUsername` returns `""` on a hub-proxied spoke with no session and no `X-Hive-User`, because `s.hubProxied()` short-circuits the owner-token fallback.

The spoke half is indeed already in place: `trustProxyIdentity` is invoked in the `isPublicPath` branch of the auth middleware (`server.go`), so the injected identity is honoured on public paths as soon as the proof header arrives. Good call not touching it.

Three things I would like a second look at.

**`publicPathGuestRole = "read"` for a grant-less user.** The reasoning in the comment checks out where it matters — `requireContributorWrite` demands `owner`/`read-write`, `paneTailViewer` returns false for `read`, and `roleEnforcement` explicitly exempts `/api/contribute/` from the read-only write gate, so registration and relay writes still work for a `read` caller. But it is worth stating that this is a deliberate widening: before this PR nothing on a public path ever saw a role for a user with no grant on the hive, and `read` is a real role elsewhere in the spoke. I took the narrower option (forward identity only when the caller actually resolves to a role on this hive, owner elevation included, and leave a grant-less hub user anonymous), which fixes the reported owner case with strictly less surface. If the goal is that a contributor on hive A can see their own stats on hive B's Operations tab, then `read` is the right call and the PR body should say that is the intent, because it is not obvious from the issue.

**`loadSaaSUser(username) == nil` still gets identity.** In `setPublicPathIdentity` the `loadSaaSUser` result is only consulted for the grant; a hub session whose `SaaSUser` record does not exist falls through to `publicPathGuestRole` and is forwarded as a signed-in caller. The gated branch twenty lines below treats exactly that state as `401 not authenticated`. Deliberate or not, the two branches now disagree about who counts as authenticated; an early `return` when `user == nil` would remove the question.

**Availability, not authorization.** Adding `auth-url` to `hive-contribute` puts the hub on the critical path for traffic that previously had no hub dependency at all: every `/api/contribute/ws` upgrade and every registration request now fires an nginx `auth_request` subrequest first, and if the hub is unreachable or slow, `auth_request` turns that into a 500 rather than the 200 the handler would have produced. That is the same trade `hive` and `hive-api-xhr` already make, so it is probably acceptable, but it is a new failure mode for the anonymous contributor relay specifically and deserves a line in the PR body (and it is not something any unit test can catch).

Finally, on `Closes #7453`: the Ingress manifest is applied per hive at provision time, so this fixes newly provisioned spokes and leaves every existing hosted spoke — including `hosted-projectbluefin-common-nmq5`, the one in the report — on its old annotation-less `hive-contribute` until it is reprovisioned or the two annotations are patched on by hand. The hub half alone does nothing there, since nginx never asks. `Refs #7453` plus an explicit operator note would be more honest, or at minimum the body should spell out the manual `kubectl annotate` for existing tenants. I could not verify the live round trip either — no access to the hosted spoke — so the reporter's DevTools capture remains the acceptance check.

clubanderson:
Merged `v4` into this branch to clear a conflict that appeared when #7456 landed.

The only conflicting file was `src/docs/api-reference.md`, whose `file:line` citations drift whenever `pkg/dashboard` route registrations move. I resolved it by taking v4's version and regenerating rather than hand-merging line numbers, which is the mechanism the repo provides for exactly this:

```console
```

It reports `0 rewritten, 0 left unresolved` and the checker then passes clean, so no citation actually needed rewriting once v4's copy was taken — the conflict was purely textual.

No commits of yours were rewritten: this is a merge, not a rebase, so nothing was force-pushed. Your changes to `pkg/hub` and the provisioning annotations are untouched.

Verified after the merge: `go build ./...` clean, `go test ./pkg/hub/` green (145.9s) and `go test ./pkg/dashboard/` green (54.7s).

clubanderson (approved):
Open-redirect-safe trampoline (blocks //, /\, CRLF/NUL), auth-url identity pass-through matches the security-model doc update. LGTM.

PR #7458, opened 2026-09-17
🐛 fix(dashboard): DASHBOARD URL is hub-owned — show it read-only and refuse the write

### LINKED ISSUE #7451: 🌱 Settings → Hub: DASHBOARD URL is editable but the heartbeat overwrites it, and a bad value can break OAuth login first
## Settings → Hub: `DASHBOARD URL` is editable but an edit cannot survive, and can break login on the way out

The Hub tab renders `DASHBOARD URL` as a normal editable text input next to `HUB URL`. On a hub-managed spoke an operator's edit there is not durable, and while it lasts it is actively harmful. It should be informational.

### An edit is reverted by the next heartbeat

`hub.dashboard_url` is owned by the hub, not the spoke. The heartbeat's project-config push adopts the hub's value whenever it differs, and persists it — `cmd/hive/main.go:5125`:

```go
if pc.DashboardURL != "" && cfg.Hub.DashboardURL != pc.DashboardURL {
    logger.Info("adopting vanity dashboard URL from hub heartbeat",
        "was", cfg.Hub.DashboardURL, "now", pc.DashboardURL)
    cfg.Hub.DashboardURL = pc.DashboardURL
}
```

There is a second, earlier path for a URL-only push that saves immediately (`cmd/hive/main.go:5080`). The write the dashboard performs — `cfg.Hub.DashboardURL = body.DashboardURL` at `pkg/dashboard/api_governor.go:884` — has no precedence over either.

The reason this is quiet today rather than visibly flapping is that a matching value short-circuits: `vanityMatched := pc.DashboardURL == "" || cfg.Hub.DashboardURL == pc.DashboardURL`, and a reconciled beat returns before the adopt block. So the field looks stable purely because nobody has changed it. On `hive-hosted-hosted-projectbluefin-knuckle-gjvq` the stored value already equals the hub's vanity URL and there are zero adopt log lines in the last three hours.

### Until the revert lands, a bad value is not inert

`hub.dashboard_url` is the second-choice source for the OAuth callback origin — `pkg/dashboard/oauth_origin.go:33`:

```go
if v := strings.TrimSpace(cfg.Dashboard.PublicURL); v != "" { return ... }
if v := strings.TrimSpace(cfg.Hub.DashboardURL); v != "" { return ... }
```

With `dashboard.public_url` unset, editing this field changes where an authorization code is redeemed. A typo here points the callback at a host that is not this server, and the operator can lock themselves out of the dashboard they are editing it from. It also feeds the hub's webhook target (`pkg/hub/webhook.go:224` builds `spokeURL` from `hive.DashboardURL`) and Discord message links (`pkg/discord/bot.go:146`).

So the field combines the two worst properties: the change does not stick, and it does damage before it is undone.

### Why editing it cannot achieve what an operator would want

The dashboard is reachable at whatever host the ingress and DNS terminate. Changing a string in config does not move it. On this spoke the hostname comes from the hosted ingress and the `hive.hivecommons.dev` wildcard; the field is a record of that fact, not a control over it.

### Suggested fix

Render it read-only when the hub is enabled, with the provenance shown the way the update block already does it ("Policy source: the hub (heartbeat)"), so the operator can see and copy the value without believing it is a setting. If an escape hatch is wanted for unmanaged spokes, gate editability on `hub.enabled == false`.

If a local override genuinely is needed — for a spoke behind a custom vanity host — `dashboard.public_url` is the field that already wins over this one for OAuth and is not overwritten by the heartbeat. Pointing operators at that instead would be clearer than making a hub-owned field look writable.

Worth applying the same check to the other hub-owned values on this tab (`NAMESPACE`, `SNAPSHOT / PREVIEW URL`), which look similarly derived.

clubanderson:
Closes #7451

Settings → Hub rendered `DASHBOARD URL` as an ordinary editable text input next to `HUB URL`, but `hub.dashboard_url` is owned by the hub, so that input was a promise the spoke cannot keep. I verified each claim in the issue against the code on `v4` before changing anything. The heartbeat's project-config callback in `cmd/hive/main.go` does adopt the hub's value whenever it differs — in two places, the URL-only push that saves immediately (now at line 5084, not 5080 as the issue says) and the full project reconcile after the already-reconciled short-circuit (now line 5129, not 5125). The `vanityMatched` short-circuit is exactly as described, which is why the field looks stable on a spoke nobody has edited. The dashboard's own write, `cfg.Hub.DashboardURL = body.DashboardURL`, was at `pkg/dashboard/api_governor.go:883-885` (the issue says 884) with no precedence over either adopt path. The lockout risk is real: `oauthPublicOrigin` in `pkg/dashboard/oauth_origin.go` takes `dashboard.public_url` first and `hub.dashboard_url` second, before falling back to the request headers, so with `dashboard.public_url` unset a typo in this field changes where an authorization code is redeemed — and the operator can lock themselves out of the dashboard they typed it into. The downstream consumers the issue names also check out: `pkg/hub/webhook.go` builds the spoke URL from `hive.DashboardURL`, and `pkg/discord/bot.go` uses it for message links. So the only inaccuracies I found were stale line numbers, which I re-derived rather than repeated.

The fix has two halves because a read-only input on its own is cosmetic. On the client, `renderGovHub()` now branches the Dashboard URL row on `hub.enabled`: a hub-managed spoke gets a `readonly disabled` input styled like the existing `NAMESPACE` row, plus the provenance line in the wording this tab already uses for the update block — "Policy source: the hub (heartbeat)" — and a pointer to `dashboard.public_url` as the override that genuinely wins for OAuth and is never touched by the heartbeat. An unmanaged spoke (`hub.enabled == false`) keeps the editable input with its `markDirty` wiring, which is the escape hatch the issue suggests. The row uses only `data-change-action` attributes, so the CSP inline-handler guard still passes (`go test ./pkg/dashboard/ -run 'CSP|Inline'` is green). On the server, `handleGovernorHub` now rejects a `dashboard_url` that differs from the stored value with `409 Conflict` and a message naming `dashboard.public_url` as the alternative, while an unchanged resubmit is a no-op — that matters because the Hub tab PUTs the whole form on every save, so a strict reject-on-present would 409 on unrelated edits. The enabled flag is read after `body.Enabled` is applied, so an operator who turns the hub off and sets the URL in one request gets an unmanaged spoke and the write lands.

On the other hub-ish fields the issue asks about: `NAMESPACE` is already read-only and never sent back on save, and `SNAPSHOT / PREVIEW URL` is not hub-owned in the direction the issue implies — `snapshot_url` appears in the spoke→hub heartbeat payload (`pkg/hub/heartbeat.go`) and in `HiveEntry` on the hub, but `HeartbeatProjectConfig` has no snapshot field and nothing in `cmd/hive/main.go` adopts one, so it is genuinely spoke-owned and stays editable. That is why I used `Closes` rather than `Refs`: the two fields the issue flagged for a second look need no change, and I have said so rather than quietly skipping them.

Tests are in `pkg/dashboard/hub_dashboard_url_owned_test.go`. One pins the markup — the enabled arm is `readonly disabled`, carries the provenance line and the `dashboard.public_url` pointer, and has no `markDirty` wiring; the disabled arm keeps it. Four pin the handler: a differing write while the hub is enabled 409s and leaves both the config and the resolved OAuth origin untouched, a same-value form resubmit still applies the rest of the body, an unmanaged spoke's write lands, and a single PUT that disables the hub and sets the URL lands. I mutation-tested both halves. Restoring the unconditional `cfg.Hub.DashboardURL = body.DashboardURL` produced a genuine `--- FAIL: TestGovernorHubRejectsDashboardURLWriteWhileHubEnabled` (not a compile error); restoring the editable input in the `hub.enabled` arm produced `--- FAIL: TestHubDashboardURLRenderedReadOnlyWhenHubEnabled` with all four assertions firing. Both passed again after reverting the mutations.

One pre-existing test needed updating and a reviewer should look at it: `TestHandleGovernorHubUpdate` in `dashboard_handlers_extra_test.go` PUT `enabled:true` together with a differing `dashboard_url`, which is precisely the case now rejected. I dropped `dashboard_url` from that smoke-test body and left a comment pointing at the new coverage, rather than weakening the guard to keep an incidental assertion green. The other thing worth a careful read is the 409-versus-silent-ignore choice: silently dropping the write would have kept every existing client working, but it would be a silent failure, and the whole point of the issue is that this field lies to operators about what it does. `go build ./...` and the full `go test ./pkg/dashboard/` pass (56s, including the known-flaky stale-generation lease test). `gofmt -d` on `api_governor.go` still shows only the pre-existing map-alignment drift on `v4`, none of it in my lines, and `check-api-reference-citations.sh --fix` reported zero drift since no route registration moved.

What I could not verify: none of this was exercised against a live hub-managed spoke, so the claim that the heartbeat would have reverted an edit rests on reading `ProjectConfigCallback` rather than on watching an adopt log line appear.


clubanderson:
Closing in favor of #7460, which also closes #7451. Adjudication: this PR gates read-only on `hub.enabled` alone, but a self-hosted spoke that registered with the hub uses `hub.dashboard_url` to *tell* the hub where its dashboard lives — locking the field there is wrong. #7460 gates on hub-proxied OR an actual heartbeat vanity-URL push (`SetHubPushedDashboardURL` wired from both adopt sites, covering OpenShift-Route hosted spokes), refuses the write before applying any bundled field, and exposes `dashboard_url_owner` in the GET API. Thanks — the 409-over-silent-drop reasoning here matches and survives in #7460.

PR #7460, opened 2026-09-17
🌱 feat(dashboard): make the hub-owned Dashboard URL read-only on the Hub tab

### LINKED ISSUE #7451: 🌱 Settings → Hub: DASHBOARD URL is editable but the heartbeat overwrites it, and a bad value can break OAuth login first

Danathar:
## Summary

**Settings → Hub** rendered **Dashboard URL** as an ordinary editable input. On a hub-managed spoke the hub owns `hub.dashboard_url`: the heartbeat's project-config push adopts and persists the hub's vanity URL whenever it differs, so an operator's edit is reverted by the next beat — and until it is, the edited value is the second-choice OAuth callback origin (`oauth_origin.go`), so a typo points the code redemption at a host that is not this server and can lock the operator out of the dashboard they typed it into. This PR makes the server say who owns the field, refuse a hub-owned change, and render the field read-only with its provenance — pointing at `dashboard.public_url`, the operator-owned field that already wins for OAuth and is never overwritten, for anyone who genuinely needs a different host.

Closes #7451.

## What changed

**Ownership (`pkg/dashboard/api.go`).** `hubOwnsDashboardURL()` is true on a hub-proxied (hosted) spoke — the hub's ingress terminates the hostname — and on any spoke that has received a vanity URL on the heartbeat in this process (`SetHubPushedDashboardURL`, wired from both adopt sites in `cmd/hive/main.go`; this covers hosted spokes reached over an OpenShift Route, which are not hub-proxied). A self-hosted spoke that merely registered with the hub keeps the editable field — there `hub.dashboard_url` is how the spoke *tells* the hub where its dashboard lives, so gating on `hub.enabled` alone (the issue's fallback suggestion) would have been wrong.

**API (`api_governor.go`).** `GET /api/config/governor` reports `hub.dashboard_url_owner` (`"hub"` | `"spoke"`) and `hub.public_url`. `PUT /api/config/governor/hub` refuses a hub-owned `dashboard_url` change with `409` **before applying any field** (so a bundled `hub.url` change is not half-applied), naming `dashboard.public_url` as the right knob. Echoing the current value back — what the read-only tab does — saves normally.

**Hub tab (`static/index.html`).** `renderHubDashboardURLField` renders a disabled input with *Owned by the hub (heartbeat) — read-only here.* (plus the `dashboard.public_url` override when one is set) for a hub-owned field, the way the auto-update block shows its policy source; the editable input survives only for spoke-owned fields. The other hub-derived values the issue mentions were already handled: **Namespace** is read-only, and **Snapshot / Preview URL** is locked while auto-publish is on and is a real operator setting otherwise.

Docs: a row in `hosted-hub.md`'s problems table; `api-reference.md` citations re-pinned with `check-api-reference-citations.sh --fix` (the five-line `Server` struct addition shifted them).

## Testing

`pkg/dashboard/hub_dashboard_url_owner_test.go` (4 tests): ownership by default, hub-proxied, blank push (ignored), real push; the `409` leaves `dashboard_url` *and* a bundled `hub.url` untouched, echoing the value saves other fields, and `GET` carries `dashboard_url_owner` + `public_url`; a self-hosted spoke still accepts the edit and reports `owner=spoke`; the tab helper has both branches and the page has exactly one editable `dashboard_url` input, inside the owner gate. Per CONTRIBUTING's test policy, on the parent commit a hub-proxied spoke accepts the edit and the tab is unconditionally editable (both probed and failing there). `go test ./pkg/dashboard/` green, `cmd/hive` builds, `check-inline-js.js` and the citation check pass, `go vet` clean.


clubanderson:
Merged `v4` into this branch to clear a conflict that appeared as other PRs landed.

The only conflicting file was `src/docs/api-reference.md`, whose `file:line` citations drift whenever `pkg/dashboard` route registrations move. I took v4's copy and regenerated rather than hand-merging line numbers:

```console
```

That rewrote 11 drifted citations with 0 left unresolved, and the checker then passes clean.

This is a merge, not a rebase — nothing was force-pushed and none of your commits were rewritten. Verified afterwards: `go build ./...` clean and `go test ./pkg/dashboard/` green (58.7s).

clubanderson (approved):
Ownership model handles the self-hosted-registered case correctly; 409-before-apply and test coverage are solid. LGTM.

Issue #7455, opened 2026-09-17
🐛 Duplicate PRs: kick PR list uses a flat prefix cut, so 14 of 16 repos are never visible to any agent


clubanderson:
## Duplicate PRs: the kick's PR list is truncated with a flat prefix cut, so most repos are never visible to any agent

Agents open PRs that duplicate work already open in the same repo. The usual reading is that the agent ignored the list it was given. On `projectbluefin` that is not what happened — the PR the agent needed to see was never in the list.

### The mechanism

`formatPRListWithPolicy` in `pkg/scheduler/scheduler.go` builds the `${PR_LIST}` substitution by walking open PRs in repo order and cutting at the cap:

```go
limit := s.prCap()
...
if i >= limit { break }   // flat prefix cut
```

The cut is positional. It has no per-repo awareness, so it does not truncate each repo — it truncates the *repo list*. Every repo past the cut contributes nothing at all.

### Measured on the live spoke

`hive-hosted-hosted-projectbluefin-knuckle-gjvq`, 16 repos, open PR counts as of 2026-09-17:

| repo | open | cumulative |
|---|---|---|
| common | 32 | 32 |
| bluefin | 37 | 69 |
| bluefin-lts | 32 | 101 |
| actions | 10 | 111 |
| testsuite | 40 | 151 |
| server | 33 | 184 |
| fsdk-containers | 26 | 210 |
| finpilot | 19 | 229 |
| dakota-iso | 45 | 274 |
| utah | 38 | 312 |
| utah-packages | 35 | 347 |
| documentation | 11 | 358 |
| website | 16 | 374 |
| review | 12 | 386 |
| bluefin-bling | 15 | 401 |
| chairlift | 22 | 423 |

**423 open PRs.** At the default `max_prs: 50` the cut lands at position 50 — inside `bluefin`, the second repo. `common` is shown in full, `bluefin` is shown to 18 of its 37, and **the other 14 repos contribute zero rows to every kick, every time.** 88% of open PRs are structurally invisible.

This is stable, not random: the order is deterministic, so the same 14 repos are invisible on every kick. An agent working in `chairlift` has never once seen a `chairlift` PR.

### The duplicate this actually produced

`projectbluefin/utah#108` (human, opened 16:55) and `projectbluefin/utah#110` (scanner, opened 21:22) touch the same two files for the same goal, +25/-1 against +24/-1.

[...]

PR #7467, opened 2026-09-17
🐛 Release the manager lock between kick keystroke chunks (209s fleet-wide stall per kick)

### LINKED ISSUE #7417: 🐛 Global Manager.mu held across tmux/subprocess restart stalls every agent settings dialog (11s+ lock holds)

clubanderson:
Fixes #7417.

## The freeze is not the kick being slow — it is the kick being slow while holding the global lock

`SendKick` takes `m.mu` and holds it through `deliverKickLocked`, which types the kick into tmux in `chunkSize`-rune chunks with a `chunkDelay` sleep between each (`manager_kick.go:908-938`). Lock hold therefore scales with prompt size, about one second per 400 runes. The sleeps are essentially all of it; the actual work is microseconds.

## Measured on the live projectbluefin spoke

With `max_prs` raised to 350 against 423 open PRs, scanner's 18:14 kick:

```
18:14:58.101  audit: governor kicking agent          <- m.mu taken
18:15:00.129  archived kick log
              ... no normal server activity at all ...
18:15:20.729  hub heartbeat collect timed out; loop keeps ticking so liveness stays green
18:15:20.730  hub heartbeat collect timed out — sending LAST-GOOD cached stats (marked stale)
18:15:33.414  copilot sniff: client request read timed out
18:15:44.706  proxy client request read timed out
18:15:45.739  copilot sniff: client request read timed out          (+3 more)
18:16:24.847  copilot sniff: client request read timed out
              ... still nothing ...
18:18:27.855  agent metrics collected                 <- m.mu released
```

**209 seconds.** Two consequences here are worse than a sluggish dashboard:

- **The hub heartbeat shipped stale stats.** The spoke sent `LAST-GOOD cached stats (marked stale)` to stay online, so the hub's view of it is knowingly wrong for three and a half minutes while the spoke still reports healthy.
- **The LLM proxy dropped requests.** Six read timeouts inside the window: one agent's kick delivery starves other agents' in-flight model calls.

## The change

Nothing needs the manager lock while we are waiting for the CLI to keep up, so release it across each pause. Delivery paces itself exactly as before — the pacing is deliberate, so the CLI's input handling does not overrun — but the lock is now held only for the individual tmux writes.

This is not a new pattern. `SendKick` already unlocks around `waitForInputPromptForAgent` and revalidates with `kickEpochChangedFn` before continuing (#7363); `kickChunkPauseUnlocked` is the same shape applied to the chunk loop.

Two details that follow from releasing the lock:

**Revalidation checks identity, not just the epoch.** If the agent restarted mid-send, the remaining chunks would be typed into a pane belonging to a different run. A replacement `AgentProcess` for the same name starts its `kickEpoch` at zero, so an epoch comparison alone could alias a fresh object as unchanged — the check compares the pointer as well. On failure the remaining chunks are abandoned and `LastKick` is left unset, so a kick that never landed is not recorded as delivered.

**Delivery needs an exclusive claim.** Two concurrent deliveries to one pane become reachable for the first time here; previously the lock alone made it impossible. Interleaved chunks would produce a single corrupt prompt. `kickDelivering` therefore moves from `Store(true)` to `CompareAndSwap(false, true)`, refusing the second typist. It was already an atomic read by the pane poller, so no new synchronisation is introduced.

## Why this matters now

#7455 is the duplicate-PR root cause: the kick PR list was cut with a flat prefix, so 14 of 16 repos were never visible to any agent, and agents duplicated work they could not see. The fix (#7456) gives every repo a fair share and offers `max_prs: 0` for full visibility — but "show the agent more of the truth" directly lengthens the stall this issue describes. This unblocks that one.

## Tests

`kick_chunk_pause_test.go` is new. The load-bearing one asserts a second goroutine can actually acquire `m.mu` while a pause is in flight — if that regresses, the fleet-wide stall is back and nothing else would notice. Also covers the lock-held-on-return contract on both paths, all four revalidation outcomes (unchanged, restarted, removed, replaced-at-same-epoch), and the exclusivity of the delivery claim.

Passes under `-race`. Full `./pkg/agent/` suite green (355s).


clubanderson:
Second live measurement of this fault, captured on the projectbluefin spoke — this
one with an operator action blocked in the middle of it, which shows the blast
radius better than the first.

An 84,842-character kick to `quality` held `m.mu` for **221 seconds**:

```
19:07:14.605  audit: governor kicking agent   agent=quality trigger=governor-eval
19:08:16.809  hub heartbeat collect timed out; timeout=10s
19:08:16.811  hub heartbeat collect timed out — sending LAST-GOOD cached stats (marked stale)
19:10:16.809  hub heartbeat collect timed out; timeout=10s
19:10:16.811  hub heartbeat collect timed out — sending LAST-GOOD cached stats (marked stale)
19:10:55.856  audit: agent kicked             name=quality message_len=84842
```

84842 / 400 runes per chunk ≈ 212 chunks × 1s ≈ 212s of pure sleeping, against a
measured 221s hold. The work itself is microseconds; essentially the entire hold
is `chunkDelay`.

**What it blocked.** An operator was changing the reviewer's model from the
dashboard during this window. The toast "Switching reviewer model →
claude-fable-5 (restarting session)…" span for over a minute, and the change
landed **one second** after the lock was released:

```
19:10:55.856  audit: agent kicked            name=quality        <- lock released
19:10:56.181  audit: model set               agent=reviewer model=claude-fable-5 trigger=dashboard-api
```

So this is not only a frozen dashboard. In this window:

- the **hub was told twice** that this hive's stats were stale, while the hive
  itself looked healthy
- **proxy client requests timed out** for two different agents
- an **operator action was blocked** for the duration

There is also a knock-on failure. The reviewer's own kick was dispatched at
19:10:55.987, immediately behind quality's, and collided with the restart that
the now-unblocked model switch triggered:

```
19:11:00.985  WARN failed to send kick  agent=reviewer
              error="kick cancelled by restart: agent reviewer restarted while the kick was waiting for its input prompt"
19:11:01.063  WARN agent CLI crashed or stuck on consent screen, restarting before kick  name=reviewer
```

That kick was lost. Serializing every kick behind one lock means a long delivery
does not just delay the next one — it bunches them against whatever operator
action was waiting on the same lock.

Prior measurement for comparison: 209s on a 350-PR kick (#7417).

PR #7488, opened 2026-09-17
📖 docs(design): record the v6 GitHub @-mention trigger design

### LINKED ISSUE #7483: 📝 v6: @-mention triggers — let a human summon an agent from a GitHub thread
## Summary

Hive has no way for a human to **summon an agent by @-mentioning it** on a GitHub issue or PR. Every trigger today is hive-initiated. A human who wants an agent to look at something has no inbound path — they have to wait for a cadence kick to happen to select their PR, or ask an operator to drive the dashboard.

This is proposed as a **v6 feature**, not something to retrofit into v4/v5.

## Why this is the missing half

The interesting part is that **we already built this for Linear** and never built the GitHub equivalent.

`pkg/linearagent/oauth.go:54-56` requests `app:assignable,app:mentionable`, and `pkg/linearagent/webhook.go` turns an inbound mention into an agent session:

```go
// PromptText returns the best available task text for the event: the user's
// message on `prompted`, else Linear's promptContext, else the mention
// comment, else the issue title.
```

So on Linear a human can mention the agent and it wakes up with the mention as its prompt. On GitHub — where the overwhelming majority of the fleet's work actually happens — there is nothing.

The conversational half is already in place too. `AgentCapabilities.Converse` (`pkg/agent/capabilities.go:25-32`) exists precisely for this shape:

```go
// An ADVISORY agent with Converse can reply on a thread it was
// mentioned in; it still cannot file, edit or relabel anything.
```

The code comment already describes replying to a mention. We have the capability to *reply*; we have no way to *hear*.

## What triggers exist today

All hive-initiated, all poll or file-relay:

| Path | Direction | Trigger |
|---|---|---|
| Governor cadence | outbound | timer + mode thresholds |
| Review swarm dispatch | outbound | queue scan |
| `issue_request_watcher.go` | outbound | agent writes a file |
| `pr_request_watcher.go` | outbound | agent writes a file |
| `review_request_watcher.go` | outbound | agent writes a file |
| `merge_request_watcher.go` | outbound | agent writes a file |
| Linear webhook | **inbound** | **human mention** |

The last row is the only inbound trigger in the system, and it is not GitHub.

## Proposed shape

- Inbound GitHub events (`issue_comment`, `pull_request_review_comment`, `issues`) delivered by webhook or polled from the App installation.
- A mention of the App login, optionally with an agent selector — `@hive review this`, `@hive ask scanner ...` — resolves to a target agent. Note `pkg/config/config.go:1681` already contemplates resolution "when empty and exactly one agent is..." for delegations and mentions, so some of the routing vocabulary exists.
- The mention body becomes the kick text, with repo/issue/PR/comment id as structured context — the same payload shape the review swarm already builds.
- Reply goes out through the existing sanctioned path (review/issue request watcher, App-authored, on the audit trail), gated by `Converse`, so an ADVISORY agent can answer a mention without gaining any artifact-production authority.

## Guards this will need

Mentions are attacker-reachable in a way cadence kicks are not — anyone who can comment on a public repo can type the App's name.

- Allowlist who may summon (org membership / write access / explicit list), not "any commenter".
- Rate limit per user, per repo, and per thread; reuse the per-thread attempt cap `review_request_watcher.go` already enforces for thread replies.
- Ignore mentions authored by the App or any configured `classification.review_bots` login, or two agents will mention each other into a loop.
- The mention body is untrusted input that becomes a prompt — it must flow through the existing canary/ioscan fail-closed handling, exactly as review bodies do (`kubestellar/hive#4960`).
- Never let a mention escalate mode. A mention is an *input*; authority stays on the ladder + capabilities.

## Out of scope

- Auto-merge on mention.
- Mode escalation by mention.
- Any change to v4/v5 trigger behavior.

## References

- `pkg/linearagent/webhook.go` — the inbound mention model to mirror
- `pkg/linearagent/oauth.go:54-56` — `app:mentionable`
- `pkg/agent/capabilities.go:25-36` — `Converse`, documented in terms of mentions
- `pkg/config/config.go:1681` — existing agent-resolution note for "delegations and mentions"
- `pkg/github/review_request_watcher.go` — the App-authored reply path to reuse

Danathar:
## Summary

#7483 proposes, for **v6**, the first inbound GitHub trigger: a human summons an agent by @-mentioning the App on an issue or PR — the GitHub twin of the Linear agent-session path that already exists. This PR is the **design record** for it, in `src/docs/design/` with the status that index requires (proposed for v6, design only), so the feature has a reviewed plan before anyone builds it. It changes no behaviour on v4 or v5.

Closes #7483 (linked post-merge so the issue shows its PR; the keyword does not re-fire after merge, and #7483 deliberately stays OPEN — it is the v6 feature, this was only its design)

## What the page settles

- **Mirror `pkg/linearagent`, don't reinvent it.** The receiver/responder split, `PromptText` precedence, the `session_agent` resolution rule, the kick shape and its 10,000-rune bound, `Converse` as the reply gate — each cited to the v4 line that implements it today.
- **Poll first, webhook as an accelerator.** Most spokes are on pull-only clusters with no public URL for GitHub to deliver to (the [wrapped master delivery](../blob/v4/src/docs/design/master-delivery-wrapped.md) premise), so a webhook-only design would work on the hosted hub and nowhere else. The poller lists comments since a per-repo watermark on each eval tick; the phase-3 webhook feeds the same handler and just advances the watermark.
- **A mention grammar** with `ask <agent>` routing. Bare verbs are prompt text, not routing — intent classification is what the agent is for.
- **No new outbound write surface.** Replies go through the issue-request (`Kind: "comment"`) and review-request watchers as the App bot: canary-gated, scrubbed, audited, gated by `Converse` at the proxy exactly as today. The one new write is an 👀 reaction as the ack, posted only after every guard passes.
- **Seven guards, each mapped to something that already has a runtime:** who may summon via the dashboard role list (`trustedMergerFunc`'s lookup), fail closed; no oracle on a declined mention; per-user / per-repo / per-thread limits with the per-thread cap reusing `classification.review_bots.max_attempts_per_thread` (the counter *is* the thread's App-authored replies, no state file); a loop list of the App login, `review_bots.logins` and `[bot]` logins; `ioscan.EnforceInput` on the body (ADR-0008); no escalation of any kind; `created`-only, never `edited`.
- **The #5591 lesson, named.** `ChannelConfig` once accepted `webhook`/`discord`/`schedule`/`bead` types with no runtime, and agents declaring them sat dormant with no diagnostics. The `mention` channel type ships with its runtime in the same change, or not at all.
- **Three open questions for maintainers:** the role floor for summoning (`read-write` vs `merger`), in-thread vs PR-level reply for `pull_request_review_comment` mentions, and whether `ask a,b` fan-out is wanted (proposed: no).

Every code reference was checked against `v4` at `cf57dd3d3`; the page says so and names symbols beside line numbers so drift is survivable.

## Files

- `src/docs/design/github-mention-triggers.md` — the record.
- `src/docs/design/README.md`, `src/docs/README.md` — index entries with status, following the pattern the Discord promotion design uses.

## Checks

- `python3 src/scripts/check-docs-links.py src/docs` → `checked 154 files under src/docs / all relative links and anchors resolve`
- Docs only — no Go changes, no tests to run; `no-changelog` per CONTRIBUTING's docs exemption.


clubanderson (commented):
I picked up #7483 independently and was about to write the same design record. This PR already does it, and does it well, so I am declining to open a duplicate and reviewing instead. What follows is an independent verification of the page against `v4` at `cf57dd3d3` plus the gaps I found.

## Verification

I checked every code citation on the page against the tree, not against the issue. All twenty-four resolve to the symbol they claim: `Scopes = "read,write,app:assignable,app:mentionable"` at `oauth.go:56`, `WebhookReceiver.ServeHTTP` at `webhook.go:164`, `SessionEvent.PromptText` at `webhook.go:105`, `Responder.HandleSessionEvent` and `buildKickMessage` at `responder.go:102`/`225`, the `SessionAgent` resolution comment at `config.go:1681`, the `Converse` doc comment at `capabilities.go:30`, `AllowedByModeCaps` at `rules.go:249`, `IssueRequest.Kind` at `issue_request_watcher.go:70`, `CreateIssueComment` at `client.go:1118`, `handleGitHubWebhook` at `hub/webhook.go:53`, the removed declarative channel types at `config.go:660`, `AuthorizedRole` at `config.go:4182`, `trustedMergerFunc` at `main.go:8076`, `EnforceInput` at `ioscan/enforce.go:24`, `Scheduler.enforceIssueText` at `scheduler/ioscan_enforce.go:122`, `filterReviewThreads` at `review_threads.go:131` and the `MaxAttempts()` denial at `review_request_watcher.go:374`. The issue's own citations are the ones that had drifted (`capabilities.go:25-32`, `oauth.go:54-56`); this page corrected them silently, which is the right call. `python3 src/scripts/check-docs-links.py src/docs` passes for me too — 154 files, all links and anchors resolve — and both index entries carry the status string the design index requires. `Refs` rather than `Fixes` is correct: the issue is the feature, this is its plan.

I also confirmed the guard-1 identity namespace holds, because the whole allowlist rests on it. `identityMatchKey` lower-cases and strips only the `github:` prefix, so a GitHub comment author matches an allowlist entry in either the bare or the hub's canonical form, while `ibmid:`/`google:` entries keep their prefix and therefore can never accidentally match a GitHub login. That is fail-closed and exactly what this guard needs — I'd add a sentence saying so, since a future reader could reasonably assume the dashboard list is a separate identity space and conclude the guard is unsound.

## The one thing I think is a genuine defect

Guard 7 says only the `created` action is a trigger, so a mention cannot be re-armed by editing an old comment. The poll transport cannot honour that as written. `Issues.ListComments`' `Since` filters on **`updated_at`**, not `created_at`, and the poller's watermark is described as "the newest `updated_at` seen". So a comment written a year ago and edited today to add `@hive` comes back in the poll window, has never been seen before, and therefore sails past the node-id dedupe and fires — which is precisely the re-arm guard 7 forbids. The benign case is bad too: every ordinary edit resurfaces an already-handled comment, so suppression rests entirely on a dedupe set that survives restarts and is never trimmed.

The fix is small but belongs in the page, because the page's premise is that both transports feed the same handler: in the poll path, compare each comment's `created_at` against the watermark and drop anything older, and persist the dedupe set alongside the watermark rather than treating it as incidental. The webhook path gets `action` for free; the poller has to reconstruct that predicate itself.

## Smaller corrections

The reaction ack is not the reuse it reads as. `fleet_report.go:42` is the only reaction call site in `pkg/github` and it is `Reactions.CreateIssueReaction` with `+1` at **issue** level; there is no comment-level reaction helper, so `CreateIssueCommentReaction` is new client surface, not an existing one. The page is honest that the ack is a new write, but the citation implies more precedent than exists. Worth copying the `isAlreadyExists(err)` tolerance from that call site while you are there, since a re-poll will retry the same reaction.

Guard 1 has a default-value trap that will make the feature look dead. `AuthorizedRole` defaults an entry with no explicit role to `owner` for the **first** entry and `read` for every other. With the proposed `min_role: read-write`, a typical allowlist authorizes exactly one summoner — the owner — and silently drops every other maintainer, who will conclude the feature is broken. That is the #5591 dormancy failure mode this very page warns about, arriving through config defaults instead of a missing runtime. I'd make the decline audit entry carry the reason at that granularity (`unauthorized: role=read < read-write`) and say in the config section that roles must be written explicitly.

The "one or two calls per repo per tick" claim is true only via a non-obvious API detail: `Issues.ListComments` and `PullRequests.ListComments` list repo-wide when the issue/PR number is `0`. Worth stating, or an implementer will reach for the per-issue form and write a call per issue. Relatedly, nothing on the page addresses quota, and #7439 is open right now to keep the App's quota surviving a cycle. A per-repo-per-tick unconditional list is a standing new cost against the budget that PR is defending; the poller should go through the conditional-request/ETag cache from #7430 from day one rather than be retrofitted later.

Finally, the trigger table inherits the issue's claim that Linear is the only inbound trigger. The dashboard already summons agents on a human's action — `api_linear_agent.go:104`, `inception_handlers.go:593` and the `ocSendKick` control in the dashboard UI all reach `Manager.SendKick`. The distinguishing property is not "inbound" but **attacker-reachable**: every existing human trigger sits behind the operator's authenticated console, and this one would not. Saying it that way makes the guards section follow from the problem statement instead of sitting beside it.

## One guard I'd add

Detection is author-blind in a way guard 4 cannot cover. Two cases: a login-prefix collision, where `@hive` matches `@hive-ci` or an address in pasted text unless matching is word-boundaried; and, more importantly, mentions inside blockquotes and fenced code. GitHub's "Quote reply" reproduces the quoted body verbatim, so a human quoting an earlier comment that named the App re-summons the agent, and a human pasting a log or a config snippet containing the App login summons it by accident. The loop guard cannot catch either, because the author is a genuine authorized human every time. Skipping `>`-quoted lines and fenced/inline code spans before scanning for the login closes both, and it is the convention most GitHub bots already follow. That is the only guard I found that neither the issue nor this page lists.

None of the above changes the shape of the design, which I think is right — particularly poll-first, and the refusal to add a `mention` channel type ahead of its runtime.

clubanderson:
Following up on my review above with the two checks it did not spell out explicitly: guard-by-guard coverage against the issue, and confirmation that this stays a proposal.

**Guard coverage is complete.** Every guard #7483 asks for has a home on the page, and each is tied to a mechanism that already has a runtime rather than to new policy. The allowlist is guard 1, keyed on `DashboardConfig.AuthorizedRole` at `read-write` or above with `github.mentions.summoners` as the widening knob and "unset means off" as the default — the same lookup `trustedMergerFunc` uses, so it is not a second permission model. Rate limiting is guard 3 on all three keys the issue names, and the per-thread cap genuinely reuses `classification.review_bots.max_attempts_per_thread` rather than reimplementing it: the counter is the thread's App-authored replies, which is how `filterReviewThreads` and the watcher's `MaxAttempts()` denial already work, so there is no new state file to desynchronise. Loop prevention is guard 4, covering the App login, `review_bots.logins` and any `[bot]` suffix. Untrusted input is guard 5, routing the body through `ioscan.EnforceInput` before it becomes kick text, which is the same treatment `Scheduler.enforceIssueText` gives issue titles under ADR-0008, with outbound replies canary-gated by the existing comment path. No escalation is guard 6, and the page is right to frame it as a boundary rather than a later phase. The page then adds two the issue did not ask for — no-oracle silent drops so the allowlist cannot be enumerated, and `created`-only with node-id dedupe — which is the correct instinct on both counts, though as I noted above the `created`-only one does not survive contact with the poll transport as currently described.

So on the four items I would have flagged as blocking if they were missing — who may summon, rate limits, mention loops, and prompt injection from untrusted mention text — the document is silent on none of them. My remaining concerns are the `updated_at` watermark defeating the edit guard, and quote/code-fence detection, both above.

**It stays a proposal.** The diff is three files and 354 added lines, all under `src/docs`: the design page plus one index line in `src/docs/README.md` and a status entry in `src/docs/design/README.md`. Nothing under `src/pkg` or `src/cmd`, no Go, no config schema change — so no v6 behaviour has been retrofitted onto `v4`, which was the main thing to get wrong here. The design index does require a status string per entry and this entry carries one ("proposed for v6, design only"), matching the pattern the Discord promotion and token-metering entries use. `no-changelog` is the right label for a docs-only change under CONTRIBUTING's exemption rather than inventing a fragment category.

I would merge this once the watermark point is resolved, either by amending the transport section or by recording it as a fourth open question.

Issue #7478, opened 2026-09-17
🌱 dashboard: a green PR pill should mean "the sweep would merge this now", with amber for mergeable-but-outstanding


Danathar:
Follow-up to #7471 / #7475. That fix makes the ✓ on a REPOSITORIES PR pill track GitHub's verdict correctly. This issue is about what the ✓ should mean once it does.

## The problem

After #7475, ✓ means `mergeable_state ∈ {clean, has_hooks, unstable}`. `unstable` is "no conflicts and nothing *required* is failing" — on a repo without branch protection that includes PRs with red CI and no review. Two from today's projectbluefin sample, both green after #7475:

| PR | GitHub state | checks | review |
|---|---|---|---|
| bluefin#1253 | unstable | 1 failing | none |
| server#167 | unstable | 1 failing | none |

The auto-merge sweep will not merge either of them: `CIStatus == "failure"` puts a PR in the `ci-failing` bucket unless the operator has declared a required-check set and only optional checks are red. The sweep also excludes held PRs, drafts, PRs without review approval when `require_review_approval` is on, and PRs with a failing DCO or intent verdict. The card checks none of that. So an operator sees a green pill that sits there for days, and nothing on the card says why.

Green on an operator-facing card reads as "this will merge". The only honest definition of that is the sweep's own verdict.

## Proposal

1. **Green = the sweep would merge this now.** The sweep already computes exactly that and writes `merge-eligible.json`; the card should key off that verdict instead of re-deriving a looser one from `mergeable_state`.
2. **Three states, not two.** Green for sweep-eligible. Amber for "mergeable per GitHub but something outstanding" — red optional checks, review pending, held, DCO unknown — with the reason in the tooltip. Nothing for dirty / blocked / unknown. Amber is the honest colour for `unstable`.
3. **Queue auto-merge only on green or amber.** #7475 already disables it on `"no"`; this keeps that.

Not proposing any change to the sweep's policy of ignoring non-required checks — that was a deliberate call and it is the display that misleads, not the merge decision.

Issue #7471, opened 2026-09-17
🐛 bug(dashboard): PR pills show the green ✓ for mergeable="no" — the frontend still treats the tri-state string as a bool


Danathar:
## What happens

On the dashboard's REPOSITORIES cards, nearly every open PR pill shows the green ✓ ("merge eligible") and a **Queue auto-merge** button. Open the PRs and many are not ready at all: conflicts, review required, failing checks.

Screenshot from the projectbluefin hive (`hosted-projectbluefin-common-nmq5`) today: ~90% of the pills are green.

## Sampled against GitHub

I picked PRs off the screenshot that show ✓ and asked GitHub (`gh pr view --json mergeStateStatus,reviewDecision,statusCheckRollup`):

| PR (shows ✓ on the card) | mergeStateStatus | review | checks |
|---|---|---|---|
| bluefin#1259 | **DIRTY** (conflicts) | — | 2 failing |
| bluefin-lts#603 | **BLOCKED** | REVIEW_REQUIRED | — |
| dakota-iso#169 | **BLOCKED** | REVIEW_REQUIRED | 3 failing |
| testsuite#825 | **BLOCKED** | REVIEW_REQUIRED | 3 failing |
| utah#105 | **BLOCKED** | REVIEW_REQUIRED | 13 passing |
| bluefin#1253 | UNSTABLE | — | 1 failing |
| server#167 | UNSTABLE | — | 1 failing |
| bluefin#1242 | CLEAN | — | 4 passing |

REST `mergeable_state` agrees (`blocked` / `dirty` / `unstable`), so this is not a GitHub-side disagreement — the hive has the right verdict and the card draws it wrong.

## Cause

#2365 (2026-07-31) changed `PullRequest.Mergeable` from a `bool` to a tri-state string (`"yes"` / `"no"` / `""` for unknown) so that "never fetched" could not read as "no". That was the right change on the Go side, but the dashboard was never updated:

- `src/pkg/dashboard/static/index.html` still tests truthiness — `p.mergeable ? '✓' : ''` (around line 8116), and the same `filter(p => p.mergeable)` in the mergeable counters (lines ~4753 and ~6116). In JavaScript `"no"` is truthy, so every PR whose mergeability *was* fetched gets the ✓, whichever way it came back. Only PRs still at `""` (unknown) show no icon.
- `src/pkg/dashboard/status_builder.go` `mergeableCount` does `m["mergeable"].(bool)`, which never matches a string, so that count is now always 0.

[...]

PR #7494, opened 2026-09-18
🌱 v5: top-up from v4 through 82308e22f (12 commits, v4.53.0)

clubanderson:
Preceded by a `-s ours` merge of v4@cfc3a6989 so git records that #7481
already carried that content (its commit was created after `checkout -b`
dropped MERGE_HEAD and so had a single parent).

- cmd/hive: v4's classifyMergeEligibility/MergeVerdict refactor (#7478)
  with v5's holdDriftPRs parameter (#5589), eligiblePR/failingPR
  Labels+CreatedAt, and the shared intent.Verdict.BlocksMerge gate (#6258)
- agent: manager_poll.go extraction (v4) merged at declaration level;
  startSandboxKickLocked keeps v5's effects mutation boundary and takes
  v4's TokenTierForRole
- github: PullRequest.MergeableState takes v4's documented block (#7471)
- ci: v2-tests.yml R=4 rest shards + hosted-runner runs-on, keeping v5's
  `!cancelled()` aggregate guard (#6293)
- docs/api-reference.md: v5 route table kept; citations re-pinned


clubanderson:
Reviewed this as the #7463 top-up (I had independently resolved the same merge on `sync/v4-to-v5` before finding this PR, so the notes below are a comparison against a second, independent resolution rather than a read-through).

**The thing a reviewer should fear here — a silently reverted v5 commit — does not appear to have happened.** I checked the merged tree for a sentinel symbol from each v5 divergence that touches a conflicted file, and all of them survive: `issuePriorityNote` (#389248ba, human issues ranked first), `startupKickInFlight`/`startupLaunchQueued` (56edf44b, governor kicks deferred during startup), `enforceHoldGuard` + the `holdDriftPRs` parameter threaded into `writeMergeEligible` (#5589), `webstatic.MichromaFontFaceCSS` (#5565 leaf-package extraction), `checkBlockedThrash` still wired inside `pollTmuxOutputForAgent` (which `thrash_poller_wiring_test.go` asserts, and which v4's extracted `manager_poll.go` did not carry), `ActiveReposFor` (#6203/#6204 per-repo pause), `syncAutoMergePolicyToGitHubClient` (v5's replacement for v4's `SetMergeRequestPolicy`), `limitActivity`/`retained`/`capacity` (#5711), and `sandboxMutation` (#6064 mutation boundary). Conversely the code v5 *deliberately deleted* has correctly stayed deleted: the legacy session-based tmux layer (`pollTmuxOutput`, `watchForTrustPrompt`, `captureTmuxPane` — #5795/#5808) is absent rather than resurrected from v4's side, and the dead trust-tier render helpers (`trustTierColor` et al, #6214) are gone. Those two are the easiest mistakes to make in this merge because v4's side of the hunk looks like an addition; taking v4 wholesale there would have re-added several hundred lines of code v5 removed on purpose.

I also agree with the two judgement calls in the body: keeping v5's `!cancelled()` aggregate guard while taking v4's `runs-on:` in `v2-tests.yml` is right (the other nine `runs-on` lines in that file are already v4's hosted form), and folding v5's separate `intent.Verdict.BlocksMerge` gate into v4's `classifyMergeEligibility` is semantically a no-op, since `BlocksMerge(enforce)` is literally `enforce && v.AgentPR && !v.MergeAllowed()` — exactly what the classifier inlines.

**One real gap, which is why I don't think this fully closes #7463.** This tops up through `82308e22f`, but `v4` has since advanced by two more commits that are not in this branch:

- `bf1c4b6b` — 🌱 make the hub-owned Dashboard URL read-only on the Hub tab (#7451) (#7460)
- `c80eaa69` — 🐛 identify a signed-in visitor on `/api/contribute` on hosted spokes (#7453) (#7457)

Concretely, `SetHubPushedDashboardURL` and `handleAuthReturn` do not exist anywhere in this branch's tree. Both of those land in files that conflicted here (`cmd/hive/main.go`, `pkg/dashboard/api.go`, `pkg/dashboard/api_contribute.go`, `pkg/dashboard/contribute_landing.go`, `pkg/hub/saas_auth.go`), so once this is squash-merged the next v5 Top-up run will conflict again on exactly the same files — which is the failure mode #7463 is about. I'd suggest either merging `origin/v4` again into this branch before landing it, or landing this and immediately opening a follow-up top-up; either way #7463 probably shouldn't be auto-closed by this PR alone.

For what it's worth, I resolved those two commits' conflicts on my own branch and the result is clean: `go build ./...`, `go vet ./...` and the full `go test ./...` all pass (96/96 packages, no failures, including the usually-flaky `TestStaleGeneration_RevokedWorkerCannotOverwriteNewOwner`), `gofmt` reports nothing new, `check-inline-js.js` passes on `static/index.html`, and `check-api-reference-citations.sh` verifies after `--fix` — though note that three citations there (`/openrouter/callback` twice and `/api/hub/delegation-keys`) need hand-pinning to v5's refactored `pkg/hub` layout because `--fix` cannot resolve them on its own. Happy to push that as a follow-up if it's useful.

clubanderson:
Follow-up now that this has landed (`0b5eead7`) — I re-ran the review against the actual merge result on `v5` rather than the branch head, including building and testing it locally rather than trusting CI.

**Verdict: the conflict resolutions are sound. I could not find a single v5 commit whose effect was lost.**

The check I care about for a forward-merge is whether taking v4's side anywhere quietly undid v5's own work, so I did it mechanically rather than by eye: I diffed `b6d2b517` (v5 immediately before the merge) against the merge result, collected every line the merge *removed* from v5, and then asked whether that line still exists anywhere in the merged tree. Code motion and honest refactors survive that test; a silent revert does not. The results all come out clean once explained:

- `pkg/hub/saas.go` looks alarming at 1,611 removed lines but **zero** of them are absent from the merged tree — it is entirely v4's file split, every line landed in a sibling file.
- `pkg/agent/manager.go`: 446 removed, only 4 genuinely gone, and all four are `m.agentMode(x).TokenTier()` becoming v4's role-aware `TokenTierForRole(...)`. That is v4's intended change, not a lost v5 behaviour. The `manager_poll.go` extraction accounts for the rest, and critically `checkBlockedThrash` is still wired inside `pollTmuxOutputForAgent` — v4's extracted copy did not carry that call, and `thrash_poller_wiring_test.go` would have caught it, but it is the exact kind of thing that gets dropped, so worth stating explicitly that it survived.
- `cmd/hive/main.go`: 74 lines gone, all of them v5's standalone `writeMergeEligible` body and its separate `intent.Verdict.BlocksMerge` gate, superseded by v4's `classifyMergeEligibility` (#7478). This one deserved a manual read rather than a symbol count, and it holds up: at `main.go:8678-8685` the classifier still carries the `BlocksMerge(g.enforceIntent)` call, the `#6258` "one shared refusal predicate" comment, and the `ReasonAlignmentMisaligned` rationale path verbatim. v5's `holdDriftPRs` parameter (#5589) is still threaded through the new signature. So the gate moved; it did not weaken.
- `pkg/github/client.go`, `pkg/dashboard/status_builder.go` and the `index.html` hunks lose their `mergeable` bool, which is just #7471's tri-state `Mergeable`/`MergeableState` landing — `openPRIsMergeable` now compares against `github.MergeableYes` and the sparkline reads from it, so the dashboard count is intact rather than silently zeroed.
- `manager_kick.go` loses `time.Sleep(chunkDelay)` under lock, which is v4's #7417 fix, correctly taken.

Positively, the v5 divergences I'd most expect to be clobbered are all still present in the merged tree: `issuePriorityNote`, `startupKickInFlight`/`startupLaunchQueued`, `enforceHoldGuard` + `holdDriftPRs`, `webstatic.MichromaFontFaceCSS`, `ActiveReposFor`, `syncAutoMergePolicyToGitHubClient` (v5's replacement for v4's `SetMergeRequestPolicy`, which does the same two setters), `limitActivity`, and `sandboxMutation`. The code v5 deliberately *deleted* also correctly stayed deleted instead of being resurrected from v4's side — the legacy session-based tmux layer (#5795/#5808) and the dead trust-tier render helpers (#6214) are absent. In `v2-tests.yml`, v5's `!cancelled()` aggregate guard is kept at line 754 while `runs-on:` matches v4's hosted-runner form on all nine jobs, which is the right way round.

On the specific things worth verifying by running rather than reading, at `origin/v5` = `0b5eead7`: `go build ./...` clean, `go vet ./...` clean, and the **full** `go test ./...` passes with 96/96 packages and zero failures (including `TestStaleGeneration_RevokedWorkerCannotOverwriteNewOwner`, which is the usual flake here). `bash src/scripts/check-api-reference-citations.sh` prints "all pkg/<file>:<line> references verified", so that file was genuinely regenerated rather than hand-merged. `check-inline-js.js` passes both the syntax and `no-undef` passes over both script blocks of `static/index.html` (23,164 lines), so the inline JS resolved into something valid.

I'll also note approvingly that this landed as a **true merge commit with both parents** rather than a squash. That matters more than it looks: the reason #7463 keeps recurring is that #7466 and #7481 were squashed, so git had no ancestry record that v4's content was already in v5 and re-conflicted on content that was semantically already there. This merge gives the next top-up a real merge base.

**The one caveat, and the reason I don't think #7463 should be considered closed.** This tops up through `82308e22f`, and `v4` has moved on. I checked out the merge result and ran the workflow's own procedure against current `v4` (`e6347683`), and it still fails:

```
git merge --no-ff origin/v4   ->  Automatic merge failed; fix conflicts
  src/cmd/hive/main.go                        src/pkg/dashboard/api.go
  src/docs/api-reference.md                   src/pkg/dashboard/api_governor.go
  src/docs/design/master-delivery-wrapped.md  src/pkg/dashboard/contribute_landing.go
  src/pkg/agent/manager.go                    src/pkg/dashboard/contribute_ws.go
  src/pkg/agent/manager_pane_signals.go       src/pkg/hub/saas.go
```

`SetHubPushedDashboardURL` (#7451) and `handleAuthReturn` (#7453) are still absent from `v5`, and the remaining drift has since grown past those two. So the next scheduled "v5 Top-up" run will go red again on largely these same paths. Nothing wrong with this PR — it just doesn't reach current `v4` HEAD, and #7463 describes the workflow being red, which it will be again. A follow-up top-up (ideally merged, not squashed, for the same ancestry reason) would actually close it.

Issue #7469, opened 2026-09-17
✨ The reviewer cannot reduce a PR queue: no write path, no cross-PR view, and its verdicts only gate a merge sweep unused spokes never run


clubanderson:
## The reviewer cannot reduce a spoke's PR queue, and on projectbluefin it currently cannot affect anything at all

projectbluefin has **423 open PRs across 16 repos** and no auto-merge. The obvious lever is the reviewer: have it review the backlog, identify duplicates, and tell humans what to close. Tracing the reviewer end to end, it cannot do any of that today — four independent reasons, each sufficient on its own.

### 1. It never wakes

`reviewer.on_demand: true` on the spoke — the only agent of the eight with the flag set. It wakes when a change is proposed, not on a cadence, so it does not touch the standing backlog. #7447 exposed the toggle and is merged to v4, but hosted spokes run v5 and the commit is not there yet, so the toggle is not even reachable from the dashboard on this spoke.

### 2. It has no GitHub write access, deliberately

From `reviewer-advisory.md`:

> Your product is a **structured verdict**, not a GitHub comment and not an issue. You have no GitHub write access and you do not need any.
> Do NOT attempt to merge, approve, label, or comment on GitHub — you have no write access and need none.

This is not an oversight to be patched. The template is explicit that the asymmetry — the verdict can withhold a merge but never cause one — "is what makes you safe to run." Handing the reviewer a write token to post comments would dismantle the property its whole design rests on.

### 3. It has no cross-PR view, also deliberately

The reviewer judges **one** change at a time, and is specifically denied the PR body:

> You do **not** get the PR body or the author's rationale. This is intentional. The arm of the experiment that saw the author's own explanation performed *worst of all three* — zero real defects found.

Duplicate detection is inherently a comparison *between* PRs, using exactly the intent signal the reviewer is designed not to read. A per-PR, body-blind reviewer is the wrong instrument for it — not a weak one, the wrong kind.

### 4. Its verdict's only consumer is merge-eligibility, which this spoke does not use

The path is: reviewer → `PerspectiveReport` → `review.CollectAndWrite` → `Artifact` → `HasAggregateApproval(repo, number, headSHA)` → `writeMergeEligible` (`cmd/hive/main.go:9147`):

```go
if requireReviewApproval && (!reviewLoaded || !reviewArtifact.HasAggregateApproval(fullRepo, pr.Number, pr.HeadSHA)) {
    continue
}
```

[...]

PR #7499, opened 2026-09-18
🌱 narrow the reviewer token back to pull_requests: read now that the audited relay works

### LINKED ISSUE #7487: 🌱 narrow the reviewer token back to pull_requests: read now that the audited relay works
## Summary

#7472 gave the `reviewer` token tier `pull_requests: write` so an advisory reviewer could post its verdict. It merged before we established that hive **already had** a sanctioned route for this. Now that #7486 makes that route usable, the token grant should be narrowed back to `pull_requests: read`.

## Why

Hive's pattern everywhere else is that the **governor performs the write, not the agent**: `issue_request_watcher.go`, `pr_request_watcher.go`, `merge_request_watcher.go` all take a request file from an agent and execute it with the App token. `review_request_watcher.go` is the review analogue, and its doc comment is explicit that this is the intended route:

> An agent that wants to review a PR writes a request file here **INSTEAD of running `gh pr review` from its own shell**. The hive's watcher submits the review with the App token, so the review is authored by the App bot AND — unlike a direct agent-CLI review, which the hive never observes — it is recorded on the audit/activity trail.

With #7486 merged, an `ADVISORY` + `converse: true` reviewer can use that relay. The direct-token grant from #7472 is then redundant — and not neutral:

- **Attribution.** A relay review is authored by the App bot. A direct-token review is authored by the agent's own identity, outside the App's activity trail.
- **Observability.** The hive never sees a direct `gh pr review`. It is not on the audit trail, so it does not appear in activity counts or SLA accounting.
- **Canary / exfiltration.** The relay runs review bodies through `scanCanaryText` with the fail-closed contract added in `kubestellar/hive#4960`. The direct path skips it.
- **Incentive.** Leaving both routes open means the easier one wins, and the easier one is the one we cannot see.

## Proposed change

- Revert the `pull_requests: write` grant in the `reviewer` case of `pkg/github/app.go` back to `read`.
- Keep the `reviewer` tier itself — it is still the right tier, and `contents: read` / `metadata: read` are correct.
- Keep `pkg/agentmode.TokenTierForRole` and its tests; the role plumbing is independent of the permission level.
- Confirm the reviewer's config carries `converse: true` before landing, so the relay route is live when the direct route closes.

## Sequencing

Do **not** land this until a reviewer has actually posted through the relay on the projectbluefin spoke and it is observed on the audit trail. Narrowing the token first would leave the reviewer with no working route at all — which is the state #7485 described.

## Verification note

The permission assertions from #7472 already exist in `pkg/github/reviewer_tier_permissions_test.go` and pin `pull_requests` presence, so this change is a one-line flip plus updating those expectations.

clubanderson:
> [!NOTE]
> **The sequencing gate this PR was blocked on is now satisfied — see the gate-verification comment below.**
> #7486 merged at `2026-09-18T00:19:14Z`, and the projectbluefin spoke's audit trail carries 24 `agent_pr_reviewed` entries attributed to `agent=reviewer` since that merge, which can only originate from the audited relay. The reviewer therefore has a working, audited route before this PR takes the direct one away. The `converse: true` precondition is answered by the same evidence. The original blocking notice is preserved in the edit history.

Closes #7487

## What this changes

#7469 gave the `reviewer` token tier `pull_requests: write` so an ADVISORY reviewer could say the verdict it had just computed, rather than writing it to a merge-eligibility file that nobody reads on a spoke which never runs a merge sweep. That was a real problem, but it was solved by the wrong mechanism. Hive already had a sanctioned route to the same outcome, and this PR restores the tier to `pull_requests: read` so that route is the only one.

The rationale is the one the issue gives, and it holds up against the code. Everywhere else in hive the governor performs the write, not the agent: `issue_request_watcher.go`, `pr_request_watcher.go` and `merge_request_watcher.go` all take a request file from an agent and execute it with the App token. `review_request_watcher.go` is the review analogue. Routing a review through it is better than a direct `gh pr review` on three counts that the direct path cannot recover: the review is authored by the App bot rather than by the agent's own identity, it is recorded on the audit/activity trail so it counts toward activity and SLA accounting, and its body passes through `scanCanaryText` under a fail-closed contract. Leaving both routes open would mean the easier one wins, and the easier one is the one hive cannot see.

The `reviewer` tier itself is kept — it is still the right tier, and `contents: read` / `metadata: read` are correct. `pkg/agentmode.TokenTierForRole` and its tests are untouched; the role plumbing is independent of the permission level, and nothing here changes which agents resolve to which tier.

## What I verified against the actual code, versus what the issue asserted

Every load-bearing claim in the issue checks out against `v4`, with one correction. The `reviewer` case in `src/pkg/github/app.go` did in fact grant `PullRequests: gh.Ptr("write")` (the issue's line references were stale, so I re-derived them — the case now sits around line 324 and the permission literal around line 358 before this change). `src/pkg/github/review_request_watcher.go` exists, and the doc comment on `ReviewRequestDir` says verbatim what the issue quotes, including the "INSTEAD of running `gh pr review` from its own shell" clause and the explicit statement that the relayed review is recorded on the audit/activity trail unlike a direct agent-CLI review the hive never observes. `scanCanaryText` is genuinely applied to `req.Body` in that watcher, at two call sites, both gated on `c.canaryFailClosed`.

The correction: the issue says the existing assertions in `src/pkg/github/reviewer_tier_permissions_test.go` "pin `pull_requests` presence". They did not — they already pinned the value, asserting `perms["pull_requests"] != "write"` as a failure. That is a better starting point than described, but it means those assertions actively contradicted the new behaviour and had to be inverted rather than tightened. They now pin `read`, again by value.

## Tests

`TestScopedToken_ReviewerTierCanWritePullRequests` is replaced by `TestScopedToken_ReviewerTierPullRequestsIsReadOnly`, which pins `pull_requests == "read"` and also pins `contents` and `metadata` at `read`. `TestScopedToken_ReviewerTierGainsNothingElse` gains an explicit assertion that `pull_requests` is not `write`. A new `TestScopedToken_ReviewerTierMatchesAdvisorPermissions` asserts the reviewer tier is permission-identical to the advisor tier across `pull_requests`, `contents`, `metadata`, `issues`, `workflows`, `administration` and `checks` — that is the assertion that would catch a future re-widening of *any* permission on this tier, not just this one. `TestScopedToken_AdvisorTierStillReadOnly` is unchanged and still passes.

I mutation-tested the result: flipping the production literal back to `gh.Ptr("write")` produced three genuine `--- FAIL` results (not build errors) across all three reviewer tests; reverting the mutation returned the package to green. So these assertions have teeth, and specifically they have teeth on the *value*, which a presence-only assertion would not.

`go build ./...` passes, and `go test ./pkg/github/ ./pkg/agentmode/ ./pkg/agent/ ./pkg/proxy/` is green. `gofmt -l` reports nothing on either changed file.

## The `converse: true` precondition — please read this

The issue asks for confirmation that the reviewer's config carries `converse: true`, because the relay route only opens for an `ADVISORY` agent that has it. **I could not confirm this, and it is not confirmable from this repository.** `converse` is a per-agent capability set in the operator-owned `hive.yaml` on each spoke (`agent.AgentCapabilities.Converse`, parsed in `pkg/agent/capabilities.go`); there is no shipped reviewer agent config in this repo to carry it. `src/deploy/hive.yaml` defines no agents, and a repo-wide search for `converse` in YAML/JSON returns nothing — it appears only in Go code, tests, and documentation (`src/docs/agent-configuration.md` documents it as something "an operator writes").

So the confirmation the issue asks for has to happen on the projectbluefin spoke's own config, not here. I deliberately did **not** try to set it in this PR. Adding a `converse: true` default to a shipped config would be a behaviour change to agent capabilities — a widening, applied to whoever adopts it — and it belongs nowhere near a PR whose entire purpose is to *narrow* a grant; mixing the two would make this change impossible to reason about and impossible to revert cleanly. If a spoke's reviewer turns out to lack `converse: true`, the fix is an operator config change on that spoke (or a separate, separately-reviewed PR if some shipped default genuinely should carry it), and it must happen *before* this PR lands, not as part of it.

## Other places asserting the old grant

I searched the repo, including `src/docs/`, for anything else pinning `pull_requests: write` for the reviewer. The only other occurrence is the released-history entry in `CHANGELOG.md` describing #7469. I left it alone on purpose: it is an accurate record of what shipped at the time, and rewriting released changelog history to match current behaviour would be wrong. The new fragment supersedes it. No documentation under `src/docs/` describes the reviewer tier's permission level, so nothing there needed updating, and no other test asserts it.

## Changelog

`changelog.d/security-7487-reviewer-token-read-only.md`, under `security` rather than `changed`. This removes a capability from a minted credential and closes a path that bypassed the fail-closed canary/exfiltration scan — the security category describes that more honestly than `changed` does, even though the change is also, trivially, a behaviour change.

## What a reviewer should scrutinise

The sequencing gate is the thing to scrutinise, and it is not satisfied yet: #7486 is open. Beyond that, the question worth asking is whether the relay actually covers every use the `write` grant was serving. #7469's comment justified the grant partly on the reviewer's ability to **request reviewers** to route a PR to the right human, and the relay's `ReviewRequest` struct supports `approve`, `request_changes`, `comment` and `resolve_thread` — it has no field for requesting a reviewer. If that routing capability is in live use anywhere, this PR removes it, and that gap should be closed (or consciously accepted) before the gate is lifted. I found no caller in this repo depending on it, but I cannot see what runs on the spokes.


clubanderson:
## Sequencing gate: satisfied

This PR was opened as a draft behind a gate with two preconditions. Both now hold, verified against the live **projectbluefin** spoke (`hive-hosted-hosted-projectbluefin-knuckle-gjvq`).

**1. #7486 has landed.** It merged at `2026-09-18T00:19:14Z`. The enabling change the gate named as missing is no longer missing.

**2. A reviewer has posted reviews through the audited relay, and they are on the audit trail.** `/data/audit.jsonl` on the spoke carries **109** `agent_pr_reviewed` entries, **24 of them after #7486 merged**. All 24 record `agent=reviewer`, which is the discriminator that matters: `AuditActionPRReviewed` is emitted from exactly two places, and the governor's own path (`client.go:1255`) hard-codes `AttributionAgentGovernor`. An entry attributed to `reviewer` can only have come from `review_request_watcher.go:329` — the relay. Sample, verbatim:

```
{"ts":"2026-09-18T05:51:48Z","user":"system","action":"agent_pr_reviewed",
 "detail":"repo=projectbluefin/common, number=932, state=commented, agent=reviewer, ...","agent":"reviewer"}
```

Spread across `common`, `utah`, `utah-packages`, `bluefin-lts` and `fsdk-containers`, so this is the reviewer's normal working route and not a single lucky call.

That also resolves the **`converse: true`** question the PR body could not answer from the repository. #7486 opened the relay *only* to converse-capable ADVISORY agents; 24 relayed reviews are proof the spoke's reviewer satisfies that predicate, which is stronger evidence than reading the config would have been.

## Narrowing cannot break the relay

Worth stating explicitly, because it is the obvious fear: the relay submits via `c.client.PullRequests.CreateReview` (`review_request_watcher.go:287`) — the governor's **App installation** client. The tier this PR narrows mints the token the *agent's own shell* carries. They are different credentials, so taking `pull_requests` back to `read` removes the direct `gh pr review` path and leaves the relay untouched. That is precisely the intended outcome: close the route the hive cannot see, keep the one it audits.

## The "request reviewers" concern

The PR body correctly flagged that #7469 justified the grant partly on requesting reviewers, and that `ReviewRequest` has no field for it. That capability is obsolete rather than lost. Routing to a human is now done by applying the configured human-decision label (on this spoke, `3-human-queue`), not by assigning a reviewer — and per the spoke operator that label is a bluefin-specific convention carried as config, which is why nothing in this repo requests reviewers and no caller depends on it. The paragraph documenting it should go with the grant.

Marking ready for review.

PR #7500, opened 2026-09-18
✨ feat(review): let review-swarm reviewers publish their verdict on the PR

### LINKED ISSUE #7469: ✨ The reviewer cannot reduce a PR queue: no write path, no cross-PR view, and its verdicts only gate a merge sweep unused spokes never run

clubanderson:
Closes #7469.

## The problem, and why the template fix didn't work

The reviewer on the projectbluefin spoke does genuinely good work. On `common#971` it cloned the repo at the PR head, checked `system_files/nvidia/` against the flatpak sync script, then *ran* `generate_skill_index.py --check` and diffed the output to see whether the hand-edited `index.json` matched what the generator produces.

Then it wrote that to a JSON verdict and threw it away.

The verdict feeds `HasAggregateApproval` → `writeMergeEligible`. With no auto-merge at ACMM L5, **nothing consumes it**.

The obvious fix — rewrite the agent's `kick_template` — does nothing, because the reviewer's kick is not built from a template. It is built in Go by `BuildPerspectivePrompt` (`pkg/review/prompts.go`), which constructs the whole message and never mentions publishing. I confirmed this on the live spoke: after pointing `kick_template` at a new policy file, the kicks were still `[review-perspective:correctness] Review PR …`, unchanged.

## The fix

Add `review.post_comments`. When set, the kick gains a publish step:

```
PUBLISH YOUR VERDICT.
After you produce the JSON, post your findings as a PR comment so a human sees them:
  hive-review 971 --repo projectbluefin/common --comment --body-file <file>
Use hive-review, never `gh pr review` — it is submitted with the App token and recorded on the audit trail.
Only --comment. Do NOT approve, request changes, merge, close, or label; a human decides those.
```

`hive-review` is the existing shim whose own header says agents should call it *instead of* `gh pr review`, because the relay submits with the App installation token and is audited as `agent_pr_reviewed`. So this adds no new mechanism — it tells the agent to use the one already installed.

## Design choices worth calling out

**Off by default.** The zero value reproduces the existing prompt exactly. `TestPromptPublishIsAdditive` asserts the opted-in prompt is a *strict superset* of the default, so the JSON aggregate that feeds the merge gate can never be traded away for a comment.

**Comment-only.** `approve`/`request_changes` set a review state that influences merge; that stays with a human. `TestPromptPublishWithholdsMergeAuthority` scans the publish block and fails if `--approve`, `--request-changes` or `--resolve-thread` ever appear in it.

**It licenses silence.** This is the part I'd most like reviewed. The measured grounding experiment found the weakest arm produced confident, uncitable findings. A false positive used to land in a file nobody read; it now lands on a contributor's PR and costs them time to refute. So the prompt requires `file:line` for every claim and explicitly tells the agent to skip the comment rather than pad it. At a 423-PR queue, a prompt that does not permit silence produces noise at scale.

## Verification

- `pkg/review` suite green; `go vet` clean on `pkg/review`, `pkg/config`, `cmd/hive`.
- Six new tests: default-omits, additive-superset, exact relay command, merge-authority withheld, noise discouraged, and all five perspectives covered.
- `pkg/config` has one failing test locally (`TestEntrypointHardensRuntimeConfigItRecreates`) — **pre-existing**, reproduced on clean `origin/v4` in a worktree. It is a macOS `chown` limitation, not related to this change.

## Rollout

Needs `review.post_comments: true` plus a reviewer that can reach the relay. Note the relay is currently gated on `AuthorizePROpen` (`CanPush()`), which #7486 fixes for `converse`-capable advisory agents.


clubanderson:
Flagging an overlap rather than a conflict: I've opened #7502 against the same issue, reaching the same goal from the other direction, and I don't think both should land as-is.

#7500 has the reviewer publish its own verdict through the `hive-review` relay. #7502 has the **spoke** publish on the reviewer's behalf — `Client.PublishReviewVerdict` renders `review-verdicts.json` into a PR comment from the governor's eval cycle, with the agent's prompt and `reviewer-advisory.md` untouched.

I went that way because #7469 is emphatic on the point ("Handing the reviewer a write token to post comments would dismantle the property its whole design rests on") and the advisory template currently reads "Do NOT attempt to merge, approve, label, or **comment** on GitHub." This PR is careful — comment-only, audited relay, App token rather than an agent token, and `TestPromptPublishWithholdsMergeAuthority` is a good guard — so it hands the agent no new credential. But it does move the agent from *cannot comment* to *is instructed to comment*, and the advisory template would need to change to match, which the issue asks not to do.

Two things I checked in the code that may be useful either way:

`refreshReviewVerdicts` (`cmd/hive/main.go`) is gated on `cfg.Review.RequireApproval`. On the spoke this issue is about — no auto-merge, so no merge gate, so no `require_approval` — the verdict artifact is never re-collected. That doesn't affect #7500, since the agent writes its own comment from its own analysis, but it does mean `review-verdicts.json` on projectbluefin is likely stale or absent, which is worth knowing before reasoning about what the merge gate currently sees.

Also, the issue's `cmd/hive/main.go:9147` citation for the `HasAggregateApproval` gate is stale; on current `v4` it is `:9089`, in `classifyMergeEligibility`. Same code, drifted line.

The genuine advantage of your approach, which you name yourself and which I can't match: an agent composing its own comment can stay silent when it has nothing citable. A deterministic renderer publishes whatever the artifact contains. At a 423-PR queue that asymmetry may well dominate, and if the maintainers judge it does, #7502 is the one that should be closed.

PR #7502, opened 2026-09-18
✨ feature(review): publish reviewer verdicts as PR comments, opt-in per spoke

clubanderson:
Refs #7469.

This implements capability **(A)**, the verdict publisher, on the spoke side — and deliberately not capability (B). It also lands alongside #7500, which reaches for the same goal from the other direction; the relationship between the two is spelled out at the bottom, and I think it is the most interesting thing to review here.

## What I verified against the code before building anything

The issue's diagnosis holds up, with one line number correction. `Aggregate` (`src/pkg/review/review.go`) really does carry everything a human-readable comment needs: `Findings []PerspectiveFinding`, each wrapping an `outputschema.Finding` with `File` and `Line`, plus `Perspectives map[Perspective]Verdict`, the aggregate `Verdict`, and `Reasons`. So a renderer built on the artifact is not speculative — the data is already there and already written by `review.CollectAndWrite`.

The consumer set is as narrow as claimed. `HasAggregateApproval` has exactly one caller, `classifyMergeEligibility` inside `writeMergeEligible`, and the gate is `cmd/hive/main.go:9089` on current `v4` rather than `:9147` — the shape of the code is identical to the issue's quote, only the line has drifted. Nothing else reads a verdict. `reviewer-advisory.md` says what the issue says it says, and I have not touched it.

One thing the issue does not mention but which turned out to matter: `refreshReviewVerdicts` was itself gated on `cfg.Review.RequireApproval`. So on exactly the spoke the issue is about — no auto-merge, so no merge gate, so no `require_approval` — the artifact would never have been re-collected at all, and a publisher hung off it would have had nothing current to publish. Widening that gate is a small change but it is load-bearing, and `TestRefreshReviewVerdicts_PublishVerdictsAloneRefreshesArtifact` pins it.

## What this does

`review.publish_verdicts`, off by default, makes the hive post each current verdict as a comment on the PR it judged: the findings with their `file:line` citations, the per-perspective breakdown as a table, the aggregate verdict, and the reasons the aggregation recorded. It follows the existing opt-in lane pattern exactly — a field on `ReviewConfig`, a pointer field on the owner-only `/api/config/review` handler, and a toggle in the Review Gate block of `renderGovFeatures()`. No new configuration mechanism.

Rendering lives in `pkg/review/publish.go` and does no I/O. Publishing lives in `pkg/github/review_verdict.go` and knows nothing about aggregation. The `cmd/hive` glue is thirty lines.

**The reviewer gains nothing.** The write is `Client.PublishReviewVerdict`, called from the governor's eval cycle with the App token the spoke already holds. Nothing in that path is reachable from the reviewer agent's capabilities; it still emits `review-verdicts.json` and nothing else. This is the whole reason the issue points at the spoke rather than at the agent, and it is why `reviewer-advisory.md` needed no edit.

**Idempotency is structural rather than remembered.** The rendered body opens with `<!-- hive:review-verdict -->` and carries a second marker naming the head commit. The renderer is deterministic — sorted perspectives, no timestamp — so re-publishing an unchanged verdict at an unchanged head produces a byte-identical body and the publisher performs no forge write at all. A re-review at a new head renders a different body and edits the same comment in place. I reused the `ensureSweepComment` / `ensureLevelHoldNotice` marker idiom rather than inventing one; I did look at `findDigestCommentDetail` first, but it is bound to the advisory digest's own prefix and its newest-first paginated scan solves a problem (thousands of comments on a pinned issue) that a PR thread does not have.

**Stale verdicts are never published.** `SelectPublishable` drops any aggregate whose head SHA is not the PR's current head, and any that carries no head SHA at all. A verdict for a superseded commit presented as current is worse than no comment, because the reader cannot tell.

**Someone else's comment is never overwritten.** When the App bot login is known, only the bot's own marker comments are adopted. The body also goes through `advisory.NeutralizeMentions`, the canary scan and `logscrub` before it is written, because it is agent-sourced prose on the same footing as an advisory digest — a finding quoting `@someone` would otherwise re-notify that person on every re-render.

## What I did not build, and why

No duplicate sweep. The issue is candid that (B) is blocked on #7456 and #7467, and that file-set clustering is a candidate generator rather than a verdict, so its output has to be a suggestion a human acts on. That is a second feature with its own failure modes, and half of it would be worse than none of it. `Refs`, not `Closes`.

I also cannot exercise this against a live spoke. Every idempotency claim above is proven against an in-memory fake forge that counts creates and edits separately, not against GitHub. The first real-world thing to watch is whether the App bot login is populated on a hosted spoke: if it is empty the publisher falls back to a marker-only match, which is the same fallback `findDigestCommentDetail` makes for a PAT client, but it is a weaker guarantee than the authorship check.

## Tests, and evidence they have teeth

I mutated the production code once per behaviour and confirmed a genuine `--- FAIL` rather than a build error each time. Dropping the `publish_verdicts` gate fails the off-by-default test; making the marker unmatchable fails three idempotency tests; removing the unchanged-body short circuit fails the "second publish adds nothing" tests in both packages; deleting the `file:line` line from the renderer fails the citation tests in both packages; neutering the head-SHA comparison fails the staleness tests; narrowing the refresh gate back to `require_approval` fails the collection test; removing the authorship check lets the fake human comment be clobbered and fails; removing mention neutralisation fails; and dropping the handler's `publish_verdicts` branch fails the config round-trip test. One earlier attempt at the marker mutation produced an unused-import compile error, which I discarded as not a valid mutant and replaced.

`go build ./...` is clean, `gofmt -l` reports nothing on any file I touched, and `pkg/review`, `pkg/github`, `pkg/dashboard` and `cmd/hive` are green. `pkg/config` fails only `TestEntrypointHardensRuntimeConfigItRecreates`, the known macOS chown limitation. The api-reference citation checker and the inline-JS checker both pass.

## Its relationship to #7500

#7500 reaches the same goal by telling the reviewer agent to publish its own verdict through the `hive-review` relay. This PR has the spoke publish on the reviewer's behalf and leaves the agent's prompt untouched. They are genuine alternatives, and I do not think both should land as-is.

I built this one because #7469 is emphatic on the point: "Handing the reviewer a write token to post comments would dismantle the property its whole design rests on," and `reviewer-advisory.md` says "Do NOT attempt to merge, approve, label, or comment on GitHub." #7500 is careful — comment-only, via an audited relay that already holds the App token, with a test that fails if `--approve` ever appears in the block — so it does not hand the agent a token. But it does move the agent from "cannot comment" to "is instructed to comment," and the advisory template would have to change to match.

The trade runs the other way too, and #7500's author names it himself: an agent composing its own comment can exercise judgement and stay silent when it has nothing citable, whereas a deterministic renderer publishes whatever the artifact contains. That is a real advantage at a 423-PR queue, and it is the strongest argument against the approach in this PR. My answer is narrower than his: publish only for the current head, cite every finding that carries a location, and say plainly in the comment that it is advisory. It does not solve the noise problem, and if the maintainers judge that noise is the dominant risk then #7500 is the better patch and this one should be closed.

The two are also not strictly exclusive. The renderer and `SelectPublishable` here are pure functions over the artifact and would survive intact if the publishing trigger moved.

## Worth scrutinising

The App-token boundary, first. `PublishReviewVerdict` is a new write path on the forge client and should be read as one: it writes only issue comments, only to a PR the artifact already names, only under the opt-in flag, and it audits every create and edit as `review_verdict_published` while deliberately not auditing unchanged cycles. Second, `truncateDigest` is now a thin wrapper over a generalised `truncateComment` so the verdict path can reuse the UTF-8-safe line-boundary cut; the advisory footer text is unchanged, but it is a shared-function refactor on a hot path and deserves a look.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


clubanderson:
Closing in favour of **#7500**, which merged at 01:06Z while this was still in CI.

Both PRs answer #7469's capability (A) — get the reviewer's verdict in front of a human — and this PR's own description says it plainly: *"They are genuine alternatives, and I do not think both should land as-is."* #7500 got there first and is now on `v4`, so landing this too would mean two publish paths and two config knobs (`post_comments` vs `publish_verdicts`) for one outcome. The `ReviewConfig` conflict when I merged `v4` in is exactly that collision made concrete.

**The design concern this PR raised is still worth recording, so I've left it on the issue rather than deleting it with the branch.** #7469 says "Handing the reviewer a write token to post comments would dismantle the property its whole design rests on," and `reviewer-advisory.md` says "Do NOT attempt to merge, approve, label, or comment on GitHub." #7500 does not hand the agent a token — it routes through the audited `hive-review` relay and has a test that fails if `--approve` ever appears — but it does move the reviewer from *cannot comment* to *is instructed to comment*, and the advisory template now has to stay consistent with that. Worth a maintainer's eye on the next advisory-template edit.

Nothing here is lost that can't be recovered: the branch `feat/7469-verdict-publisher` stays, and the renderer plus `SelectPublishable` are pure functions over the artifact that would survive intact if the publishing trigger ever needs to move back to the spoke. Reopen if maintainers prefer the spoke-side approach after all.

#7469 stays open — capability (B), the duplicate sweep, is unbuilt and still blocked on #7456/#7467.

PR #7507, opened 2026-09-18
🐛 fix(agents): keep an operator's mode and converse on agents the ACMM pack does not manage

### LINKED ISSUE #7503: 🐛 Agent `mode: ADVISORY` and `converse: true` silently dropped — reviewer runs at the ACMM default with more authority than configured
## Summary

On the `hosted-projectbluefin-knuckle-gjvq` spoke, the `reviewer` agent is configured `mode: ADVISORY` in **every** config layer, but runs as **`ISSUES_AND_PRS`** — two rungs up the trust ladder. Its `converse: true` is dropped in the same way: the capability file is written empty.

The agent therefore holds push and PR-open authority that no config file grants it, and lacks the one capability that was deliberately granted.

## Evidence

All three layers agree on `ADVISORY`:

```
/data/hive.yaml.dashboard -> mode= ADVISORY converse= None kick= reviewer-advisory.md
/data/hive.yaml.runtime   -> mode= ADVISORY converse= None kick= reviewer-advisory.md
/etc/hive/hive.yaml       -> mode= ADVISORY converse= None kick= reviewer-advisory.md
```

The per-agent overlay agrees, and is a valid file with top-level keys (verified with `cat -et`, no nesting or trailing whitespace):

```
$ sed -n '31,34p' /data/agent-configs/reviewer.yaml
mode: ADVISORY
on_demand: false
caveman_mode: ""
converse: true
```

The runtime disagrees:

```
$ cat /tmp/.hive-mode-reviewer   ->  ISSUES_AND_PRS
$ cat /tmp/.hive-caps-reviewer   ->  (empty, 0 bytes)
```

`ISSUES_AND_PRS` is exactly `DefaultAgentMode(*, 5)` (`pkg/agent/manager.go`, `case 5: return ModeIssuesAndPRs`), i.e. the value taken when `agent.Config.Mode == ""`. An empty caps file is `AgentCapabilities{}.String()`, i.e. `Config.Converse == nil`. **Both are the "config said nothing" branch**, so the Manager's in-memory `AgentConfig` for `reviewer` has neither field set, while every file on disk sets both.

## Ruled out

- **Not a parse failure.** `ParseAgentMode("ADVISORY")` matches `modeNames[0]` exactly. The tag is `yaml:"mode"` (`config.go:997`), `yaml:"converse,omitempty"` (`config.go:1011`).
- **Not overlay rejection.** `ValidateKickTemplateName("reviewer-queue.md")` passes (bare name, no `/` or `..`). No `rejected invalid per-agent overlay file` line appears in 24h of logs.
- **Not a tombstone skip.** The boot tombstone list is `[guide brainstorm telemetry strategist operations]`; `reviewer` is absent, and no `merge: skipped tombstoned agent` line was logged.
- **Not write-ordering.** `reviewer.yaml` mtime `00:08:06Z`, process start `00:09:17Z`, state files `00:12Z` — the file predates boot.
- **Not staleness.** Both state files were written *after* this boot.
- **Not `applyKnownAgentDefaults`.** It sets emoji/color/aliases/sort order only; it never touches `Mode`.
- **Not the ACMM pack.** `ACMM pack applied` reports `created: 0, updated: 0` with `skipped_agents: [supervisor scanner ci-maintainer quality sec-check architect]` — `reviewer` is not pack-managed.

The other five agents cannot distinguish the two cases: their configured mode is `ISSUES_AND_PRS`, which is *also* the L5 default, so they look correct whether or not their config is being read. `reviewer` is the only agent on this spoke whose configured mode differs from the level default — which is likely why this has gone unnoticed.

## A probable contributor

There are **two** review agents in the runtime:

```
all: [architect ci-maintainer quality review reviewer scanner sec-check supervisor]
```

`/data/agent-configs/review.yaml` (`id: review`, `mode: ISSUES_AND_PRS`, `kick_template: review.md`, mtime Sep 17 18:16) sits alongside `reviewer.yaml`. Only `reviewer` has state files, so `review` appears to be an orphan — but two agents whose names are prefixes of one another, one carrying exactly the mode that is leaking, is worth ruling in or out before looking further afield.

## Why it matters

1. **Authority not granted.** ADVISORY was chosen deliberately: this agent reviews and comments, it does not push or open PRs. It currently can do both.
2. **It silently inverts #4492.** `converse` is the capability that lets an advisory agent talk. Configured on, it is delivered off, so the agent is simultaneously over-privileged on the ladder and under-privileged on the axis meant to replace it.
3. **It fails silently.** `agentMode` logs `SyncModeFiles: Config.Mode override` only when the override *is* applied. The dropped case emits nothing, so the only signal is reading `/tmp/.hive-mode-<agent>` and comparing by hand.

## Suggested fix

Two parts, the second useful regardless of the first:

1. Find where the `reviewer` `AgentConfig` loses `Mode`/`Converse` between load and `Manager`.
2. **Log the fallback.** In `agentMode`, when `Config.Mode` is empty *and* the agent has a per-agent overlay file, log at INFO that the default is being used. A one-line "config said nothing, using the level default" would have made this a five-minute diagnosis. The same applies to the capability write.

A regression test should assert that an agent configured below its level default actually *runs* below it — the current tests pin the parse and the default table separately, and this bug lives in the gap between them.

Danathar:
## Summary

An agent configured `mode: ADVISORY` and `converse: true` was running as `ISSUES_AND_PRS` with an empty capability file — two rungs more authority than any config file granted it, and without the one capability that was deliberately granted. This PR fixes the two separate paths that dropped those fields, adds the fallback log the issue asked for, and pins all of it with regression tests that fail on `v4`.

Fixes #7503

## What was actually going wrong

The issue's diagnosis was right: both symptoms are the "config said nothing" branch. Two different pieces of code produced that state.

**`mode` — `ClearAllModeOverrides` in the pack apply.** `applyPack` (which runs on every restart — the `ACMM pack applied created: 0, updated: 0` line in the issue is this firing) called `AgentMgr.ClearAllModeOverrides()`, which wiped `Config.Mode` for *every* agent in the process table, not only the pack's roster. `SyncModeFiles` then wrote `DefaultAgentMode("reviewer", 5)` = `ISSUES_AND_PRS`. `PUT /api/packs/level` did the same, and additionally blanked `Mode` for every agent in the *persisted* config, so the loss survived reloads. The clear exists so a previous level's pack-seeded mode does not stick — but `reviewer` is in no pack; its mode was never pack-seeded, so what got cleared was the operator's setting.

**`converse` — `LoadWithDashboardOverlay`.** In Kubernetes the load re-applies the dashboard overlay's agents on top of the seed + per-agent overlays, and `MergeAgentOverrides` replaces each entry wholesale. The dashboard overlay's `reviewer` entry was silent on `converse` (nil, not `false`), so the per-agent file's `converse: true` — the documented layer for agent fields, and per `layers.go` the one that outranks the dashboard overlay — was discarded on every boot and every reload. `Converse` is a pointer precisely so "unset" and "explicitly false" stay distinguishable across this overlay; the merge was not honouring that distinction.

The orphan `review.yaml` the issue flagged as a probable contributor is not involved: neither path keys on agent name prefixes.

## Changes

- **`pkg/agent`**: `ClearAllModeOverrides()` → `ClearModeOverrides(names []string)`. `SyncModeFiles` now logs `SyncModeFiles: no Config.Mode, using level default` (with `default`, `converse_configured`, `overlay_file`) alongside the existing override line, so an agent running at the default because its config said nothing is a grep rather than a by-hand read of `/tmp/.hive-mode-<agent>`.
- **`pkg/config`**: `ACMMPack.AgentNames()` and `ACMMPackManagedAgentNames()` (union of every level's roster, sorted). `LoadWithDashboardOverlay` carries a lower layer's `converse` forward when — and only when — the dashboard overlay entry is silent on it; an explicit `converse: false` still revokes.
- **`pkg/dashboard`**: both `applyPack` and `handlePackSetLevel` scope the mode clear (process table *and* persisted config) to `ACMMPackManagedAgentNames()`.

### Why the union of all packs, not the target level's roster

On a downgrade (L5 → L2) `strategist`/`architect` are not in the L2 pack but stay in the config carrying the L5 pack's `ISSUES_AND_PRS` (only `telemetry`/`operations` are gate-removed). That is a stale pack mode and must still be cleared, or an operator resume would run them at L5 authority on an L2 hive. Scoping to the union keeps behaviour byte-identical for every agent any pack has ever listed; only agents no pack manages are exempt. `TestSetLevelDowngradeStillClearsHigherPackAgentMode` pins this.

### What this deliberately does not do

It does not make the per-agent overlay file win over the dashboard overlay for *all* fields, even though `layers.go` documents that precedence. `applyPack`'s update path persists reconciled pack fields to the dashboard overlay but not to the per-agent file, so flipping precedence wholesale would let a stale `scanner.yaml` from a lower level revert pack reconciliation on every reload across the fleet. That is a separate, fleet-affecting change; `converse` is safe to fix on its own because no pack seeds it and the dashboard API writes it to both layers.

## Tests

All fail on the parent commit and pass with the fix, per CONTRIBUTING's test policy:

- `pkg/dashboard`: `TestApplyPackKeepsOperatorModeOnNonPackAgent`, `TestSetLevelKeepsOperatorModeOnNonPackAgent` — assert the **effective** mode via `Manager.AgentCapabilities`, the exact gate `AuthorizePROpen` enforces, for a non-pack `reviewer` configured `ADVISORY` at L5 (and that pack agent `scanner` is still re-derived). `TestSetLevelDowngradeStillClearsHigherPackAgentMode` pins the union scoping.
- `pkg/config`: `TestLoadWithDashboardOverlay_SilentOverlayKeepsAgentOverlayConverse` (fails on parent: `got <nil>`), `TestLoadWithDashboardOverlay_ExplicitFalseStillRevokesConverse`, `TestACMMPackAgentNames`, `TestACMMPackManagedAgentNames`.
- `pkg/agent`: the two `ClearAllModeOverrides` unit tests converted to the scoped method; one now also asserts an unnamed agent's mode is untouched.

```
go build ./...                                      ok
go vet ./pkg/agent ./pkg/config ./pkg/dashboard     ok
go test ./pkg/dashboard                             ok  (48.98s)
go test ./pkg/config -run 'TestACMMPack|TestLoadWithDashboardOverlay'   ok
go test ./pkg/agent  -run 'TestClearModeOverrides|TestSyncModeFiles'    ok
```

Full `./pkg/config` and `./pkg/agent` suites: 45 tests fail in this sandbox on *both* pristine `upstream/v4` and this branch (tmux / read-only `/tmp` / ip6tables — environmental); a sorted diff of the two failure lists is identical apart from timings. No new failures.

## Deployment note for the reporter

Once this lands and rolls to the projectbluefin spoke, `reviewer`'s `/tmp/.hive-mode-reviewer` should read `ADVISORY` and `/tmp/.hive-caps-reviewer` should carry `converse` after the startup pack apply, with a `SyncModeFiles: Config.Mode override … override=ADVISORY converse_configured=true` line in the log. If the dashboard overlay is later re-saved it will then also carry `converse: true`, since the in-memory config now holds it.


clubanderson (approved):
Correct scoping: pack applies now clear only the pack roster's modes, silent overlays no longer revoke converse, and the fallback log makes the gap greppable. Regression tests pin all three. LGTM.

clubanderson:
## Independent verification of #7507 against #7503 — reproduced, mutation-tested, concur

I was asked to fix #7503 and found this PR already open, so I did not open a competing one. Instead I verified every factual claim in the issue against `v4` (`c25ab30d`) and then checked this PR's fix and tests empirically on a local checkout of `fix/7503-advisory-converse-dropped`. **Everything below is from running the code, not from reading the diff.**

### The issue's claims — all verified true on `v4`

| Claim | Verdict | Evidence on `v4` @ `c25ab30d` |
|---|---|---|
| `ISSUES_AND_PRS` is exactly `DefaultAgentMode(*, 5)` | ✅ true | `src/pkg/agent/manager.go:3833-3834` — `case 5: return ModeIssuesAndPRs` |
| Empty caps file ⇒ `Config.Converse == nil` | ✅ true | `Converse *bool` at `src/pkg/config/config.go:1011`, tag `yaml:"converse,omitempty"` — exactly as the issue quoted |
| `mode` tag is `yaml:"mode"` at `config.go:997` | ✅ true — line number is still exact | `src/pkg/config/config.go:997` |
| "Not `applyKnownAgentDefaults`" | ✅ true | confirmed, it never touches `Mode` |
| "Not the ACMM pack" (`reviewer` not pack-managed) | ✅ true *as stated*, but this is where the bug lives | `reviewer` is in no pack roster — yet the pack **apply** was what cleared its mode |
| "The orphan `review.yaml` is a probable contributor" | ❌ **false** — ruled out | neither failing path keys on agent name or prefix; I re-read `ClearAllModeOverrides` and `LoadWithDashboardOverlay` and both are name-agnostic |

The one claim I could **not** verify from this repository is anything about the live `hosted-projectbluefin-knuckle-gjvq` spoke — `converse: true` is spoke-config-only and the state files under `/tmp/.hive-*-reviewer` cannot be exercised from here. I treat the reporter's `cat` output as given and verified only the code paths that produce it.

### The PR's diagnosis — both paths confirmed present on `v4`

1. **`mode`**: `ClearAllModeOverrides` at `src/pkg/agent/manager.go:3429` loops `for _, agent := range m.agents` unconditionally, and is called from `src/pkg/dashboard/api_packs.go:405` (`applyPack`, which runs on every restart) and `:610` (`handlePackSetLevel`). `SyncModeFiles(level)` runs immediately after at `api_packs.go:407`. That is a complete chain from "operator set ADVISORY" to "`/tmp/.hive-mode-reviewer` says `ISSUES_AND_PRS`".
2. **Persisted loss**: `api_packs.go:601-604` on `v4` blanks `ac.Mode` for **every** entry in `s.deps.Config.Agents` before `saveConfig()` — so on the set-level path the loss survives reloads, exactly as the PR says.
3. **`converse`**: `LoadWithDashboardOverlay` calls `cfg.MergeAgentOverrides(overlay.Agents)` (`config.go:4407` on `v4`), and `MergeAgentOverrides` (`pkg/config/agent_overlay.go:125`) replaces each entry wholesale — a dashboard entry silent on `converse` (nil) overwrites a per-agent file's `true`.

### Mutation testing — I re-introduced each bug and confirmed real assertion failures

Every mutation below is a **behavioural** mutant (compiles cleanly); none is a build error.

| # | Mutation | Result |
|---|---|---|
| 1 | Delete the `converse` carry-forward loop in `LoadWithDashboardOverlay` (i.e. restore `v4`) | `--- FAIL: TestLoadWithDashboardOverlay_SilentOverlayKeepsAgentOverlayConverse` — `got <nil>` |
| 2 | **Negative control**: make the carry-forward *unconditional* (drop the `if oa.Converse != nil { continue }` guard) so an explicit `converse: false` is also overwritten | `--- FAIL: TestLoadWithDashboardOverlay_ExplicitFalseStillRevokesConverse` — the pair genuinely distinguishes *unset* from *explicitly false*, it is not string-matching one spelling |
| 3 | `applyPack` passes all config agent names to `ClearModeOverrides` (restore `v4` semantics) | `--- FAIL: TestApplyPackKeepsOperatorModeOnNonPackAgent` — *"reviewer configured ADVISORY can open PRs after ApplyPack(5)"* **and** `process-table Mode = "" want ADVISORY` |
| 4 | `handlePackSetLevel` blanks persisted `Mode` for all agents (restore `v4`) | `--- FAIL: TestSetLevelKeepsOperatorModeOnNonPackAgent` (`persisted reviewer Mode = ""`) **and** `--- FAIL: TestSetLevelDowngradeStillClearsHigherPackAgentMode` |

Mutation 3 is the one that matters most: it asserts the **effective** capability via `AgentMgr.AgentCapabilities("reviewer")` — the same gate `AuthorizePROpen` reads — not just a config field. That is precisely the "assert an agent configured below its level default actually *runs* below it" test the issue asked for, and it closes the gap between the parse test and the default-table test.

Clean-tree validation on the PR branch:
```
go build ./...                                                          ok
go vet ./pkg/agent/... ./pkg/config/... ./pkg/dashboard/...             ok
go test ./pkg/config  -run 'TestLoadWithDashboardOverlay_|TestACMMPack' ok
go test ./pkg/dashboard -run 'TestApplyPackKeeps|TestSetLevel...'       ok  (1.6s)
go test ./pkg/agent   -run 'TestClearModeOverrides|TestSyncModeFiles'   ok
```
I also saw the new fallback line fire in the mutant run, which is the observability half of the issue's ask:
```
INFO SyncModeFiles: no Config.Mode, using level default agent=scanner level=5 default=ISSUES_AND_PRS converse_configured=false overlay_file=""
```

### Relationship to #7486 — residual gap, not a regression

#7486 (`ccef331a`) touched `src/pkg/agent/manager.go`, `src/cmd/hive/main.go` and `src/pkg/proxy` to let converse-capable advisory agents review through the audited relay. It does **not** touch `ClearAllModeOverrides`, `api_packs.go` or `LoadWithDashboardOverlay`, and `git log -S'ClearAllModeOverrides'` shows this PR is the first change to that function. So #7503 is a **pre-existing gap, not a regression from #7486** — but it is the gap that *defeats* #7486 in practice: #7486 made converse-capable advisory agents able to review, while this bug ensured the capability was never delivered and the agent was pushed above ADVISORY anyway. Fixing this is what makes #7486 observable on a spoke.

### Two notes, neither blocking

1. The PR's stated non-goal — not flipping per-agent-file precedence over the dashboard overlay for *all* fields despite `layers.go` documenting it — is the right call and the reasoning (a stale `scanner.yaml` reverting pack reconciliation fleet-wide) holds. Worth a follow-up issue so the documented precedence and the implemented precedence are reconciled deliberately rather than by accident.
2. `seedAdvisoryReviewer` calls `t.Skip` if a pack ever starts managing `reviewer`. That is the correct fail-safe direction, but a silently skipped regression test is a regression test that stops guarding. `TestACMMPackManagedAgentNames` already asserts `!seen["reviewer"]`, so the skip can only trigger alongside that test failing loudly — fine as-is, just noting the coupling is load-bearing.

**Concur with the approval.** The diagnosis matches the code, the scoping to the union of all pack rosters is correctly justified by the downgrade case, and the tests have teeth under four independent mutations including a negative control. No competing PR opened.

Issue #7232, opened 2026-09-16
🌱 cmd/hive: extract runEvalCycle and main() seams (follow-up to #7224)


clubanderson:
Follow-up to #7224, which is now closed by #7230. That PR took only recommendation **#1** (`runHub`) — the "smallest win" the issue itself named. Recommendations **#2 (`runEvalCycle`)** and **#3 (`main`)** were deliberately left out so the change stayed reviewable, and they are currently untracked. This issue tracks them.

## Measured state on `v4` (post-#7230)

Numbers below are measured, not inherited — #7224 quoted 44.1% for this package when the real figure was 51.2%, so please re-measure rather than quoting this issue later.

```
go test ./cmd/hive/ -count=1 -coverprofile=cov.out
go tool cover -func=cov.out
```

Package `cmd/hive`: **51.9%** (2080/4007 statements).

Exactly **three** functions remain at 0%:

| function | location | lines | statements | coverage |
|---|---|---|---|---|
| `main` | `main.go:1272-5423` | 4,151 | 1,491 | 0.0% |
| `runEvalCycle` | `main.go:6299-7190` | 891 | 315 | 0.0% |
| `runHub` | `main.go:9808` | 3 | 3 | 0.0% (delegating wrapper — expected, ignore) |

Those two functions are **1,806 of main.go's 1,900 uncovered statements — 95%**. Everything else in the package is already covered.

Ceiling, if both were fully exercised:

- `runEvalCycle` alone: 51.9% → **59.8%**
- both: 51.9% → **97.0%**

This is why #7230 moved the package only +0.7pp despite taking its target from 0% to 83.7%. Nothing else in `cmd/hive` will move the number; these two functions *are* the remaining gap.

## The pattern is proven — three times now

Each of these extracted a blocking/uncancellable effect behind an injectable seam, leaving a thin production wrapper:

- `StartPermissionsWatcher` → `runPermissionsWatcher(ctx, logger, interval)` — `pkg/agent/permissions_watcher.go`
- `watchHubRollout` → `watchHubRolloutWithInterval(...)` — `pkg/hub/saas.go` (#7220)
- `runHub` → `runHubWithDeps(ctx, logger, configPath, hubDeps{...})` — `cmd/hive/hub_boot_seams.go` (#7230)

[...]

PR #7442, opened 2026-09-17
🌱 refactor(dashboard): split api_contribute.go god file — profiles, admin, federation, leaderboard (#7435)

clubanderson:
## What

**Move-only refactor.** Stages 2–4 of the #7435 god-file split of `src/pkg/dashboard/api_contribute.go` (9,407 lines, 81 funcs, 6+ domains). Four unrelated concerns are cut **verbatim** into new files in the **same package** (`package dashboard`), so there are **no exported-symbol changes, no signature changes, no import churn elsewhere and no call-site edits anywhere in the repo**.

Nothing was renamed, re-signatured, re-ordered or "improved" while moving. No drive-by fixes.

| New file | Concern | Lines |
|---|---|---|
| `src/pkg/dashboard/contribute_profiles.go` | Contributor profile store (`loadContributorProfile`, `saveContributorProfileCAS`, `listContributorProfiles`, `createContributorProfile`, `findContributor`, the `ContributorProfile`/`ContributorRateLimits` types, …) + invite/registration tokens (`inviteSigningSecret`, `mintInviteToken`, `verifyInviteToken`, `reissueContributorToken`) | 445 |
| `src/pkg/dashboard/contribute_admin.go` | Contributor management handlers — trust, agent-role grants/assignment, revoke, requeue, delete | 444 |
| `src/pkg/dashboard/contribute_federation.go` | Federation registry + `/api/hives` handlers | 253 |
| `src/pkg/dashboard/contribute_leaderboard.go` | Contributor + agent leaderboard, trust-tier/rank presentation helpers | 262 |

`src/pkg/dashboard/api_contribute.go`: **9,407 → 8,069 lines** (−1,338).

## Relationship to #7436

#7436 (open) is **stage 1** — the 6,390-line embedded landing/dossier page. This PR deliberately does **not** touch that region; it takes the remaining concerns the issue lists. The two are logically disjoint but both delete from the same file, so they **will conflict textually** if both land. Resolution is mechanical: both sides are *pure deletions* from non-overlapping regions, so the merged result is the union of the deletions (plus the union of the import prunes). Happy to rebase this branch after #7436 merges — the split is script-generated and trivially reproducible on top of it.

Marked `Refs #7435` rather than `Closes` because stage 1 is owned by #7436; the issue should close once both have landed.

## Behaviour-preservation evidence (all commands actually run)

[...]

PR #7510, opened 2026-09-18
🌱 cmd/hive: extract five business-logic domains from main.go (9,795 → 8,072 lines)

clubanderson:
Refs #7238. **This is a stage, not the whole issue — #7238 should stay open.**

## First: the issue's headline claim is already stale

`main_helpers.go` **does not exist on `v4`**. The 4,789-line grab-bag this issue is named after was dissolved across #7241, #7245, #7246, #7253, #7291, #7293, #7294, #7296, #7306 and #7308. I re-measured rather than trusting the title:

| | issue filed | on `v4` today |
|---|---|---|
| `package main`, non-test | ~13,000 | 12,059 |
| `main_helpers.go` | 4,789 | **gone** |
| `main.go` | — | **9,795** |

So the *grab-bag* half is genuinely delivered. But the **other half of the title — "13K-line package main concentrates business logic" — still held**: `main.go` alone was 9,795 of the package's 12,059 non-test lines. That is what this PR addresses.

Scope note: I deliberately did **not** touch `main()` or `runEvalCycle` — those are #7232's territory (#7347 landed the first stage there). Everything moved here is business logic that merely *lived alongside* the entrypoint.

## What moved

Five self-contained domains, each into one file in the same `package main`:

| new file | concern | lines |
|---|---|---|
| `appkeyfile.go` | GitHub App private-key file management — where per-App-ID and per-hive key files live, how a key is written, resolved, fingerprinted, and how a missing key is described | 389 |
| `selfupgrade.go` | self-upgrade bookkeeping — the upgrade marker and last-outcome files, the retry/backoff budget, and boot-time reconciliation of the two | 212 |
| `login_scan.go` | login-required scanning — detecting a backend dropping to an interactive login prompt, and debouncing that across sightings so one noisy line cannot pause an agent | 366 |
| `merge_eligibility.go` | merge eligibility — hold gates, required-check state, bucket classification, the trusted-merger authz binding, and the merge-eligible report | 578 |
| `intent_verdicts.go` | intent verdicts — PR/issue evidence gathering, alignment summary, the advisory record, and the intent-verdicts report | 366 |

**`main.go`: 9,795 → 8,072 lines** at the time of the split. `v4` has since grown `main.go` (#7512, #7516) and been merged in twice, so the current comparison is **9,917 on `v4` → 8,150 on this branch, −1,767**.

Worth noting as evidence the split holds up under concurrent traffic: #7516 landed on `v4` *after* this branch was cut and changed `classifyMergeEligibility`, `notMergeableReason`, and added `mergeabilityUnknownReason` — all three squarely inside the merge-eligibility domain. The merge routed every one of them into `merge_eligibility.go` rather than into `main.go`, and the declaration diff below is still clean against the newest `v4`. The domain boundary is real enough that an unrelated PR's changes land in the right file on their own.

Each of these is now reviewable on its own instead of only as a slice of a 9.8K-line file. The five domains were chosen because each already had its own dedicated test files (`appkeyfile_test.go`, `upgrade_marker_test.go`, `login_scan_decision_test.go`, `merge_eligible_buckets_test.go`, `write_intent_verdicts_test.go`, …) — i.e. the package's own tests already treated them as separate concerns; only the source file disagreed.

## This is pure code motion — here is the proof

Declarations were moved **byte-identically**, carrying their doc comments. Nothing was renamed, reformatted, reordered within a declaration, or "improved". Same package, so no import path changes; only the new files' import blocks were generated (via `goimports`, safe because `main.go` was already gofmt-clean).

I verified this structurally rather than settling for "it compiles". A script parses every top-level `func` (including methods, keyed by receiver type), `type`, `const` and `var` — **including individual members of grouped `const (...)` / `var (...)` blocks** — from the original `main.go` and from the union of the six resulting files, keys them by name, and compares bodies **byte-exactly**:

```
BEFORE files: 1  declarations: 242
AFTER  files: 6  declarations: 242

LOST: 0
ADDED: 0
CHANGED: 0
DUPLICATE (before): 0
DUPLICATE (after): 0

RESULT: PURE CODE MOTION ✅
```

Re-run against the newest `v4` after both merges (baseline is v4's `main.go`, now 245 declarations after #7512 and #7516):

```
v4 main.go decls: 245   branch (6 files): 245
LOST  : []
ADDED : []
CHANGED: []
RESULT: PURE CODE MOTION ✅
```

Two independent corroborations:

1. **Cross-checked against `go/ast`** — a real Go parser reports **241 unique declaration names before and 241 after, with identical name sets** (the 242/241 difference is only my keying of one comma-grouped `var`). The regex parser was validated against the AST parser before being trusted.
2. **The `main.go` diff is `1723 deletions(-)` and `0 additions(+)`.** Nothing was edited in place — every changed line is a line that left the file.

```
$ git diff --stat -- src/cmd/hive/main.go
 src/cmd/hive/main.go | 1723 --------------------------------------------------
 1 file changed, 1723 deletions(-)
```

### A gofmt defect inherited from `v4`

The `v4` merge surfaced one thing worth flagging: `v4`'s own `main.go` is **not gofmt-clean**. #7512 landed this, with `runDuplicateSweepIfDue`'s closing brace glued onto the next declaration's doc comment and the blank line lost:

```go
	})
} // mergeEligiblePath is a var (not a const) only so tests can point
// mergeTargetEligible at a temp file; production never reassigns it.
var mergeEligiblePath = "/var/run/hive-metrics/merge-eligible.json"
```

Since `mergeEligiblePath` and its comment move to `merge_eligibility.go`, that glued line had to be separated. The declaration diff above flags this as the single byte-level difference in `runDuplicateSweepIfDue`, and it is exactly the stray comment text — the function's own body is untouched. All six files are gofmt-clean on this branch.

### One deliberate omission

This grouped block stays in `main.go`:

```go
var (
	ciFailingPath      = "/var/run/hive-metrics/ci-failing.json"
	intentVerdictsPath = "/var/run/hive-metrics/intent-verdicts.json"
)
```

Its two members belong to *different* extracted domains. Splitting the group would have meant **rewriting** a declaration rather than moving it, which would have broken the byte-identity guarantee above. Leaving two path vars behind is the cheaper price.

## Verification

- `go build ./...` — clean
- `go vet ./cmd/hive/...` — clean
- `go test ./cmd/hive/...` — `ok github.com/hivecommons/hive/cmd/hive 9.477s`
- `gofmt -l` on all six touched Go files — clean
- `src/scripts/check-api-reference-citations.sh` (without `--fix`) — exit 0

One **cross-package** source-reading guard did need updating: `TestF3AuthorizerIsWiredInMain` in `pkg/github` reads `cmd/hive/main.go` and asserts both that startup installs the trusted-merger authorizer *and* that `trustedMergerFunc` itself fails closed. The latter moved to `merge_eligibility.go`, so the guard failed on its "renamed or removed" branch. It now scans the whole `cmd/hive` package for the func body (so it survives the rest of this split) while keeping the wiring assertion pinned to `main.go`, since that one is about *where startup installs it*. Neither half is relaxed.

The three source-reading guard tests in `cmd/hive` itself (`TestNoBlankAutomaxprocsImport`, `TestMainRegistersBothPreShutdownHooks`, `TestAllVersionConsumersUseReportedVersion`) all `os.ReadFile("main.go")` and pin content that lives in `main()` / the version helpers — none of which moved — so they still pass unmodified. No test needed updating.

## Doc citation drift

`main.go` shrinking by 1,723 lines shifts every `src/cmd/hive/main.go:<line>` citation in the design docs (#7493 is the standing complaint about exactly this). Rather than guess, I **content-anchored** each one: read the cited line's exact text out of the original `main.go`, then located that same text in the new tree.

- App-key citations in `master-delivery-wrapped.md` now point at `src/cmd/hive/appkeyfile.go`.
- `trustedMergerFunc` citations in `github-mention-triggers.md` now point at `src/cmd/hive/merge_eligibility.go:51`.
- Citations whose target stayed in `main.go` had their line numbers shifted to the verified new position.

Note: several of these citations were **already drifted before this PR** (e.g. `main.go:8076` for `trustedMergerFunc` actually landed on `return escalated`). I carried those forward by content anchor and repointed the ones naming a symbol I moved; I did not silently rewrite unrelated pre-existing drift, which stays #7493's business.

## Remaining work — why #7238 stays open

`main.go` is still 8,072 lines. The remaining bulk is `main()` (~4,179 lines) and `runEvalCycle` (~902), which are tracked under #7232 so the work is not duplicated under two issue numbers. Use `Refs`, not `Closes`.


clubanderson:
Pushed a fix for the `test (rest 3/4)` failure.

`TestF3AuthorizerIsWiredInMain` (in `pkg/github/f3_trusted_merger_source_test.go`) is a **source-reading guard**: it `os.ReadFile`s `cmd/hive/main.go` and asserts `trustedMergerFunc`'s body still fails closed. This PR moves that declaration to `cmd/hive/merge_eligibility.go`, so the guard reported:

```
f3_trusted_merger_source_test.go:169: trustedMergerFunc not found — it was renamed or removed;
update this test deliberately, do not delete the case
```

which is the test doing its job — it just cannot tell code motion from deletion. Worth taking seriously: that comment exists because F3 was fixed in `160f585c` and then **silently reverted** by a sync merge, so this guard is load-bearing.

**Fix:** the func-body assertions now scan every non-test `.go` file in `cmd/hive` (new `f3ReadPackage` helper) instead of one filename, so the guard follows the declaration wherever the ongoing `cmd/hive` split puts it next. The *wiring* assertion stays pinned to `main.go` deliberately — that one is about startup installing the authorizer, and it should still fail if the call leaves startup.

**I verified the guard keeps its teeth** rather than assuming it, mutating production code three times and confirming a genuine assertion failure each time (not a build error):

| mutation | caught by |
|---|---|
| floor lowered to `config.RoleRead` | "no longer requires at least `config.RoleMerger`" |
| `SetMergerAuthorizer` call deleted from `main.go` | "does not install the trusted-merger authorizer … INERT" |
| `trustedMergerFunc` renamed away | "not found — it was renamed or removed" |

All reverted; `go test ./pkg/github/ -run TestF3` is green and `gofmt` reports nothing.

Note for the other in-flight splits: this is the same class of breakage as the F22/F25 and owner-gate guards hit by #7442 — **grep the test tree for the filename you are moving code out of** before pushing.

clubanderson:
Resolved the conflict with `v4` (merge commit, no force-push).

**Cause:** #7512 landed `duplicateSweepInterval` and `runDuplicateSweepIfDue` into `cmd/hive/main.go` while this PR was moving declarations *out* of it, so both conflict hunks were "v4 added code here / this branch emptied this region".

**Resolution:** took v4's new code wholesale, minus the two declarations this PR relocates — `trustedMergerFunc` and `mergeEligiblePath` — which now live only in `cmd/hive/merge_eligibility.go`. Taking v4's side verbatim would have reintroduced both as duplicate declarations in the same package; the compiler caught `mergeEligiblePath` (`redeclared in this block`), which is exactly why I rebuilt rather than trusting the resolve.

**Verified after resolving**, not assumed:
- `go build ./...` clean.
- I re-ran a duplicate-declaration scan across every non-test file in `cmd/hive` — **zero** names now declared in both `main.go` and any split file.
- `go test ./cmd/hive/... ./pkg/github/` both `ok` — including `TestF3AuthorizerIsWiredInMain`, the guard I repaired earlier in this PR, so the `runDuplicateSweepIfDue` arrival didn't disturb the F3 wiring assertion.
- `gofmt -l ./cmd/hive/` reports nothing.
- `api-reference citations: all pkg/<file>:<line> references verified.`

`runDuplicateSweepIfDue` is deliberately left in `main.go` rather than swept into one of this PR's new domain files — it arrived after this split was designed, and folding it in would mix a merge resolution with new scope.

PR #7508, opened 2026-09-18
🌱 hub: finish saas.go split — dashboard HTML to go:embed asset, extract pages/clusters/crypto

clubanderson:
Part-of: #7278 (leave open — one summary comment to follow before it can close).

This finishes the `saas.go` split to the pre-adoption v5 file boundaries:

- **`assets/dashboard.html` (12,728 lines)** — the embedded SaaS dashboard document leaves Go raw-string space for a real HTML file served via `go:embed` (`saas_dashboard_html.go` is byte-for-byte the v5 file). Editors, JS linters and CI can now see the frontend, and backticks are no longer banned in it.
- **`saas_pages.go`** (9 blocks, 204 lines) — page handlers
- **`saas_clusters.go`** (8 blocks, 110 lines) — cluster helpers
- **`saas_crypto.go`** (5 blocks, 64 lines) — crypto helpers

`saas.go`: **13,762 → 623 lines** (16,879 when #7278 was opened). The v5 `saas_vanity_budget.go` symbols already live in v4's `saas_provision.go` and stay there.

Test pins that source-grep the dashboard JS re-pointed at the asset; `check-api-reference-citations.sh --fix` run for the line drift.

Verified: `go build ./...`, `go vet ./pkg/hub/`, full `go test ./pkg/hub/ ./pkg/reach/ ./pkg/github/ ./pkg/dashboard/` green.

Closes #7278

PR #7513, opened 2026-09-18
🌱 Forward-merge v4 into v5 through b2c98f6bb (15 commits)

clubanderson:
Manual resolution of the conflicting `v5 Top-up` run tracked in #7463,
on the workflow's own `sync/v4-to-v5` branch.

- agent/hub/dashboard/config: v4's manager_pane_signals.go, saas_upgrade.go,
  contribute_{task_prompt,lease,cooldown,ledger,token}.go extractions merged
  at declaration level; v5-local edits carried into the moved decls
- config: #7503 converse carry-forward composed with v5's
  RejectInvalidAgentOverlays gate
- cmd/hive: #7451 SetHubPushedDashboardURL hooks and #7482
  applyModeUnscheduledAlert ported into v5's main()
- dashboard: contribute_landing keeps v5's webstatic seam with v4's
  HIVE_HUB_PROXIED substitution
- tests: v4's new #7474/#7482 fixtures used `reviewer` as a cadence-only
  agent, but v5 ships `reviewer` on-demand in the L5/L6 packs (#5002) —
  renamed to `auditor`; ACMMPackManagedAgentNames reverse-containment check
  made generic
- docs/api-reference.md: v5 route table kept; citations re-pinned

Closes #7463


clubanderson:
Independent verification of this top-up. I had re-derived the same merge locally before this PR appeared (I stopped rather than open a competing one), so what follows is a comparison against a second, separately-resolved merge of the same two branches rather than a read-through. **It holds up, and I'd merge it.**

The two resolutions converge on the same judgement calls, including the non-obvious one. I hit the `reviewer` collision too: v4's new #7474/#7482 fixture uses `reviewer` as a plain cadence-scheduled agent, but `OnDemandAgentsFromPacks()` reads the embedded pack YAMLs globally and v5's L5/L6 packs ship `reviewer` with `on_demand: true`, so `agentUnscheduledInMode` short-circuits and `TestBuildAgents_UnscheduledInMode` fails with `UnscheduledInMode = false, want true` on exactly the `reviewer`/`reviewer-2` rows and nothing else. Renaming the fixture agent to `auditor` is the right call — it keeps v5's pack untouched (which is the thing that must not be reverted) while preserving precisely what v4's test asserts. Two agents arriving at the same fix independently is reassuring; it also means this is a real semantic collision rather than a resolution artifact, and it is worth remembering that it will recur every time a v4 test picks an agent name v5's packs have claimed.

On the specific silent-revert risks in this merge — the ones created by v4's god-file splits landing on v5's divergent layout — I checked the merged tree for a sentinel from each and they all survive:

- **Digest pin (#6290).** This is the one I'd have bet on being lost, because v4's `saas_upgrade.go` extraction was cut from a v4 `saas.go` that never had the guards. All three are present at `saas_upgrade.go:440`, `:623` and `:1235` — the upgrade refusal, the branch/channel-switch refusal, and the auto-upgrade sweep's silent skip. Taking v4's file wholesale without re-applying these would have silently re-armed upgrades on digest-pinned hives, with no test to catch it.
- **`trustedSpokeSelfServiceUser`.** v5 renamed and generalised v4's `trustedSpokeUpgradeUser`; v4's extracted upgrade file still called the old name. The merged tree uses the v5 name at both call sites and keeps the definition in `saas_auth.go`. (This one at least fails the build, so it can't be missed — but it confirms the extraction was reconciled rather than pasted.)
- **`omp onboarding`.** A v5-only `blockingPrompt` entry that lived in the region v4 moved into `manager_pane_signals.go`; it is present in the merged file. Easy to lose and entirely silent if lost.
- v5's other divergences are intact: `issuePriorityNote`, `startupKickInFlight`/`startupLaunchQueued`, `enforceHoldGuard` + the `holdDriftPRs` parameter, `webstatic.MichromaFontFaceCSS`, `checkBlockedThrash` still wired inside `pollTmuxOutputForAgent`, `ActiveReposFor`, `syncAutoMergePolicyToGitHubClient`, `limitActivity`, and `sandboxMutation`.
- v5's deliberate **deletions** stayed deleted, which matters as much: `watchForTrustPrompt`, `pollTmuxOutput`, `captureTmuxPane`, `waitForCLIReady` and `tmuxPaneHasCLI` (the legacy session-based tmux layer, #5795/#5808) are all absent. Those sit on the *v4* side of the `manager.go` hunks and look like additions, so taking v4 wholesale there resurrects several hundred lines v5 removed on purpose — the single easiest mistake in this file.

I ran it rather than trusting CI, at `53d406a9`: `go build ./...` clean, `go vet ./...` clean, and the full `go test ./...` green at 96/96 packages with zero failures. `check-api-reference-citations.sh` passes with no `--fix` needed. `check-inline-js.js` passes both the syntax and `no-undef` passes over both script blocks of `static/index.html` (23,224 lines), so the inline JS resolved into something valid.

Two small notes, neither blocking. First, `v4` has moved again while this was open — `9a2817e5` (#7439), `a432de5c` (#7508), `63f4e0ec` (#7504), `3b2a3838` (#7509) and `f53a076f` (#7511) are not in this branch, and #7508/#7504/#7511 are more splits of the same three god files, so the next run will conflict on `saas.go`, `manager.go` and `contribute_ws.go` again. `Refs #7463` rather than `Closes` is exactly right for that reason. Second, and more of a suggestion for whoever lands it: please **merge rather than squash**, as #7494 was. The reason this issue keeps reopening is that the earlier top-ups (#7466, #7481) were squashed, which left git with no ancestry record that v4's content was already in v5, so it re-conflicted on content that was semantically already there. #7494 gave us a real merge base at `82308e22` and this merge was visibly cheaper as a result — 11 conflicts instead of 16, and a genuine three-way base in every hunk.

Issue #2674, opened 2026-08-05
F5 fatal-egress: NET_ADMIN drift + OpenShift/OVN iptables limitation block v4 rollout to non-OKE hives


clubanderson:
## Context
The F5 fix (#2664, v4) makes a failure to establish the forced iptables egress redirect FATAL by default (container exits rather than running with advisory-only proxy enforcement), with an explicit `HIVE_PROXY_ADVISORY_OK=true` opt-out. This is correct security behavior. But rolling the F5 image to the existing fleet will crash-loop hives for two distinct reasons discovered 2026-08-05:

## Blocker 1 — NET_ADMIN drift (pre-#1222 hives)
The provisioning template (`saas_provision.go:2148`) requests `NET_ADMIN`, but hives provisioned before #1222 have `securityContext: {}` on their live deployment and never had it re-applied. On plain OKE this is patchable (verified: patching the deploy makes iptables come up — done for `kubestellar-console-4vkt` and `projectbluefin-knuckle-gjvq`).

**Fix:** the hub should reconcile the securityContext (NET_ADMIN) onto every hive's deployment on upgrade/heartbeat, so drift can't persist. Until then, F5 rollout must be gated behind a per-hive NET_ADMIN check.

## Blocker 2 — OpenShift/OVN blocks iptables nat-write beneath NET_ADMIN (the harder one)
On the vllm-d cluster (OpenShift), hive pods run under the **privileged** SCC WITH NET_ADMIN in the pod spec, and iptables is installed and can READ the nat table — yet `iptables -t nat -N` (chain creation) fails with "need NET_ADMIN". The capability is effectively denied at a layer BELOW the SCC/pod-capability grant (node/CNI/OVN or seccomp). Granting a different SCC does not help (already privileged). **F5 forced-egress enforcement is therefore UNAVAILABLE on OpenShift/OVN clusters** — the MITM confinement is advisory-by-platform there, not by misconfiguration.

**Consequence:** when v4 reaches the vllm-d OpenShift hives (currently all on `v2-latest`, so NOT yet affected), all ~32 will crash-loop on the F5 fatal check unless `HIVE_PROXY_ADVISORY_OK=true` is set on them.

## Required before any v4 rollout beyond OKE
1. Hub-side reconcile of NET_ADMIN onto deployments (fixes Blocker 1 fleet-wide).
[...]

PR #7521, opened 2026-09-18
🐛 fix(hub): reconcile #7457's auth-url onto the hive-contribute Ingress of existing spokes

### LINKED ISSUE #7517: 🐛 bug(contribute): #7457's auth-url on the hive-contribute Ingress never reaches spokes provisioned before it — /api/contribute/me still 401 on hosted-projectbluefin-common
Follow-up to #7453 / #7457. That fix is correct, but only half of it reaches a spoke that already existed when it merged.

## What I saw

On `hosted-projectbluefin-common-nmq5.hive.hivecommons.dev/contribute/operations`, signed in to the hub, the Operations tab still says I am not signed in. Firefox console shows the same capture #7453 was filed on:

```
GET /api/contribute/me  →  401
```

The spoke is running code that includes #7457: `/api/contribute/status` reports `served_sha: f73e94d`, which is after `c80eaa69` (the #7457 merge).

Separately, clicking **Sign in with GitHub** from that tab bounced between the spoke and the hub about twenty times before the browser gave up. I have not captured the redirect chain yet; that part may or may not share the cause below, so I am only recording it here for now.

## Why the 401 is still there

#7457 shipped two halves:

1. **Code** — `setPublicPathIdentity` on the hub auth-check, `/auth/return` and `ccSignInHref` on the spoke. Deployed.
2. **Ingress** — `nginx.ingress.kubernetes.io/auth-url` and `auth-response-headers` added to the `hive-contribute` Ingress in `k8sManifestTemplate` (`src/pkg/hub/saas_provision.go`), so nginx actually calls the auth-check on `/api/contribute/*`.

The template is rendered and `kubectl apply`'d by `provisionHive`, which runs when a hive is created (and via the pool replenisher). Nothing re-applies it to spokes that already exist. The hub has reconcilers for other per-spoke state (`netadmin_reconcile.go`, `perhive_env_reconcile.go`, the vanity-host JSON patch at `saas_provision.go` ~L3853) but none for Ingress annotations.

So every hosted spoke provisioned before 2026-09-18 still has a `hive-contribute` Ingress with no `auth-url`. nginx never asks the hub who the caller is, `X-Hive-User` never arrives, and `/api/contribute/me` keeps answering 401 — exactly the state #7453 described. The PR body anticipated this: *"Not verified here (needs the hosted cluster): the live round trip once the hub and a **re-provisioned** spoke carry this."*

I have not been able to read the Ingress object itself (no cluster access); the inference is from the code path plus the live 401. If someone with cluster access can confirm `kubectl -n <spoke-ns> get ingress hive-contribute -o yaml` lacks the `auth-url` annotation, that settles it.

## What "fixed" looks like

- Existing spokes get the `hive-contribute` `auth-url` / `auth-response-headers` annotations without being re-provisioned — a hub reconciler on the same pattern as the vanity-host patch (JSON-patch the two annotations onto `hive-contribute` for every hosted spoke, idempotent, logged), or at minimum a documented one-off `kubectl annotate` for the current fleet.
- `/api/contribute/me` on `hosted-projectbluefin-common-nmq5` answers 200 with the signed-in user, and the Operations tab's "Your contribution" card shows them.
- More generally: any future change to `k8sManifestTemplate` that alters an existing Ingress needs a reconcile path, or a note in the PR that existing spokes must be re-provisioned. Worth a line in `docs/security-model.md` or the provisioning skill.

## The sign-in loop

Recorded above but not yet traced. I will add the Network-tab redirect chain (URLs + status codes for the first few hops) when I have it. If it turns out to be the same missing annotation, this issue covers it; if the loop is on the gated Ingress (`/auth/return` → hub `/login` → back → 401 again), that is a cookie/session-scope question and should get its own issue.

Danathar:
## Summary

Hosted spokes that existed before #7457 merged still answer `401` on `/api/contribute/me` to their own signed-in owner, because #7457's two nginx annotations live in the provisioning template and nothing re-applies the template to a spoke that already exists. This PR adds the missing re-apply: a hub reconcile sweep that puts `auth-url` / `auth-response-headers` onto the `hive-contribute` Ingress of every hosted spoke on an nginx cluster, idempotently, without re-provisioning and without rolling any pod. It also writes down the general rule so the next template change to an existing object does not repeat this.

Fixes #7517

## What was wrong

#7457 shipped two halves. The code half (`setPublicPathIdentity`, `/auth/return`, `ccSignInHref`) rolled out with the next image — `hosted-projectbluefin-common-nmq5` reports `served_sha: f73e94d`, past the merge. The Ingress half is `kubectl apply`ed only by `provisionHive`, at creation time. The hub has reconcilers for other per-spoke state (`netadmin_reconcile.go`, `perhive_env_reconcile.go`, the vanity-host patch) but none for Ingress annotations, so on every pre-existing spoke nginx never asks the hub who is calling, `X-Hive-User` never arrives, and `/api/contribute/me` keeps answering `401` — the #7453 symptom, still there. The #7457 PR body anticipated exactly this ("needs … a **re-provisioned** spoke").

## What changed

`src/pkg/hub/contribute_ingress_reconcile.go` (new) — on the NET_ADMIN / per-hive-env pattern:

- `reconcileContributeIngressIfDue()` runs at most every 15 minutes from the SHA poller (same guarding mutex, same throttle shape).
- For every hosted hive not in `provisioning`, on a cluster that is nginx (not `openshift-route` — those render Routes, there is no Ingress) and kubectl-reachable (not pull-only), not recently-unreachable: `kubectl get ingress hive-contribute -o json`, compute the drift, `kubectl patch --type merge` only the annotations that are missing or stale.
- Pure decision function `contributeIngressAnnotationPatch(raw, want)` → merge-patch body or `""` when converged; an unparseable object is an error and **no patch** (never patch blind).
- Expected values from `contributeIngressAuthAnnotations(hubPublicURL(), hiveID)` — the same two inputs the template renders from. `TestContributeIngressReconcileMatchesTheTemplate` renders `k8sManifestTemplate` and pins them byte-equal, so the sweep and provisioning cannot disagree about what "converged" means.
- Sweep accounting: warns if the filter selects nobody on a populated registry (the failure the NET_ADMIN lane shipped with, #2674); Info summary only when something was patched or failed; per-hive Info line on each patch.

`server.go` — `lastContributeIngressReconcile` field. `saas_sha_poller.go` — the call, next to its siblings.

`src/docs/security-model.md` — new subsection "Provisioning-template changes reach existing spokes only through a reconcile", with #7457/#7517 as the worked example (the issue asked for this).

`changelog.d/fixed-7517-contribute-ingress-auth-url-reconcile.md`.

## What the operator will see

Within one sweep of the hub picking this up: one log line per drifted spoke — `reconciled auth-url onto hive-contribute ingress (#7517)` with `hive_id` / `cluster` / `namespace` — then `/api/contribute/me` on `hosted-projectbluefin-common-nmq5` answers `200` with the signed-in user on the next request (nginx re-reads the Ingress; nothing restarts). Subsequent sweeps are silent for that spoke.

Interim one-off for the current fleet, if anyone wants it before the hub rolls (the sweep will find nothing to do afterwards):

```
kubectl -n hive-hosted-<id> annotate ingress hive-contribute --overwrite \
  nginx.ingress.kubernetes.io/auth-url="https://hive.hivecommons.dev/api/saas/auth-check?hive=<id>&uri=\$request_uri" \
  nginx.ingress.kubernetes.io/auth-response-headers="X-Hive-User,X-Hive-Role,X-Hive-Proxy-Auth"
```

## Tests

`src/pkg/hub/contribute_ingress_reconcile_test.go` (new):

- template ↔ reconcile values byte-equal (both TLS modes)
- pre-#7457 Ingress (issuer + timeouts, no auth-url) → patch carries exactly the two annotations, leaves the rest alone, adds no `auth-signin`, keeps the literal `$request_uri` — **this is the case that reproduces the bug**
- converged Ingress → no patch (the sweep can run forever without touching a healthy fleet)
- stale `auth-url` (old hub host) → only that key in the patch
- Ingress with no `annotations` key at all → both added
- empty / `NotFound` text / non-JSON → error, no patch
- per-hive uniqueness of the auth-url; eligibility predicate; throttle; poller call site pinned

```
go vet ./pkg/hub/            clean
go test ./pkg/hub/           ok  (103.8s, whole package)
gofmt -l                     clean on touched files
```

Not verified here (no cluster access): the live patch on the hosted cluster. The kubectl shapes (`get ingress -o json`, `patch ingress --type merge`) are the ones `addVanityHostToIngress` already uses against the same objects.

## Out of scope

The sign-in bounce the issue records in passing — its redirect chain has not been captured, and the issue itself says it may be a cookie/session-scope question deserving its own issue. If it was the missing annotation (spoke `/auth/return` → hub → back with no identity → again), this fixes it; if it persists after the sweep, please file it.


clubanderson:
## Review of #7521 against #7517 — verified against `v4` @ `ff7630e` and this branch @ `8bdb0b3`

I was asked to fix #7517 independently, found this PR already open, and reviewed it instead of opening a competing one. Everything below is checked against the code, not read off the descriptions. I have **no cluster access**, so nothing here asserts live spoke behaviour.

### The issue's and the PR's factual claims — what held

| Claim | Verdict | Evidence |
|---|---|---|
| #7457 put `auth-url` / `auth-response-headers` on the `hive-contribute` Ingress in the provisioning template | **true** | `src/pkg/hub/saas_provision.go:3681-3682`, inside the `name: hive-contribute` block starting `:3664` |
| Those annotations are applied only at provision time | **true** | the template is rendered/applied by `provisionHive`; no other writer touches an Ingress annotation except `addVanityHostToIngress` (host list, not these keys) |
| The hub has no Ingress-annotation reconciler on `v4` | **true** | only `netadmin_reconcile.go` and `perhive_env_reconcile.go` exist under `src/pkg/hub/*reconcile*.go` |
| The reconcile's expected values equal what the template renders | **true, and pinned** | template data is `"HubPublicURL": hubPublicURL()` (`saas_provision.go:2590`) and `"ID": h.ID` (`:2490`) — the same two inputs `contributeIngressAuthAnnotations` takes; `TestContributeIngressReconcileMatchesTheTemplate` renders and compares |
| OpenShift-Route clusters have no Ingress to patch, so skipping them is right | **true** | the whole Ingress section of the template is inside `{{- if .IsNginxIngress}}`; `hive-contribute` is rendered only there |
| #7457 changed only `hive-contribute` among existing objects, so a one-Ingress reconcile is the right scope | **true** | `git show c80eaa69` — "Provisioning: hive-contribute carries the same auth-url and auth-response-headers…" |

The design (pure decision function, patch only drifted keys, error-and-don't-patch on unparseable input, `provisioning` the only excluded status, merge patch so cert-manager/timeout/vanity annotations survive) is right, and the test file covers the pre-#7457 case that *is* the bug. I ran it: `go build ./...` clean, `go vet ./pkg/hub/` clean, all 11 `-run ContributeIngress` tests pass, `gofmt -l` clean on both new files.

Three things, one of which I think blocks.

---

### 1. Blocking — the sweep is wired into the poller's **startup preamble only**, so it runs once per hub process, never on a tick

On this branch:

- `s.reconcileContributeIngressIfDue()` — `src/pkg/hub/saas_sha_poller.go:710`, inside `StartLatestSHAPoller` (`:670`), in the one-shot preamble that runs before `ticker := time.NewTicker(...)`.
- `pollLatestSHAsTick` (`:760`) — the actual ticker body — calls `s.reconcileNetAdminIfDue()` (`:774`) and `s.reconcilePerHiveEnvIfDue()` (`:775`) and **not** the new one.

Both siblings the PR models itself on appear in **both** places: net-admin at `:695` and `:774`, per-hive-env at `:702` and `:775`. The new lane only got the first half.

Consequences, all of which contradict what the code and PR body promise:

- `contributeIngressReconcileInterval` (15m) is effectively dead: the second call never happens, so the throttle can never admit a second sweep.
- `"patch failed — will retry next sweep"` is unreachable. There is no next sweep.
- A cluster that is down, or suppressed by `clusterRecentlyUnreachable` (`clusterUnreachableTTL = 10m`, `saas_upgrade.go:36`) at the moment the hub boots, stays unconverged until someone restarts the hub — and the fleet-wide upgrade lane routinely sets that same breaker, so this is not a hypothetical.
- The `"selected NO hives"` warning — the accounting the PR added specifically because the NET_ADMIN lane shipped dead — fires at most once, at boot, when `listSaaSHives()` is most likely to be thin.

One line fixes it: add `s.reconcileContributeIngressIfDue()` to `pollLatestSHAsTick`, immediately after `s.reconcilePerHiveEnvIfDue()` at `:775`. Keep the preamble call as well, matching the siblings, so a hub restart converges immediately.

### 2. The wiring guard test cannot see that — it is placement-blind

`TestContributeIngressReconcileIsWiredIntoThePoller` does `os.ReadFile("saas_sha_poller.go")` + `strings.Contains`, which is satisfied by the string appearing anywhere in the file. I mutated it both ways:

| Mutation | Result | Reading |
|---|---|---|
| Delete `s.reconcileContributeIngressIfDue()` from `saas_sha_poller.go` entirely | `--- FAIL: TestContributeIngressReconcileIsWiredIntoThePoller … never calls reconcileContributeIngressIfDue()` — a genuine assertion failure, not a build error | the test does have teeth against *total* unwiring |
| Move the call out of the preamble and into `pollLatestSHAsTick` (i.e. the *correct* placement, and equally the current *incorrect* one) | `ok github.com/hivecommons/hive/pkg/hub` | the test cannot distinguish once-at-boot from once-per-tick — which is exactly the defect in §1 |

Suggested strengthening: assert the call appears inside `pollLatestSHAsTick`'s body — e.g. slice the source from `func (s *HubServer) pollLatestSHAsTick(` to the next top-level `func ` and require the call in that slice. That is the assertion that would have caught this, and it is the assertion #2674's lesson actually asks for. A negative control comes free: the same slice must *not* contain a name that was never wired.

### 3. Blocking (mechanical) — api-reference citation drift, CI is red on it now

`relative links and anchors resolve` is failing with `api-reference citations: 25 drifted citation(s)` ([run 35299277990](https://github.com/hivecommons/hive/actions/runs/35299277990/job/105458283447)). It is this PR's doing, not pre-existing — I ran the checker on both:

```
v4  @ ff7630e : api-reference citations: all pkg/<file>:<line> references verified.   (exit 0)
pr  @ 8bdb0b3 : 25 drifted citation(s)  (DRIFT doc:538 GET /fleet -> pkg/hub/server.go:1587 … )
```

Cause: the 5-line `lastContributeIngressReconcile` field added at `server.go:1076` shifts every route registration below it by 5. Fix is `bash src/scripts/check-api-reference-citations.sh --fix` from the repo root, commit the rewritten `api-reference.md`, then re-run **without** `--fix` to confirm.

### Smaller, non-blocking

- `TestContributeIngressReconcileThrottle` re-implements the due/stamp logic inside the test rather than calling `reconcileContributeIngressIfDue()`. A real bug in the production throttle (forgetting to stamp `lastContributeIngressReconcile`, say) would not be caught. With an empty hive registry the real method is safe to call, so this could be tightened.
- `considered++` runs before the `ns == ""` guard, so a hive with no resolvable hosted namespace counts as "considered" and can mask the `considered == 0` warning. Cosmetic.
- Marking the whole cluster unreachable on a *patch* failure, moments after a successful `get` on the same cluster, conflates "RBAC rejected this object" with "cluster is down" and suppresses other lanes for 10m — but this is exactly what `netadmin_reconcile.go:255` and `perhive_env_reconcile.go:962` already do, so it is consistent precedent and I would not change it here.

---

### On the sign-in bounce the issue records separately

**I read it as independent of this fix, and worth its own issue.** The reasoning is structural, and does not need a cluster:

A redirect loop requires something that *issues redirects*. The pre-#7457 `hive-contribute` Ingress has neither `auth-url` nor `auth-signin`, so nginx does not call the hub and does not redirect anything on `/api/contribute/*` — it proxies straight through. That is precisely why the symptom there is a flat `401`, not a bounce. And the post-fix Ingress still deliberately carries **no** `auth-signin` (`saas_provision.go:3676-3682` and the comment above it; `TestContributeIngressPatchAddsTheMissingAnnotations` asserts the reconcile never adds one).

The "Sign in with GitHub" link goes to `/auth/return?to=<here>` (`pkg/dashboard/contribute_landing.go:2661-2665` → `handleAuthReturn`, `pkg/dashboard/api_contribute.go:247,261`), which is a **page** path served by the gated Ingress — the one with `auth-signin: {{.HubPublicURL}}/login?redirect=…` at `saas_provision.go:3566-3569`. That Ingress has carried both annotations since long before #7457 and is untouched by both that PR and this one. So a ~20-hop bounce there is a hub-session / cookie-scope question on the gated lane (auth-check answering 401 to a caller the hub's own `/login` considers already signed in, returning them immediately), not the missing `hive-contribute` annotation.

So: this PR can fix the `401` and leave the loop exactly as it is. I would keep the PR's "Out of scope" note but sharpen it from "if it was the missing annotation, this fixes it" to "this cannot be the missing annotation — if it reproduces, file it against the gated `/auth/return` lane", so nobody closes #7517 believing the bounce went with it.

### What I could not verify

The live round trip: whether `hosted-projectbluefin-common-nmq5`'s `hive-contribute` Ingress actually lacks the annotation today, whether the patch lands, and whether `/api/contribute/me` then answers `200`. I have no cluster access, same as the issue author. The `kubectl get ingress -o json` / `patch --type merge` shapes are the ones `addVanityHostToIngress` already uses against these objects, which is the strongest check available from the repo alone.

With §1 and §3 addressed I think this is the right fix for #7517.

clubanderson (approved):
Correct gap-fill: the reconcile sweep re-applies #7457's auth-url annotations to pre-existing spokes, idempotent, with tests; citations fixed on the branch. LGTM.

PR #7538, opened 2026-09-18
✨ dashboard: link the plan view from the governor PLANNING tile

### LINKED ISSUE #7537: ✨ dashboard: PLANNING tile promises a plan-review view that does not exist
**Target: v5**

The governor PLANNING tile's tooltip tells the operator: *"N plan(s) awaiting your review — approve them in the plan view before agents start the work"* — but there is no plan view.

Evidence (v4, `pkg/dashboard/static/index.html`):
- line ~15811: `if (typeof openPlanReview === 'function' && data.epic_id) openPlanReview(data.epic_id);` — `openPlanReview` is never defined anywhere.
- The tile (line ~4744) counts `planning.awaiting_review` and turns amber, but offers no navigation.

The backend is already complete (`pkg/dashboard/plan_handlers.go`):
- `GET /api/plan/{epicID}` — PlanTree (children, execution tags, dependency edges)
- `POST /api/plan/{epicID}/approve` / `/reject`
- `POST /api/plan/{epicID}/child/{childID}` — retag / remove

Needed:
1. A way to **list** draft epics (either a `GET /api/plans` listing endpoint or filter `/api/beads` by `type=epic`, surfaced in the UI).
2. A **plan-review view** (modal or panel): show the epic + ordered children with execution tags and dependency edges; retag/remove; approve/reject buttons.
3. Make the PLANNING tile's amber state **clickable**, opening that view; define `openPlanReview(epicID)` so the ⧉ Plan flow's existing call-site starts working.

Until then, operators can only approve plans with raw curl against the API — the tooltip's instruction is impossible to follow.

clubanderson:
The governor PLANNING tile's tooltip says *"approve them in the plan view"* — but there was no plan view: `openPlanReview` was invoked only behind a `typeof` guard and never defined, so operators could only approve plans with raw `curl`.

This PR wires the whole path:

- **`GET /api/plans`** (new, read-only like `GET /api/plan/{epicID}`): `planning.ListPlans` enumerates every plan across all bead stores — drafts awaiting review first, then epics queued for the architect, then approved plans — with issue tracing and child counts.
- **Plan list modal** (`openPlanList`): every plan with its status badge and source issue; click a row to drill in.
- **Plan-review modal** (`openPlanReview`, the function the ⧉ Plan flow already tried to call): the epic's children with execution tags and dependency edges, retag (agent/human) and remove controls while draft, and **Approve** / **Return to draft** buttons — all against the existing owner-gated endpoints.
- **Both governor PLANNING tiles** (classic `gov-stat` and oc `gov-metric`) are now clickable and open the plan view; tooltips say so.

Tests: `ListPlans` selection/ordering/nil-stores in pkg/planning; `GET /api/plans` happy path + no-stores in pkg/dashboard. Inline-JS lint (`check-inline-js.js`) passes. api-reference citations re-pointed (33 drifted from the api.go route insertion).

Fixes #7537


clubanderson:
Reviewed this against #7537 and against the code on both branches, plus ran the repo's gates locally on `feat/plan-review-view` @ fd070b51. The design is right and the wiring is genuinely correct — but **there is one blocking CI failure this PR introduces**, and one meaningful test gap.

## Issue claims: all verified, nothing stale

| Claim | Verdict |
|---|---|
| `openPlanReview` called but never defined | ✅ `src/pkg/dashboard/static/index.html:15811` is the *only* occurrence on `v4` |
| tile counts `planning.awaiting_review`, offers no navigation | ✅ `index.html:4744` (oc `gov-metric`) **and** `index.html:7881` (classic `gov-stat`) — the issue mentioned one, there are two, and this PR correctly fixes both |
| backend complete | ✅ `plan_handlers.go:221/243/271/300` + `api.go:322-326` |
| no listing reader existed | ✅ `BuildPlanning` (`status_builder.go:1808`) only *counts*; nothing names plans |

One thing worth recording for posterity: the reason this shipped broken is that `.github/scripts/check-inline-js.js:119-137` deliberately exempts a `no-undef` hit whose own line carries a `typeof NAME` guard — and the comment there literally uses `openPlanReview` as its example. So the one linter that could have caught it was documented to ignore exactly this case.

## 🔴 Blocking: `GET /api/plans` is missing from `dashboard/openapi.json`

Both red jobs are the same root cause:

```
openapi_route_parity_test.go:388: dashboard/openapi.json has drifted from the registered dashboard routes
  (317 registered, 307 documented, 1 problem(s)):
  registered but undocumented: GET /api/plans — add it to dashboard/openapi.json or,
  if it genuinely should stay unpublished, add a routeParityException with a reason
--- FAIL: TestOpenAPISpecCoversEveryRegisteredRoute (0.73s)
```

(`test (rest 2/4)` job 105473189823 and `test (dashboard shuffle 1/2)`; reproduced locally, and the same test is `ok` on a clean `v4` worktree — so it is this PR, not a flake.)

Adding this block next to `"/api/plan/{epicID}"` in `dashboard/openapi.json` turns it green locally (verified: `go test ./pkg/dashboard/ -run TestOpenAPI` → `ok`):

```json
"/api/plans": {
  "get": {
    "tags": ["Plan"],
    "summary": "List plans",
    "description": "Lists every plan (an epic carrying a plan_status) across all bead stores, drafts awaiting review first. Not owner-gated, matching GET /api/plan/{epicID}.",
    "responses": {
      "200": {
        "description": "Plan list",
        "content": {
          "application/json": {
            "schema": {
              "type": "object",
              "properties": {
                "ok": { "type": "boolean" },
                "plans": {
                  "type": "array",
                  "items": { "type": "object", "description": "planning.PlanSummary" }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

Please hand-insert rather than round-tripping the file through a JSON serializer — mine re-escaped every em dash in the spec.

## 🟠 Also: `src/docs/api-reference.md` has no row for the new endpoint

All five sibling plan routes are documented at `src/docs/api-reference.md:405-409`; the diff here only re-points drifted citations and never adds a row. `check-api-reference-citations.sh` is clean (it only validates existing citations, it doesn't enforce coverage), so nothing fails — but the table is now incomplete. Suggested row, matching the neighbours' auth wording:

```
| `GET` | `/api/plans` | Dashboard auth/session | Plan List | `pkg/dashboard/api.go:324` |
```

…then re-run `bash src/scripts/check-api-reference-citations.sh` (no `--fix`) to confirm.

## 🟠 Test gap: the half the issue is actually about has zero coverage

The four new Go tests are real — I mutated `listOrder` (`plan_list.go`, draft rank `0` → `2`) and got:

```
--- FAIL: TestListPlans_OrderingDraftsFirst (0.02s)
    plan_list_test.go:56: wrong order: a draft epic, b approved epic, b draft epic
```

But they only cover the *listing reader*. I reverted `src/pkg/dashboard/static/index.html` to its `v4` state — deleting `openPlanReview`, `openPlanList`, both `data-action` tiles, i.e. the entire defect #7537 describes — and **every gate still passed**: `check-inline-js.js` reported `All script blocks passed syntax and no-undef checks` (with `1 typeof-guarded reference(s) exempted` — the very exemption that hid the bug), and `go test ./pkg/dashboard/` was clean. Nothing would stop this regressing.

`pkg/dashboard/static_index_test.go:157` (`TestStaticTerminalLinksRenewAssertionBeforeOpening`) is the established pattern for pinning `index.html` wiring. This one is mutation-proven — **FAIL** on `v4`'s `index.html` (`plan-review wiring missing "async function openPlanReview(epicID)"`), **ok** on this branch:

```go
// TestStaticPlanReviewWiring pins the dashboard half of #7537: the governor
// PLANNING tile promised "approve them in the plan view", and the ⧉ Plan flow
// already called openPlanReview() behind a `typeof` guard, but the function was
// never defined — and no gate noticed, because check-inline-js.js deliberately
// exempts typeof-guarded references.
func TestStaticPlanReviewWiring(t *testing.T) {
	body, err := os.ReadFile("static/index.html")
	if err != nil {
		t.Fatal(err)
	}
	html := string(body)
	for _, want := range []string{
		"async function openPlanReview(epicID)",
		"async function openPlanList()",
		"function planApprove(epicID)",
		"function planReject(epicID)",
		"function planChildRetag(epicID, childID, execution)",
		"function planChildRemove(epicID, childID)",
		`data-action="openPlanList"`,
		`data-action="openPlanReview"`,
		"fetch('/api/plans')",
	} {
		if !strings.Contains(html, want) {
			t.Fatalf("plan-review wiring missing %q", want)
		}
	}
	if !strings.Contains(html, "if (typeof openPlanReview === 'function' && data.epic_id) openPlanReview(data.epic_id);") {
		t.Fatal("the ⧉ Plan call-site for openPlanReview disappeared")
	}
}
```

## 🟡 Minor: `data-stop="1"` on the issue link is inert

In `openPlanList`, the row is `<div data-action="openPlanReview" …>` and the issue link inside it carries `data-stop="1"`. The dispatcher reads `data-stop` off the element it resolved the *action* from — `el = event.target.closest('[data-action]')`, then `if (el.dataset.stop === '1')` (`index.html:15361,15376`) — and that's the row `div`, which has no `data-stop`. So clicking the `org/repo#123` link opens GitHub in a new tab **and** opens the plan-review modal behind it. The precedent at `index.html:8278` puts `data-stop="1"` on the element that also carries `data-action`. Simplest fix: give the anchor its own no-op `data-action` + `data-stop="1"`, or make the row's drill-in a button next to the link instead of the whole row.

## 🟡 Minor: approve / reject / retag / remove render for every role

Those four endpoints are owner-gated (`plan_handlers.go:248,276,306` via `requireOwnerRole`), so a read-write member gets a 403 and an error toast — safe, but a dead-end button. The dashboard already hides owner-only affordances by role (`index.html:25234-25236`); worth the same treatment.

## 🟡 Nit: the tile's number and the list's buckets can disagree

`AwaitingReview` counts **every** epic with `plan_status == draft` (`status_builder.go:1834-1837`), including the issue-minted ones that are still `decompose_pending` (`issue.go:135-136` sets both at mint). `listOrder` puts those in the "queued for architect" bucket instead (`plan_list.go`). So a tile reading "3 to review" can open a list with one row under *awaiting your review* and two under *queued for architect*. Every plan is still listed — just a momentary "where did they go?". A sentence in the empty/queued area, or aligning the tile's count, would close it.

## Things I checked and found correct

- Every `PlanSummary`/`PlanTree` field the JS reads matches the Go JSON tags (`epicId`, `epicTitle`, `planStatus`, `approved`, `children[].dependsOn`, `planRef`, `execution`) — `plan_review.go:19-49`, `plan_list.go`.
- `data-arg-types="s,s,v"` on the retag `<select>` is valid: `v` → `el.value` (`index.html:15334,15348`), and `change` falls through to `hiveDispatchAction` via `default:` (`index.html:15452`). All the new functions are top-level declarations in the same classic script, so `window[action]` resolves them.
- `planShowModal` matches the existing dynamic-overlay recipe (`index.html:12495` does the same `className = 'nous-config-overlay'`), and the class is in `OVERLAY_SELECTOR` (`index.html:4196`), so Esc/backdrop handling picks it up. `showToast(msg, type, persistent)` / `dismissToast(el, newMsg, newType)` signatures match `index.html:15523,15567`. All interpolated values go through `escapeHtml`; no inline `on*=`, so CSP stays intact.
- `handlePlanList` being un-gated is consistent with `handlePlanTree`, and it is nil-safe on `deps`/store values (`ListPlans` skips `nil` stores).
- `changelog.d/added-7537-plan-review-view.md` starts with a `- ` bullet and uses the right category — guard is green.
- Locally: `go build ./...` clean, `gofmt -l` clean for both new files, `pkg/planning` `ok`, `check-api-reference-citations.sh` (no `--fix`) verified, `check-inline-js.js` on `index.html` passes with **no** typeof exemption remaining (proof `openPlanReview` is now really defined).

## Base branch

The issue says *Target: v5*; this targets `v4`. The defect exists on both (`origin/v5:src/pkg/dashboard/static/index.html:16111` has the same orphan call), and `v4` is the default branch that forward-merges into `v5` (#7525), so `v4` is the right base and `v5` inherits it — noting only so the mismatch with the issue header isn't read as an oversight.

I did not open a competing PR for #7537, since this one already covers it end to end. Fix the openapi entry and this is in good shape.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

PR #7570, opened 2026-09-18
✨ feat(review): tell a human what to merge next

clubanderson:
## What this is for

The bluefin spoke reviews a queue of **411 open PRs across 16 repositories** and has **no auto-merge** — deliberately. The maintainers do not want the hive merging anything yet, and that is a reasonable position to hold while trust is still being earned.

But a hive that may not merge can still do the expensive half of the work: read every open PR and establish which ones are *actually* mergeable right now. That answer currently exists only inside one eval cycle and is thrown away. This publishes it.

## What it produces

One issue per repository, titled `🐝 Hive recommendations — what to merge next`, rewritten in place on the review cadence. It leads with the answer:

> **3 pull requests are ready to merge right now**, out of 8 open.
>
> **Start here:** #1011 fix: handle empty image list on rebase — ready to merge, open 12 days.

then a summary table and one section per bucket — **✅ Ready to merge**, **🔀 Needs a rebase**, **🔴 Failing CI**, **🧑 Waiting on a human decision**, **💤 Stale drafts** — each with a real GitHub query link and a collapsed, copy-pasteable `gh` block.

## The three rules in `render.go`

1. **Lead with the answer.** Not a report to interpret — a number and a place to start.
2. **Never `@`-mention.** The body is rewritten in place, so one raw mention would re-notify that person on every refresh. `NeutralizeMentions` is applied unconditionally.
3. **No checkboxes.** A tick would silently vanish on the next rewrite, which would teach the reader the issue is untrustworthy.

## Why it will not become spam

This is the part I care most about, given that the same spoke was, until this morning, leaving five review comments per PR.

- **One issue, found by title, edited** — not a new comment and not a new issue. GitHub does not notify on an edit, so a subscriber gets *one* notification, ever.
- **The skip-if-unchanged baseline is the body currently on the issue**, fetched each cycle — never process memory. A spoke that restarts nine times a day (as this one did on 2026-09-18) would otherwise forget what it had written and rewrite after every restart. Deriving the decision from the target makes restarts irrelevant by construction. This is the same lesson as #7551/#7558, applied up front rather than after the fact.
- **It will not open an issue to say there is nothing to say.** `MinReadyToOpen` (default 1) gates the first post only; once the issue exists it is kept current even when the answer is "nothing is ready", because that is a useful thing to be able to read.
- **Off by default.** Opening an issue in someone else's repository is not something to do uninvited.

## Safety

- **Held PRs are excluded.** A hold is a maintainer's explicit "not now"; recommending one would discredit the digest on its first outing.
- **Every PR in a `gh` block is listed on its own line**, never a query-driven mass merge, so the reader can delete lines before running it. The inspect blocks are read-only.
- **The digest states plainly** that the hive does not merge, approve, or close anything.
- Canary-scanned, `logscrub`bed and mention-neutralized on the same fail-closed contract as `CreateIssue`/`PostAdvisoryDigest`.
- Labels are applied **only on creation**, so a maintainer who removes one keeps it removed, and the hive never invents a label in someone else's taxonomy.

## Cost

It reads the PR set the eval cycle has **already** enumerated and CI-enriched, so it adds no GitHub reads beyond one issue lookup per repo. A 16-repo hive on a 15-minute cadence could not afford a second enumeration.

## Also in here

`max_perspectives_per_pr`'s config doc still described the per-cycle behaviour that #7562 replaced with a lifetime per-head-SHA budget, and still claimed "No coverage is lost" — the exact opposite of what that change deliberately does. Operators read this text in the dashboard, and it is the knob that decides whether a PR gets one review comment or five, so it is corrected here.

## Testing

- `pkg/recommend`: 16 tests — bucket precedence, strict readiness (`unstable`/`pending` are not ready), fork conflicts, query-link honesty, batch-block safety.
- `cmd/hive`: held PRs cannot leak into a digest; a PR with no repo cannot produce an empty-repo group.
- `go build ./...` clean; `pkg/recommend`, `pkg/github`, `pkg/review` green. The `pkg/config` entrypoint failure reproduces on a clean `v4` and is unrelated.


clubanderson:
Pushed b35931b7: coverage tests for pkg/recommend (82.0% → 99.5%), clearing the 90% floor. Merge watcher armed — will admin-merge on green.

Issue #7571, opened 2026-09-18
🌱 cmd/hive: split main() into boot phases behind injectable deps (follow-up to #7232)


clubanderson:
Follow-up to #7232 (closed). Steps 1 and 2 of that issue landed in #7555, #7561, #7566, #7569; `cmd/hive` measured **58.4%** on v4 afterwards, `runEvalCycle` **52.3%**.

## What is left

`main()` is still **4,142 lines at 0%** (`cmd/hive/main.go:968`). Every *decision* I could identify has been lifted into `boot_seams.go` (config path, hub target, ACMM plan, upgrade-marker verdict, badge URL, primary repo, fleet identity, policy dirs). What remains is **wiring** — constructing ~40 collaborators and starting goroutines — which no amount of decision-lifting will cover.

## Recommendation

The only way to move `main()` off 0% is the #7230 pattern one level up: a `runMainWithDeps(ctx, cfg, deps mainDeps)` whose deps struct carries constructors for the blocking/side-effecting collaborators (GitHub client, dashboard listener, agent manager/tmux, proxies, watchers), with `main()` reduced to flag parsing + `defaultMainDeps()`. Do it in phases, each one PR:

1. Split `main()` at its natural phase boundaries into `bootConfig`, `bootGitHub`, `bootStores`, `bootDashboard`, `bootAgents`, `bootWatchers`, `runLoop` — pure moves, no behaviour change, each returning a small struct. This alone makes each phase measurable.
2. Give each phase a deps struct and a test that drives it with fakes (the `evalCycleFixture` in `eval_cycle_run_7232_test.go` is the worked example for the loop end).

Re-measure with `go test ./cmd/hive/ -count=1 -coverprofile=cov.out && go tool cover -func=cov.out | grep -E 'main\s|^total'` rather than quoting the numbers above. CI floor is `[cmd/hive]=35`; there is headroom, so this is not urgent — it is here so the gap stays visible.

clubanderson:
Step 1 (boot-phase split, b132f4b72) is now on v5 as well via #7625 — v5's main() was re-cut into the same `boot` struct/phase methods (89 fields; bootAdvisory returns bool to preserve v5's ledger-failure early returns, which look worth a look on their own).

clubanderson:
[...]

PR #7602, opened 2026-09-18
🌱 cmd/hive: split main() into boot phases on a shared boot struct (#7571, step 1)

Danathar:
## Summary

`main()` in `cmd/hive` was 4,142 lines at 0% coverage because nothing could call into the middle of it. This PR splits it into 19 boot-phase methods on a new `boot` struct, called in the original order from a 24-line `main()`. Nothing is reordered and nothing behaves differently — it is the "pure moves" step 1 of #7571. Each phase now reports its own coverage number, and the five phases that are cheap to stand up already have tests. Step 2 (a deps struct per phase so the heavier collaborators can be faked) is a follow-up.

Refs #7571 (step 1 of 2; leaving the issue open for step 2).

## What changed

**`src/cmd/hive/boot.go`** (new)
- `type boot struct` — the ~40 locals `main()` used to share, as fields grouped by the phase that sets them.
- `deferStack` — `push`/`run`, LIFO. `main()` defers `run()` once; phases push what they would have deferred.
- `hiveStatePath` — the `/data/hive-state.json` const, moved from inside `main()` to package level.

**`src/cmd/hive/main.go`**
- `main()` → `bootConfig` (returns false on the `--version` / hub-mode fast paths), `bootGitHub`, `bootGovernor`, `bootAdvisory`, `bootAgents`, `bootState`, `bootDashboard`, `bootStores`, `bootCollectors`, `bootKnowledge`, `bootSupervision`, `bootDashboardAPI`, `bootPolicies`, `bootWatchers`, `bootProxy`, `bootLaunch`, `bootHeartbeat`, `bootLanes`, `runLoop`.
- Each method starts by rebinding the fields it reads to locals of the original names, so the moved body reads unchanged; outputs are stored on `b` at the end.

The phase boundaries follow the code's actual order rather than the issue's suggested list exactly, because a pure move can't reorder: e.g. the agent manager is built before the dashboard, and the GitHub request relays need the manager, so they live in `bootAgents`; the bead stores open between the dashboard server and the collectors, so `bootStores` sits there.

### Why one shared struct instead of a return value per phase

Three closures **reassign** `ghClient` and `appAuth` long after boot — the dashboard's `ReinitGitHubFunc`, the config watcher's reload handler, and the heartbeat's GitHub App config callback — and every later reader (the eval cycle, `refreshDashboard`, the Re-check button, the canary scanner, the sweeps) must see the replacement. In `main()` that worked because they all captured the same local. Here it works because they all read `b.ghClient` through the same pointer; those two are never rebound to a local. Per-phase return structs would have to be threaded by pointer through every subsequent phase to preserve that, which is this struct with more plumbing.

### The only in-body edits (everything else is a verbatim move)

| rewrite | why |
| --- | --- |
| `ghClient` → `b.ghClient`, `appAuth` → `b.appAuth` | reassigned by three closures after boot; see above |
| `beadStores` → `b.beadStores` inside `bootDashboard` only | the advisory/sandbox sinks it installs captured the local *before* it was assigned in what is now `bootStores`; they read the field the same way — a nil map until then |
| `&lastActionable` → `&b.lastActionable` | `atomic.Pointer` must not be copied; addressed in place |
| `preShutdownHooks` → `b.preShutdownHooks` | registered in `bootConfig`, added to in `bootAgents`/`bootDashboard`, run by the signal goroutine |
| `defer X` → `b.cleanup.push(func() { X })` for `main()`'s own five defers | they must run when `main()` returns, not when the phase does; LIFO order preserved (the `procLock.Release` inside the singleton `if` included). Defers inside closures/goroutines and `runLoop`'s two ticker defers are untouched — `runLoop` is the last call in `main()`, so they fire at the same moment |
| `statePath` → `hiveStatePath` | a `const` inside `main()` used by three phases |
| `return` → `return false` (×2) in `bootConfig` | `--version` and `HIVE_MODE=hub` return from `main()` |

Verified mechanically: stripping the phase scaffolding (headers, rebinds, output stores) and undoing the rewrites above reproduces the old `main()` body byte-for-byte, apart from the five declarations that became fields or the const (`var preShutdownHooks`, `var beadStores`, `var lastActionable`, `const statePath`, and `ghClient, appAuth :=`). `git diff --color-moved=dimmed-zebra` shows the same thing interactively.

### Deferred-cleanup order, before and after

`main()` had seven defers. Original LIFO on return: `agentTicker.Stop`, `ticker.Stop`, `close(tokenStop)`, `dashSrv.CloseContributeHub`, tracing shutdown, `cancel`, `procLock.Release`. After: the first two are real defers in `runLoop` (fire when it returns, i.e. when `main()` does), the other five are pushed in registration order and `run()` pops them LIFO — same sequence. `os.Exit` skips all of them in both versions.

## Tests

`src/cmd/hive/boot_phases_7571_test.go` (new):
- `TestDeferStack_RunsLIFOAndIgnoresNil`
- `TestBootConfig_VersionFastPathSkipsBoot` — `--version` / `version` return false without touching flags, config, the lock, or the cleanup stack
- `TestBootStores_OpensEnabledStoresAndCountsFailures` — enabled store opens, disabled agent's dir is never created, an unopenable dir is dropped and counted
- `TestBootGovernor_WiresSchedulerAndPrimer` — governor/scheduler/definition resolver built; primer follows `knowledge.enabled`
- `TestBootSupervision_WatchdogFollowsConfiguredMode` — `mode: off` → nil watchdog; default → observe mode; Linear credential resolver falls back to the trimmed work-source key
- `TestBootLanes_BuildsOnlyTheRunnableLanes` — replan on by default / off explicitly; trajectory lane stays nil with no reviewer endpoint; retro off

```
$ cd src && go test ./cmd/hive/ -count=1 -coverprofile=cov.out
ok  	github.com/hivecommons/hive/cmd/hive	5.498s	coverage: 57.0% of statements   (v4 before: 54.9%)

$ go tool cover -func=cov.out | grep -E 'main\.go:[0-9]+:\s+(main|boot[A-Za-z]*|runLoop)\s|boot\.go|^total'
boot.go:153:   push              100.0%
boot.go:161:   run               100.0%
main.go:968:   main                0.0%
main.go:998:   bootConfig          1.8%
main.go:1283:  bootGitHub          0.0%
main.go:1317:  bootGovernor       74.7%
main.go:1467:  bootAdvisory        0.0%
main.go:1582:  bootAgents          0.0%
main.go:1878:  bootState           0.0%
main.go:1984:  bootDashboard       0.0%
main.go:2138:  bootStores         43.5%
main.go:2209:  bootCollectors      0.0%
main.go:2353:  bootKnowledge       0.0%
main.go:2681:  bootSupervision    68.9%
main.go:2787:  bootDashboardAPI    0.0%
main.go:3152:  bootPolicies        0.0%
main.go:3219:  bootWatchers        0.0%
main.go:3447:  bootProxy           0.0%
main.go:3707:  bootLaunch          0.0%
main.go:3817:  bootHeartbeat       0.0%
main.go:5024:  bootLanes          72.7%
main.go:5128:  runLoop             0.0%
total:                            57.0%
```

Also: `go build ./...`, `go vet ./cmd/hive/`, `gofmt -l` clean; `src/scripts/check-dco-trailers.sh` passes. `golangci-lint` is not installed in this environment, so CI's run is the gate for `staticcheck`/`unused`; the new code introduces no new identifiers a phase doesn't use (the script that generated the split only rebinds names the body references, and the compiler confirmed it).

The existing structural guard `TestMainRegistersBothPreShutdownHooks` still holds (its needles are substrings of the `b.`-prefixed calls) and still passes.

## Not in scope

- No behaviour change, so no changelog fragment (`no-changelog`).
- Step 2 of #7571 — per-phase deps structs and fakes for `bootGitHub`, `bootAgents`, `bootDashboardAPI`, `bootHeartbeat`, etc.
- Design-doc line citations into `main.go` (`docs/design/agent-turn-model.md`, `copilot-cost-capture.md`, `knowledge-curator.md`) were already stale before this PR (e.g. `runHub()` cited at :7994, actually :8000) and are not gate-checked; they shift by ~100 lines here and would be a separate docs refresh.


Danathar:
**CI: `test (rest 3/4)` was red — fixed in cd3f7af19.**

The failure was `pkg/github.TestRequestWatchersStayGatedInMain`, a source-level guard that greps `cmd/hive/main.go` for the usable-App gate by its literal spelling `if ghClient != nil && cfg.GitHub.HasUsableApp() {`. The split reads the client through the boot struct, so the same gate is now `if b.ghClient != nil && …` and the needle stopped matching. I only ran `./cmd/hive/` locally before opening the PR, which is why I missed it.

The fix keeps the guard's teeth: it now matches `if (b\.)?ghClient != nil && cfg\.GitHub\.HasUsableApp\(\) \{` — the optional receiver and nothing else — so weakening or removing the gate still fails it, and the ordering checks (`PrepareRequestDirs` before the gate, both watchers after it) are unchanged.

I also ran every other package whose tests read `cmd/hive/main.go` (`pkg/classify`, `pkg/config`, `pkg/dashboard`, `pkg/governor`, `pkg/hub`, `pkg/intent`, `pkg/proxy`, `pkg/reach`, plus `pkg/github`): no other guard depends on a local's name. The three tests that fail in my sandbox (`TestHiveOpenPRScript_MultilineBodyIsValidJSONWithoutPython3`, `TestGateFailsClosedWithoutIP6Tables`, `TestHandleConnectDirectAnthropicReroute`) fail on missing `python3`/`ip6tables`/egress here and passed on CI's first run.

Issue #2356, opened 2026-07-31
/contribute sending "dupe" PRs


castrojo:
hive assigned me an issue and then it went and found one but someone else had sent a PR for the same thing already - if it would have looked before starting work it would have saved a cycle.

castrojo:
- https://github.com/projectbluefin/common/pull/921
- https://github.com/projectbluefin/common/pull/919

This just happened this morning so I'm not sure if the fix propagated or if it's my hive catching up, etc to reporting it

castrojo:
- https://github.com/projectbluefin/bluefin-lts/pull/495
- https://github.com/projectbluefin/bluefin-lts/pull/494

PR #7772, opened 2026-09-19
🐛 fix(contribute-ws): keep an external item's canonical key across a reconnect resume

### LINKED ISSUE #7770: 🐛 bug(contribute-ws): a reconnect drops an external (Linear/Jira) work item's canonical key, so it escapes the double-assignment guard and books no cooldown
**A contributor who reconnects while working an external (Linear/Jira) work item loses that item's double-assignment protection: the hub can hand the same item to a second contributor, and completing it books no cooldown.**

Found by code audit of `v4` at `ff0b345` (v4.67.0), not yet observed live — so treat the production impact below as the consequence the code implies, not a confirmed incident. The code paths and line numbers are exact.

## What happens

Every guard that stops two contributors working the same item keys on the item's *canonical identity* — `owner/repo#42` for a GitHub issue, `owner/repo!ENG-123` for a Linear/Jira item (`worksource.Ref.Key()`, introduced in #4245). The in-flight double-assignment guard in `selectTask` is one of them:

```go
// src/pkg/dashboard/contribute_select.go ~337
if key := conn.currentTask.identityKey(); key != "" {
    activeIssues[key] = true
}
```

When a contributor's socket drops and the relay reconnects, the hub rebuilds `currentTask` from the server-issued lease in the `task_progress` resume path:

```go
// src/pkg/dashboard/contribute_ws.go ~2339
s.contributor.currentTask = &WSTaskAssign{
    TaskID: lease.taskID,
    Kind:   msg.Kind,
    Repo:   lease.repo,
    Number: lease.number,
    Title:  msg.Title,
}
```

It copies `Repo` and `Number` but **not** `Key` — even though the lease carries it (`lease.key`, set at assignment via `recordLeaseForKey(..., chosen.ref.Key(), ...)`). For an external item `Number` is `0` and `ExternalID` is empty, so the rebuilt task's identity collapses:

```go
// src/pkg/dashboard/contribute_ws.go ~432
func (t *WSTaskAssign) identityKey() string {
    if t.Key != "" { return t.Key }
    return worksource.Ref{Repo: t.Repo, Number: t.Number}.Key()
}
```

`Ref{Repo: "acme/repo", Number: 0}.Key()` returns `""` (a zero-numbered ref with no external id has no identity — `worksource/key.go`). So after a resume, a live connection that is actively working `acme/repo!ENG-123` reports `identityKey() == ""`.

Two consequences follow, both from that empty key:

1. **The item drops out of the double-assignment guard.** The `if key != ""` check skips it, so `activeIssues` never contains it. Once the original socket's stale entry is swept from `h.connections` (heartbeat timeout, ~90s, or immediately on a clean close), nothing marks the item in flight. A second contributor sending `ready` — or the same account under another session label — is offered the identical external item and works it concurrently. That is the duplicate-PR class the whole lease/generation/`activeIssues` machinery exists to prevent, reintroduced for external work on the resume path.

2. **Completion and failure book cooldowns against an empty key.** `handleTaskComplete` calls `markTaskCompletedVerdictKeySignal(completedTask.identityKey(), …)`, which returns early on `key == ""` (`contribute_ledgers.go` ~685). So finishing a resumed external item records no completion cooldown, and `recordTaskFailureForTask` (keyed on `identityKey()`) records no failure cooldown either. The item is instantly re-offerable.

GitHub-backed issues are unaffected: `Number > 0`, so the rebuild's `Number` is enough and `identityKey()` returns `owner/repo#number` correctly. The bug is specific to zero-numbered external items (Linear, Jira), which #4245 deliberately made offerable to contributors.

## When it bites

1. A hive is configured with a Linear or Jira work source (`worksource` factory: `linear`/`jira`).
2. An external item (`acme/repo!ENG-123`) is assigned to contributor A; the lease stores `key = "acme/repo!ENG-123"`.
3. A's WebSocket drops and the relay reconnects within the lease window (routine — this is the flow the reconnect section of `contributor-relay.md` describes).
4. A re-asserts the task via `task_progress` with the correct `task_gen`; `lookupLease` matches (it skips the `number` check when `number == 0`), and the hub rebuilds `currentTask` **without** `Key`.
5. A's old connection ages out of `h.connections`.
6. Contributor B (or A under a second session label) sends `ready`. `selectTask` scans candidates; `ENG-123` is not in `activeIssues`, passes admission, and is assigned to B while A is still working it.

## Why it happens

#4245 taught the *assignment* path to carry the canonical key end to end (`task_assign`, the lease, `activeIssues`, cooldowns, holds, queue order — all covered by tests in `worksource_identity_production_test.go`). The *resume rebuild* in `handleTaskProgress` was not updated with it, so it reconstructs `currentTask` from `repo`/`number` alone — exactly the pre-#4245 shape that collapses external items. The lease already holds the correct key; the rebuild just doesn't copy it.

## What "fixed" looks like

- After a reconnect resume, a live connection working an external item reports the item's real canonical key from `currentTask.identityKey()` (the lease's `key`), so `activeIssues` marks it in flight and a second contributor is not offered it.
- Completing or failing a resumed external item books the cooldown against the real key, not against `""`.
- No behavioural change for GitHub-backed items.

The minimal change is to carry the lease's identity into the rebuilt task (the lease stores `key`; `identityKey()` prefers `Key` when set). Preserving `SourceType`/`ExternalID` too would keep the resumed task's metadata complete, but only `Key` is required for the guards above.

A hub-level test: assign an external item (`externalItem("linear", "ENG-1", …)`), simulate a `task_progress` resume against its lease, then assert (a) the resumed connection's `currentTask.identityKey()` equals `acme/repo!ENG-1`, and (b) a second contributor's `selectTask` does not re-offer `ENG-1` while the first still holds it. The existing `worksource_identity_production_test.go` external-item tests are the right neighbours.

## Affected components

- `contribute_ws.go`: `handleTaskProgress` resume rebuild (~2339), `WSTaskAssign.identityKey()` (~432), `selectTask` `activeIssues` scan via `contribute_select.go` (~337).
- `contribute_leases.go`: `taskLease.key` is populated but not propagated on resume.
- `contribute_ledgers.go`: `markTaskCompletedVerdictKeySignal` early-returns on empty key (~685).
- `worksource/key.go`: `Ref.Key()` returns `""` for a zero-numbered ref with no external id.

Danathar:
## Summary

A contributor that reconnected while working a Linear or Jira item lost that item's double-assignment protection on the hub: once the old socket aged out of the connection table, the same item could be offered to a second contributor while the first was still working it, and completing or failing the resumed item booked no cooldown, so it was instantly re-offerable. Found by code audit of v4.67.0, not yet observed live. GitHub issues were unaffected.

Fixes #7770

## Cause

Every guard that stops two contributors working one item keys on the item's canonical identity (`worksource.Ref.Key`, #4245) — `owner/repo#42` for a GitHub issue, `owner/repo!ENG-123` for an external item. The `task_progress` resume path rebuilt `currentTask` from the lease's `repo` and `number` alone, the pre-#4245 shape. `identityKey()` prefers `Key` and falls back to `repo#number`; for an external item `Number` is 0, so the fallback is `""`: the resumed task had no identity, the `if key != ""` scan in `selectTask` skipped it, and `markTaskCompletedVerdictKeySignal` / `recordTaskFailureKey` returned early. The lease has carried the key since #4245 — the rebuild was the one consumer that never read it.

## Fix

- **Resume rebuild** (`contribute_ws.go`): copy `Key: lease.key`, and recover `ExternalID` from that key for a zero-numbered item so the task displays its native key rather than `#0`. `SourceType` is not in the lease and a client-declared value is not adopted — exactly as `repo`/`number` are not.
- **The reconnect-window hedge had the same gap from the other side**: the #2356 release cooldown that covers an item while its relay reconnects was booked only for `Number > 0`, so an external item had no cover at all before the resume even happened. `bookReleaseCooldown` / `clearReleaseCooldown` gain key-based forms and the disconnect and resume paths use the task's `identityKey()`. For a GitHub issue the key is byte-for-byte the `repo#number` string the old form wrote, so persisted ledgers read back unchanged; a synthetic pr-review task has no identity and books nothing, which is what the old `Number > 0` guard was for. The repo/number forms remain for existing callers and now route through `worksource.Ref.Key()`, so a zero-numbered ref books nothing instead of a fabricated `repo#0`.

## Tests

`TestReconnectResume_ExternalItemKeepsCanonicalKey` — over a real hub, the frames a relay sends: assign a Linear item, drop the socket, check the hedge is booked by the item's key, resume via `task_progress`, then assert (a) the live connection's `currentTask.identityKey()` is `myorg/repo1!ENG-1` and the hedge is withdrawn, (b) a second contributor's `ready` gets `task_unavailable` rather than the held item, (c) `task_complete` books the completion cooldown against the real key.

- On the parent it fails at the hedge step; with only the `Key: lease.key` line removed it fails at the identity check (`identityKey() = "", want "myorg/repo1!ENG-1"`).
- `TestReconnectResume_GitHubItemUnchanged` pins the no-regression leg: identity and hedge key for a GitHub issue are exactly as before.

```
```

Docs citation checker passes (lines shifted in two cited files); changelog fragment included.

## Notes for review

The hedge generalization is the one piece beyond the issue's minimal ask. I included it because it is the same defect on the same reconnect flow — without it a resumed external item is protected only *after* the resume, and the 1–60 s reconnect window that #2356 exists to cover still has no guard for external work. It is behaviour-preserving for GitHub items by construction (same key string) and is covered by both tests. Happy to split it out if you'd rather land the rebuild fix alone.

— hive: backend=claude model=claude-opus-4-8


Danathar:
/hold

Danathar:
/hold cancel

Reviewed against the issue it fixes and found correct. Verified: full `go test -race ./pkg/dashboard/` and `go vet` green locally on this head; CI green. The sleep-ratchet fix landed as a test-only commit.

Danathar:
/hold

Re-holding: these core-path changes (lease keying, offer/claim, cooldown keying) were written and reviewed tonight and should get a human maintainer's review before they merge. A review handoff summarizing what each PR changes, what was verified, and the merge order is being prepared for Andy. Nothing wrong was found on this PR — this is caution, not a defect.

PR #7780, opened 2026-09-19
🐛 fix(contribute-ws): keep one task lease per task, not one per identity

### LINKED ISSUE #7774: 🐛 bug(contribute-ws): the task lease is keyed per identity, so a second concurrent task overwrites the first's lease — a flap then revokes healthy in-flight work
**When one contributor identity holds more than one task at a time (allowed for the contributor tier and above), the hub only remembers a lease for the most recent one. A brief disconnect on any of the earlier tasks then loses it: the relay is told the task was revoked, its agent is interrupted mid-turn, and the issue is requeued as if failed.**

Found by code audit of `v4` at `ff0b345` (v4.67.0) and reproduced with a Go test against the real hub. Treat the production impact as the consequence the code implies.

## What happens

The server-issued lease registry is keyed by **contributor identity**, one entry per identity:

```go
// contribute_leases.go — recordLeaseForKey
h.leases[identity] = &taskLease{ taskID, repo, number, key, tier, gen, expiresAt }
```

But the concurrency gate deliberately lets one identity hold several tasks at once — `max_concurrent` defaults to 2 for the `contributor` tier and 5 for `trusted`/`merger` (`pkg/config`), counted across all of that identity's live connections (`contribute_select.go:371`). So an identity can be assigned task X on one connection and task Y on another; the second assignment **overwrites** the lease, and the registry now knows only Y.

If the connection working X then drops and reconnects, the relay re-asserts X, but `lookupLease` finds the entry for Y — the `taskID` does not match — so the resume is rejected and the hub sends `task_revoke`. The relay treats a revoke as terminal: it interrupts the agent mid-turn, drops the task's credential, and relaunches the CLI. X, which was progressing normally, is booked into the short failure cooldown and re-offered as a fresh task. This is exactly the #5681 symptom (a resume that should have worked ends in a spurious revoke), recurring for any identity that holds more than one task.

## Reproduction

`TestAudit_SecondTaskForSameIdentityReplacesFirstLease` (run against the real handlers via `go test -overlay`) assigns two tasks to one identity, shows only the second has a lease, and confirms a flap on the first ends in `task_revoke`. It passes.

## When it bites

Reachable whenever a single contributor identity holds two or more concurrent tasks. That needs two live connections under the same identity — the same GitHub account running two relays with the **same** session label (or two relays of the same backend, which default to the same label). The multi-session feature (`HIVE_SESSION`) gives *distinct* labels distinct identities and distinct lease keys, so it is not affected; the collision is specifically same-identity, multiple-connection, which the docs discourage but the hub still accepts and assigns multiple tasks to.

## Why it happens

The lease was introduced (#4260/#5681) as "the one task the hub issued to this identity," written before `max_concurrent > 1` made "this identity" able to hold several tasks. The map key never moved from identity to the task's own key, so a second assignment silently evicts the first task's resumability.

## What "fixed" looks like

Every task the hub has assigned and not released is independently re-adoptable for its lease window, regardless of how many the identity holds. A resume for an earlier task is not rejected merely because a later task was assigned to the same identity. The natural shape is to key the lease registry by the task (identity + task key) rather than by identity alone, allowing one entry per held task.

## Impact

For contributors running more than one concurrent relay under one identity: any brief network flap on a non-latest task interrupts a healthy agent mid-turn, discards its progress, and requeues the issue with a failure cooldown — the work the lease mechanism was built to protect. Silent apart from the `task_revoke` and the re-assignment in the hub log.

Assign two tasks to one identity across two connections; flap the first; assert its resume is honored (not revoked) and the second's lease is untouched.

## Affected components

- `contribute_leases.go`: `recordLeaseForKey` (~91, `h.leases[identity] = …`), `lookupLease`, `revokeLease`, the `leases map[string]*taskLease` field.
- `contribute_select.go`: the per-identity concurrency gate that permits `max_concurrent > 1` (~371).
- `contribute_ws.go`: `handleTaskProgress` resume (~2299) where the mismatch produces `task_revoke`.

Danathar:
## Summary

A contributor identity holding two tasks at once — allowed for the contributor tier and above (`max_concurrent` 2, trusted/merger 5) across two connections under one identity — only had a lease for the most recent one. A brief disconnect on the connection working the earlier task then lost it: the resume found the *other* task's lease, the task ids did not match, the hub answered `task_revoke`, and the relay interrupted a healthy agent mid-turn, dropped its credential, relaunched the CLI, and the issue was requeued with a failure cooldown.

Fixes #7774

## Cause

The lease registry was `leases[identity]`, written when an identity held one task at a time. The concurrency gate in `selectTask` has since allowed several, and a second assignment simply overwrote the first task's record.

## Fix

The registry is keyed by `leaseKey(identity, taskID)` — one entry per held task:

- `recordLeaseForKey` adds a lease without touching the identity's others (re-recording the *same* task still replaces that task's lease).
- `renewLease` / `lookupLease` / `revokeLease` address exactly one task. A lease for a different task the same identity holds is simply not the one asked for, rather than a mismatch that rejects the resume. `revokeLease` with an empty task id revokes all of the identity's leases (no production caller passes one; the meaning is kept explicit).
- **Persistence is unchanged in shape** — each record already carried identity and task id — so a lease file written before this change loads unchanged, and one written after may hold several records for one identity, each of which comes back.
- `pruneExpiredLeases` / `leasedIssueKeys` iterate values; the latter reads the identity off the lease rather than the map key. Its "the requester's own restored leases are never an exclusion" exception is kept; its now slightly wider reach for a multi-task identity during the two-minute post-restart grace window is noted in place.

## Tests

- `TestLeasePerTask_SecondAssignmentKeepsFirstLease` — registry level: two leases for one identity; renew and revoke touch only their own; both survive a restart.
- `TestLeasePerTask_FlapOnFirstTaskResumes` — the issue's reproduction through the real protocol: one identity, two connections, two tasks; the first connection flaps and re-asserts its task; the resume is honoured (no `task_revoke`, no fresh `task_assign`), the hub holds both tasks, and the second task's lease is untouched.
- With the old one-lease-per-identity eviction re-inserted into `recordLeaseForKey`, both fail — the flap test at `both tasks must hold a lease after the second assignment: x=false y=true`.
- Existing lease tests that indexed the map by identity now go through `leaseKey` / `leaseForLocked` / an `onlyLeaseOf` helper.

```
```

Docs: the multi-session paragraph in `src/docs/contributor-relay.md` no longer says leases are keyed per identity; citation and link checkers pass. Changelog fragment included.

## Notes for review

- Relationship to #7772 (also mine, also open): that PR changes the *rebuild* of `currentTask` in the same resume handler; this one changes the *registry* it looks up. They touch different lines and both apply to `v4` independently.
- The `\x1f` separator in `leaseKey` is a control character neither an identity (`ContributorID` or `ContributorID#session`, sanitized) nor a hub-minted task id can contain; both halves are also stored on the lease, so nothing parses a key back apart.


Danathar:
/hold

Danathar:
/hold cancel

Reviewed against the issue it fixes and found correct. Verified: full `go test -race ./pkg/dashboard/` and `go vet` green locally on this head; CI green. The sleep-ratchet and the -race test-helper fix landed as a test-only commit.

Danathar:
/hold

Re-holding: these core-path changes (lease keying, offer/claim, cooldown keying) were written and reviewed tonight and should get a human maintainer's review before they merge. A review handoff summarizing what each PR changes, what was verified, and the merge order is being prepared for Andy. Nothing wrong was found on this PR — this is caution, not a defect.

Issue #5322, opened 2026-08-31
🐛 bug(contribute-ws): a reconnect can silently drop the task assignment — hub shows the contributor idle and re-queues the issue while the agent is still working it


Danathar:
## Summary

After a WebSocket reconnect, the hub can end up believing a contributor is **idle with no work in flight** while that contributor's agent is actively implementing the task — and the issue is returned to the ready queue, where any contributor can pick it up. Two agents can then implement the same issue, and the second cannot see the first's work because it lives in an unpushed worktree.

Observed live today. The hub showed `current_task: null` and the dashboard read "No work items in flight" / contributor **idle**, with #5227 back at position 2 of the ready queue, while the agent's tmux pane was mid-implementation on #5227 (worktree created, config key added, scheduler wired, docs being edited).

This is distinct from #5151. There, a reconnect books a departure but the relay re-adopts within ~1s, so the task survives and the cost is feed churn. Here **re-adoption never happened** and the assignment was simply lost.

Blast radius: any contributor task that spans a reconnect. Silent — nothing reports an error; the hub just stops believing work is happening.

## Evidence (2026-08-31, hosted spoke `hosted-kubestellar-hive-ojv6`)

Activity feed, verbatim:

```
17:11:45Z  picked up   issue kubestellar/hive#5227: ✨ scheduler: thread a per-repo checkout root …
17:19:24Z  joined      -
```

State immediately after, from three sources at the same moment (~17:22Z):

| Source | Says |
|---|---|
| `/api/contribute/fleet` | `current_task: null`, `connected_at: 17:19:24Z`, `last_activity: 17:22:24Z` |
| Dashboard "My work" | `0 items — No work items in flight`; contributor badge **idle** (0 working, 1 idle) |
| Ready-work queue | **#5227 listed at position 2, available** |
| Agent tmux pane | actively working #5227 — `esc to interrupt`, `1 shell` running, mid-edit on scheduler wiring and docs |

The relay process never died — same PID across the event (only the TCP source port changed, 34160 → 48428), so this is a socket-level reconnect, not a restart. The hub is still receiving heartbeats: `last_activity` advances normally. It knows the contributor is alive; it has simply forgotten what it is doing.

## The reconnect that lost it had an anomalous signature

Every earlier flap today produced a full triplet:

```
16:41:25Z  left
16:41:25Z  released: connection lost   issue …#5227
16:41:44Z  joined
```

[...]

PR #7762, opened 2026-09-19
✨ feature(contribute): report omp's primary and advisor model+effort in fleet, run rows and the PR trailer

Danathar:
## Summary

An omp contributor running a primary model **and** an advisor (`--advisor`) appeared everywhere in hive as `omp` with `model: null` — fleet view, run rows, activity rail, and the `— hive:` trailer on the PRs it opened. Two models did the work; hive named neither. omp picks its models from its own config rather than a flag, so contributors never export `AGENT_MODEL`, the relay had no omp detector, and even with one the hub had exactly one model slot per connection.

Fixes #7760 — the relay detector, the hub plumbing, and the display, end to end.

## Relay (`bin/contributor-relay.js`)

`detectOmpSelection()` reads what omp already records: `~/.omp/agent/config.yml` (`modelRoles.default`, `modelRoles.advisor`, `advisor.enabled`), the newest session's `model_change` record, and the advisor's `__advisor.jsonl` sidecar. Each `provider/model:level` selection is split into a model (provider kept inside, as for pi) and an effort; only omp's known thinking levels are split off, so an Ollama tag like `llama3:8b` stays part of the model name.

- Primary: the existing precedence `AGENT_MODEL` → detected → `''`; effort follows the same rule with `AGENT_REASONING_EFFORT`.
- Advisor: no env var exists, so it is always what omp records; reported as `advisor_model` / `advisor_reasoning_effort` on `auth_response` and every `task_progress`. Additive and omitted when empty — a bare omp, or any single-model backend, reports exactly what it did before.
- Refreshed on the same tick as the model, so a mid-task `/model` switch is reflected. The sidecar is never a candidate for the primary (it would otherwise win the mtime race).

## Hub

The pair lands on the connection and the profile at auth (sanitized like every stored contributor string; **not** checked against the accepted-models list — it reviewed the work, it did not do it), refreshes on `task_progress` under the same non-empty-overwrites rule as the model, and is carried to:

- `FleetClanker` → `/api/contribute/fleet` (`advisor_model`, `advisor_effort`)
- every `TaskRunRecord` → `/api/contribute/runs` (completed, failed, abandoned)
- `ActivityEntry` → `/api/contribute/activity`, via an optional trailing `advisorInfo` argument to `addActivity`, so the many call sites with no connection at hand are untouched
- `InvocationMeta` → the trailer and audit detail render `advisor=<model> advisor_effort=<v>` after the primary's effort; omitted when unknown, so every existing trailer is byte-identical:

[...]

PR #7781, opened 2026-09-19
🐛 fix(contribute): hold an issue for as long as its lease can be resumed

### LINKED ISSUE #7773: 🐛 bug(contribute): a 10–30 min contributor outage double-assigns its issue — the release cooldown (10m) lapses before the resume lease (30m) expires
**A contributor whose connection drops for between 10 and 30 minutes can have its in-flight issue handed to a second contributor while it is still able to resume — both then work the same issue and open duplicate PRs.**

Found by code audit of `v4` at `ff0b345` (v4.67.0) and reproduced with a Go test against the real hub over local sockets. Treat the production impact as the consequence the code implies.

## What happens

Two independent clocks are supposed to keep one issue with one contributor:

- **The release cooldown.** When a socket drops, `releaseOnDisconnect` clears the in-flight task and books a short cooldown so the issue is not instantly re-offered during the reconnect window. That cooldown is `failedTaskCooldownMinutes` = **10 minutes** (`contribute_cooldowns.go:19`, booked by `bookReleaseCooldown` at `contribute_ws.go:1870`).
- **The lease.** The disconnect deliberately keeps the server-issued lease so the returning relay can resume its task. The lease stays re-adoptable for `leaseTTL` = `wsTaskTimeout` = **30 minutes**, measured from the last progress report (`contribute_leases.go:64`).

The two windows are different lengths, and nothing else covers the gap between them. `selectTask` excludes an issue from re-offer only if a **live connection** currently holds it (`contribute_select.go:337`) or if it is in cooldown; a lease is not consulted as a hold outside the brief 2-minute post-restart grace. So between **10 and 30 minutes** after the last progress report from a disconnected relay:

- the relay is gone, so the issue is not in the live in-flight set,
- the 10-minute release cooldown has lapsed, so the issue is not excluded by cooldown,
- but the 30-minute lease is still alive, so the moment the relay reconnects it re-adopts the task.

If a second contributor asks for work during that window, `selectTask` offers them the same issue. When the original relay reconnects (it keeps its task locally and re-asserts it on `auth_ok`, `bin/contributor-relay.js:4622`), `lookupLease` matches and it resumes — `resumeTaskToken` even mints it a fresh push credential. Now two contributors hold the same issue with valid tokens.

## When it bites

The design note for #5322 reasons that the release cooldown "comfortably outlasts the reconnect backoff," which is true for a normal flap: the relay reconnects within its 1–60s backoff, well inside the 10-minute cooldown, and repopulates the live in-flight set before it lapses. The gap opens for a **medium-length** outage — 10 to 30 minutes — that the relay survives but does not reconnect through quickly:

1. Contributor A is assigned issue X; last progress report at t=0.
2. A's machine sleeps, or its network/VPN drops, at t≈2 (the relay process stays alive; its CLI keeps the task locally).
3. At t=10 the release cooldown lapses.
4. At t=13 contributor B sends `ready` → `selectTask` offers X to B (not in cooldown, no live holder).
5. At t=15 A's relay reconnects, re-asserts X, `lookupLease` matches (lease alive until t=30), A resumes with a fresh token.
6. A and B both work X and both open a PR.

## Reproduction

`TestAudit_ResumeAfterReleaseCooldownLapsesDoubleAssigns` (run against the real handlers via `go test -overlay`) drives exactly this ordering and asserts both connections end up holding X. It passes = the double-assignment reproduces.

## Why it happens

The disconnect hedge (#2356) and the resumable lease (#4260/#5681) were designed against different timescales — one to bridge the reconnect backoff, the other to let a long-running task survive a drop — and their windows were never reconciled. A lease that is still re-adoptable is, in effect, a hold; but only the live-connection scan and the cooldown are consulted when deciding whether to re-offer an issue, so a leased-but-not-live issue falls through.

## What "fixed" looks like

An issue that is still re-adoptable under a live lease is not offered to another contributor. Either the release cooldown lasts at least as long as the lease remains re-adoptable, or `selectTask` treats an unexpired lease held by a not-currently-connected identity as an in-flight exclusion (the way the post-restart grace already treats a restored lease), so the offerable window and the resumable window cannot overlap.

## Impact

Duplicate PRs and wasted contributor work — the exact failure #2356 exists to prevent — for any outage between the release-cooldown length and the lease TTL. The hub log shows both a "task assigned" (to B) and a "task resumed" (by A) for the same issue key.

A hub test: assign an issue, disconnect, advance the clock past the release cooldown but within the lease TTL, have a second contributor `ready`, and assert the issue is not re-offered while the lease is live; then reconnect the first relay and assert it still resumes.

## Affected components

- `contribute_ws.go`: `releaseOnDisconnect` (~1780), the `bookReleaseCooldown` call (~1870).
- `contribute_cooldowns.go`: `failedTaskCooldownMinutes` (19), the failure-cooldown check.
- `contribute_leases.go`: `leaseTTL` (64), `lookupLease`.
- `contribute_select.go`: the `activeIssues` live-connection scan (~337) that is the only in-flight exclusion.

Danathar:
## Summary

A contributor whose connection dropped for between ten and thirty minutes could have its in-flight issue handed to a second contributor while it was still able to resume it. When it came back, `lookupLease` matched, `resumeTaskToken` minted it a fresh credential, and two contributors held the same issue with valid tokens — duplicate PRs, the exact failure #2356 exists to prevent.

Fixes #7773

## Cause

Two clocks were meant to keep one issue with one contributor after a socket dropped, and they were never reconciled: #2356's release cooldown kept the issue out of `selectTask` for `failedTaskCooldownMinutes` (10 min), while the lease kept it re-adoptable for `leaseTTL` (30 min from the last progress report). Nothing covered the twenty minutes between — the issue was out of the live-connection scan, out of cooldown, and still resumable. #5322's "comfortably outlasts the reconnect backoff" was true for a flap and false for a laptop sleeping or a VPN dropping.

## Fix

The issue offers two shapes; this takes the second: **an unexpired lease is an in-flight exclusion in `selectTask` for every identity but its holder, for exactly as long as it is re-adoptable.** The moment the task is released (`task_complete`, `task_failed`, ready-abandon, operator requeue, the wedged-task backstop) `revokeLease` removes it; the moment it expires `pruneExpiredLeases` drops it. #5681 had applied this rule to *restored* leases during a two-minute post-restart grace; `leaseHoldGraceAfterStart` is gone and it is now the rule for every lease (`restored` and `startedAt` become informational).

- The requester's own leases are still never an exclusion — asking for work is the statement that it holds nothing. That leaves a narrower same-account case (one identity, several connections, one of them mid-outage) that this does not close; the comment records it.
- The cost #5681 weighed is real and accepted: a relay that never returns parks its issue for the lease's length rather than the ten-minute hedge. It is the price of the invariant the issue asks for — the same item is never offerable to one contributor while resumable by another.

## Tests

- `TestLeaseHold_OutageBetweenCooldownAndLeaseDoesNotDoubleAssign` — the issue's ordering over the real protocol: assign, drop, age the release hedge past ten minutes with the lease still live, a second contributor's `ready` gets `task_unavailable` (it got `task_assign` before), the first relay resumes without a revoke, and `task_complete` clears the hold at once.
- `TestLeaseHold_ExpiredLeaseReleasesTheIssue` — the other edge: an expired lease holds nothing and the issue is re-offered, while a live lease minted in this process still holds its item.
- `TestLeaseRestart_HoldLapsesAfterGrace`, which pinned the two-minute grace, becomes `TestLeaseRestart_HoldLastsWhileResumable`.
- All three fail on the parent (`5c9ab1719`).
- `TestSelectTask_DeclaredCapabilitiesDoNotAffectSelection` made two sequential picks for two contributors and expected the *same* item back — the very double assignment now prevented — so it releases the first pick before the second asks, which is the question it was actually asking (does *declaring* change the pick).

```
```

Docs: the lease-persistence section of `src/docs/contributor-relay.md` describes the new hold rule; citation and link checkers pass. Changelog fragment included.

## Relationship to my other open lease PRs

#7780 (one lease per task) and #7772 (external items keep their key on resume) touch neighbouring lines in `contribute_leases.go` / `contribute_ws.go`. Each applies to `v4` independently; whichever lands second will need a trivial rebase. If #7780 lands first, this rule automatically covers each of an identity's leases individually.


Danathar:
/hold

Danathar:
**Rebased onto #7780 — this PR now depends on #7780 and must merge after it.**

Both PRs change `leasedIssueKeys` in `contribute_leases.go` and conflicted. Worse than the textual conflict: #7780 keys the lease map by `identity\x1ftaskID`, and this branch compared the **raw map key** to the requester identity (`identity == exceptIdentity`), so once #7780 landed the requester's own leases would never have been excluded — a requester would have been blocked from re-taking its own item. Resolved to the combined form:

```go
for _, l := range h.leases {
    if l == nil || l.identity == exceptIdentity { continue }   // per-task key (#7780), every lease a hold (#7773)
    …
```

with the post-restart grace window and the `restored` gate removed as this PR intends. Two follow-up test-only commits: the hold-gap test now looks leases up with `leaseForLocked(identity, taskID)` instead of indexing by bare identity (which matched nothing / nil-dereferenced under #7780), and one fixed sleep became `testutil.Eventually` for the sleep-ratchet gate.

Verified on this head: `go vet`, full `go test -race ./pkg/dashboard/`, and `internal/testutil` all green. History is `#7780 → #7781 → test fixes`; force-pushed with a lease.

clubanderson:
Merged your other seven (#7772 #7780 #7785 #7786 #7787 #7788 #7792) plus #7762 — this one now conflicts with #7780 in `src/pkg/dashboard/contribute_leases.go`: both branches add `leasedIssueKeys`, and #7780's version carries an extra `!l.restored` skip (with its comment noting the post-#7774 grace-window widening) that this branch's version lacks. Since the difference is semantic, could you rebase onto v4 and pick the intended condition? I'll merge as soon as it's green.

Danathar:
Rebased onto v4 (`bee1815f`) — the branch is now just this PR's three commits on top of the merged #7780, and it applied cleanly; head is `a851d097`.

On the semantic question: **the intended condition is the one without the `!l.restored` skip.** That skip was the pre-#7773 rule — hold an item only during the two-minute post-restart grace window for leases read back from disk (#5681). This PR's whole point is to generalise that rule: *every* unexpired lease held by someone other than the requester is a hold, whether the lease was minted in this process or restored at boot, because the double-assignment gap (#7773) opens for any dropped socket, not just a restarted hub. So `restored` no longer gates anything here; it stays on the struct as a diagnostic (the comment on the field says so). The combined form that landed:

```go
for _, l := range h.leases {
    if l == nil || l.identity == exceptIdentity { continue } // per-task key (#7780), every lease a hold (#7773)
    if l.expiresAt.IsZero() || now.After(l.expiresAt) { continue }
    if l.key != "" { keys[l.key] = true }
}
```

`l.identity` (not the map key) is what's compared, since #7780 keys the map by `identity\x1ftaskID`.

Verified on the rebased head: `go vet ./pkg/dashboard/` clean, `go test -race -count=1 ./pkg/dashboard/` green (103.8s).

Danathar:
/unhold

Rebased onto v4 at a851d097 with the condition you asked about (no `!l.restored` skip, reasoning above), and every check is green. Releasing the hold so Tide can take it once it has your lgtm.

clubanderson (approved):
/lgtm
/approve

Green on the rebased head; merging.

Issue #7002, opened 2026-09-14
🐛 contributor quota: published readings never expire — a dead publisher admits forever on frozen numbers


clubanderson:
Found while verifying #7001 (the #6987 publishing-default-on change, merged `bab972ac`). **This is pre-existing and was not introduced by #6987 — but #6987 widens its blast radius from opt-in operators to every default install, which is why it is worth dispositioning now.**

## The relay defends against a state that no producer can emit

`bin/contributor-relay.js:661` holds fail-closed on a stale reading:

```js
if (state === 'unknown' || state === 'stale') return { admit: false, wait: true, reason: state };
```

Repo-wide, **nothing ever publishes `state: "stale"`.** The publisher emits exactly two states (`src/pkg/rotation/contributorreading.go`):

- `:149` → `ContributorReading{State: "unknown", Cause: ...}`
- `:164` → `ContributorReading{State: "available", Limits: limits}`

A repo-wide grep for `"stale"` outside tests returns only unrelated subsystems (`pkg/knowledge`, `pkg/hub`, `pkg/hub/spoke`). So the `stale` arm is **dead code**: a defense that is never armed.

## The reading carries no timestamp, so freshness is structurally unjudgeable

`ContributorReading` has no `captured_at`/`observed_at` field — the full set of JSON tags is `state`, `cause`, `limits`, and per-window `id`/`kind`/`pct_remaining`/`reset_epoch`/`resets_at`. The relay therefore **cannot** compute the age of a reading even if it wanted to.

I checked whether `reset_epoch` provides accidental self-healing. It does not: its only uses are override expiry (`continue_until_reset` at `:444`, `pause_until_reset` at `:462`) and diagnostics echo (`:691`). **An elapsed `reset_epoch` never invalidates the reading itself.**

## Consequence: the failure mode is fail-open, and it does not self-heal

A publisher that writes one healthy reading and then dies (crash, disabled rotation, revoked credentials, upgrade that drops the unit) leaves `state: "available"` with the last-known percentages on disk. The relay keeps reading that file and **keeps admitting forever against frozen numbers**, including long after every window it describes has rolled.

[...]

PR #7785, opened 2026-09-19
🐛 fix(contribute-ws): stop holding the selection lock across GitHub calls

### LINKED ISSUE #7775: 🐛 bug(contribute-ws): selectTask holds selectMu across two GitHub API calls — slow GitHub stalls dispatch fleet-wide and drops queued contributors on a false heartbeat timeout
**Every contributor's request for work is serialized behind a single lock that is held across two GitHub API calls. When GitHub is slow, contributor dispatch stalls fleet-wide, and a contributor that waits long enough is dropped for a heartbeat timeout — after which the hub still commits an assignment to the now-dead socket, booking a cooldown on an issue nobody worked.**

Found by code audit of `v4` at `ff0b345` (v4.67.0). The lock scope and the network calls inside it are confirmed by reading; the fleet-wide-stall impact follows from that and from GitHub latency, so treat the impact as conditional on a degraded GitHub.

## What happens

`selectTask` takes a single process-wide mutex for its whole body:

```go
// contribute_select.go
func (h *ContributeWSHub) selectTask(c *ContributorConnection) *WSMessage {
	h.selectMu.Lock()
	defer h.selectMu.Unlock()
	...
	ghToken, err := h.mintScopedToken(c.profile.TrustTier, chosen.repoFull) // line 809 — GitHub App token mint
	canPush := h.contributorCanPush(chosen.repoFull, ownUsername)            // line 836 — GitHub GetPermissionLevel
	...
}
```

Both calls hit GitHub synchronously while `selectMu` is held: `mintScopedToken` (App installation-token mint, ~8s timeout) and `contributorCanPush` (`Repositories.GetPermissionLevel`, 5s timeout, `repoPermissionTimeout`). `selectMu` is one mutex for the entire hub, so **every** contributor's `ready` is serialized behind these network round-trips. Under normal GitHub latency this is invisible; under GitHub degradation each `ready` holds the lock for many seconds and the whole fleet's dispatch queues up behind one contributor's GitHub calls.

The second half is worse than latency. `handleReady` runs on the connection's WebSocket **read goroutine** (the `HandleWS` read loop), and that same goroutine is the only place `lastPong` is updated (from `pong` frames / the pong handler). A contributor whose `ready` is queued behind several slow `selectTask` calls has its read goroutine blocked on `selectMu`, so it cannot read the hub's pong traffic. The heartbeat loop closes the socket after `wsHeartbeatTimeout` = 90s of no pong. So a contributor queued behind ~7 slow (≈13s) selections is hung up on. When its `selectTask` finally runs, it commits the assignment (records the lease, the rate-window slot, and a "picked up" activity entry), then `send`s to a socket that is already closing; the send fails, the handler returns, and `releaseOnDisconnect` books a release cooldown on an issue that contributor never actually received.

## Why it happens

The lock is meant to make offer→claim atomic (scan the candidates, pick one, commit it) so two contributors cannot be handed the same issue — and it does that correctly. But the two GitHub calls that decorate the chosen assignment (mint the scoped token, decide fork-vs-push) were placed inside the same critical section, so a lock that only needs to guard in-memory selection state now also serializes the fleet behind GitHub I/O.

## What "fixed" looks like

`selectMu` is held only around the in-memory selection and commit, not across GitHub network calls. The token mint and the push-permission lookup happen outside the lock (before it, cached, or after the item is claimed), so one contributor's slow GitHub round-trip cannot stall every other contributor's dispatch or starve a read goroutine into a heartbeat timeout.

## Impact

- Under a degraded GitHub, contributor dispatch is serialized fleet-wide behind GitHub latency even in the common case.
- A contributor queued long enough is dropped on a false heartbeat timeout, then handed an assignment that goes to a dead socket — which books a release cooldown on an issue nobody worked and leaves a lease to expire.

A test that injects a slow `mintScopedToken`/`contributorCanPush` and asserts that a second contributor's `ready` is not blocked for the duration (i.e. `selectMu` is not held across the network calls), and that a selection whose send fails does not leave a rate-window slot consumed for a task never delivered.

## Affected components

- `contribute_select.go`: `selectTask` `selectMu` scope (261), `mintScopedToken` call (809), `contributorCanPush` call (836).
- `contribute_ws.go`: `handleReady` on the read goroutine (~2186), `repoPermissionTimeout` (62), `wsHeartbeatTimeout` (33), the send-failure→`releaseOnDisconnect` path.

Danathar:
## What was wrong

When one contributor asked the hub for work, every other contributor's request had to wait until the first one's GitHub calls had finished. The hub picks tasks under a single fleet-wide lock so two contributors are never handed the same issue — and that lock was also held while the hub minted the contributor's scoped GitHub token and looked up whether it can push to the repo. Those are network calls that take seconds, or the full 5-second permission timeout when GitHub is slow, and every other `ready` queued behind them.

That queue ran on each connection's read goroutine, so a contributor waiting its turn also stopped reading heartbeat pongs. The heartbeat loop counted it as timed out and closed its socket. When its turn finally came, the hub committed an assignment to that dead socket, the send failed, and the disconnect path "released" a task the contributor had never received: a release cooldown on an issue nobody worked, a lease left to expire, and one of the contributor's hourly slots spent on nothing.

Fixes #7775.

## What changed

- **`selectTask` holds the lock only for the in-memory claim.** Scan candidates, pick one, commit it to the connection (task, lease, rate-window slot) — then release the lock. The committed claim is what every other selection's scan reads, so it *is* the reservation; the GitHub round-trips that decorate it (token mint, push-permission lookup) now run unlocked. Once they return, the prompt and pending token are attached only if the claim is still on the connection; if the socket dropped in the meantime and the disconnect path already released it, nothing is sent.
- **A claim that cannot ship is rolled back** (`rollbackAssignment` + `unrecordAssignment`): a failed token mint still returns the same `token_mint_failed` refusal, but first clears the task, revokes its lease and hands back the rate-window stamp; a failed `task_assign` send in the ready handler does the same before the disconnect path runs, so no cooldown is booked on an issue the contributor never saw.
- The `task_assign` wire shape, the double-assignment guarantee and the mint-failure refusal are unchanged.

## How it was verified

New `contribute_select_lock_scope_test.go`, all run against the parent to confirm they discriminate:

| test | on `v4` (`5c9ab1719`) | with this PR |
|---|---|---|
| a gated app-auth stub parks contributor A inside its token mint; contributor B's selection must complete meanwhile, with a *different* issue | B's selection blocked 2 s — **FAIL** | PASS (B assigned #7002 while A holds #7001) |
| a `task_assign` whose socket is already closed leaves no task / lease / rate slot, and `releaseOnDisconnect` then books no cooldown | claim left on the dead connection — **FAIL** | PASS |
| a claim released mid-mint sends nothing | `task_assign` returned for a released item — **FAIL** | PASS |
| a failed mint leaves no task / lease / slot / cooldown and the issue is re-offerable | PASS (mint ran before the claim) — guards the new ordering | PASS |
| `unrecordAssignment` removes exactly one stamp | n/a | PASS |

Full `go test ./pkg/dashboard/` is green. `gofmt`, `go vet` and `scripts/check-docs-citations.py` are clean; changelog fragment `changelog.d/fixed-7775-select-lock-scope.md` added.


Danathar:
/hold

Holding for review: part of tonight's audit-driven fix sweep; being checked against the issue it fixes before merge.

Issue #7732, opened 2026-09-19
🐛 bug(relay): task completion advertises `ready` twice — hub hands the freshly assigned task back and assigns another, relay rejects it


Danathar:
## Summary

After a task completes, the interactive relay advertises `ready` **twice** in the same synchronous turn: once from `finishCurrentTask()` and once from `armCLIReadyWait()`'s resolve handler. The hub answers the first with a new assignment, then reads the second as "relay asked for new work while still holding this task" (#2545 path), books the just-assigned task as `abandoned_handback`, and assigns a *third* task — which the relay rejects because it is already running the second one. Hub and relay now disagree about what the contributor is working on.

Observed live 2026-09-19 14:56Z, `just contribute-hive omp local`, hive `c73d9028b` (v4.66.x), omp 18.2.6. The cited lines are unchanged at current `v4` head `8df74e39a`.

## What the hub recorded

`/api/contribute/runs`, both rows stamped the **same second**:

```
2026-09-19T14:56:54Z projectbluefin/utah#18   completed  scenario=verdict_complete  signal=verdict verdict=no_work_needed
2026-09-19T14:56:54Z projectbluefin/utah#168  abandoned  scenario=abandoned_handback signal=null    verdict=null
```

`/api/contribute/fleet` immediately after: `current_task = projectbluefin/utah#175`.

## What the relay recorded (stdout, same window)

```
Task ct-projectbluefin/utah-18-1789829448 completed — signal=verdict (HIVE_VERDICT: no_work_needed)
Detected no_work_needed verdict for ct-projectbluefin/utah-18-1789829448: …
CLI ready — accepting tasks
Relaunching omp after a task exit: HIVE_WORKSPACE_DIR=$HOME/workspace GOCACHE=… omp
Task assigned: issue projectbluefin/utah#168 — ci: bump the pinned Bluefin parity ref on a schedule
Task prompt sent to CLI
Rejecting task projectbluefin/utah#175 from wss://…/api/contribute/ws — already working on projectbluefin/utah#168
```

The relay is now running #168 (the omp session file shows the #168 prompt as its second user message at 14:56:56Z) while the hub has #168 booked as handed back and shows the contributor on #175. Whatever #168 produces will be reported under a task the hub has already released.

## Mechanism (`bin/contributor-relay.js` @ `8df74e39a`)

`finishCurrentTask()` (L3456):

[...]

PR #7786, opened 2026-09-19
🐛 fix(relay): drop a queued task prompt when its task is revoked or fails

### LINKED ISSUE #7779: 🐛 bug(relay): a queued task prompt survives task_revoke and is typed into the agent later, with no task held
**A task prompt that was queued because the CLI was still starting survives a `task_revoke`, and is then typed into the agent after the CLI comes up — with no task held. The agent starts working an issue the hub has already taken back and given to someone else.**

Found by code audit of `v4` at `ff0b345` (v4.67.0) and reproduced by loading the relay in test mode.

## What happens

When a task prompt is sent while the CLI is not yet confirmed ready, `tmuxSendKeys` queues it in `pendingTask` instead of typing it:

```js
// bin/contributor-relay.js — tmuxSendKeys (~2083)
if (!cliReady) { pendingTask = text; return; }
```

`pendingTask` is cleared in exactly two places: the CLI-readiness **failure** handler (`armCLIReadyWait().catch`, ~1910), and `flushPendingTask` (~3256), which **types it into the CLI** once the CLI is ready. Nothing on the task-exit paths clears it: the `task_revoke` handler does not touch `pendingTask`, and neither does `failCurrentTask` (confirmed by search — both leave it set).

So this sequence types a revoked task's prompt into the agent:

1. A task is assigned while the CLI is relaunching → its prompt is queued in `pendingTask`.
2. Before the CLI is ready, the task is revoked (an operator yank, or a resume rejected by the hub). The revoke handler clears `currentTask` and relaunches the CLI, but leaves `pendingTask` set.
3. The CLI reaches its prompt → `armCLIReadyWait`'s success path calls `flushPendingTask` → the revoked task's prompt is typed into the agent.

The agent now works an issue that `currentTask` says it does not hold, unsupervised — while the hub has already revoked the lease and may have handed that issue to another contributor. When the next real `task_assign` arrives, its prompt is typed on top of the turn the agent is already running.

## Reproduction

Loaded under `HIVE_RELAY_TEST_MODE=1`: `pendingTask queued at assign = true; still queued after revoke = true; currentTask after revoke = null; revoked prompt typed into CLI later = true`.

## Why it happens

`pendingTask` is a prompt-delivery queue with only two lifecycle hooks — "CLI failed, drop it" and "CLI ready, type it." The task-exit paths (revoke, fail) that invalidate the task the prompt belongs to were never taught to invalidate the queued prompt as well, so a prompt outlives the task it was for.

## What "fixed" looks like

A queued prompt is discarded when the task it belongs to ends. The `task_revoke` and `failCurrentTask` paths clear `pendingTask` (ideally keyed to the task id, so only the queued prompt for the ended task is dropped), so `flushPendingTask` can never type a prompt for a task the relay no longer holds.

## Impact

After a yank or a hub-side resume rejection that lands while the CLI is relaunching, the agent silently works a revoked issue with no lease — the duplicate-PR / unsupervised-work shape — and the relay's own `currentTask` disagrees with what the agent is doing. Related to the revoke-path `ready` double-send flagged on #7732, but a distinct root cause (the prompt queue, not the ready count).

Queue a prompt while the CLI is not ready, deliver a `task_revoke`, bring the CLI ready, and assert `pendingTask` was dropped and nothing was typed into the pane.

## Affected components

- Repo `hivecommons/hive`, `bin/contributor-relay.js`: `pendingTask` queue in `tmuxSendKeys` (~2083), its only clears (`armCLIReadyWait().catch` ~1910, `flushPendingTask` ~3256), and the `task_revoke` handler / `failCurrentTask` that end a task without clearing it.

Danathar:
## What was wrong

The relay could type a task's prompt into the agent **after the hub had taken that task away**. When an assignment arrives while the agent CLI is still starting, the relay queues the prompt and delivers it once the CLI reaches its prompt. That queue had only two lifecycle hooks — "the CLI never came up, drop it" and "the CLI is ready, type it" — and nothing on the task-exit paths touched it.

So if the hub revoked the task (an operator yank, a rejected resume), or the relay itself failed the task, while the CLI was relaunching:

1. the revoke cleared `currentTask` and relaunched the CLI, but the prompt stayed queued;
2. the relaunch's readiness callback flushed the queue and typed the revoked task's prompt into the fresh CLI;
3. the agent went to work on an issue the hub had already taken back — and possibly handed to another contributor — with the relay's own `currentTask` saying it held nothing. The next real assignment's prompt then landed on top of that turn.

Fixes #7779.

## What changed

`bin/contributor-relay.js`:

- **The queue belongs to a task.** `queuePendingTask()` stamps the queued prompt with the id of the task it is for (`pendingTaskId`); the three places that used to write `pendingTask` directly go through it.
- **Every task-exit path discards it.** `task_revoke`, `failCurrentTask`, `finishCurrentTask` and the existing CLI-never-ready path all call `discardPendingTask()`, which logs why the prompt was dropped.
- **`flushPendingTask()` is the backstop.** It compares the queued prompt's owner with the task the relay currently holds and drops the prompt on a mismatch — no task held, or a different task — so a future exit path that forgets the queue cannot reintroduce the bug. Delivery for a task the relay still holds is unchanged, as is the single `ready` a revoke advertises (#7732).
- Test seam: `setPendingTask` stamps the owner from `currentTask` exactly as `queuePendingTask` does; `getPendingTaskId` added.

## How it was verified

New tests in `bin/contributor-relay.test.js`, run against the parent (`v4` at `5c9ab1719`) to confirm they discriminate:

| test | parent | this PR |
|---|---|---|
| the issue's reproduction: prompt queued while the CLI is not ready → `task_revoke` → CLI comes up; the queue must be empty and nothing typed | prompt still queued after the revoke — **FAIL** | PASS |
| the same with `failCurrentTask` | **FAIL** | PASS |
| `flushPendingTask` with no task held, and with a *different* task held, types nothing | prompt typed with no task held — **FAIL** | PASS |
| positive control: a still-held task's queued prompt is typed once the CLI is ready | PASS | PASS |

Full relay suite: **453/453** passed (`node bin/contributor-relay.test.js`). Changelog fragment `changelog.d/fixed-7779-queued-prompt-outlives-revoke.md` added.


Danathar:
/hold

Holding for review: part of tonight's audit-driven fix sweep; being checked against the issue it fixes before merge.

PR #7788, opened 2026-09-19
🐛 fix(relay): an unwritable TASK_FILE no longer crashes the relay on every task_assign

### LINKED ISSUE #7777: 🐛 bug(relay): an unwritable TASK_FILE crashes the relay on every task_assign — the write is unguarded unlike the token-cache write beside it
**If the relay cannot write its task file, it crashes on every task assignment — a crash loop, not a degraded mode. The write is unguarded, unlike the token-cache write right next to it that was deliberately wrapped for exactly this reason.**

Found by code audit of `v4` at `ff0b345` (v4.67.0) and reproduced by loading the relay in test mode.

## What happens

When a `task_assign` arrives, the relay writes the task metadata to `TASK_FILE` before it acknowledges the task:

```js
// bin/contributor-relay.js — task_assign handler (~4744)
const { github_token: _omittedToken, ...taskFileRecord } = msg;
fs.writeFileSync(TASK_FILE, JSON.stringify(taskFileRecord, null, 2), { mode: 0o600 });
try { fs.chmodSync(TASK_FILE, 0o600); } catch (_) {}
send({ type: 'task_accepted', seq: nextSeq(), task_id: msg.task_id, task_gen: msg.task_gen });
```

That `writeFileSync` is not wrapped in try/catch. If it throws — an unwritable directory, a full filesystem, a stale file owned by another user, a bad `HIVE_TASK_FILE` — the exception propagates out of `handleMessage`, out of the WebSocket `message` listener, and there is no `uncaughtException` handler, so the process dies. The relay reconnects, gets the same or another assignment, and dies again: a crash loop. `currentTask` is set and the token file is written by the time it throws, but `task_accepted` is never sent, so from the hub's side each cycle is an assignment that goes nowhere and books a disconnect-release cooldown on the issue.

The token-cache write immediately above this, in `injectGhToken`, **is** wrapped in try/catch, with a comment stating precisely why: "A failed write must never throw out of handleMessage: task_assign calls this before task_accepted is sent, so an unwritable cache path would crash the relay on every assignment that carries a token — a crash loop, not a degraded mode." The task-file write has the same shape and the same call site but not the same guard.

## Reproduction

Loaded under `HIVE_RELAY_TEST_MODE=1` with `HIVE_TASK_FILE` pointing at a non-existent directory: `task_assign threw = true (ENOENT … open '/nonexistent-dir-xyz/contributor-task.json'); currentTask set = true; task_accepted sent = false`.

## When it bites

The default `TASK_FILE` is `/tmp/contributor-task.json`. Triggers seen in ordinary operation: `/tmp` (a tmpfs) full, a stale `contributor-task.json` owned by a different uid on a shared host, or a misconfigured `HIVE_TASK_FILE`. Any of these turns every assignment into a crash.

## What "fixed" looks like

A failed task-file write degrades rather than crashes: wrap the write in try/catch and log loudly, exactly as `injectGhToken` already does, so a bad task-file path costs at most the task-file feature, not the whole relay. The relay should also not leave `currentTask` set / the token written if it is going to abort before `task_accepted`.

## Impact

A relay on a host with a full or unwritable `/tmp` (or a stale task file) crash-loops instead of working, and each cycle books a release cooldown on the assigned issue. The failure names a filesystem error, not "task file unwritable," so it reads as a broken relay rather than a config problem.

A test that points `HIVE_TASK_FILE` at an unwritable path and asserts `task_assign` is handled without throwing (the relay logs and continues, `task_accepted` still sent), mirroring the existing `injectGhToken` failure test.

## Affected components

- Repo `hivecommons/hive`, `bin/contributor-relay.js`: the `task_assign` handler's `fs.writeFileSync(TASK_FILE, …)` (~4744); contrast `injectGhToken` (~626) which guards the same failure.

Danathar:
**What a contributor sees today:** if the relay's task file cannot be written, the relay crashes the moment the hub assigns it work — and again on the next assignment, forever. The default path is `/tmp/contributor-task.json`, so a full `/tmp` (a tmpfs), a stale copy owned by a different user on a shared host, or a bad `HIVE_TASK_FILE` all trigger it. The log shows a filesystem error, not "task file unwritable," so it reads like a broken relay rather than a config problem. Each cycle also books a disconnect-release cooldown on the assigned issue hub-side.

**Why:** in the `task_assign` handler, `fs.writeFileSync(TASK_FILE, …)` was the one unguarded call. The exception left `handleMessage`; there is no `uncaughtException` handler; the process died with `currentTask` already set and the token already written but `task_accepted` never sent. The token-cache write a few lines above (`injectGhToken`) already guards exactly this failure, with a comment explaining the crash-loop risk — the task-file write just never got the same treatment.

**The change:** wrap the task-file write (and its chmod) in a try/catch that logs loudly and continues, mirroring `injectGhToken`. The task file is observability state no task depends on, so a failed write now costs that file, not the relay.

**Test:** `#7777 task_assign with an unwritable task file path does not crash the relay` — points `HIVE_TASK_FILE` at a path whose parent is a regular file, asserts `handleMessage` does not throw, `task_accepted` is still sent, and the failure is logged. It fails on the old code (449/450) and passes with the fix (450/450, CI's own `node bin/contributor-relay.test.js` invocation).

Fixes #7777

https://claude.ai/code/session_01VXwvRu29c2MkwftzkgAHcC


Danathar:
/hold

Holding to match the rest of tonight's audit-driven batch: waiting on a human maintainer's review before merge.

PR #7802, opened 2026-09-20
🐛 fix(dashboard): register API routes once during boot

hanthor:
The v5 boot path registers the dashboard API through the injected `registerAPI` dependency and then calls `RegisterAPI` directly a second time. On Go's `http.ServeMux`, the second registration panics at startup; a v4→v5 canary reproduced this on `/contribute` before the dashboard became ready.

This removes the duplicate direct call and adds a regression test that exercises the production registration callback, where any second registration would panic.

Validation:
- `go test ./cmd/hive -run 'TestBootDashboardAPIWith_(RegistersDependenciesFromTheBootStruct|RegistersRoutesExactlyOnce)' -count=1`
- upstream lint, security, unit, dashboard-shuffle, Docker, and multi-architecture build checks pass
- patched v5 image canaried successfully on Hanthor, Reef, and School; all public health endpoints are green with zero container restarts


hanthor:
The v5 base advanced during this PR and now already contains `changelog.d/fixed-7802-dashboard-double-register.md` (commit 9f136370f). I rebased the code fix onto that base and added `no-changelog` only to satisfy the guard; the user-visible fix is still covered by that existing fragment.

PR #7895, opened 2026-09-20
✨ feat(ops): show PRs produced in the last 24 hours on the "Your contribution" card

### LINKED ISSUE #7894: 🌱 Operations page: show PRs produced in the last 24 hours on the "Your contribution" card
**I want the "Your contribution" card to tell me how many PRs I produced in the last 24 hours, next to the all-time count it already shows.**

The card on the Operations page has four tiles: *Issues worked (24h)*, *Issues worked (total)*, *PRs produced* (with a "shipped no PR" sub-line), and *Failed*. Right now on my hive it reads 41 / 281 / 92 / 25.

Issues get both a 24-hour figure and an all-time figure. PRs only get the all-time figure. So I can see that I worked 41 issues today, but not how many of those turned into a pull request today, which is the number I actually care about when I check in on the fleet: the footnote under the card says itself that *PRs produced* "is what auto-promotion reads, not the bare completion count".

The footnote's last sentence ("The 24h figure is summed from the same hourly series as the sparklines") also reads as if it applies to PRs, but there is no 24h PR figure to apply it to.

## Why it is not there yet (what I think, from reading the code)

`GET /api/contribute/me` (`src/pkg/dashboard/api_contribute.go`, `handleContributeMe`) returns `tasks_completed_24h` by summing the caller's trailing 24 buckets of the hourly ring in `contribute_metrics.go`. That ring (`perUserDone`) is built each rollup from the delta of each profile's `TasksCompleted`. Nothing rolls up `TasksWithPR` the same way, so there is no per-hour series to sum for PRs; the card only has the cumulative `total_tasks_completed_with_pr` from the profile.

## What "fixed" looks like

- A fifth tile, or a sub-line on the existing *PRs produced* tile, showing PRs produced in the last 24h, with the same honesty rules the issues tile already has: an em dash and "no hourly history yet" when there is no series, and "Nh of history so far" when fewer than 24 buckets back it.
- `/api/contribute/me` gains `prs_produced_24h` (name open), derived from a second per-user ring rolled up from `TasksWithPR` deltas exactly like `perUserDone` is from `TasksCompleted`, so it stays consistent with the sparklines and survives restarts the same way.
- The footnote updated so the "24h figure" sentence is true of both tiles.

## Open questions

- Should the leaderboard's per-contributor series get the same PR ring, so the sparkline could later show PRs rather than completions? Not needed for this, but the rollup change is the same one.
- The new ring adds one `map[string][]int` to the persisted metrics file. Fine for the hives I run; worth a look at what that costs on a large hub.

Danathar:
## Summary

The "Your contribution" card on the Operations page told you how many issues you worked in the last 24 hours and all-time, but for pull requests it only had the all-time number. So you could see that you worked 41 issues today without knowing how many of those actually became a PR today — which is the number that matters, since PRs produced is what auto-promotion reads. The card's own footnote ("the 24h figure is summed from the same hourly series as the sparklines") also read as if it covered PRs when there was no 24h PR figure at all.

The card now shows **PRs produced (24h)** as a fifth tile, next to **PRs produced (total)**, with the same honesty rules the issues tile already has: an em dash and "no hourly history yet" until the spoke has a PR series for you, and "Nh of history so far" while fewer than 24 hours back it. The footnote now speaks for both 24h figures.

Fixes #7894

## What changed

- **Metrics store** (`contribute_metrics.go`): a second per-contributor hourly ring, `per_user_pr`, rolled up from the delta of each profile's verified-PR counter exactly the way the completion ring is rolled up from the completion counter — same first-tick baseline seed, same clamp on a counter that went down, same zero for an idle registered contributor, same pruning of a departed contributor with an all-zero ring, same 168-bucket cap. The rollup's delta and per-user append steps are factored into two small helpers so both rings go through one code path. The ring is persisted in the same metrics file so the figure survives restarts.
- **`GET /api/contribute/me`**: gains `prs_produced_24h`, `pr_window_hours_covered` and `pr_history_available`, summed from that ring. The PR figure carries its own availability/coverage pair instead of borrowing the completion series' — see the one design difference below.
- **Card** (`contribute_landing.go`): the fifth tile; the "number or em dash plus sub-line" logic is factored into one helper shared by both 24h tiles so they cannot drift in what they call honest; the footnote is updated. The all-time PR tile is relabelled "PRs produced (total)" to mirror the issues pair.
- **Docs**: `openapi.json` gets the three new fields; the `api-reference.md` row mentions the 24h PR figure; a `changelog.d/added-…` fragment.

### One deliberate difference from the completion ring

The PR ring is **tail-aligned, not left-padded** to the shared timeline. The completion ring pads a late-starting ring with leading zeros so the sparkline can draw index *i* as the same hour in every series. Nothing draws the PR ring — it is not served by `/api/contribute/metrics` (so the per-poll sparkline payload does not grow on a large hub, one of the issue's open questions) and it is only ever summed from its tail — so head alignment buys nothing, while leaving it unpadded keeps its length equal to the hours actually measured.

That matters for the upgrade that introduces the ring. With padding, the first post-upgrade rollup would give every contributor a full day of zeros and the tile would print, say, "3" labelled as 24h with no hint, when the real count might be ten. Without padding, a spoke with a day of completion history and three post-upgrade hours reports "3h of history so far" on the PR tile, and a metrics file from before this change (no `per_user_pr` key) loads as "no PR history yet" rather than as zeros. Both cases are pinned by tests.

## Testing

`go build ./... && go vet ./pkg/dashboard/` clean; `gofmt -l` reports nothing for the touched files.

New tests (all pass; the branch fails to compile without the change, and switching the PR ring to padded makes `TestPRRingIsTailAlignedNotPadded` and `TestContributeMePRsShortHistoryIsHonest` fail — checked):

- `contribute_metrics_pr_ring_test.go` — PR buckets are the verified-PR delta (not the completion delta, not the cumulative total; a counter drop books zero); trailing-window sum with the same known/covered contract; persisted and restored across a restart; a legacy file without `per_user_pr` yields "unknown", not zeros; tail alignment after an upgrade (28-bucket completion ring, 3-bucket PR ring, coverage 3); idle contributors get explicit zeros and departed all-zero rings are dropped; the 168-bucket cap; `/api/contribute/metrics` does not serve the ring.
- `contribute_me_stats_test.go` — `prs_produced_24h` sums the caller's own PR ring and stays distinct from `tasks_completed_24h`; short PR history reports its real depth while the completion window reports a full day; no PR rollups yet reports `pr_history_available: false` with 0/0 even when the completion series exists; the lookup keys off the stored username case; the rendered Operations page carries the new tile, helper, fields and footnote.

```
$ go test ./pkg/dashboard/ -count=1
ok  	github.com/hivecommons/hive/pkg/dashboard	56.706s
$ scripts/check-api-reference-citations.sh
api-reference citations: all pkg/<file>:<line> references verified.
$ python3 scripts/check-docs-citations.py
docs citations: 879 repo-local file:line reference(s) verified.
```

Not run: a browser render of the card (no dashboard in this sandbox); the tile markup goes through the existing `ccMineTile` helper and the auto-fit grid, and the page-level test pins the rendered strings.


clubanderson (approved):
/lgtm
/approve

Green on the rebased head; merging.

Issue #7721, opened 2026-09-19
📝 v5 migration plan: move hives to v5 next week; re-base stable/candidate on v5 and edge on v6


clubanderson:
> **Deviation record (2026-09-21).** Hub Admin decision on 2026-09-21 superseded the slip rule below: Phase 1/3 were executed as an emergency cut-over (#8058/#8059/#8060, promote run 35632232764 with `soak-hours=0`) *before* the Phase 0 announcement was posted. The pre-migration `v5-migration-announcement.md` was rewritten as a post-hoc operator notice in #8105; #8062 carries the post-hoc soak close-conditions. Declaration: https://github.com/hivecommons/hive/issues/6016#issuecomment-5766618218. Filed via #8101.

## Goal

Starting the week of **2026-09-22**, move hives from the v4 line to v5, and re-base the three release channels so that **`stable` and `candidate` are built from `v5` and `edge` is built from `v6`**. v4 then enters the accepted feature freeze (#6346, mechanics in #7719) and serves security/critical fixes only until EOL.

This issue is the single plan + tracker for that move. It links, rather than repeats, the GA bar (#6016), the freeze runbook (`src/docs/v4-freeze-runbook.md`), and the channel docs (`src/docs/release-channels.md`). Check a row only with a linked evidence artifact.

## Where we are today

| Channel | Built from | Mechanism |
|---|---|---|
| `stable` | v4 (promoted by digest after soak) | `promote-stable.yml` (`RELEASE_BRANCH: v4`, 24h default soak) |
| `candidate` | v4 (every green merge) | `docker.yml` on v4: `publish-image-tags.sh … v4 true candidate` |
| `edge` | v5 (every green merge) | `docker.yml` on v5: `publish-image-tags.sh … v5 false edge` |
| — | v6 | `docker.yml` on v6 still carries the v5 table; `v6` is **not** in `LONG_LIVED`, so no `v6-latest` and no channel is published |

`latest` follows v4 (`INCLUDE_LATEST=true` only on the v4 row). GA bar (#6016) Release-train rows still open: `edge` digest-pinned to the GA candidate; dual v4/v5 hub-spoke validation (#6112 format).

## What is new in v5 (operator-facing)

297 commits on `v5` that are not on `v4`. The ones an operator will notice, grouped:

**Agent governance & review**
[...]

PR #7904, opened 2026-09-20
🐛 fix(ci): v6 must not publish hive-contributor/hive-hub :latest

clubanderson:
Closes #7893 (v6 half; v5 counterpart is the sibling PR)

`.github/workflows/docker.yml` L688 (`hive-contributor`) and L879 (`hive-hub`): `v5 true edge` → `v5 false edge`, matching the `hive` row (L482). `publish-image-tags.sh` pushes `:latest` whenever `INCLUDE_LATEST=true` regardless of the branch guard, so these two images' `:latest` flapped between v4/v5/v6 on every merge while `hive:latest` stayed v4 — a mixed-version triplet for anyone pulling `:latest`, and a contradiction of #7721's "latest follows v4" premise two days before migration week.

Same one-word edits on v6, whose rows still carry `v5 … edge` and will be rewritten to `v6` in #7721 Phase 1. Workflow-only + changelog fragment. Pushed with maintainer credentials since the contributor-tier token cannot touch `.github/workflows/`.


clubanderson:
Follow-up commit: the cross-org mirror steps carry a lockstep `INCLUDE_LATEST` env (L729/L920) that I missed — flipped to `"false"` too, otherwise the mirror would still advance `:latest`. `test-publish-image-tags.sh` pinned the old `v5 true edge` row; it now asserts 3× `v5 false edge` and 3× `INCLUDE_LATEST: "false"` and fails on any `true`. Self-test passes locally.

Issue #7879, opened 2026-09-20
🐛 bug(relay): #7759 drops advisor notes that land after the final verdict — they are on the pane at finalization and never recorded; also allow one more follow-up only for a new ⟦blocker⟧ (hard cap 2)


Danathar:
Follow-up to #7759 / #7806, from a live case where the bound worked as specified and still lost a real (minor) finding.

## What happened (projectbluefin/contribute#638 → PR #641, omp Sonnet 5 + Opus 5 xhigh advisor, relay `92e00565d`)

| time (UTC) | event |
|---|---|
| ~17:52 | first `HIVE_VERDICT: complete` |
| 17:53:40 | relay: `1 advisor concern(s) … asking the agent once` (JSONC in renovate.json — valid, fixed, regression test added) |
| ~17:57:44 | second `HIVE_VERDICT: complete` |
| **17:58:00** | advisor posts a *new* concern: the PR body still says `gitAuthor` is rejected by the validator — false (only `username` is self-hosted-only). Prose only; code correct. |
| 17:59:40 | relay finalizes on the second verdict — the note has been on the pane for 100 s, but `postVerdictReviewRequested` is already true, so the relay never looks |
| 17:59:50 | relaunch, next task |

The advisor's own closing note: *"One item I raised remains unaddressed and is worth recording for whoever reviews PR #641."* Nothing recorded it. The inaccuracy is now in the PR body for a maintainer to catch or miss.

This is the accepted trade-off of #7759 ("second verdict is final") doing its job, so the fix must not reopen the loop. Two additive layers:

## 1. Record instead of drop (no extra turns — cannot regress)

At finalization, if `⟦concern⟧`/`⟦blocker⟧` notes are on the pane that post-date the verdict being finalized (same "new since" test `maybeRequestPostVerdictReview` already uses), the relay:

- appends them to the `task_complete` `summary` under a heading like *"Advisor notes not addressed before completion"*, and
- when a PR was detected for the task, posts them as one PR comment (same task credential the agent used; one comment, never edited later), so the reviewer sees exactly what the advisor saw.

The information already exists on the pane; today it dies with the relaunch.

## 2. One more follow-up, only for a NEW `⟦blocker⟧`, hard cap 2

[...]

Issue #7862, opened 2026-09-20
🐛 bug(contribute): an agent can claim `HIVE_VERDICT: complete — PR opened` with zero commits and the hub books a completed task — the PR-less verdict is never checked against the claim


Danathar:
Observed 2026-09-20 14:37Z, omp 18.2.6 with `gemini-3.8-flash` (high) primary + `claude-opus-5` (high) advisor, local mode, relay `92e00565d`, hub `273aa02`, task projectbluefin/chairlift#137 ("Add rollback to chairlift").

## What happened

2 min 40 s after the prompt, the agent printed:

```
 HIVE_VERDICT: complete — PR opened
```

No PR exists. The omp session transcript for the task has **30 tool calls, all read-only** — `read`, `grep`, `glob`, `git log`, `git show`, `gh issue list`, `gh pr list --search` — and **zero** `git checkout -b`, `git commit`, `git push`, `gh pr create`. The checkout was never touched (still on the previous task's branch). The contributor's newest chairlift PR predates the task by an hour.

Relay:

```
Task ct-projectbluefin/chairlift-137-…: HIVE_VERDICT: complete is on the pane — running the progress tick now … (#7841)
Task ct-projectbluefin/chairlift-137-… completed — signal=verdict (HIVE_VERDICT: complete)
```

(no `Detected PR` line — `resolveTaskPR` correctly found nothing). Hub run row:

```
2026-09-20T14:37:56Z projectbluefin/chairlift#137 completed / verdict_complete  verdict=idle  model=google-antigravity/gemini-3.8-flash
```

## Why the ledger accepts it

- Relay: the verdict text and the PR scan are never compared. `resolveTaskPR` returning empty just means `pr_url` is omitted from `task_complete`; the claim in the reason ("PR opened") is passed through as prose.
- Hub (`contribute_ledgers.go`): `isEvidenceLessCompletion(prURL, verdict, signal)` is true only when `signal == chrome_idle`. A PR-less `complete` that arrived **with a verdict** is treated as a normal completion: `handleTaskComplete` does `TasksCompleted++` (`contribute_ws.go` ~L2760; only `TasksWithPR` is gated on a verified PR), and `markTaskCompletedVerdictKeySignal` takes the `advanceNoPRStreakLocked` branch — a real completion cooldown on the issue rather than the short `completedNoPRCooldownHours` reserved for evidence-less finishes.

[...]

Issue #7841, opened 2026-09-20
🐛 bug(contribute): claude keeps working after HIVE_VERDICT — a finished background shell re-wakes the agent, and the relay waits up to 120 s to act on a verdict already on the pane


Danathar:
## Summary

Claude Code re-enters the agent when a background shell it started completes. If the agent printed `HIVE_VERDICT` while such shells were still running, the completion notification wakes it for another turn *after* the sentinel — the same "keeps working past its own verdict" shape as omp's todo-reminder in #7733, from a different runtime feature. The relay finalizes on the verdict at its next tick (≤ `PROGRESS_REPORT_INTERVAL_MS` = 120 s) and kills the CLI, so the booking is right; but for up to two minutes the agent runs on, and anything it does in that window (a follow-up push to the PR it just reported, a comment) happens after the task has been credited.

Observed 2026-09-20 11:31Z, claude backend (Claude Code 2.1.267) in container mode, relay `60be2c93f`, task projectbluefin/review#187.

## Pane

Verdict, with three background shells still alive:

```
  HIVE_VERDICT: complete — PR #635 opened against main fixing the dead advisory-reviewer tool names found under #187, with an evidenced finding posted …
✻ Brewed for 7m 0s · 3 shells still running
❯ check if CI passed
  ⏵⏵ bypass permissions on · 3 shells · ← for agents · ↓ to manage
```

Twenty seconds later, no input from anyone, the agent is mid-turn again:

```
  ⎿  $ cat …/tasks/bg1ksercw.output …/tasks/bg20yjb48.output 2>/dev/null; echo "[end]"
* Creating… (9s · ↓ 506 tokens · thought for 2s)
❯
  ⏵⏵ bypass permissions on · 2 shells · esc to interrupt · ← for agents · ↓ to manage
```

One of the three shells had finished; Claude Code delivered its output as a task notification and the model responded. 2 of 5 tasks this run ended with `N shells still running` on the verdict line (`1 shell` on utah-packages#212, `3 shells` here).

## Why the shells exist

In this run they were polls for CI / a review bot that the agent started because of instructions staged in from the operator's `~/.claude/CLAUDE.md` (#7836). But the mechanism is general: any `run_in_background` shell the agent leaves alive at the sentinel re-wakes it when it exits, and long-running polls are exactly the kind of thing an agent backgrounds before declaring itself done.

## Proposal (either or both)

[...]

Issue #7759, opened 2026-09-19
✨ feature(contribute): let an omp agent address advisor notes that land after HIVE_VERDICT before the relay finalizes


Danathar:
## Summary

omp's `--advisor` runtime "passively reviews each turn and injects notes" — so its review of the agent's *final* turn lands on the pane **after** `HIVE_VERDICT` has been printed. With #7733 fixed, the relay now correctly finalizes on the sentinel and kills the CLI, which means every advisor note on the closing turn is discarded unread. Two live tasks today each ended with a stack of `⟦concern⟧`/`⟦nit⟧` notes under the verdict; at least one was a real, cheap fix the agent would have made if it had seen it.

Observed 2026-09-19 on `just contribute-hive omp local`, hive `273aa0280`, omp 18.2.6 with `--advisor` (advisor model: claude-opus-5).

## What the pane looked like at the end of `projectbluefin/utah#14` (PR #198 opened)

```
 HIVE_VERDICT: complete — PR #198 delivers gated production ISO artifact retention against main.
 ⓘ Advisor 1 note
   ▎ ⟦concern⟧ Base branch: repo has no CONTRIBUTING/PR template, but .github/workflows contains
   ▎ promote-testing-to-main.yml and sync-main-to-testing.yml — a promotion model where `testing` is the
   ▎ integration branch and `main` is released. Branch from upstream/testing and open the PR against `testing`, …
 ⓘ Advisor 1 note
   ▎ ⟦concern⟧ Working tree is dirty from a prior task (M docs/skills/kernel-cache.md,
   ▎ scripts/verify-rpm-contract.py, tests/test_verify_rpm_contract.py) … can leak into your commit …
 ⓘ Advisor 1 note
   ▎ ⟦nit⟧ Set `compression-level: 0` on the Retain production ISO upload — upload-artifact zips at level 6 by
   ▎ default, and an ISO that is already a zstd squashfs will burn many minutes of CPU per flavor for near-zero
   ▎ size gain.
```

The relay completed the task on the verdict line (correctly), relaunched omp, and the notes were gone. Scoring them after the fact:

- the `compression-level: 0` nit is right and unaddressed — PR #198's `upload-artifact@…v7` step has no `compression-level`, so CI will re-zip a squashfs ISO per flavor;
- the dirty-tree concern turned out moot (the PR contains only its own three files);
- the base-branch concern is debatable (the promotion workflows exist, but 11 of the last 12 merged utah PRs target `main`).

[...]

PR #7910, opened 2026-09-20
🐛 fix(relay): judge HIVE_VERDICT on a settled capture, not the one that found it

### LINKED ISSUE #7907: 🐛 bug(relay): #7886 note collection races omp's end-of-turn advisor delivery — verdict fast path finalizes ~1 s before the notes render, so they are dropped again (utah#218)
## What happened

Same contributor and build as the #7886 verification (omp container mode, Sonnet 5 medium + Opus 5 xhigh advisor, relay from the `:latest` image carrying #7886, hub `37522cf`).

`projectbluefin/utah#169` → PR projectbluefin/utah#218. After the one #7759 follow-up the agent printed its second `HIVE_VERDICT: complete`. Two `⟦concern⟧` notes were on the pane directly under that verdict — one of them saying *"Skip the rebase. Rewriting ffeca63 solely to change 'via omp' → 'via GitHub Copilot' needs a …"* — and the relay finalized without recording either of them. No `recording them (#7879)` line, no PR comment, nothing in the summary. The same relay had done exactly that 12 minutes earlier for chairlift#187 (3 notes recorded and posted).

## Why

omp queues advisor notes generated mid-turn and appends them to the transcript **at the end of the agent's turn**, in the same instant as the agent's final message. The session JSONL for this task shows it exactly:

```
21:52:36.313Z  assistant  (tool call)
21:52:36.660Z  toolResult main false https://github.com/projectbluefin/utah/pull/218
21:52:40.438Z  assistant  "PR #218 confirmed open against `main` … HIVE_VERDICT: complete — …"
21:52:40.438Z  custom_message  <advisory severity="concern"> Correction to my last note: index.md has no per-skill date …
21:52:40.438Z  custom_message  <advisory severity="concern"> Bound this: …
21:52:40.439Z  custom_message  <advisory severity="concern"> Skip the rebase. Rewriting ffeca63 solely to change "via omp" → "via GitHub Copilot" …
21:52:40.439Z  custom_message  <advisory severity="concern"> Disregard my …
```

The relay's verdict watch (`VERDICT_WATCH_INTERVAL_MS = 5000`, `bin/contributor-relay.js:280`) saw the verdict line while the final message was still streaming and ran the finalizing tick immediately:

```
2026-09-20T21:52:39  Task …utah-169…: HIVE_VERDICT: complete is on the pane — running the progress tick now … (#7841)
2026-09-20T21:52:40  Detected PR for …: https://github.com/projectbluefin/utah/pull/218
2026-09-20T21:52:40  Task …utah-169… completed — signal=verdict (HIVE_VERDICT: complete)
```

`verdictWatchTick()` (`bin/contributor-relay.js:4431-4449`) calls `progressTick()` the moment `detectCompletionVerdict()` matches, and the #7886 collection (`postVerdictNoteBlocks(paneScanLines, …)` at `:5237-5243`) runs on the pane captured in *that* tick. The notes rendered ~1 s later, under the verdict, on a pane nobody looked at again. My own pane sampler at 21:52:42Z shows both `⟦concern⟧` blocks under the verdict.

Whether a task's late notes get recorded therefore depends on the phase of a 5 s timer. chairlift#67/#187 landed on the good side of it; utah#169/#218 did not. (Verdicts printed inside `TASK_GRACE_PERIOD_MS` (`:3899`, 180 s) are unaffected — the tick refuses to judge until the grace ends, which is why the first verdicts of short tasks always picked up their notes tonight; that is an accident of the grace period, not a settle.)

There is a second, longer window this leaves open: the advisor's review of the *final* turn itself (Opus xhigh took 30–100 s per review tonight; in this task its last two outputs, 21:52:38Z and 21:52:43Z, happened to be "Silent"). A note from that review is generated after finalization and has nowhere to go at all. That one is harder — bounding a wait on advisor activity — and may not be worth it; the end-of-turn delivery race is cheap to close and is the one that actually lost notes.

## What it cost

The dropped "Skip the rebase" note was about the agent rewriting an already-pushed commit solely to change its attribution trailer from `via omp` to `via GitHub Copilot` (utah's AGENTS.md prescribes the Copilot wording). The agent had already done it by the time the note was delivered, and the reviewer of #218 will never see the advisor's objection.

## Suggested fix

On the finalizing tick, if the verdict line was first seen less than one watch interval ago, re-capture once before collecting: i.e. `verdictWatchTick()` records `verdictFirstSeenAt`, and the collection at `:5237` uses a fresh `captureTmuxLines()` taken at least ~2 s after that (or simply defers finalization to the next 5 s watch tick when `Date.now() - verdictFirstSeenAt < VERDICT_WATCH_INTERVAL_MS`). Bounded at one extra interval, no new turns, and it makes the "record notes under the final verdict" guarantee independent of timer phase.

Test shape: pane that shows the verdict alone on capture N and the verdict + `ⓘ Advisor 1 note` block on capture N+1 → the notes must be in `unaddressedNotes`.

Relay `bin/contributor-relay.js` at v4 `dc698c8` for line numbers; behaviour observed on the `:latest` container image (relay carrying #7886) at 2026-09-20T21:52Z.

clubanderson:
Closes #7907

## Why

omp queues the advisor notes it generates mid-turn and appends them at the end of the agent's turn — same millisecond as the final message in the session JSONL — rendering them under `HIVE_VERDICT` about a second later. The #7841 glance (`verdictWatchTick`, 5 s) saw the verdict while the final message was still streaming and ran the finalizing tick on a pane the notes had not reached yet. So the #7879 recording (`postVerdictNoteBlocks`) collected nothing and the #7759 follow-up never saw a new concern. utah#169/#218 lost four `⟦concern⟧`s; chairlift#187 kept three on the same relay 12 minutes earlier. Timer phase.

## What

- `VERDICT_SETTLE_MS = 2000`. The first `progressTick()` to see a given verdict **line** records the sighting (`noteVerdictSighting`) and defers: reports `working`, judges nothing, and **restores the #7759 concern snapshot** so notes that render during the settle still read as new on the tick that finally looks (without that the follow-up silently stopped firing — caught by the test).
- Placed ahead of every branch that reads the verdict, so the #7759 follow-up, the #7862 PR-claim follow-up and the #7879 recording all see the settled pane.
- `verdictWatchTick()` doesn't spend the fast path on the discovering run: while the settle is pending it glances again next interval; it marks the line consumed only on the run that judges. Cost: at most one extra glance interval, no new turns.
- The line is remembered, not a boolean: a second verdict after a follow-up is a new sighting and settles on its own.
- Test harness `loadRelay()` sets the settle to 0 (`__setVerdictSettleMs`) so the existing tick-by-tick tests are unchanged; the #7907 tests opt in.

## Tests

`node bin/contributor-relay.test.js` → 499/499 (was 496). New:
- the exact shape from the issue: verdict alone on capture N, verdict + notes on N+1 → deferred, then the concern earns the #7759 follow-up; the second verdict races the same way and the late concern under it is recorded (#7879) with no extra turn;
- the glance defers, does not re-fire while the settle is pending, then judges and consumes the line;
- zero-settle default judges on the discovering capture.

Not covered (as the issue says): the advisor's review of the *final* turn itself (30–100 s after the verdict). Documented in `contributor-relay.md` as a separate decision.


Danathar:
I was dispatched on #7907 and had an equivalent fix ready when this PR landed on the same branch name, so I am not opening a second PR. I ran the suite here (499/499) and then checked one shape the new tests don't cover, and it is a gap worth closing before merge:

**A second verdict re-printed verbatim after the follow-up gets no settle.** Both sightings are keyed on the line's *text* (`verdictFirstSeen.line`, `verdictFastPathLine`), and the #7759 docs already call out that an agent answering the follow-up may re-print its conclusion byte-for-byte (`postVerdictReviewAnswered` exists for exactly that reason). When it does:

- `verdictWatchTick()` returns at `paneVerdict.line === verdictFastPathLine` — the second verdict never takes the fast path (pre-existing from #7841, unchanged here), so it waits for the scheduled tick, up to 120 s;
- when that tick lands, `noteVerdictSighting(line)` sees the same line first seen minutes ago → *settled* → the tick judges the capture that first shows the re-print. That is the utah#218 race again for the second verdict: the #7879 collection reads a pane omp's end-of-turn notes are not on yet.

Reproduced on this branch by appending this to the test file — it fails with `instead it finalized on the capture that found it`:

```js
test('#7907 a second verdict re-printed VERBATIM after the follow-up gets its own settle', async () => {
  let pane = ompPane(' ⠋ Editing docs/skills/index.md', '╰─');
  const relay = loadRelay({ backend: 'omp', paneText: () => pane, prMeta: new Error('gh: offline') });
  const log = console.log; console.log = () => {};
  const err = console.error; console.error = () => {};
  try {
    relay.__setVerdictSettleMs(40);
    dispatchTask(relay, 'ct-7907-verbatim', 169);
    const first = [' Opened https://github.com/foo/bar/pull/218', OMP_UTAH218_VERDICT, ...OMP_UTAH218_LATE_NOTES];
    pane = ompPane(first, OMP_UTAH14_TAIL);
    relay.__crashTick();                       // first sighting: defers
    await settleElapsed(60);
    relay.__crashTick();                       // settled: the follow-up fires
    assert.strictEqual(relay.getPostVerdictReviewCount(), 1);

    // The agent answers and re-prints its verdict VERBATIM; this capture shows
    // it alone — the advisor's notes on that turn land a second later.
    const second = ompPane(first, `> ${relay.POST_VERDICT_REVIEW_MESSAGE}`, ' Left ffeca63 alone; noted the trailer in the PR body.', OMP_UTAH218_VERDICT);
    pane = ompPane(second, OMP_UTAH14_TAIL);
    relay.__crashTick();
    assert.strictEqual(relay.__sent.filter(m => m.type === 'task_complete').length, 0,
      'a verbatim re-print is a new printing and must be held for the settle');
  } finally { console.log = log; console.error = err; teardown(relay); }
});
```

(Side note on the existing #7907 test: `first` nests `OMP_UTAH218_LATE_NOTES` without spreading, so `ompPane(first, …)` renders the note block as one comma-joined line. It doesn't affect that test's assertions — the notes sit above the second verdict there — but the fixture is not the pane it looks like.)

**Suggested fix, which I have working locally (499/499 with three tests of its own):** key a sighting on the line text *plus how many copies of it the scan window holds*, and use that same key for the fast-path latch, so a verbatim re-print is a new printing with its own settle and its own fast path:

```js
// verdictSightingKey identifies ONE printing of a verdict line on the pane.
// The line's text alone cannot: after a #7759/#7862 follow-up the agent may
// re-print its conclusion verbatim, and that second printing is a new
// statement. A count that changes because an earlier copy scrolled out of the
// window reads as a new printing, which costs one more glance — harmless.
function verdictSightingKey(lines, verdict) {
  let copies = 0;
  for (const line of lines) if (line === verdict.line) copies++;
  return `${copies}:${verdict.line}`;
}
```

with `verdictWatchTick()` capturing `paneScanLines` once, comparing `verdictSightingKey(paneScanLines, paneVerdict)` against `verdictFastPathKey`, and `noteVerdictSighting` taking `(lines, verdict)` instead of the bare line. Happy to push that as a commit on this branch if you want it, or leave it as a follow-up — the main fix here is right and I would not hold the PR on it.

Issue #7678, opened 2026-09-19
🐛 bug(contribute): omp container mode starts with no credentials, and the relay reads the setup wizard as ready and types a task into the login box


Danathar:
Follow-up to #7661 and #7663. omp is in the contributor image now. Two more things stop it from working in container mode.

**`just contribute-hive omp` starts omp as a fresh install with no credentials, and the relay then reads omp's setup wizard as "ready" and types a task into the login box. I want container mode to sign in the way local mode does, and the relay to wait when omp is asking for a login.**

## What happened

Hive `v4` at `d1e4dd4b5`, image digest `05ef51cf…`, docker.

`just contribute-hive omp` started the container. The relay log:

```
omp CLI is installed; authentication will be checked in its interactive pane.
...
Authenticated, but CLI is not ready yet — withholding ready until the CLI reaches its prompt
CLI ready — accepting tasks
Task assigned: issue projectbluefin/utah-packages#25 — supply-chain: lock factory buildroots ...
Task prompt sent to CLI
```

The omp pane at that moment was its first-run wizard:

```
                    omp
              Setup step 1 of 5

Set up your providers
Sign in and pick a web search provider. Press Esc when you're done.

Providers:  [Sign in]   Web search

Signing in to openai-codex
Browser login: Open login URL

Paste the authorization code (or full redirect URL):
>

A browser window should open. Complete login to finish.
```

So the task prompt went into the "Paste the authorization code" field. I stopped the run.

Local mode does not have this problem: `HIVE_OMP_DANGEROUSLY_RUN_UNCONFINED=1 just contribute-hive omp local` picks up my sign-in with no extra steps, because it runs the host's omp with my real home directory.

## Why

Two things.

**1. Nothing copies omp's config into the container.** `contribute-hive` stages each backend's config into a temporary directory and mounts that: `~/.claude`, `~/.copilot`, `~/.config/goose`, `~/.codex`, `~/.pi` (`Justfile:1405-1491`). There is no `omp)` case. On my host omp keeps its state in `~/.omp/agent/` (`config.yml`, `agent.db`, `models.db`, ...). None of it reaches `/home/dev` in the container, so omp starts from nothing. #7663 added the binary to the image; it did not add the config path.

[...]

PR #7931, opened 2026-09-21
🐛 fix(container): sync a refreshed omp OAuth credential back on a timer; withhold ready for a disabled provider

### LINKED ISSUE #7922: 🐛 bug(container): staging a COPY of an OAuth credential — the first in-container refresh revokes the host's refresh token; later launches get 'No API key found', omp falls back to an unreachable ollama model, and the relay still reports Sonnet to the hub
## What happened

omp container mode, hosted spoke `7dde068`, relay image at v4.69.0, primary `anthropic/claude-sonnet-5` via an **OAuth** credential (Claude subscription login, not an API key). Four containers launched from the same host store tonight:

| launch (Z) | container | result |
|---|---|---|
| 21:03 | `e47087b4` | fine, 1 task |
| 21:22 | `3f0f0168` | fine, ran ~2 h, 6 tasks + a review cycle |
| 23:20 | `c6bea302` | pane: `Error: No API key found for anthropic. Use /login, set an API key environment variable, or create /home/dev/.omp/agent/agent.db` |
| 23:23 | `3996434e` | same; omp's model picker showed `○ anthropic` unavailable and `● ollama 21`, and the footer read `⬢ qwen3-coder:30b` — a model the container cannot reach |

Host store after the 23:20 launch (`~/.omp/agent/agent.db`, table `auth_credentials`, non-secret columns only):

```
provider=anthropic  credential_type=oauth
disabled_cause="oauth refresh failed: OAuthError: anthropic token refresh failed: 400 {"error": "invalid_grant", "error_description": "Refresh token not found or invalid"}"
updated_at=2026-09-20 23:21:37Z
```

`/login` on the host restored it (23:24:55Z); the fifth container is fine.

## Why

`bin/omp-backend.js --stage` copies `agent.db` + `agent.db-wal` into the stage dir (`OMP_STAGED_AGENT_ENTRIES`, `bin/omp-backend.js:42-48`) and narrows the `auth_credentials` rows to the selected provider (`:50`, `stageOmp` at `:314`). The container then runs against a **copy** of an OAuth record. OAuth refresh tokens are single-use: the first client to refresh gets a new refresh token and the old one is revoked. So once a container's omp refreshes (its access token expires mid-run — `3f0f0168` ran two hours), the host's copy holds a revoked refresh token. Every later launch stages the revoked token, and the host's own omp is broken too, until the operator logs in again. Nothing in the launch path can tell the operator this happened.

Two aggravations on top of the root cause:

1. **The relay never noticed.** It logged `Detected running model from omp session transcript: anthropic/claude-sonnet-5` and the hub fleet showed `model: claude-sonnet-5 / reasoning_effort: high` for both broken containers, while the pane's footer said `qwen3-coder:30b`. The detection reads the session's configured model, not what omp actually resolved. Had the hub had matching work, a task would have been "worked" by an unreachable local 30B model under Sonnet's name — the #7862 shape, from the credential side.
2. **omp's fallback is silent.** With anthropic unusable it picked the first provider that had any models listed (the host's ollama entries staged via `models.db`) and sat at the prompt as if fine.

The same shape exists for the other staged OAuth credentials: claude container mode stages `.credentials.json` (`Justfile:1421-1436`, #7836), which is an OAuth record with a refresh token; pi stages `auth.json` (`bin/pi-backend.js:92,124`). I did not reproduce those tonight; the mechanism is identical.

## Suggested fix

Pick one of the first two; the third is cheap and worth doing regardless.

- **Bind-mount the live credential store instead of copying it** (rw), so a refresh inside the container lands in the host record and vice versa. For omp that means the DB file itself; the provider-narrowing then has to happen by view/permission rather than by copy, or accept that the container sees every signed-in provider (the #7681 narrowing was about not exposing unrelated providers — an rw mount of a narrowed *copy* cannot also propagate refreshes).
- **Copy the refreshed credential back on container exit** (and on a timer while running): `docker cp` / read the stage DB's `auth_credentials` row for the selected provider and upsert it into the host store if its `updated_at` is newer. Keeps the narrowing; closes the window except for a crash mid-run.
- **Detect and refuse at launch.** Before staging, if the selected provider's row has a non-null `disabled_cause`, fail the launch with that text and "run `omp` → `/login` on the host", instead of starting a container that will fall back silently. And the relay should compare the model it reports to the hub with the one the CLI actually resolved (omp's footer / session `model_change` event) and refuse `ready` on a mismatch — that is the guard that would have kept a task off `qwen3-coder:30b`.

Cites at v4 `7dde068`.

Danathar:
## Summary

**Rebased onto v4 over #7923.** While this PR was open, #7923 (caa77cd44) landed the narrow fix for #7922 in the same files: `bin/omp-backend.js --sync-back <omp-dir> <stage-dir> [model]` copies the selected provider's refreshed credential row back to the host at container exit (only the existing host row, only its credential columns, only when strictly newer — `syncBackOmp`), `listProviderRows` reads `disabled_cause`, and `--stage`/`--describe` refuse a disabled row. This PR now keeps that implementation as-is and layers only what #7923 does not cover:

1. **A timer-driven sync-back while the container runs.** `just contribute-hive omp` runs the same `--sync-back` (same arguments, same narrowing) every `HIVE_OMP_CREDENTIAL_SYNC_SECONDS` (default 300; `0` = exit-only) reading the staged store in place through the bind mount, so the host is not left holding a revoked refresh token for the length of a run — a second container launched from the same host, or the host's own omp, would otherwise stage/refresh the dead token. `cleanup_container` kills the timer (`OMP_SYNC_PID`) before its own exit pass so the two never run over the same row at once.
2. **Hardening of the store access** in `bin/omp-backend.js`: `PRAGMA busy_timeout` on every read and write (the timer reads a store omp is writing to); SQL fed to the `sqlite3` CLI fallback on stdin rather than argv, so a credential never appears in the process table; `printableCause()` strips control characters from the untrusted `disabled_cause` and caps it before it reaches the terminal; `hasDisabledCauseColumn()` so a store from an older omp without the column still describes, stages and syncs back.
3. **The relay half**: `bin/contributor-relay.js` no longer advertises `ready` and a model for an omp whose configured provider's credential is disabled (see "The relay half" below), and `bin/lib/pane-classifier.js` classifies omp's `No API key found for <provider>. Use /login` as blocked on a human.
4. Tests and docs for the above; the changelog fragment covers only these additions (#7923's fragment already covers the sync-back and the refusal).

Dropped from this PR as duplicates of #7923: its own `--sync-back` implementation (`syncOmpCredentialsBack`, the compare-and-swap on `data`, the "revive a disabled host row whatever the timestamps" rule, the OAuth-record validation), its own row-reading helpers (`readRows`/`readCredentialRows`/`updateCredentialRow`), its `disabledCauses`/`blockedProviders` report shape and refusal wording, and the tests for those. v4's `queryRows`/`execStatements`/`syncBackOmp`/`disabledProviders` and its message texts are what ships.

<details><summary>Original summary (before the rebase)</summary>

`just contribute-hive omp` runs the omp coding agent in a container and, so the container can never write into the operator's real config, gives it a *copy* of the host's omp sign-in (the `~/.omp/agent/agent.db` credential store). For a subscription (OAuth) login that copy is a problem the copy cannot see: an OAuth refresh token is single-use, so the first time omp inside the container refreshed its access token, the **host's** token was revoked. The host's own omp then failed to refresh and marked the credential disabled, and every later container launch staged that dead credential — omp in the container quietly switched to the host's local ollama model (unreachable from the container) while the relay told the hub the container was running `claude-sonnet-5`, ready for work. On a hosted spoke on 2026-09-20: two good launches (one two hours long), then two broken ones until the operator ran `/login` on the host by hand.

This PR (1) copies the credential the container refreshed back into the host's store — on a timer while the container runs and once more at exit — so the host keeps a live token; (2) refuses to launch a container when the selected provider's stored credential is already disabled, printing the cause omp recorded and the fix; and (3) stops the relay from advertising `ready` and a model for an omp whose configured provider cannot actually run.

Blast radius: omp container mode (`Justfile` `contribute-hive`, `bin/omp-backend.js`) and the omp branch of the relay's readiness/model detection. The one new host write is a single credential row, gated as described below; no other backend is touched.

</details>

## Root cause

`bin/omp-backend.js --stage` copies `agent.db` into the ephemeral staging dir and narrows its `auth_credentials` rows to the selected provider (#7678, #7681). The container mounts that copy. omp 18.2.x refreshes an OAuth access token ~60 s before `expires` (`#kt`/`#Ft` in the bundle) and the provider rotates the refresh token on every refresh, so after the container's first refresh the host row holds a revoked `refresh`. The host's omp, on its next refresh, gets `400 invalid_grant` and writes `disabled_cause = "oauth refresh failed: …"`; omp reads only rows `WHERE disabled_cause IS NULL`, so a staged disabled row is invisible to the container's omp and it resolves whatever other provider has models listed.

The relay half: omp writes no session transcript until its first turn, so at auth time `detectOmpSelection()` fell through to `config.yml`'s `modelRoles.default` — what omp is *set* to run — and logged it as "Detected running model from omp session transcript". The pane looked ready because it was: the same footer chrome, a different model in it.

## Change

> After the rebase, the **`bin/omp-backend.js`** and **`Justfile`** paragraphs below describe the pre-rebase design and are superseded by the Summary: the sync-back that ships is #7923's `syncBackOmp` (strictly-newer, selected provider's existing row, called with `"${HOME}/.omp" "${CLI_STAGE}/.omp" "${AGENT_MODEL:-}"`), plus the timer, the busy-timeout/stdin/printable-cause/old-schema hardening. The relay, pane-classifier and docs paragraphs still apply.

**`bin/omp-backend.js`**

- `--sync-back <omp-dir> <stage-dir>` (`syncOmpCredentialsBack`): for each staged `oauth` row, find the host row with the same `id` and `provider` (the copy kept the ids) and write the staged `data` over it only when all of: the staged row is not disabled; it is a JSON object with non-empty `refresh` and `access` strings and ≤ 64 KiB; it differs from the host's; and it is newer than the host row **or the host row is disabled** (a disabled host row is a token the provider already rejected — the incident's exact ordering, where the host disables its row *after* the container rotated). The `UPDATE` is a compare-and-swap on the host row's current `data` — omp's own guard between its processes — so a host omp that rotated first keeps its token; it clears `disabled_cause`, and it goes through omp's `auth_change_revision` triggers so a running host omp notices. API-key rows are never touched. A staged row the *container* disabled is reported as a warning (the host's token may be dead too), never written. Never throws: every problem is a warning line, exit 0, because the trap that calls it must still delete the staging dir. The sqlite3-CLI fallback feeds the SQL on stdin so a credential never appears in argv; the cause text read from the container's copy is stripped of control characters and capped before it can reach the host terminal.
- `--stage` / `--describe`: `listProviderRows` now reads `disabled_cause` (guarded by `PRAGMA table_info`, so an older omp schema still works). A provider is *disabled* when every row it has is; `describeOmpHost` reports `disabledCauses` and `blockedProviders` (the selected ones — for a `source: none` selection, only when nothing usable is left), warns in the preflight, and `stageOmp` throws **before copying anything** when a selected provider is blocked.

**`Justfile`** (`contribute-hive`, container mode)

- After the startup check, when `BACKEND=omp` and a store was staged: a background loop runs `--sync-back` every `HIVE_OMP_CREDENTIAL_SYNC_SECONDS` (default 300; `0` disables it). The staged store is a bind mount, so the host reads it in place.
- `cleanup_container`: stop the container → kill the timer → `--sync-back` → `rm -rf "${CLI_STAGE}"`, in that order.

**`bin/contributor-relay.js`**

- `ompProviderCredentialState(provider)` reads the store omp runs against (`HIVE_OMP_AGENT_DIR`, i.e. the staged copy in container mode) via `node:sqlite`, returning `usable` / `disabled` (+ cause) / `absent` / `unknown`. Only `disabled` is acted on: an absent row may still be served by an API key in the environment, and no `node:sqlite`, no store, an old schema or a read error are all `unknown`.
- `getCLIState()`: for omp, a pane that classifies `ready` while the configured provider (AGENT_MODEL if it names one, else `modelRoles.default`) is `disabled` becomes `needs-login`, with the cause and the fix logged once. This uses the existing login-banner machinery; no new readiness state.
- `detectOmpSelection()` gains `source: 'transcript' | 'config' | ''`; a `config`-sourced model whose provider is disabled is not reported (`''`), and `refreshDetectedModel` logs `from omp config.yml (no session transcript yet)` vs `from omp session transcript` honestly. A transcript record still wins and is reported as-is — in the incident that would have been `ollama/qwen3-coder:30b`, the truth.

**`bin/lib/pane-classifier.js`**

- `paneShowsLoginRequiredError` also matches omp's `No API key found for <provider>` + `Use /login` (same line or the next), so a turn that dies at the credential is `BLOCKED_ON_HUMAN`, not a completed task.

**Docs**: `docs/backend-setup.md` omp row, `src/docs/contributor-relay.md` model-detection caveat, `src/docs/env-vars.md` (`HIVE_OMP_CREDENTIAL_SYNC_SECONDS`), `bin/README.md`, changelog fragment.

### What this does not do, and why

- **Not a read-write bind mount of the live store** (the issue's option 1): the H6/CWE-668 boundary and its guard `TestOmpStagingMountsTheHostSignIn` deliberately forbid mounting `~/.omp` itself, and an rw mount of a narrowed copy cannot propagate refreshes anyway. The copy-back is the narrowest write that fixes the incident: one row, that the container already held, only in the forward direction.
- **Not a footer-vs-config comparison for refusing `ready`** (the issue's second half of option 3): the footer shows display names (`Sonnet 5`, `GPT-5.6-Terra`) that changed shape between omp 18.1 and 18.2; a parser there would have to be right on every omp release or it would silently keep a healthy relay from ever going ready — the failure class #6413/#6639 already paid for. The credential store is deterministic evidence of the same condition and is what the gate reads.
- **Two containers running from one host store at the same time** still race each other's refreshes; that is inherent to single-use refresh tokens, is documented in `backend-setup.md`, and the timer keeps the *host* right through it.

## Related issues

Fixes #7922

## Testing

After the rebase onto v4 (head `836f63c05`):

- [x] `node --check bin/contributor-relay.js bin/omp-backend.js`
- [x] `node bin/contributor-relay.test.js` — 519/519. #7923's three tests (`omp sync-back writes …`, `… does not roll a host sign-in back …`, `omp container staging refuses a sign-in omp has disabled …`) pass unchanged against the merged `omp-backend.js`. This PR's tests: the sqlite3-CLI path round-trips a token containing `'` via stdin and bumps omp's `auth_change_revision`; a hostile `disabled_cause` (terminal escape, bell, 1 KiB) is printed with no control characters and capped in both the preflight and the `--stage` refusal; a store without the `disabled_cause` column still describes, stages and syncs back; and the five relay tests (ready-looking pane + disabled provider → `needs-login` and no model; the gate stands aside for usable/absent/unknown/other-backend; `AGENT_MODEL` picks the provider checked and a transcript still wins; the source-labelled log line; the `No API key found … Use /login` login wall).
- [x] `cd src && go test ./pkg/dashboard/ -run 'Omp|Backend|Relay|EnvVars' -count=1` — ok, including the two Justfile guards (`TestOmpRefreshedCredentialIsCopiedBackBeforeTheStageIsDeleted`, adapted to #7923's `[ "${BACKEND}" = "omp" ]` gate and `|| echo …` warning; `TestOmpRefreshedCredentialIsCopiedBackWhileTheContainerRuns`).
- [x] `just -n contribute-hive omp` rendered → `bash -n` ok, `shellcheck -S warning` clean. The `kill` + `wait` of the timer subshell verified quiet (no "Terminated" line) under `set -euo pipefail`.
- [ ] Not run: a live container run (no container runtime in the sandbox that produced this).

## Contributor checklist

- [x] PR targets `v4`.
- [x] Docs updated where behavior changes.
- [x] `changelog.d/fixed-7931-omp-sync-timer.md` carries the entry (the sync-back and refusal are in #7923's `fixed-7922-omp-oauth-refresh.md`).


Danathar:
Rebased onto v4 over #7923 (caa77cd44), which landed the narrow fix for #7922 in the same files while this was open. The result is v4's implementation plus only what this PR adds on top; new head is 836f63c05.

Dropped as duplicates of #7923: this PR's own `--sync-back` (`syncOmpCredentialsBack`, its compare-and-swap on `data`, the "revive a disabled host row whatever the timestamps" rule, the OAuth-record validation), its row-reading helpers (`readRows`, `readCredentialRows`, `updateCredentialRow`), its `disabledCauses`/`blockedProviders` report shape and refusal wording, its changelog fragment, and the tests for all of that. What ships is v4's `queryRows`/`execStatements`/`syncBackOmp`/`disabledProviders`, v4's `--sync-back "${HOME}/.omp" "${CLI_STAGE}/.omp" "${AGENT_MODEL:-}"` call in `cleanup_container`, and v4's warning text on failure.

What this PR still adds:

- The same `--sync-back` also runs every `HIVE_OMP_CREDENTIAL_SYNC_SECONDS` (default 300, `0` = exit only) while the container runs, reading the staged store in place through the bind mount. `cleanup_container` kills the timer (`OMP_SYNC_PID`) and waits for it before its own exit pass, so the two never run over the same row at once. The timer pass discards `--sync-back`'s per-pass status line so it does not interleave with the relay logs; a failure still prints the same warning as the exit pass.
- Hardening in `bin/omp-backend.js`: `PRAGMA busy_timeout` on every read and write (the timer reads a store omp is writing to); SQL fed to the `sqlite3` CLI on stdin rather than argv; `printableCause()` strips control characters from the untrusted `disabled_cause` and caps it before it reaches the terminal; `hasDisabledCauseColumn()` so a store from an older omp without the column still describes, stages and syncs back (`syncBackOmp` leaves the column out of its `UPDATE` in that case).
- The relay half, unchanged from before: `getCLIState()` reports `needs-login` instead of `ready` for an omp whose configured provider's credential is disabled, `detectOmpSelection()` records its source and withholds a config-sourced model whose provider is disabled, and the pane classifier treats omp's `No API key found for <provider>. Use /login` as a login wall.
- One changelog fragment for the timer and the relay change (`changelog.d/fixed-7931-omp-sync-timer.md`); the sync-back and the refusal are already in #7923's fragment.

Tests: `node bin/contributor-relay.test.js` 519/519 (v4's three #7922 tests pass unchanged; this PR's three `omp-backend` tests now cover the sqlite3-CLI stdin path, the printable cause, and the old schema, plus the five relay tests). `cd src && go test ./pkg/dashboard/ -run 'Omp|Backend|Relay|EnvVars'` ok; the Justfile guard `TestOmpRefreshedCredentialIsCopiedBackBeforeTheStageIsDeleted` was adapted to accept #7923's `[ "${BACKEND}" = "omp" ]` gate and `|| echo` fallback. `just -n contribute-hive omp` renders, `bash -n` and `shellcheck -S warning` are clean. `git diff --stat origin/v4...HEAD` touches only the Justfile, `bin/{README.md,contributor-relay.js,contributor-relay.test.js,lib/pane-classifier.js,omp-backend.js}`, the docs, the Go test, and the fragment. The PR body's Summary and Testing sections say the same.

Danathar:
@clubanderson one design question for you, since it touches #7923's rule rather than this PR's additions.

`syncBackOmp` only writes when the staged row is strictly newer than the host's. The incident this came from has the opposite ordering: the container rotates the token first, then the host's own omp tries to refresh, gets `invalid_grant`, and writes `disabled_cause` with a later `updated_at`. At that point the host row is both newer and dead, so the exit sync leaves it alone and the operator still has to `/login` by hand. The timer sync in this PR narrows that window (the container's fresh row usually reaches the host before the host's next refresh) but does not close it.

The pre-rebase version of this PR revived a disabled host row whenever the staged row for the same provider was usable, regardless of timestamps. I dropped that when layering onto #7923 because it changes your strictly-newer guarantee. Do you want it back as a narrow exception (host row has `disabled_cause` set, staged row for the same provider has none, so copy the staged credential and clear the cause), or would you rather keep strictly-newer and accept the manual `/login` in that ordering? Happy to add the exception in a small follow-up if you want it.

clubanderson:
Reviewed: 519/519 relay tests; `node --check` clean; Justfile parses; sqlite3 stdin path verified by hand (`.timeout` emits no row in `-json`); trap ordering (`trap - EXIT` before timer start, `kill`/`wait` `|| true`) checks out; `--sync-back` is not gated by the `stageOmp` refusal so a host row that goes disabled mid-run cannot block the write-back.

One hardening nit, inherited from #7923 but made more reachable by the 300 s timer: `syncBackOmp` (bin/omp-backend.js ~L513) only checks `updated_at`, so a staged row the **container's** omp disabled (host refreshed first → container refresh fails → `disabled_cause` set with a bumped stamp) would be copied over the host's working row. A disabled row never helps the host; it should be skipped and reported. I'll land that as a small follow-up on v4 rather than hold this.

/lgtm
/approve

Issue #7871, opened 2026-09-20
🐛 bug(hub/select): #6869 settle scan re-offers issues already settled by a merged PR — when a no_work_needed verdict names the settling PR/commit, record it as a merged claim instead of discarding it


Danathar:
## Summary

#6869 (closing #6867) suppresses offers for issues already settled by a merged PR — but only when the PR *references* the issue (closing keyword or issue number the claim scan can recover). Fixes that land without mentioning the issue are invisible to it, and today they were the majority of the wasted cycles: 5 of 7 tasks on one contributor ended `no_work_needed` because the work was already merged or already open under another PR.

The agent, however, *does* find the settling PR and says so in its verdict. The hub currently discards that and applies the generic no-PR cooldown. Proposal: parse the verdict reason for the PR/commit it cites, verify it, and record it as a merged claim in the `prclaims` ledger — so the issue is suppressed the same way a referencing PR would suppress it.

## Evidence (2026-09-20, hosted projectbluefin spoke `92e0056`, omp + Sonnet 5 / Opus 5 advisor)

| task | verdict reason (as sent in `task_complete`) | settling PR references the issue? |
|---|---|---|
| actions#533 | `all four workflow entries #533 asks for are already merged into upstream/main's unit-tests.yml paths filters (via #532)` | **no** — actions#532 has no closing refs and never mentions #533 |
| testsuite#866 | `e2e.yml already has migration-target/extra-tags restored via merged PR #867 (2026-09-20)` | **no** — testsuite#867 has no closing refs and never mentions #866 |
| utah#18 (twice, 09-19 and 09-20) | `…already merged in upstream/main (e6d3de3)` | direct commit, no PR |
| testsuite#794 | `already fully covered by open PR #808 … explicitly "Resolves #793 and #794"` | yes — open PR, so a live claim the scan *should* see (offered anyway?) |
| fsdk-containers#306 | `PR #277 must be merged first to avoid conflicting implementations` | open PR, no reference |

Each of these cost a full agent cycle (clone/fetch, issue + repo research, 2–9 min of a primary model plus an xhigh advisor) to discover what the previous agent had already discovered. The escalating no-PR cooldown (`advanceNoPRStreakLocked`, 4 h doubling) bounds the repeats but still re-offers the issue to a fresh contributor each window; utah#18 was worked to the same `no_work_needed` conclusion by two different models on two different days.

## Proposal

In `handleTaskComplete`, when `verdict == no_work_needed`:

[...]

Issue #7861, opened 2026-09-20
🐛 bug(relay): when an agent prints both HIVE_VERDICT lines, the relay keeps the last one — a PR-less `complete` overrides a real `no_work_needed` and the hub books `idle`


Danathar:
Observed 2026-09-20 14:30Z, omp 18.2.6 with `gemini-3.8-flash` (high) primary + `claude-opus-5` (high) advisor, local mode, relay `92e00565d`, task projectbluefin/fsdk-containers#305.

The agent printed both sentinels, one after the other:

```
 HIVE_VERDICT: no_work_needed — The issue is explicitly blocked by a pending maintainer decision in issue #305.
 HIVE_VERDICT: complete — The problem is identified as blocked on a maintainer decision.
```

The prompt forbids exactly this ("If you printed the no_work_needed line above, that already counts as your completion — do not print both"), but a model that narrates a closing "complete" after its real verdict will keep doing it, and Flash did it on its second task.

Relay (`bin/contributor-relay.js`): `detectCompletionVerdict()` → `detectHiveVerdict(lines, [HIVE_VERDICT_COMPLETE, HIVE_VERDICT_NO_WORK])`, which resolves via `lines.lastIndexOf(verdictLine)` — the **last** sentinel on the pane wins. So the task was reported `complete`:

```
Task ct-projectbluefin/fsdk-containers-305-…: HIVE_VERDICT: complete is on the pane — running the progress tick now … (#7841)
Task ct-projectbluefin/fsdk-containers-305-… completed — signal=verdict (HIVE_VERDICT: complete)
Ignoring PR https://github.com/projectbluefin/fsdk-containers/pull/268 … — not this task's work (merged 2026-09-12 …)
```

No PR (correctly — #268 was researched, not shipped), so the hub's completion ledger maps this to `completionVerdictIdle` ("returned to idle, no PR", `contribute_ledgers.go`). The run row now says `idle` for a task whose actual, more useful outcome — `no_work_needed`, blocked on a maintainer decision, with the reason text — was one line above.

## Proposal

[...]

PR #7933, opened 2026-09-21
🐛 fix(contribute): let a verdict say blocked — hold it for the full cooldown and apply the blocked label

### LINKED ISSUE #7924: 🐛 bug(contribute): a no_work_needed that names an external blocker is re-offered on the 4h backoff and the finding is thrown away — let the verdict say 'blocked' and hold for the full cooldown / apply the existing 'blocked' label (utah#100)
## What happened

projectbluefin/utah#100 ("nautilus missing from image, extensions ship orphaned"), omp container mode, Sonnet 5 high + Opus 5 xhigh advisor, hub `7dde068`, 2026-09-20T23:30–23:40Z.

Ten minutes of real research, two advisor corrections (a stale "recipes don't exist" claim from a 09-17 issue comment; a misread GHCR digest), and an accurate verdict:

> `HIVE_VERDICT: no_work_needed — nautilus's dependency closure (libportal/exiv2/libgexiv2) now has recipes on utah-packages main, but no successful factory build has published since they merged (latest GHCR tag is the frozen Aug 30 image, older than utah's pin) …`

That is a **blocked-on-another-repo** state: nothing in `utah` can change until `utah-packages`' factory publishes an image with those packages, at which point #100 is a one-line pin bump.

## What the hub does with it

Nothing durable. The verdict names no merged PR or commit, so #7884 records no settle claim; the issue carries no label; so the booking is an ordinary no-PR completion and the issue re-enters the queue on the no-PR streak backoff — `completedNoPRCooldownHours = 4` (`src/pkg/dashboard/contribute_ledgers.go:27`), doubling per streak entry (`advanceNoPRStreakLocked`, `:150-160`). Every retry re-runs the same ten minutes to re-discover "still no factory build", and the hub learns nothing from the reason string the agent handed it.

Meanwhile the hub already has the right primitive: an issue labelled `blocked` is withheld as "not contributor work until its dependency is cleared" (`src/pkg/dashboard/contribute_admission.go:11-30`, `hasBlockedWorkflowLabel`, enforced in both `ReadyQueue` and `selectTask`). utah#100 simply doesn't carry it — the last human comment predates today's recipe merge — and nothing connects the agent's finding to the label.

## Suggested fix (either half stands alone)

1. **Let the verdict say it.** Accept `HIVE_VERDICT: blocked — <reason>` (or a `no_work_needed` reason prefixed `blocked:`) alongside the two existing sentinels (`bin/contributor-relay.js:3021-3022`). The hub books it as `no_work_needed` with a **blocked** marker and holds the issue for the full with-PR cooldown rather than 4 h — one wasted cycle per cooldown period at worst, and the `verdict_reason` on the ledger row tells an operator what it is waiting on.
2. **Close the loop on the repo.** When the task credential carries `issues:write` for the repo (`src/pkg/github/app.go:582`, `IssuesPerm`), the relay applies the `blocked` label and posts the reason as a comment in the task's own attribution style ("— hive: backend=… model=…"), so the existing admission gate takes over and a human sees why. Without the permission, fall back to (1) only. Lifting the label stays human — hive should not un-block on its own.

Not proposing that the hub parse free-text reasons for "blocked"; the agent should say it explicitly, the same way it says `no_work_needed`.

## Same shape earlier tonight

projectbluefin/server#197 (23:00Z): `no_work_needed` because the maintainer closed the same-titled PR #203 with a design ruling and the open supersede #210 removes the subsystem. Also unsettlable by #7884 (closed PR, open PR), also no label, also back on the 4 h backoff. #7911/#7915 fixes the part where #210's `Refs #197` should have held it; the explicit-blocked verdict would have covered it regardless.

Danathar:
## Summary

The contributor relay hands a hive issue to an agent, watches for the agent's closing `HIVE_VERDICT:` line, and reports it to the hub, which decides how long to keep that issue out of the queue. Until now an agent had two ways to end a task: `complete` (a PR is open) or `no_work_needed` (nothing shippable).

Some issues are correct and genuinely stuck, but for a reason *outside* their own repository. projectbluefin/utah#100 ("nautilus missing from image") is one: the missing packages now have recipes in utah-packages, but no factory build has published an image with them, so nothing in utah can change until that other repo ships — and then the fix is a one-line pin bump. The agent found exactly that after ten minutes of research and could only say `no_work_needed`. The hub booked it as an ordinary "nothing happened" completion: the issue came back on the 4-hour backoff, the next agent redid the same ten minutes to rediscover the same dependency, and the reason the first agent had written was never used for anything.

This PR lets the agent say `HIVE_VERDICT: blocked — <what it is waiting on>`. The hub holds a blocked issue for the full week-long cooldown right away (instead of 4h → 8h → …), keeps the reason on the ledger row so an operator can see what it is waiting on, and — when the task's GitHub credential allows — the relay applies the repository's existing `blocked` label, which the hub already treats as "not contributor work until a human clears it". The assignment prompt now also asks the agent to leave one signed comment on the issue naming what covers or blocks it before printing either verdict, so the finding lives on GitHub and not only in one hub's ledger, and never to edit somebody else's PR description to add `Fixes #N`.

Blast radius: contribute queue only. `no_work_needed` and `complete` behave exactly as before; a relay and hub of different ages stay compatible (see wire shape below). The hub itself still never labels or closes anything — the label is applied by the relay with the per-task credential.

## Detail

**Relay (`bin/contributor-relay.js`).** `blocked` joins `HIVE_VERDICT_TOKENS` and is parsed by the same anchored, echo-guarded scanner (`<…>` placeholder echo rejected, `\b` token boundary, Markdown emphasis and CLI bullets tolerated). `no_work_needed — blocked: <reason>` is promoted to the blocked verdict with the marker stripped; a reason that merely mentions being blocked mid-sentence stays `no_work_needed`. `isNoWorkVerdict()` treats both as the "nothing to ship" family everywhere the relay used to compare against `HIVE_VERDICT_NO_WORK` (tick-loop finalization, lease-expiry crediting, headless path, the #7861 keep-the-informative-verdict rule); a PR this task opened still overrides it via `resolveTaskPR`'s `suppressesVerdict`. `markIssueBlocked()` runs `gh issue edit <issue> --add-label blocked` with the task credential *before* `stopAgentForTaskExit` drops it (same ordering as the #7879 PR comment), only for GitHub issue tasks and only when the hub's `auth_ok` `permissions` include `issues:write` (now remembered per hub). If the repo lacks the label it is created once and the add retried once; any further failure is a log line, never the completion. The relay never removes the label.

**Wire shape.** `verdictWireFields()` sends a blocked verdict as `verdict: "no_work_needed"` + `verdict_blocked: true`. A hub older than this PR reads the `no_work_needed` it already books (long offer suppression); sending a new token would have made it normalize to a bare `idle` and re-offer on the short cooldown — worse than today. The new hub also accepts `verdict: "blocked"` outright and advertises `blocked_verdict` in `server_capabilities`.

**Hub (`src/pkg/dashboard`).** `normalizeCompletionVerdict(reported, blocked, verifiedPR)` yields `completionVerdictBlocked` for either spelling; a verified PR still wins. `markTaskCompletedVerdictKeySignal` books `configuredWithPRCooldown()` at once for a blocked verdict, writes the `noWorkVerdicts` row with `Blocked: true`, and leaves the no-PR streak alone (it exists to escalate; this is already the ceiling). `isEvidenceLessCompletion` treats blocked as affirmative evidence like `no_work_needed` (`TasksCompleted++`, no PR credit, no promotion). `handleTaskComplete` does not feed a blocked reason to the #7871 settle path: it names what the issue is waiting on, not what settled it, and a merged PR in *another* repo must not become a claim on this one. The assignment prompt gains the `blocked` sentinel, the issue-comment contract, and the "don't edit someone else's PR body" rule; the attribution trailer instruction now covers issue comments too.

**Docs.** `src/docs/contributor-relay.md`: a row in the completion-contract table and a section on the blocked verdict. Changelog fragment `changelog.d/fixed-7924-blocked-verdict.md`.

**Not in this PR.** The issue's follow-up comment also proposes a `hive/claimed` label for the covered-by-open-PR case, honoured by admission like an open-PR claim and released when that PR closes. That is a new admission mechanism with its own lifecycle (a sweep that knows which PR a label stands for) and is better as its own change; the prompt's comment contract here already creates the PR→issue cross-reference for that case, and #7884 holds it on the hub side.

## Related issues

Fixes #7924

## Testing

- [x] `cd src && go test ./pkg/dashboard/` — full package green (57s), including 5 new `TestBlockedVerdict_*` cases: full cooldown vs the 4h ladder, held past the ladder sweep even with issue activity newer than the verdict, affirmative-evidence predicate, ledger round-trip through a restart, and the real `task_complete` handler driven with the relay's wire shape (cooldown, marked row, `TasksCompleted` but no PR credit, `blocked_verdict` advertised, and the settle verifier never reached — verified by flipping the marker off, which makes the verifier get called and the cooldown book 4h).
- [x] `node bin/contributor-relay.test.js` — 519/519, including 8 new `#7924` cases (parsing and echo guard, the older `no_work_needed — blocked:` spelling, the #7861 preference, end-to-end wire shape + label call, no-`issues:write` path, label create-and-retry bound, plain `no_work_needed` unchanged, GitHub-only scope). On the parent commit the end-to-end cases fail: `HIVE_VERDICT: blocked` is not a verdict there, so the tick never completes the task.
- [x] `node --check` on the relay and pane classifier, `bash -n bin/contributor-relay.sh`, `bash bin/contributor-agent.test.sh`, `go vet ./pkg/dashboard/`, `gofmt -l` on touched files, `src/scripts/check-docs-citations.py` and `check-api-reference-citations.sh` (all citations still resolve).
- [ ] Other / not run: the live label call (`gh issue edit … --add-label blocked` against a real repo) is exercised only through the recorded-command fake; the first real blocked verdict on a hosted spoke is the end-to-end check.

## Contributor checklist

- [x] PR targets `v4` (the assignment's base; the template's `v2` predates the line policy).
- [x] A `changelog.d/fixed-7924-blocked-verdict.md` fragment carries the changelog entry.


Danathar:
CI's only failure was `TestSleepRatchet`: the new `contribute_ws_blocked_verdict_test.go` added one fixed `time.Sleep(50ms)` between sending the `task_complete` and reading the contributor's counters, which took the test-code sleep count to 199 against the 198 baseline. Replaced it with `testutil.Eventually` waiting on the observable condition (`TasksCompleted == 1`), folded into the same commit; head is now `5b6f5929c`. Blocked-verdict and no-work-verdict tests pass locally (three runs), and the ratchet is back at baseline.

clubanderson:
Reviewed: 519/519 relay tests; `go test ./pkg/dashboard/` green incl. new Blocked*/NormalizeCompletionVerdict tests under -race; no new `time.Sleep` in tests. Wire shape stays backward-compatible (`verdict: no_work_needed` + `verdict_blocked: true`, key absent otherwise), ledger `Blocked` is `omitempty`, blocked branch takes `completedMu` like its siblings, `settleIssueFromVerdict` still only fires for plain no_work_needed, label application is gated on `issues:write` and never blocks `task_complete`. Good design.

/lgtm
/approve

PR #7934, opened 2026-09-21
🐛 fix: clamp contributor frames to the hub's 64 KiB read limit

### LINKED ISSUE #7932: 🐛 bug: headless task_complete frames exceed hub 64 KiB ws limit → code=1009 reconnect loop after successful tasks
## What happened?

After the `HIVE_RELAY_MAX_OUTPUT_BYTES` fix (#7739 / #7744) unblocked pi headless tasks, a fresh-pod contributor now **completes real tasks (exit 0)** — but the relay falls into a **reconnect loop**: immediately after sending `task_complete`, the hub closes the WebSocket with `code=1009 message too big`, the relay reconnects, the hub re-assigns the **same task**, it completes again, and 1009 repeats.

Observed on the live Bluefin hub (`hosted-projectbluefin-knuckle-gjvq.hive.hivecommons.dev`, spoke surface). Four tasks completed `(exit 0)` (projectbluefin/utah#24 three times + utah#126), each followed by:

```
Headless task ct-projectbluefin/utah-24-1789949133 completed (exit 0)
Connection to wss://…/api/contribute/ws closed (code=1009 message too big). Reconnecting in 1000ms...
```

One completed run opened a real PR (`projectbluefin/utah#185`).

## Expected behavior

A `task_complete` frame the relay sends should fit within the hub's message-size limit, or the relay should truncate the `tmux_output` tail to fit. A successful task should not be re-assigned in a loop because the result frame is oversized.

## Root cause

The relay sends the captured-output **tail** in `task_complete` with **no size cap**:

```js
// bin/contributor-relay.js
const outTail = outLines.slice(-TMUX_TAIL_LINES);   // last 15 LINES
send({ type: 'task_complete', …, tmux_output: outTail, … });
```

`outLines` comes from `createBoundedOutputCapture(HEADLESS_MAX_OUTPUT_BYTES)` (default now 16 MiB). For a JSON-streaming backend like pi (`--mode json`), a single output **line** can be a large `tool_execution_end` event with an embedded diff, so 15 lines can reach hundreds of KiB to MiB. The `ws` client sends it as one frame.

The hub's WebSocket read limit is a hard **64 KiB**:

```go
// src/pkg/dashboard/contribute_ws.go
wsMaxMessageSize = 64 * 1024
// …
conn.SetReadLimit(wsMaxMessageSize)
```

gorilla/websocket then closes with `CloseMessageTooBig` (1009) on the relay's oversized frame.

## Reproduction steps

1. Deploy the stock contributor image with `CONTRIBUTOR_MODE=headless`, `AGENT_BACKEND=pi`, `AGENT_MODEL=opencode-go/<model>`, `OPENCODE_API_KEY`, and the new image (post-#7744) so tasks don't die at 1 MiB.
2. Hub assigns a real task → pi runs → `Headless task … completed (exit 0)`.
3. Immediately: `Connection … closed (code=1009 message too big)`; task is re-assigned; loop repeats.

## Environment

- Branch/version: `v4` (`hive-contributor:latest` image, `contributor-relay.js` with `HIVE_RELAY_MAX_OUTPUT_BYTES`)
- Deployment method: Kubernetes (ArgoCD), `ghcr.io/hivecommons/hive-contributor@sha256:d09fe9…` (== `latest`/`v4-latest`, built 9/20)
- Backend: `pi` 0.84.1, provider `opencode-go`; headless status-file probes
- Hub: hosted Bluefin spoke (API v1.2)

## Logs

```
Headless task ct-projectbluefin/utah-24-1789945687 completed (exit 0)
Connection to wss://…/api/contribute/ws closed (code=1009 message too big). Reconnecting in 1000ms…
Task assigned: issue projectbluefin/utah#24 — tracking: close Utah release-readiness gaps against Dakota
Headless: running pi (one-shot) for projectbluefin/utah#24
Headless task ct-projectbluefin/utah-24-1789947292 completed (exit 0)
Connection to wss://…/api/contribute/ws closed (code=1009 message too big). Reconnecting in 1000ms…
…(same task re-assigned and re-completed 4×)…
```

## Suggested fixes

- **Relay side (minimal, unblocks pi/chatty backends):** cap the `tmux_output` (and `diagnostic`) payloads sent in `task_complete`/`task_failed` to a few KiB (e.g. truncate to the hub's limit, or slice lines AND bytes). The interactive path's `TMUX_TAIL_LINES` output is small, but headless JSON backends can emit huge single lines.
- **Hub side:** consider whether 64 KiB is the right ceiling now that headless backends stream more output; if raised, raise both sides together and document the bound.

This blocks pi (and any verbose backend) from completing headless pod tasks end-to-end even though the CLI itself succeeds.

Danathar:
Fixes #7932

## What was happening

The hub reads contributor frames under a hard 64 KiB cap — `wsMaxMessageSize`, installed with `conn.SetReadLimit` (`src/pkg/dashboard/contribute_ws.go`). gorilla/websocket does not truncate an oversized message: it closes the connection with `1009 message too big` and **the frame is lost**. For a `task_complete` that is not a dropped log line, it is a reconnect loop — the completion never lands, the hub's lease outlives the close, the same task is handed back, and the agent redoes work it already shipped. @SNThrailkill observed exactly that on the live Bluefin spoke: four `(exit 0)` completions against the same two issues, one of the runs having opened a real PR.

The relay sent the captured-output tail as fifteen **lines** with no byte bound. Fifteen terminal rows cannot be large — but a JSON-streaming backend like pi (`--mode json`) puts a whole `tool_execution_end` event, embedded diff and all, on one line, so the same fifteen lines is routinely hundreds of KiB.

## What changed

**Relay — clamp at the choke point.** `sendTo()` is the one function every frame passes through, so the bound goes there rather than at each call site that happens to build a tail: the oversized frame was built by a call site that had no idea a limit existed, and a guard per call site is the shape that lets the next one reintroduce it.

- `tmux_output` is held to 8 KiB on **every** frame — it is an audit tail for a human to read, not a transcript — and the whole serialized frame is then held to the hub's budget.
- Only payload fields shrink (`tmux_output`, `prompt`, `summary`, `title`, `reason`, `verdict_reason`). `task_id`, `task_gen`, `result`, `pr_url` and `verdict` always survive intact: a trimmed frame says less, never something different. An allowlist, not a denylist, so a field nobody has thought about yet stays protocol.
- A single line larger than the whole budget keeps its **tail** rather than being dropped — that is exactly the pi case. A truncated tail carries a visible marker line; a trim is logged.

**Protocol — the hub states the limit instead of enforcing it silently.** `auth_ok` now carries `max_message_bytes`, and the relay clamps to whatever the hub advertises (less a small headroom), falling back to 64 KiB for any hub that says nothing — which is every hub released before this. That answers the issue's "raise both sides together": a hub that raises its ceiling now carries its relays up with it, and the number cannot drift between the halves. Additive and `omitempty`; no protocol-version bump, since an optional field is backward-compatible by construction.

**Hub — log the disconnect it used to swallow.** `ErrReadLimit` is not a `*CloseError`, so `IsUnexpectedCloseError` never matched it: the hub dropped the connection with *nothing* in its log while the relay logged `code=1009 message too big` at the other end of the same socket. It now names the bound that was exceeded.

The 64 KiB ceiling itself is unchanged — whether it is still the right number is a maintainer call the issue flags, and this PR makes raising it a one-line change on the hub that relays pick up automatically.

Also pins `HIVE_HUB` in the relay test harness: the suite inherits a real contributor container's multi-hub value against its single test token and dies on the relay's one-token-per-hub `FATAL` at `require()` time, before its first assertion. That is how the suite behaved in the hive's own contributor pods.

## Tests

New coverage, each failing on the parent commit for the right reason:

- `bin/contributor-relay.test.js` — a headless `task_complete` whose output is one huge pi-shaped JSON line must fit the hub's limit and keep its `pr_url`. On the parent: `the frame the hub reads is 409920 bytes, past its 65536-byte read limit — this is the 1009 close`. Plus: a fitting frame is passed through untouched (same object, no behaviour change for the interactive path), protocol fields survive a clamp, an oversized `reason` is trimmed head-first, a line bigger than the budget keeps its tail, and the hub-advertised limit is honoured (including the floor for an absurdly small one, and the default for a hub that advertises nothing).
- `src/pkg/dashboard/contribute_ws_frame_limit_test.go` — `auth_ok` advertises `wsMaxMessageSize`, and an oversized frame both closes the connection with 1009 and is logged with the bound. On the parent, the log assertion fails with the exact diagnostic gap it closes: `an oversized frame disconnected a contributor with nothing in the hub log`.

Commands run:

```
node bin/contributor-relay.test.js          # 518/518 passed (511/511 on the parent)
bash bin/test_bin_suites_wired.sh           # 31 suites, 0 orphaned
cd src && go build ./... && go vet ./pkg/dashboard/
cd src && go test ./pkg/dashboard/ -count=1 # ok, 89s
gofmt -l on both touched Go files           # clean
```

Docs updated: a new section in `src/docs/contributor-relay.md`, and the `HIVE_RELAY_MAX_OUTPUT_BYTES` row in `src/docs/env-vars.md` now distinguishes what the relay *holds* from what it *sends*. Changelog fragment: `changelog.d/fixed-7932-relay-frame-limit.md`.

— hive: backend=claude
---
🐝 **Hive Agent**: `contributor` | **SHA:** `ec6c28919`


clubanderson:
Reviewed: 518/518 relay tests; `go test ./pkg/dashboard/` green incl. `TestWS_AuthOKAdvertisesMaxMessageBytes` / `TestWS_OversizedFrameIsClosedAndLogged`. Byte-boundary cuts via `Buffer.byteLength` stay valid UTF-8, tail keeps the END with a marker, protocol fields are allowlist-protected, bound is re-measured post-serialization. The reported failure mode is fixed.

One latent nit in `clampFrame` (bin/contributor-relay.js ~L770–790): the loop truncates fields in allowlist order regardless of which one is oversized, and `room` is computed with the still-huge later field inside `rest`, so it goes negative and `truncateFrameField(value, 0)` empties earlier fields that already fit (repro: small tail + normal summary + 200 KiB `reason` → `tmux_output: []`, `summary: ''`, no marker). Not reachable today except via an unbounded `verdict_reason`; I'll land a small follow-up that only shrinks a field by the actual deficit.

/lgtm
/approve

Issue #7925, opened 2026-09-20
✨ feature(container): the contributor image cannot run most repos' own checks — no pip/pytest/pyyaml/make/just/shellcheck; add a small baseline and honour a repo-declared toolchain


Danathar:
## The gap

`src/Dockerfile.contributor:36-38` installs `bash curl git jq tmux python3 ca-certificates gnupg openssh-client bubblewrap`, then Node and `gh`. Go is present. That is enough to run the relay and the CLI; it is not enough to run most repositories' own checks. Inventory of the running image tonight (Debian 12, v4.69.0 `:latest`):

| present | absent |
|---|---|
| python3, go, node, npm, jq, git, gh, curl, tmux | **pip / uv, pytest, pyyaml, jsonschema, requests, make, just, shellcheck, yq**, podman/skopeo/crane/oras/buildah |

So a container-mode agent can *edit* a Python or Make- or just-driven repo but cannot run its test suite, its `just check`, or its `make ci`.

## What it cost tonight (omp container mode, Sonnet 5 + Opus 5 xhigh advisor, projectbluefin repos)

- **utah#218** — PR body had to say *"`just`/`pip` were not available in this environment, so `just check` and `scripts/generate_skill_index.py --write` could not be run directly; the skill catalog was hand-synced"*. A generated file was edited by hand because the generator's runtime was missing.
- **utah#198 (review cycle)** — six wasted tool calls (`pip`, `python -m pip`, `venv`, `apt`, `node`, `bun`) trying to conjure a YAML parser to run one test whose delta was a substring; the advisor had to issue a `⟦blocker⟧` to stop it.
- **chairlift#188** — `CONTRIBUTING.md` requires `make ci` before opening a PR and the PR template has a `make ci` checkbox; no `make`, so it ran one Go test instead. Advisor concern posted under the verdict.
- **server#213 (review cycle)** — the agent ran pytest-style tests under `python3 -m unittest`, got `Ran 0 tests … OK`, and was about to cite it as green. The advisor caught it: *"That 'Ran 0 tests ... OK' is a false green — do not cite it."*

The last one is the argument that this is a correctness problem, not a convenience: when the repo's real checks cannot run, the agent either defers to CI (fine) or reaches for a proxy that *looks* green (not fine — this repo's reviewers police false verification claims, and the same contributor had to post a correction on chairlift#187 an hour earlier for exactly that).

## Proposal

[...]

Issue #2613, opened 2026-08-05
contribute: hive-contributor image composition measurements, and three low-risk size and update-cost options


castrojo:
Part of #2534

Downstream field note from `projectbluefin/review`. We build a separate, much narrower contributor image for our own use, and while measuring ours we ended up measuring `hive-contributor` beside it. The composition data looked worth passing along rather than keeping.

**To be clear up front: this is not a suggestion that you adopt or switch to our image.** Ours is deliberately a narrow single-backend task runtime and it would be a bad fit for `hive-contributor`, which has to support every backend `/contribute` offers. Everything below is scoped to changes inside your own `v2/Dockerfile.contributor`, and each one is independently adoptable. You may already have reasons for the current shape that we cannot see from outside.

## Measurements

All probes run **2026-08-05T00:20Z–00:30Z UTC** against `ghcr.io/kubestellar/hive-contributor:latest`, digest `sha256:706b08d1d22551bf9dc220df3d1629c441ed685148576c981748d3028a0ea357`, image created `2026-07-31T15:45:53Z`. Compressed figures are the sum of `linux/amd64` layer sizes from the OCI manifest; unpacked figures are from `du -sh` inside the running container.

| | compressed (pull) | unpacked | layers |
|---|---|---|---|
| `hive-contributor:latest` | 1250.6 MB | 2.7 GiB | 19 |

Largest contributors, unpacked:

| path | size | source |
|---|---|---|
| `/usr/lib/node_modules` | 1.3 G | `@github` 374M, `@openai` 350M, `@anthropic-ai` 263M, `@earendil-works` 174M, `bobshell` 102M, `npm` 15M |
| `/usr/local/bin/goose` | 288 M | single binary |
| `/usr/local/go` | 281 M | Go 1.24.4 toolchain |
| `/usr/lib/x86_64-linux-gnu` | 83 M | base |
| `/usr/lib/python3.11` | 27 M | base |

`/var/lib/apt/lists`, `/var/cache/apt`, `/usr/share/man` and `/usr/share/locale` are already at 0, and `/usr/share/doc` is 2.8M — the usual Debian cleanup is clearly already done, so there is nothing to suggest there.

## Observation 1 — the Goose release publishes a smaller artifact for the same version

[...]

PR #7936, opened 2026-09-21
✨ feature(container): ship a repository-check baseline in the contributor image

Danathar:
## What this changes

The contributor image (`src/Dockerfile.contributor`) can run the relay and the agent CLIs, but not the **checked-out repository's own checks**. It shipped no `pip`, `pytest`, `pyyaml`, `jsonschema`, `requests`, `make`, `just`, `shellcheck` or `yq` — so a container-mode agent could edit a Python-, Make- or `just`-driven repository and then not run its test suite, its `just check`, or its `make ci`.

This adds the small baseline proposed in the first half of #7925.

| | |
| --- | --- |
| apt | `make`, `shellcheck`, `python3-pip`, `python3-venv`, `python3-pytest`, `python3-yaml`, `python3-jsonschema`, `python3-requests` |
| pinned release binaries | `just` 1.58.0, `yq` (mikefarah) 4.53.6 |
| new | `/opt/hive/pyenv` — a `--system-site-packages` virtualenv owned by `dev`, first on `PATH` |

## Why it is a correctness fix, not a convenience one

An agent that cannot run the real check reaches for a proxy that *looks* green. From #7925: pytest-style tests were run under `python3 -m unittest`, reported `Ran 0 tests ... OK`, and were about to be cited as a passing run. Deferring to CI is a fine answer when a check cannot run locally; citing a false green is not, and this repository's reviewers police exactly that.

## Notes on the implementation

- **`python3-pip` alone would not have fixed "no pip".** Debian marks its system Python externally-managed (PEP 668) and this container runs as the unprivileged `dev` user, so a bare `pip install` fails twice over. Hence the venv: `--system-site-packages` so the apt-installed libraries stay importable through it, owned by uid 1000, first on `PATH` with `VIRTUAL_ENV` set to match. `pip install <anything>` now works inside a task with no privilege and no flags, and nothing outside the venv is modified.
- **`pytest` on `PATH` is a shim onto `python3 -m pytest`** in that venv. apt's `/usr/bin/pytest` has the *system* python3 in its shebang and would not see a plugin an agent just pip-installed, so the name and the module would disagree. A later `pip install pytest` simply replaces the shim with the real console script pointing at the same venv.
- **`yq` comes from mikefarah/yq v4, not apt.** Debian 12's `yq` 3.1.0 is a different project (the Python jq wrapper, kislyuk/yq); installing it would satisfy `command -v yq` while failing every `yq eval` a CI script actually writes. `just` has no Debian 12 package at all.
- **Supply chain.** Both binaries are fetched as a pinned release and verified with `sha256sum -c` *before* install, the way this file already installs the NodeSource key, Go, muse and omp — never a curl-pipe-to-shell installer. `JUST_VERSION` matches the pin `.github/workflows/v2-ci.yml` already uses, so CI and the container run the same `just`. Unlike the optional backend CLIs, a download failure here is **fatal** rather than tolerated: a missing backend is surfaced honestly at startup by `detect_cli`, but a silently missing `just` is surfaced nowhere and would reproduce the very failure this removes.
- **Layering.** Both new layers sit with the other rarely-changing apt content, above Node and the CLIs, so the file's stable-to-mutable ordering (#2613) is preserved and a CLI version bump still does not invalidate them.
- **Container tooling stays out.** `podman`/`skopeo`/`buildah`/`crane` need privileges the contributor container should not hold, exactly as #7925 argues.

## What is *not* in this PR

The **repo-declared toolchain layer** — the second half of #7925 — is deliberately left out, and I do not think it can land as specified without a maintainer decision:

- The proposal is "apt + pip lines" installed into the throwaway container. The contributor container runs as non-root `dev` with no `sudo`, so the **apt half cannot work at all** without granting the agent a privilege it does not have today. That is a threat-model change, not an implementation detail.
- There is also no checkout to read at the point the proposal names. The agent clones the repository itself as its first step (`contribute_task_prompt.go`), so "before the prompt is typed" there is nothing on disk to inspect — wiring this would mean hive pre-cloning, which is a separate structural change.

The pip half alone *is* now unblocked by this PR (`pip install -r requirements-dev.txt` into the venv works unprivileged), so a follow-up can implement it without touching the container's privileges — but I would rather that be its own reviewed change than a half-working feature shipped alongside the baseline. `src/docs/contributor-relay.md` now states this explicitly and points at the existing entrypoint-hook seam as the supported way to add tools meanwhile.

## Testing

```
$ go test ./pkg/config/ -run TestContributorDockerfile -v
--- PASS: TestContributorDockerfileInstallsPiWithoutCurlPipeShell
--- PASS: TestContributorDockerfileLinksClaudeNativeBinary
--- PASS: TestContributorDockerfileInstallsOmpPinnedAndChecksummed
--- PASS: TestContributorDockerfileShipsRepoCheckBaseline
--- PASS: TestContributorDockerfileMakesPipUsableByTheAgentUser
--- PASS: TestContributorDockerfileInstallsJustAndYqPinnedAndChecksummed
ok      github.com/hivecommons/hive/pkg/config

$ bash src/deploy/test_supply_chain_pins.sh
=== 17 passed, 0 failed ===

$ go vet ./pkg/config/        # clean
$ go test ./pkg/dashboard/ -run 'Dockerfile|Contributor'   # ok
```

The three new tests **fail on the parent commit** (verified by reverting `Dockerfile.contributor` and re-running):

```
--- FAIL: TestContributorDockerfileShipsRepoCheckBaseline
    Dockerfile.contributor no longer apt-installs "make" — a container-mode agent
    cannot run the repository's own checks without it (#7925)
--- FAIL: TestContributorDockerfileMakesPipUsableByTheAgentUser
    Dockerfile.contributor missing "python3 -m venv --system-site-packages /opt/hive/pyenv"
--- FAIL: TestContributorDockerfileInstallsJustAndYqPinnedAndChecksummed
    Dockerfile.contributor missing ".../just-${JUST_VERSION}-${JUST_TARGET}.tar.gz"
```

They assert against the **apt install list itself**, not against the file text, so a tool named only in a comment or in a build-time verification step cannot stand in for the package it is meant to prove.

**The image itself builds.** `build-contributor` is green on **both** architectures for this branch — [linux/amd64](https://github.com/hivecommons/hive/actions/runs/35550811622/job/106185099206) (3m23s) and [linux/arm64](https://github.com/hivecommons/hive/actions/runs/35550811622/job/106185099359) (4m8s). That is the strongest evidence available here, because both new layers end in verification steps that fail the build loudly rather than shipping a silently incomplete image: `make --version`, `shellcheck --version`, `import yaml, jsonschema, requests`, `pytest --version`, `pip --version`, `just --version`, `yq --version`, plus both `sha256sum -c` gates — all of them ran and passed on amd64 and arm64. I have no container runtime locally, so this was the first actual build; I had pre-verified out of band that every Debian package name resolves on `packages.debian.org/bookworm`, that `python3-pytest` ships `/usr/bin/pytest`, that both release URLs download with the four digests recorded here (the `yq` pair matches the release's published `checksums` file, the `just` amd64 digest matches the one `v2-ci.yml` already pins), and that `chown -R` with the default `-P` does not dereference symlinks (coreutils `chown.c` sets `affect_symlink_referent = false` whenever `-R` is used with `FTS_PHYSICAL`), so chowning the venv cannot reach `/usr/bin/python3`.

What a green build does **not** prove is runtime ergonomics inside a real task — that an agent's `pip install -r requirements-dev.txt` lands in the venv and its `pytest` picks up the plugin. That needs a live container-mode task against the built image.

Refs #7925


clubanderson:
Reviewed: just 1.58.0 and yq v4.53.6 SHA-256 pins verified against upstream SHA256SUMS/checksums-bsd for both arches; JUST_VERSION matches v2-ci.yml; new layers precede `useradd dev`, numeric chown, PATH change leaves node untouched; `go test ./pkg/config/` green.

/lgtm
/approve

PR #7939, opened 2026-09-21
🐛 fix(gh-wrapper): inject the identity footer into every body flag gh accepts

### LINKED ISSUE #7937: 🐛 bug(gh-wrapper): identity footer injection only knows --body/--body= — 'gh pr comment -F file' fails (two body flags) and '-b text' would post only the footer
## What happened

omp container mode, relay image v4.69.0, PR review cycle 2026-09-21T01:21–01:34Z. The agent tried to post a multi-paragraph PR comment with `gh pr comment -F <file>` and it failed; the advisor's note under the verdict: *"`-F` can't work here — something is already supplying `--body` (the same wrapper that appends the attribution footer)…"*. The agent fell back to inlining the body.

## Why

`bin/gh-wrapper.sh` `_inject_identity()` (`:941-969`) appends the hive identity footer by rewriting the argument list. It recognises exactly two spellings of the body — `--body <val>` (`:948`) and `--body=<val>` (`:954`) — and when it sees neither, it **adds** `--body "<footer>"` (`:965-966`). It runs for `issue/comment|pr/comment` (`:1134-1135`) and the create paths (`:1037`, `:1119`).

Two failure modes follow:

1. **`-F <file>` / `--body-file <file>`** → the wrapper adds `--body <footer>`; gh rejects the pair (`specify only one of --body or --body-file`). The comment is not posted at all. This is the case observed.
2. **`-b <text>`** (gh's documented short flag for `--body`) → not matched either, so the wrapper adds a second body flag. gh's flag parser keeps the last value, so the comment posts with **only the footer** — the agent's text is silently dropped. Not observed tonight, inferred from the same code path; worth a test.

Agents reach for `-F` precisely when the comment is long or contains quotes/backticks — i.e. the comments most worth posting correctly. `--body-file` is also the shape the relay's own advice and many repos' CONTRIBUTING docs suggest for PR bodies.

## Fix

In `_inject_identity`, handle all four spellings:

- `-b <val>` / `-b<val>` → same as `--body`.
- `-F <file>` / `--body-file <file>` / `--body-file=<file>` → read the file, append the footer, and pass the result as `--body` (or write a temp file and rewrite the `--body-file` argument — keeping the file form avoids any argv length limit on long bodies). `-` (stdin) needs the same treatment: read stdin once.

Add the four cases to the wrapper's test file next to the existing `--body` cases.

Cites at v4 `7dde068`.

Danathar:
## What

`bin/gh-wrapper.sh` `_inject_identity()` recognised exactly two spellings of the
body — `--body <val>` and `--body=<val>` — and, when it saw neither, **added**
`--body "<footer>"`. Every other spelling `gh` accepts therefore reached `gh`
carrying two body flags:

| invocation | what happened |
| --- | --- |
| `gh pr comment -F notes.md` | `gh` refused the call: *"specify only one of --body or --body-file"*. **The comment was never posted.** (Observed live, 2026-09-21, omp container mode.) |
| `gh pr comment --body-file notes.md` | same |
| `gh pr comment -b 'text'` | two `--body` flags; `gh` keeps the **last**, so the comment posted as the **footer alone** and the agent's text was dropped with no error at all |

`-F` is precisely what an agent reaches for when the comment is long or contains
quotes and backticks — i.e. the comments most worth posting correctly — and
`--body-file` is the shape the relay's own advice suggests for PR bodies.

## How

`_inject_identity()` now handles all six spellings — `--body`/`-b` and
`--body-file`/`-F`, each separated, attached (`-btext`, `-Fnotes.md`) or
`=`-joined — plus `-` for stdin, and emits one normalised `--body` carrying the
caller's text followed by the footer.

A file or stdin body is **read here and re-emitted as `--body`** rather than
rewritten to a temporary `--body-file`. Most call sites in this `case` end in
`exec`, so no EXIT trap could ever clean such a temp file up, and argv size is
not the binding constraint: GitHub caps a body at 65536 characters, and the
relay this hands off to already passes the whole body through argv itself
(`hive-open-issue.sh` does `BODY="$(cat "$BODY_FILE")"`). Reading stdin in the
wrapper is also what makes `-F -` work at all — the body must be consumed
exactly once, and the process that would otherwise read it is replaced by that
same `exec`.

An **unreadable** body file is passed through untouched, with no `--body` added
on top of it, so `gh` (or the relay) reports "file not found" in its own words
instead of a confusing flag conflict.

One adjacent line changed for the same reason: the `pr/review` fallback used to
`exec "$REAL_GH" "$@"` — the *original* argv — which both dropped the footer it
had just injected and, now that the wrapper consumes a `-F -` body itself, would
have handed `gh` a stdin already at EOF. It now execs `"${args[@]}"`.

The three relays downstream (`hive-open-issue.sh`, `hive-open-pr.sh`,
`hive-review.sh`) already understand `--body`/`-b`/`--body-file`/`-F`; the
wrapper was the remaining gap.

## Testing

`bin/gh-wrapper.test.sh` gains a `#7937` section — 13 cases asserting on **the
argv `gh` actually receives**, recorded byte-for-byte by the mock (one file per
argument: bodies are multi-line and contain quotes, backticks and emoji). The
argv is what has to be asserted here, because the `-b` failure mode **exits 0**:
a test checking only the exit code would have passed throughout the bug's life
and would pass again if it came back.

Nine of the cases fail on the parent commit and pass on the fix (`-F <file>`,
`--body-file <file>`, `--body-file=<file>`, `-F<file>`, `-F -`, `-b <text>`,
`-b<text>`, `pr create -F`, `pr review -F`, unreadable file); the other three
pin the spellings that always worked (`--body`, `--body=`, and no body flag at
all, which must still add one).

```
$ bash bin/gh-wrapper.test.sh
...
=== Identity footer injection knows every body spelling (#7937) ===
PASS: pr comment -F <file> keeps the body and adds the footer (the reported failure)
PASS: pr comment --body-file <file> keeps the body and adds the footer
PASS: pr comment --body-file=<file> keeps the body and adds the footer
PASS: pr comment -F<file> (attached short form) keeps the body and adds the footer
PASS: issue comment -F - reads stdin once and adds the footer
PASS: pr comment -b <text> keeps the text (it used to post the footer alone)
PASS: pr comment -b<text> (attached short form) keeps the text
PASS: pr create -F <file> keeps the body and adds the footer
PASS: pr review -F <file> reaches gh with the injected body
PASS: pr comment --body <text> still gets exactly one footer
PASS: pr comment --body=<text> still gets exactly one footer
PASS: pr comment with no body flag still gets --body <footer>
PASS: pr comment -F <missing file> passes the flag through without adding a second body

Results: 56 passed, 0 failed
```

Against the parent commit (fail-first check), with the fix reverted and the new
tests kept: `Results: 46 passed, 9 failed`.

Also run: `bash bin/test_hive_open_issue.sh` (44 passed), `bash
bin/test_hive_review.sh` (24 passed), `bash -n` on both changed scripts.
`bash bin/test_hive_open_pr.sh` reports 47 passed / 1 failed ("fallback body
round-trips newline/tab/backslash/quote", the no-`python3` JSON escaper) — that
failure reproduces unchanged on `upstream/v4` with this branch stashed, so it is
pre-existing and untouched by this PR.

Fixes #7937


clubanderson:
Reviewed: `bash bin/gh-wrapper.test.sh` 56/56; `_inject_identity` is scoped to issue/pr create|comment|review so `gh api -F` is unaffected; `-B` is case-distinct; `pr/review` fallback switching to `${args[@]}` is the right fix for the drained stdin.

/lgtm
/approve

Issue #7996, opened 2026-09-21
🐛 A contributor whose Claude login has expired keeps reporting ready, fails every task on the 30-minute watchdog, and nothing tells the operator


Danathar:
**A contributor with a dead Claude session ran all night, was handed nine tasks, failed every one after 30 minutes with "no observed progress", and the dashboard showed nothing but green.**

## What I saw

On my host Hive (`hive-wild-mole`, hub on `:candidate`, contributor container `ghcr.io/hivecommons/hive-contributor:latest` with the claude backend) the contributor's Claude CLI lost its session some time before 23:23 EDT on 2026-09-20. From then until I noticed at 07:40 EDT the next morning:

| Time (EDT) | Task | Outcome |
|---|---|---|
| 23:23 | goodreads-mcp#111 | assigned, "abandoned without completion" 0 s later |
| 23:33 | goodreads-mcp#111 | assigned, failed at 00:03 |
| 00:13 | goodreads-mcp#111 | assigned, failed at 00:43 |
| 00:43 | sensi#241 | assigned, failed at 01:13 |
| 01:23 | sensi#241 | assigned, failed at 01:54 |
| 02:04 | sensi#241 | assigned, failed at 02:34 |
| 02:34 | zfs-kinoite-complex#221 | assigned, failed at 03:04 |
| 04:38 | sensi#242 | assigned, failed at 05:08 |
| 06:43 | goodreads-mcp#111 | assigned, failed at 07:13 |
| 07:23 | goodreads-mcp#111 | assigned (fifth attempt) |

Every failure carried the same reason from the relay:

```
Task ct-Danathar/goodreads-mcp-111-1789961600 failed [environment]: no observed progress for 30min — the agent CLI is not visibly working
Relaunching claude after a task exit: claude --dangerously-skip-permissions ...
CLI ready — accepting tasks
```

The tmux pane inside the container showed what was actually going on. The task prompt had been typed in, and directly under it:

```
● Login expired · Please run /login
✻ Churned for 0s
❯
```

So the CLI answered every prompt with "Login expired", returned to its prompt, the relay read that prompt as "CLI ready", told the hub `ready for work` (the hub journal has two of those a minute all night), accepted the next lease, and the cycle repeated every 30 minutes until a human ran `/login`.

## What the operator saw

[...]

Issue #6541, opened 2026-09-10
🐛 bug: agy quota exhaustion is invisible to the relay — tasks are dispatched into a quota-blocked CLI and misreported as [environment] failures


Danathar:
## What happened?

An `agy` contributor completed 14 PRs in four hours, then hit its provider quota. The relay could not see that, kept dispatching tasks into a CLI that could not serve them, and booked each one as an `[environment]` failure twenty minutes later — blaming the contributor's machine for a provider rate limit.

Pane at the moment of failure (`tmux capture-pane -t contributor -p`):

```text
> You are a contributor to the projectbluefin/lab hive. Work on issue projectbluefin/lab#468 …
▸ Thought for 3s, 499 tokens
  Initiating repository fork and clone
● Bash(echo "HIVE_WORKSPACE_DIR: $HIVE_WORKSPACE_DIR"; pwd; ls -la)
▸ Thought for 2s, 633 tokens
  Verifying Workspace Directory
● Bash(ls -la /home/dev/workspace)
⚠ Individual quota reached. Please upgrade your subscription to increase your limits. Resets in 38m29s.
Error ID: cd45be83-870d-4c25-b297-c608fa533664-12
 How's the CLI experience so far? Help us improve:
 [1] Good  [2] Fine  [3] Bad  [0] Skip
```

agy took the task, ran two tool calls, hit the quota, and stopped. Relay log for the same task and the two before it:

```text
Task prompt sent to CLI
Pane unchanged for 20+ minutes — confirming before giving up on ct-projectbluefin/documentation-1139-… (1/2)
Task ct-projectbluefin/documentation-1139-… failed [environment]: no pane activity for 20+ minutes, confirmed over 2 checks — the agent CLI is not visibly working
…
Task ct-projectbluefin/review-401-… failed [environment]: no pane activity for 20+ minutes …
…
Task ct-projectbluefin/lab-466-… failed [environment]: no pane activity for 20+ minutes …
```

Three tasks lost in an hour, each holding a lease for its full 20-minute stall window, and the relay kept accepting more.

## Why it is invisible

`paneShowsUnretryableAPIError()` in `bin/lib/pane-classifier.js` is gated on `"API Error:"` chrome:

```js
const lower = line.toLowerCase();
if (!lower.includes('api error:')) return false;
if (UNRETRYABLE_API_ERROR_PATTERNS.some((pat) => lower.includes(pat))) return true;
```

[...]

PR #8004, opened 2026-09-21
🐛 fix: park the relay when the agent CLI is signed out instead of failing every task on the watchdog

Danathar:
## What this fixes

A contributor's Claude session expired at 23:23 and the relay worked the same wall until the operator noticed at 07:40. Nine assignments; nine identical failures reading `[environment] no observed progress for 30min — the agent CLI is not visibly working`; nine hive issues put in failure cooldown. The pane said exactly what was wrong for eight hours:

```
● Login expired · Please run /login
✻ Churned for 0s
❯
```

Three things made that a loop rather than one bad task:

1. **The pane at rest is indistinguishable from a healthy one.** The CLI refused the prompt and returned to full ready chrome, which is what readiness reads, so the relay reported `ready` twice a minute all night.
2. **The refusal was classified as "the agent asked a question."** `BLOCKED_ON_HUMAN` is right for a question somebody can answer while the work stays parked. Here the work never started, so the pane never changed again, the progress lease ran down, and the task was handed back thirty minutes later described as a *dead host* — which is what `not visibly working` means to whoever reads it.
3. **Nothing said any of it outside the container.** The dashboard's green AGENT dots are the hub's *own* agents, which have their own login and were fine. Every number on screen was accurate and the picture was wrong.

## What changed

Relay-side, modelled directly on the provider-quota hold (#6541) — the other condition that refuses every task rather than this one.

- **`paneLoginWallLine()`** (`bin/lib/pane-classifier.js`) picks the CLI's own sign-in refusal out of the visible tail, under the same two-independent-signals rule the quota and unretryable detectors already use: the CLI's own chrome (`●`, `⚠`, `Error:`, `API Error:`) **plus** the wording. This repository's sources contain every one of those strings, and an agent reading them into its pane must not park its own relay. A 403 or a quota banner on the same line is vetoed and keeps its existing bucket — #4400's rule that `/login` fixes a 401 and fixes nothing about a 403 is unchanged.
- **The task is handed back at once**, within a progress interval instead of at the watchdog, with a reason that names the backend, quotes the CLI's line verbatim, and gives the attach command. That reason string is the whole operator surface for this condition, so it carries the remedy.
- **`ready` is withheld** until a person has been at the tmux session — at the same single choke point the quota hold uses, so no future `ready` call site can reintroduce it — and a pushed assignment is declined the same way, so a hub offer does not walk into the same wall.
- **The relay keeps saying why.** The hold restates its banner on a fixed cadence rather than printing once and scrolling away, and the relaunch log no longer prints `CLI ready — accepting tasks` while the hold stands.
- **Confirmed before acted on**, over consecutive ticks, exactly as the pane-stall and CLI-death detectors are. And while somebody is *actively* at the pane (a recency question, not a connection one — #5277) the task is held for them rather than taken away, which preserves what #5094 established for this same pane; #7996 only bounds how long that wait can last.
- **Release** is tmux client activity newer than the hold — somebody attached and typed, which is the act the banner asked for. Not proof the sign-in worked, and it does not need to be: releasing costs one task, and that task re-arms the hold if the wall is still there. A six-hour ceiling backs it up so a tmux that cannot report client activity at all costs one task per window instead of the contributor's place in the fleet.

## What is deliberately not here

The issue's second and third bullets are hub and dashboard changes, not relay ones, and I have left them: there is still no `not-ready` contributor state on the wire, a held lease on a signed-out contributor still reads as "Implementing" on the triage ladder, and there is no hub-side per-contributor consecutive-failure breaker. The relay simply stops asking, which is what stops the loop — and the failure reason it now sends is what puts the cause in the Operations feed. Both remaining pieces are noted in the docs section this PR adds.

On the open question in the issue — no, there is no way for the relay to run `/login` non-interactively from a host credential for the claude backend, so a person is always needed. The notification really is the fix, which is why the banner, the failure reason, and the repeat cadence are all part of this.

## Testing

`node bin/contributor-relay.test.js` — **552/552 pass**. Ten new tests, all of which **fail on the parent commit** (verified: 542/552 there):

- `paneLoginWallLine` returns the incident's line; refuses the same wording in prose; refuses a 403 and a quota banner
- a signed-out CLI is not read as ready, and the mid-session wording is chrome-gated against prose
- the wall is handed back with the real reason, not the 30-minute watchdog, and no `ready` follows
- a person actively at the pane keeps the task instead of parking the relay
- a tick without the wall resets the confirmation count
- the progress-lease expiry names the wall instead of "not visibly working"
- `ready` is withheld at the choke point across reconnect and task-exit
- a pushed assignment during a hold is declined
- the hold ends when somebody has been at the pane, and is bounded when nobody ever comes

Also run: `node --check` on both files, `bash -n bin/contributor-relay.sh`, `bash bin/test_bin_suites_wired.sh`, `bash bin/test_backend_smoke.sh` (all pass). `bash bin/contributor-agent.test.sh` fails on this machine at an unrelated pre-existing assertion (`expected missing codex binary to be NOT_INSTALLED; got: NOT_AUTHED`) — it fails identically on the parent commit.

Docs: a new section in `src/docs/contributor-relay.md` next to the quota-hold one, and a changelog fragment.

Refs #7996

---
🐝 **Hive Agent**: `contributor` | **SHA:** `ac5cf2f18`


clubanderson:
Heads-up: this collided with #8002, which merged to `v4` at 12:28Z (about 12 minutes before this opened) for the same #7996 and is why the branch now shows conflicts in `bin/contributor-relay.js` / `.test.js`.

#8002 landed the same core shape you describe — login hold modelled on #6541: unattended `/login` wall confirmed over `LOGIN_WALL_GRACE_TICKS` ticks → immediate `[environment]` hand-back naming `/login`, `ready` withheld at the `sendTo` choke point, pushed assignments declined, no relaunch over the evidence, human-at-pane keeps #5094's blocked-on-human — plus the hub half (a stall streak in `contribute_failure_streak.go`: 3 consecutive ≥15-min failures pause the identity with a `/login` message).

What this PR has that #8002 does not, and that I think is worth keeping:

1. **Chrome-gated detection** (`paneLoginWallLine` in `pane-classifier.js`) — requiring the CLI's own `●`/`⚠`/`Error:` chrome alongside the wording, so an agent reading this repo's sources into its pane can't park its own relay. #8002 reuses `paneShowsLoginRequiredError`, which is wording-only.
2. **`classifyReadiness` refusing a signed-out pane** — #8002 leaves the readiness latch alone and relies on the hold.
3. **Banner repeat cadence** and suppressing `CLI ready — accepting tasks` while held.
4. **Release by tmux client activity newer than the hold, with a 6-hour ceiling.** #8002 releases on the pane showing a signed-in state (`Login successful`/`Logged in as`, or a ready pane with a human present) — no ceiling. Your activity-based release plus ceiling is arguably more robust; happy to take either or both.
5. The `contributor-relay.md` section.

Suggest rebasing onto `v4` and reducing to those deltas on top of #8002's `enterLoginHold`/`probeLoginHold`/`releaseLoginHold` (exports are in the test hook block, `#7996` tests are next to the `#5094` ones). If you'd rather I fold them in, say so and I'll open it as a follow-up crediting this PR.

Danathar:
Superseded by #8002, which merged to `v4` while this was in flight and closed #7996.

#8002 covers the same relay behaviour — chrome-gated detection of the sign-in wall, fast `[environment]` failure naming `/login` instead of the 30-minute watchdog, `ready` withheld at the send choke point, pushed assignments declined, and a grace period so a human at the pane keeps the #5088 blocked-on-human path — and it also does the hub half this PR deliberately left out (the stall-failure streak in `contribute_failure_streak.go`). It additionally keeps the CLI un-relaunched so the wall stays on the pane as evidence, which is a better call than what is here.

Closing rather than rebasing: this would be a competing rewrite of code that just landed. Three small things in here were not in #8002, and I am noting them rather than re-opening the fix — happy to send any of them as a focused follow-up if a maintainer wants one:

1. The wall detector in #8002 (`paneShowsLoginRequiredError`, plus a `use /login` substring) is not chrome-gated, so an agent whose pane carries the phrase in prose — this repository's own sources contain it — can count toward the grace. The version here required the CLI's own `●`/`⚠`/`Error:` chrome *plus* the wording, and vetoed a line that also carried a 403 or a quota banner so #4400's ordering could not be crossed.
2. The progress-lease expiry path (`onTaskProgressLeaseExpired`) still reports "no observed progress for 30min — the agent CLI is not visibly working" when the wall has scrolled out of the tick window, or when the tick loop is not running at all — which is the case that timer exists for.
3. `classifyReadiness` for claude matches "Please run /login" but not a bare "Login expired", and `src/docs/contributor-relay.md` has no section for the new hold next to the quota one.

Issue #7896, opened 2026-09-20
🌱 Repositories card: a held PR counts in the PRs number but never appears in the pills, so an operator cannot see what is waiting on them


Danathar:
**I want the repo card to show held PRs alongside the actionable ones, visibly marked as held, because those are exactly the ones waiting for me and today they are the only PRs the card hides.**

## What an operator sees today

On my hive the `Danathar/arch-bootc` card says **3 PRs** with a sparkline, and under it three pills: `#313`, `#312`, `#316`. Those three are *issues*. None of the three open PRs (#315, #311, #143) is listed, and nothing on the card says why. The `Danathar/sensi` card says **1 PR** and lists nothing at all.

My first read was "the dashboard is missing my PRs". What is actually going on, from the code:

- The number is the repo's total open PR count (`TotalByRepo[repo].PRs` in `github/client.go`, wired through `status_builder.go` as `FrontendRepo.PRs`).
- The pills come from `FrontendRepo.OpenPrs`, which is filled from `actionable.PRs.Items` only. `fetchPRs` moves any PR carrying the `hold` label out of that set into `PRs.Held` / `Hold.Items`.
- All three arch-bootc PRs and the sensi one carry `hold`: two are the ACMM level gate ("held for human review by the hive's ACMM level gate"), one is a Renovate bump held with `needs-human`. So the count says 3 and the list says nothing.
- Held items do exist in the payload, but the only place the UI shows them is the tooltip on the fleet-wide HOLD stat at the top of the page (`holdItems` in `index.html`), which lists them as text across all repos.

So the card is internally consistent, but the thing it hides is the thing I most need to act on: an ACMM hold is by definition a PR waiting for a human, and this card is where a human looks.

## What "fixed" looks like

- **Two columns of pills, matching the two stat columns already on the card.** Issue pills sit under the issues number (blue, as today) and PR pills sit under the PRs number (purple, the PR sparkline's colour). Position then says what a pill is, so three issue pills under a "3 PRs" number can never read as the PRs again.
- **Colour within each column means state, not kind.** Held PRs and held issues are shown in their column with a distinct held tint (the card already tints `needs-human` pills differently from mergeable ones, `repo-pr-pill needs-human` vs `mergeable` in `index.html`, so this is the same pattern), with a tooltip saying who holds it and why: ACMM level gate, `needs-human`, or a plain `hold` label.
- The count and its column then agree: 3 PRs, 3 pills in the PR column, 3 of them tinted held.
[...]

PR #8007, opened 2026-09-21
✨ feature(dashboard): drag a repository card wider to read its pill titles

### LINKED ISSUE #8000: 🌱 Repositories page: let me drag a repo card wider or narrower so the pills show their whole title
**I want to resize any one repository card on the Repositories page by dragging its edge, so I can read the full text of the pills in that card without hovering each one.**

On a hive that watches a lot of repos (my projectbluefin hive has six cards across and up to 28 PRs and 91 issues per card) every card is the same width, set by the grid: `.repo-grid { grid-template-columns: repeat(auto-fit, minmax(240px, 1fr)) }` in `index.html`. The pills inside are two to a row, and the title on each is cut to about ten characters (`MAX_PILL_TITLE_LEN = 50` is then clipped again by the pill's `max-width: 100%`), so a column reads as `#192 Pa...`, `FIX #1109 fix...`, `TESTS #106...`. To know what any of them is I hover one at a time or open the repo on GitHub. When I am triaging a card with 30 pills that is the whole job.

I do not want every card wider; the six-across overview is right for most repos. I want to widen the one I am working in, and narrow the quiet ones to make room.

## What "fixed" looks like

- Each card has a drag handle on its right edge (the usual cursor `col-resize` affordance). Dragging it changes that card's width; the grid reflows the others.
- A wider card shows more of each pill title (the pill row can grow to three or four across, or the titles can simply run longer), so at some width the titles are complete and the hover is unnecessary.
- Widths persist per repo in the browser (localStorage keyed by hive plus repo full name is enough), so the layout I set survives a reload and the governor's repaint.
- Double-click on the handle resets that card to the default width. A small "reset layout" affordance next to the Rescan button resets all of them.
- Keyboard: the handle is focusable and arrow keys nudge the width, for the same reason the other controls are keyboard-reachable.

## Notes for whoever picks this up

- The cards repaint on every governor cycle (`renderRepos` rebuilds the grid HTML), so the width has to be reapplied after each repaint from the stored map, not set once on the element.
- CSS grid with `auto-fit` sizes tracks, not items; a per-card width probably means switching the container to flex-wrap with explicit `flex-basis` per card, or keeping grid and setting `grid-column: span N` in steps. Either is fine; the drag should feel continuous, not snap in 240px jumps.
- The pill truncation is a separate knob from card width. If the card is wide enough, `MAX_PILL_TITLE_LEN` should stop cutting the title at 50 characters; a wider card with the same ten-character clip gains nothing.
- Related: #7896 (held PRs missing from the pill row and two pill columns per card). If that lands first, the resize should apply to the two-column layout.

## Open questions

- Should widths be per repo (survive the card moving position) or per column position? Per repo is what I expect.
- Is there any reason to sync the layout server-side (per operator) rather than per browser? Per browser is enough for me.

Danathar:
Fixes #8000

## The problem

Every card on the Repositories page is exactly as wide as every other one, because `.repo-grid` is a CSS grid with `repeat(auto-fit, minmax(240px, 1fr))` — and grid sizes **tracks**, not items. On a hive watching six repos with up to 28 PRs and 91 issues per card, a pill column reads `#192 Pa…`, `FIX #1109 fix…`, `TESTS #106…`. The only way to learn what one of them is is to hover it, one at a time, or open the repo on GitHub. Triaging a card with 30 pills is that hover, thirty times.

The six-across overview is right for most repos, so this does not widen them all. It lets an operator widen the one card they are working in and narrow the quiet ones to make room.

## What changed

All of it in `src/pkg/dashboard/static/index.html`.

- **Flex, not grid.** `.repo-grid` becomes `display: flex; flex-wrap: wrap`, and `.repo-card` carries `flex: 1 1 240px`. Flex sizes items, so one card can hold its own `flex-basis` while the rest reflow around it. The default basis reproduces what `auto-fit` did — a line takes as many cards as fit at 240px and they share the leftover evenly — so an untouched hive looks exactly as it did. Light mode's tighter 200px packing moves from the grid's `minmax()` onto the item.
- **A drag handle on each card's right edge**, with the usual `col-resize` cursor, modelled on the sidebar splitter next door. The drag is continuous, not snapped to 240px steps.
- **Double-click the handle resets that card.** A **"Reset layout"** button beside Rescan resets all of them; it only appears once there is a layout to reset.
- **Keyboard.** The handle is focusable and reports the width it controls (`role="separator"`, `aria-orientation="vertical"`, `aria-valuenow`). ← / → nudge by 16px, Shift by 64px, Home/End go to the ends, Esc resets. It is hidden in the phone layout, like `.oc-sidebar-resize`, where there is no width to trade between cards and an 8px edge strip would only eat taps meant for the pills.
- **Persistence.** One `localStorage` map per hive (`hive-repo-card-widths:<hiveId>`), keyed by repo full name. Widths survive a reload and follow the repo when the list reorders.
- **Reapplied on every repaint, not set once.** `renderRepos()` rebuilds the grid HTML each governor cycle, so the width is emitted *with* the card HTML from the stored map rather than assigned to the node afterwards — one source of truth however the width was changed. A repaint landing mid-drag is held, since replacing the node under the pointer would drop the pointer capture with it.
- **`MAX_PILL_TITLE_LEN` is no longer a fixed 50.** The character budget grows with the card (capped at 240), because a wider card with the same ten-character clip gains nothing — which is the other half of the issue.

The two pill columns from #7896 are untouched: the resize applies to that layout, and position still says whether a pill is an issue or a PR.

## The open questions in the issue

- **Per repo or per column position?** Per repo, as expected — the width follows the repo when the list reorders. `TestRepoCardWidthStore` pins that the map is keyed by repo full name.
- **Sync server-side?** No. This is a view preference, not hive state: per browser needs no write permission and no round trip. Two hives open in one browser get separate maps, which is also tested.

## Tests

`src/pkg/dashboard/repo_card_resize_8000_test.go`:

- `TestRepoGridSizesItemsNotTracks` — the flex layout is in place and **both** track-sizing rules are gone; either one left behind would pin every card to the same width again and the drag would do nothing visible.
- `TestRepoCardResizeHandleMarkup` — the handle's markup and ARIA, the width reaching the card in the rendered HTML, the repaint hold during a drag, and that the fixed `const MAX_PILL_TITLE_LEN = 50;` is gone.
- `TestRepoCardResizeInteractionsAreWired` — drag, double-click reset, reset-all through the generic `data-action` dispatcher (no inline handler attribute — CSP `script-src-attr` is `'none'`), and the keyboard bindings.
- `TestRepoCardWidthStore` — the store and truncation rules **executed** under node against a stubbed `localStorage`: clamping and rounding, the per-hive key, round trip, two hives not colliding, reset deleting the entry rather than storing a zero, the key disappearing when the last card is reset, a corrupt/non-object/null store reading as empty without throwing, and the title budget growing with the card and staying capped. It skips with an explicit note when `node` is absent; the structural tests still run.

```
$ go test ./pkg/dashboard/
ok  	github.com/hivecommons/hive/pkg/dashboard	92.290s

$ go build ./... && go vet ./pkg/dashboard/
(clean)
```

The JS was also syntax-checked with `node --check` on the extracted `<script>` body, since the Go compiler cannot see it.

Not verified in a browser — there is no display in this environment. The layout equivalence argument for an untouched hive (`flex: 1 1 240px` reproducing `minmax(240px, 1fr)` under `auto-fit`) is reasoned from the spec and pinned by test, not observed; a reviewer with the dashboard open is the right check on the visual result.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `39e27a00f`


clubanderson (approved):
Reviewed against `v4` locally: `node .github/scripts/check-inline-js.js` (syntax + no-undef) green, `go test ./pkg/dashboard/` green, `#7896` pill-column tests unaffected.

Design notes I checked and agree with:
- flex-wrap + `flex: 1 1 240px` reproduces auto-fit for untouched hives; `.repo-card-sized { flex-grow: 0 }` keeps the dragged width without stopping a lone card from shrinking to the viewport.
- `renderRepos` returns early while `_repoResizeDrag` is set, so a governor repaint cannot replace the handle holding pointer capture; widths are baked into the markup from the stored map so `setIfChanged` stays stable across cycles.
- Persisting the drag's *requested* width rather than the measured one avoids the narrow-viewport ratchet — good catch.
- `pillTitleLimit` scaling the character budget with card width (rather than a single threshold) means the drag pays off continuously.
- Per-hive storage key and click-without-move as a no-op are both nice.

One low note, not blocking: `.repo-card { min-width: 0 }` lets a card be squeezed below `REPO_CARD_MIN_W` (180px) when several sized cards share a line narrower than their sum — acceptable since the stored width is untouched and they recover on a wider viewport.

/lgtm
/approve

PR #8008, opened 2026-09-21
✨ dashboard: drag-to-resize Repositories cards (#8000)

### LINKED ISSUE #8000: 🌱 Repositories page: let me drag a repo card wider or narrower so the pills show their whole title

clubanderson:
Closes #8000

Repositories cards get a drag handle on the right edge so an operator can widen a card until its pill titles are whole, instead of hovering each pill.

## What

- `.repo-grid` switches from grid `auto-fit` (which sizes tracks, not items) to flex-wrap; default cards flex from `--repo-card-basis` (240px dark / 200px light) exactly as the old `minmax(…, 1fr)` track did, and a dragged card gets an explicit `flex-basis` with `flex-grow: 0`, so the drag is continuous and the width set is the width kept.
- Per-card `.repo-resize-handle` (`col-resize`, `role="separator"`, `tabindex="0"`): pointer drag with capture; double-click / `Home` resets; `←`/`→` nudge 16px (`Shift` 64px); `End` takes the grid width.
- Header-level **↔ Reset layout** button next to Rescan, shown only when any width is stored, carries `data-stop="1"` like its neighbours.
- Widths persist per repo full name in `localStorage` (`hive-repo-card-widths`) and are re-read from that map on every governor repaint — `renderRepos` bakes the width into the card markup, never sets it once on an element. It also returns early mid-drag so the repaint can't replace the handle holding pointer capture.
- A card at/past 480px lifts the 50-character pill clip (`MAX_PILL_TITLE_LEN` → `Infinity`); crossing that threshold triggers one repaint after the drag ends.

## Tests

- `repo_card_resize_8000_test.go` pins the flex-wrap grid, the per-card handle markup, the stored-width read on repaint, the mid-drag guard, the pointer/dblclick/keyboard listeners, and the Reset layout button (wired + `data-stop`).
- `node .github/scripts/check-inline-js.js src/pkg/dashboard/static/index.html` — syntax + no-undef green.
- `go test ./pkg/dashboard/` green (existing #7896 pill-column tests unaffected).


clubanderson:
Superseded by #8007, which opened first for #8000 with the same shape (flex-wrap grid, per-card handle, localStorage widths re-read on repaint, header Reset layout) plus a per-hive storage key, a proportional pill-title budget and aria-valuenow. Reviewing that one instead.

Issue #8011, opened 2026-09-21
🌱 dashboard: show which issues are connected to an epic and how its plan is progressing


clubanderson:
## Summary

Once an issue has been planned (📋 button or plan label), the dashboard gives no way to see that the issue is connected to an epic, what state the plan is in, or how its tasks are progressing — other than opening the governor "planning" tile → Plans modal and finding it by hand. The pill on the Repositories card looks identical before and after planning, and the child tasks (beads) are invisible on GitHub entirely.

This is the visibility half of making plan mode useful; #8010 is the "nothing gets decomposed" half. Both are Phase 1 of #7993.

## Today

- Governor **planning** tile → **Plans** modal (`GET /api/plans`): epic title, status badge, source `repo#N`, `open/total tasks`.
- Row → **Plan review** modal (`GET /api/plan/{epic}`): children with retag / remove / ✓ Approve / ↩ Return to draft. Works, but because of #8010 it always reads "No children yet".
- The issue pill has no state; the linkage is one-hop (epic → one source issue); children are beads, not GitHub issues, so "which issues belong to this epic" has no answer on either side.

## Proposed (Phase 1 deliverables)

1. **Pill state chip.** When an issue has an epic, its 📋 button is replaced by a chip that opens `openPlanReview(epicId)` for that epic: `⧗ planning` (decompose pending), `● N tasks · review` (draft), `✓ done/total` (approved), `⚠ stuck` (decompose failed / revision cap, see #8010). Data is already in `/api/plans` (`issueRepo`, `issueNumber`, `planStatus`, `pendingDecompose`, `childrenOpen`, `childrenTotal`); `renderRepos` joins on `repo#number`. The plans list needs to ride the status payload (or be fetched alongside it) so the card does not make N requests.
2. **Review modal shows who has each child**: claiming agent, PR link when one exists, status colour — all already on the bead.
3. **Planning tile lists what is waiting on a human**: designs awaiting approval (Phase 2), task lists awaiting approval, stuck epics — not a bare count that reads as "working".
4. **Mirror the plan onto GitHub** (opt-in, `planning.mirror_to_issue`): post the approved task list as a `- [ ]` checklist comment on the source issue, edited in place as children close. GitHub renders tasklist progress natively, so contributors see the linkage without the dashboard. Real sub-issues per child are deferred to Phase 2 of #7993.

## Out of scope

- Turning children into GitHub issues.
[...]

Issue #8010, opened 2026-09-21
🐛 bug: "Plan this issue" and the plan label never decompose: nothing turns the architect's reply into child beads


Danathar:
## Summary

"Plan this issue" (dashboard) and the `plan`/`epic` label trigger both mint an epic and kick the architect, but no code path turns the architect's reply into child beads. `planning.Decompose` and `planning.DecomposeFromOutput` are the only functions that create children and clear `decompose_pending`, and their only production caller is the `bd decompose` CLI. Every epic created from the button or the label sits in the PLANNING tile as pending until an operator runs the CLI by hand with the plan text pasted in. The label trigger additionally re-kicks the architect for the same epic on every eval cycle, because it is written to keep kicking until the marker clears and nothing ever clears it.

This is the live bug that Phase 1 of #7993 exists to fix. The RFC builds on top of it; this issue is just the bug.

## Evidence (origin/v4 at 00eeeb6ce, 2026-09-21)

This is from code analysis only. I have not yet reproduced it on a running hub by clicking the button or labeling an issue and watching the epic stay pending. If someone has seen the button produce children on a live L5+ hub, please say so and I will re-check.

- `src/pkg/planning/decompose.go:148` (`Decompose`) and `:174` (`DecomposeFromOutput`) are the only writers of child beads, and `DecomposeFromOutput` is the only place `decompose_pending` is cleared (line 260). `git grep` for either function outside `_test.go` files returns exactly one caller: `src/cmd/bd/decompose.go:62`.
- `src/pkg/agentparse/agentparse.go:283` `ParseTaskList` has one non-test caller, `DecomposeFromOutput`. Nothing tails the architect's session or its kick reply looking for a task list.
- `src/pkg/planning/decompose.go:120` the kick prompt ends with "Do NOT open PRs or create beads — output only the plan." So the architect is explicitly told not to close the loop itself.
- `src/cmd/bd/decompose.go` documents its planner as "a passthrough of the file/stdin content; ... no agent runs." The CLI expects the operator to supply the plan text.
- `src/pkg/dashboard/plan_handlers.go:71` says the UI "polls GET /api/plan/{epicID} and the architect fills the plan in". `src/pkg/planning/issue.go:256` says "the pending marker stays until the architect materializes children (DecomposeFromOutput clears it)". Both describe a hand-off that has no implementation on the receiving side.
[...]

Issue #7993, opened 2026-09-21
[RFC] Design review before implementation for issues labeled `Epic`


Danathar:
> **Hold — design discussion, not a task.** This is a proposal for maintainer and community discussion. Please do not implement it, decompose it into sub-issues, or relay it to contributors until the design is settled. The `hold` label should stay applied.

**Status:** Proposal
**Scope:** `pkg/planning` label trigger and plan materialization, the dashboard plan entry point, the ACMM `plan_auto_approve` knob, planning docs
**Out of scope (deferred):** any change to how ordinary (non-epic) issues flow, new agents, the code-review path after implementation
**Prior art / related:** #5074, #5075 (the `planning:` block and `plan_from_label`), #7095 (ioscan fail_mode closed at L5+)

New to Hive? There is a **glossary in section 10**. Terms defined there are used throughout: bead, epic, Architect, ACMM level, the Hive pool, `plan_status`, `plan_auto_approve`.

## 1. Summary

Today, labeling an issue `plan` or `epic` can ask the Architect to break it into tasks. This proposal adds a **design step before the breakdown**, with a **human gate on each**.

```
  issue labeled "epic"
          |
          v
  Architect writes a design and posts it on the issue
          |
          v
  GATE 1: a human approves the design
          |
          v
  Architect turns the approved design into a task list (draft)
          |
          v
  GATE 2: a human approves the task list
          |
          v
  tasks enter the Hive pool and agents can claim them
```

The goal is to do the arguing before the code gets written, in public, where anyone can join in.

**One decision comes first.** The ask is for a mandatory human gate. But the maturity packs for levels 5 and 6 turn that gate off, and planning only runs at level 5 and above. So by default, every Hive that can use this feature is configured to skip the gate being asked for. Section 4 lays out the options.

## 2. Why

Large features get worse when the design argument happens inside the pull request. By then someone has spent a day writing it, and has an understandable reason to defend the shape they picked. Moving the argument earlier means options get compared before one is built, and blast radius gets named before it lands.

The second reason is the one the requester cares most about: **reviewing a design is a real way for new contributors to participate.** It does not require a working dev environment or knowledge of the whole codebase. It requires reading and an opinion. That is a much lower bar than a first pull request, and it is genuinely useful work. The Architect already writes good design documents. We just do not route them anywhere a person can react to them.

## 3. How it works today

The parts exist. They are not connected.

[...]

PR #8013, opened 2026-09-21
✨ feature(dashboard): show an issue's epic and its plan progress on the pill

Danathar:
## What

Once an issue has been planned, the dashboard gave no way to see that the issue is connected to an epic, what state the plan is in, or how its tasks are progressing — other than opening the governor **planning** tile → **Plans** modal and finding it by hand. The pill on the Repositories card looked identical before and after planning.

This lands **Phase 1 deliverables 1–3** of #7993, as described in #8011.

### 1. Pill state chip

An issue that **has** an epic gets a state chip in place of its 📋 button, and clicking it opens `openPlanReview(epicId)` for that epic:

| Chip | Meaning |
|---|---|
| `⧗ planning` | An epic exists and is queued for the architect to decompose. |
| `● N tasks · review` | Plan drafted; its N tasks are withheld from the fleet until you approve it. |
| `✓ done/total` | Approved and executing — how many of its tasks are finished. |
| `⚠ stuck` | Queued for the architect past the stuck threshold with nothing built. |

The chip is drawn from the plan listing that now **rides the status payload** (`planning.plans`, the same set `GET /api/plans` returns), joined onto the pills by `issueRepo#issueNumber` — as the issue asked, so a card with forty issue pills makes zero extra requests. The listing is not truncated: a silently-capped list would drop the chip from a planned issue and make the card lie again.

The chip is shown whatever the ACMM level. An epic minted at L5 does not stop existing if the level later drops, and hiding it would strand the plan.

### 2. Review modal shows who has each child

`PlanChild` now carries `actor`, `prRef` and `prUrl`, all read off the bead — the actor is the lane the task was decomposed into, and the PR is whatever the working agent recorded (`pr_url`, `pr_repo` + `pr_number`, or an `external_ref` that is itself a PR URL, matching what `pkg/retro` already reads). Each row gets a status colour. A `repo#number` with no URL is rendered as plain text rather than linked to a guessed forge host — this package does not know the hive's forge, and a `github.com` guess is wrong on GHE.

### 3. The planning tile lists what is waiting on a human

New `stuck` state: an epic that has been `decompose_pending` for longer than `planning.DecomposeStuckAfter` (12h — three missed architect cycles at ACMM L5). Stuck plans are **subtracted from the reassuring "+N queued" suffix**, raise the tile's amber, sort first in `ListPlans`, and are **named** in the tile tooltip and at the top of the Plans modal rather than being folded into a count that reads as "working".

That state is not cosmetic: while #8010 is open, "queued forever" is the *normal* outcome of planning an issue, and `BuildPlanning` counted it as work in flight. The elapsed-time derivation is the only signal available today; when #8010 lands an explicit failure marker it should feed `DecomposeStuck` **alongside** this fallback, not replace it — an architect that answers and is never heard from again leaves no marker at all.

## What is NOT here

**Deliverable 4** (mirror the approved task list onto the source issue as a `- [ ]` checklist comment, behind `planning.mirror_to_issue`) is deliberately left out. It is a GitHub *write-path* feature rather than a dashboard one — it needs an edit-comment client path, a marker-comment upsert, and an eval-cycle call site that posts to real issues — and it is inert until #8010 makes decomposition produce children at all. It remains the open piece of #8011; happy to take it as a follow-up, or fold it in here if a maintainer prefers one PR.

Also out of scope per the issue: turning children into GitHub issues, any change to how the plan is produced (#8010), and the design gate (#7993 Phase 2). `planWaitingList` returns `{label, title}` entries specifically so Phase 2's "design awaiting approval" rows can join it without reshaping callers.

## Testing

```
$ cd src && go build ./... && go vet ./pkg/planning/... ./pkg/dashboard/...
$ go test ./pkg/planning/... ./pkg/dashboard/...
ok      github.com/hivecommons/hive/pkg/planning        0.143s
ok      github.com/hivecommons/hive/pkg/dashboard       89.0s
$ node .github/scripts/check-inline-js.js src/pkg/dashboard/static/index.html
All script blocks passed syntax and no-undef checks.
$ python3 src/scripts/check-docs-links.py src/docs        # all relative links and anchors resolve
$ python3 src/scripts/check-docs-citations.py src/docs    # 881 file:line references verified
$ bash src/scripts/check-api-reference-citations.sh       # all references verified
```

New tests, each of which fails on the parent commit:

- `src/pkg/planning/plan_progress_8011_test.go` — the stuck threshold (not before it, at it, never once the pending marker clears, never for a nil epic), stuck sorting first ahead of a draft, the issue linkage the chip join needs, and `childPR` across all three recorded shapes plus its rejections (an *issue* `external_ref` must not resolve as a PR; a repo+number must not get a synthesized URL).
- `src/pkg/dashboard/plan_progress_ui_8011_test.go` — `buildPlanningAt` carrying the listing and a `stuck` count derived from the same read (so they cannot disagree), the JSON round-trip the SPA actually reads, `plans` omitted on a hive with no plans, and the SPA wiring for all three deliverables (every helper defined, the index built from the status payload rather than a per-pill fetch, the chip replacing the button, the four chip states, the child actor/PR/colour, and both governor tiles consulting the waiting-on-a-human helper).

`src/docs/api-reference.md` carried 11 `server.go:NNN` citations that my `FrontendPlanning` fields shifted; remapped with `check-api-reference-citations.sh --fix` in this commit, per CONTRIBUTING's cite-remap rule.

Refs #8011, #7993.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `f9527ee23`


Danathar:
Closing this as superseded by #8014, which the issue author opened for the same issue and which covers strictly more of it.

Overlap check before closing — #8014 lands everything here and then some:

| #8011 deliverable | this PR | #8014 |
|---|---|---|
| 1. Pill state chip joined off the status payload | ✅ | ✅ (`planning.issues[]`) |
| 2. Review modal shows claimant + PR link + status colour | ✅ | ✅ (`claimedBy` / `prUrl`) |
| 3. Planning tile names what is waiting on a human | ✅ | ✅ (`planning.waiting_on_human[]`) |
| 4. `planning.mirror_to_issue` checklist comment | ❌ deferred | ✅ |

Two further reasons not to keep this one alive:

- #8012 (the #8010 fix) merged into `v4` while this was in review and introduced an **explicit** decompose-failed marker with a 30m/3-attempt throttle. My `⚠ stuck` state here is derived from elapsed time (`DecomposeStuckAfter`, 12h) because no such marker existed when I wrote it. Rebasing would mean reimplementing state that has already shipped, from a weaker signal. #8014 reads the real marker via `PlanStateOf`.
- That same merge is why this branch now shows a conflict against `v4` (`plan_list.go`, `index.html`, `status_builder.go`).

One small thing from here that may be worth folding into #8014 if you want it — `childPRURL` there returns the bare URL, so the review modal renders a full URL as the link text. Parsing `owner/repo#N` out of the `.../owner/repo/pull/N` path (host-agnostic, so it works on GHE as well as github.com) makes the row read `acme/widgets#7` instead. Entirely cosmetic; the substring check in #8014 is otherwise equivalent and simpler.

No action needed on my side — #8014 is the better base.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `f9527ee23`

clubanderson:
Thanks — this is a good read of #8011, and we collided: I had #8014 in flight for the same issue, stacked on #8012 (the #8010 fix), which merged an hour ago and is why this branch now shows CONFLICTING on `plan_list.go`, `status_builder.go` and `index.html`.

Where the two differ, and what I did about it:

- **Stuck semantics.** #8012 landed an explicit marker: the label path re-kicks ≥30m apart, caps at 3 attempts, then sets `decompose_failed`. Your point that an elapsed-time fallback is still needed is right — a *button*-kicked epic is kicked once and never re-kicked, so it can never hit the cap. I folded that into #8014 as `planning.DecomposeStuckAfter` (12h) inside `DecomposeStuck()`, which is `failed || (pending && last kick ≥ 12h ago)`, credited to you in the const comment. So both signals feed the same chip, as you suggested.
- **Deliverable 4 (mirror).** #8014 includes it behind `planning.mirror_to_issue` (off by default), posting on approve via the existing `GHClient.CreateIssueComment` with a `#plan=<epic>` deep link back. Snapshot-on-approve, not an upsert — I took your "not the source of truth" framing and wrote it into the comment footer.
- **Chip vocabulary** is effectively the same (`⧗ / ● N / ▶ d/N / ✓ / ⚠`); yours shows the chip at any ACMM level, and I agree with that reasoning — #8014 does too (the chip is keyed off `planning.issues`, not `planningAllowed()`).
- **PR shape on children:** you also read `pr_repo`+`pr_number`; #8014 reads `pr_url` and a PR-shaped `external_ref`. If `pkg/retro` really emits `pr_repo`/`pr_number`, that's a small follow-up worth adding to `childPRURL` — want to open it against v4 after #8014 lands?

Given the conflict and the overlap I'd propose closing this one in favour of #8014 rather than asking you to rebase across #8012. Your call — if you'd rather carry it, say so and I'll close #8014 instead and review yours once it's rebased.

Issue #8034, opened 2026-09-21
[v6] standby S0: design doc for RFC #7629 (protocol, config, state machine, open questions)


Danathar:
Step S0 of RFC #7629 (standby contributors), target v6. This is the design gate: no production code, docs only.

## Deliverables

1. `docs/rfc-7629-standby-contributors.md`: the RFC text from #7629 committed in tree, the way #6825 was (`docs/rfc-6825-capability-aware-contributor-task-assignment.md`).
2. `src/docs/design/standby-contributors.md`, plus an index entry, covering:
   - **Relay protocol.** The standby-mode messages: how a relay declares "standing by for lane X on hive Y", what configuration it reports (model, harness, reasoning effort, advisor; reuse the `auth_response` fields in `src/pkg/dashboard/contribute_ws.go`), how the hub acknowledges and how the relay leaves standby. All additive and `omitempty`, so an older relay or hub is unaffected.
   - **Config schema.** The per-lane `standby:` block (`enabled`, `min_model_capability` default T1, `daily_cap_per_contributor`) and the approved-contributor list, with validation rules and the rule that nothing in the UI suggests lowering the floor.
   - **State machine.** Lane paused for budget, standby offered, dispatched by hand, PR opened hold-gated, outcome recorded, contributor suspended or cleared. Which component owns each transition (scheduler, hub, relay, dashboard).
   - **Matching.** Lane floor versus item tier, `unknown` never qualifies, the whole configuration is matched, daily cap accounting.
   - **Suspend rule.** N consecutive closed-unmerged donated PRs (default 2) suspends that configuration for that hive until the owner clears it. Where the outcome is recorded and how "reworked by a human first" is detected.
   - **Answers to the RFC's open questions:** floor per lane or per repo; whether donated PRs count against the hive's PR budget and cadence; private repos and what the approval screen says; whether hold-gated review of a T3 item is cheaper than doing it by hand, and how narrow the T3 list is.
   - **Phase map.** S1 to S8 as listed on #7629, each naming the files it touches, so later issues can be filed from the doc.
   - **Testing.** The doc ships with the test plan, and each of S1 to S8 cites which parts it must add:
[...]

Issue #7629, opened 2026-09-18
[RFC] [v6] Standby contributors: a lane paused for budget hands its queue to volunteers, behind a model floor


Danathar:
> **Hold - design discussion, not a task.** This is a proposal for maintainer discussion. Please do not implement it, decompose it into sub-issues, or relay it to contributors until the design is settled.

**Status:** Proposal  
**Target:** Hive v6  
**Scope:** Contributor relay (standby mode), scheduler budget-pause signalling, per-lane model floor, donated-PR marking and outcome tracking, dashboard  
**Initially out of scope:** Sharing credentials or tokens in any form, automatic dispatch (step 3 of the rollout), benchmark-based tiering beyond what #6825 already proposes  
**Related design:** [RFC #5698: Backend capacity, model inventory, pacing, and placement](https://github.com/hivecommons/hive/blob/v5/docs/rfc-5698-backend-capacity-model-inventory-placement.md) · [RFC #6825: Capability-aware contributor task assignment](https://github.com/hivecommons/hive/blob/v5/docs/rfc-6825-capability-aware-contributor-task-assignment.md)

---

This is a feature idea, not a bug. The outcome I want: **when a hive owner runs out of tokens, contributors who volunteered ahead of time can pick up that lane's queue on their own machines, and a cheap model can never quietly replace a good one.**

## What an operator sees today

I run out of subscription tokens. The quality lane pauses. The queue keeps growing. Nothing happens until I pay again, and for a hobby project that can be weeks.

At the same time, the contributor relay already lets someone else run an agent on their own machine, with their own model and their own credentials, and hand the result back as a normal PR. But they have to come by and pick a task. Nothing tells them a lane is stuck, and there is no way to say "use my machine whenever this hive is short."

## What I am proposing

The missing piece between the two: a lane that is paused for budget offers its queue to approved standby contributors, if their model clears the bar for that lane.

Two things it is not:

- **Not sharing tokens or keys.** Nobody's credentials move. A contributor donates a running agent, never a key. That is the relay's existing rule and it does not change.
- **Not a way around review.** Donated work is hold-gated no matter what the hive's ACMM level is, and it lands as a PR from the contributor's own account. It gets reviewed like any outside contribution.

## The risk: people will donate pennies, not quarters

[...]

Issue #5491, opened 2026-09-01
🔧 repo-root docs/ is never link-checked — docs-link-check.yml covers only src/docs/ and the wiki


clubanderson:
Deferred from #5398's path-filter sweep, which called it "worth its own issue". It still has none. Filing narrowly; not fixing it here.

## The gap

`.github/workflows/docs-link-check.yml` runs `check-docs-links.py` over exactly two trees:

```
python3 src/scripts/check-docs-links.py src/docs
python3 src/scripts/check-docs-links.py src/deploy/data/wiki --vault-root
```

and its `paths:` filter, on both `push` and `pull_request`, lists only `src/docs/**`, `src/deploy/data/wiki/**`, the two script paths, and the workflow itself.

**Repo-root `docs/` appears in neither.** It holds 13 markdown files — `architecture.md`, `troubleshooting.md`, `federation-design.md`, `backend-setup.md`, `HUB_DISASTER_RECOVERY.md`, `advisory-staleness.md`, `outreach-antispam.md`, `LEGACY-v1-architecture.md` among them. A PR that breaks a relative link or a heading anchor in any of them runs no link check at all, and no later PR catches it either, because nothing ever points the checker at that tree.

This is the same shape as #5309, where the wiki vault was checked by nothing, and the same shape as the `dashboard/openapi.json` exemption named in #5388: the guard exists, is correct, and is simply never aimed at the surface.

## Why it is worth doing now

#5398 deferred this partly on the expectation that widening the scope would be "likely red on arrival" — roughly 20 never-checked files. That is not the case today. Against the current tree:

```
$ python3 src/scripts/check-docs-links.py docs
checked 13 files under docs
all relative links and anchors resolve
```

So the fix can land **green**, as a pure scope widening with no documentation rewrites. That will not stay true indefinitely — every unchecked PR is a chance for it to go red — which is the argument for taking it while it is cheap.

## Suggested fix

1. Add `docs/**` to both `paths:` filters in `docs-link-check.yml`.
2. Add a step running `python3 src/scripts/check-docs-links.py docs`.

Plain mode, not `--vault-root`: root `docs/` is served from the repo, so parent-relative links resolve normally, unlike the flat `/data/wiki/` layout #5309 covered.

Per the standard #5388 set, the PR should demonstrate the new step failing — break one relative link or anchor under `docs/`, show the step red, restore it, show green.

## Scope

[...]

PR #8038, opened 2026-09-21
📖 docs: standby contributors design (S0) and RFC #7629 in tree

Danathar:
Step **S0** of [RFC #7629](https://github.com/hivecommons/hive/issues/7629) (standby contributors), the design gate the RFC's own rollout asks for before any code. **Docs only — no production code, no behaviour change.**

## What lands

| File | What it is |
| --- | --- |
| `docs/rfc-7629-standby-contributors.md` | The RFC text from #7629 committed verbatim, so it can be reviewed as a diff and amended in place rather than edited in an issue body — the same treatment [RFC #6825](https://github.com/hivecommons/hive/blob/v5/docs/rfc-6825-capability-aware-contributor-task-assignment.md) got. |
| `src/docs/design/standby-contributors.md` | The design document. |
| `src/docs/design/README.md` | Index entry, with status. |

## The design, in short

**Relay protocol.** Four additive messages — `standby_declare`, `standby_ack`, `standby_release`, `standby_state` — gated on one new `standby_v1` capability token in `RELAY_CAPABILITIES` / `serverCapabilities()`. The configuration a relay stands by with is the five `auth_response` fields that already exist (`cli_backend`, `model`, `reasoning_effort`, `advisor_model`, `advisor_reasoning_effort`), re-reported rather than inherited so a CLI restart on a different model refreshes what the hub matches on. `contributorProtocolVersion` is **not** bumped: the precedent is `token_refresh_failed`, both peers already ignore unknown message types (the hub's switch has no default case, the relay logs and continues), and `standby_state` is sent *only* to connections in standby — so an older relay or hub sees no new traffic at all. The doc carries the full mixed-version matrix.

Standby is **connection state, not durable state**: approval is durable and lives in config, availability is a property of a machine that is on right now. A hub restart forgets it and relays re-declare on reconnect. That is the correct semantics — a hub that remembered standby would offer work to machines that are off.

**Config schema.** Per-lane `standby:` (`enabled`, `min_model_capability` default `T1`, `daily_cap_per_contributor` default `0`) under `agents.<lane>`, plus hive-wide `hub.standby_contributors` and an owner-authored `standby_model_tiers` that ships **empty** — so the out-of-the-box outcome on every hive is "0 qualify" until an owner writes the mapping deliberately. Six validation rules, each a named unit test; `unknown` is rejected as a floor at parse, and `enabled: true` with an empty approved list is a load error rather than a silently inert setting.

**The anti-nudge rule, made structural.** The RFC's central worry is an owner lowering the bar until something qualifies. The doc's answer is that **the floor has no write API** — readable everywhere, writable only by editing `hive.yaml`, asserted by a test that no route writes it. A rule about what the UI *says* can be violated by a well-meant copy change; an absent endpoint cannot. `standby_ack` rejections carry the reason class and never the delta, so the contributor-facing surface cannot supply lobbying material either.

**State machine.** Lane and contributor-configuration machines with an owner named per transition (governor / hub / relay / dashboard). One consequence is worth flagging for review: the contributor's agent opens the PR with the contributor's own credential, so it does **not** travel the `pr_request_watcher` path that applies `hold` today. The hub already verifies the reported PR (`verifyReportedPR`) and must apply the label itself there — **the first hub-side GitHub label write on the contributor path**, deliberate because a gate the client applies is not a gate.

**Matching and the suspend rule** as pure functions in a new `pkg/standby`: whole-configuration tier keys (so nobody can offer Opus and run something else — changing the reasoning effort changes the key), `unknown` never qualifies both structurally and via an explicit guard whose test fails when the guard is removed, and the daily cap on the existing trailing-window convention (`rateLimitDayWindow`) rather than a calendar bucket. The suspend rule walks an append-only outcome ledger newest-first; "reworked by a human first" reuses `holdguard`'s commit-author drift rather than inventing a second answer to "who touched this after we recorded it".

**The four open questions each get one written answer:**

1. **Floor per lane or per repo?** Per lane — that is where policy is already written (`agents.<name>`, `LaneConfig`, `filterByLane`). The cost is named, and the escape hatch is the existing per-repo agent (`agents.<name>.repos`).
2. **Do donated PRs count against the hive's PR budget and cadence?** Split: **yes** against cadence and queue accounting (add no exemption — that is what stops a generous donor flooding a repo), **no** against the token budget (no hive tokens were spent; counting them would re-pause the lane for spend that did not happen).
3. **Private repos / the approval screen.** Opt-in, default off (`hub.standby_allow_private_repos`), refusal reason `private_repo`, failing closed when visibility cannot be read — plus the exact sentence the approval control must state about read-access equivalence.
4. **Is hold-gated review of a T3 donated PR cheaper than doing it by hand?** Often not, and the doc says so about the RFC's own changelog-entry example. The break-even is not size but whether the reviewer must re-derive the work, so T3 eligibility is defined by verifiability, the shipped list is **empty**, three candidate entries are named with their signals, and the evidence that revisits it is reviewer-minutes recorded in the S5 runbook.

**Phase map S1 to S8**, each row naming the files it touches so the follow-on issues can be filed straight from the table, and the **test plan** each phase must satisfy: unit tests in `pkg/`, protocol and UI against the existing `contribute_ws_*_test.go` and `contributor-relay.test.js` fakes, the live-hive runbook recorded verbatim, and one keyless standby scenario added to the backend-smoke stub tier that v2-ci already runs on every PR — real relay against a real hub, no new CI hardware and no model spend.

## Testing

Not run (docs only) — no Go or JS changed. Verified instead with the repo's own docs gates, locally:

```
$ python3 src/scripts/check-docs-links.py src/docs
checked 171 files under src/docs

$ python3 src/scripts/check-docs-links.py docs
checked 20 files under docs

$ python3 src/scripts/check-docs-links.py . --no-recurse
checked 9 files under .

$ python3 src/scripts/check-docs-citations.py src/docs
docs citations: 940 repo-local file:line reference(s) verified.
```

Every `file:line` cite in the new design doc was additionally checked by symbol, not just by line range, per [CONTRIBUTING's citation rule](https://github.com/hivecommons/hive/blob/v6/CONTRIBUTING.md#fileline-citations-in-docs). (Confirmed the citation checker actually covers the new file by breaking one cite and watching it go red, then restoring it.)

`no-changelog`: docs-only, per CONTRIBUTING's changelog-fragment scope.

## For the reviewer

The acceptance criteria on #8034 ask for a maintainer review and one written answer per open question. The four answers are recommendations to accept or reject, not settlements — **#8034 stays open** (`Refs`, not `Closes`) until they have been reviewed and S1 to S4 are filed from the phase map. The `hold` on #7629 comes off when this merges.

Refs #7629
Refs #8034

---
🐝 **Hive Agent**: `contributor` | **SHA:** `a6cdd77cb`


Danathar:
One thing noticed while verifying, out of scope for this PR but worth a maintainer's eye: `.github/workflows/docs-link-check.yml` filters both its `push` and `pull_request` triggers to `v4` and `v5` only, so **no docs PR on the `v6` line runs the relative-link, anchor, or `file:line` citation gates** — including this one (the check does not appear in the list above). That is the same shape as #5491 and #7829, where a surface was simply never aimed at.

That is why the verification in the description was run locally rather than cited from CI. Adding `v6` to those two branch filters is a one-line change; I have deliberately kept it out of this diff so the design gate stays docs-only, and am happy to file it as a separate issue if that is wanted.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `a6cdd77cb`

